PR #35220, opened 2025-10-04
Catch Throwables in DefaultBuildEventsListenerRegistry

rhernandez35:
As is, an OperationCompletionListener that throws Throwable or Error from its `onFinish` method its not appropriately quarantined. Something worse happens: the unhandled Throwables throw out of the task defined in DefaultBuildEventsListenerRegistry$AbstractListener and are lost. They're logged to the daemon's stderr by the default thread uncaught exception handler, but not in the build output. The listener will continue to be invoked despite the failure.

### Context
I had an OperationCompletionListener that was failing to load a class first referenced in its `onFinish` method. The resulting NoClassDefFoundError was impossible to find unless you went looking through daemon stderr logs for some reason. There were no indications that the listener was failing; I found the bug by chance.

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/language-java/src/crossVersionTest/groovy/org/gradle/api/internal/tasks/compile/tooling/r51/JavaCompileTaskOperationResultCrossVersionTest.groovy
+++ b/platforms/jvm/language-java/src/crossVersionTest/groovy/org/gradle/api/internal/tasks/compile/tooling/r51/JavaCompileTaskOperationResultCrossVersionTest.groovy
@@ -22,7 +22,6 @@ import org.gradle.integtests.tooling.fixture.ToolingApiSpecification
 import org.gradle.language.fixtures.HelperProcessorFixture
 import org.gradle.tooling.events.OperationType
 import org.gradle.tooling.events.task.java.JavaCompileTaskOperationResult
-import spock.lang.Issue
 
 import java.time.Duration
 


hegyibalint:

hegyibalint:
Thank you for your interest in Gradle.

---

We think that catching `Throwables` is not the right solution here, and a potential solution needs more discussion first. 

The fact that the error was not propagated to the build output is the main issue here, and we think we should approach this problem from that angle first. 

Please open an issue first, so we can have more discussion there.

rhernandez35:
Opened https://github.com/gradle/gradle/issues/35518.

Why is catching Throwable the wrong answer here? The only other alternative is to set an UncaughtExceptionHandler (which is invoked implicitly by Thread's own `catch Throwable`), but once we've escaped the gradle worker context there's no way to report the failure.

The current behavior of doing nothing and reporting nothing seems wrong to me, and I know of no way other than catching a Throwable to handle it.

eskatos:
Thank you for the issue. We will reconsider this contribution.

platforms/jvm/language-java/src/crossVersionTest/groovy/org/gradle/api/internal/tasks/compile/tooling/r46/JavaCompileTaskOperationResultCrossVersionTest.groovy
```diff
@@ -0,0 +1,76 @@
...
+            @Helper class SomeClass {}
+        """
+        def processorProjectDir = file("processor")
+        def fixture = new HelperProcessorFixture()
+        fixture.writeApiTo(processorProjectDir)
+        fixture.writeAnnotationProcessorTo(processorProjectDir)
+        fixture.writeSupportLibraryTo(processorProjectDir)
+    }
+
+    @TargetGradleVersion("<5.1")
```
eskatos: 💅 This can be merged with the class level annotation
```java
@TargetGradleVersion(">=4.6 <5.1")
```

jbartok: I know, I intentionally didn't do that. My reasoning was that the meaning of the class level annotation should be "these tests should be run from version X onwards" and then the fact that some or the other test applies only until a certain version B, that is a peculiarity of the test. Although, this logic would make more sense if there were multiple tests... I don't feel strong about it, will change.

jbartok:
@bot-gradle test this

PR #35238, opened 2025-10-07
Deprecate using legacy Usage attribute values

jvandort:
These legacy attribute values were deprecated in Gradle 5.6, with the intention to add deprecation warnings in 6.0 Better late than never, so we're adding the deprecations now.

Trying to use any of these legacy Usage values in build logic will result in a deprecation warning, as the mutable attribute container emits a warning if it encounters a legacy value.
Additionally, we specially handle legacy Usage attributes in published GMM parsing, as published GMM with these attributes must always be handled in a backwards compatible manner.
In all other cases, traces of these legacy attributes have been removed from the core attribute container implementation. Attribute containers should be agnostic about the attributes they hold, and should not treat certain attributes specially compared to others

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -48,6 +48,25 @@ The method on the link:{javadocPath}/org/gradle/api/tasks/wrapper/Wrapper.html[W
 
 Use link:{javadocPath}/org/gradle/api/tasks/wrapper/Wrapper.DistributionType.html#values()[`Wrapper.DistributionType.values()`] to obtain the available distribution types instead.
 
+[[deprecate_legacy_usage_values]]
+==== Deprecated legacy Usage attribute values
```
```suggestion
==== Deprecated legacy `Usage` attribute values
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -48,6 +48,25 @@ The method on the link:{javadocPath}/org/gradle/api/tasks/wrapper/Wrapper.html[W
...
+
+Since Gradle 5.6, the link:{javadocPath}/org/gradle/api/attributes/Usage.html[`Usage`] attribute was split into a separate link:{javadocPath}/org/gradle/api/attributes/LibraryElements.html[`LibraryElements`] attribute.
```
```suggestion
Since Gradle 5.6, the link:{javadocPath}/org/gradle/api/attributes/Usage.html[`Usage`] attribute has been split into a separate link:{javadocPath}/org/gradle/api/attributes/LibraryElements.html[`LibraryElements`] attribute.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -48,6 +48,25 @@ The method on the link:{javadocPath}/org/gradle/api/tasks/wrapper/Wrapper.html[W
...
+In the JVM ecosystem, `Usage` controls whether a variant is intended for compilation or runtime, while `LibraryElements` controls the format of the artifact (for example, JAR or classes directory).
```
```suggestion
In the JVM ecosystem, `Usage` indicates whether a variant is intended for compilation or runtime, while `LibraryElements` specifies the format of the artifact (for example, a JAR file or a classes directory).
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -48,6 +48,25 @@ The method on the link:{javadocPath}/org/gradle/api/tasks/wrapper/Wrapper.html[W
...
+
+In order to ease the migration, Gradle automatically maps legacy `Usage` values to the corresponding `Usage` and `LibraryElements` values:
```
```suggestion
To ease migration, Gradle has automatically mapped legacy `Usage` values to their corresponding `Usage` and `LibraryElements` pairs:
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -48,6 +48,25 @@ The method on the link:{javadocPath}/org/gradle/api/tasks/wrapper/Wrapper.html[W
...
+|===
```
lkasso, line 59:
```suggestion

|===
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -48,6 +48,25 @@ The method on the link:{javadocPath}/org/gradle/api/tasks/wrapper/Wrapper.html[W
...
+| Legacy `Usage`           | Replaced `Usage` | Replaced `LibraryElements`
+| `java-api-jars`          | `java-api`       | `jar`
+| `java-api-classes`       | `java-api`       | `classes`
+| `java-runtime-jars`      | `java-runtime`   | `jar`
+| `java-runtime-classes`   | `java-runtime`   | `classes`
+| `java-runtime-resources` | `java-runtime`   | `resources`
+|===
+
+Starting in Gradle 10, Gradle will no longer perform this automatic mapping when a legacy `Usage` value is added to an `AttributeContainer` in build logic.
```
```suggestion
Starting with Gradle 10, this automatic mapping will no longer occur when legacy `Usage` values are added directly to an `AttributeContainer` in build logic.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -48,6 +48,25 @@ The method on the link:{javadocPath}/org/gradle/api/tasks/wrapper/Wrapper.html[W
...
+In order to maintain backward compatibility with published metadata, Gradle will continue to translate legacy `Usage` values found in published Gradle Module Metadata.
```
```suggestion
To maintain backward compatibility, Gradle will continue translating legacy `Usage` values found in published Gradle Module Metadata.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
tresat: ❗ Doesn't the change to `JavaEcosystemAttributesDescriber` start to remove this mapping _now_?

subprojects/core/src/main/java/org/gradle/api/internal/attributes/UsageCompatibilityHandler.java
```diff
@@ -83,19 +76,38 @@ private static <T> Attribute<T> getAs(Attribute<?> key, Class<T> type) {
...
+            usageValue = usage.getName();
+        } else if (key.getName().equals(Usage.USAGE_ATTRIBUTE.getName())) {
+            usageValue = value.toString();
+        } else {
+            return null;
+        }
+
+        if (getReplacementUsage(usageValue) != null) {
+            return usageValue;
+        }
```
tresat: ❓ Is this correct?  It reads weird: "If there is a non-null replacement usage for the given usageValue, then ignore it and return the given usageValue"?

subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultMutableAttributeContainer.java
```diff
@@ -174,6 +179,20 @@ private void assertAttributeValueIsNotNull(@Nullable Object value) {
...
     }
 
+    @SuppressWarnings("deprecation")
+    private static <T> void maybeWarnOnLegacyUsageValue(Attribute<T> key, T value) {
+        String legacyUsageValue = UsageCompatibilityHandler.maybeGetLegacyUsageValue(key, value);
+        if (legacyUsageValue != null) {
+            // In Gradle 10, we can remove deprecation entirely instead of making it an error.
+            DeprecationLogger.deprecateAction("Declaring a Usage attribute with a legacy value")
+                .withContext("A Usage attribute was declared with value '" + legacyUsageValue + "'.")
+                .withAdvice("Declare a Usage attribute with value '" + UsageCompatibilityHandler.getReplacementUsage(legacyUsageValue) + "' and a LibraryElements attribute with value '" + UsageCompatibilityHandler.getLibraryElements(legacyUsageValue) + "' instead.")
```
tresat: 💭 There shouldn't be a need to calculate the replacement usage twice in this method.

subprojects/core/src/main/java/org/gradle/api/internal/attributes/UsageCompatibilityHandler.java
tresat: 🤔 This and `getReplacementUsage` seem like they ought to be clearer.

`maybeGetLegacyUsageValue` maybe gets the current value (that it is given) and maybe ignores it to return null (iff the attribute is a `USAGE_ATTRIBUTE).  This is called on every attribute though?  Why not just on the usage attributes?

I think `maybeWarnOnLegacyUsageValue` would be clearer as something like:

```
if (isUsageAttribute(key)) {
   String replacement = maybeGetReplacementUsage(value);
   if (replacement != value) {
     // deprecation warning/failure
   }
}

File: subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultMutableAttributeContainer.java [hunk 10]
```diff
@@ -192,7 +211,27 @@ public Provider<Map<Attribute<?>, AttributeEntry<?>>> getEntriesProvider() {
...
+    /**
+     * Concatenates the given entry to the given attributes, taking care of legacy {@link Usage} values.
+     * <p>
+     * Starting in Gradle 10, we can replace this with a simple call to {@link AttributesFactory#concat(ImmutableAttributes, Attribute, Isolatable)}.
+     */
+    @SuppressWarnings("deprecation")
+    private <T> ImmutableAttributes concatEntry(ImmutableAttributes attributes, AttributeEntry<T> entry) {
+        Attribute<T> key = entry.getKey();
+        if (key.equals(Usage.USAGE_ATTRIBUTE) || key.getName().equals(Usage.USAGE_ATTRIBUTE.getName())) {
+            return attributesFactory.concatPotentiallyLegacyUsage(attributes, key, entry.getValue());
```
tresat: ❓ Why `potentially`?  Aren't we sure at this point it's a `USAGE` attribute?

subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultMutableAttributeContainer.java
```diff
@@ -192,7 +211,27 @@ public Provider<Map<Attribute<?>, AttributeEntry<?>>> getEntriesProvider() {
     public ImmutableAttributes asImmutable() {
         Map<Attribute<?>, AttributeEntry<?>> realizedState = doRealize(Provider::get);
         assertNoDuplicateNames(realizedState.keySet());
-        return attributesFactory.fromEntries(realizedState.values());
+
+        ImmutableAttributes result = ImmutableAttributes.EMPTY;
+        for (AttributeEntry<?> entry : realizedState.values()) {
+            result = concatEntry(result, entry);
```
tresat, line 234: 🤔 I don't like how now the mutable container now once again relies on knowing the implementation details of the immutable container.  Can't we leave this knowledge in the factory?

subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultAttributesFactory.java
```diff
@@ -83,14 +81,16 @@ public <T> ImmutableAttributes concat(ImmutableAttributes node, Attribute<T> key
...
-        } else {
-            return doConcatEntry(node, new DefaultImmutableAttributesEntry<>(key, value));
-        }
+        return doConcatEntry(node, new DefaultImmutableAttributesEntry<>(key, value));
+    }
+
+    @Override
+    @Deprecated
+    public <T> ImmutableAttributes concatPotentiallyLegacyUsage(ImmutableAttributes node, Attribute<T> key, Isolatable<T> value) {
+        return usageCompatibilityHandler.doConcat(this, node, key, value);
```
tresat, line 92: ❗  These 2 methods form a weird API.  It's now the responsibility of the caller to know whether or not they might contain a `USAGE` attribute and call the appropriate method.  This seems unreasonable - why not just route every concatenation through the compatibility handler?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/metadata/DefaultGradleModuleMetadataSource.java
```diff
@@ -111,4 +112,30 @@ public void listModuleVersions(ModuleComponentSelector selector, ComponentOverri
             versionLister.listVersions(module, metaDataArtifact, ivyPatterns, result);
         }
     }
+
+    public static void handleMavenSnapshotCompatibility(MutableModuleComponentResolveMetadata metaDataFromResource) {
```
tresat, line 116: 🤔 This seems unrelated and maybe better for a separate PR?

tresat (dismissed):

subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultMutableAttributeContainer.java
jvandort: It's potentially legacy, we indeed do know it's a usage.

I'm open to other names if you'd prefer a different one

subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultAttributesFactory.java
jvandort: The goal of this PR is to make the factory logic agnostic of individual attribute types. The factory shouldn't be silently treating some attributes differently than others, or worse yet converting one attribute to different attributes. At the moment this behavior is silent to callers who put one attribute in a container and get two different ones out.

In practice the automatic migration we added back in 5.x for the Usage attribute is a specific application of the attribute ecosystem and therefore should be handled outside the attribute container ecosystem.

This PR doesn't get us all the way there, but it does pull out logic related to the Usage attribute further to the edges.

subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultMutableAttributeContainer.java
tresat: OK, I didn't think that way at all.

`concatUsageAttribute`?

subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultAttributesFactory.java
tresat: OK, that makes sense then.  It's intentional that callers have to know to use the right method for now.  Can you just add a doc comment explaining this?

platforms/jvm/plugins-java-base/src/testFixtures/groovy/org/gradle/java/compile/AbstractJavaGroovyCompileAvoidanceIntegrationSpec.groovy
```diff
@@ -96,7 +96,10 @@ abstract class AbstractJavaGroovyCompileAvoidanceIntegrationSpec extends Abstrac
             allprojects {
                 configurations.apiElements.outgoing.variants {
                     classes {
-                        attributes.attribute(USAGE_ATTRIBUTE, objects.named(Usage, org.gradle.api.internal.artifacts.JavaEcosystemSupport.DEPRECATED_JAVA_API_CLASSES))
+                        attributes {
+                            attribute(Usage.USAGE_ATTRIBUTE, project.objects.named(Usage, Usage.JAVA_RUNTIME))
+                            attribute(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE, project.objects.named(LibraryElements, LibraryElements.JAR))
```
big-guy: Is this right? Shouldn't this still be CLASSES_AND_RESOURCES? It looks like we're adding the compile task's output and not the jar

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/parser/GradleModuleMetadataParser.java
```diff
@@ -620,11 +624,25 @@ private ImmutableAttributes consumeAttributes(JsonReader reader) throws IOExcept
...
             }
         }
         reader.endObject();
 
+        // If a legacy Usage value was encountered, only add the corresponding LibraryElements value if there is not already one provided.
+        if (libraryElementsValue != null && attributes.findEntry(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE.getName()) == null) {
+            attributes = attributesFactory.concat(attributes, Attribute.of(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE.getName(), String.class), new CoercingStringValueSnapshot(libraryElementsValue, instantiator));
+        }
+
         return attributes;
```
big-guy, line 648: I'm not sure which side of the persistent caching we are here, but this may require bumping the cached metadata versions.

e.g., imagine that Gradle 9.0.0 resolves a module with GMM that has the deprecated attribute values.  The parsed and interpreted metadata gets cached in GUH.

Gradle 9.3.0 resolves the same module but looks at the cached metadata only. If our re-interpretation of these attributes happened after this cached step, then 9.3.0 would no longer do that and would see the deprecated attribute values.

Are there any places where a newer Gradle will see metadata from an older Gradle that will not have this compatibility step done?

big-guy, line 648: Before this change, did metadata rules see the original attributes or the compatibility modified attributes? 

IOW, could a plugin exist that looks for the deprecated attributes and does something about them that will no longer see the deprecated attribute values?

platforms/jvm/plugins-java-base/src/testFixtures/groovy/org/gradle/java/compile/AbstractJavaGroovyCompileAvoidanceIntegrationSpec.groovy
jvandort: This seems incorrect. 

DEPRECATED_JAVA_API_CLASSES -> Usage=JAVA_API LibraryElements=CLASSES

I'll go through these replacements and make sure they're all correct

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/parser/GradleModuleMetadataParser.java
jvandort: This is before the persistent cache -- this parser directly reads JSON and extracts the attributes. 

Prior to this change, whenever we tried to create a new `ImmutableAttributes` instance, via the `concat` method, we would silently transform any deprecated attributes to the non-deprecated forms. 

The GMM parser always outputs an `ImmutableAttributes` instance, meaning it never contains the deprecated attribute. 

The persistent cache layer would always see the non-deprecated forms of the attribute, as it was impossible to create an `ImmutableAttributes` instance with the deprecated attribute values.

For this reason, I don't think we need to update the cached metadata versions. 

---

> Are there any places where a newer Gradle will see metadata from an older Gradle that will not have this compatibility step done?

The plan in Gradle 10 is to remove the automatic conversion of attributes in _userspace_ mutable attributes containers. So, any build logic that is using these deprecated attribute values in 9.x will get deprecation warnings. in 10.0, we will stop checking for these deprecated values and it would be possible to use them if you ignored the warnings in 9.x. We _could_ continue checking for them in 10.x+ and throw an exception instead, but the goal here was to remove this logic from core `AttributeContainer` infrastructure. An attribute container should be agnostic of the attributes being stored in it (it should not treat a Usage different from a Foo)

jvandort: Metadata rules saw the modified attributes. We read JSON directly into an `ImmutableAttributes`, which performed the conversion silently. It was impossible (or nearly impossible) to create an `ImmutableAttributes` instance _without_ the conversion.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
jvandort: It just changes the way we describe these deprecated attribute values in failure scenarios -- values which have been deprecated since Gradle 5.0. We will still print their values, but will use their "generic" descriptions, as we've just removed their special-case handling here.

This does not change behavior in the normal case -- attribute containers still perform automatic migration of legacy values, variant matching behavior shouldn't change, etc.

subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultMutableAttributeContainer.java
jvandort: One of these calculates the replacement usage, the other calculates the replacement library elements. 

In any case, this has been simplified as a result of another comment.

subprojects/core/src/main/java/org/gradle/api/internal/attributes/UsageCompatibilityHandler.java
jvandort: Looking back this could definitely be a little cleaner. I've made this a little more straightforward.

subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultAttributesFactory.java
jvandort: I've added some additional documentation to the `concatUsageAttribute` method, in additional to the existing `@deprecated` tag, describing that it implements legacy behavior that should not be relied on for new code.

subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultMutableAttributeContainer.java
jvandort: The mutable container does not rely on the implementation details of the immutable container. It leverages a widely used `AttributesFactory` API to creating `ImmutableAttributes` instances, which hides the implementation details of `ImmutableAttributes`. (nowhere in here does the mutable container know a `DefaultImmutableAttributes` backs the instance it gets back, or that that instance is cached, how it is cached, etc). 

I think this is a reasonable use of this API. 

The point of this PR was to remove as much special Usage-specific logic as possible from the `AttributesFactory`. As long as the mutable attribute container needs to do this weird `Usage` special-casing, I don't think this logic should live in the `AttributesFactory`.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/metadata/DefaultGradleModuleMetadataSource.java
jvandort: Previously `GradleModuleMetadataCompatibilityConverter` was in charge of this logic and in charge of performing mapping of legacy Usage attribute values to their modern counterparts. 

With the usage responsibilities moved to common infrastructure, it doesn't really make sense to have a separate static "compatibility converter" class to hold a single method. This change just moves the remaining single method to the one place where it is used.

jvandort:
@bot-gradle test this

big-guy (approved):

PR #35248, opened 2025-10-07
Allow configuring failure reporting behavior on DefaultRootTestEventReporter

tresat:
Introduces a flag to control whether the test reporter throws an exception when closing if test failures are detected.

This allows callers to handle test failures in a custom way, for example, the AbstractTestTask chooses to opt-out of this; while retaining a default behavior of throwing test failures for custom test tasks.

Specifically, the `createTestEventReporter` method in `TestEventReporterFactory` now accepts a `closeThrowsOnTestFailures` parameter. This parameter determines whether the reporter should throw an exception upon close if the root node has been failed. The `AbstractTestTask` is modified to handle test failures directly instead of relying on the reporter to throw an exception.

### DIFF
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
@@ -531,8 +531,8 @@ public abstract class AbstractTestTask extends ConventionTask implements Verific
 
             // Throw an exception with rendered test results, if necessary
             TestEventReporterFactoryInternal.TestReportResult testReportResults = handleCollectedResults(testCountLogger);
-            if (testReportResults instanceof TestEventReporterFactoryInternal.TestReportResult.TestFailureDetected) {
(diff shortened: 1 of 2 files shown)


platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
```diff
@@ -526,6 +528,12 @@ public void executeTests() {
                 testOutputListenerSubscriptions.removeAllListeners();
                 testListenerInternalBroadcaster.removeAll();
             }
+
+            // Throw an exception with rendered test results, if necessary
+            TestEventReporterFactoryInternal.TestReportResult testReportResults = handleCollectedResults(testCountLogger);
```
octylFractal, line 533: IMO we should get rid of `TestReportResult` and just revert to the original exception code.

tresat: I'm open to the idea, but let's do it separately afterwards.

PR #35251, opened 2025-10-08
Update Documentation: update public apis list

lkasso:
### Details

### Context
- Update public_api.adoc page with better information

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
@@ -18,192 +18,68 @@
 
 Gradle provides many public APIs that allow developers to interact with various build system components.
 
-Any class name containing `Internal`, or that is located in a package containing the segment `.internal.` is by definition not part of the public Gradle API and should not be relied upon for stable usage.
+Any class or package name containing `Internal`, or located in a package segment `.internal.`, is not part of the public Gradle API and should not be relied upon for stable usage.
+Use of Gradle internal APIs in plugins and build scripts can break builds when either Gradle or plugins change.
 
-The following table summarizes public Gradle APIs and their common use cases.
-This is *not* an exhaustive list:
+The list of public API packages is maintained in the link:https://github.com/gradle/gradle/blob/master/buildSrc/subprojects/configuration/src/main/kotlin/org/gradle/gradlebuild/public-api.kt[Gradle public API definition] (excluding any subpackages containing internal).
+It is also available below:
 
-[cols="1,1,2"]
+[source,text]
+----
+org/gradle/*
+org/gradle/api/**
+org/gradle/authentication/**
+org/gradle/build/**
+org/gradle/buildconfiguration/**
+org/gradle/buildinit/**
+org/gradle/caching/**
+org/gradle/concurrent/**
+org/gradle/deployment/**
+org/gradle/external/javadoc/**
+org/gradle/ide/**
+org/gradle/ivy/**
+org/gradle/jvm/**
+org/gradle/language/**
+org/gradle/maven/**
+org/gradle/nativeplatform/**
+org/gradle/normalization/**
+org/gradle/platform/**
+org/gradle/plugin/devel/**
+org/gradle/plugin/use/*
+org/gradle/plugin/management/*
+org/gradle/plugins/**
+org/gradle/process/**
+org/gradle/testfixtures/**
+org/gradle/testing/jacoco/**
+org/gradle/tooling/**
+org/gradle/swiftpm/**
+org/gradle/model/**
+org/gradle/testkit/**
+org/gradle/testing/**
+org/gradle/vcs/**
+org/gradle/work/**
+org/gradle/workers/**
+org/gradle/util/**
+----
+
+== Avoiding internal APIs
+
+Some older Gradle internals are still used in the wild but are **not supported** and may break between Gradle releases.
+Use the following public alternatives instead:
+
+[cols="1,3",options="header"]
 |===
-| API Package | Description | Common Use Cases
+| Internal API (do not use) | Supported alternative
 
-| *`org.gradle.api`*
-| Core Gradle API
-| - Defines key interfaces like `Project`, `Task`, and `Plugin` +
-- Used for configuring projects and tasks
+| `org.gradle.internal.reflect.Instantiator`
+| Use link:{javadocPath}/org/gradle/api/model/ObjectFactory.html[ObjectFactory] to create and wire new objects in tasks or plugins.
 
-| *`org.gradle.authentication`*
-| Authentication API
-| - Handles authentication for repositories and services +
-- Supports username/password, OAuth, and other mechanisms
+| `org.gradle.api.internal.ConventionMapping`
+| Use link:{javadocPath}/org/gradle/api/provider/Property.html[Property] and link:{javadocPath}/org/gradle/api/provider/Provider.html[Provider] to model user inputs and lazy configuration.
 
-| *`org.gradle.build`*
-| Build Lifecycle API
-| - Manages build phases and execution +
-- Useful for controlling the order of tasks
+| `org.gradle.internal.os.OperatingSystem`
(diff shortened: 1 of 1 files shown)


platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
```diff
@@ -18,192 +18,68 @@
...
-Any class name containing `Internal`, or that is located in a package containing the segment `.internal.` is by definition not part of the public Gradle API and should not be relied upon for stable usage.
-
-The following table summarizes public Gradle APIs and their common use cases.
-This is *not* an exhaustive list:
-
-[cols="1,1,2"]
+Any class or package name containing `Internal`, or located in a package segment `.internal.`, is not part of the public Gradle API and should not be relied upon for stable usage.
+Use of Gradle internal APIs in plugins and build scripts can break builds when either Gradle or plugins change.
+
+The list of public API packages is maintained in the link:https://github.com/gradle/gradle/blob/master/buildSrc/subprojects/configuration/src/main/kotlin/org/gradle/gradlebuild/public-api.kt[Gradle public API definition] (excluding any subpackages containing internal).
```
octylFractal: I don't know if we should document this here. It's an implementation detail of our build, and this link could easily break when moving it around. I think just including the list is fine.

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
```diff
@@ -18,192 +18,68 @@
...
+It is also available below:
+
+[source,text]
+----
+org/gradle/*
```
octylFractal: A couple issues:
- I think these should be listed with `.` instead of `/`.
- I think this is unclear regardless -- these are referred to as "packages" but the globs here actually refer to classes, i.e. `org.gradle.Foo` is public, but `org.gradle.something.Foo` is internal. But that's not clear from the `*` vs `**` alone. Perhaps we can give this particular entry a special note that it doesn't include subpackages?

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
```diff
@@ -18,192 +18,68 @@
...
+org/gradle/model/**
+org/gradle/testkit/**
+org/gradle/testing/**
+org/gradle/vcs/**
+org/gradle/work/**
+org/gradle/workers/**
+org/gradle/util/**
+----
+
+== Avoiding internal APIs
```
octylFractal: Is there a reason you added this section? I personally am not aware of widespread usage of these particular APIs in the wild, and the `util` package is actually public API that we don't discourage. The internal parts are in `.internal.` already.

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
lkasso: Agreed. Removed.

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
lkasso: It was removed from Grade 7 and put it back. 
Removed. TY.

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
lkasso: 1. Update the list to use . instead of /
2. Added a note in the list and after the list.

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
```diff
@@ -18,192 +18,48 @@
...
+org.gradle.testkit.**
+org.gradle.testing.**
+org.gradle.vcs.**
+org.gradle.work.**
+org.gradle.workers.**
+org.gradle.util.**
+----
+
+NOTE: The `\*` wildcard matches only classes directly in the package (e.g., `org.gradle.Foo`), but **does not include subpackages** like `org.gradle.something.Foo`.
+For subpackages, `**` must be used.
```
octylFractal:
```suggestion
For indicating that subpackages are included, `**` is used.
```

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
```diff
@@ -18,192 +18,48 @@
...
-
-| *`org.gradle.util`*
-| Utility API
-| - Provides various utility functions used internally by Gradle +
-- Used for Gradle build logic and plugin development
-|===
+Any class or package name containing `Internal`, or located in a package segment `.internal.`, is not part of the public Gradle API and should not be relied upon for stable usage.
+Use of Gradle internal APIs in plugins and build scripts can break builds when either Gradle or plugins change.
+
+The list of public API packages:
```
octylFractal: Feels a little awkward. Perhaps:
```suggestion
The list of public API packages is as follows:
```

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
```diff
@@ -18,192 +18,48 @@
...
-| Worker API
-| - Enables parallel execution of Gradle tasks +
-- Used for running tasks in worker threads
-
-| *`org.gradle.util`*
-| Utility API
-| - Provides various utility functions used internally by Gradle +
-- Used for Gradle build logic and plugin development
-|===
+Any class or package name containing `Internal`, or located in a package segment `.internal.`, is not part of the public Gradle API and should not be relied upon for stable usage.
```
eskatos: 🤔 Stating this first makes it quite unclear that there are packages without 'internal' in the name that are internal. The only valid rule is: only what is listed as public is public, everything else is internal.

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
```diff
@@ -18,192 +18,48 @@
...
+Use of Gradle internal APIs in plugins and build scripts can break builds when either Gradle or plugins change.
+
+The list of public API packages:
+
+[source,text]
+----
+org.gradle.*  (top-level classes only — excludes subpackages)
```
eskatos, line 30: ❌ This must be generated from the `PublicApi` definition. We must have only one source of truth for this. Otherwise this will be outdated as soon as we change the definition of what is public.

eskatos: I've just read on another channel that this PR is just a quick fix for 9.2.0 and that the actual generation will be put into place after that. I'm fine with that as long at this PR clearly states this.

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
lkasso: Updated accordingly.

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
lkasso: Ack

platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/public_apis.adoc
```diff
@@ -18,192 +18,50 @@
...
+org.gradle.testkit.**
+org.gradle.testing.**
+org.gradle.vcs.**
+org.gradle.work.**
+org.gradle.workers.**
+org.gradle.util.**
+----
+
+NOTE: The `\*` wildcard matches only classes directly in the package (e.g., `org.gradle.Foo`), but **does not include subpackages** like `org.gradle.something.Foo`.
+For indicating that subpackages are included, `**` is used.
```
jvandort:
```suggestion
Conversely, the `**` wildcard matches the noted package and its subpackages.
```

jvandort:
@bot-gradlt test bd

PR #35254, opened 2025-10-08
Provider failures if fetch is used to get ResilientGradleBuild

reinsch82:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


testing/architecture-test/src/changes/archunit-store/internal-classes-internal-suffix.txt
```diff
@@ -59,5 +59,6 @@ Class <org.gradle.tooling.internal.provider.InternalCancellationTokenAdapter> do
 Class <org.gradle.tooling.internal.provider.InternalCancellationTokenAdapter> has simple name starting with 'Internal' in (InternalCancellationTokenAdapter.java:0)
 Class <org.gradle.tooling.internal.provider.action.BuildActionSerializer$InternalTaskSpecSerializer> does not have simple name ending with 'Internal' in (BuildActionSerializer.java:0)
 Class <org.gradle.tooling.internal.provider.action.BuildActionSerializer$InternalTaskSpecSerializer> has simple name starting with 'Internal' in (BuildActionSerializer.java:0)
+Class <org.gradle.tooling.internal.provider.runner.DefaultInternalFetchModelResult> does not have simple name ending with 'Internal' in (DefaultInternalFetchModelResult.java:0)
```
asodja: ❓ I think we should respect naming conventions we have. And use a different name. 

Alternatively, maybe we should we add exception for classes that extend interfaces from `org.gradle.tooling.internal.protocol.` WDYT?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r920/ResilientGradleBuildSyncCrossVersionSpec.groovy
```diff
@@ -171,17 +238,22 @@ class ResilientGradleBuildSyncCrossVersionSpec extends ToolingApiSpecification {
...
 
-    static class CustomModelAction implements BuildAction<MyCustomModel>, Serializable {
+    static class FetchModelAction implements BuildAction<FetchModelResult<Model, ResilientGradleBuild>>, Serializable {
+        @Override
+        FetchModelResult<Model, ResilientGradleBuild> execute(BuildController controller) {
+            return controller.fetch(null, ResilientGradleBuild, null, null)
+        }
+    }
 
+    static class CustomModelAction implements BuildAction<MyCustomModel>, Serializable {
```
asodja: ❓ Do we need to test this custom one now that we have fetch API?

testing/architecture-test/src/changes/archunit-store/internal-classes-internal-suffix.txt
reinsch82: That sounds good.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
```diff
@@ -0,0 +1,160 @@
...
+        settingsKotlinFile << """
+            blow up !!!
+        """
+        withStackTraceChecksDisabled()
+
+        when:
+        def model = runFetchModelAction()
+
+        then:
+        model.failures != null // it failed
```
asodja: ❌ I would prefer we return empty list failures instead of null if it didn't fail. Is there any reason we would want to return null?

testing/architecture-test/src/changes/archunit-store/internal-classes-internal-suffix.txt
```diff
@@ -1,56 +1,24 @@
...
-Class <org.gradle.api.problems.internal.InternalProblem> has simple name starting with 'Internal' in (InternalProblem.java:0)
-Class <org.gradle.api.problems.internal.InternalProblemBuilder> does not have simple name ending with 'Internal' in (InternalProblemBuilder.java:0)
-Class <org.gradle.api.problems.internal.InternalProblemBuilder> has simple name starting with 'Internal' in (InternalProblemBuilder.java:0)
-Class <org.gradle.api.problems.internal.InternalProblemReporter> does not have simple name ending with 'Internal' in (InternalProblemReporter.java:0)
-Class <org.gradle.api.problems.internal.InternalProblemReporter> has simple name starting with 'Internal' in (InternalProblemReporter.java:0)
-Class <org.gradle.api.problems.internal.InternalProblemSpec> does not have simple name ending with 'Internal' in (InternalProblemSpec.java:0)
-Class <org.gradle.api.problems.internal.InternalProblemSpec> has simple name starting with 'Internal' in (InternalProblemSpec.java:0)
-Class <org.gradle.api.problems.internal.InternalProblems> does not have simple name ending with 'Internal' in (InternalProblems.java:0)
```
asodja, line 17: ❓ Hmm, Is this change ok? For example, `org.gradle.api.problems.internal.InternalProblems` was removed from this list, but it implements `org.gradle.api.problems.Problems` that is not in `org.gradle.tooling.internal.protocol..`

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/DefaultBuildController.java
```diff
@@ -165,28 +170,23 @@ public void dispatch(Object value) {
...
         } catch (Exception e) {
-            return createDefaultFetchModelResult(target, null, ImmutableList.of(DefaultFailure.fromThrowable(e)));
+            ImmutableList.Builder<InternalFailure> builder = ImmutableList.builder();
+            builder.addAll(getRecordedFailures());
+            builder.add(DefaultFailure.fromThrowable(e));
+            return createDefaultFetchModelResult(target, null, builder.build());
         }
     }
 
+    private List<InternalFailure> getRecordedFailures() {
```
asodja: ❓ What happens if we query multiple resilient model, one after another. E.g. we query GradleBuild twice in the same action. Do we get duplicated failures on the second query?

I wonder if we should clear all failures collected in `ResilientIssuesRecorder` in `finally` block in `fetch` method.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
```diff
@@ -0,0 +1,160 @@
...
+                .run()
+        }
+    }
+
+    def "receive root project with broken settings file [#actionType]"() {
+        given:
+        settingsKotlinFile << """
+            blow up !!!
+        """
+        withStackTraceChecksDisabled()
```
asodja: 💅  Is `withStackTraceChecksDisabled()` needed?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
asodja: I actually see we return empty list if there is no failures, so I think this check is incorrect and it should be:
```
!model.failures.isEmpty()
```

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
```diff
@@ -0,0 +1,160 @@
...
+        settingsKotlinFile << """
+            rootProject.name = "root"
+        """
+        blowUpBuildGradleKts()
+
+        when:
+        def model = runFetchModelAction()
+
+        then:
+        model.failures != null // it failed
```
asodja: ❓ I would expect this scenario not fail, since we don't evaulate build.gradle.kts for GradleBuild. Is this really failing?

I think we should do a check:
```
model.failures.isEmpty()
```

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
```diff
@@ -0,0 +1,160 @@
...
+        model.failures.toString().contains("Script compilation error")
+    }
+
+    def "receive root project with broken root build file [#actionType]"() {
```
asodja: 💅 It seems `[#actionType]` is some leftover, should we remove it?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
```diff
@@ -0,0 +1,160 @@
...
+        included.file("settings.gradle.kts") << """
+            rootProject.name = "included"
+        """
+        blowUpBuildGradleKts(included)
+
+        when:
+        def model = runFetchModelAction()
+
+        then:
+        model.model != null
```
asodja: 💅 I guess this query also doesn't fail, since we don't evaluate configuration. Should we also add a check `models.failures.isEmpty()`?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
asodja: It seems this is used in other test names in this class too

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
```diff
@@ -0,0 +1,160 @@
...
+        includedPlugin.file("settings.gradle.kts") << """
+        rootProject.name = "included-plugin"
+    """
+        blowUpBuildGradleKts(includedPlugin)
+
+        when:
+        def model = runFetchModelAction()
+
+        then:
+        model.failures != null // it failed
```
asodja: 💅 Probably should be:
```
!model.failures.isEmpty()
```

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
```diff
@@ -0,0 +1,160 @@
...
+            settings boom !!!
+        """
+        blowUpBuildGradleKts(included)
+        withStackTraceChecksDisabled()
+
+        when:
+        def model = runFetchModelAction()
+
+        then:
+        model.failures != null // it failed
```
asodja: 💅 Probably should be:
```
!model.failures.isEmpty()
```

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
```diff
@@ -0,0 +1,160 @@
...
+        model.model != null
+        model.failures.toString().contains("Script compilation error")
+        model.model.includedBuilds.getAt(0) != null
```
asodja: 🤔  With a debugger I inspected what we return here for:
```
model.model.includedBuilds.getAt(0)
```

And I see every field of this GradleBuild is null (kinda expected). But I thought that I would be able to get `rootDir` via:
```
model.model.editableBuilds[0].buildIdentifier.rootDir
```
But `buildIdentifier` is null.  Is that expected or can we at least provider `buildIdentifier` with `rootDir` here?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
```diff
@@ -0,0 +1,160 @@
...
+            rootProject.name = "root"
+            includeBuild("included")
+        """
+
+        def included = file("included")
+        included.file("settings.gradle.kts") << """
+            settings boom !!!
+        """
+        blowUpBuildGradleKts(included)
+        withStackTraceChecksDisabled()
```
asodja: 💅 Is `withStackTraceChecksDisabled()` needed?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
asodja: That reminds me that we should probably also check what are the values for returned models for these failed tests. Just to be sure we get back something meaningful and models are as expected. So basically not just check that model exists, but to also check actual values of it.

We can do that also in a separate PR. WDYT?

asodja (approved):
LGTM, let's merge this and do improvements in a separate PR as discussed in Slack

Most importantly, we need to make `DefaultResilientIssuesRecorder` thread safe

PR #35255, opened 2025-10-08
Fix issue with nested classes not handling exclude filters

### LINKED ISSUE #31304: `filter.excludeTest` doesn't work for a JUnit5 nested test class inside another nested test class
### Current Behavior

A test is not excluded when test task is configured with a `filter.excludeTest("A$B$C", null)` where `A$B$C` is nested inside `A$B` which is also nested inside `A`. 

Consider the following example test class:
```Java
public class TestClass {
    @Test
    public void test() {}

    @Nested
    public class FirstLevelNestedTestClass {
        @Test
        public void firstLevelNestedTest() {}

        @Nested
        public class SecondLevelNestedTestClass {
            @Test
            public void secondLevelNestedTest() {}
        }
    }
}
```
When task is configured with ` filter.excludeTest("TestClass${'$'}FirstLevelNestedTestClass${'$'}SecondLevelNestedTestClass", null)` `secondLevelNestedTest()` should be skipped but it is not.
```Kotlin
tasks.register<Test>("testExcludeSecondLevelNested") {
    filter.excludeTest("TestClass${'$'}FirstLevelNestedTestClass${'$'}SecondLevelNestedTestClass", null)
}
//> Task :testExcludeSecondLevelNested
//===========================================
//test()
//firstLevelNestedTest()
//secondLevelNestedTest() - BUG
//===========================================
```

### Examples bellow work as expected:
When exclude configured for all classes `secondLevelNestedTest()` is excluded as expected
```Kotlin
tasks.register<Test>("testExcludeAll") {
    filter.excludeTest("TestClass", null)
    filter.excludeTest("TestClass${'$'}FirstLevelNestedTestClass", null)
    filter.excludeTest("TestClass${'$'}FirstLevelNestedTestClass${'$'}SecondLevelNestedTestClass", null)
}
//> Task :testExcludeAll
//===========================================
//===========================================
```

When exclude is configured for all nested classes `secondLevelNestedTest()` is excluded as expected
```Kotlin
tasks.register<Test>("testExcludeAllNested") {
    filter.excludeTest("TestClass${'$'}FirstLevelNestedTestClass", null)
    filter.excludeTest("TestClass${'$'}FirstLevelNestedTestClass${'$'}SecondLevelNestedTestClass", null)
}
//> Task :testExcludeAllNested
//===========================================
//test()
//===========================================
```

`secondLevelNestedTest()` and `firstLevelNestedTest()` are not excluded as expected
```Kotlin
tasks.register<Test>("testExcludeTopLevel") {
    filter.excludeTest("TestClass", null)
}
//> Task :testExcludeTopLevel
//===========================================
//firstLevelNestedTest()
//secondLevelNestedTest()
//===========================================
```

`secondLevelNestedTest()` is not excluded as expected
```Kotlin
tasks.register<Test>("testExcludeFirstLevelNested") {
    filter.excludeTest("TestClass${'$'}FirstLevelNestedTestClass", null)
}
//> Task :testExcludeFirstLevelNested
//===========================================
//test()
//secondLevelNestedTest()
//===========================================
```

### Expected Behavior

Configuring `filter.excludeTest` excludes a nested test inside another nested test in all cases.

### Context (optional)

[Kotlin](http://github.com/JetBrains/kotlin) project has a lot of nested tests generated from test data. The nesting level may be quite deep (For example: [FirJsCodegenBoxWithInlinedFunInKlibTestGenerated$Ranges$Stepped$Unsigned$Literal$Until$Reversed](https://github.com/JetBrains/kotlin/blob/828c4bdbc4fb84f1b99ecdb0760792ee838b16d9/js/js.tests/tests-gen/org/jetbrains/kotlin/js/test/fir/FirJsCodegenBoxWithInlinedFunInKlibTestGenerated.java#L34480)). In some of our CI builds we use [TeamCity's "Parallel Tests﻿"](https://www.jetbrains.com/help/teamcity/parallel-tests.html) feature and noticed that some of the tests are duplicated across all buckets. It seems the issue is the reason for the duplications.

[test-filter-bug.zip](https://github.com/user-attachments/files/17774278/test-filter-bug.zip)
Run `./gradlew testExcludeSecondLevelNested` to reproduce the issue

### Gradle version

8.11

### Build scan URL (optional)

_No response_

_No response_

ghale:
This fixes an issue where JUnit 5 nested classes were not being excluded properly because we would match on an included parent class and improperly determine that the excluded class should be included.

Note that this also fixes an underlying issue with the pattern matching when including/excluding a test class.  Adding a classname like `com.foo.SampleTest` resulted in a pattern like `com.foo.SampleTest.*` which would never match on a classname (only on a method name).  We now use a pattern without the wild card which means the pattern is `com.foo.SampleTest`.  There is an edge case where `com.foo.SampleTest` _could_ match on a class named `foo` in the `com` package with a method named `SampleTest`, but this seems exotic enough that it's not worth the additional complexity to handle it.

Fixes #31304 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestClassProcessor.java
+++ b/platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestClassProcessor.java
@@ -64,6 +64,7 @@ public class JUnitPlatformTestClassProcessor extends AbstractJUnitTestClassProce
     private final JUnitPlatformSpec spec;
     private final IdGenerator<?> idGenerator;
     private final Clock clock;
+    private final TestSelectionMatcher matcher;
 
     private CollectThenExecuteTestClassConsumer testClassExecutor;
     private BackwardsCompatibleLauncherSession launcherSession;
@@ -74,6 +75,7 @@ public class JUnitPlatformTestClassProcessor extends AbstractJUnitTestClassProce
         this.spec = spec;
         this.idGenerator = idGenerator;
         this.clock = clock;
+        this.matcher = new TestSelectionMatcher(spec.getFilter());
     }
 
     @Override
@@ -114,7 +116,7 @@ public class JUnitPlatformTestClassProcessor extends AbstractJUnitTestClassProce
         @Override
         public void consumeClass(@NonNull TestClassRunInfo testClassInfo) {
             Class<?> klass = loadClass(testClassInfo.getTestClassName());
-            if (isInnerClass(klass) || (supportsVintageTests() && isNestedClassInsideEnclosedRunner(klass))) {
+            if (anyParentIsIncluded(klass, matcher) || (supportsVintageTests() && isNestedClassInsideEnclosedRunner(klass))) {
                 return;
             }
             testClasses.add(klass);
@@ -171,7 +173,24 @@ public class JUnitPlatformTestClassProcessor extends AbstractJUnitTestClassProce
         }
     }
 
-    private boolean isInnerClass(Class<?> klass) {
+    /**
+     * Returns true if any parent class of the given class is included by the given matcher.
+     * This is used to exclude inner classes of included classes, as they will be executed
+     * as part of the parent class's execution.
+     */
+    private static boolean anyParentIsIncluded(Class<?> klass, TestSelectionMatcher matcher) {
+        if (isInnerClass(klass)) {
+            Class<?> parent = klass.getEnclosingClass();
+            if (matcher.mayIncludeClass(parent.getName())) {
+                return true;
+            } else {
+                return anyParentIsIncluded(parent, matcher);
+            }
+        }
+        return false;
+    }
+
+    private static boolean isInnerClass(Class<?> klass) {
         return klass.getEnclosingClass() != null && !Modifier.isStatic(klass.getModifiers());
     }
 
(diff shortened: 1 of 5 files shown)


ghale:
@bot-gradle test this

platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestClassProcessor.java
```diff
@@ -175,7 +177,24 @@ private boolean supportsVintageTests() {
         }
     }
 
-    private boolean isInnerClass(Class<?> klass) {
+    /**
+     * Returns true if any parent class of the given class is included by the given matcher.
+     * This is used to exclude inner classes of included classes, as they will be executed
+     * as part of the parent class's execution.
+     */
+    private static boolean anyParentIsIncluded(Class<?> klass, TestSelectionMatcher matcher) {
```
tresat, line 181: 🤔 It's odd how we do some filtering of classes here, before creating the class selectors in the discovery request, and then we also add some filters to the discovery request as well.  

Can't we create another `PostDiscoveryFilter` similar to `ClassMethodNameFilter` for entire classes and implement this filtering there instead, simplifying the `CollectThenExecuteTestClassConsumer`?

Possibly something to think about after the Non-Class-Based testing is done, I'll add to our list.

ghale: I think the problem here is that the post discovery filters act on the _discovered_ classes and we want the nested classes to be discovered.  So if we filtered it after discovery, we would never execute the nested class.  The problem we resolve by doing the filtering here is that we prevent it from being _selected_ twice.  The way I understand it, since we have two diffrent selectors that capture it (i.e. the parent class selector and the nested class selector) and those classes could get distributed to two different forked processes, we could end up selecting the nested class twice.  So we filter this before distributing the classes to the forked processes which ensures that it only gets selected once.

Another approach (which we've talked about) is pulling the discovery out into a step that occurs prior to distributing the classes (e.g. a JUnit5 specific `TestClassDetector`).  Then we wouldn't need the test class selectors at all and we wouldn't have this duplication problem.

PR #35262, opened 2025-10-09
Improve performance of Gradle source directory extraction

jvandort:
The Kotlin DSL buildscripts model builder uses the FindGradleSources artifact transform to download and extract gradle source code from a gradle source distribution.

Previously, it extracted the entire distribution into a temporary directory, then copied the files of interest into a separate directory, then deleted the temporary directory. This process has been observed to take anywhere from 10 to 35 seconds.

This new implementation optimizes the extraction process, copying only the files of interest to the transform outputs without copying any unwanted files. This reduces the sync time of a gradle build in a fresh GUH. The new time to execute this operation is <2s

Previously we spent most of our time extracting and deleting temporary files:
image

Now, we still spend most of our time copying files, but only the subset we need, and don't spend time deleting files
image

Before:
image

After:
image

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/FindGradleSources.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/FindGradleSources.kt
@@ -24,10 +24,12 @@ import org.gradle.api.file.FileSystemLocation
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity
-import org.gradle.kotlin.dsl.support.unzipTo
+import org.gradle.internal.file.PathTraversalChecker.safePathName
 import org.gradle.work.DisableCachingByDefault
 import java.io.File
-import java.io.IOException
+import java.io.InputStream
+import java.util.zip.ZipEntry
+import java.util.zip.ZipFile
 
 
 /**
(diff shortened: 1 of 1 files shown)


jvandort:
@bot-gradle test this

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/FindGradleSources.kt
```diff
@@ -47,54 +49,111 @@ abstract class FindGradleSources : TransformAction<TransformParameters.None> {
...
-                    }
-                }
+        ZipFile(input.get().asFile).use { zip ->
+            val it = zip.entries()
+            if (!it.hasMoreElements()) {
+                // Zip contains no files.
+                return
+            }
+
+            // We assume the zip contains a single root directory
```
alllex, line 59: ❓  Do we have a test for this?
It can be the test just around the code that creates the sources distribution, with a comment that if that changes the code here should change as well

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/FindGradleSources.kt
```diff
@@ -47,54 +49,111 @@ abstract class FindGradleSources : TransformAction<TransformParameters.None> {
...
+                processProjectEntry(platformDirOffset, outputs, entry, zip)
+            }
+            return
         }
     }
 
-    private fun distroDirFrom(unzippedDistroDir: File): File? =
-        unzippedDistroDir.listFiles()?.singleOrNull()
+    private fun processProjectEntry(offset: Int, outputs: ProjectSourceOutputs, entry: ZipEntry, zip: ZipFile) {
+        // Offset marks the beginning of the project directory name
```
jvandort, line 100:
```suggestion
        // `offset` marks the beginning of the project directory name
```

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/FindGradleSources.kt
eskatos: The TAPI cross-version tests in `kotlin-dsl-integ-tests` assert that the model is populated with actual Gradle sources. Look at tests in `KotlinBuildScriptModelCrossVersionSpec`* that make assertions on the `sourcePath` from the received TAPI model.

alllex: Do you mean this?
https://github.com/gradle/gradle/blob/5d2baecc621322e150d0b5498dddabb75cce6511/platforms/core-configuration/kotlin-dsl-tooling-builders/src/crossVersionTest/groovy/org/gradle/kotlin/dsl/tooling/builders/r54/KotlinBuildScriptModelCrossVersionSpec.groovy#L575

From the initial look, this feels way too involved compared to what I am looking for.

Ideally, we have a test that directly validates the assumptions we are making in the production code here, so that if they are even broken it's clear, and we also have a note in the test that the place here needs to be updated.

PR #35265, opened 2025-10-10
Align ProjectState.parent with Project.parent

alllex:
Replaces the implementation of `ProjectState.parent` with `ProjectState.buildParent`, which is aligned with the public version from `Project.parent` and behaves intuitively.

---

The sneaky `ProjectState.parent` method does not do what most Gradle developers expect. Instead of returning null for the root project of an included build, this method returns the **root project of some other included build**, depending on where the current build is located in the build tree.

The only legitimate-looking use case for this behavior seems to be in the areas of tooling model builders, when IDEs expect projects (across the build-tree) to have unique names. With a recent refactoring (https://github.com/gradle/gradle/pull/35270), this implementation no longer relies on the `ProjectState.parent`.

There were only two more uses of `ProjectState.parent`, but they didn't seem legitimate. One was in the area of `Project` instance comparison/sorting, where we used the "parent" computation to compute project "depth". This was refactored in https://github.com/gradle/gradle/pull/35268 to avoid the usage of `ProjectState.parent` and also align the implementation with the public counter-part -- `Project.depth`.

This leaves only the last usage of `ProjectState.parent`, which is a bug that is addressed in this PR: the usage in the `RelevantProjectsRegistry` for determining dependencies between projects. The changes in the PR look like that file is not touched, but that is only because we don't change the call-site of `project.parent`, but instead change the meaning of that call.

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/RelevantProjectsRegistry.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/RelevantProjectsRegistry.kt
@@ -57,7 +57,7 @@ class RelevantProjectsRegistry(
         if (!projects.add(project)) {
             return
         }
-        val parent = project.buildParent
+        val parent = project.parent
         if (parent != null) {
             collect(parent, projects)
         }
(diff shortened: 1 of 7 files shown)


alllex:

alllex:

alllex:

PR #34990, opened 2025-09-15
Move services from Gradle to Build scope

alllex:
This PR moves us closer toward the goal of getting rid of the `Gradle` service scope. It's a scope that is somewhat different from most others, because it does not have a unique lifetime. In that sense, it has very little difference with the `Build` scope. By removing it, we'll simplify reasoning about services, if only just a little.

This PR:
- Moves a number of services into the `Build` scope
- Removes some services in favor of direct creation, when there is no dependencies and no lifecycle management
  - This is mostly relates to the pattern of exposing event broadcasters as services, where they actually have a "Listener" type, but can be injected to act as broadcasters.

---

Previous related changes:
- https://github.com/gradle/gradle/pull/29534
- https://github.com/gradle/gradle/pull/28706

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #35266, opened 2025-10-10
Remove internal Gradle services scope

alllex:
Moves the remaining services from `Gradle` scope into `Build` scope, removing the need for `Gradle`-scope existance.

Gradle-scope has been around recently mostly for histrorical reasons, while the majority of related services, including the `GradleInternal` service/state, were living in the Build-scope.

---

Previous preparation and further history:
- https://github.com/gradle/gradle/pull/34990

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java
+++ b/subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java
@@ -102,9 +102,9 @@ public abstract class DefaultGradle extends AbstractPluginAware implements Gradl
         this.buildState = buildState;
         this.startParameter = startParameter;
         this.buildScopeServices = buildScopeServices;
-        this.crossProjectConfigurator = this.buildScopeServices.get(CrossProjectConfigurator.class);
-        this.isolatedProjectEvaluationListenerProvider = this.buildScopeServices.get(IsolatedProjectEvaluationListenerProvider.class);
-        this.gradleLifecycleActionExecutor = this.buildScopeServices.get(GradleLifecycleActionExecutor.class);
+        this.crossProjectConfigurator = buildScopeServices.get(CrossProjectConfigurator.class);
+        this.isolatedProjectEvaluationListenerProvider = buildScopeServices.get(IsolatedProjectEvaluationListenerProvider.class);
+        this.gradleLifecycleActionExecutor = buildScopeServices.get(GradleLifecycleActionExecutor.class);
 
         this.buildListenerBroadcast = getListenerManager().createAnonymousBroadcaster(BuildListener.class);
         this.projectEvaluationListenerBroadcast = getListenerManager().createAnonymousBroadcaster(ProjectEvaluationListener.class);
@@ -116,7 +116,7 @@ public abstract class DefaultGradle extends AbstractPluginAware implements Gradl
 
                 if (!rootProjectActions.isEmpty()) {
                     gradleLifecycleActionExecutor.executeBeforeProjectFor(rootProject);
-                    DefaultGradle.this.buildScopeServices.get(CrossProjectConfigurator.class).rootProject(rootProject, rootProjectActions);
+                    buildScopeServices.get(CrossProjectConfigurator.class).rootProject(rootProject, rootProjectActions);
                 }
                 if (isolatedListener != null) {
                     projectEvaluationListenerBroadcast.add(isolatedListener);
@@ -126,7 +126,7 @@ public abstract class DefaultGradle extends AbstractPluginAware implements Gradl
         });
 
         if (buildState.getParent() == null) {
-            this.buildScopeServices.get(GradleEnterprisePluginManager.class).registerMissingPluginWarning(this);
+            buildScopeServices.get(GradleEnterprisePluginManager.class).registerMissingPluginWarning(this);
         }
     }
 


alllex:

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/use/internal/DefaultPluginRequestApplicator.java
```diff
@@ -63,15 +64,15 @@ public class DefaultPluginRequestApplicator implements PluginRequestApplicator {
...
+        CorePluginRegistryProvider corePluginRegistryProvider,
         PluginResolverFactory pluginResolverFactory,
         PluginArtifactRepositoriesProvider pluginRepositoriesProvider,
         PluginResolutionStrategyInternal pluginResolutionStrategy,
         PluginInspector pluginInspector,
         PluginVersionTracker pluginVersionTracker,
         ListenerManager listenerManager
     ) {
-        this.pluginRegistry = pluginRegistry;
+        this.corePluginRegistry = corePluginRegistryProvider.getCorePluginRegistry();
```
bamboo, line 75: 💅 It feels odd that the constructor must extract something from this one argument.

subprojects/core/src/main/java/org/gradle/internal/service/scopes/GradleUserHomeScopeServices.java
```diff
@@ -197,6 +200,12 @@ ClassLoaderScopeRegistry createClassLoaderScopeRegistry(
         return new DefaultClassLoaderScopeRegistry(classLoaderRegistry, classLoaderCache, listenerManager.getBroadcaster());
     }
 
+    @Provides
+    CorePluginRegistryProvider createCorePluginRegistryProvider(ClassLoaderScopeRegistry scopeRegistry, PluginInspector pluginInspector) {
+        DefaultPluginRegistry corePluginRegistry = new DefaultPluginRegistry(pluginInspector, scopeRegistry.getCoreAndPluginsScope());
+        return () -> corePluginRegistry;
```
bamboo, line 206: ❓ Why not expose the registry directly then?

subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java
```diff
@@ -99,13 +98,13 @@ public abstract class DefaultGradle extends AbstractPluginAware implements Gradl
...
-    public DefaultGradle(BuildState buildState, StartParameter startParameter, ServiceRegistryFactory parentRegistry) {
+    public DefaultGradle(BuildState buildState, StartParameter startParameter, ServiceRegistry buildScopeServices) {
         this.buildState = buildState;
         this.startParameter = startParameter;
-        this.services = parentRegistry.createFor(this);
-        this.crossProjectConfigurator = services.get(CrossProjectConfigurator.class);
-        this.isolatedProjectEvaluationListenerProvider = services.get(IsolatedProjectEvaluationListenerProvider.class);
-        this.gradleLifecycleActionExecutor = services.get(GradleLifecycleActionExecutor.class);
+        this.buildScopeServices = buildScopeServices;
+        this.crossProjectConfigurator = this.buildScopeServices.get(CrossProjectConfigurator.class);
```
bamboo: ❓ Why `this.buildScopeServices` instead of just `buildScopeServices`?

subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java
```diff
@@ -611,7 +610,7 @@ public ClassLoaderScope getClassLoaderScope() {
      * {@link Closure} overloads for the {@link IsolatedAction} based methods.
      */
     private DefaultGradleLifecycle instantiateGradleLifecycle() {
-        return services.get(ObjectFactory.class).newInstance(DefaultGradleLifecycle.class, this);
+        return buildScopeServices.get(ObjectFactory.class).newInstance(DefaultGradleLifecycle.class, this);
```
bamboo: 💅 I understand that was already the structure of the code but it seems extracting a `getBuildScopeService(Class<...>)` method would make sense—it would shorten lines and provide for useful call-hierarchy navigation.

subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java
alllex: No reason, artifacts of refactoring. I'll polish this

subprojects/core/src/main/java/org/gradle/internal/service/scopes/GradleUserHomeScopeServices.java
alllex: It is because we moved the other registry from the `Gradle` scope up into the `Build` scope. So if we keep exposing the core PluginRegistry in the Build scope, they would conflict. But since we also moved the core PluginRegistry up to the UserHome scope, the conflict does not take place. However, the previous consumers of the core PluginRegistry in the build scope would have started getting the Gradle-scoped plugin registry as their argument, which is the wrong plugin registry for them. That's why a dedicated type is required.

Introducing this `CorePluginRegistryProvider` type was one of the solutions. The other alternative is to introduce a marker interface `interface CorePluginRegistry extends PluginRegistry {}`. Since it's in the UserHome scope, there will be no conflict, and the services that needed to consume exactly the core plugin registry would be able to do so via the specific type.

Would you prefer the second solution?

Maybe there are other alternatives that I didn't consider.

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/use/internal/DefaultPluginRequestApplicator.java
alllex: With the current implementation that's the only way to consume the right plugin registry here, since all arguments are passed implicitly to the constructor:
https://github.com/gradle/gradle/blob/cc33958488019a9ddd4438724ef412f75a6002bd/platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/internal/PluginUseServices.java#L127

We could change that registration code, but for me it's better that we know this Applicator service needs exactly the core plugin registry and not just any.

---

Also, see my other comment on how we could introduce `CorePluginRegistry` instead. Then here we'd just inject that directly.

subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java
alllex: Just for this one call, it does not seem like it's a lot of savings, so I presume you refer to many `.get(Class)` calls in the constructor. Though, the character savings are only -2 characters per call.

I don't much like calling methods on the same instance before we finish initializing, and using the `getBuildScopeService()` in the constructor would require that.

I think what would be better is to actually make all those service dependencies explicit for the `DefaultGradle` and list them in the class'es constructor.

bamboo: > I presume you refer to many `.get(Class)` calls in the constructor

I had the impression that there were many occurrences of the `buildScopeServices.get(<class>)` pattern outside the constructor. If that's not the case then it's definitely not worth it.

bamboo (approved):
👍

PR #35267, opened 2025-10-10
Add internal javadoc for Project display name

alllex:
The display name of the project is not exactly a public contract, but it's also what `Project.toString()` evaluates to. The latter is used ubiquitously for print debugging and in our own tests.

Without the javadoc, one has to go through multiple hoops to discover the implementation in a non-trivial location:
https://github.com/gradle/gradle/blob/2879727a0a4d76fd2498d9d0780bbbb99b5744e9/subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java#L289-L296

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
+++ b/subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
@@ -875,6 +875,15 @@ public abstract class DefaultProject extends AbstractPluginAware implements Proj
         return projectToEvaluate;
     }
 
+    /**
+     * Returns the display name of this project in a human-readable format.
+     * <p>
+     * Currently:
+     * <ul>
+     *     <li>For the root project: {@code root project 'projectName'}</li>
+     *     <li>For subprojects: {@code project ':identity:path:of:project'}</li>
+     * </ul>
+     */
     @Override
     public String getDisplayName() {
         return owner.getDisplayName().getDisplayName();


subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
```diff
@@ -875,6 +875,13 @@ private Project evaluationDependsOn(ProjectInternal projectToEvaluate) {
         return projectToEvaluate;
     }
 
+    /**
+     * Returns the display name of this project in a human-readable format.
+     * <ul>
+     *     <li>For the root project: {@code root project 'projectName'}</li>
+     *     <li>For subprojects: {@code project ':identity:path:of:project'}</li>
```
abstratt, line 884: ❓ 

Is this the spec for this method, or a description of the current implementation? Made a suggestion to make it is more like the latter

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
```diff
@@ -875,6 +875,13 @@ private Project evaluationDependsOn(ProjectInternal projectToEvaluate) {
         return projectToEvaluate;
     }
 
+    /**
+     * Returns the display name of this project in a human-readable format.
```
abstratt, line 879:
```suggestion
     * Returns the display name of this project in a human-readable format. Currently:
```

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
alllex: It's about the current implementation, but if it ever changes it will affect a lot of places.

I think your suggestion makes sense

PR #35268, opened 2025-10-10
Always compute project depth per build

alllex:
Consolidates the logic to compute "project depth" and ensures a single way of doing so.

We have a public API `Project.depth` and an internal API `ProjectState.depth`. Previously, their computations could have produced different results in composite builds. 

The `ProjectState.depth` might have been different from `Project.depth`, depending on the location of the owner build in the build tree, as we were relying on `ProjectState.parent` method to compute the depth of the current project. The `ProjectState.parent` method is peculiar, because it can return a "parent" project that comes from another build.

The places where this change can potentially have a visible effect is either a direct use of `Project.depthCompare` or comparison/sorting of projects via `Project.compareTo`. However, the change might only be observable, when comparing/sorting projects **from different included builds**, which is rather obscure use case.

### DIFF
--- a/subprojects/core/src/test/groovy/org/gradle/api/internal/project/DefaultProjectTest.groovy
+++ b/subprojects/core/src/test/groovy/org/gradle/api/internal/project/DefaultProjectTest.groovy
@@ -81,6 +81,7 @@ import org.gradle.groovy.scripts.ScriptSource
 import org.gradle.initialization.ClassLoaderScopeRegistryListener
 import org.gradle.internal.Actions
 import org.gradle.internal.Describables
+import org.gradle.internal.build.BuildState
 import org.gradle.internal.deprecation.DeprecationLogger
 import org.gradle.internal.instantiation.InstantiatorFactory
 import org.gradle.internal.logging.LoggingManagerInternal
@@ -125,6 +126,7 @@ class DefaultProjectTest extends Specification {
 
     DefaultProject project, child1, child2, childchild
     ProjectState projectState, child1State, child2State, chilchildState
+    BuildState buildState
 
     ProjectEvaluator projectEvaluator = Mock(ProjectEvaluator)
 
@@ -266,6 +268,9 @@ class DefaultProjectTest extends Specification {
         build.isRootBuild() >> true
         build.getIdentityPath() >> Path.ROOT
 
+        buildState = Stub(BuildState)
+        buildState.getIdentityPath() >> Path.ROOT
+
         serviceRegistryMock.get((Type) ObjectFactory) >> Stub(ObjectFactory)
         serviceRegistryMock.get((Type) DependencyLockingHandler) >> Stub(DependencyLockingHandler)
         serviceRegistryMock.get((Type) DynamicCallContextTracker) >> Stub(DynamicCallContextTracker)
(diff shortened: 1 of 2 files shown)


alllex:

alllex:
@bot-gradle test RfR

alllex:

alllex:
@bot-gradle test APT

subprojects/core/src/test/groovy/org/gradle/api/internal/project/DefaultProjectTest.groovy
```diff
@@ -1014,15 +1023,3 @@ class TaskContainerDynamicObject {
         closure.call()
     }
 }
-
-class TestConvention {
```
alllex, line 1018: ℹ️  unused

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
```diff
@@ -383,7 +381,7 @@ public File getProjectDir() {
 
         @Override
         public int getDepth() {
-            return depth.get();
+            return getProjectPath().segmentCount();
```
alllex, line 384: ℹ️  This is the main change. Now `ProjectState.depth` returns the same values that `Project.depth` has been returning. So `Project.depth` can delegate here to ensure consistency.

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectOrderingUtil.java
```diff
@@ -23,12 +23,17 @@ public static int compare(Project left, Project right) {
     }
 
     public static int compare(ProjectState left, ProjectState right) {
+        int buildCompare = left.getOwner().getIdentityPath().compareTo(right.getOwner().getIdentityPath());
+        if (buildCompare != 0) {
+            return buildCompare;
```
alllex, line 28: ℹ️  Ensures that (in the rare case) when we are sorting projects from different included builds, projects from one build will always come before all projects of another build.

Previously, this compare method was actually not deteministic, since it could have return `0` for two projects with the same `projectPath`, if they happened to be located in the included builds of the same depth.

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectState.java
```diff
@@ -100,8 +100,11 @@ public interface ProjectState extends ModelContainer<ProjectInternal> {
...
     /**
-     * Returns the nesting level of a project in a multi-project hierarchy. For single project builds this is always
-     * 0. In a multi-project hierarchy 0 is returned for the root project.
+     * Returns the nesting level of a project in a multi-project hierarchy.
+     * <p>
+     * Returns 0 for the root project, 1 for its direct children, etc.
+     * <p>
+     * The depth is computed independently for each build in the build tree.
      */
     int getDepth();
```
jvandort, line 109: Do we even need a depth on project state?

I can't really think of a useful use case for exposing this method other than implementing depth compare, and it can easily be implemented using the project path's segment count.

I think we would be better off removing this method entirely

alllex: This would mean duplicating the meaning of project depth in two places again: one for the `Project.depth` public API that we can’t remove, and the other one is the inline knowledge that the segment count represents depth. I would like to avoid that.
So I think it’s better to keep the actual definition of depth in the `ProjectState.depth`, which would return segment count. But both `Project.depth` and `ProjectOrderingUtil` would use the same `ProjectState.depth` in their logic. This way, there is no ambiguity.

jvandort: Do we have any known use cases for Peoject.depth or Project.depthCompare?

alllex: Probably not

jvandort: What are our thoughts on the API evolution for APIs like these? 

IMO these act as noise for the user and we need to maintain code like this which we don't actually see a purpose for.

Would we ever consider deprecating these?

alllex: I think this is a good question to raise with the whole team. These specific APIs were there practically from day-1 (2008). Maybe that's why there is an inherited view of never touching them, especially given they are cheap to maintain.

My view is that, unless there are actual use cases we can identify, we should deprecate these APIs in sources. So that any new compilations observe them, while binary plugins that ever touched them just work.

PR #35270, opened 2025-10-10
Avoid ProjectState.parent when computing unique project names for IDEs

alllex:
This is pure refactoring PR, so no behavior changes should be expected.

Refactors the logic to assign unique names to projects across **the build tree**, primarily to be provided to the IDEs, such that project names don't have clashes.

The motivation for the change is to avoid using the sneaky `ProjectState.parent` method that does not do what most Gradle developers expect. Instead of returning null for the root project of an included build, this method returns the root project of some other included build, depending on where the current build is located in the build tree.

While this is not the expected behavior for the majority of cases when a parent project is needed, there is a legitimate use case in the tooling model builders.

The model builders are among very few callers of the `ProjectState.parent` method. So to avoid confusing and future bugs, this PR refactors the logic of model builders to avoid using that method, and instead be explicit about cross-build "jumps", when considering a "parent" of a project.

### DIFF
--- a/platforms/ide/ide/src/main/java/org/gradle/plugins/ide/internal/configurer/AbstractUniqueProjectNameProvider.java
+++ b/platforms/ide/ide/src/main/java/org/gradle/plugins/ide/internal/configurer/AbstractUniqueProjectNameProvider.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


alllex:
@bot-gradle RfR

alllex:
@bot-gradle test RfR

alllex:

alllex:
> 💅 Some of the logic in the two ProjectPathDeduplicationAdapter implements HierarchicalElementAdapter implementations is now duplicated. Would it be worth introducing a common parent?

@eskatos, this was in the back of my mind as well. Now that you mention it, it's probably worth deduplicating. I'll update the PR.

alllex:
@jbartok It seems we already have a number of tests for deduplicated names:
https://github.com/search?q=repo%3Agradle%2Fgradle+path%3AAbstractIdeDeduplicationIntegrationTest.groovy+dedup&type=code

Issue #35221, opened 2025-10-04
gradle -version should not show Gradle JDK warning when setup is correct


miracletim:
### Issue type

Polishing

### Problem description

# Description / Issue
Running `gradle -version` with `JAVA_HOME` set correctly prints a warning about `(no JDK specified, using current Java home)` even though Gradle is fully operational.

While harmless, this warning gives the impression that the setup is incomplete or broken, which can confuse users, especially newcomers.

# Steps to reproduce

1. Set a valid `JAVA_HOME`

2. Install Gradle.

3. Run: `gradle -version`

# Observed behavior

- Gradle prints version info correctly.

- Additional warning about Gradle JDK appears unnecessarily.

# Expected behavior

- Only display version information.

- Only show warnings if there is a real problem.

# Suggested fix

- Suppress the Gradle JDK warning if Gradle can detect a valid Java installation.

- Optionally, update documentation to guide users on specifying a custom Gradle JDK if needed.

# Rationale

- Improves first-time user experience by avoiding misleading warnings.

- Keeps `gradle -version` output clean and trustworthy.

# Environment

*Gradle version:* `Gradle 9.1.0`

*Java version:* `openjdk version "17.0.16" 2025-07-15`

*OS:* `Windows 10 10.0 amd64`

### Context (optional)

<img width="1104" height="297" alt="Image" src="https://github.com/user-attachments/assets/742254c3-2350-4ab1-9434-fd5bdeb2a9aa" />

hegyibalint:
We think the message you are referring to is not a warning but a factual statement about how the JDK was selected.
However, we think we could do the following:
 - [ ] the "Daemon JVM" message could use a bit of tuning to communicate that we are talking about daemon toolchains.
 - [ ] the section title [Daemon JVM criteria](https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:daemon_jvm_criteria) should be renamed, so users could easily find what daemon toolchains are after reading the message

PR #35279, opened 2025-10-12
Update Daemon JVM selection message when executing `gradle -version`

lkasso:
Fixes https://github.com/gradle/gradle/issues/35221

### Context
- Update defaults message from `(no JDK specified, using current Java home)` to `(no Daemon JVM specified, using current Java home)`
- Updates the Daemon docs

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs/src/docs/release/notes-template.md
+++ b/platforms/documentation/docs/src/docs/release/notes-template.md
@@ -45,7 +45,6 @@ For Java, Groovy, Kotlin, and Android compatibility, see the [full compatibility
 
 ================== TEMPLATE ==============================
 
-<a name="FILL-IN-KEY-AREA"></a>
 ### FILL-IN-KEY-AREA improvements
 
 <<<FILL IN CONTEXT FOR KEY AREA>>>
--- a/platforms/documentation/docs/src/docs/release/notes.md
+++ b/platforms/documentation/docs/src/docs/release/notes.md
@@ -45,7 +45,6 @@ For Java, Groovy, Kotlin, and Android compatibility, see the [full compatibility
 
 ================== TEMPLATE ==============================
 
-<a name="FILL-IN-KEY-AREA"></a>
 ### FILL-IN-KEY-AREA improvements
 
 <<<FILL IN CONTEXT FOR KEY AREA>>>
@@ -73,7 +72,6 @@ For Wistia, contact Gradle's Video Team.
 ADD RELEASE FEATURES BELOW
 vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv -->
 
-<a name="build-authoring"></a>
 ### Build authoring improvements
 
 #### New `AttributeContainer.named()` method
@@ -94,7 +92,6 @@ configurations.resolvable("foo") {
 }
 ```
 
-<a name="stream-testkit-output"></a>
 ### Stream TestKit output
 
 Gradle TestKit's `BuildResult` now offers a new method for accessing the build console output efficiently, especially for builds that produce a large volume of logs.
--- a/platforms/documentation/docs/src/test/groovy/org/gradle/docs/releasenotes/StaticReleaseNotesTest.groovy
+++ b/platforms/documentation/docs/src/test/groovy/org/gradle/docs/releasenotes/StaticReleaseNotesTest.groovy
@@ -52,7 +52,7 @@ class StaticReleaseNotesTest extends Specification {
         def brokenAnchorLinks = []
         def links = renderedDocument.select("a")
         def ids = renderedDocument.allElements.findAll { it.id() }*.id()
-        def anchors = links.findAll { it.attr("name") }*.attr("name")
+        def anchors = links.findAll { it.attr("id") }*.attr("id")
 
         links.each {
             def href = it.attr("href")
@@ -66,6 +66,15 @@ class StaticReleaseNotesTest extends Specification {
 
         then:
         brokenAnchorLinks.empty
+    }
+
+    def "does not use <a> name attribute"() {
+        when:
+        def links = renderedDocument.select("a")
+        def nameUsages = links.findAll { it.attr("name") }*.attr("name")
+
+        then:
+        assert nameUsages.empty : "all <a> elements used as targets should use id or be removed if possible"
     }
 
     def "no absolute links to docs.gradle.org"() {


platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
```diff
@@ -16,6 +16,32 @@
 = Gradle Daemon
 :keywords: daemon, status, jps, no-daemon, stop, org.gradle.daemon, updateDaemonJvm, gradle-daemon-jvm.properties, jvm-version, jvm-vendor
 
+The Gradle Daemon is a long-lived, persistent process that runs in the background and hosts Gradle's execution engine.
+Its primary benefit is dramatically reducing build times by eliminating the significant overhead associated with JVM startup and initialization on every command execution.
```
alllex: 💅  I'd ping the JVM team about the phrasing here. I am, personally, not 100% sure it's the skipped JVM startup that improves things dramatically, but rather it is a number of runtime-only caches that we have. So I guess it's much more about "initialization" of some kind. Though I might be wrong.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
```diff
@@ -16,6 +16,32 @@
...
+To run any Gradle command, your system starts two separate Java Virtual Machine (JVM) processes, which can use different Java versions.
+
+Understanding their distinct roles and how their respective JVM versions are determined is key to effective build configuration and debugging:
+
+[cols="1,1,2", options="header"]
+|===
+| Process | Role | JVM Version Source
+
+| *Gradle Client JVM*
+| Short-lived launcher that starts when you run `./gradlew`. It forwards the request to the Daemon and handles console I/O.
```
alllex: > Short-lived

💭  I'd note that the client lives at least as long as the build invocation. Right now it seems like we use it to start and then discard it immediately, but it's not the case.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
```diff
@@ -16,6 +16,32 @@
...
+| Whatever JVM launches the script (e.g., `JAVA_HOME`, `java` on `PATH`, or settings in your IDE).
+
+| *Gradle Daemon JVM*
+| Long-lived process that executes the build and caches state between runs.
+| By default, the *same JVM as the Gradle Client*.
```
alllex: ❔  Is this actually true?

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
```diff
@@ -16,6 +16,32 @@
...
+Can be overridden by `--java-home` (CLI), `org.gradle.java.home` (properties), the <<tooling_api.adoc#tooling_api,Tooling API>>, or the <<#sec:configuring_daemon_jvm,Daemon JVM toolchains>>.
```
alllex: 💭  Probably a much more important consideration is the difference in the jvmargs. If jvmargs are different, then a second JVM will always be spawned. And jvmargs is something people change most often, normally to increase the daemon memory.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
lkasso: This is not entirely accurate—at least, it's not well-worded. I updated it to a list, so it makes more sense.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
lkasso: Thanks for this feedback, I updated the text accordingly.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
lkasso: This is what's already on the Gradle Daemon [doc page today.](https://docs.gradle.org/current/userguide/gradle_daemon.html) @tresat can you comment?

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
lkasso: Added an asterix about this.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
tresat: I agree this language is overselling one-time startup time benefits and underselling the ongoing caching.   In the linked doc page, "The Gradle Daemon reduces build times by:" lists 4 different items, only one is reducing startup costs.

It's not a big deal, but probably worth changing to something like:

> It dramatically reduces build times using caching, runtime optimizations, and eliminating JVM startup overhead.

lkasso: Fixed.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
```diff
@@ -16,6 +16,39 @@
 = Gradle Daemon
 :keywords: daemon, status, jps, no-daemon, stop, org.gradle.daemon, updateDaemonJvm, gradle-daemon-jvm.properties, jvm-version, jvm-vendor
 
+The Gradle Daemon is a long-lived, persistent process that runs in the background and hosts Gradle's execution engine.
+It dramatically reduces build times using caching, runtime optimizations, and eliminating JVM startup overhead.
+
+== The Gradle Client vs. the Gradle Daemon
+
+To run any Gradle command, your system starts two separate Java Virtual Machine (JVM) processes, which can use different Java versions.
```
tresat, line 24: 💅 Unless using `--no-daemon`.

big-guy: I wouldn't include this because it's (mostly) not true anymore.  `--no-daemon` means "do not reuse the build process", which might be the current process _or_ a single use daemon process.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
```diff
@@ -16,6 +16,39 @@
...
+
+Understanding their distinct roles and how their respective JVM versions are determined is key to effective build configuration and debugging:
+
+[cols="1,2,3", options="header"]
+|===
+| Process | Role | JVM Version Source
+
+| *Gradle Client JVM*
+| Process that starts when you run `./gradlew` and stays alive for the *entire build invocation*. +
```
big-guy: It also runs when you run `gradle` (e.g., you've installed Gradle on your $PATH)

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
```diff
@@ -16,6 +16,39 @@
...
+It connects to (or starts) a Daemon, sends the build request, and streams output back to the console.
+| Uses the JVM that launches the wrapper script (e.g. `JAVA_HOME`, `java` on `PATH`, or your IDE).
+
+| *Gradle Daemon JVM*
+| Long-lived process that executes the build and caches state between runs.
+| Uses the JVM set by (in order): +
+1) `--java-home` (CLI) +
```
big-guy: This isn't an option?

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
```diff
@@ -16,6 +16,39 @@
...
+2) `org.gradle.java.home` (properties) +
+3) <<tooling_api.adoc#tooling_api,Tooling API>> requests (e.g., IDE) +
+4) <<#sec:configuring_daemon_jvm,Daemon JVM toolchains>> +
+5) Falls back to the JVM that launched the Gradle Client or reuses an existing Daemon if conditions are right*
```
big-guy: I think this is conflating two different things.

1. Given no other opinions, Gradle will use the JVM from the client.  
2. The client will look for compatible daemons and, by default, start a new one if none are available.

We only attempt to use the client process to run the build if the daemon has been disabled (--no-daemon or -Dorg.gradle.daemon=false) and the client process is compatible. It almost never is compatible, so the client will start a single use daemon instead.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
```diff
@@ -16,6 +16,39 @@
...
+| Process | Role | JVM Version Source
+
+| *Gradle Client JVM*
+| Process that starts when you run `./gradlew` and stays alive for the *entire build invocation*. +
+It connects to (or starts) a Daemon, sends the build request, and streams output back to the console.
+| Uses the JVM that launches the wrapper script (e.g. `JAVA_HOME`, `java` on `PATH`, or your IDE).
+
+| *Gradle Daemon JVM*
+| Long-lived process that executes the build and caches state between runs.
+| Uses the JVM set by (in order): +
```
big-guy: I think it's confusing to refer to this as an "order" because something like `org.gradle.java.home` can be set in different locations (Gradle user home, project directory, command-line, etc) and whether or not it overrides another way to set the Java home depends on which location is more preferred.

That means something from TAPI can almost always win over something set in gradle.properties because it's more similar to the CLI which has a higher precedence than any gradle.properties files.

big-guy (changes requested):
I don't mind the change in the message, but the docs aren't quite there.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
lkasso: I rewrote that section so many times that I ended up adding this in the wrong place, and it's supposed to be an env var! Thanks for catching this. It's now fixed.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
lkasso: I had `gradle` in the first version and then I changed it after Alex's round of feedback. I will go ahead and put both `gradle` and `./gradlew`. TY. Fixed.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
lkasso: I also re-wrote this more times than I can count. Let me know what you think of this newer instance. I aimed to make this as simple as possible (but I oversimplified imho).

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
lkasso: Resolving this with @big-guy's latest comment.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/gradle_daemon.adoc
lkasso: I turned it into a general list. Let me know what you think.

PR #35290, opened 2025-10-13
Fixes false positive in incompatible constraints conflict describer

tresat:
The module rejected incompatible constraints failure describer incorrectly triggered when there were multiple unique constraint paths that all targetted the same version.

This change ensures that the describer only triggers when there are truly incompatible constraints that ask for different versions.


tresat:
@bot-gradle test and merge

tresat:
@bot-gradle test this

tresat:
@bot-gradle test and merge

PR #35292, opened 2025-10-13
Enable NullAway in base-services

mlopatkin:
`:base-services` has more polynull methods, so I've added `@Contract` instead of modifying all the callsites. I also suspect that many users are depending on GUtil and TextUtil despite them being internal.

### DIFF
--- a/platforms/core-runtime/base-services/src/main/java/org/gradle/util/Path.java
+++ b/platforms/core-runtime/base-services/src/main/java/org/gradle/util/Path.java
@@ -61,7 +61,7 @@ public class Path implements Comparable<Path> {
         validatePathInternal(path);
     }
 
-    @Contract("null -> fail") // TODO(mlopatkin) Maybe exposing Contract in the public API is fine?
+    @Contract("null -> fail")
     private static void validatePathInternal(@Nullable String path) {
         if (Strings.isNullOrEmpty(path)) {
             throw new InvalidUserDataException("A path must be specified!");
--- a/testing/architecture-test/src/test/java/org/gradle/architecture/test/PublicApiCorrectnessTest.java
+++ b/testing/architecture-test/src/test/java/org/gradle/architecture/test/PublicApiCorrectnessTest.java
@@ -17,10 +17,15 @@
 package org.gradle.architecture.test;
 
 import com.tngtech.archunit.base.DescribedPredicate;
+import com.tngtech.archunit.core.domain.JavaAnnotation;
 import com.tngtech.archunit.core.domain.JavaClass;
+import com.tngtech.archunit.core.domain.JavaCodeUnit;
 import com.tngtech.archunit.junit.AnalyzeClasses;
 import com.tngtech.archunit.junit.ArchTest;
+import com.tngtech.archunit.lang.ArchCondition;
 import com.tngtech.archunit.lang.ArchRule;
+import com.tngtech.archunit.lang.ConditionEvents;
+import com.tngtech.archunit.lang.SimpleConditionEvent;
 import kotlin.Pair;
 import kotlin.jvm.functions.Function1;
 import kotlin.reflect.KClass;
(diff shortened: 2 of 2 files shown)


mlopatkin:
@bot-gradle test QFL without PTS

mlopatkin:
@bot-gradle test QFL without PTS

mlopatkin:
@bot-gradle test QFL without PTS

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/exceptions/FormattingDiagnosticsVisitor.java
```diff
@@ -59,6 +61,7 @@ public DiagnosticsVisitor candidate(String displayName) {
 
     @Override
     public DiagnosticsVisitor example(String example) {
+        Preconditions.checkState(current != null, "add a candidate() first");
```
mlopatkin, line 64: 💭 This was failing with NPE before

platforms/core-runtime/base-services/src/main/java/org/gradle/api/internal/cache/StringInterner.java
```diff
@@ -31,7 +30,8 @@ public StringInterner() {
     }
 
     @Override
-    public String intern(@Nullable String sample) {
+    @SuppressWarnings("NullAway") // For legacy reasons, while unmarked code exists.
+    public String intern(String sample) {
         if (sample == null) {
             return null;
         }
```
mlopatkin, line 37: 💭 Returning `null` actually violates the contract of the `Interner` interface.

I think I've checked all the callsites and they're safe, but let's keep the extra protection for some time.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/graph/CachingDirectedGraphWalker.java
```diff
@@ -146,7 +147,7 @@ private Set<T> doSearch() {
                 }
 
                 for (N connectedNode : details.successors) {
-                    NodeDetails<N, T> connectedNodeDetails = seenNodes.get(connectedNode);
+                    NodeDetails<N, T> connectedNodeDetails = Objects.requireNonNull(seenNodes.get(connectedNode));
```
mlopatkin, line 150: 💭 this is not adding a new failure (NPE happens right before), but just makes our assumption explicit.

Most of new `requireNonNull` here are like this.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/DefaultResourceLockCoordinationService.java
```diff
@@ -79,13 +80,14 @@ public void withStateLock(final Runnable action) {
...
-        final MutableReference<T> result = MutableReference.empty();
+    public <T extends @Nullable Object> T withStateLock(final Supplier<T> action) {
+        final MutableReference<@Nullable T> result = MutableReference.empty();
         withStateLock(resourceLockState -> {
             result.set(action.get());
             return ResourceLockState.Disposition.FINISHED;
         });
-        return result.get();
+        // result now holds a valid value of T supplied by the action.
+        return Cast.unsafeStripNullable(result.get());
```
mlopatkin, line 90: 💭 We have `action`s that return `null`.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultResourceLockStatistics.java
```diff
@@ -47,7 +48,7 @@ public DefaultResourceLockStatistics(BuildOperationRunner buildOperationRunner)
 
     @Override
     public void measureLockAcquisition(Iterable<? extends Describable> locks, Runnable runnable) {
-        measure("Blocked on", locks, () -> {
+        measure("Blocked on", locks, (Factory<@Nullable Void>) () -> {
```
mlopatkin, line 51: 💭 An explicit type is needed for NullAway 🫤. At least, IDEA doesn't complain.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/notations/DependencyStringNotationConverter.java
```diff
@@ -57,8 +58,7 @@ private T createDependencyFromString(String notation) {
 
         ParsedModuleStringNotation parsedNotation = splitModuleFromExtension(notation);
         StrictVersionParser.RichVersion version = strictVersionParser.parse(parsedNotation.getVersion());
-        T moduleDependency = instantiator.newInstance(wantedType,
```
mlopatkin, line 60: 💭 This was the only unsafe use of `intern` I've found.

mlopatkin:
@bot-gradle test this

PR #35295, opened 2025-10-13
Bump to Swift 6 for testing where possible

octylFractal:
Also remove `SWIFTC_4_OR_OLDER` (just use `SWIFTC_4`)

I tried to remove `SWIFTC_5_OR_OLDER` but had issues with the C++ interoperability breaking. I don't need to fix this for our puposes so I left those alone, but at some point we will need to deal with that. It seems like we're not even using the proper interop mode documented at https://www.swift.org/documentation/cxx-interop/#enabling-c-interoperability ...

### Context
We need to run the XCTest integration tests locally, but getting Swift 5 to work on modern macOS is nearly impossible with our current Swift integration.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


octylFractal:
This seems to be failing in many places and is not critically important, so I'm going to close it for now.

Issue #35040, opened 2025-09-18
Add Gradleception IP test


alllex:
Something similar to `GradleBuildSanityCheckConfigurationCacheSmokeTest`

The main goal of the test is to catch any IP violations that might be added to our build via plugin version updates or build logic changes.

Hint: make the test deterministic by disabling parallelism and caching, for now:
* `org.gradle.internal.isolated-projects.parallel=false`
* `org.gradle.internal.isolated-projects.caching=false`

The test should capture existing violations, if any. This is useful, because accepting violations can unblock other work, e.g. a necessary plugin upgrade, while IP is still experimental.

PR #35308, opened 2025-10-14
Introduce Gradleception test for IP

abstratt:
...using `sanityCheck`.

Issue: #35040

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractGradleBuildIsolatedProjectsSmokeTest.groovy
+++ b/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractGradleBuildIsolatedProjectsSmokeTest.groovy
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.smoketests
+
+import org.gradle.initialization.StartParameterBuildOptions
+
+/**
+ * Smoke test building gradle/gradle with isolated projects enabled.
+ */
+abstract class AbstractGradleBuildIsolatedProjectsSmokeTest extends AbstractGradleBuildConfigurationCacheSmokeTest {
+
+    protected void setMaxIsolatedProjectProblems(int maxProblems) {
+        maxConfigurationCacheProblems = maxProblems
+    }
+
+    void isolatedProjectsRun(List<String> tasks, int daemonId = 0) {
+        def ipOptions = [
+            "-D${StartParameterBuildOptions.IsolatedProjectsOption.PROPERTY_NAME}=true",
+            // caching is not stable enough
+            "-Dorg.gradle.internal.isolated-projects.caching=false",
+            // to improve determinism
+            "-Dorg.gradle.internal.isolated-projects.parallel=false",
+            // we do not expect execution-time issues, so save some build time
+            "--dry-run",
+        ]
+        configurationCacheRun(tasks + ipOptions, daemonId)
+    }
+}
--- a/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
+++ b/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.smoketests
+
+class GradleBuildIsolatedProjectsSmokeTest extends AbstractGradleBuildIsolatedProjectsSmokeTest {
+
+    def "can run Gradle build tasks with isolated projects enabled"() {
+        given:
+        def tasks = [
+            "build",
+            "sanityCheck",
+            "test",
+            "embeddedIntegTest",
+            // AsciidoctorTask only became CC compatible in version 5 (https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/564),
+            // skip these tasks to avoid non-IP problems
+            "-x", ":docs:samplesMultiPage",
+            "-x", ":docs:userguideMultiPage",
+            "-x", ":docs:userguideSinglePageHtml"
+        ]
+
+        when:
+        maxIsolatedProjectProblems = 13
+        isolatedProjectsRun(tasks)
+
+        then:
+        result.assertConfigurationCacheStateStoreDiscarded()
+        result.output.contains "13 problems were found storing the configuration cache, 5 of which seem unique."
+        result.output.contains "- Plugin 'org.jetbrains.kotlin.jvm': Project ':declarative-dsl-core' cannot dynamically look up a property in the parent project ':'"
+        result.output.contains "- Plugin 'org.jetbrains.kotlin.jvm': Project ':declarative-dsl-evaluator' cannot dynamically look up a property in the parent project ':'"
+        result.output.contains "- Plugin 'org.jetbrains.kotlin.jvm': Project ':declarative-dsl-tooling-models' cannot dynamically look up a property in the parent project ':'"
+        result.output.contains "- Plugin 'org.jetbrains.kotlin.jvm': Project ':kotlin-dsl-plugins' cannot dynamically look up a property in the parent project ':'"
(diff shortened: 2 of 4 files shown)


abstratt:
@bot-gradle test AST

abstratt:
@bot-gradle test AST

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildSanityCheckIsolatedProjectsSmokeTest.groovy
```diff
@@ -0,0 +1,48 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.smoketests
+
+class GradleBuildSanityCheckIsolatedProjectsSmokeTest extends AbstractGradleBuildIsolatedProjectsSmokeTest {
```
abstratt: Will rename the class if we agree on specifying more tasks than just `sanityCheck`- since this is a dry-run, the impact of specifying additional tasks on build time is not as bad, and may help uncover more incompatibilities.

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractGradleBuildIsolatedProjectsSmokeTest.groovy
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Smoke test building gradle/gradle with isolated projects enabled.
+ */
+abstract class AbstractGradleBuildIsolatedProjectsSmokeTest extends AbstractGradleBuildConfigurationCacheSmokeTest {
+
+    protected void setMaxIsolatedProjectProblems(int maxProblems) {
+        maxConfigurationCacheProblems = maxProblems
+    }
+
+    void isolatedProjectsRun(List<String> tasks, int daemonId = 0) {
```
6hundreds, line 30: ❓ Why to pass `daemonId` ? Don't see usages of it

6hundreds: I see it's being used in `configurationCacheRun`, then we should pass it there, if we really need it

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildSanityCheckIsolatedProjectsSmokeTest.groovy
```diff
@@ -0,0 +1,48 @@
...
+            // AsciidoctorTask only became CC compatible in version 5 (https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/564),
+            // skip these tasks to avoid non-IP problems
+            "-x", ":docs:samplesMultiPage",
+            "-x", ":docs:userguideMultiPage",
+            "-x", ":docs:userguideSinglePageHtml"
+        ]
+
+        when:
+        maxIsolatedProjectProblems = 13
+        isolatedProjectsRun(tasks, 0)
```
6hundreds: 💅 `0` is default I believe, we can omit it then

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractGradleBuildIsolatedProjectsSmokeTest.groovy
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.gradle.smoketests
+
+import org.gradle.initialization.StartParameterBuildOptions
+
+/**
+ * Smoke test building gradle/gradle with isolated projects enabled.
+ */
+abstract class AbstractGradleBuildIsolatedProjectsSmokeTest extends AbstractGradleBuildConfigurationCacheSmokeTest {
```
6hundreds, line 24: ❓ `AbstractGradleBuildConfigurationCacheSmokeTest` has some preconditions on it. Will it be inherited for this class? I see `AbstractGradleceptionSmokeTest` and `AbstractGradleBuildConfigurationCacheSmokeTest` have duplicated set of preconditions despite the single hierarchy, what is make me think that preconditions may be not a subject for inheritance.

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractGradleBuildConfigurationCacheSmokeTest.groovy
```diff
@@ -50,7 +50,7 @@ abstract class AbstractGradleBuildConfigurationCacheSmokeTest extends AbstractGr
 
     protected int maxConfigurationCacheProblems = 0
 
-    void configurationCacheRun(List<String> tasks, int daemonId = 0) {
+    void configurationCacheRun(List<String> tasksAndOptions, int daemonId = 0) {
```
6hundreds: 💅 Maybe `args`? Not insist tho

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
```diff
@@ -0,0 +1,48 @@
...
+            "-x", ":docs:userguideSinglePageHtml"
+        ]
+
+        when:
+        maxIsolatedProjectProblems = 13
+        isolatedProjectsRun(tasks, 0)
+
+        then:
+        result.assertConfigurationCacheStateStoreDiscarded()
+        result.output.contains "13 problems were found storing the configuration cache, 5 of which seem unique."
```
alllex, line 41: ❓  Can we use some of our test fixtures for parsing build output / CC reports?

The current validation is not too bad, because there is a connection between 5 unique problems and 5 further checks, but this seems too adhoc. If it's too hard to use the fixtures, maybe it's okay as is.

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractGradleBuildIsolatedProjectsSmokeTest.groovy
abstratt: Forgot to pass it, good catch, thanks

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractGradleBuildIsolatedProjectsSmokeTest.groovy
abstratt: Good question, but they are indeed inheritable. 

Not sure why the duplication, maybe it was be to be independent from the super class conditions changing, but whatever works for the CC tests should be good for IP tests.

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
abstratt: AFAIU, in smoke tests, we are using `GradleRunner`, which is a TestKit abstraction. 

The fixtures we use for checking for build problems (including configuration cache report fixtures) are tied to `GradleExecuter` and `ExecutionResult`, which are part of our internal integration test framework.

What are you missing with this approach?

abstratt:
@bot-gradle test AST

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
alllex: As I said, right now this seems fine, but I have concerns about the error-proneness of this validation.

- It's easy to change the first line checking the number of projects, but not assert all individual violations, especially, if the number of violations grows for some reason.
- If there will be more than 10 unique violations, they won't be printed to the console.

These problems might get compounded in a way, if we start introducing more similar tests and copy-pasting this validation there.

---

> The fixtures we use for checking for build problems (including configuration cache report fixtures) are tied to GradleExecuter and ExecutionResult, which are part of our internal integration test framework.

Can you take a look at the infrastructure around IP/CC violations validation that @6hundreds introduced and that we already have for the IDE Sync tests? `IsolatedProjectsIdeSyncFixture` and its usages. I don't think those rely on the internal integration test framework.

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractGradleBuildIsolatedProjectsSmokeTest.groovy
alllex, line 24: ❓  @abstratt, can you make sure that if anyone changes this abstract class or any of the concrete test classes, that the IP team gets added as a reviewer?

abstratt:
@bot-gradle test PRF

abstratt:
@bot-gradle test this

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
abstratt: Thanks - you are right, we are using the bits of CCProblemFixture that are not reliant on `ExecutionResult`/`ExecutionFailure` 

I will pull this functionality into `internal-integ-testing`, so I can reuse it in the smoke tests. Will do it a follow-up PR.

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractGradleBuildIsolatedProjectsSmokeTest.groovy
abstratt: See: https://github.com/gradle/gradle/pull/35371/files#diff-3d36a1bf06148bc6ba1ce2ed3d19de32ea708d955fed212c0d27c536f0bd4da7R35

PR #35311, opened 2025-10-14
Remove dynamic task container accesses from DM tests

jvandort:
We will eventually want to deprecate the behavior of dynamically looking up tasks from the task container when doing dynamic lookups against a project. This gets ahead of the curve for dependency management tests

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

jvandort:
Before we go enabling IP tests on DM were blocked on https://github.com/gradle/gradle/pull/35425 and the issue with dynamic groovy, containers, and implicit object creation.

Without resolving those two there would just be way too many test to manually ignore for IP

Issue #31243, opened 2024-11-11
Internal options are not customizable for sync


alllex:
Gradle has many [internal options](https://github.com/gradle/gradle/blob/14cbec5f92744710f42684f794a0150e6a48db63/platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/InternalOption.java#L24) used for debugging, testing and troubleshooting. The values for those options are determined by the system properties map in the `StartParameter`.

These system properties need to be available before the build tree is created, so they currently don't include the usual locations for system property customization like `gradle.properties` file.
https://github.com/gradle/gradle/blob/d5527c11b1c25b0d8c607f5eb649e40e729c036c/subprojects/core/src/main/java/org/gradle/internal/buildtree/BuildTreeScopeServices.java#L140

Some are read even before the build session, like the build operations tracing property:
https://github.com/gradle/gradle/blob/4a4d8403b784ae441d533257ab4ba0759a7f568f/platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationTrace.java#L161-L162

When invoking Gradle on the command line, even these "early" system properties can be customized via the `-Dmy.prop` arguments.
However, when it comes to syncing in IDEA, there is no such option. In fact, currently, there is no way to customize those options for sync without creating an IDE plugin.

There should be a way for Gradle developers and troubleshooting users to reliably customize the corresponding system properties that works equally well on CLI and in the IDE.

alllex:
One approach to consider is to fill internal properties from the `gradle.properties` files as we do for the regular build scope:
https://github.com/gradle/gradle/blob/29e0f2adbd333e29a1c9d92d63c0b75eb2bd6a39/subprojects/core/src/main/java/org/gradle/initialization/DefaultGradlePropertiesLoader.java#L47-L49

Normally, the properties from the files have lower priority than the ones provided in the `StartParameter`:
https://github.com/gradle/gradle/blob/b7607da8bfef113f3c48982a09d655632fd5bf37/subprojects/core/src/main/java/org/gradle/initialization/properties/DefaultSystemPropertiesInstaller.java#L52-L53

mlopatkin:
We should start with a discovery spec and outline the current situation.

alllex:
When running a build from CLI, the build operations trace flag can be specified with `-D` on the command line. However, it also works if it's specified as a `systemProp` in the `gradle.properties` file:
```properties
[...]

PR #35357, opened 2025-10-17
Enforce prefix for internal options

alllex:
Preparation for https://github.com/gradle/gradle/issues/31243

Enforces that all internal options have `org.gradle.internal.` prefix.

Most internal options already follow this rule, but there was no validation to ensure this for all internal options.

We keep a few exceptions for now to be addressed later:
- Configuration Cache using `org.gradle.configuration-cache.internal.*` naming
- Gradle API generation using `unsafe` instead of `internal` for an internal flag

### DIFF
--- a/platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/InternalOption.java
+++ b/platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/InternalOption.java
@@ -45,6 +45,11 @@ public abstract class InternalOption<T extends @Nullable Object> implements Opti
 
     public abstract T convert(String value);
 
+    @Override
+    public String toString() {
+        return "InternalOption('" + getPropertyName() + "', default=" + getDefaultValue() + ")";
+    }
+
     private static boolean isInternalOption(String name) {
         return name.startsWith(INTERNAL_PROPERTY_PREFIX) ||
             name.startsWith("org.gradle.unsafe.") || // TODO: avoid reading public 'unsafe' properties via internal options


alllex:

alllex:

alllex:

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/InternalOption.java
```diff
@@ -24,9 +24,18 @@
  * @param <T> The value of the option.
  */
 public interface InternalOption<T extends @Nullable Object> extends Option {
```
mlopatkin: 🤔  I wonder if we need the separate interface now. Can we convert `InternalOption` into an abstract class and merge `AbstractInternalOption` into it? 

Otherwise, it is still possible to bypass the checks by simply implementing the interface. I know it is a weak-ish argument, but I have another - one less class to think about is also nice.

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/DefaultFeatureFlags.java
```diff
@@ -18,27 +18,27 @@
 
 import org.gradle.internal.event.ListenerManager;
 
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
 
 public class DefaultFeatureFlags implements FeatureFlags {
```
mlopatkin, line 25: 💭 Interesting that this thing is actually backing a public API, it should never have used `InternalOptions` to begin with.

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/InternalOption.java
alllex: Having to use `AbstractInternalOption` at the field declaration sides is a bit of a mouthful. Even the `InternalOption` gets verbose when used together with a nullable argument type.

I opted for splitting the validation into two parts. The second part validates that we only use `AbstractInternalOption` instances when requesting values of options:
https://github.com/gradle/gradle/pull/35357/files#diff-4ba831befa40be165dd469ed64f36012e69f02c7e1cf5689cce6d5b4661964b1R32

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/DefaultFeatureFlags.java
alllex: My thoughts exactly

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/InternalOption.java
mlopatkin: > Having to use `AbstractInternalOption` at the field declaration sides is a bit of a mouthful. Even the `InternalOption` gets verbose when used together with a nullable argument type.

Noting prevents us from using `InternalOption` as the abstract base type name.

alllex: Right!

PR #34827, opened 2025-08-28
Define proper nullability for Factory's generic argument

mlopatkin:
This helps to avoid null checks in places where the producer returns non-null value.

This commit also fixes some of the places that use Factory with improper nullability.

Part of #34769.

PR #35360, opened 2025-10-17
Better nullability support for build options and internal options

alllex:
This change is an entirely internal refactoring.

Mainly, annotates classes in `org.gradle.internal.buildoption` with `@Nullable`.

Also make `Option.Value` null-parameteric, such that definitions of options can decide whether they provide nullable or non-nullable option values. For instance, what was previously a `StringInternalOption` with undefined nullness, is now replaced with `InternalOption<String>` and `InternalOption<@Nullable String>`. This also plays well with Kotlin null-inference. Overall, this is similar to what was done in https://github.com/gradle/gradle/pull/34827

### DIFF
--- a/platforms/core-configuration/core-kotlin-extensions/src/main/kotlin/org/gradle/internal/extensions/core/InternalOptionsExtensions.kt
+++ b/platforms/core-configuration/core-kotlin-extensions/src/main/kotlin/org/gradle/internal/extensions/core/InternalOptionsExtensions.kt
@@ -25,9 +25,9 @@ fun InternalOptions.getInternalFlag(systemPropertyName: String, defaultValue: Bo
     getOption(InternalFlag(systemPropertyName, defaultValue)).get()
 
 
-fun InternalOptions.getInternalString(systemPropertyName: String, defaultValue: String): String =
+fun InternalOptions.getString(systemPropertyName: String, defaultValue: String): String =
     getOption(StringInternalOption.of(systemPropertyName, defaultValue)).get()
 
 
-fun InternalOptions.getInternalStringOrNull(systemPropertyName: String): String? =
+fun InternalOptions.getStringOrNull(systemPropertyName: String): String? =
     getOption(StringInternalOption.of(systemPropertyName)).get()
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/ConfigurationCacheStartParameter.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/ConfigurationCacheStartParameter.kt
@@ -26,7 +26,7 @@ import org.gradle.internal.buildtree.BuildModelParameters
 import org.gradle.internal.cc.impl.ConfigurationCacheLoggingParameters
 import org.gradle.internal.cc.impl.Workarounds
 import org.gradle.internal.extensions.core.getInternalFlag
-import org.gradle.internal.extensions.core.getInternalStringOrNull
+import org.gradle.internal.extensions.core.getStringOrNull
 import org.gradle.internal.extensions.stdlib.unsafeLazy
 import org.gradle.internal.service.scopes.Scope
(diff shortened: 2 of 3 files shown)


alllex:

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/AbstractBuildOption.java
```diff
@@ -18,62 +18,72 @@
...
+    @Nullable
     protected final String deprecatedProperty;
 
-    public AbstractBuildOption(String property) {
-        this(property, null, Collections.<V>emptyList());
+    public AbstractBuildOption(@Nullable String property) {
+        this(property, null, emptyList());
     }
 
     @SuppressWarnings("unchecked")
```
mlopatkin: 🤔 (a bit unrelated) Is it still unchecked? My IDEA complains about redundant suppression here

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/AbstractBuildOption.java
```diff
@@ -18,62 +18,72 @@
...
 
+    @Nullable
     protected final String property;
     protected final List<V> commandLineOptionConfigurations;
+    @Nullable
     protected final String deprecatedProperty;
 
-    public AbstractBuildOption(String property) {
-        this(property, null, Collections.<V>emptyList());
+    public AbstractBuildOption(@Nullable String property) {
```
mlopatkin: ❓ Does it make sense to have an option without any way to specify it - no Gradle property, no command line?

platforms/core-configuration/core-kotlin-extensions/src/main/kotlin/org/gradle/internal/extensions/core/InternalOptionsExtensions.kt
```diff
@@ -25,5 +25,11 @@ fun InternalOptions.getInternalFlag(systemPropertyName: String, defaultValue: Bo
     getOption(InternalFlag(systemPropertyName, defaultValue)).get()
 
 
-fun InternalOptions.getInternalString(systemPropertyName: String, defaultValue: String?) =
+fun InternalOptions.getInternalString(systemPropertyName: String, defaultValue: String): String =
+    getOption(StringInternalOption(systemPropertyName, defaultValue)).get()
+
+
+@JvmName("getInternalStringOrNull")
+@Suppress("RedundantNullableReturnType") // Option.Value can't carry nullability information
```
mlopatkin: 🤔 as we discussed, `Option.Value` can be defined as `Value<T extends @Nullable Object>`, then it will carry nullability information. Proper definition is generally better than suppression, and maybe it isn't that hard to add to this PR rather than leaving it?

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/AbstractBuildOption.java
alllex: Yes, if you remove this suppression, then Errorprone fails the compilation with some flavor of unsafe varargs error.

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/AbstractBuildOption.java
alllex: Good catch

platforms/core-configuration/core-kotlin-extensions/src/main/kotlin/org/gradle/internal/extensions/core/InternalOptionsExtensions.kt
alllex: With your guidance, it turned out not too hard to make this work properly by making `Value` null-parametric. I initially wanted to do this in a follow-up, but it seems the change is not too large, so it can be included as part of this PR.

PR #35366, opened 2025-10-17
Align parsing of public and internal flag values

alllex:
Before this change, public and internal flags had different parsing results: (both are case insensitive)
- Public flags: `"true" -> true`, anything else is `false`
- Internal flags `"false" -> false` anything else is `true`

While the behavior has been obviously misaligned, one practical difference I can make out is that supplying an **empty string** for internal flags resulted in parsing them to `true`. This might have saved typing a few characters, but many internal options actually default to `true` already, so that parsing isn't favoring an overwhelming majority.

Since we shouldn't change the public behavior without good reasons, this PR makes internal flag parsing behave exactly like the public one does, preventing any surprises for Gradle developers.

### DIFF
--- a/platforms/core-runtime/build-option/src/test/groovy/org/gradle/internal/buildoption/DefaultInternalOptionsTest.groovy
+++ b/platforms/core-runtime/build-option/src/test/groovy/org/gradle/internal/buildoption/DefaultInternalOptionsTest.groovy
@@ -22,35 +22,34 @@ class DefaultInternalOptionsTest extends Specification {
     def sysProps = [:]
     def options = new DefaultInternalOptions(sysProps)
 
-    def "locates value for boolean option"() {
-        sysProps["prop1"] = "true"
-        sysProps["prop2"] = ""
-        sysProps["prop3"] = "false"
-        sysProps["prop4"] = "not anything much"
+    def "locates value for boolean option #description"() {
+        sysProps["prop1"] = sysProp
 
-        expect:
-        def value1 = options.getOption(new InternalFlag("prop1", false))
-        value1.get()
-        value1.explicit
+        when:
+        def value = options.getOption(new InternalFlag("prop1"))
 
-        def value2 = options.getOption(new InternalFlag("prop2", false))
-        !value2.get()
-        value2.explicit
+        then:
+        value.explicit
+        value.get() == result
 
-        def value3 = options.getOption(new InternalFlag("prop3", true))
-        !value3.get()
-        value3.explicit
-
-        def value4 = options.getOption(new InternalFlag("prop4", false))
-        !value4.get()
-        value4.explicit
+        where:
+        description                    | sysProp         | result
+        "for true"                     | "true"          | true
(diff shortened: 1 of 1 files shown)


alllex:

platforms/core-runtime/build-option/src/test/groovy/org/gradle/internal/buildoption/DefaultInternalOptionsTest.groovy
```diff
@@ -34,15 +34,15 @@ class DefaultInternalOptionsTest extends Specification {
...
         def value2 = options.getOption(new InternalFlag("prop2", false))
-        value2.get()
+        !value2.get()
         value2.explicit
 
         def value3 = options.getOption(new InternalFlag("prop3", true))
         !value3.get()
         value3.explicit
 
         def value4 = options.getOption(new InternalFlag("prop4", false))
```
abstratt: 💭 

Since the default values are not relevant in this test (all options use explicit values), I would use single-arg constructor instead. 

(took me a bit to understand whether there was difference between line 40 and 44 - the only difference is in the explicitly assigned values

platforms/core-runtime/base-services/src/integTest/groovy/org/gradle/internal/work/ResourceLockStatisticsIntegrationTest.groovy
```diff
@@ -53,7 +53,7 @@ class ResourceLockStatisticsIntegrationTest extends AbstractIntegrationSpec {
         """
 
         when:
-        executer.withArguments("--parallel", "-D${DefaultBuildOperationsParameters.VERBOSE_LOCKING_OPERATIONS_OPTION.systemPropertyName}")
+        executer.withArguments("--parallel", "-D${DefaultBuildOperationsParameters.VERBOSE_LOCKING_OPERATIONS_OPTION.systemPropertyName}=true")
```
abstratt, line 56: 👍

platforms/core-runtime/build-option/src/test/groovy/org/gradle/internal/buildoption/DefaultInternalOptionsTest.groovy
alllex: Good point, I've updated the tests!

platforms/core-runtime/build-option/src/test/groovy/org/gradle/internal/buildoption/DefaultInternalOptionsTest.groovy
```diff
@@ -22,35 +22,34 @@ class DefaultInternalOptionsTest extends Specification {
...
-        !value3.get()
-        value3.explicit
+        then:
+        value.explicit
+        value.get() == result
 
-        def value4 = options.getOption(new InternalFlag("prop4", false))
-        value4.get()
-        value4.explicit
+        where:
```
abstratt, line 35: 👍

PR #35371, opened 2025-10-17
Adopt CC html problem report fixtures in IDE sync tests

abstratt:
Issue: #35040

This change set ensures `ConfigurationCacheProblemFixture` can be used by smoke tests.

One main limitation it had is that `ConfigurationCacheProblemFixture` was originally too tied to `GradleExecuter`/`ExecutionResult`/`ExecutionFailure`.

 * The `GradleExecuter`-related functionality was moved to a new, more specialized fixture, `ConfigurationCacheExecutionResultProblemsFixture`. Clients that were `GradleExecuter` based now use this fixture implementation, instead of `ConfigurationCacheProblemFixture`. This new implementation adds a layer that understands `ExecutionResult` and can find the report location from the output as found in `ExecutionResult`.
* `ConfigurationCacheProblemFixture` assert methods now support three use cases: client provides a root dir and the console output (the report file location should be found in the console output), client provides a root dir (which should contain a report dir at the standard location), or client provides a base-report dir (which is really used in the unit tests)  
* Adopted `CCPF` in the IDE sync tests and  the `Gradleception` smoke test (introduced as the fix for #35040)
* `IsolatedProjectsIdeSyncFixture` is now just a subclass for `ConfigurationCacheProblemFixture` that adds zero features). Its own version of `HasConfigurationCacheProblemsInReportSpec` has been removed.   
* Also changes CODEOWNERS so @gradle/bt-isolated-projects-project shows as reviewer for IP smoke tests


abstratt:
@bot-gradle test AST

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGracefulDegradationIntegrationTest.groovy
```diff
@@ -241,7 +241,6 @@ class ConfigurationCacheGracefulDegradationIntegrationTest extends AbstractConfi
         problems.assertResultHtmlReportHasProblems(result) {
             totalProblemsCount = 2
             withProblem(INVOCATION_OF_TASK_PROJECT_AT_EXECUTION_TIME)
-            withProblem(INVOCATION_OF_TASK_PROJECT_AT_EXECUTION_TIME)
```
abstratt, line 244: Unique problems should not appear more than once

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheExecutionResultProblemsFixture.groovy
```diff
@@ -0,0 +1,141 @@
...
+import org.gradle.integtests.fixtures.executer.ExecutionResult
+import org.gradle.util.internal.ConfigureUtil
+import org.hamcrest.Matcher
+
+import static org.junit.Assert.assertThrows
+
+/**
+ * A `ConfigurationCacheProblemsFixture` that supports `ExecutionResult`.
+ */
+final class ConfigurationCacheExecutionResultProblemsFixture extends ConfigurationCacheProblemsFixture {
```
abstratt: Extracted from `ConfigurationCacheProblemsFixture`. This now only handles assertions based on `ExecutionResult` and `ExecutionFailure`

abstratt:
@bot-gradle test this

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -333,45 +223,63 @@ final class ConfigurationCacheProblemsFixture {
...
+    protected static void assertInputs(
         String output,
         File rootDir,
         HasConfigurationCacheProblemsSpec spec
     ) {
         assertItems('input', output, rootDir, spec.inputs)
     }
 
-    private static void assertIncompatibleTasks(
+    protected static void assertInputs(
```
abstratt, line 243: Needs to be exposed to the `ExecutionResult` -based problem fixture

testing/smoke-ide-test/src/smokeIdeTest/groovy/org/gradle/ide/sync/fixtures/IsolatedProjectsIdeSyncFixture.groovy
```diff
@@ -17,96 +17,10 @@
...
-import java.nio.file.Path
-import java.nio.file.SimpleFileVisitor
-import java.nio.file.attribute.BasicFileAttributes
-
-class IsolatedProjectsIdeSyncFixture {
-
-    private final ConfigurationCacheProblemsFixture configurationCacheProblemsFixture
-    private final TestFile rootDir
-
+class IsolatedProjectsIdeSyncFixture extends ConfigurationCacheProblemsFixture {
```
abstratt, line 22: Keeping it around as a subclass to `ConfigurationCacheProblemsFixture` instead of removing it and replacing all call sites to use `ConfigurationCacheProblemsFixture` instead to reduce churn and as a potential home for IDE-sync related functionality that we may require. Right now, it provides zero additional features over `ConfigurationCacheProblemsFixture`.

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
```diff
@@ -38,11 +41,17 @@ class GradleBuildIsolatedProjectsSmokeTest extends AbstractGradleBuildIsolatedPr
 
         then:
         result.assertConfigurationCacheStateStoreDiscarded()
-        result.output.contains "13 problems were found storing the configuration cache, 5 of which seem unique."
-        result.output.contains "- Plugin 'org.jetbrains.kotlin.jvm': Project ':declarative-dsl-core' cannot dynamically look up a property in the parent project ':'"
-        result.output.contains "- Plugin 'org.jetbrains.kotlin.jvm': Project ':declarative-dsl-evaluator' cannot dynamically look up a property in the parent project ':'"
-        result.output.contains "- Plugin 'org.jetbrains.kotlin.jvm': Project ':declarative-dsl-tooling-models' cannot dynamically look up a property in the parent project ':'"
-        result.output.contains "- Plugin 'org.jetbrains.kotlin.jvm': Project ':kotlin-dsl-plugins' cannot dynamically look up a property in the parent project ':'"
-        result.output.contains "- Unknown location: Project ':docs' cannot dynamically look up a property in the parent project ':'"
+
```
abstratt, line 44: Now leveraging `CCPF`  to assert on HTML report contents instead of asserting on console output directly

abstratt:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/fixtures/AbstractConfigurationCacheOptInFeatureIntegrationTest.groovy
```diff
@@ -18,18 +18,18 @@ package org.gradle.internal.cc.impl.fixtures
...
 import org.gradle.initialization.StartParameterBuildOptions
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
-import org.gradle.integtests.fixtures.configurationcache.ConfigurationCacheProblemsFixture
+import org.gradle.integtests.fixtures.configurationcache.ConfigurationCacheExecutionResultProblemsFixture
 
 abstract class AbstractConfigurationCacheOptInFeatureIntegrationTest extends AbstractIntegrationSpec {
     static final String WARN_PROBLEMS_CLI_OPT = "--${StartParameterBuildOptions.ConfigurationCacheProblemsOption.LONG_OPTION}=warn"
 
-    protected ConfigurationCacheProblemsFixture problems
+    protected ConfigurationCacheExecutionResultProblemsFixture problems
```
abstratt: `GradleExecuter`-based tests should use this subclass

testing/internal-integ-testing/src/test/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixtureTest.groovy
```diff
@@ -0,0 +1,360 @@
...
+import org.gradle.test.fixtures.file.TestNameTestDirectoryProvider
+import org.junit.Rule
+import spock.lang.Specification
+
+import static org.gradle.integtests.fixtures.configurationcache.ConfigurationCacheProblemsFixture.resolveConfigurationCacheReport
+
+/**
+ * A unit test for {@link ConfigurationCacheProblemsFixture}
+ */
+class ConfigurationCacheProblemsFixtureTest extends Specification {
```
abstratt, line 29: Testing fixture suitability for smoke tests got tiring very quickly. Added this fixture test so I could iterate on fixture changes more quickly.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -45,234 +49,127 @@ import static org.hamcrest.CoreMatchers.notNullValue
...
     protected static final String PROBLEMS_REPORT_HTML_FILE_NAME = "configuration-cache-report.html"
 
-    private final File rootDir
+    protected final TestFile rootDir
 
     ConfigurationCacheProblemsFixture(File rootDir) {
-        this.rootDir = rootDir
-    }
-
-    void assertFailureHasError(
```
abstratt, line 61: Moved all assertions taking `ExecutionResult|Failure` to the new `ConfigurationCacheExecutionResultProblemsFixture `

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -45,234 +49,127 @@ import static org.hamcrest.CoreMatchers.notNullValue
...
-        ExecutionFailure failure,
-        String error,
-        @DelegatesTo(value = HasConfigurationCacheErrorSpec, strategy = Closure.DELEGATE_FIRST) Closure<?> specClosure
-    ) {
-        assertFailureHasError(failure, error, ConfigureUtil.configureUsing(specClosure))
+        this.rootDir = rootDir instanceof TestFile ? rootDir : new TestFile(rootDir)
```
abstratt, line 60: wrapping plain `File` objects as `TestFile` as a convenience here, instead of requiring call sites to do it

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -45,234 +49,127 @@ import static org.hamcrest.CoreMatchers.notNullValue
...
     }
 
-    void assertFailureHasError(
-        ExecutionFailure failure,
-        String error,
-        Action<HasConfigurationCacheErrorSpec> specAction = {}
```
abstratt, line 72: These methods used to have three variants: 
- `Closure`-based
- `Action`-based 
- `HasConfigurationCacheProblemsSpec`-based

With the need to support a new axis of variation (client provides a report dir vs client provides a root dir + console output which can then be used to find a report location) this was getting out of hand, so I inlined calls to the second (action-based) variant.

testing/smoke-ide-test/src/smokeIdeTest/groovy/org/gradle/ide/sync/IsolatedProjectsGradleceptionSyncTest.groovy
```diff
@@ -35,10 +35,12 @@ class IsolatedProjectsGradleceptionSyncTest extends AbstractIdeSyncTest {
         then:
         report.assertHtmlReportHasProblems {
             totalProblemsCount = 12
-            withLocatedProblem("Gradle runtime", "Project ':declarative-dsl-core' cannot dynamically look up a property in the parent project ':'")
```
abstratt, line 38: @6hundreds I am assuming `withLocatedProblem` is not different than `withUniqueProblem`(s) - other than maybe asserting the location.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheExecutionResultProblemsFixture.groovy
alllex: 💅 `ConfigurationCacheProblemsExecutionResultFixture` to keep "problems" attributed to their issuer

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -1,5 +1,5 @@
 /*
```
abstratt, line 1: Apologies for the huge diff... I tried to clarify the motivation for the changes 
 the PR description, but please ask for clarifications on anything that does not seem obvious

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -502,6 +427,29 @@ final class ConfigurationCacheProblemsFixture {
         }
     }
 
+    File findReportDir() {
+        return resolveSingleConfigurationCacheReportDir(rootDir)
+    }
+
+    private static TestFile resolveSingleConfigurationCacheReportDir(TestFile rootDir) {
```
abstratt, line 431: This was moved from `IsolatedProjectsSyncFixture`.

testing/smoke-ide-test/src/smokeIdeTest/groovy/org/gradle/ide/sync/fixtures/IsolatedProjectsIdeSyncFixture.groovy
```diff
@@ -17,96 +17,10 @@
...
-
-    private static HasConfigurationCacheProblemsInReportSpec createSpec(Closure<?> specClosure) {
-        def spec = new HasConfigurationCacheProblemsInReportSpec()
-        specClosure.delegate = spec
-        specClosure()
-        spec.validateSpec()
-        return spec
-    }
-
-    private static TestFile resolveSingleConfigurationCacheReportDir(TestFile rootDir) {
```
abstratt, line 95: Moved to `ConfigurationCacheProblemsFixture`

testing/smoke-ide-test/src/smokeIdeTest/groovy/org/gradle/ide/sync/fixtures/IsolatedProjectsIdeSyncFixture.groovy
```diff
@@ -17,96 +17,10 @@
...
-
-    void assertHtmlReportHasProblems(
-        @DelegatesTo(value = HasConfigurationCacheProblemsInReportSpec, strategy = Closure.DELEGATE_FIRST) Closure<?> specClosure
-    ) {
-        def spec = createSpec(specClosure)
-        def jsModel = readJsModelFromReport()
-        checkProblemsAgainstModel(spec, jsModel)
-    }
-
-    private void checkProblemsAgainstModel(HasConfigurationCacheProblemsInReportSpec spec, Map<String, Object> jsModel) {
```
abstratt, line 53: This is now relying on `CCFProblems`

abstratt:
@bot-gradle test this

abstratt:
@bot-gradle test this

.github/CODEOWNERS
```diff
@@ -26,15 +26,16 @@ testing/internal-testing/                   @gradle/bt-developer-productivity
...
-testing/soak/                           @gradle/bt-developer-productivity
-testing/distributions-integ-tests       @gradle/bt-developer-productivity
-testing/public-api-tests                @gradle/bt-developer-productivity
+packaging/core-platform/                        @gradle/bt-product-operations
+packaging/distributions-dependencies/           @gradle/bt-product-operations
+packaging/distributions-full/                   @gradle/bt-product-operations
+packaging/public-api                            @gradle/bt-product-operations
+testing/performance/                            @gradle/bt-developer-productivity
+testing/smoke-test/                             @gradle/bt-product-operations
+testing/smoke-test/src/**/*IsolatedProjects*    @gradle/bt-isolated-projects-project @gradle/bt-product-operations
```
abstratt: Added this line, as suggested here:

https://github.com/gradle/gradle/pull/35308#discussion_r2435041400

testing/smoke-ide-test/src/smokeIdeTest/groovy/org/gradle/ide/sync/IsolatedProjectsGradleceptionSyncTest.groovy
6hundreds: @abstratt Don't we care about locations anymore? I think it might be useful to assert on locations as well, wdyt?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -543,7 +489,7 @@ final class ConfigurationCacheProblemsFixture {
         return (jsModel.diagnostics as List<Object>).count { it['problem'] != null && it['error']?.getAt('parts') != null }
     }
 
-    private static ProblemsSummary extractSummary(String text) {
+    protected static ProblemsSummary extractSummary(String text) {
```
6hundreds, line 492: 💭  I see this method is the only one not related to html report in this class?
As a possible option - move it to the call site `CCPERF`, and rename this class to `ConfigurationCacheProblemsHtmlReportFixture`. 

Currently, `smoke-ide-tests` can't leverage this method facility - it's not possible to extract Gradle console log from the IDE process, but I can't be 100% confident it will remains like this in future.

.github/CODEOWNERS
alllex: 💅 I'd follow the pattern of putting the main owners first, and more specific later, like in other places in the file. This means `@gradle/bt-product-operations` would go first.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsExecutionResultFixture.groovy
```diff
@@ -0,0 +1,141 @@
...
+
+    void assertFailureHasError(
+        ExecutionFailure failure,
+        String error,
+        @DelegatesTo(value = HasConfigurationCacheErrorSpec, strategy = Closure.DELEGATE_FIRST) Closure<?> specClosure
+    ) {
+        assertFailureHasError(failure, newErrorSpec(error, ConfigureUtil.configureUsing(specClosure)))
+    }
+
+    void assertFailureHasError(
```
alllex, line 44: 🤔  If I didn't miss anything, there are no real usages of `assertFailureHasError` and its overloads on `ConfigurationCacheProblemsExecutionResultFixture`, and also no real usages of `newErrorSpec` and its overloads on `ConfigurationCacheProblemsFixture`.

If that is indeed the case, should we remove the methods altogether?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -463,18 +385,18 @@ final class ConfigurationCacheProblemsFixture {
...
                 }
             }
         } else {
-            assertThat("Unexpected HTML report URI found", reportDir, nullValue())
+            assertThat("Unexpected HTML report URI found", reportDir?.with { it.directory ? it : null }, nullValue())
         }
     }
 
-    static Map<String, Object> readJsModelFromReportDir(TestFile reportDir) {
+    private static Map<String, Object> readJsModelFromReportDir(File reportDir) {
```
alllex, line 399: 🤔 This should probably keep the `TestFile` argument. The IDE complains about the `reportDir.file()` call below, because it thinks a different method is called.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -502,6 +424,29 @@ final class ConfigurationCacheProblemsFixture {
         }
     }
 
+    File findReportDir() {
```
alllex, line 427: 💅 should probably return `TestFile`

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -813,3 +759,4 @@ class HasConfigurationCacheProblemsSpec {
         return this
     }
 }
+
```
alllex: 💅 An extra newline?

testing/smoke-ide-test/src/smokeIdeTest/groovy/org/gradle/ide/sync/IsolatedProjectsGradleceptionSyncTest.groovy
alllex: I agree with Sergey, that it would be nice not to lose the location info. Though, we should probably do this in a follow-up PR. This depends on how much changes are required.

For reference, in IP tests we use basically the same fixture methods, and the location information is simply included in this string that we assert. Maybe we can do the same here?

https://github.com/gradle/gradle/blob/30dd9b0619e7af7ef8e1e1d5bde75b435801fa0b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsProblemReportingIntegrationTest.groovy#L42-L44

abstratt: In the existing tests that assert on HTML report problems we haven't cared about location (also, the only case we were using the removed capability in the IP Sync tests had always the same location - "Gradle Runtime" - so at this time we were not benefitting much from it). 

Note that the example you pointed to is asserting on the console output - HTML report problem verification is not performed by default (see [`checkReportProblems`](https://github.com/search?q=repo%3Agradle%2Fgradle%20checkReportProblems&type=code)). There is no support for location matching yet that works for both console and report problems.  We can add that capability back if we miss it in the future.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
abstratt: It is one of the few indeed, but there are some other methods that rely on it from here (`assertHasConsoleSummary`, `assertOutputHasError`). I felt the a similar impulse to rename this class, and further separate the two kinds of functionalities, but the HTML and console assertion facilities are still closely related, and share the same spec types, so didn't feel there was enough pressure yet to make it happen.

> Currently, smoke-ide-tests can't leverage this method facility - it's not possible to extract Gradle console log from the IDE process, but I can't be 100% confident it will remain like this in future.

Do you mean, in the future we may need to assert on console output from tests that do not rely on `GradleExecuter`?  That is possible already via `assertOutputHasError`, for example, which will validate the console output, and perform some basic checks on the report.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsExecutionResultFixture.groovy
abstratt: They were unused before this change set, including `HasConfigurationCacheErrorSpec`. It seems these were used originally for asserting against the "Configuration cache state could not be cached: <reason>" CC store/load error message but no call sites remain. I may look into deleting them in a follow-up PR.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
abstratt: I changed instead the code so it would work with a regular `java.io.File` as we don't need `TestFile` capabilities anyways:

```
new File(reportDir, PROBLEMS_REPORT_HTML_FILE_NAME)
```

(or else I would have to change many places and force clients to provide `TestFile` where `File` would have been enough)

abstratt:
@bot-gradle squash and merge

PR #32672, opened 2025-03-12
Re-implement AbstractTestTask using new report APIs

octylFractal:
Does not yet handle JUnit XML or TestReporter, just HTML reports

Fixes #32317

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #35374, opened 2025-10-17
Improve FileUtils.toSafeFileName length limiting

octylFractal:
This commit makes it truncate in bytes instead of characters, which is appropriate for Linux. Additionally, instead of just truncating to the limit, which might cause issues with collisions when used for test names, use a shorter limit and add a decent hash to the name.

### Context
This fixes bugs in #32672.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/SafeFileLocationUtils.java
+++ b/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/SafeFileLocationUtils.java
@@ -39,8 +39,19 @@ import java.util.Arrays;
 public final class SafeFileLocationUtils {
     public static final int WINDOWS_PATH_LIMIT = 260;
 
+    // SipHash-2-4 provides decent collision resistance of 64 bits while being fast to compute
     private static final HashFunction HASHER = Hashing.sipHash24();
     private static final BaseEncoding BASE_ENCODING = BaseEncoding.base32Hex().omitPadding();
+
+    /**
+     * A short, distinctive prefix to indicate that a file name has been truncated.
+     *
+     * <p>
+     * Uses underscores as they are likely valid characters on all filesystems,
+     * and visually distinctive. Uses "cut" instead of "truncated" to keep it short.
+     * </p>
+     */
+    private static final byte[] TRUNCATED_PREFIX_BYTES = "_cut_".getBytes(StandardCharsets.UTF_8);
 
     /**
      * The maximum file name length in bytes for most filesystems (e.g. ext4, NTFS).
@@ -60,7 +71,7 @@ public final class SafeFileLocationUtils {
      * </p>
      */
     @VisibleForTesting
-    static final int MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES = MAX_FILE_NAME_LENGTH_IN_BYTES - 1 - (HASHER.bits() + 4) / 5;
+    static final int MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES = MAX_FILE_NAME_LENGTH_IN_BYTES - TRUNCATED_PREFIX_BYTES.length - 1 - (HASHER.bits() + 4) / 5;
 
     /**
      * The character used to replace illegal characters in file names.
@@ -89,10 +100,15 @@ public final class SafeFileLocationUtils {
             .replace('\r', ILLEGAL_CHAR_REPLACEMENT);
     }
 
-    private static final CharsetEncoder UTF_8_ENCODER_WITH_ILLEGAL_CHAR_REPLACEMENT = StandardCharsets.UTF_8.newEncoder()
-        .onMalformedInput(CodingErrorAction.REPLACE)
-        .onUnmappableCharacter(CodingErrorAction.REPLACE)
-        .replaceWith(String.valueOf(ILLEGAL_CHAR_REPLACEMENT).getBytes(StandardCharsets.UTF_8));
+    // CharsetEncoder is not thread-safe, so use ThreadLocal to hold one per thread
+    // These are not very expensive memory-wise, so this should be OK
+    private static final ThreadLocal<CharsetEncoder> UTF_8_ENCODER_WITH_ILLEGAL_CHAR_REPLACEMENT =
+        ThreadLocal.withInitial(() ->
+            StandardCharsets.UTF_8.newEncoder()
+                .onMalformedInput(CodingErrorAction.REPLACE)
+                .onUnmappableCharacter(CodingErrorAction.REPLACE)
+                .replaceWith(String.valueOf(ILLEGAL_CHAR_REPLACEMENT).getBytes(StandardCharsets.UTF_8))
+        );
 
     /**
      * Remove malformed or un-mappable characters and shorten the name if necessary.
@@ -103,7 +119,7 @@ public final class SafeFileLocationUtils {
     private static String cleanAndShortenName(String name) {
         byte[] rawName;
         try {
-            ByteBuffer buffer = UTF_8_ENCODER_WITH_ILLEGAL_CHAR_REPLACEMENT.encode(CharBuffer.wrap(name));
+            ByteBuffer buffer = UTF_8_ENCODER_WITH_ILLEGAL_CHAR_REPLACEMENT.get().encode(CharBuffer.wrap(name));
             rawName = new byte[buffer.remaining()];
             buffer.get(rawName);
         } catch (CharacterCodingException e) {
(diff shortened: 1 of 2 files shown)


octylFractal:
I had to refactor out `SafeFileLocationUtils` in order to not add additional code to the tooling-api shaded JAR. Let me know if I should do something else.

platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/SafeFileLocationUtilsTest.groovy
```diff
@@ -0,0 +1,139 @@
...
+        input                       | output
+        'A' * 256                   | 'A' * MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES + '-9G531233RIUS4'
+        ('A' * 253) + 'Θ'           | 'A' * MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES + '-JJTN3CQ249KPK'
+        // Hash should preserve extension
+        ('A' * 256) + '.html'       | 'A' * (MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - Utf8.encodedLength('.html')) + '-EG3TV8Q6QCCQS.html'
+        ('A' * 256) + '.Θ'          | 'A' * (MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - Utf8.encodedLength('.Θ')) + '-N32UNHUNL5N5O.Θ'
+        'Θ' + ('A' * 300) + '.html' | 'Θ' + ('A' * (MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - Utf8.encodedLength('Θ') - Utf8.encodedLength('.html'))) + '-5O3MPEP5S1RIC.html'
+        'Θ' + ('A' * 300) + '.Θ'    | 'Θ' + ('A' * (MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - Utf8.encodedLength('Θ') - Utf8.encodedLength('.Θ'))) + '-27045PA297J5I.Θ'
+        // Extension is only preserved if it fits, otherwise normal truncation occurs.
+        'A.' + ('B' * 300)          | 'A.' + ('B' * (MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - 'A.'.length())) + '-3CF50NO32LPNI'
```
tresat: ❓ I think we might want to try harder to preserve the extension.  At least for extensions of reasonable length.  Extension-less files are annoying.

Could we do something like: If the extension is <= 7 chars, include it as it (truncating up to the extra 7 chars from the name if needed); otherwise replace the extension with .unknown (7 char extension)?

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/SafeFileLocationUtils.java
```diff
@@ -0,0 +1,215 @@
...
+        .onUnmappableCharacter(CodingErrorAction.REPLACE)
+        .replaceWith(String.valueOf(ILLEGAL_CHAR_REPLACEMENT).getBytes(StandardCharsets.UTF_8));
+
+    /**
+     * Remove malformed or un-mappable characters and shorten the name if necessary.
+     *
+     * @param name the original name
+     * @return the cleaned and potentially shortened name
+     */
+    private static String cleanAndShortenName(String name) {
```
tresat, line 119: 🤔 I wonder if it might be helpful to also add a prefix, so that all affected file names are sorted together.

Something like: "TRUNCATED-Some-Original-Name-Fragment-HASHCHARS.extension"

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/SafeFileLocationUtils.java
```diff
@@ -0,0 +1,215 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+/**
+ * Sibling class to {@link FileUtils}, focused on obtaining safe file locations and names.
+ */
+public final class SafeFileLocationUtils {
+    public static final int WINDOWS_PATH_LIMIT = 260;
+
+    private static final HashFunction HASHER = Hashing.sipHash24();
```
tresat, line 43: 🤔 Can you add a note as to why this particular hash function is used?

tresat (approved):
I'm going to merge this now, so we can rebase the feature branch and keep it moving.  I think it might be worth following this up (either now or later) with a few tweaks.

platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/SafeFileLocationUtilsTest.groovy
octylFractal: This is how it works aside from `.unknown`, see the `('A' * 256) + '.html'` test case.

If an extension is that long I think it's just part of the filename and not an extension. We might get dots because of test names in here, so I think truncating is a better behavior if we can't preserve it.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/SafeFileLocationUtils.java
octylFractal: Added as `_cut_`.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/SafeFileLocationUtils.java
octylFractal: Done.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/Binary2JUnitXmlReportGenerator.java
```diff
@@ -100,7 +100,7 @@ public void execute(final TestClassResult result) {
     }
 
     private String getReportFileName(TestClassResult result) {
-        return REPORT_FILE_PREFIX + FileUtils.toSafeFileName(result.getClassName()) + REPORT_FILE_EXTENSION;
+        return REPORT_FILE_PREFIX + SafeFileLocationUtils.toSafeFileName(result.getClassName()) + REPORT_FILE_EXTENSION;
```
big-guy, line 103: Doesn't this technically mean we'll generate a file name that's too long when the classname exceeds the safe file name length - REPORT_FILE_EXTENSION.length?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/ClassTestResults.java
```diff
@@ -46,7 +46,7 @@ public ClassTestResults(long id, String name, String displayName, PackageTestRes
         this.name = name;
         this.displayName = displayName;
         this.packageResults = packageResults;
-        baseUrl = HTML_REPORT_PREFIX + FileUtils.toSafeFileName(name) + HTML_EXTENSION;
+        baseUrl = HTML_REPORT_PREFIX + SafeFileLocationUtils.toSafeFileName(name) + HTML_EXTENSION;
```
big-guy, line 49: Same thing here

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
```diff
@@ -71,7 +71,7 @@ public static String getFilePath(org.gradle.util.Path path) {
         if (path.segmentCount() == 0) {
             filePath = "index.html";
         } else {
-            filePath = String.join("/", Iterables.transform(path.segments(), FileUtils::toSafeFileName)) + "/index.html";
+            filePath = String.join("/", Iterables.transform(path.segments(), SafeFileLocationUtils::toSafeFileName)) + "/index.html";
```
big-guy, line 74: Here is OK because we're just shortening an intermediate path

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/Binary2JUnitXmlReportGenerator.java
octylFractal: Hmm, you're right. We should probably pass the prefix and extension with the rest of the file name.

octylFractal: https://github.com/gradle/gradle/pull/35390

PR #35390, opened 2025-10-21
More safe file name fixes

octylFractal:
### Context
Follow-up to #35374 

Also fixes a bug where the chars weren't filtered in uncut names.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-execution/persistent-cache/build.gradle.kts
+++ b/platforms/core-execution/persistent-cache/build.gradle.kts
@@ -17,7 +17,6 @@ dependencies {
 
     api(libs.jspecify)
 
-    implementation(projects.baseServices)
     implementation(projects.io)
     implementation(projects.time)
 
--- a/platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFileLockManager.java
+++ b/platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFileLockManager.java
@@ -32,7 +32,6 @@ import org.gradle.cache.internal.filelock.LockStateAccess;
 import org.gradle.cache.internal.filelock.LockStateSerializer;
 import org.gradle.cache.internal.filelock.Version1LockStateSerializer;
 import org.gradle.cache.internal.locklistener.FileLockContentionHandler;
-import org.gradle.internal.SafeFileLocationUtils;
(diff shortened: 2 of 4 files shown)


platforms/core-runtime/base-services/src/main/java/org/gradle/internal/SafeFileLocationUtils.java
```diff
@@ -117,24 +123,56 @@ public static String toSafeFileName(String name) {
...
-        byte[] rawName;
-        try {
-            ByteBuffer buffer = UTF_8_ENCODER_WITH_ILLEGAL_CHAR_REPLACEMENT.get().encode(CharBuffer.wrap(name));
-            rawName = new byte[buffer.remaining()];
-            buffer.get(rawName);
-        } catch (CharacterCodingException e) {
-            // We use the REPLACE action, so this should never happen
-            throw new AssertionError("should never encounter an encoding error", e);
+        Utf8EncodingResult nameResult = encodeIntoUtf8WithReplacement(name);
+        if (nameResult.cleanBytes.length <= MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES) {
```
tresat: 🤔 Would it be worth adding an early check ahead of attempting encode: If given a short string made up of entirely safe charactors, just return as is?  

This could probably could avoid the work of encoding the vast majority of the time.  Just check char-by-char for Strings of length < 20?30? chars for chars in the basic ASCII range minus the few illegal ones.

Maybe an idea for later.

tresat (approved):

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/SafeFileLocationUtils.java
octylFractal: In terms of performance, this is low on the list of things that could be tried to improve it I think. Some of this code is already purposefully prioritizing readability over performance (e.g., in the buffer usages) because I think this is not called often enough for it to matter that much, and the strings should generally be quite short anyways.

tresat (approved):

tresat:
@bot-gradle test RfN

PR #35394, opened 2025-10-21
Consolidate base script and buildSrc names

alllex:
with a primary goal of removing dependencies from layout-related classes
on classes like `Settings` and `DefaultProjectDescriptor` for the purposes of platformization.

### DIFF
--- a/subprojects/core-api/src/main/java/org/gradle/api/initialization/Settings.java
+++ b/subprojects/core-api/src/main/java/org/gradle/api/initialization/Settings.java
@@ -84,6 +84,8 @@ import java.util.Arrays;
 public interface Settings extends PluginAware, ExtensionAware {
     /**
      * <p>The default name for the settings file.</p>
+     *
+     * @implNote For internal purposes, prefer constants from {@code BuildLogicFiles}.
      */
     String DEFAULT_SETTINGS_FILE = "settings.gradle";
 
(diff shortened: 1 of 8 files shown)


alllex:

subprojects/core-api/src/main/java/org/gradle/api/initialization/Settings.java
```diff
@@ -85,7 +86,7 @@ public interface Settings extends PluginAware, ExtensionAware {
     /**
      * <p>The default name for the settings file.</p>
      */
-    String DEFAULT_SETTINGS_FILE = "settings.gradle";
+    String DEFAULT_SETTINGS_FILE = BuildLogicFiles.DEFAULT_SETTINGS_FILE;
```
lptr: For the constants already in _internal_ classes I think it can make sense to pull them out into an single internal class as you do. However, for this one, it actually makes things worse, as here you are introducing a reference from a public API interface to an internal type. Arrows should point in the opposite direction. I think it would make sense to always refer to this as `Settings.DEFAULT_SETTINGS_FILE` instead.

If you insist on moving this constant out as well, I think deprecating and removing this constant would be the way. That way you essentially declare that the default file name is not part of the public API; a reasonable thing to say I think.

alllex: Good catch!

At the same time, we don't want `BuildLogicFiles` depending on the `Settings`, nor do we want other places to depend on `Settings`. Unfortunately, `Settings` brings with it a heap of compilation dependencies that are not needed in the majority of places and hence depending on it becomes a blocker for platformization.

Deprecating the public field in Settings makes sense. Though, I wouldn't want to do it in the same PR.

I'll restore the constant here and add a comment for now.

Issue #35211, opened 2025-10-03
Support third-party model builders to run after configuration failures, but only for successful builds


asodja:
Third-party model builders should be able to run on configuration failures, but only for builds in the builds tree that were succesfully configured. 

For example, if we have a project with an incuded build as a build logic and included build configuration is successfuly but main build configuration fails, we should allow third-party model builders to run for included build that was successful.

Example project:
```
settingsKotlinFile << """
    pluginManagement {
        includeBuild("build-logic")
    }
    rootProject.name = "root"
    plugins {
        id("build-logic")
    }
"""
def included = file("build-logic")
included.file("settings.gradle.kts") << """
    rootProject.name = "build-logic"

    pluginManagement {
        repositories {
            mavenCentral()
            gradlePluginPortal()
        }
    }
"""
included.file("build.gradle.kts") << """
    plugins {
        `kotlin-dsl`
    }

    repositories {
        mavenCentral()
        gradlePluginPortal()
    }
"""
def settingsPlugin = included.file("src/main/kotlin/build-logic.settings.gradle.kts") << """
    broken !!!
"""
```
In this case `build-logic` build with configure successfully, but `main` build will fail at evalution of settings, since applied `build-logic.settings.gradle.kts` is broken. Third-party model builders should still be allowed to query `build-logic` build, since configuration of it was successful.

asodja:
I added a test here: https://github.com/gradle/gradle/blob/4e1be54e1668f9c9839584fdfdaca0ba85077942/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy

We should implement this test with this feature.

PR #35395, opened 2025-10-21
 Allow querying all models on successful parts of a build

asodja:
Fixes https://github.com/gradle/gradle/issues/35211


asodja:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
```diff
@@ -41,15 +41,19 @@ public ResilientBuildToolingModelController(
...
+            super.configureProjectsForModel(target, modelName);
         } catch (GradleException e) {
-            rethrowExceptionIfNotResilientModel(modelName, e);
+            rethrowExceptionIfNotResilientModel(target, modelName, e);
         }
     }
 
-    private static void rethrowExceptionIfNotResilientModel(String modelName, GradleException e) {
+    private static void rethrowExceptionIfNotResilientModel(ProjectState target, String modelName, GradleException e) {
+        if (target.getOwner().isProjectsConfigured()) {
```
alllex: ❓  How does configure-on-demand fit into this?

CoD is an "incubating" feature, but we know many users, including large ones, are using it. And it will come in some form for IP in the future. Let's make sure the API we provide here does not paint us into a corner.

asodja: What is the behaviour with `configure-on-demand` with model builders? Aren't all projects always configured with model building?

alllex: With configure-on-demand, the usual stage that configured projects ("loads the build") is effectively skipped:
https://github.com/gradle/gradle/blob/e40b729bd700dcca4149458fa804c5dd909f88d0/subprojects/core/src/main/java/org/gradle/configuration/DefaultProjectsPreparer.java#L43

This means that each individual project will be configured only when it is directly asked to do so. For this reason, each individual model builder (for a project-scope) will make sure it's project is configured, before running it:
https://github.com/gradle/gradle/blob/2879727a0a4d76fd2498d9d0780bbbb99b5744e9/subprojects/core/src/main/java/org/gradle/internal/build/DefaultBuildToolingModelController.java#L139-L140

But there is no top-level point-in-time, where we'd know all projects have been configured. It will basically depend on how the models are queried. To be more specific, with CoD, if a build action only requests a model for the root project, then only the root project will be configured.

---

To be clear, I didn't look in the changes in the PR in detail, so I am not sure, how the lifecycle of projects is managed for resilient model builders, if it's any different.

asodja: The lifecycle is the same. So then I guess with the CoD mode we would allow to query more: we would allow all successfully configured projects in a build until one project fails.

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/ResilientKotlinDslScriptsModelBuilder.kt
```diff
@@ -1,49 +0,0 @@
...
-
-import org.gradle.api.Project
-import org.gradle.api.internal.GradleInternal
-import org.gradle.internal.problems.failure.FailureFactory
-import org.gradle.kotlin.dsl.tooling.builders.AbstractKotlinDslScriptsModelBuilder
-import org.gradle.tooling.provider.model.ToolingModelBuilder
-import org.gradle.tooling.provider.model.internal.ToolingModelBuilderResultInternal
-
-internal
-class ResilientKotlinDslScriptsModelBuilder(
```
asodja: 💭 We actually don't need it with new logic 🎉

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,268 @@
...
+        description                     | extraGradleProperties        | expectedSuccesfulProjects         | expectedFailedProjects
+        ""                              | [""]                         | ['build-logic']                   | ['root', 'a', 'b', 'c']
+        " with configuration-on-demand" | IP_CONFIGURE_ON_DEMAND_FLAGS | ['root', 'a', 'c', 'build-logic'] | ['b']
+    }
+
+    /**
+     * Requires GradleBuild model to run on runtime failures
+     */
+    @ToBeImplemented
+    def "can query custom model for included build without build configuration errors, even if main settings fail"() {
```
asodja: 💭 This test will start working once we merge https://github.com/gradle/gradle/pull/35478

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
```diff
@@ -16,68 +16,149 @@
...
+            }
+
+            // For configure-on-demand, projects are configured individually, so failure at this point means that the project failed to configure.
+            // For normal mode assume project was successfully configured if the owner build was fully configured, e.g., full included build configuration was successful.
+            return !isConfigurationOnDemand && isOwnerBuildSuccessfullyConfigured;
+        }
+
+        private static boolean canRunEvenIfProjectNotFullyConfigured(String modelName) {
+            // Some internal model builders can run even if the project is not fully configured.
+            return RESILIENT_MODELS.contains(modelName);
```
asodja: 💭 We could expose actual builder type by adding a new to `ToolingModelBuilderLookup.Builder`. And then  `resilient model builders` could implement some new interface.

But I left this out for now to not have too many changes here.

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
```diff
@@ -16,68 +16,149 @@
...
         @Override
         ToolingModelBuilderLookup.Builder locateBuilder() throws UnknownModelException {
-            // Force configuration of the target project to ensure all builders have been registered, but ignore failures
-            try {
-                target.ensureConfigured();
-            } catch (GradleException e) {
-                rethrowExceptionIfNotResilientModel(modelName, e);
-            }
+            // Force configuration of the target project to ensure all builders have been registered
+            ConfigurationResult projectConfigurationResult = tryRunConfiguration(target::ensureConfigured);
```
asodja: 💭 First I had a different logic here, something like:

```
ConfigurationResult configurationResult = isConfigureOnDemand || fullBuildConfigurationResult.isSuccess()
? tryRunConfiguration(target::ensureConfigured)
: fullBuildConfigurationResult;
```

Which translates to:

> Run ensureConfigured only if full build configuration was succcesful or is configure-on-demand. 

But I figured that this is a problem for `KotlinDslScriptsModelBuilder`.
With current logic on master we run `KotlinDslScriptsModelBuilder` even if project/build configuration failed. That is why I changed the logic to always call  `tryRunConfiguration(target::ensureConfigured)`. With that `KotlinDslScriptsModelBuilder` is registered even if `fullBuildConfigurationResult` failed before. And we can get `KotlinDSLScriptsModels` even for failed projects. I wonder if there could be some problems with it.

This is something we need to discuss.

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/ResilientKotlinDslScriptsModelBuilder.kt
asodja, line 27: 💭 We actually don't need this class with the new logic 🎉

subprojects/core/src/main/java/org/gradle/internal/build/DefaultBuildToolingModelController.java
```diff
@@ -56,28 +57,39 @@ public ToolingModelScope locateBuilderForTarget(String modelName, boolean param)
...
-        configureProjectsForModel(modelName);
-        return doLocate(target, modelName, param);
+        ConfigurationResult configurationResult = configureProjects();
+        return doLocate(target, modelName, param, configurationResult);
     }
 
-    protected void configureProjectsForModel(String modelName) {
-        buildController.configureProjects();
+    protected ConfigurationResult configureProjects() {
+        return tryRunConfiguration(buildController::configureProjects);
```
asodja, line 81: 💭 We could actually use `buildController::configureProjectsIgnoringLaterFailures` also here, that way we would allow to run model builders on successful included builds also in resilient mode

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/KotlinScriptingModelBuildersRegistrant.kt
```diff
@@ -0,0 +1,63 @@
...
+class KotlinScriptingModelBuildersRegistrar(
+    private val modelParameters: BuildModelParameters,
+    private val intermediateModelProvider: IntermediateToolingModelProvider
+) : ToolingModelBuilderRegistrar {
+
+    override fun registerForProject(registry: ToolingModelBuilderRegistry, isRootProject: Boolean) {
+        registry.register(KotlinBuildScriptModelBuilder)
+        registry.register(IsolatedScriptsModelBuilder)
+
+        if (isRootProject) {
```
bamboo, line 41: 🎗️ These models should likely become build-level model builders since they also carry `Settings` information

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/KotlinScriptingModelBuildersRegistrant.kt
```diff
@@ -0,0 +1,63 @@
...
+                else -> KotlinDslScriptsModelBuilder
+            }
+            registry.register(builder)
+        }
+    }
+
+    /**
+     * A [ProjectConfigureAction] that registers the [KotlinDslModelsParameters.PREPARATION_TASK_NAME] task required for Kotlin DSL scripts models.
+     */
+    class KotlinScriptingModelBuildersTaskRegisterAction : ProjectConfigureAction {
```
bamboo, line 53: 🚀 We should consider introducing a `RootProjectConfigureAction` extension point so this action doesn't have to be applied to every single project for no reason

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/KotlinScriptingModelBuildersServices.kt
```diff
@@ -39,4 +46,14 @@ class KotlinScriptingModelBuildersServices : AbstractGradleModuleServices() {
                 }
             }
     }
+
+    private object ProjectScopeToolingServices : ServiceRegistrationProvider {
+        @Provides
+        fun createToolingModelBuilderRegistrar(
+            modelParameters: BuildModelParameters,
+            intermediateModelProvider: IntermediateToolingModelProvider
+        ): ToolingModelBuilderRegistrar {
```
bamboo: ❓ Are there other examples of project configure actions that could be transformed into registrars?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+            rootProject.name = "root"
+            include("a", "b", "c")
+            includeBuild("build-logic")
+        """
+
+        def included = file("build-logic")
+        included.file("settings.gradle.kts") << """
+            rootProject.name = "build-logic"
+
+            pluginManagement {
```
bamboo, line 97: 💅 Isn't there a constant for `pluginManagement` repository declarations that could be interpolated here?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+                    gradlePluginPortal()
+                }
+            }
+        """
+        included.file("build.gradle.kts") << """
+            plugins {
+                `kotlin-dsl`
+            }
+
+            repositories {
```
bamboo: 💅 Isn't there a constant for project level repository declarations that could be interpolated here?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+                .run()
+        }
+
+        then:
+        result.failedToQueryProjects == expectedFailedProjects
+        result.successfullyQueriedProjects == expectedSuccesfulProjects
+
+        where:
+        description                     | extraGradleProperties        | expectedSuccesfulProjects         | expectedFailedProjects
+        ""                              | [""]                         | ['build-logic']                   | ['root', 'a', 'b', 'c']
```
bamboo: ❓ Does `extraGradleProperties` need the empty string element? Wouldn't an empty list work?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+        " with configuration-on-demand" | IP_CONFIGURE_ON_DEMAND_FLAGS | ['root', 'a', 'c', 'build-logic'] | ['b']
+    }
+
+    def "can query custom model for included build without build configuration errors, even if main settings fail"() {
+        settingsKotlinFile << """
+            pluginManagement {
+                includeBuild("build-logic")
+            }
+            rootProject.name = "root"
+            plugins {
```
bamboo, line 154: 💅 `plugins` block should appear before any statements

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+                id("build-logic")
+            }
+            include("a")
+        """
+
+        def included = file("build-logic")
+        included.file("settings.gradle.kts") << """
+            rootProject.name = "build-logic"
+
+            pluginManagement {
```
bamboo, line 165: ❓ Same as for other test

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+                    gradlePluginPortal()
+                }
+            }
+        """
+        included.file("build.gradle.kts") << """
+            plugins {
+                `kotlin-dsl`
+            }
+
+            repositories {
```
bamboo: ❓ Same as for other test

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    public void apply(Project project) {
+        println "Registered CustomBuilder for project: " + (project != null ? project.name : "<no project>")
+    }
+}
+"""
+    }
+
+    def "can query custom model for included build without build configuration errors, even if main project configuration fails#description"() {
```
bamboo: 👍

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+        result.failedToQueryProjects == expectedFailedProjects
+        result.successfullyQueriedProjects == expectedSuccesfulProjects
+
+        where:
+        description                     | extraGradleProperties        | expectedSuccesfulProjects         | expectedFailedProjects
+        ""                              | [""]                         | ['build-logic']                   | ['root', 'a', 'b', 'c']
+        " with configuration-on-demand" | IP_CONFIGURE_ON_DEMAND_FLAGS | ['root', 'a', 'c', 'build-logic'] | ['b']
+    }
+
+    def "can query custom model for included build without build configuration errors, even if main settings fail"() {
```
bamboo: 👍

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+            }
+        """
+        included.file("src/main/kotlin/build-logic.gradle.kts") << """
+            broken !!!
+        """
+        file("a/build.gradle.kts") << """
+            plugins {
+                id("java")
+            }
+
```
bamboo: 💅 Spurious empty line

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    static class ModelAction implements BuildAction<ModelResult>, Serializable {
+
+        @Override
+        ModelResult execute(BuildController controller) {
+            GradleBuild gradleBuild = controller.getModel(GradleBuild.class)
+            List<String> successfulQueriedProjects = []
+            List<String> failedQueriedProjects = []
+            for (BasicGradleProject project : gradleBuild.projects) {
```
bamboo, line 235: 🧪 It might be good to test with queries in a different order as well, e.g., `editableBuilds` before projects, projects in different order, etc

subprojects/core/src/main/java/org/gradle/internal/build/BuildLifecycleController.java
```diff
@@ -58,9 +58,19 @@ public interface BuildLifecycleController {
...
+     *
+     * When calling it after a project already configured, and any later state failed, e.g., tasks, this method will fail too.
      */
     void configureProjects();
 
+    /**
+     * Configures the projects of the build, if not already done.
+     * Can be called multiple times.
+     *
+     * Ignores any failures that occur in later states of the build, while {@link #configureProjects()} will fail if any later state fails.
```
bamboo: ❓ Please clarify _later states of the build_

subprojects/core/src/main/java/org/gradle/internal/build/DefaultBuildToolingModelController.java
bamboo: > that way we would allow to run model builders on successful included builds also in resilient mode

❓ Isn't that exactly what this PR is making possible?

subprojects/core/src/main/java/org/gradle/internal/build/DefaultBuildToolingModelController.java
```diff
@@ -142,4 +154,45 @@ ToolingModelBuilderLookup.Builder locateBuilder() throws UnknownModelException {
             return lookup.locateForClientOperation(modelName, parameter, target, target.getMutableModel());
         }
     }
+
+    protected static class ConfigurationResult {
```
bamboo: ❓ Do we actually need a new type? Isn't this basically `Optional<GradleException>` with some helper functions?

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
```diff
@@ -16,68 +16,127 @@
...
+    private final FailureFactory failureFactory;
+
     public ResilientBuildToolingModelController(
         BuildState buildState,
         BuildLifecycleController buildController,
-        ToolingModelBuilderLookup buildScopeLookup) {
+        ToolingModelBuilderLookup buildScopeLookup
+    ) {
         super(buildState, buildController, buildScopeLookup);
+        this.failureFactory = buildController.getGradle().getServices().get(FailureFactory.class);
```
bamboo: ❓ Shouldn't this be a constructor parameter?

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
```diff
@@ -16,68 +16,127 @@
...
-        return new ResilientProjectToolingScope(target, modelName, param);
+    protected ToolingModelScope doLocate(ProjectState target, String modelName, boolean param, ConfigurationResult configurationResult) {
+        return new ResilientProjectToolingScope(target, failureFactory, modelName, param, configurationResult);
     }
 
     private static class ResilientProjectToolingScope extends ProjectToolingScope {
-        public ResilientProjectToolingScope(ProjectState target, String modelName, boolean parameter) {
+
+        private final FailureFactory failureFactory;
+        private final ConfigurationResult fullBuildConfigurationResult;
```
bamboo: ❓ `ownerBuildConfigurationResult` perhaps?

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
```diff
@@ -16,68 +16,127 @@
...
+            return delegate.getParameterType();
+        }
+
+        @Override
+        public Object build(@Nullable Object parameter) {
+            if (configurationResult.isSuccess()) {
+                return delegate.build(parameter);
+            }
+
+            Object model = canRunEvenIfProjectNotFullyConfigured(modelName) ? delegate.build(parameter) : null;
```
bamboo: 🤔 It seems this decision could/should have been made in `locateBuilder`?

subprojects/core/src/main/java/org/gradle/tooling/provider/model/internal/ToolingModelBuilderRegistrar.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Registrar that registers one or more {@link org.gradle.tooling.provider.model.ToolingModelBuilder} to {@link ToolingModelBuilderRegistry}.
+ *
+ * It's used to register internal builders at project scope that can run even if the project is not configured (e.g., Kotlin DSL scripts model).
+ *
+ * Normally builders should be registered at project configuration time and {@link ToolingModelBuilderRegistrar} should be used just for exceptional cases.
+ */
+@NullMarked
+@ServiceScope(Scope.Project.class)
+public interface ToolingModelBuilderRegistrar {
```
bamboo: 🤔 The suffix `Registrar` doesn't feel right here as it gives me the impression that this services keeps the records while it simply acts as a one-shot registration action.

I think `RegistrationProvider` would work better as a suffix here since its role is similar to [ServiceRegistrationProvider](https://github.com/gradle/gradle/blob/94d5a6eaa078bb5e67994854ff02d8290da9430c/platforms/core-runtime/service-provider/src/main/java/org/gradle/internal/service/ServiceRegistrationProvider.java#L140).

How about renaming to `ToolingModelBuilderRegistrationProvider`?

subprojects/core/src/main/java/org/gradle/internal/build/DefaultBuildToolingModelController.java
asodja: Yes, you are right, the comment is incorrect. It should say `in non-resilient mode`

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/KotlinScriptingModelBuildersServices.kt
asodja: It seems the Kotlin action is the only model registration action

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy
asodja: Good catch, this should be removed

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
asodja: There is for repositories, changed to using `$repositoriesBlock` now

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy
asodja: Good catch, removed this

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
asodja: Fixed

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
asodja: Fixed

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy
asodja: Fixed

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/CustomResilientModelCrossVersionSpec.groovy
asodja: Removed spurious empty line

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
asodja: I am now testing more variations, either we query editable builds first or root build first

subprojects/core/src/main/java/org/gradle/internal/build/BuildLifecycleController.java
asodja: Improved the JavaDoc

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
asodja: It would be nicer yes, fixed

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
asodja: Makes sense, renamed

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
asodja: I moved call to `canRunEvenIfProjectNotFullyConfigured(modelName)` to `locateBuilder`. 

Did you image that we move some more logic there and `ResilientToolingModelBuilder` builder would just handle "failed configuration" case? We could avoid additional builder wrapper when build configuration is succesful

subprojects/core/src/main/java/org/gradle/internal/build/DefaultBuildToolingModelController.java
asodja: I think that with a special type it's easier to understand what this result represents. `Optional<GradleException>` could be anything, while for `ConfigurationResult` you know it's a result of configuration operation. 

Additioanly wheny you pass `Optional` in different scopes/methods, it's easy to loose context.

subprojects/core/src/main/java/org/gradle/tooling/provider/model/internal/ToolingModelBuilderRegistrar.java
asodja: Renamed to `ToolingModelBuilderRegistrationProvider`, it may be more specific than `registrar`. I would expect `Provider` to provide something (return a value), but it's probably not too confusing in this context.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/KotlinScriptingModelBuildersRegistrant.kt
asodja: That would make sense indeed. 👍 I haven't tackle this in this PR though

asodja:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/tooling/provider/model/internal/ToolingModelBuilderRegistrar.java
bamboo: The suffix `Registrant` just occurred to me. `ToolingModelBuilderRegistrant` is clear and shorter.

subprojects/core/src/main/java/org/gradle/internal/build/DefaultBuildToolingModelController.java
bamboo: Reading the code again, it seems this is more like a `Try<Void>`, anyway...

I agree that having a more specific name to the type helps with comprehension but I'm not fully convinced we need so much code, specially if some of these methods are only used once.

If this was implemented in Kotlin I'd probably have something like `value class ConfigurationResult(val result: Try<Unit>)`

asodja: I removed `ConfigurationResult` and I use just `Try<Void>` now

subprojects/core/src/main/java/org/gradle/tooling/provider/model/internal/ToolingModelBuilderRegistrar.java
asodja: That makes sense, renamed to ToolingModelBuilderRegistrant

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/internal/build/BuildLifecycleController.java
```diff
@@ -58,9 +58,22 @@ public interface BuildLifecycleController {
     /**
      * Configures the projects of the build, if not already done.
      * Can be called multiple times.
+     *
+     * If called after build have already been configured and a later build lifecycle state
```
bamboo:
```suggestion
     * If called after the build has already been configured and a later build lifecycle state
```

platforms/core-configuration/kotlin-dsl-tooling-builders/build.gradle.kts
```diff
@@ -9,19 +9,18 @@ dependencies {
     api(projects.core)
     api(projects.serviceProvider)
     api(libs.kotlinStdlib)
+    api(projects.baseServices)
```
bamboo: 💅 Keep `api(projects.*)` together and sorted by name

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.groovy
```diff
@@ -587,17 +587,15 @@ class ResilientKotlinDslScriptsModelBuilderCrossVersionSpec extends ToolingApiSp
...
         rootBuildFailure?.with {
             expectFailureToContain(resilientModels.failures[settingsKotlinFile.parentFile], it)
         }
-        includedBuildFailure?.with {
-            expectFailureToContain(resilientModels.failures[included], it)
-        }
+        resilientModels.failures[included] == null
 
         where:
         queryStrategy         | expectedNoOfFailures | rootBuildFailure                               | includedBuildFailure
```
bamboo: ❓ Is `includedBuildFailure` even used anymore?

platforms/core-configuration/kotlin-dsl-tooling-builders/build.gradle.kts
asodja: Thanks, fixed!

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.groovy
asodja: Good catch, we can remove it. 

Done!

asodja:
@bot-gradle test this

PR #35399, opened 2025-10-21
Remove old codepath from `kotlin-dsl` plugin support for precompiled scripts

eskatos:
In Gradle 9.0 we moved to Kotlin 2. As a best effort and to support a seamless upgrade, we kept the `kotlin-dsl` plugin support for precompiled scripts work with previous versions of the `kotlin-dsl` plugin. This is a best effort because this is actually not supported.

In this change we remove the support for the old way. Users not using the blessed version of the `kotlin-dsl` plugin are getting a warning that this is not supported since Gradle 9.0.

This change is a step on the way to get rid of the long-time unsupported Kotlin 1 internal embeddable compiler APIs that stop working with the upcoming Kotlin 2.3.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


eskatos:

jbartok (approved):

Issue #35320, opened 2025-10-15
Make DefaultResilientIssuesRecorder thread safe


asodja:
In https://github.com/gradle/gradle/pull/35254 we added a `DefaultResilientIssuesRecorder`. But it's currently not thread safe. 

Since Tooling API allows parallel model fetching we should make `DefaultResilientIssuesRecorder` and associate returned models with exceptions collected in builders.

asodja:
Also a thread with some more context how parallel fetch works:
https://gradle.slack.com/archives/C092LN2SS3C/p1760539448870359?thread_ts=1760538189.096559&cid=C092LN2SS3C

PR #35400, opened 2025-10-21
Introduce ToolingModelBuilderResultInternal and remove ResilientIssuesRecorder

asodja:
And with that fix issue with DefaultResilientIssuesRecorder not being thread safe.

### Context
Previously we had `ResilientIssuesRecorder` and implementation `DefaultResilientIssuesRecorder` where `ResilientGradleBuildBuilder` could report failures for a model. We figured out that `DefaultResilientIssuesRecorder` needs to be thread safe if parallel model fetch is enabled. But with parallel fetch it's then difficult to associate a failure with a model. And a problem becomes even trickier if model caching is enabled, since we need to cache also failures.

With this PR internal `ToolingModelBuilder` can return a `ToolingModelBuilderResultInternal` that can contain additional information, like failures. We then unpack this result in `DefaultBuildToolingModelController`. With that we also don't have any problem with caching failures if model caching is enabled, since failures are cached with a model as `ToolingModelBuilderResultInternal`.

Fixes https://github.com/gradle/gradle/issues/35320


asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

PR #35407, opened 2025-10-21
Forbid inheriting from DefaultServiceRegistry

alllex:
Introduces a check that forbids unknown inheritors of `DefaultServiceRegistry`, basically only allowing itself and `ScopedServiceRegistry`.

This allows to get rid of _another way of doing things_, where one could declare `@Provides` service methods directly on the `ServiceRegistry` class. This capability has already been replaced by the explicit use of `ServiceRegistrationProviders` in all production code a while ago, so what has remained is the usage in tests. This reduction in flexibility simplifes the mental model of service registries, which is much needed, given its complexity.

As such, this change is largerly about cleaning up the test code to use registry and service composition over inheritance.

### DIFF
--- a/platforms/core-execution/persistent-cache/src/integTest/groovy/org/gradle/cache/internal/DefaultFileLockManagerContentionIntegrationTest.groovy
+++ b/platforms/core-execution/persistent-cache/src/integTest/groovy/org/gradle/cache/internal/DefaultFileLockManagerContentionIntegrationTest.groovy
@@ -286,24 +286,27 @@ class DefaultFileLockManagerContentionIntegrationTest extends AbstractIntegratio
                 }
             }
 
-            class ZincCompilerServices extends DefaultServiceRegistry {
+            class ZincCompilerServices {
                 private static def instance
 
-                private ZincCompilerServices(File gradleUserHome) {
-                    super(NativeServices.getInstance())
-
-                    add(OutputEventListener.class, OutputEventListener.NO_OP)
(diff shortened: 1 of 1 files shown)


alllex:

alllex:

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -84,13 +63,21 @@
...
-public class DefaultServiceRegistry implements CloseableServiceRegistry, ContainsServices, ServiceRegistrationProvider {
+public class DefaultServiceRegistry implements CloseableServiceRegistry, ContainsServices {
     private enum State {INIT, STARTED, CLOSED}
 
     private final static ServiceRegistry[] NO_PARENTS = new ServiceRegistry[0];
     private final static Service[] NO_DEPENDENTS = new Service[0];
     private final static Object[] NO_PARAMS = new Object[0];
 
+    // Simulation of a sealed class with public constructors in Java 8
+    private final static Set<Class<? extends DefaultServiceRegistry>> ALLOWED_IMPLEMENTATIONS = new HashSet<>();
```
abstratt: 💅 Use an unmodifiable set

```
private final static Set<Class<? extends DefaultServiceRegistry>> ALLOWED_IMPLEMENTATIONS = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
    DefaultServiceRegistry.class,
    ScopedServiceRegistry.class
)));    
```

alllex: Good idea, updated

PR #35082, opened 2025-09-22
Enable Non-Class-Based Testing

tresat:

PR #35409, opened 2025-10-21
Change DirectoryBasedTestDefinition to only store relative path from …

tresat:
…project root

Avoid sending absolute paths to test workers.


tresat:
@bot-gradle test this

tresat:
This was an experiment that doesn't seem like it's needed for #35082.

PR #35410, opened 2025-10-21
Support Java 26

### LINKED ISSUE #35405: Support compiling and testing on Java 26
https://jdk.java.net/26/

### Steps to complete
- [ ] Add JDK to JDK yaml
- [ ] Bump ASM to 9.9 (https://asm.ow2.io/versions.html)
- [ ] Update `AsmConstants`
- [ ] Remove unsupported warning from `JavaVersion`
- [ ] Add future entry to `JavaVersion` to maintain 2-year window of future version constants
- [ ] Upgrade Groovy
- [ ] Update `JacocoCoverage`, `GroovyCoverage`, `ScalaCoverage`, `PmdCoverage` JDK versions if necessary
- [ ] Update compatibility table
- [ ] Update `JavaVersionSpec`

octylFractal:
Fixes #35405 and fixes https://github.com/gradle/gradle/issues/35406

Pmd doesn't support 26 yet, and Scala seems like it won't be clear of `Unsafe` until 3.8.

The ASM `Remapper` constructor got an `api` parameter added to it, which affects what methods are called. But we don't use the affected methods, so I just added our `ASM_LEVEL`.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/code-quality/src/integTest/groovy/org/gradle/api/plugins/quality/SamplesCodeQualityIntegrationTest.groovy
+++ b/platforms/jvm/code-quality/src/integTest/groovy/org/gradle/api/plugins/quality/SamplesCodeQualityIntegrationTest.groovy
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010 the original author or authors.
+ * Copyright 2026 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,10 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.gradle.integtests.samples
+package org.gradle.api.plugins.quality
 
 
-import org.gradle.api.plugins.quality.PmdPlugin
 import org.gradle.integtests.fixtures.AbstractSampleIntegrationTest
 import org.gradle.integtests.fixtures.Sample
 import org.gradle.integtests.fixtures.UsesSample
--- a/platforms/documentation/samples/build.gradle.kts
+++ b/platforms/documentation/samples/build.gradle.kts
@@ -16,7 +16,6 @@ dependencies {
     integTestImplementation(libs.samplesCheck) {
         exclude(group = "org.codehaus.groovy", module = "groovy-all")
     }
-    integTestImplementation(testFixtures(projects.codeQuality))
     integTestImplementation(testFixtures(projects.core))
     integTestImplementation(testFixtures(projects.modelCore))
 


jvandort:
We should update this as well https://github.com/gradle/gradle/blob/master/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/DefaultGradleDistribution.groovy#L32

octylFractal:
That is for when we support running Gradle on 26, which I had as #35406 

Honestly, the running Gradle support has gotten so small that I was debating doing the PRs together from now on. It used to be that we would need to significantly delay for Groovy & Kotlin, but Groovy releases have been better and I did a workaround for Kotlin. I'll combine them together.

jvandort:
Ah my bad I knew we tracked support for these separately but wasn't thinking apparently

.teamcity/jdks.yaml
```diff
@@ -49,6 +49,20 @@ jdks:
     vendor: "temurin"
     version: "jdk-25+36"
     sha256: "ee04de95ab9da7287d40bd2173076ecc2a6dd662f007bedfc6eb0380c0ef90e8"
+  - params:
+      - "linux.java26.openjdk.aarch64"
```
cobexer, line 53: ❓ as there are no references to this in the `.teamcity` Kotlin code, no Gradle instance will detect Java 26 on CI

cobexer: https://github.com/gradle/gradle/pull/35422 will make this better, but it's not that simple ... thus not today^^

octylFractal: I did forget to update the values in `Jdk`, but there is a concern around removing 21 from being tested at all (I was planning to update them to 17/25/26). Some of our tests even require specifically JDK 21 right now. What do you think should be done?

cobexer: We need to take a look at the implications

big-guy: are the tests that need 21 specifically asking for it? I think we can test with 17/25/26 and have individual tests still request a wider range (e.g., 8-25)

platforms/documentation/docs/src/docs/userguide/releases/compatibility.adoc
```diff
@@ -58,7 +58,7 @@ See the table below for the Java version supported by a specific Gradle release:
 | 23| 8.10 | 8.10 and after
 | 24| 8.14 | 8.14 and after
 | 25| 9.1.0 | 9.1.0 and after
-| 26| N/A | N/A
+| 26| 9.3.0 | 9.3.0
```
big-guy: and after?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/runtimeshaded/ImplementationDependencyRelocator.java
```diff
@@ -54,6 +55,7 @@ protected void doExecute(InputStream thing) throws Exception {
     }
 
     public ImplementationDependencyRelocator(RuntimeShadedJarType type) {
+        super(AsmConstants.ASM_LEVEL);
```
big-guy, line 58: What impact does this have?

.teamcity/jdks.yaml
octylFractal: Yes, they specifically ask for it, e.g. `DaemonToolchainCrossVersionTest` or the `custom-test-task` sample. I could adjust them, but it would be some effort.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/runtimeshaded/ImplementationDependencyRelocator.java
octylFractal: I partially mentioned it in the PR description, but the full impact is that `Remapper.mapInvokeDynamicMethodName(String, String)` is no longer called, and  the new method `Remapper.mapInvokeDynamicMethodName(String, String, Handle, Object...)` is. We don't use these so it doesn't affect us.

.teamcity/jdks.yaml
cobexer, line 1: ❌ in the current form this PR does not execute any tests with Java 26 on CI.

CI Agents will download Java 26 into a location that is not known to any started Gradle processes.

cobexer (changes requested):
No CI coverage currently

.teamcity/jdks.yaml
cobexer: Let's make sure Java 21 stays visible to the Gradle builds but move the Java versions up as we usually do:

`MAX_LTS_VERSION` = 25
`MAX_VERSION` = 26

in https://github.com/gradle/gradle/blob/dcba4837a0b09669e28a904ab8aac82756ee7681/.teamcity/src/main/kotlin/common/JvmCategory.kt#L24-L25C5


Iff something ends up breaking on Java 21 only, users always have the option to upgrade to Java 25 (they will eventually anyway). That is way better than the reverse situation IMO.

We discussed this in todays status with @ljacomet, @ov7a and @blindpirate and all consider this the correct strategy.

octylFractal: Does what I just pushed look correct?

cobexer: looks about right!

octylFractal:
@bot-gradle test this

octylFractal:
@bot-gradle test this

cobexer:
@bot-gradle test RfN without PTS

octylFractal:
We decided that the Java 26 compiler bug is breaking too core of a feature for us to simply skip tests like `JavaProjectIntegrationTest.compilationFailureBreaksBuild` and merge this, so this PR will be parked for 9.4.0 (when we assume this bug will be fixed) and I'll open a separate one to merge in the parts that don't require CI changes, like JaCoCo or ASM updates.

octylFractal:
@bot-gradle test RfN without PTS

octylFractal:
@bot-gradle test RfN

platforms/jvm/code-quality/src/integTest/groovy/org/gradle/api/plugins/quality/SamplesCodeQualityIntegrationTest.groovy
```diff
@@ -15,14 +15,21 @@
...
+import org.gradle.quality.integtest.fixtures.PmdCoverage
 import org.gradle.test.fixtures.file.TestFile
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.UnitTestPreconditions
+import org.gradle.util.internal.VersionNumber
 import org.junit.Rule
 
+import static org.junit.Assume.assumeTrue
+
 class SamplesCodeQualityIntegrationTest extends AbstractSampleIntegrationTest {
```
big-guy, line 32: Could we move this to code-quality?

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AsciidoctorPluginSmokeTest.groovy
```diff
@@ -91,6 +91,8 @@ class AsciidoctorPluginSmokeTest extends AbstractPluginValidatingSmokeTest {
                     "Consult the upgrading guide for further information: ${BASE_URL}/userguide/upgrading_version_8.html#deprecated_startparameter_is_configuration_cache_requested",
                 "https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/751"
             )
+            // Asciidoc plugin currently triggers an --enable-native-access warning on Java 24+
+            runner.withJdkWarningChecksDisabled()
```
big-guy, line 95: Does this not trigger today with 25?

big-guy (approved):

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AsciidoctorPluginSmokeTest.groovy
octylFractal: We only run smoke tests on `MIN_VERSION` and `MAX_LTS_VERSION` (used to be 21), so it didn't run on 25 before.

octylFractal:
@bot-gradle test this

platforms/jvm/code-quality/src/integTest/groovy/org/gradle/api/plugins/quality/SamplesCodeQualityIntegrationTest.groovy
octylFractal: Done.

platforms/documentation/docs/src/docs/userguide/releases/compatibility.adoc
```diff
@@ -58,7 +58,7 @@ See the table below for the Java version supported by a specific Gradle release:
 | 23| 8.10 | 8.10 and after
 | 24| 8.14 | 8.14 and after
 | 25| 9.1.0 | 9.1.0 and after
-| 26| N/A | N/A
+| 26| 9.3.0 | 9.3.0 and after
```
reinsch82:
```suggestion
| 26| 9.4.0 | 9.4.0 and after
```
right?

octylFractal: Yes, thank you.

octylFractal:
@bot-gradle test this

octylFractal:
@bot-gradle test this

PR #34373, opened 2025-07-23
Extract build-discovery subproject

alllex:
As part of platformization efforts, this PR introduces two new subprojects:
- `core-runtime/build-discovery-api`
- `core-runtime/build-discovery`

extracting their respective contents from `core` and `core-api`.

---

See https://github.com/gradle/gradle/pull/34323 for more context and future plans for classes touched in this PR

PR #34323, opened 2025-07-18
Merge root and settings directory representation

alllex:
This is a refactoring PR that aims to simplify and clarify handling of the root-of-the-build directory handling.

There is more work to be done, but this change provides a start at the simplification:

- Represent build root directory and settings directory as a single state. They already are in the way the build layout is discovered, but it's not apparent from the way the implementation evolved historically.
- Introduce `BuildTreeLocations` service in the build-session scope and remove `BuildLayout` services from being build-session scoped. Previously, injecting `BuildLayout` would either inject the layout of the current build at the build scope or its children scope, or it inject the root-build layout at any parent scope of build-scope.

---

### Higher level goals

We should get to a point where internally we use only a single concept to represent the root of the build. Historically, there were two: root of the build and settings directory. At some point, it became impossible for these directories to be different, and we should leverage that fact, simplifying the codebase.

The required refactoring mostly resolve around `SettingsLocation` and `org.gradle.initialization.layout.BuildLayout`. Both are internal, but we know they have been used in the wild due to the ease of injection. We should be a bit more careful, when refactoring them: given their tenure, a deprecation and removal in a major version is warranted. But this is better done in a separate PR, for easy of reviewing.

Note, there is also a public counter-part of the build layout: `org.gradle.api.file.BuildLayout`. However, it's only available in the settings scope, and not in the project scope.

PR #35417, opened 2025-10-22
Extract build-discovery classes out of core, take 2

alllex:
Second take
- on https://github.com/gradle/gradle/pull/34373
- after https://github.com/gradle/gradle/pull/34566 happened

What is different this time?
- The class `GradleScript` has not been moved out from `core-api`, since there weren't direct dependencies on it, and it was previously the root case of classloading issues
- We now have (on master) the `KotlinDslIntellijPlatformIntegrationTest` [test](https://github.com/gradle/gradle/blob/0dd9a3c53b3ebb54ff6693c142fa0debf47599e3/platforms/core-configuration/kotlin-dsl-integ-tests/src/integTest/kotlin/org/gradle/kotlin/dsl/integration/KotlinDslIntellijPlatformIntegrationTest.kt#L40) that validates that necessary classes are available on the IntelliJ-crafted classpath to load Kotlin models
- We now also have (on master) the `GradleDslBaseScriptModelCrossVersionSpec` [test](https://github.com/gradle/gradle/blob/0dd9a3c53b3ebb54ff6693c142fa0debf47599e3/platforms/core-configuration/kotlin-dsl-tooling-builders/src/crossVersionTest/groovy/org/gradle/kotlin/dsl/tooling/builders/r92/GradleDslBaseScriptModelCrossVersionSpec.groovy) that validates the classpath we provide to the IDE can load Kotlin models as well.

---

As part of platformization efforts, this PR introduces two new subprojects:
- `core-runtime/build-discovery`
- `core-runtime/build-discovery-impl`

extracting their respective contents from `core` and `core-api`.

---

See https://github.com/gradle/gradle/pull/34323 for more context and future plans for classes touched in this PR


alllex:

platforms/core-runtime/client-services/build.gradle.kts
```diff
@@ -49,7 +51,6 @@ dependencies {
     // The client should not depend on core or core-api or projects that depend on these.
     // However, these project still contains some types that are shared between the client and daemon.
     api(projects.core)
-    api(projects.coreApi)
```
mlopatkin, line 52: 🎉  (I see that the implementation dependency is still there, but this is an improvement regardless)

platforms/core-runtime/build-discovery-impl/build.gradle.kts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+    id("gradlebuild.distribution.implementation-java")
```
mlopatkin, line 18: ❓ This module doesn't define a public API, right? I would avoid naming it `-api` then. Maybe a scheme proposed [here](https://docs.google.com/document/d/1fSIxzet6TKQXROfQxkGDpNfdqy1zTIH8hwTG6-7rqxs/edit?tab=t.0#heading=h.ou963jmcpepi) would work better:
- `build-discovery-api` -> `build-discovery`
- `build-discovery` -> `build-discovery-impl`

platforms/core-runtime/build-discovery-impl/build.gradle.kts
mlopatkin, line 1: ~~❓ Why do we need a separate "api" module? It even has more dependencies than the implementation.~~

NVM, I see that `StartParameter` pulls in all kinds of implementation dependencies 😿

mlopatkin (changes requested):
I have a suggestion for project naming, otherwise LGTM

platforms/core-runtime/build-discovery-impl/build.gradle.kts
alllex: I don't feel strongly and wasn't aware of that document.

@big-guy wdyt?

alllex: We definitely seem to have more `-api` modules than `-impl` today, some of which are not "public API"

mlopatkin: > We definitely seem to have more `-api` modules than `-impl` today, some of which are not "public API"

Most of them are, though. I don't think the pattern of `-api` not being API should proliferate.

alllex: It does make sense to avoid `-api` in the naming. I renamed the projects as you suggested.

However, that naming is also at odds with the naming suggested in the platformization document: now many modules, including `core` depend on a `-impl`. This suggests that the second module (`-impl` after your suggested changes) is actually another internal API module.

Maybe `build-discovery` is not a good name for these two?

We can continue brainstorming the names, however, I think there is a lot of value to actually keep moving the platformization forward. As also noted in the documented, we plan a second pass that can re-collapse some of the modules, once we actually understand and extract the pieces from the core[-api].

mlopatkin: >We can continue brainstorming the names

I think we don't need to block on this after clearing the potential API-vs-internal confusion.

mlopatkin (approved):

PR #35421, opened 2025-10-22
Improve file hasher error messages

tresat:
Updates the file hasher to provide more context in error messages, including the file path and specific IO errors encountered.

This prevents falsely reporting File Not Found errors, when the file exists and is not readable.


tresat:
The "as it does not exist" message had me scratching my head when I saw it during other testing and confirmed that the file **did in fact exist**.  I think emitting the message from the `FNFE` that explains the why is much more clear.

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

PR #35426, opened 2025-10-22
Polish `DefaultServiceRegistry` implementation assertion

bamboo:
- Replace `HashSet` by direct comparison
- Move test and throw closer together for clarity
- Test most common condition first

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
+++ b/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
@@ -75,9 +75,8 @@ public class DefaultServiceRegistry implements CloseableServiceRegistry, Contain
         if (impl != ScopedServiceRegistry.class && impl != DefaultServiceRegistry.class) {
             throw new IllegalArgumentException(
                 String.format(
-                    "Inheriting from %s is not allowed. Use %s instead.",
-                    DefaultServiceRegistry.class.getSimpleName(),
-                    ScopedServiceRegistry.class.getSimpleName()));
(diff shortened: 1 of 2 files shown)


platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -71,10 +71,15 @@ private enum State {INIT, STARTED, CLOSED}
...
-        DefaultServiceRegistry.class,
-        ScopedServiceRegistry.class
-    )));
+    private static void assertAllowedImplementation(Class<? extends DefaultServiceRegistry> impl) {
+        if (impl != ScopedServiceRegistry.class && impl != DefaultServiceRegistry.class) {
+            throw new IllegalArgumentException(
+                String.format(
+                    "Inheriting from %s is not allowed. Use %s instead.",
+                    DefaultServiceRegistry.class.getSimpleName(),
+                    ScopedServiceRegistry.class.getSimpleName()));
```
alllex: ❌  It was actually a typo in my PR. This should be `ServiceRegistryBuilder` instead. Because we don't want developers to instantiate registries directly.

alllex (approved):

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
bamboo: Done. Notice the class is not visible from that site though.

PR #35428, opened 2025-10-22
Support Non-Class-Based Testing

tresat:
Enables the use of JUnit Test Engines that define tests without using JVM class files.

### DIFF
--- a/platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/IncorrectSetupNonClassBasedTestingIntegrationTest.groovy
+++ b/platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/IncorrectSetupNonClassBasedTestingIntegrationTest.groovy
@@ -25,7 +25,7 @@ class IncorrectSetupNonClassBasedTestingIntegrationTest extends AbstractNonClass
         return [TestEngines.BASIC_RESOURCE_BASED]
     }
 
-    def "empty test definitions directory skips"() {
+    def "all test definitions dirs are non-existent skips"() {
         given:
         buildFile << """
             plugins {
@@ -52,7 +52,7 @@ class IncorrectSetupNonClassBasedTestingIntegrationTest extends AbstractNonClass
         testTaskWasSkippedDueToNoSources()
     }
 
-    def "non-existent test definitions directory fails"() {
+    def "some test definitions dirs are non-existent warns and testing proceeds with other dirs"() {
         def badPath = "src/test/i-dont-exist"
 
         given:
@@ -75,20 +75,85 @@ class IncorrectSetupNonClassBasedTestingIntegrationTest extends AbstractNonClass
             }
         """
 
-        // Write some test def to default dir (which is still scanned), not badPath, needed to avoid "no sources" skip
-        def defaultTestDefsDir = file(DEFAULT_DEFINITIONS_LOCATION)
-        defaultTestDefsDir.mkdirs()
-        defaultTestDefsDir.file("SomeTestDefinition.xml").createNewFile()
+        writeTestDefinitions(DEFAULT_DEFINITIONS_LOCATION)
 
         when:
-        fails("test")
+        succeeds("test", "--info")
 
         then:
-        failureCauseContains("Test definitions directory does not exist: " + testDirectory.file(badPath).absolutePath)
+        outputContains("Test definitions directory does not exist: " + testDirectory.file(badPath).absolutePath)
+        nonClassBasedTestsExecuted()
     }
 
-    def "non-directory test definitions directory fails"() {
-        def badPath = "src/test/i-dont-exist.txt"
+    def "all test definitions dirs are empty skips"() {
+        def otherPath = "src/test/other-defs"
+
+        given:
+        buildFile << """
+            plugins {
+                id 'java-library'
+            }
+
+            ${mavenCentralRepository()}
+
+            testing.suites.test {
+                ${enableEngineForSuite()}
+
+                targets.all {
+                    testTask.configure {
+                        testDefinitionDirs.from("$DEFAULT_DEFINITIONS_LOCATION")
+                        testDefinitionDirs.from("$otherPath")
+                    }
+                }
+            }
+        """
+
+        [otherPath, DEFAULT_DEFINITIONS_LOCATION].each { path ->
+            file(path).createDir()
+        }
+
+        when:
+        succeeds("test", "--info")
+
+        then:
+        testTaskWasSkippedDueToNoSources()
+    }
+
+    def "some test definitions dirs are empty proceeds silently with other dirs"() {
+        def emptyPath = "src/test/other-defs"
+
+        given:
+        buildFile << """
+            plugins {
+                id 'java-library'
+            }
+
+            ${mavenCentralRepository()}
+
+            testing.suites.test {
+                ${enableEngineForSuite()}
+
+                targets.all {
+                    testTask.configure {
+                        testDefinitionDirs.from("$DEFAULT_DEFINITIONS_LOCATION")
+                        testDefinitionDirs.from("$emptyPath")
+                    }
+                }
+            }
+        """
+
+        writeTestDefinitions()
+        file(emptyPath).createDir()
+
+        when:
+        succeeds("test", "--info")
+
+        then:
+        nonClassBasedTestsExecuted()
+    }
+
+    def "all test definitions dirs are not directories proceeds and fails with no tests found"() {
+        def badPath = "src/test/im-a-file.txt"
         file(badPath).createFile()
 
         given:
@@ -114,7 +179,41 @@ class IncorrectSetupNonClassBasedTestingIntegrationTest extends AbstractNonClass
         fails("test")
 
         then:
-        failureCauseContains("Test definitions directory is not a directory: " + testDirectory.file(badPath).absolutePath)
+        sourcesPresentAndNoTestsFound()
+    }
+
+    def "some test definitions dirs are not directories warns and testing proceeds with other dirs"() {
+        def badPath = "src/test/im-a-file.txt"
+        file(badPath).createFile()
+
+        given:
+        buildFile << """
+            plugins {
+                id 'java-library'
+            }
+
+            ${mavenCentralRepository()}
+
+            testing.suites.test {
+                ${enableEngineForSuite()}
+
+                targets.all {
+                    testTask.configure {
+                        testDefinitionDirs.from("$DEFAULT_DEFINITIONS_LOCATION")
+                        testDefinitionDirs.from("$badPath")
+                    }
+                }
+            }
+        """
+
+        writeTestDefinitions()
+
+        when:
+        succeeds("test", "--info")
+
+        then:
+        outputContains("Test definitions directory is not a directory: " + testDirectory.file(badPath).absolutePath)
+        nonClassBasedTestsExecuted()
     }
 
     def "missing test classes and/or definitions is skipped or fails when appropriate (scan for test classes = #scanForTestClasses, has test classes = #hasTestClasses, add test defs dir = #addTestDefsDir, has test defs = #hasTestDefs)"() {
--- a/platforms/jvm/testing-jvm/src/main/java/org/gradle/api/internal/tasks/testing/detection/DefaultTestScanner.java
+++ b/platforms/jvm/testing-jvm/src/main/java/org/gradle/api/internal/tasks/testing/detection/DefaultTestScanner.java
@@ -16,7 +16,6 @@
 
 package org.gradle.api.internal.tasks.testing.detection;
 
-import org.gradle.api.GradleException;
 import org.gradle.api.file.EmptyFileVisitor;
 import org.gradle.api.file.FileTree;
 import org.gradle.api.file.FileVisitDetails;
@@ -26,6 +25,8 @@ import org.gradle.api.internal.tasks.testing.ClassTestDefinition;
 import org.gradle.api.internal.tasks.testing.DirectoryBasedTestDefinition;
 import org.gradle.api.internal.tasks.testing.TestClassProcessor;
 import org.gradle.api.internal.tasks.testing.TestDefinition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.util.Set;
@@ -36,6 +37,8 @@ import java.util.regex.Pattern;
  * a detection or filename scan is performed to find test classes.
  */
 public class DefaultTestScanner implements TestDetector {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultTestScanner.class);
+
     private static final Pattern ANONYMOUS_CLASS_NAME = Pattern.compile(".*\\$\\d+");
     private final FileTree candidateClassFiles;
     private final Set<File> candidateDefinitionDirs;
@@ -80,15 +83,18 @@ public class DefaultTestScanner implements TestDetector {
             }
         });
         candidateDefinitionDirs.forEach(dir -> {
-            if (dir.exists()) {
-                if (dir.isDirectory()) {
-                    TestDefinition testDefinition = new DirectoryBasedTestDefinition(dir);
-                    testClassProcessor.processTestDefinition(testDefinition);
-                } else {
-                    throw new GradleException("Test definitions directory is not a directory: " + dir.getAbsolutePath());
-                }
+            boolean isValid = false;
+            if (!dir.exists()) {
+                LOGGER.warn("Test definitions directory does not exist: " + dir.getAbsolutePath());
+            } else if (!dir.isDirectory()) {
+                LOGGER.warn("Test definitions directory is not a directory: " + dir.getAbsolutePath());
             } else {
-                throw new GradleException("Test definitions directory does not exist: " + dir.getAbsolutePath());
+                isValid = true;
+            }
+
+            if (isValid) {
+                TestDefinition testDefinition = new DirectoryBasedTestDefinition(dir);
+                testClassProcessor.processTestDefinition(testDefinition);
             }
         });
     }
--- a/platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/shared/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/ResourceBasedSelectorResolver.java
+++ b/platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/shared/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/ResourceBasedSelectorResolver.java
@@ -70,9 +70,12 @@ public class ResourceBasedSelectorResolver implements SelectorResolver {
                     .map(Optional::get)
                     .map(Match::exact)
                     .collect(Collectors.toSet());
-            return Resolution.matches(tests);
-        } else {
-            return Resolution.unresolved();
+
+            if (!tests.isEmpty()) {
+                return Resolution.matches(tests);
+            }
         }
+
+        return Resolution.unresolved();
     }
 }
--- a/platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/multi-file-rbt-engine/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/MultiFileResourceBasedSelectorResolver.java
+++ b/platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/multi-file-rbt-engine/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/MultiFileResourceBasedSelectorResolver.java
@@ -61,10 +61,12 @@ public class MultiFileResourceBasedSelectorResolver implements SelectorResolver 
                 .map(Optional::get)
                 .map(Match::exact)
                 .forEach(tests::add);
-            return Resolution.matches(tests);
-        } else {
-            return Resolution.unresolved();
+
+            if (!tests.isEmpty()) {
+                return Resolution.matches(tests);
+            }
         }
+        return Resolution.unresolved();
     }
 
     private boolean hasTestHalves(File dir) {


tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DirectoryBasedTestDefinition.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+@NullMarked
+public final class DirectoryBasedTestDefinition implements TestDefinition {
+    private final File resourceFile;
+
+    public DirectoryBasedTestDefinition(File resourceFile) {
+        this.resourceFile = resourceFile;
+    }
+
+    public File getTestDefintionFile() {
```
octylFractal: Typo.

```suggestion
    public File getTestDefinitionFile() {
```

platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestClassProcessor.java
```diff
@@ -131,6 +160,44 @@ private void processAllTestClasses() {
                 launcher.execute(discoveryRequest, executionListener);
             }
         }
+
```
octylFractal, line 177: Moving these methods seems like it just adds noise to the diff, is it necessary? Or if we are going to move them here, can we try to make `CollectThenExecuteTestDefinitionConsumer` `static` so some of these can be as well?

octylFractal: Also I don't see why `addEnginesFilter`, etc. weren't also moved.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
```diff
@@ -659,6 +662,10 @@ public ModularitySpec getModularity() {
      */
     @Override
     protected JvmTestExecutionSpec createTestExecutionSpec() {
+        if (!getTestFramework().supportsNonClassBasedTesting() && getScanForTestDefinitions().get()) {
```
octylFractal: I think this should also fail if `getCandidateDefinitionDirs()` or `getTestDefinitionDirs()` has anything in it.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
```diff
@@ -1234,6 +1261,19 @@ public FileTree getCandidateClassFiles() {
         return getTestClassesDirs().getAsFileTree().matching(patternSet);
     }
 
+    /**
+     * Returns the classes files to scan for test classes.
+     *
+     * @return The candidate class files.
```
octylFractal: :x: This still refers to classes instead of dirs.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
octylFractal: Actually, is there any reason to have `getScanForTestDefinitions()` at all? Shouldn't this just always be active, and you can unwire the test definitions if you don't want them?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/worker/TestEventSerializer.java
```diff
@@ -109,16 +112,28 @@ public void write(Encoder encoder, CompositeIdGenerator.CompositeId value) throw
...
+        @Override
+        public DirectoryBasedTestDefinition read(Decoder decoder) throws Exception {
+            return new DirectoryBasedTestDefinition(new File(decoder.readString()));
         }
 
         @Override
-        public void write(Encoder encoder, DefaultTestClassRunInfo value) throws Exception {
-            encoder.writeString(value.getTestClassName());
+        public void write(Encoder encoder, DirectoryBasedTestDefinition value) throws Exception {
+            encoder.writeString(value.getTestDefintionFile().getAbsolutePath());
```
octylFractal: 🤔 This makes the resulting definition different after serialization. I think it might be better to call `getAbsoluteFile()` in the `DirectoryBasedTestDefinition` constructor.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/ClassTestDefinition.java
```diff
@@ -16,23 +16,49 @@
...
 
 import org.apache.commons.lang3.StringUtils;
+import org.gradle.api.internal.tasks.testing.filter.TestSelectionMatcher;
+import org.jspecify.annotations.NullMarked;
 
-public class DefaultTestClassRunInfo implements TestClassRunInfo {
+/**
+ * A test definition which denotes a single test class.
+ *
+ * @implNote Kept non-{@code final} to enable mocking.
```
octylFractal: :x: I only found one mock usage (`SuiteTestClassProcessorTest`), let's just replace it with `new ClassTestDefinition("<class-name>")` instead of letting it affect production code design.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestDefinition.java
```diff
@@ -0,0 +1,44 @@
...
+import org.gradle.api.Describable;
+import org.gradle.api.internal.tasks.testing.filter.TestSelectionMatcher;
+
+/**
+ * Represents a container of tests to be executed, for instance, a test class or a directory
+ * containing non-class-based tests.
+ */
+public interface TestDefinition extends Describable {
+    /**
+     * A stable identifier for this test class run info. Used for tracking test classes across test runs.
```
octylFractal: :x: Javadoc needs to be updated to call it a definition.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/DefaultRootTestEventReporter.java
octylFractal, line 1: :x: Remove the whitespace-only change so the file is unchanged.

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
octylFractal, line 1: :x: Remove the whitespace-only change so the file is unchanged.

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/testengine/descriptor/ResourceBasedFileDescriptor.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.gradle.testing.testengine.descriptor;
+
+import org.junit.platform.engine.UniqueId;
+import org.junit.platform.engine.support.descriptor.AbstractTestDescriptor;
+
+import java.io.File;
+
+public class ResourceBasedFileDescriptor extends AbstractTestDescriptor {
```
octylFractal: :x: Appears unused, delete?

platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestClassProcessor.java
tresat: I took the first step towards making this consumer static, but there's so much non-static access, I didn't want to completely rework it yet.

But thinking about it again, I think it's worth the cost of copying references to all those fields to make this type static, as it make the processor itself much simpler.  I'll update this.

tresat:
@bot-gradle test this

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/worker/TestEventSerializer.java
tresat: I'm not sure this matters, but I'm fine with calling `getAbsoluteFile()`.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
tresat: `getScanForTestDefinitions()` is the master on switch that allows opting into non-class-based testing.  Having this as a separate property allows setting a convention for resource-based test definitions in `src/test/definitions`, which will help build authors know where to put their definitions and keep builds using this feature more similar to each other.  So I'm in favor of keeping it working this way.

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
octylFractal: Can we then add tests for JUnit 4 and TestNG to verify that if dirs are provided but no (test) classes, it fails with the "no tests run" message?

tresat: I'm not quite sure what you're suggesting.

There is already a test `"can't do resource-based testing with unsupported test framework = #testFrameworkName"` that ensures that we short-circuit and fail with `"The $testFrameworkName test framework does not support resource-based testing."` if we enable scanning for test definitions with either of those frameworks.  Having dirs or not won't change the behavior here (in fact this test already writes test defs to the conventional location).

There's also existing tests for the case of empty test classes dirs that fail with "no tests run".

Is this a different scenario?

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
Testing is still being added and adjusted (see https://github.com/gradle/gradle/pull/35441), but major work here is hopefully done now.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/internal/tasks/testing/testng/TestNGTestFramework.java
```diff
@@ -114,4 +114,8 @@ public void close() throws IOException {
         detector = null;
     }
 
+    @Override
+    public String getDisplayName() {
+        return "Test NG";
```
big-guy: I think they use "TestNG"

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
```diff
@@ -203,6 +202,10 @@ public Object call() {
         javaLauncher.finalizeValueOnRead();
         getDryRun().convention(false);
         testFramework = objectFactory.property(TestFramework.class).convention(objectFactory.newInstance(JUnitTestFramework.class, this.getFilter(), this.getTemporaryDirFactory(), this.getDryRun()));
+
+        getTestDefinitionDirs().from(getProject().getLayout().getProjectDirectory().file(TEST_DEFINITIONS_DIR));
```
big-guy: This needs to be defined outside the task.  We shouldn't be calling `getProject` in a task.

I think this means by default we do not specify any test definition dir and it has to be manually added for now.

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/shared/test-engine-build/src/main/java/org/gradle/testing/testengine/util/DirectoryScanner.java
```diff
@@ -0,0 +1,51 @@
...
+                        filesFound.add(file);
+                    }
+                    filesFound.addAll(scanDirectory(file));
+                } else if (testFileParser.isValidTestDefinitionFile(file)) {
+                    filesFound.add(file);
+                }
+            }
+        }
+        return filesFound;
+    }
```
leonard84, line 50: Why don't you use a `FileWalker`?

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/shared/test-engine-build/src/main/java/org/gradle/testing/testengine/util/TestFileParser.java
```diff
@@ -0,0 +1,67 @@
...
+     * @throws RuntimeException if there is an error parsing the XML file.
+     */
+    public List<String> parseTestNames(File testDefinitionsFile) {
+        List<String> names = new ArrayList<>();
+
+        try {
+            DocumentBuilderFactory dbFactory = DocumentBuilderFactory.newInstance();
+            DocumentBuilder dBuilder = dbFactory.newDocumentBuilder();
+            Document doc = dBuilder.parse(testDefinitionsFile);
+            doc.getDocumentElement().normalize();
```
leonard84, line 52: FYI, this is probably vulnerable to common XML vulnerabilities. Which might be ok if this is only test code.

| Vulnerability             | Default Risk | Safe Mitigation |
|----------------------------|---------------|-----------------|
| **XML External Entity (XXE) Injection** | ✅ Enabled | Disable DOCTYPE declarations and external entities:<br>`dbf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);`<br>`dbf.setFeature("http://xml.org/sax/features/external-general-entities", false);` |
| **Billion Laughs (Entity Expansion)** | ✅ Enabled | Disable DOCTYPE declarations (prevents entity expansion). |
| **External DTD Fetching** | ✅ Enabled | Disable loading external DTDs:<br>`dbf.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);` |
| **XInclude Injection** | ✅ Enabled | Disable XInclude:<br>`dbf.setXIncludeAware(false);` |
| **Schema / XPath Injection** | ⚠️ Possible | Validate safely and avoid untrusted schemas. |

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestClassProcessor.java
```diff
@@ -20,8 +20,10 @@
 
 /**
  * A processor for executing tests. Implementations are not required to be thread-safe.
+ *
+ * @param <D> the type of test definition supported by this processor
  */
-public interface TestClassProcessor extends Stoppable {
+public interface TestClassProcessor<D extends TestDefinition> extends Stoppable {
```
leonard84, line 26: Is this a backwards-compatible change?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestClassProcessor.java
```diff
@@ -30,12 +32,12 @@ public interface TestClassProcessor extends Stoppable {
...
     /**
-     * Accepts the given test class for processing. May execute synchronously, asynchronously, or defer execution for
+     * Accepts the given test definition for processing. May execute synchronously, asynchronously, or defer execution for
      * later.
      *
-     * @param testClass The test class.
+     * @param testDefinition The test definition.
      */
-    void processTestClass(TestClassRunInfo testClass);
+    void processTestDefinition(D testDefinition);
```
leonard84, line 40: This seems like a hard brake. Wouldn't it be better to create a new interface and to rework this to adapt to the new one?

tresat:
> In general, this looks like it won't be compatible with plugins built against the older API.

This is internal API, so plugins shouldn't be using it.  I didn't see any notes on this type saying that it needed to be preserved.  Do you know of any specific plugins that are using it?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestClassProcessor.java
tresat: No, but it's internal API.

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/shared/test-engine-build/src/main/java/org/gradle/testing/testengine/util/TestFileParser.java
tresat: Everything here `org.gradlel.testing.testengine` is strictly a sample TestEngine used for internal testing.  It won't be shipped, and the only XML it parses is stuff I wrote.

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/shared/test-engine-build/src/main/java/org/gradle/testing/testengine/util/DirectoryScanner.java
tresat: I was just getting something up and running quickly for testing.  I could update, but there's little benefit.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/internal/tasks/testing/detection/DefaultTestScanner.java
```diff
@@ -67,8 +75,24 @@ private void filenameScan() {
...
             @Override
             public void visitClassFile(FileVisitDetails fileDetails) {
-                TestClassRunInfo testClass = new DefaultTestClassRunInfo(getClassName(fileDetails));
-                testClassProcessor.processTestClass(testClass);
+                TestDefinition testDefinition = new ClassTestDefinition(getClassName(fileDetails));
+                testClassProcessor.processTestDefinition(testDefinition);
+            }
+        });
+        candidateDefinitionDirs.forEach(dir -> {
+            if (dir.exists()) {
```
big-guy: This check might be unnecessary depending on how we get the set of files from the file collection.  e.g., .getFiles() might already filter for existence

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/testengine/util/Inputs.java
```diff
@@ -13,8 +13,18 @@
...
-    String getTestClassName();
+package org.gradle.testing.testengine.util;
+
+import java.io.File;
+
+public interface Inputs {
+    String PROPERTY_PREFIX = "org.gradle.testing.fixture.testengine.";
+
+    String TEST_RESOURCES_ROOT_DIR_PROP = PROPERTY_PREFIX + "testResourcesRootDir";
+
```
big-guy: Is this needed if we're providing the directory to search through discovery?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestDefinition.java
```diff
@@ -0,0 +1,44 @@
...
+     */
+    String getId();
+
+    /**
+     * Should this test definition be processed according to the given matcher.
+     *
+     * @param matcher the matcher to check
+     * @return {@code true} if so; {@code false} otherwise
+     */
+    boolean matches(TestSelectionMatcher matcher);
```
big-guy, line 43: I'm not sure how this is going to be used. It doesn't look like it's needed right now.

I don't think we need unique ids for the definition and we do not need to do selection on top of the definitions.  The test selection is going to happen down in the discovery process and not at the definition level.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
```diff
@@ -1149,6 +1167,15 @@ public void setScanForTestClasses(boolean scanForTestClasses) {
...
     }
 
+    /**
+     * Whether to scan for test resources for resource-based testing.
+     *
+     * @since 9.3.0
+     */
+    @Incubating
+    @Input
+    public abstract Property<Boolean> getScanForTestDefinitions();
```
big-guy: I don't think we need this at all.  We should just always send test definition dirs over

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
```diff
@@ -1234,6 +1261,19 @@ public FileTree getCandidateClassFiles() {
...
+     *
+     * @return The candidate directories holding non-class-based test definition files.
+     * @since 9.3.0
+     */
+    @Incubating
+    @InputFiles
+    @SkipWhenEmpty
+    @IgnoreEmptyDirectories
+    @PathSensitive(PathSensitivity.RELATIVE)
+    public abstract ConfigurableFileCollection getCandidateDefinitionDirs();
```
big-guy: I don't think this is a pattern we need to follow for the new definitions.

The reason we have this for classes is that we encouraged or supported a pattern of defining a single test source set and then using multiple Test tasks to only run certain classes out of the compiled set of classes.

IOW, if you had:
build/classes/test/A.class
build/classes/test/B.class
build/classes/test/C.class

You could have a A-Test, B-Test and C-Test task that only runs their respective classes while pointing to the same classes dir (build/classes/test/).  Changes to A.class would not impact the up-to-date checks for B-Test. 

This is a bit tied up into how we want to do filtering, but I wouldn't start with this pattern for now.  Just a single property is fine.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
```diff
@@ -1234,6 +1261,19 @@ public FileTree getCandidateClassFiles() {
...
 
+    /**
+     * Returns directories to scan for non-class-based test definition files.
+     *
+     * @return The candidate directories holding non-class-based test definition files.
+     * @since 9.3.0
+     */
+    @Incubating
+    @InputFiles
+    @SkipWhenEmpty
```
big-guy: I'm not sure what happens when a task has two properties with `SkipWhenEmpty`.  Do we check both of them and only skip if both are empty?

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
big-guy: Alternatively, we could just push this down to the test class processors.  If they receive a test definition they can't process, they throw/fail in a way that shows up in reports?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+ * Tests that exercise and demonstrate Non-Class-Based Testing using the {@code Test} task
+ * and a sample resource-based JUnit Platform Test Engine defined in this project's {@code testFixtures}.
+ */
+class NonClassBasedTestingIntegrationTest extends AbstractIntegrationSpec {
+    private static engineJarLibPath
+
+    def setupSpec() {
+        def version = IntegrationTestBuildContext.INSTANCE.getVersion().getBaseVersion().version
+        // TODO: there's probably a better place to put this and/or way to get this on the path
+        def path = IntegrationTestBuildContext.TEST_DIR.file("../../software/testing-base/build/libs/gradle-testing-base-$version-test-fixtures.jar").path
```
big-guy: This can't be merged as-is, but another alternative I didn't think of:

We publish local jars to `IntegrationTestBuildContext.localRepository`.  We can probably reuse that to publish this test engine. I'm not entirely sure how that's wired today because we only use it for TAPI.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+    def "non-readable test definitions directory fails"() {
+        def badPath = "src/test/i-cant-be-read"
+        // create the directory and make it non-readable at the filesystem level
+        def dir = file(badPath).createDir()
+
+        // try to remove read permission for everyone (owner=false, ownerOnly=false)
+        // fall back to single-arg version if the two-arg form isn't supported/has no effect
+        if (!dir.setReadable(false, false)) {
+            dir.setReadable(false)
+        }
```
big-guy: Maybe assert that the directory isn't readable for a sanity check at this point?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
big-guy: The most straightforward thing to do is still to just copy the test engine source into the build-under-test and compile it every time.

Looking at the integration tests below, many of them don't actually require the test engine. e.g., `empty test definitions directory skips` never needs the test engine.  We can probably move those into a separate integ test entirely.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/internal/tasks/testing/detection/DefaultTestScanner.java
```diff
@@ -67,8 +75,24 @@ private void filenameScan() {
...
+                if (dir.isDirectory()) {
+                    if (dir.canRead()) {
```
big-guy: Looking at the tests, this cannot ever be hit anyways because it would imply that we have non-readable inputs.  The task would never be started

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+                }
+            }
+        """
+
+        when:
+        fails("test")
+
+        then:
+        failureCauseContains("Cannot access input property 'candidateDefinitionDirs' of task ':test'. Accessing unreadable inputs or outputs is not supported.")
+        failureCauseContains("java.nio.file.AccessDeniedException: ${dir.absolutePath}")
```
big-guy: It looks like this isn't really testing something new, but testing what we do for any unreadable task input. We probably don't need this test and can assume that task execution works like normal

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+
+        cleanup:
+        // restore read permission for cleanup
+        dir.setReadable(true, false)
+    }
+
+    def "non-readable files in test definitions directory fails"() {
```
big-guy: Same thing here as above, this is testing the execution engine and not something specific to Test

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+    private static engineJarLibPath
+
+    def setupSpec() {
+        def version = IntegrationTestBuildContext.INSTANCE.getVersion().getBaseVersion().version
+        // TODO: there's probably a better place to put this and/or way to get this on the path
+        def path = IntegrationTestBuildContext.TEST_DIR.file("../../software/testing-base/build/libs/gradle-testing-base-$version-test-fixtures.jar").path
+        engineJarLibPath = TextUtil.convertLineSeparatorsToUnix(path)
+    }
+
+    def "empty test definitions directory skips"() {
```
big-guy: We should check the combinations when both resources and classes are defined:
- Classes exist and definitions do not (not skipped)
- Classes do not exist and definitions do (not skipped)
- Classes and definitions do not exist (skipped with NO SOURCE)
- Classes and definitions exist (not skipped)

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+            failureCauseContains("Cannot access input property 'candidateDefinitionDirs' of task ':test'. Accessing unreadable inputs or outputs is not supported.")
+            failureCauseContains("Failed to create MD5 hash for file: ${badFile.absolutePath} (Permission denied)")
+        }
+
+        cleanup:
+        // restore read permission for cleanup
+        badFile.setReadable(true, false)
+    }
+
+    def "resource-based test engine detects and executes test definitions (excluding jupiter engine = #excludingJupiter)"() {
```
big-guy: What's the significance of excluding Jupiter?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+    }
+
+    private void classBasedTestsExecuted() {
+        outputContains("Tested!")
+    }
+
+    private void nonClassBasedTestsExecuted() {
+        outputContains("INFO: Executing test: Test [file=SomeTestSpec.rbt, name=foo]")
+        outputContains("INFO: Executing test: Test [file=SomeTestSpec.rbt, name=bar]")
+        outputContains("INFO: Executing test: Test [file=subSomeOtherTestSpec.rbt, name=other]")
```
big-guy: This should rely on our test execution fixtures that verifies the report/output files

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+
+            ${mavenCentralRepository()}
+
+            testing.suites.test {
+                ${setupSuiteWithEngineFixture()}
+
+                targets.all {
+                    testTask.configure {
+                        scanForTestDefinitions = true
+                        testDefinitionDirs.setFrom(project.layout.projectDirectory.file("$customLocation"))
```
big-guy: did you want to purposefully exclude the default here?

Alternatively, you could just use `testDefinitionDirs.from("$customLocation")`

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+        writeTestDefinitions(customLocation)
+
+        when:
+        succeeds("test", "--info")
+
+        then:
+        nonClassBasedTestsExecuted()
+    }
+
+    def "empty custom test definitions location skips"() {
```
big-guy: I think this is a duplicate test if we no longer have a default test definition dir.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+        """
+
+        when:
+        succeeds("test", "--info")
+
+        then:
+        testTaskWasSkippedDueToNoSources()
+    }
+
+    def "can't do resource-based testing with unsupported test framework = #testFrameworkName"() {
```
big-guy: This is another test that doesn't need the test engine

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+                id 'java-library'
+            }
+
+            ${mavenCentralRepository()}
+
+            testing.suites.test {
+                $testFrameworkMethod
+
+                dependencies {
+                    implementation 'org.testng:testng:${TestNGCoverage.NEWEST}'
```
big-guy: This shouldn't be necessary with test suites

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,498 @@
...
+        then:
+        failure.assertHasCause("The $testFrameworkName test framework does not support resource-based testing.")
+
+        where:
+        testFrameworkName | testFrameworkMethod
+        "Test NG"         | "useTestNG()"
+        "JUnit"           | "useJUnit()"
+    }
+
+    def "missing test classes and/or definitions is skipped or fails when appropriate (scan for test classes = #scanForTestClasses, has test classes = #hasTestClasses, scan for test defs = #scanForTestDefs, has test defs = #hasTestDefs )"() {
```
big-guy: I think this is overcomplicating the cases we need to check.  We don't care about `scanForTestClasses` and I don't think we should have a `scanForTestDefinitions`. 

We just need to demonstrate:
- Test task with both definitions and classes can execute and both are seen in the results
- Test task with neither definitions nor classes are skipped
- Test task with classes but no definitions is executed 
- Test task with no classes but definitions is executed (this is probably covered above in a different test)

By "no definition", I mean that we set the test definition dirs for the Test, but we do not write any files into that path. 

I don't think we have a way to get into the failing case you see below because we do not have a way of filtering test definitions.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
tresat: Lots of testing updates are in progress [here](https://github.com/gradle/gradle/pull/35441).  I think it should address these concerns.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
tresat: It looks like other tasks inject `ProjectLayout` directly.  I'll switch to an internal property and avoid accessing the `project` at construction time.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/internal/tasks/testing/detection/DefaultTestScanner.java
tresat: The `"non-existent test definitions directory fails"` test suggests it's still needed to provide a cleaner error message.  

Without this check, we get "Test definitions directory is not a directory".  Without _that_ check, we get "Cannot read test definitions directory:".  Without _that other_ check, we get only "Could not execute tests in directory".  So I think this is worth leaving, for the better UX.

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/testengine/util/Inputs.java
tresat: No, this is just something I forgot to delete when I was copying the test engine.  I'll remove it in the testing branch.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestDefinition.java
tresat: This is a refactoring of existing functionality that used to be performed directly in `PatternMatchTestClassProcessor`.  I think it's cleaner this way, even if nothing furthur is needed for non-class-based testing.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
tresat: I disagree here, I think we should mirror the existing scan for test classes functionality, where if you don't scan you have to provide a `--tests` arg to specify exactly what to run.

I also like that this functions as a master on switch, where a build isn't affected by the new functionality unless you opt-in.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
tresat: OK, I wasn't sure about this, but figured I'd copy what was there before.  I'll consolidate the props.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
tresat: Yes, I hit this case during development.  This is tested (indirectly) by the existing tests.

tresat:
> Alternatively, we could just push this down to the test class processors. If they receive a test definition they can't process, they throw/fail in a way that shows up in reports?

@big-guy - We could do this, but I think it's better to fail-faster in the `Test` task itself - it already has enough info to know this won't succeed, so why go any further?

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/internal/tasks/testing/detection/DefaultTestScanner.java
tresat: True, I've remove this check.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
tresat: OK, I'll remove these.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
tresat: Is this not all tested in `"missing test classes and/or definitions is skipped or fails when appropriate (scan for test classes = #scanForTestClasses, has test classes = #hasTestClasses, scan for test defs = #scanForTestDefs, has test defs = #hasTestDefs"`?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
tresat: To make sure that if the Jupiter engine is available (which it will be by default), it doesn't interfere with a non-class-based engine.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
tresat: It should be updated after we tackle reporting, but for now I don't want to rely on any particular reporting structure or details.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
tresat: Yes.  If I use `.from` this is adding a new dir to the existing default dir.  Then the test fails since the default dir doesn't exist.

I'm going to add a new test now for default + custom dirs can be used.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
tresat: I think it's important to keep a conventional definitions dir, to promote similarity across builds.  It addresses a constant community concern.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
tresat: In the testing branch, I've moved these to an `IncorrectSetupNonClassBasedTestingIntegrationTest` test class and removed the engine dep from them.  I'll merge this to the main branch next.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
tresat: I'd strongly prefer to leave this test, assuming we leave `scanForTestDefinitions`, which I also think we should leave.  It clearly demonstrates the behavior for each interesting combination in the table, and I've found it very useful.

subprojects/core/src/test/groovy/org/gradle/configuration/DefaultImportsReaderTest.groovy
```diff
@@ -23,7 +23,7 @@ import spock.lang.Specification
 
 class DefaultImportsReaderTest extends Specification {
     @Rule
-    public Resources resources = new Resources()
+    public Resources resources = new Resources(null)
```
big-guy: Is this being used at all? It seems like this might be copying nothing?

big-guy: Maybe to make this explicit, we add a constructor to Resources with no arguments so we don't have this strange looking null copied elsewhere.

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/resource-and-class-engine/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/ClassAndResourceBasedTestEngine.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutorService;
+import org.junit.platform.engine.support.hierarchical.Node;
+import org.junit.platform.engine.support.hierarchical.ResourceLock;
+import org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService;
+import org.junit.platform.engine.support.hierarchical.ThrowableCollector;
+
+import java.io.File;
+import java.util.stream.Collectors;
+
+public class ClassAndResourceBasedTestEngine implements TestEngine {
```
big-guy, line 45: Do you think this would expose some different behavior on the Gradle side?

There are two scenarios I was thinking of:
1. The Test task is configured with only a single test engine. It's either resource based or class based.
2. The Test task is configured with multiple test engines. One is class based and one is resource based.

This adds a third option:
- The Test task is configured with a single test engine that can run both resource and class based tests.

Do you think this is different from Gradle's POV as (2)?  Do we need to test it separately?

I think for something like cucumber, are they actually running the test classes (annotated with something like `RunWith(...)`) through the cucumber test engine _and_ running resource based files? Or are they just running JUnit Jupiter + a separate test engine for resources?

platforms/jvm/testing-jvm/src/testFixtures/groovy/testengines/TestEnginesFixture.groovy
```diff
@@ -0,0 +1,103 @@
...
+        // Switch to engine build directory for this setup
+        GradleDistribution distribution = new UnderDevelopmentGradleDistribution(IntegrationTestBuildContext.INSTANCE)
+        GradleExecuter engineBuilder = new GradleContextualExecuter(distribution, customTestDirectoryProvider, IntegrationTestBuildContext.INSTANCE)
+        engineBuildDir = customTestDirectoryProvider.testDirectory.file(ENGINE_COPY_TO_DIR_NAME)
+        engineBuilder.inDirectory(engineBuildDir)
+            .withRepositoryMirrors()
+            .withConsole(ConsoleOutput.Verbose)
+
+        // Build the test engine jar
+        engineBuilder.withTasks("build").run()
```
big-guy: I think `ArtifactBuilder` does something like this already

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
big-guy: No, we should not be setting this property at all here.  This is entirely a plugin's decision to set this property.  It doesn't make sense that every Test task is going to be looking in `src/test/definitions`.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestDefinition.java
big-guy: Ah, I see where the TestSelectionMatcher comes from, but this is in the wrong place.

We don't have ids for TestClassRunInfo now (the existing API that ClassTestDefinitions came from) and the matcher only applies to the class files. I don't think that's going to extend to test definitions in general.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
big-guy: I think the switch is based on if we have anything defined in the CFC.  If we do, then we're going to send them to JUnit Platform for discovery. If we don't, then we don't have anything to pass. 

Not now, but our goal should be to get rid of the flag for classes too. AFAICT, this has been enabled since before Gradle 1.0 and the justification for it is pretty weak (I think it can be faster to start executing tests).

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
tresat: Discussed, going to leave as is.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
tresat: Discussed, resolved.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
tresat: Discussed, resolved.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestDefinition.java
tresat: Discussed, resolved.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
tresat: Deleted.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/NonClassBasedTestingIntegrationTest.groovy
tresat: Adjusted to deal with removal of scan for test defs, as discussed.

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/shared/test-engine-build/src/main/java/org/gradle/testing/testengine/util/DirectoryScanner.java
tresat: I looked at using Files.walk and Files.walkFileTree and the code CoPilot generated looked actually more complex than this, so I'll just keep it simple for now.

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/resource-and-class-engine/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/ClassAndResourceBasedTestEngine.java
tresat: Discussed.

platforms/jvm/testing-jvm/src/testFixtures/groovy/testengines/TestEnginesFixture.groovy
tresat: Looking at ArtifactBuilder, it's about programatically building the jar file by file, not copying pre-defined source files to build.  I think I'd rather leave this separate, since the engine examples are more complex.

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/ClassAndNonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+
+                        options {
+                            excludeEngines("junit-jupiter")
+                        }
+                    }
+                }
+            }
+
+            // Ensure the definitions directory exists even if no definitions are added
+            project.layout.projectDirectory.file("$DEFAULT_DEFINITIONS_LOCATION").getAsFile().mkdirs()
```
big-guy, line 53: Why?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/AbstractNonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,75 @@
...
+    protected void testTaskWasSkippedDueToNoSources() {
+        result.assertTaskSkipped(":test")
+        outputContains("Skipping task ':test' as it has no source files and no previous output files.")
+    }
+
+    // Once reporting is addressed, this should use more robust verification using existing report-checking fixtures
+    protected void nonClassBasedTestsExecuted() {
+        outputContains("INFO: Executing resource-based test: Test [file=SomeTestSpec.rbt, name=foo]")
+        outputContains("INFO: Executing resource-based test: Test [file=SomeTestSpec.rbt, name=bar]")
+        outputContains("INFO: Executing resource-based test: Test [file=subSomeOtherTestSpec.rbt, name=other]")
```
big-guy: Do the existing fixtures not work? I would expect them to

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/CucumberNonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,115 @@
...
+                useJUnitJupiter()
+
+                dependencies {
+                    implementation("io.cucumber:cucumber-java:7.15.0")
+                    implementation("io.cucumber:cucumber-junit-platform-engine:7.15.0")
+                }
+
+                targets.all {
+                    testTask.configure {
+                        testDefinitionDirs.from(project.layout.projectDirectory.file("$testLocation"))
```
big-guy: nit, I think you could just use `from("$testLocation")` for this kind of thing.  It's automatically interpreted as relative to the project dir

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/FailsExecutionNonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,92 @@
...
+        """
+
+        writeTestDefinitions()
+
+        when:
+        fails("test")
+
+        then:
+        def results = resultsFor()
+        def testPath = results.testPath(":UnknownClass")
```
big-guy, line 63: Do we see this because the test engine fails very early in execution or would we see this for _any_ failure that happens after discovery?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/IncorrectSetupNonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,233 @@
...
+        """
+
+        when:
+        succeeds("test", "--info")
+
+        then:
+        testTaskWasSkippedDueToNoSources()
+    }
+
+    def "non-existent test definitions directory fails"() {
```
big-guy: Does this mean that we fail if testDefinitionDirs = `[src/test/definitions` and we have something in classes dir? 

I don't think this is consistent with how other tasks work and it makes `SkipWhenEmpty` make no sense.

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm/src/testFixtures/groovy/testengines/TestEnginesFixture.groovy
```diff
@@ -0,0 +1,112 @@
...
+import org.gradle.test.fixtures.file.TestNameTestDirectoryProvider
+import org.gradle.util.internal.TextUtil
+
+/**
+ * Trait to provide custom JUnit Test Engines to integration tests.
+ * <p>
+ * Note that this trait can <strong>NOT BE SHARED</strong> by multiple test classes in the same project
+ * that share a class name prefix (e.g. MyClassRedTest and MyClassBlueTest) because of the way our testing
+ * infrastructure sets up test directories.  Both of these test classes would use the same engine copy-to
+ * dir, and thus interfere with each other.
```
octylFractal: ❓ Why document this instead of just adding the full class name to the path?

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/fails-discovery-rbt-engine/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/FailsDiscoveryResourceBasedTestEngine.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public String getId() {
+        return ENGINE_ID;
+    }
+
+    @Override
+    public TestDescriptor discover(EngineDiscoveryRequest discoveryRequest, UniqueId uniqueId) {
+        LOGGER.info(() -> {
+            return "Discovering tests with engine: " + uniqueId;
+        });
```
octylFractal: 💅 Keep this short.
```suggestion
        LOGGER.info(() -> "Discovering tests with engine: " + uniqueId);
```

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/shared/test-engine-build/src/main/java/org/gradle/testing/testengine/descriptor/ResourceBasedTestDescriptor.java
```diff
@@ -0,0 +1,59 @@
...
+        return file;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public String toString() {
+        return "Test [file=" + file.getName() + ", name=" + name + "]";
```
octylFractal: 💅 The extra space here seems weird?
```suggestion
        return "Test[file=" + file.getName() + ", name=" + name + "]";
```

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/shared/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/ResourceBasedSelectorResolver.java
```diff
@@ -0,0 +1,77 @@
...
+        List<File> contents = new ArrayList<>(directoryScanner.scanDirectory(selector.getDirectory(), true));
+
+        if (!contents.isEmpty()) {
+            Set<DiscoverySelector> selectors = new LinkedHashSet<>();
+            contents.stream()
+                .filter(File::isFile)
+                .forEach(file -> selectors.add(DiscoverySelectors.selectFile(file.getAbsolutePath())));
+            contents.stream()
+                .filter(File::isDirectory)
+                .forEach(file -> selectors.add(DiscoverySelectors.selectDirectory(file.getAbsolutePath())));
```
octylFractal: 💭 This is faster and IMO clearer, unless there's a reason you need to keep the order as files first?:
```suggestion
            for (File file : contents) {
                if (file.isFile()) {
                    selectors.add(DiscoverySelectors.selectFile(file.getAbsolutePath()));
                } else if (file.isDirectory()) {
                    selectors.add(DiscoverySelectors.selectDirectory(file.getAbsolutePath()));
                }
            }
```

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/ClassTestDefinition.java
octylFractal: :x: Note is outdated.
```suggestion
```

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/DefaultRootTestEventReporter.java
octylFractal: This is still not done.

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
octylFractal: This is still not done.

platforms/jvm/testing-jvm/src/testFixtures/groovy/testengines/TestEnginesFixture.groovy
tresat: I just didn't want to make a bigger change to our testing infrastructure, and figured this might be the cause of other flakiness in our tests, so be worth resolving beyond just this fixture.  

But you make an excellent point, this fixture could just use a different nested dir per class, as the names are present.  I'm going to try this.

platforms/jvm/testing-jvm/build.gradle.kts
```diff
@@ -81,3 +82,6 @@ packageCycles {
 tasks.isolatedProjectsIntegTest {
     enabled = false
 }
+
+// We do a lot of test file extraction here, and sometimes the cleanup service doesn't finish deleting in time
```
big-guy: What is this about? Is there an open issue for it?

platforms/jvm/testing-jvm/src/testFixtures/groovy/testengines/TestEnginesFixture.groovy
tresat: Updated this, it's now better and reliable.

platforms/jvm/testing-jvm/build.gradle.kts
tresat: Removed with changes addressing https://github.com/gradle/gradle/pull/35428#discussion_r2492249809.

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/shared/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/ResourceBasedSelectorResolver.java
tresat: Speed wasn't a huge concern here, and I just default to streams for iteration.  I'll update though.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/ClassTestDefinition.java
tresat: Good catch!

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/ClassAndNonClassBasedTestingIntegrationTest.groovy
tresat: This is needed if we are testing the condition where no test defs exist, otherwise we get a "defs directory does not exist".  I'll expand the comment to make this more clear.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/AbstractNonClassBasedTestingIntegrationTest.groovy
tresat: They do work, but the report hierarchy isn't final.  I'd rather leave it like this until we finalize reporting and then switch to the new fixtures throughout these tests.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/FailsExecutionNonClassBasedTestingIntegrationTest.groovy
tresat: IDK how deeply you want to investigate this.  A `TestEngine` provides 2 callbacks: `discover()` and `execute()`.  So _any_ failure after discovery _is_ by definition during execution, unless something odd is happening.  Perhaps `getId()` could fail?

So I think the answer is "yes", but I'm not 100% on that.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/IncorrectSetupNonClassBasedTestingIntegrationTest.groovy
tresat: You're right, this is inconsistent with how `testClassesDirs` works when given a bad path, but I think it's worth the inconsistency, as this is more desirable behavior.

As a build author, I'd rather fail-fast if I tell a task to get test defs from a dir that doesn't exist.  I've probably misconfigured my build at that point.

I don't see why this makes `SkipWhenEmpty` nonsensical.  The "empty" there means "the property is empty", not "the property has a value available, which is a non-existant directory".

If you really want this, I suppose we could change the behavior to "fail only if ALL test def dirs don't exist", but I think that's slightly worse - it's silently letting probably incorrect behavior succeed.  

If a build author really wants to supply possibly non-existent dirs to this prop, they can just create the dirs in a `doFirst`.

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/FailsExecutionNonClassBasedTestingIntegrationTest.groovy
big-guy: What I meant was why do we see "UnknownClass" vs something related to the test engine or resources or something like that?

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/IncorrectSetupNonClassBasedTestingIntegrationTest.groovy
big-guy: OK, I didn't catch this earlier.  This is a bad behavior. 

Let me try to convince you:

- You apply a plugin that adds a task `:foo` that has a CFC with `@SkipWhenEmpty` and a convention of `src/`. Running `gradle build` will try to run this task.
- You do not write anything into `src/`
- You run `gradle build`

Does the build fail? No, because `src/` is empty and we skip `:foo`

- You put something into `src/`. Running `gradle build` now runs `:foo` successfully.
- You decide that you'd prefer if your sources were in `source/`. You add `foo.sources.from("source/")`
- You run `gradle build` and it fails because `source/` doesn't exist. You know how to fix this: you add some files into `source/`. You happen to keep everything else in `src/`
- Later, you decide to move everything remaining in `src/` to `source/`. You run `gradle build`, it passes because `src/` is empty but still exists. 
- On a fresh checkout, `gradle build` now fails because `src/` does not exist.

We've now seen the full gamut:
- `src/` not existing -> build works
- `src/` not existing -> build fails
- `src/` empty -> build works
- `src/` not empty -> build works

My point is that we should follow `@SkipWhenEmpty`'s behavior and treat a missing and empty directory as the same thing. 

> As a build author, I'd rather fail-fast if I tell a task to get test defs from a dir that doesn't exist. I've probably misconfigured my build at that point.

I believe the bad behavior you're trying to prevent is that someone configures the task the wrong way (e.g., instead of `src/test/definitions` they type `source/test/definitions`). Not catching this would give someone the false impression that their tests were passing.

To catch this consistently, we would need to change the way `@SkipWhenEmpty` works. Otherwise, we would only catch these problems when there are multiple directories configured.

The problem is that the vast majority of the time, missing directories are OK and expected. We can't know what was intended. This is a general problem with FileCollections and something that's come up before when combining Providers with FileCollections.

> If you really want this, I suppose we could change the behavior to "fail only if ALL test def dirs don't exist", but I think that's slightly worse - it's silently letting probably incorrect behavior succeed.

We can catch this later.  This seems like it's another form of "we looked for tests, but didn't find any".  We don't need to catch this at the input level.

>  The "empty" there means "the property is empty", not "the property has a value available, which is a non-existant directory".

It's the other way around. SkipWhenEmpty means the directory is empty, not that the property is empty. 

> If a build author really wants to supply possibly non-existent dirs to this prop, they can just create the dirs in a doFirst.

I don't think we should suggest that.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/IncorrectSetupNonClassBasedTestingIntegrationTest.groovy
```diff
@@ -0,0 +1,233 @@
...
+        true                | true              | false             | false         || false            || false
+        true                | false             | false             | false         || true             || false
+        false               | true              | true              | true          || false            || false
+        false               | false             | true              | true          || false            || false
+        false               | true              | false             | true          || false            || true
+        false               | true              | true              | false         || false            || true
+        false               | false             | false             | true          || true             || false
+        false               | false             | true              | false         || true             || false
+        false               | true              | false             | false         || false            || true
+        false               | false             | false             | false         || true             || false
```
big-guy, line 289: You know I dislike conditionals in expectations for tests :)

If I'm looking at this right, shouldFail is only true when scanForTestClasses=false and hasTestClasses=true, I think this is actually just testing existing behavior where we don't find tests when we think we should. It doesn't matter what we do with test definitions.

I also don't think we care about scanForTestClasses.  They don't impact how we select from test definitions.

Any test that has `addTestDefsDir` as false isn't exercising the test definition changes either.

Maybe part of what you're trying to do is mixed up in applying `java-library` because that comes with certain defaults. I think we only care about checking how skipped tests work with and without test definitions.

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/FailsExecutionNonClassBasedTestingIntegrationTest.groovy
tresat: It has to do with how we track hierarchy.  If the engine fails before any tests have started (as in this example engine), then we literally don't know which class is responsible.

Though in this case, it's more properly "what TestDescriptor" is responsible.  I think updating the internal constant in `JunitSupport` to read "UnknownSource" and be more general would be slightly better.

I'll add a followup item for this.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/IncorrectSetupNonClassBasedTestingIntegrationTest.groovy
tresat: I still find this table really clear and useful.  With the changes to remove properties and changes to what happens on non-existent tests, there is now some duplication here, yes.  But I'd rather be comprehensive and have this fully describe all combinations with their intended results and have it catch any regressions.

I don't think there's harm in a little redundancy here, I'd rather leave it.

tresat:
@bot-gradle test this

ljacomet:
This work was reverted for Gradle 9.3.0 and will be released in a later version.

See
* #35730

Issue #30101, opened 2024-08-06
Use permalink in the generated gradlew file


vlsi:
### Expected Behavior

The links should point to a static resource rather than to an ever-changing one.

For instance, the following links would do better than `HEAD`:
* https://github.com/gradle/gradle/blob/849ab53bb942b2541747ec46b19707c74a62a48d/platforms/jvm/plugins-application/src/main/resources/org/gradle/api/internal/plugins/unixStartScript.txt
* https://github.com/gradle/gradle/blob/v8.9.0-RC2/platforms/jvm/plugins-application/src/main/resources/org/gradle/api/internal/plugins/unixStartScript.txt

### Current Behavior (optional)

```diff
 #   (3) This script is generated from the Groovy template
-#       https://github.com/gradle/gradle/blob/HEAD/subprojects/plugins/src/main/resources/org/gradle/api/internal/plugins/unixStartScript.txt
+#       https://github.com/gradle/gradle/blob/HEAD/platforms/jvm/plugins-application/src/main/resources/org/gradle/api/internal/plugins/unixStartScript.txt
```

Both links point to `HEAD` which changes over time

### Context

I'm upgrading Gradle, and I noticed the link is not quite right.

cobexer:
This feature request is in the backlog of the relevant team and is prioritized by them.

ov7a:
We should point to a permalink with a release tag.

VinVorteX:
Can you assign me this issue

DeepanshuYadav-code:
hwy @vlsi i want to work on this issue>

can you please assgin this issue to me

Taz03:
> We should point to a permalink with a release tag.

like hard coded?

ploober:
I see a couple of approaches to this involving adding a new parameter to the template generation:
- The tagged version ("v8.13") would probably be considered the prettiest and most desirable to use in place of `HEAD`, but do we have confident knowledge or gurantee of what the tag is going to be at the time the template is rendered? What would we use for non-release builds?

- Using short or long `commitId` would probably be easiest and most stable, always being well known and pinning the link to the exact version of the file as it existed during build.

Then there is the question of how much of the link should be parameterized.
- Simply the new token to use in place of `HEAD`, like:
Example value: `commitId = "7d3acc3"`
Example template: `https://github.com/gradle/gradle/blob/${commitId}/platforms/jvm...`

- Using a more generic link base, like:
[...]

PR #35429, opened 2025-10-22
Fix: Use permalink in generated gradlew file instead of HEAD reference

Sourojitbanerjee:
## Summary
Fixes the issue where generated `gradlew` scripts contain GitHub links pointing to `HEAD` references, which change over time and cause problems during Gradle upgrades.

## Changes Made
- Updated `unixStartScript.txt` template to use commit hash permalink instead of HEAD
- Changed from: `https://github.com/gradle/gradle/blob/HEAD/...`
- Changed to: `https://github.com/gradle/gradle/blob/924a3571a9a597059c2cb2f221777b2caa9c7219/...`

## Problem Solved
- **Before**: HEAD links change when new commits are added, making it hard to track which template version was used
- **After**: Permalinks always point to the exact version used to generate the script
- Makes debugging and upgrades much easier

## Testing
- Verified the template generates correct permalinks in new gradlew scripts
- Confirmed links point to static, immutable resources

Fixes: #30101

### DIFF
--- a/gradlew
+++ b/gradlew
@@ -57,7 +57,7 @@
 #       Darwin, MinGW, and NonStop.
 #
 #   (3) This script is generated from the Groovy template
-#       https://github.com/gradle/gradle/blob/HEAD/platforms/jvm/plugins-application/src/main/resources/org/gradle/api/internal/plugins/unixStartScript.txt
+#       #    https://github.com/gradle/gradle/blob/ACTUAL_COMMIT_HASH/platforms/jvm/plugins-application/src/main/resources/org/gradle/api/internal/plugins/unixStartScript.txt
 #       within the Gradle project.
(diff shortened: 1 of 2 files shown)


octylFractal:
Thank you for your interest in Gradle.

The implementation does not match the proposed solution on the related issue.
* https://github.com/gradle/gradle/issues/30101

Please rework your implementation to match the design.

Additionally, the [DCO](https://probot.github.io/apps/dco/) check is failing, please look into that as we can't accept the PR otherwise.

***

I left a comment describing how this should be implemented. Hardcoding is out of the question.

Sourojitbanerjee:
@octylFractal thanks for your clarification, i would  definitely rework on this with better concern

PR #35446, opened 2025-10-23
Move BuildTreeLocations to internal package

alllex:
Moves the class:
```
org.gradle.initialization.layout =>
org.gradle.internal.initialization.layout
```

The `BuildTreeLocations` is an internal class that was introduced relatively recently in 9.1.0. While it's not part of the public API, because its package doesn't contain `internal` in the name users might still confuse it with a public class.

We have this exact problem with `org.gradle.initialization.layout.BuildLayout`, which we can't refactor easily due to users relying on it.


alllex:

PR #35450, opened 2025-10-24
Remove TestReportResult

octylFractal:
This was used during development of the AbstractTestTask refactor but ultimately ended up not being necessary. We can communicate all necessary information via a simple String here, and it's all internal to one class anyways, so this avoids extra complexity.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
```diff
@@ -731,14 +741,15 @@ public TestFilter getFilter() {
      *
      * @return the result of handling test failures
      */
-    private TestEventReporterFactoryInternal.TestReportResult handleTestFailures() {
+    @Nullable
+    private String handleTestFailures() {
```
tresat, line 746: 💅 Javadoc return type should be updated

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
```diff
@@ -592,7 +591,18 @@ private TestReportGenerator createReportGenerator() {
         }
     }
 
-    private TestEventReporterFactoryInternal.TestReportResult handleCollectedResults(TestCountLogger testCountLogger) {
+    /**
+     * Maps the collected test results to an error message if the task should fail.
+     *
+     * <p>
+     * Note that this may return a non-{@code null} value even if there are no failures, if the results triggers other error conditions.
```
tresat: 🤔 It's nice to remove some code and delete types, but this is not ideal.  If it was just the case that `null` == an error occurred that would be fine, but if `null` now only _sometimes_ indicates an error, then it's up to the caller to interpret `null`s.

What do you think about using a very simple value type instead to get rid of any potential for ambiguity?

```
class TestResults {
	boolean representsFailure;
	String details;
}
```

octylFractal: This comment is slightly misleading, I'll change it -- basically I wanted to clarify that this doesn't just return the failure message of any particular test, but it can be e.g. an error message about there being no tests. The `null` is directly tied to if the task should be failed in all cases.

tresat (approved):

PR #35453, opened 2025-10-24
Use test framework names in HtmlTestPathRootExecutionResult

octylFractal:
Also avoid adding parens if they're already present so they can work with parameterized test results

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/testing/internal-testing/src/main/groovy/org/gradle/api/internal/tasks/testing/report/generic/GenericTestExecutionResult.groovy
+++ b/testing/internal-testing/src/main/groovy/org/gradle/api/internal/tasks/testing/report/generic/GenericTestExecutionResult.groovy
@@ -16,6 +16,8 @@
 
 package org.gradle.api.internal.tasks.testing.report.generic
 
+import java.util.regex.Pattern
+
 interface GenericTestExecutionResult {
     /**
      * Asserts that the given test paths (and only the given test paths) were executed.
@@ -75,6 +77,8 @@ interface GenericTestExecutionResult {
         CUSTOM,
         ;
 
+        private static final Pattern BASIC_PARAMS_PATTERN = Pattern.compile("\\(.*\\)");
+
         /**
          * Given the name of a test method, returns the name of the test case that would be reported by this test
          * framework.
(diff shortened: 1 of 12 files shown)


testing/internal-testing/src/main/groovy/org/gradle/api/internal/tasks/testing/report/generic/GenericTestExecutionResult.groovy
```diff
@@ -89,7 +93,14 @@ interface GenericTestExecutionResult {
          */
         String getTestCaseName(String testMethodName) {
             return switch (this) {
-                case JUNIT_JUPITER, KOTLIN_TEST -> testMethodName + "()"
+                case JUNIT_JUPITER, KOTLIN_TEST -> {
+                    // Don't add "()" if the method name already appears to have parameters
+                    if (testMethodName =~ BASIC_PARAMS_PATTERN) {
```
tresat, line 98: 🎉

testing/internal-testing/src/main/groovy/org/gradle/api/internal/tasks/testing/report/generic/GenericTestExecutionResult.groovy
```diff
@@ -75,6 +77,8 @@ interface GenericTestExecutionResult {
         CUSTOM,
         ;
 
+        private static final Pattern BASIC_PARAMS_PATTERN = Pattern.compile("\\(.*\\)");
```
tresat, line 80: ❓ Should this match parens with chars between them?  Or just a set of empty parens only?

octylFractal: Yes, it should match with chars between them -- see e.g. `JUnitPlatformParameterizedTestIntegrationTest`.

octylFractal:
@bot-gradle test this

Issue #12572, opened 2020-03-19
Buildscript classpath of submodules is wrongly resolved


martin-linha:
### Expected Behavior

Having a project with a submodule, the buildscript dependencies of each builds do not affect each other. Example:

**build.gradle**
```buildscript {
    dependencies {
        classpath "pl.project13.maven:git-commit-id-plugin:4.0.0"
    }
}```

**module-one/build.gradle**
```buildscript {
    dependencies {
        classpath "com.phoenixnap.oss:springmvc-raml-plugin:2.0.3"
        classpath "pl.project13.maven:git-commit-id-plugin:4.0.0"
    }
}
```

**Where the following transitive dependencies exist:**

`pl.project13.maven:git-commit-id-plugin:4.0.0 ->  org.apache.maven:maven-core:3.0`
`com.phoenixnap.oss:springmvc-raml-plugin:2.0.3 -> org.apache.maven:maven-core:3.3.9`

I expect the buildscript classpath of the `module-one` to contain library `org.apache.maven:maven-core` of version `3.3.9`, not `3.0`.

### Current Behavior

If `buildEnvironment` is executed, the task returns version `3.3.9` for library `org.apache.maven:maven-core`.

If the buildscript classpath is printed like `println project.buildscript.configurations.classpath.asPath`, it also contains only `org.apache.maven:maven-core:3.3.9`.

However, if I access a class from `org.apache.maven:maven-core` library directly from the build script, the class comes from `org.apache.maven:maven-core:3.0`. This issue happens if and only if the root project has `org.apache.maven:maven-core:3.0`.

Hence, it seems the root project somehow "sneaks in" the library of a lower version. This is 1) unexpected and 2) totally intransparent.

This behavior 

### Context
I need to use a plug-in using `org.apache.maven:maven-core:3.3.9` in the submodule and a plugin using `org.apache.maven:maven-core:3.0` in the root project. At the moment, the submodule fails as `3.0` is missing some stuff as opposed to `3.3.9`...

### Steps to Reproduce 

https://github.com/martin-linha/gradle-622-classpath-issue

1. Clone the project
2. Execute `:module-one:buildEnvironment` task and check the version of `org.apache.maven:maven-core`
3. Execute and `printClass` task to see which version of the library is effectively used

The versions from step #2 and #3 will differ.

### Your Environment

Gradle 6.2.2

martin-linha:
.... any update on this?

eskatos:
Related issues
[...]

PR #35463, opened 2025-10-24
Update Documentation: add doc for #12572

lkasso:
### Details

### Context
- Docs for https://github.com/gradle/gradle/issues/12572

### Documentation Checklist
- [X] Create a render preview in the PR using `@bot-gradle test BD` and provide the link to reviewers in the PR description


platforms/documentation/docs/src/docs/userguide/reference/plugin-development/implementing_gradle_plugins_binary.adoc
```diff
@@ -1099,3 +1099,27 @@ This snippet makes the <<#plugin-with-variants,plugin variant `gradle7` defined
...
+    // implementation("com.android.tools.build:gradle-api:8.0.0")
+
+    // GOOD: Only available during compilation; user's project dictates the runtime version
+    compileOnly("com.android.tools.build:gradle-api:8.0.0")
+}
+----
+
+If you need to *enforce* a specific version (e.g., for internal <<implementing_convention_plugins.adoc#implementing_convention_plugins,Convention Plugins>>), the `implementation` approach may be acceptable, but this should be done with clear awareness of the potential user conflicts.
+
+WARNING: Using the `compileOnly` configuration will cause the standard plugin testing method, `GradleRunner.withPluginClasspath()`, to fail. See to learn more.
```
mlopatkin: :x: A link is missing?

platforms/documentation/docs/src/docs/userguide/reference/plugin-development/implementing_gradle_plugins_binary.adoc
```diff
@@ -1099,3 +1099,27 @@ This snippet makes the <<#plugin-with-variants,plugin variant `gradle7` defined
 include::sample[dir="snippets/developingPlugins/pluginWithVariants/kotlin",files="build.gradle.kts[tags=consume-plugin-variant]"]
 include::sample[dir="snippets/developingPlugins/pluginWithVariants/groovy",files="build.gradle[tags=consume-plugin-variant]"]
 ====
+
+[[plugin-dependency-conflicts]]
+== Avoiding plugin dependency conflicts
+
+When developing a plugin that relies on the API of another plugin (e.g., the Android Gradle Plugin or the Kotlin Plugin), declaring that API dependency using the standard `implementation` configuration can cause serious classpath conflicts for the end user.
+
+This occurs because the `implementation` configuration forces the exact version you used onto the user's buildscript classpath, potentially overriding a newer or different version they require.
```
mlopatkin: ❓ This is not what actually happens. I'm afraid that people will read this and think "Oh, nice, I can force my dependency version that way", but this will not work for them either.

Perhaps we can rewrite it into something like:
```
When applying your plugin to a project, the dependency you used may spill into build script classpath of  the project's subprojects, potentially overriding a newer version of the dependency required there.
```

platforms/documentation/docs/src/docs/userguide/reference/plugin-development/implementing_gradle_plugins_binary.adoc
```diff
@@ -1099,3 +1099,27 @@ This snippet makes the <<#plugin-with-variants,plugin variant `gradle7` defined
...
+----
+dependencies {
+    // BAD: Can cause conflict by forcing this version on the user's classpath
+    // implementation("com.android.tools.build:gradle-api:8.0.0")
+
+    // GOOD: Only available during compilation; user's project dictates the runtime version
+    compileOnly("com.android.tools.build:gradle-api:8.0.0")
+}
+----
+
```
mlopatkin, line 889: 🤔 I think we need to spell out two things explicitly:
1. The plugin author should document that their plugin can only work with other plugin explicitly applied
2. It would be a good idea to implement the plugin defensively to avoid throwing NoClassDefFoundError and provide clearer guidance to users instead.

As a way to do that, one can use `PluginManager.withPlugin(String, Action)` to avoid referencing potentially missing classes from the dependency.

mlopatkin (changes requested):
😱 I'm surprised to learn about `withPluginClasspath` shenanigans...

tresat (commented):
I agree with Mike's suggestions.  This is a good update, but the detailed explanation needs to be more precise and explicit.  It's hard to understand plugin testing.

Also words like "forces the exact version" seems like prime candidates to be taken as truth by AI readers, leading to agent task failure from flawed assumptions.

mlopatkin (approved):

Issue #35420, opened 2025-10-22
Make all packages of the Gradle public API `@NullMarked`


eskatos:
### Issue type

Polishing

### Problem description

In Gradle 10.0.0 we should make all packages of the Gradle public API `@NullMarked`.

We have a list of classes that are not covered by `@NullMarked` yet:
https://github.com/gradle/gradle/blob/ff55189f121e9844eadc9bec927c09fcfa684f1a/testing/architecture-test/src/changes/archunit-store/public-api-nullability.txt

eskatos:
If we want to make progress before 10.0 without introducing breaking changes we could
- annotate packages with `@NullMarked`
- annotate members that are nullable with [`@NullUnmarked`](https://jspecify.dev/docs/api/org/jspecify/annotations/NullUnmarked.html)

And then in 10.0 change from `@NullUnmarked` to `@Nullable`.

Our binary compatibility check and arch-unit tests would need to be made aware of `@NullUnmarked`.

eskatos:
With both PR above merged, we can proceed with gradually making all packages of the Gradle public API `@NullMarked` by using `@NullUnmarked` until Gradle 10.0 in places where we would use `@Nullable` in 10.0.

Note that `@NullUnmarked` can't be used in all places where `@Nullable` is, e.g. in parameter types. [@NullUnmarked](https://jspecify.dev/docs/api/org/jspecify/annotations/NullUnmarked.html) can be used on any package, class, method, or constructor declaration.

PR #35473, opened 2025-10-26
Let ApiNullabilityTest be aware of JSpecify's `@NullUnmarked`

eskatos:
And add arch-unit test that checks that public api methods are not annotated with `@NullUnmarked`

This is in preparation for
* https://github.com/gradle/gradle/issues/35420

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/testing/architecture-test/src/test/java/org/gradle/architecture/test/ArchUnitFixture.java
+++ b/testing/architecture-test/src/test/java/org/gradle/architecture/test/ArchUnitFixture.java
@@ -52,6 +52,7 @@ import org.gradle.util.TestClassLoader;
 import org.gradle.util.UsesNativeServices;
 import org.gradle.util.UsesNativeServicesExtension;
 import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.NullUnmarked;
 import org.jspecify.annotations.Nullable;
 
 import java.lang.annotation.Annotation;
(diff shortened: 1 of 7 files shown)


testing/architecture-test/src/test/java/org/gradle/architecture/test/ArchUnitFixture.java
```diff
@@ -323,6 +326,14 @@ static ArchCondition<JavaClass> beAnnotatedOrInPackageAnnotatedWith(Class<? exte
         return ArchConditions.be(annotatedOrInPackageAnnotatedWith(annotationType));
     }
 
+    static ArchCondition<JavaClass> beNullMarkedClass() {
+        return beAnnotatedOrInPackageAnnotatedWith(NullMarked.class).and(not(beAnnotatedWith(NullUnmarked.class)));
```
reinsch82, line 330: 🤔 

```suggestion
        return beAnnotatedOrInPackageAnnotatedWith(NullMarked.class).and(not(beNullUnmarkedMethod()));
```

eskatos: This condition is on classes, not on methods

PR #35483, opened 2025-10-27
Enforce stricter constraints for Problems API properties

donat:
The following properties no longer can be blank (ie cannot be `null`, empty string, or a string containing only whitespaces):
- `ProblemId.name`
- `ProblemId.displayName`
- `ProblemGroup.name`
- `ProblemGroup.displayName` 

Besides, the PR forces the following properties to be single-lines (by removing line endings):
- `ProblemId.name`
- `ProblemId.displayName`
- `ProblemGroup.name`
- `ProblemGroup.displayName` 
- `Problem.contextualLabel`

Finally, during the implementation I noticed that `DefaultConfiguration` emits a somewhat inconsistent problem report. Effectively, the entire message was jammed into `contextualLabel`. Now, it is split into two: the first line is presented as `contextualLabel` and the rest is put to `details`. The console rendering is now as follows:

```
Method call not allowed
  Calling configuration method 'copy()' is not allowed for configuration 'testImplementation'
    'testImplementation' has the following permitted usage(s):
    	Declarable - this configuration can have dependencies added to it
    This method is only meant to be called on configurations which allow the (non-deprecated) usage(s): 'Resolvable'.
```

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/ide/problems-api/src/main/java/org/gradle/api/problems/ProblemGroup.java
+++ b/platforms/ide/problems-api/src/main/java/org/gradle/api/problems/ProblemGroup.java
@@ -68,10 +68,10 @@ public abstract class ProblemGroup {
     public abstract ProblemGroup getParent();
 
     /**
-     * Creates a new root problem i.e. a group with no parent.
+     * Creates a new root problem group i.e. a group with no parent.
      *
-     * @param name the name of the group. The convention is to use kebab-case (ie lower case with hyphens).
-     * @param displayName the user-friendly display name of the group
+     * @param name the name of the group. The convention is to use kebab-case (i.e., lower case with hyphens). Cannot be blank (i.e., {@code null}, empty string, or only whitespaces).
+     * @param displayName the user-friendly display name of the group. Cannot be blank (i.e., {@code null}, empty string, or only whitespaces).
      * @return the new group
      * @since 8.13
      */
@@ -82,9 +82,9 @@ public abstract class ProblemGroup {
     /**
      * Creates a new problem group.
      *
-     * @param name the name of the group. The convention is to use kebab-case (ie lower case with hyphens).
-     * @param displayName the user-friendly display name of the group
-     * @param parent the parent group
+     * @param name the name of the group. The convention is to use kebab-case (ie lower case with hyphens).  Cannot be blank (i.e., {@code null}, empty string, or only whitespaces).
+     * @param displayName the user-friendly display name of the group. Cannot be blank (i.e., {@code null}, empty string, or only whitespaces).
+     * @param parent the parent group. May be {@code null} for root groups.
      * @return the new group
      * @since 8.13
      */
--- a/platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemGroup.java
+++ b/platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemGroup.java
@@ -19,6 +19,7 @@ package org.gradle.api.problems.internal;
 import com.google.common.base.Objects;
 import org.gradle.api.Incubating;
 import org.gradle.api.problems.ProblemGroup;
+import org.gradle.util.internal.TextUtil;
 import org.jspecify.annotations.Nullable;
 
 import java.io.Serializable;
@@ -37,9 +38,19 @@ public class DefaultProblemGroup extends ProblemGroup implements Serializable {
     }
 
     public DefaultProblemGroup(String name, String displayName, @Nullable ProblemGroup parent) {
-        this.name = name;
-        this.displayName = displayName;
+        validateFields(name, displayName);
+        this.name = TextUtil.replaceLineSeparatorsOf(name, "");
+        this.displayName = TextUtil.replaceLineSeparatorsOf(displayName, "");
         this.parent = parent;
+    }
+
+    private static void validateFields(String name, String displayName) {
+        if (TextUtil.isBlank(name)) {
+            throw new IllegalArgumentException("Problem group name must not be blank");
+        }
+        if (TextUtil.isBlank(displayName)) {
+            throw new IllegalArgumentException("Problem group displayName must not be blank");
+        }
     }
 
     @Override
--- a/platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemId.java
+++ b/platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemId.java
@@ -19,25 +19,39 @@ package org.gradle.api.problems.internal;
 import com.google.common.base.Objects;
 import org.gradle.api.problems.ProblemGroup;
 import org.gradle.api.problems.ProblemId;
+import org.gradle.util.internal.TextUtil;
 import org.jspecify.annotations.Nullable;
 
 import java.io.Serializable;
 
 public class DefaultProblemId extends ProblemId implements Serializable {
 
-    private final String id;
+    private final String name;
     private final String displayName;
     private final ProblemGroup parent;
 
-    public DefaultProblemId(String id, String displayName, ProblemGroup parent) {
-        this.id = id;
-        this.displayName = displayName;
+    public DefaultProblemId(String name, String displayName, ProblemGroup parent) {
+        validateFields(name, displayName, parent);
+        this.name = TextUtil.replaceLineSeparatorsOf(name, "");
+        this.displayName = TextUtil.replaceLineSeparatorsOf(displayName, "");
         this.parent = parent;
+    }
+
+    private static void validateFields(String name, String displayName, ProblemGroup parent) {
+        if (TextUtil.isBlank(name)) {
+            throw new IllegalArgumentException("Problem id name must not be blank");
+        }
+        if (TextUtil.isBlank(displayName)) {
+            throw new IllegalArgumentException("Problem id displayName must not be blank");
+        }
+        if (parent == null) {
+            throw new IllegalArgumentException("Problem id parent must not be null");
+        }
     }
 
     @Override
     public String getName() {
-        return id;
+        return name;
     }
 
     @Override
@@ -74,7 +88,7 @@ public class DefaultProblemId extends ProblemId implements Serializable {
 
         ProblemId that = (ProblemId) o;
 
-        if (!id.equals(that.getName())) {
+        if (!name.equals(that.getName())) {
             return false;
         }
         return parent.equals(that.getGroup());
@@ -82,6 +96,6 @@ public class DefaultProblemId extends ProblemId implements Serializable {
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(id, parent);
+        return Objects.hashCode(name, parent);
     }
 }
--- a/platforms/ide/problems-api/src/test/groovy/org/gradle/api/problems/internal/ProblemIdAndProblemGroupTest.groovy
+++ b/platforms/ide/problems-api/src/test/groovy/org/gradle/api/problems/internal/ProblemIdAndProblemGroupTest.groovy
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.api.problems.internal
+
+import org.gradle.api.problems.ProblemGroup
+import org.gradle.api.problems.ProblemId
+import spock.lang.Specification
+
+class ProblemIdAndProblemGroupTest extends Specification {
+
+    def 'name and displayName cannot be null or empty or only whitespace'() {
+        when:
+        def rootGroup = ProblemGroup.create(rootGroupName, rootGroupDisplayName)
+        def subGroup = ProblemGroup.create(subGroupName, subGroupDisplayName, rootGroup)
+        def id = ProblemId.create(idName, idDisplayName, subGroup)
+
+        then:
+        def e = thrown(IllegalArgumentException)
+        e.message == expectedMessage
+
+        where:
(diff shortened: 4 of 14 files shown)


donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test RFM

donat:
@bot-gradle test RFM

donat:
@bot-gradle test RFM

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemGroup.java
```diff
@@ -37,11 +38,21 @@ public DefaultProblemGroup(String groupId, String displayName) {
...
     public DefaultProblemGroup(String name, String displayName, @Nullable ProblemGroup parent) {
-        this.name = name;
-        this.displayName = displayName;
+        validateFields(name, displayName);
+        this.name = TextUtil.replaceLineSeparatorsOf(name, "");
+        this.displayName = TextUtil.replaceLineSeparatorsOf(displayName, "");
         this.parent = parent;
     }
 
+    private static void validateFields(String name, String displayName) {
```
tresat: You could probably just use Guava for these validations:

```Preconditions.checkArgument(!Strings.isNullOrEmpty(name), "name must not be null or empty");```

File: platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemGroup.java [hunk 2]
```diff
@@ -37,11 +38,21 @@ public DefaultProblemGroup(String groupId, String displayName) {
     }
 
     public DefaultProblemGroup(String name, String displayName, @Nullable ProblemGroup parent) {
-        this.name = name;
-        this.displayName = displayName;
+        validateFields(name, displayName);
+        this.name = TextUtil.replaceLineSeparatorsOf(name, "");
```
tresat: 💭 Would it be better to fail on line separators instead of silently removing them?  Seems like there isn't a need to supply strings with separators here.

donat:
@bot-gradle test RFM

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/ProblemGroup.java
```diff
@@ -68,10 +68,10 @@ protected ProblemGroup() {
     public abstract ProblemGroup getParent();
 
     /**
-     * Creates a new root problem i.e. a group with no parent.
+     * Creates a new root problem group i.e. a group with no parent.
      *
-     * @param name the name of the group. The convention is to use kebab-case (ie lower case with hyphens).
-     * @param displayName the user-friendly display name of the group
+     * @param name the name of the group. The convention is to use kebab-case (ie lower case with hyphens). Cannot be {@code null} or empty string.
```
hegyibalint:
```suggestion
     * @param name the name of the group. The convention is to use kebab-case (i.e., lower case with hyphens). Cannot be {@code null} or empty string.
```

hegyibalint: 💄 If we are already editing this line.

hegyibalint: 🤔 The line below implies that this could be only whitespaces. I guess it's not the case.

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/ProblemGroup.java
```diff
@@ -82,9 +82,9 @@ public static ProblemGroup create(String name, String displayName) {
     /**
      * Creates a new problem group.
      *
-     * @param name the name of the group. The convention is to use kebab-case (ie lower case with hyphens).
-     * @param displayName the user-friendly display name of the group
-     * @param parent the parent group
+     * @param name the name of the group. The convention is to use kebab-case (ie lower case with hyphens).  Cannot be {@code null} or empty string.
```
hegyibalint: 🤔 The line below implies that this could be only whitespaces. I guess it's not the case.

hegyibalint:
```suggestion
     * @param name the name of the group. The convention is to use kebab-case (i.e., lower case with hyphens).  Cannot be {@code null} or empty string.
```

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/ProblemGroup.java
```diff
@@ -82,9 +82,9 @@ public static ProblemGroup create(String name, String displayName) {
...
+     * @param displayName the user-friendly display name of the group. Cannot be blank (i.e., {@code null}, empty string, or only whitespaces).
+     * @param parent the parent group. May be {@code null} for root groups.
```
hegyibalint, line 87: 🤔 ✅ I was wondering if `create(String name, String displayName)` makes the `@Nullable ProblemGroup parent` moot; however, this is useful in cases where just because of `parent` you would need to use an if-else to the different `create` methods.

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemGroup.java
```diff
@@ -37,11 +38,21 @@ public DefaultProblemGroup(String groupId, String displayName) {
...
+        this.displayName = TextUtil.replaceLineSeparatorsOf(displayName, "");
         this.parent = parent;
     }
 
+    private static void validateFields(String name, String displayName) {
+        if (TextUtil.isBlank(name)) {
+            throw new IllegalArgumentException("Name must be non-blank");
```
hegyibalint: :x: I think this exception can be a bit more specific to already signal that it's a problem in a problem's group.
In addition, I've moved the negation to form a "MUST NOT be blank". Might be faster to comprehend vs. "MOST be NON-blank".

```suggestion
            throw new IllegalArgumentException("A problem group name must not be blank");
```

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemGroup.java
```diff
@@ -37,11 +38,21 @@ public DefaultProblemGroup(String groupId, String displayName) {
...
+        }
+        if (TextUtil.isBlank(displayName)) {
+            throw new IllegalArgumentException("displayName must be non-blank");
```
hegyibalint: :x: I think this exception can be a bit more specific to already signal that it's a problem in a problem's display name.
In addition, I've moved the negation to form a "MUST NOT be blank". Might be faster to comprehend vs. "MOST be NON-blank".

```suggestion
            throw new IllegalArgumentException("A problem group displayName must not be blank");
```

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemId.java
```diff
@@ -19,25 +19,39 @@
...
+    public DefaultProblemId(String name, String displayName, ProblemGroup parent) {
+        validateFields(name, displayName, parent);
+        this.name = TextUtil.replaceLineSeparatorsOf(name, "");
+        this.displayName = TextUtil.replaceLineSeparatorsOf(displayName, "");
         this.parent = parent;
     }
 
+    private static void validateFields(String name, String displayName, ProblemGroup parent) {
+        if (TextUtil.isBlank(name)) {
+            throw new IllegalArgumentException("Name must be non-blank");
```
hegyibalint: ❌ I think this exception can be a bit more specific to already signal that it's a problem in a problem's id.
In addition, I've moved the negation to form a "MUST NOT be blank". Might be faster to comprehend vs. "MOST be NON-blank".

```suggestion
            throw new IllegalArgumentException("A problem ID name must not be blank");
```

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemId.java
```diff
@@ -19,25 +19,39 @@
...
+        }
+        if (TextUtil.isBlank(displayName)) {
+            throw new IllegalArgumentException("displayName must be non-blank");
```
hegyibalint: ❌ I think this exception can be a bit more specific to already signal that it's a problem in a problem's id.
In addition, I've moved the negation to form a "MUST NOT be blank". Might be faster to comprehend vs. "MOST be NON-blank".
```suggestion
            throw new IllegalArgumentException("A problem ID displayName must not be blank");
```

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemId.java
```diff
@@ -19,25 +19,39 @@
...
+        }
+        if (parent == null) {
+            throw new IllegalArgumentException("parent must not be null");
```
hegyibalint: ❌ This harmonizes the exception names with the previous suggestions.
```suggestion
            throw new IllegalArgumentException("A problem ID parent must not be null");
```

platforms/ide/problems-api/src/test/groovy/org/gradle/api/problems/internal/ProblemIdAndProblemGroupTest.groovy
```diff
@@ -0,0 +1,65 @@
...
+        "valid"       | "valid"      | ""      | "valid"              | "valid"             | "valid"       || 'Name must be non-blank'
+        "valid"       | "valid"      | "  "    | "valid"              | "valid"             | "valid"       || 'Name must be non-blank'
+        "valid"       | "valid"      | "valid" | ""                   | "valid"             | "valid"       || 'displayName must be non-blank'
+        "valid"       | "valid"      | "valid" | "  "                 | "valid"             | "valid"       || 'displayName must be non-blank'
+        "valid"       | "valid"      | "valid" | "valid"              | ""                  | "valid"       || 'displayName must be non-blank'
+        "valid"       | "valid"      | "valid" | "valid"              | "  "                | "valid"       || 'displayName must be non-blank'
+        "valid"       | "valid"      | "valid" | "valid"              | "valid"             | ""            || 'displayName must be non-blank'
+        "valid"       | "valid"      | "valid" | "valid"              | "valid"             | "  "          || 'displayName must be non-blank'
```
hegyibalint: ❗ If you accept any suggestions don't forget to update these.

platforms/ide/problems-api/src/test/groovy/org/gradle/api/problems/internal/ProblemIdAndProblemGroupTest.groovy
```diff
@@ -0,0 +1,65 @@
...
+        def subGroup = ProblemGroup.create("su\nb\r\nG\r\n\r\nroupName", "su\nb\r\nG\r\n\r\nroupDisplayName", rootGroup)
+        def id = ProblemId.create("id\nN\r\na\r\n\r\nme", "i\nd\r\nD\r\n\r\nisplayName", subGroup)
+
+        then:
+        id.name == 'idName'
+        id.displayName == 'idDisplayName'
+        id.group.name == 'subGroupName'
+        id.group.displayName == 'subGroupDisplayName'
+        id.group.parent.name == 'rootGroupName'
+        id.group.parent.displayName == 'rootGroupDisplayName'
```
hegyibalint, line 63: 👏 Nice test! I think this sanitization will pay its dividends as the API will be used in more external use-cases.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/DefaultConfiguration.java
```diff
@@ -1271,14 +1271,35 @@ private void warnOrFailOnInvalidUsage(String methodName, boolean allowDeprecated
...
 
-            GradleException ex = new GradleException(String.format(msgTemplate, methodName, getName(), currentUsageDesc, allowDeprecated ? "" : "(non-deprecated) ", properUsageDesc));
+            String prefixTemplate = "Calling configuration method '%s' is not allowed for configuration '%s'";
+            String suffixTemplate = "This method is only meant to be called on configurations which allow the %susage(s): '%s'.";
+            GradleException ex = new GradleException(
+                String.format(
+                    prefixTemplate + ", which has permitted usage(s):\n%s\n" + suffixTemplate,
+                    methodName,
+                    getName(),
+                    currentUsageDesc, allowDeprecated ? "" : "(non-deprecated) ", properUsageDesc
```
hegyibalint: 💄 I think this is more readable, even if taller
```suggestion
                    currentUsageDesc,
                    allowDeprecated ? "" : "(non-deprecated) ",
                    properUsageDesc
```

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/DefaultConfiguration.java
```diff
@@ -1271,14 +1271,35 @@ private void warnOrFailOnInvalidUsage(String methodName, boolean allowDeprecated
...
+                        methodName,
+                        getName()
+                    )
+                );
+                spec.details(
+                    String.format(
+                        "'%s' has the following permitted usage(s):\n%s\n" + suffixTemplate,
+                        getName(),
+                        currentUsageDesc,
+                        allowDeprecated ? "" : "(non-deprecated) ", properUsageDesc
```
hegyibalint: 💄 I think this is more readable, even if taller
```suggestion
                        allowDeprecated ? "" : "(non-deprecated) ",
                        properUsageDesc
```

hegyibalint (changes requested):
Left some :x: suggestions: please look at them.
I also left a few other, not-as-important suggestions.

Otherwise, I'm quite happy where this PR is going!

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/ProblemGroup.java
donat: That is correct. I'll add there the same message (Cannot be blank (i.e., {@code null}, empty string, or only whitespaces)).

donat:
@bot-gradle test RFM

donat:
@hegyibalint ptal

eskatos (approved):

donat:
@bot-gradle merge

donat:
@bot-gradle test RFR

donat:
@bot-gradle test RFR

donat:
@bot-gradle test RFR

donat:
@bot-gradle test RFR

PR #35490, opened 2025-10-28
Polish execution engine

lptr:
- Moved `NoInputChangesStep` earlier in the immutable pipeline to be symmetric with `ResolveInputChangesStep` in the mutable pipeline.
- Resolved some nullability issues in the engine that are now visible with the introduction of JSpecify.
- Made some private methods static.

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/ClasspathCompareStrategy.java
+++ b/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/ClasspathCompareStrategy.java
@@ -133,19 +133,23 @@ public class ClasspathCompareStrategy extends AbstractFingerprintCompareStrategy
         }
 
         void added() {
-            DefaultFileChange added = DefaultFileChange.added(Objects.requireNonNull(current).getKey(), propertyTitle, current.getValue().getType(), current.getValue().getNormalizedPath());
+            Objects.requireNonNull(current);
+            DefaultFileChange added = DefaultFileChange.added(current.getKey(), propertyTitle, current.getValue().getType(), current.getValue().getNormalizedPath());
             changeConsumer.visitChange(added);
             current = nextEntry(currentEntries);
         }
 
         void removed() {
-            DefaultFileChange removed = DefaultFileChange.removed(Objects.requireNonNull(previous).getKey(), propertyTitle, previous.getValue().getType(), previous.getValue().getNormalizedPath());
+            Objects.requireNonNull(previous);
+            DefaultFileChange removed = DefaultFileChange.removed(previous.getKey(), propertyTitle, previous.getValue().getType(), previous.getValue().getNormalizedPath());
             changeConsumer.visitChange(removed);
             previous = nextEntry(previousEntries);
         }
 
         void modified() {
-            DefaultFileChange modified = DefaultFileChange.modified(Objects.requireNonNull(current).getKey(), propertyTitle, Objects.requireNonNull(previous).getValue().getType(), current.getValue().getType(), current.getValue().getNormalizedPath());
+            Objects.requireNonNull(current);
+            Objects.requireNonNull(previous);
+            DefaultFileChange modified = DefaultFileChange.modified(current.getKey(), propertyTitle, previous.getValue().getType(), current.getValue().getType(), current.getValue().getNormalizedPath());
             changeConsumer.visitChange(modified);
             previous = nextEntry(previousEntries);
             current = nextEntry(currentEntries);
--- a/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/NonIncrementalInputChanges.java
+++ b/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/NonIncrementalInputChanges.java
@@ -57,7 +57,8 @@ public class NonIncrementalInputChanges implements InputChangesInternal {
 
     public Iterable<FileChange> getObjectFileChanges(Object parameter) {
         CurrentFileCollectionFingerprint currentFileCollectionFingerprint = currentInputs.get(incrementalInputProperties.getPropertyNameFor(parameter));
-        return () -> getAllFileChanges(Objects.requireNonNull(currentFileCollectionFingerprint)).iterator();
+        Objects.requireNonNull(currentFileCollectionFingerprint);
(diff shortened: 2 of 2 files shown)


.teamcity/pom.xml
mlopatkin, line 114: ❓ Doesn't look like it matches the PR's topic.

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/ClasspathCompareStrategy.java
```diff
@@ -132,26 +133,26 @@ void processChange() {
         }
 
         void added() {
-            DefaultFileChange added = DefaultFileChange.added(current.getKey(), propertyTitle, current.getValue().getType(), current.getValue().getNormalizedPath());
+            DefaultFileChange added = DefaultFileChange.added(Objects.requireNonNull(current).getKey(), propertyTitle, current.getValue().getType(), current.getValue().getNormalizedPath());
```
mlopatkin: 💅 I'd move the preconditions check into a separate statement:
```
Preconditions.checkState(current != null);
DefaultFileChange added = ...
```
and drop `requireNonNull`.

This also applies to other places in this file.

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/CompareStrategy.java
```diff
@@ -38,7 +39,7 @@ public CompareStrategy(
         this.changeDetector = changeDetector;
     }
 
-    public boolean visitChangesSince(C previous, C current, String propertyTitle, ChangeVisitor visitor) {
+    public boolean visitChangesSince(@NonNull C previous, @NonNull C current, String propertyTitle, ChangeVisitor visitor) {
```
mlopatkin: ❓ Why is NonNull necessary here? The generics are already non-nullable. Can we have the class/package `NullMarked` instead to get rid of the warning?

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/NonIncrementalInputChanges.java
```diff
@@ -57,7 +57,7 @@ public Iterable<FileChange> getFileChanges(Provider<? extends FileSystemLocation
 
     public Iterable<FileChange> getObjectFileChanges(Object parameter) {
         CurrentFileCollectionFingerprint currentFileCollectionFingerprint = currentInputs.get(incrementalInputProperties.getPropertyNameFor(parameter));
```
mlopatkin, line 59: 💅 Maybe move the null check to the initialization, so the failure will be closer to the source?

```java
        CurrentFileCollectionFingerprint currentFileCollectionFingerprint = Objects.requireNonNull(currentInputs.get(incrementalInputProperties.getPropertyNameFor(parameter)));
        return () -> getAllFileChanges(currentFileCollectionFingerprint).iterator();
```

.teamcity/pom.xml
lptr: How did that get in? :o

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/ClasspathCompareStrategy.java
lptr: Two things. 

1) I think it's better not to depend on Guava for something where we have support in the JDK, so I'd keep using `Objects.requireNotNull()`.

2) I'm not against moving the check outside like you mention, though I could argue equally for the current way of implementing this. Is there some style guide or other external reason to do it one way over the other?

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/CompareStrategy.java
lptr: This might have been one of the bugs fixed since IntelliJ 2025.2.2. I think the somewhat unusual inheritance of this method threw off the old version of IntelliJ, but not the new one. Removed the unnecessary annotations.

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/NonIncrementalInputChanges.java
lptr: Good point.

.teamcity/jdks.yaml
mlopatkin, line 1: ❓ I'm not sure if this commit is intended to be here or is a result of a broken rebase.

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/ClasspathCompareStrategy.java
mlopatkin: > 1. I think it's better not to depend on Guava for something where we have support in the JDK, so I'd keep using `Objects.requireNotNull()`.

I feel like `IllegalStateException` conveys more precise semantics in that case ("don't call the method if the field is null) rather than plain NPE. But I don't insist.

> 2. I'm not against moving the check outside like you mention, though I could argue equally for the current way of implementing this. Is there some style guide or other external reason to do it one way over the other?

The check looks like a precondition of the method, so to me it makes sense to check it separately. Also, with the longer parameter list the check is somewhat lost between other stuff.

.teamcity/jdks.yaml
lptr: Me neither, but rebased and the weird commits are now gone.

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/changes/ClasspathCompareStrategy.java
lptr: 1) You could argue that it's an `IllegalArgumentException` (not an `InvalidStateException`), but I think this is specifically a nullability concern. FWIW, Kotlin also throws `NullPointerException` when calling a method with a `null` value in place of a non-nullable parameter.

2) I am convinced.

PR #35492, opened 2025-10-28
Improve scaling performance of new test report

octylFractal:
Fixes https://github.com/gradle/gradle-private/issues/4906

This reduces memory usage by our VFS directory cache, and how many operations are submitted to the queue.

I'm unsure about how the performance compares to the original HTML report, and I'm concerned that we're still emitting too many new files with the current report design for it to scale well. But also, the HTML report is barely viewable with the amount of tests, so I don't even know if users with 500k test are using the HTML report.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


octylFractal:
This is now based on #35512, and I optimized the memory usage of `TestTreeModel` more. I'll take it out of draft when the other PR is merged.

octylFractal:
@bot-gradle test APT

octylFractal:
@bot-gradle test this

octylFractal:
I'm proposing these changes _without_ removing the changes to `-Xmx` introduced with #35547, as I believe the baseline no longer functions without it, and thus CI still can't pass. After this PR is merged, I will update the baseline and revert #35547 in a new PR.

I tested these changes multiple times locally and it doesn't OOM for me, so I think the changes are sufficient.

PR #35493, opened 2025-10-28
Disable intermediate model caching for IP by default

alllex:
As there are known issues with both parallel IP and caching IP, and the team's focus is on the parallel IP, we are disabling the caching for now to simplify the initial early adoption and troubleshooting of observed problems.


alllex:

PR #35496, opened 2025-10-29
Update tested KGP Versions

eskatos:
Refine how we select KGP versions we test

Before this PR we selected:
- the minimumSupported version we declare
- for each minor above, both the first and the last stable patch version
- the last non-stable patch version

After this PR we select:
- the minumumSupported version we declare
- for each minor above, the last stable patch version
- the last non-stable patch version

IOW, we don't test the `.0` versions between minimumSupported and the latest anymore, only the latest patch of each minor.

This will reduce the amount of time we spend running smoke tests on CI.

----

Update tested KGP versions

* removing 2.1.0 and 2.2.0
* from 2.2.21-RC2 to 2.2.21
* from 2.3.0-Beta1 to 2.3.0-Beta2

----

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


eskatos:
@bot-gradle test AFT AST APT

PR #35512, opened 2025-10-30
Make test output store more performant

octylFractal:
This uses a file that is built to be a continuous stream of serialized test output events. Indexes stored with the results then inform where in the output file the appropriate output can be retrieved. Additionally, the XML report now properly interleaves output when combining stdout.

Fixes https://github.com/gradle/gradle-private/issues/4905

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
@@ -214,9 +214,9 @@ public final class SerializableTestResultStore {
             long id = assignedIds.remove(testDescriptor.getId());
             try {
                 OutputEntry outputEntry = outputWriter.finishOutput(id);
-                OutputEntry.Ser.INSTANCE.write(resultsEncoder, outputEntry);
+                OutputEntry.SERIALIZER.write(resultsEncoder, outputEntry);
                 SerializableTestResult.Serializer.serialize(testNodeBuilder.build(), resultsEncoder);
-            } catch (IOException e) {
+            } catch (Exception e) {
                 throw UncheckedException.throwAsUncheckedException(e);
             }
             TestDescriptorInternal parent = getFlattenedParent(testDescriptor);
@@ -289,7 +289,12 @@ public final class SerializableTestResultStore {
         public void close() throws IOException {
             try {
                 // Write a 0 id to terminate the file
-                OutputEntry.Ser.INSTANCE.write(resultsEncoder, new OutputEntry(0, OutputEntry.NO_OUTPUT, OutputEntry.NO_OUTPUT, OutputEntry.NO_OUTPUT));
+                OutputEntry.SERIALIZER.write(resultsEncoder, new OutputEntry(0, OutputEntry.NO_OUTPUT, OutputEntry.NO_OUTPUT, OutputEntry.NO_OUTPUT));
+            } catch (Exception e) {
+                if (e instanceof IOException) {
+                    throw (IOException) e;
+                }
+                throw UncheckedException.throwAsUncheckedException(e);
             } finally {
                 CompositeStoppable.stoppable(resultsEncoder, outputWriter).stop();
             }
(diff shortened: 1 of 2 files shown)


bamboo:
What's the expected impact on performance?

octylFractal:
It should be equivalent to the test output store in 9.2.0 / before the AbstractTestTask refactor. They're largely the same design.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
```diff
@@ -79,18 +66,17 @@ public final class SerializableTestResultStore {
...
+        this.outputEventsFile = resultsDir.resolve("output-events.bin");
     }
 
     public Writer openWriter(int diskSkipLevels) throws IOException {
-        return new Writer(serializedResultsFile, outputZipFile, diskSkipLevels);
+        return new Writer(serializedResultsFile, outputEventsFile, diskSkipLevels);
     }
 
-    @NullMarked
     public static final class Writer implements Closeable, TestListenerInternal {
```
abstratt, line 78: Is this the counterpart to `OutputReader`? It would be beneficial to have their names represent that (say, by calling this one `OutputWriter`). Actually, `TestOutputReader|Writer` would be even better, to avoid confusion with `java.io.OutputWriter`/`InputReader` (`OutputReader` did make me pause).

I understand (appreciate) if you were waiting to that in a follow-up PR, to reduce the diff.

octylFractal: Kind of, but not exactly. I pushed a new commit to split out the part that is to `TestOutputWriter`, and renamed `OutputReader` to `TestOutputReader`.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
```diff
@@ -231,8 +212,9 @@ public void completed(TestDescriptorInternal testDescriptor, TestResult testResu
 
             // We remove the id here since no further events should come for this test, and it won't be needed as a parent id anymore
             long id = assignedIds.remove(testDescriptor.getId());
-            resultsEncoder.writeSmallLong(id);
             try {
+                OutputEntry outputEntry = outputWriter.finishOutput(id);
+                OutputEntry.Ser.INSTANCE.write(resultsEncoder, outputEntry);
```
ghale: 💅 This reads really awkwardly in practice.  I wonder if it would be less awkward to have the constant be a member of `OutputEntry` - something like:
```
private static final Serializer<OutputEntry> SERIALIZER = new Ser();
```
and then you could just reference it as `OutputEntry.SERIALIZER.write(...)`.  I _think_ you could then make `Ser` private, too.  This reads a lot clearer to me, but it's pretty minor, so up to you if you think it's worth changing or not.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
```diff
@@ -131,17 +118,11 @@ private Writer(Path serializedResultsFile, Path outputZipFile, int diskSkipLevel
...
-            resultsEncoder.writeSmallInt(STORE_VERSION);
             try {
-                // Truncate existing output zip file
-                new ZipOutputStream(Files.newOutputStream(outputZipFile)).close();
-                try {
-                    outputZipFileSystem = FileSystems.newFileSystem(URI.create("jar:" + outputZipFile.toUri()), Collections.emptyMap());
-                } catch (FileSystemAlreadyExistsException e) {
-                    throw new InvalidUserCodeException("A previous file system exists for " + outputZipFile + ", which likely means that a previous test reporter was not closed", e);
-                }
+                resultsEncoder.writeSmallInt(STORE_VERSION);
```
ghale, line 122: ❓ Should the `STORE_VERSION` be incremented?  The format of the results file is now different right?

ghale (approved):
I left a couple of minor comments, but overall I think this looks good.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
octylFractal: This is a follow-up to #32672 which is unreleased, so I don't see a reason to bump it.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
octylFractal: I guess that makes more sense.

Issue #35427, opened 2025-10-22
Parallel model fetching with IP requires an IDE toggle


alllex:
This happens at least in the IntelliJ IDEA and we should check if the same happens for Android Studio.

This means that it's not enough to enable IP in the build to get its full benefits. Users need to know about and additionally enable the toggle in settings.

<img width="2188" height="1668" alt="Image" src="https://github.com/user-attachments/assets/94c19c8e-03ec-4161-87ce-cdba7ba7af5e" />

eskatos:
Android Studio has parallel model fetching enabled by default.

It would be much nicer if IntelliJ IDEA supported a way to set this in the source repository. That way developers working on the repo wouldn't have to tweak their IntelliJ IDEA configuration themselves once it is set in the repository.

alllex:
Related issue:
- https://github.com/gradle/gradle/issues/26506

alllex:
We've prepared a spec on the topic:
[Enabling parallel model fetching for IntelliJ - GBT Spec - Google Docs](https://docs.google.com/document/d/1DgDKp1UyYKRZ5psPIXOCnk55yngWvp6bSHSlIvjBVKM/edit?tab=t.0#heading=h.wbxzemi759rh)

alllex:
The mechanism on the Gradle side has been implemented:
- https://github.com/gradle/gradle/pull/35948

Let's keep this open until there is a version of IDEA that uses that mechanism

alllex:
The latest available IntelliJ IDEA 2026.1.1 **Preview** (build 261.23567.28) now enables parallel model fetching transparently 

- when IP is enabled

alllex:
This has been delivered as part of IntelliJ IDEA 2026.1.1 and can be observed when using Gradle 9.4.0+

PR #35515, opened 2025-10-30
Enable parallel model fetching for IDE sync for gradle/gradle

alllex:
For some time, IntelliiJ has provided an option to run tooling model builders in parallel. This is a setting that can be enabled in project settings (in the UI):
```
Settings > Build Tools > Gradle > 'Enable parallel Gradle model fetching for Gradle 7.4+'
```

This option has an observable effect when Parallel Execution is enabled in gradle.properties, which we already do in gradle/gradle. It also applies when Isolated Projects are enabled, but we are not there yet.

In any case, enabling this option can speed up sync time of gradle/gradle in the IDE. However, this option has to be enabled manually and for every checkout or worktree, where you work on gradle/gradle.

This PR checks into git an idea configuration file `.idea/gradle.xml`, which contains the setting, such that it is automatically enabled, whenever you work on Gradle.

---

So the change provides a measurable performance benefit. However, there is a potential downside.

The `gradle.xml` contains project-specific structure and settings. This means, the project structure would need to be updated, whenever new projects are added, removed or renamed. In practice, this means that whenever Gradle developers change project structure in their PR, they'd need to sync the project in the IDE and the commit changes to the `gradle.xml` file.

### DIFF
--- a/.idea/gradle.xml
+++ b/.idea/gradle.xml
@@ -306,5 +306,6 @@
         </option>
       </GradleProjectSettings>
     </option>
+    <option name="parallelModelFetch" value="true" />
   </component>
 </project>


alllex:
We will be able to avoid capturing the parallelism UI setting in a checked-in file once IntelliJ IDEA implements support for
- https://github.com/gradle/gradle/pull/35948

An issue to track this on our side is this:
- https://github.com/gradle/gradle/issues/35427

PR #35517, opened 2025-10-30
Capture ReportEntry from JUnit platform 

big-guy:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
+++ b/platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
@@ -218,8 +218,7 @@ public class JUnitPlatformTestExecutionListener implements TestExecutionListener
     }
 
     private void reportSkipped(TestIdentifier testIdentifier) {
-        assert currentTestPlan != null;
-
+        Objects.requireNonNull(currentTestPlan);
         currentTestPlan.getChildren(testIdentifier).stream()
             .filter(child -> !wasStarted(child))
             .forEach(this::executionSkipped);
@@ -352,7 +351,7 @@ public class JUnitPlatformTestExecutionListener implements TestExecutionListener
 
     @SuppressWarnings("deprecation")
     private Optional<TestIdentifier> getParent(TestIdentifier testIdentifier) {
-        assert currentTestPlan != null;
+        Objects.requireNonNull(currentTestPlan);
         try {
             return testIdentifier.getParentIdObject().map(currentTestPlan::getTestIdentifier);
         // Some versions of the JDK throw a BootstrapMethodError
@@ -400,7 +399,7 @@ public class JUnitPlatformTestExecutionListener implements TestExecutionListener
     }
 
     private boolean hasDifferentSourceThanAncestor(TestIdentifier testIdentifier) {
-        assert currentTestPlan != null;
+        Objects.requireNonNull(currentTestPlan);
 
         Optional<TestIdentifier> parent = currentTestPlan.getParent(testIdentifier);
         while (parent.isPresent()) {


platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -116,6 +122,26 @@ public JUnitPlatformTestExecutionListener(TestResultProcessor resultProcessor, C
...
+        // JUnit Platform will emit ReportEntry before a test starts if the ReportEntry is published from the class constructor.
+        if (wasStarted(testIdentifier)) {
+            resultProcessor.report(getId(testIdentifier), new DefaultTestMetadataEvent(clock.getCurrentTime(), Cast.uncheckedNonnullCast(entry.getKeyValuePairs())));
+        } else {
+            // The test has not started yet, so see if we can find a close ancestor and associate the ReportEntry with it
+            Object closestStartedAncestor = getIdOfClosestStartedAncestor(testIdentifier);
+            if (closestStartedAncestor != null) {
+                resultProcessor.report(closestStartedAncestor, new DefaultTestMetadataEvent(clock.getCurrentTime(), Cast.uncheckedNonnullCast(entry.getKeyValuePairs())));
+            }
+            // otherwise, we don't know what to associate this ReportEntry with
```
jvandort, line 139: Might be a good place for a debug log

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/JUnitTestMetadataCrossVersionSpec.groovy
```diff
@@ -0,0 +1,118 @@
...
+        then:
+        testEvents {
+            task(":test") {
+                nested("Gradle Test Run :test") {
+                    nested("Gradle Test Executor") {
+                        nested("Test class com.example.ReportEntryTest") {
+                            metadata("constructor", "value")
+                            test("Test test(TestReporter)(com.example.ReportEntryTest)") {
+                                output("Hello")
+                                metadata("beforeEach", "value")
```
jvandort: Can we emit different values for each of these (value1, value2)?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r812/CustomTestEventsCrossVersionSpec.groovy
```diff
@@ -211,14 +209,12 @@ class CustomTestEventsCrossVersionSpec extends ToolingApiSpecification implement
...
         testEvents {
             task(":customTest") {
-                root("Custom test root") {
-                    composite("My Suite") {
-                        composite("myTestMethod") {
-                            test("myTestMethod[0]") {
-                                displayName "My test method! (foo=0)"
+                nested("Test suite 'Custom test root'") {
+                    nested("Test class My Suite") {
+                        nested("Test class myTestMethod") {
```
jvandort, line 212: Odd we consider this method a class here.

platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -116,6 +122,26 @@ public JUnitPlatformTestExecutionListener(TestResultProcessor resultProcessor, C
...
+        }
+    }
+
+    @Override
+    public void fileEntryPublished(TestIdentifier testIdentifier, FileEntry file) {
+        // TODO: Capture this as well
```
jvandort, line 146: Is there an issue to track this that we can link here?

big-guy:
@bot-gradle test this

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestResultProcessor.java
```diff
@@ -48,4 +49,7 @@ public interface TestResultProcessor {
      */
     @UsedByScanPlugin("test-distribution")
     void failure(Object testId, TestFailure result);
+
+    default void report(Object testId, TestMetadataEvent keyValueEvent) {}
```
jvandort: I assume this is default since the test distribution plugin implements this class? If so, how do existing implementations of this class behave using this default implementation?

Can we add javadoc to this method?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/TestEventsSpec.java
```diff
@@ -0,0 +1,64 @@
...
+interface GroupTestEventSpec extends TestEventSpec {
+    /**
+     * The name of the group test event.
+     */
+    void nested(String name, @DelegatesTo(value = GroupTestEventSpec.class, strategy = Closure.DELEGATE_FIRST) Closure<?> spec);
+
+    /**
+     * The name of the test in the test event.
+     */
+    void test(String name, @DelegatesTo(value = TestEventSpec.class, strategy = Closure.DELEGATE_FIRST) Closure<?> spec);
```
jvandort, line 41: Might be useful to set a default value for the `spec` param to `= {}`. I see a few tests that call this method with an empty closure, which looks kinda weird

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r812/CustomTestEventsCrossVersionSpec.groovy
big-guy: I think this is from an assumption deep down that anything that's nested (and isn't the root) is a class

platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
big-guy: This is the next thing I'm doing (and have to do)

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestResultProcessor.java
big-guy: For TD/retry, we lose this information for now.

big-guy:
@bot-gradle test this

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/text/AbstractLineChoppingStyledTextOutput.java
```diff
@@ -31,7 +31,11 @@ public abstract class AbstractLineChoppingStyledTextOutput extends AbstractStyle
     private State currentState = INITIAL_STATE;
 
     protected AbstractLineChoppingStyledTextOutput() {
-        eol = SystemProperties.getInstance().getLineSeparator();
+        this(SystemProperties.getInstance().getLineSeparator());
```
tresat, line 34: Would all these line separator changes be better in a separate PR that could be merged right away?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r812/CustomTestEventsCrossVersionSpec.groovy
tresat: Can we fix this assumption (that this method is a class)?  It's only going to get more wrong.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestResultProcessor.java
```diff
@@ -48,4 +49,10 @@ public interface TestResultProcessor {
      */
     @UsedByScanPlugin("test-distribution")
     void failure(Object testId, TestFailure result);
+
+    /**
+     * Notifies this processor that metadata has been emitted for the given test.
+     */
+    default void report(Object testId, TestMetadataEvent keyValueEvent) {}
```
tresat: I think this name and comment is misleading.  The first 2 things "report" suggests to me are:

- A verb, as in "report on this test result"
- A noun, as in "the test report produced for this test"
- The other methods here are things that happened.  The test failed/completed/etc.  This would mean "The test reported", which again makes me think of the testing report produced at the end.

Also, I'm not sure it's fair to say this is only for metadata.  It could be used to transmit test params/results/etc.

How about calling this recorded? published? emitted? data?  Or even just entry?

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ProgressEvents.groovy
```diff
@@ -160,7 +160,9 @@ class ProgressEvents implements ProgressListener {
                 assert event.result.startTime == startEvent.eventTime
                 assert event.result.endTime == event.eventTime
             } else if (event instanceof TestOutputEvent || event instanceof TestMetadataEvent) {
-                // Do nothing - these are treated differently, their contents are stored separately from the events
+                assert running.containsKey(event.descriptor.parent)
+                def operation = operations.find { it.descriptor == event.descriptor.parent }
+                operation.statusEvents << new OperationStatus(event)
```
tresat, line 165: Is this logic also needed for TestOutputEvents?  Shouldn't it just be for TestMetadataEvents?

platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -116,6 +125,27 @@ public JUnitPlatformTestExecutionListener(TestResultProcessor resultProcessor, C
...
+        if (wasStarted(testIdentifier)) {
+            resultProcessor.report(getId(testIdentifier), new DefaultTestMetadataEvent(clock.getCurrentTime(), Cast.uncheckedNonnullCast(entry.getKeyValuePairs())));
+        } else {
+            // The test has not started yet, so see if we can find a close ancestor and associate the ReportEntry with it
+            Object closestStartedAncestor = getIdOfClosestStartedAncestor(testIdentifier);
+            if (closestStartedAncestor != null) {
+                resultProcessor.report(closestStartedAncestor, new DefaultTestMetadataEvent(clock.getCurrentTime(), Cast.uncheckedNonnullCast(entry.getKeyValuePairs())));
+            }
+            // otherwise, we don't know what to associate this ReportEntry with
+            LOGGER.debug("report entry published for unknown test identifier {}", testIdentifier);
```
tresat, line 140: 🤔 Should this be a failure instead?  At this point I think something has gone wrong with how we track the running test, so other things might be failing/fail too, so might as well bail out now.

platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -116,6 +125,27 @@ public JUnitPlatformTestExecutionListener(TestResultProcessor resultProcessor, C
         this.idGenerator = idGenerator;
     }
 
+    @Override
+    public void reportingEntryPublished(TestIdentifier testIdentifier, ReportEntry entry) {
```
tresat, line 129: 💭 This is a better name, we should use this for the TestExecutionListener as well.

platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -361,6 +400,8 @@ private static Optional<MethodSource> getMethodSource(TestIdentifier testIdentif
     }
 
     private boolean hasDifferentSourceThanAncestor(TestIdentifier testIdentifier) {
+        assert currentTestPlan != null;
```
tresat: 💭 These assertions are a lot of noise.  Can't we be confident that a test plan has been started by either us (dryrun) or JUnit already and drop these?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/JUnitTestMetadataCrossVersionSpec.groovy
```diff
@@ -0,0 +1,118 @@
...
+                suites {
+                    test {
+                        useJUnitJupiter()
+                    }
+                }
+            }
+        """
+    }
+
+    def "receives test metadata from JUnit platform tests"() {
```
tresat, line 55: 💭 The more I see it and read in the JUnit docs, the more I don't think metadata is the word for this.

"Publish the supplied map of key-value pairs as a report entry."

It's not data "about the test", it's anything the test chooses to publish.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/junit/platform/JUnitPlatformReportEntryIntegrationTest.groovy
```diff
@@ -0,0 +1,130 @@
...
+                public void afterEach(TestReporter testReporter) {
+                    testReporter.publishEntry("afterEach", "value4");
+                }
+            }
+        """
+        when:
+        succeeds("test")
+        then:
+        def results = resultsFor(testDirectory)
+        results.testPath('com.example.ReportEntryTest').onlyRoot().assertMetadata([constructor: "value1"]).assertChildrenExecuted("test(TestReporter)")
```
tresat, line 76: 💭 I think we should have a different assertion method for `ReportEntry` to distinguish it from the metadata asserted as in `CustomTestTaskHTMLTestReportMetadataTest`.  We should make it clear in the assertion which functionality we're checking.

tresat:
We should have tests using `@ReportEntry` annotation on a `@Test` as well.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/TestResultProcessor.java
big-guy: +1 published, that's what JUnit calls it (reportEntryPublished)

platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
big-guy: I don't think we want to be strict here because it's probably better to get _some_ signal from the test and being able to continue to work with the build vs being stuck until Gradle properly understands how to associate the ReportEntry to the right test.

platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
big-guy: Since currentTestPlan is nullable and changes from set to null, this is trading noise to resolve IDE inspections

octylFractal: Tangential, but it's recommended to use `Objects.requireNonNull`/`Preconditions` over `assert`, to ensure the check is actually performed even if currently superficial: https://github.com/gradle/gradle/blob/master/architecture/standards/0008-use-nullaway.md#decision

big-guy:
> We should have tests using `@ReportEntry` annotation on a `@Test` as well.

I don't think this made it into JUnit 5 itself. I see a post about it for a set of JUnit extensions (https://junit-pioneer.org/docs/report-entries/), but there is no ReportEntry annotation in JUnit Platform itself.

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ProgressEvents.groovy
big-guy: It's both because we assert that we see certain test output too in the test fixture.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/junit/platform/JUnitPlatformReportEntryIntegrationTest.groovy
big-guy: I don't think we should think about ReportEntry as a different kind of thing. We map ReportEntry into what we call "metadata" for tests, but we could have called metadata something else (e.g., "extra data").  Our test fixtures should be based on our generic representation (metadata) and not the underlying testing framework.

If TestNG or XCTest or something else supported this kind of information, we'd include that as "metadata" too.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r812/CustomTestEventsCrossVersionSpec.groovy
big-guy: I'll try to find where this is happening

PR #35528, opened 2025-10-31
Capture test metadata in XML reports 

big-guy:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestClassResult.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestClassResult.java
@@ -16,27 +16,24 @@
 
 package org.gradle.api.internal.tasks.testing.junit.result;
 
+import org.gradle.api.internal.tasks.testing.results.serializable.SerializedMetadata;
 import org.gradle.api.tasks.testing.TestResult;
 import org.gradle.internal.time.Clock;
-import org.jspecify.annotations.Nullable;
 
 import java.util.ArrayList;
 import java.util.List;
 
 public class TestClassResult {
-    private final List<TestMethodResult> methodResults = new ArrayList<TestMethodResult>();
+    private final List<TestMethodResult> methodResults;
     private final String className;
     private final String classDisplayName;
+    private final List<SerializedMetadata> metadatas;
     private long startTime;
     private int failuresCount;
     private int skippedCount;
     private long id;
 
-    public TestClassResult(long id, String className, long startTime) {
-        this(id, className, null, startTime);
-    }
-
-    public TestClassResult(long id, String className, @Nullable String classDisplayName, long startTime) {
+    public TestClassResult(long id, String className, String classDisplayName, long startTime, List<SerializedMetadata> metadatas) {
         if (id < 1) {
             throw new IllegalArgumentException("id must be > 0");
         }
@@ -44,6 +41,8 @@ public class TestClassResult {
         this.className = className;
         this.startTime = startTime;
         this.classDisplayName = classDisplayName == null ? className : classDisplayName;
+        this.metadatas = metadatas;
+        this.methodResults = new ArrayList<>();
     }
 
     public long getId() {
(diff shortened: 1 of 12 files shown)


big-guy:
@bot-gradle test this

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestClassResult.java
```diff
@@ -117,4 +116,8 @@ private boolean hasDefaultDisplayName() {
         // so we use this as a heuristic to determine whether the display name was customized
         return className.endsWith("." + classDisplayName) || className.endsWith("$" + classDisplayName);
     }
+
+    public List<SerializedMetadata> getMetadatas() {
+        return metadatas;
```
tresat, line 121: 💭 We should return a defensive copy here.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestMethodResult.java
```diff
@@ -109,4 +93,8 @@ public long getDuration() {
     public long getEndTime() {
         return endTime;
     }
+
+    public List<SerializedMetadata> getMetadatas() {
+        return metadatas;
```
tresat, line 98: 💭 Return a copy here?  Also update `getFailures()`?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestClassResult.java
```diff
@@ -16,34 +16,33 @@
...
     private int failuresCount;
     private int skippedCount;
     private long id;
 
-    public TestClassResult(long id, String className, long startTime) {
-        this(id, className, null, startTime);
-    }
-
-    public TestClassResult(long id, String className, @Nullable String classDisplayName, long startTime) {
+    public TestClassResult(long id, String className, String classDisplayName, long startTime, List<SerializedMetadata> metadatas) {
```
tresat, line 36: 🤔 We don't use metadatas in the vast majority of tests.  We should probably have a forwarding constructor that supplies an empty list, to avoid changing so many unrelated tetss.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestClassResult.java
big-guy: These structures are all backed by an immutable Guava list, do you think these should be defensively copied?

big-guy:
@bot-gradle test this

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestClassResult.java
tresat: They're only backed by whatever is supplied to the constructor.  It's currently an immutable list, but that's not required.

testing/internal-testing/src/main/groovy/org/gradle/integtests/fixtures/TestClassExecutionResult.java
```diff
@@ -37,6 +38,23 @@ public interface TestClassExecutionResult {
 
     TestClassExecutionResult assertTestCount(int tests, int failures);
 
+    /**
+     * Asserts that the provided metadata appears in the result file at the class level
+     */
+    default TestClassExecutionResult assertMetadata(Map<String, String> props) {
+        assert false;
```
octylFractal, line 45: These should be using JUnit asserts, unless you mean to fill them in before merging. As-is, they use Java `assert` which is only present with `-ea`, which I don't know if it's set for our tests.

tresat: I remember making this same point on another PR a long time ago.  They are always enabled in our testing, and used pretty often, so it's not a big deal.  I assume these are temporary placeholders though.

tresat (approved):

testing/internal-testing/src/main/groovy/org/gradle/integtests/fixtures/TestClassExecutionResult.java
big-guy: Yeah, we always have -ea turned on for the test workers. 

My intention was to take a pass through the test fixtures separately because TestClassExecutionResult and GenericTestExecutionResult are doing very similar things and we should only have one now that Test is backed by TestEventReporter.

PR #35534, opened 2025-10-31
Capture FileEntry from JUnit Platform

big-guy:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things
This reverts commit d84d14c.

### DIFF
--- a/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/DefaultTestEventSpec.groovy
+++ b/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/DefaultTestEventSpec.groovy
@@ -73,12 +73,12 @@ class DefaultTestEventSpec implements GroupTestEventSpec {
     }
 
     @Override
-    void metadata(String key, Object value) {
-        assert actualMetadata.remove(Collections.singletonMap(key, value)): "expected to find $key -> $value in $self"
+    void metadata(String key, String value) {
+        assert actualMetadata.remove(Collections.singletonMap(key, value)): "expected to find $key -> $value in $self. Available: $actualMetadata"
     }
 
     @Override
-    void metadata(Map<String, Object> values) {
+    void metadata(Map<String, String> values) {
         assert actualMetadata.remove(values): "expected to find $values in $self"
     }
 
--- a/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/TestEventsSpec.java
+++ b/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/TestEventsSpec.java
@@ -62,12 +62,12 @@ interface TestEventSpec {
     /**
      * Expect some metadata with a singleton key-value
      */
-    void metadata(String key, Object value);
+    void metadata(String key, String value);
 
     /**
      * Expect some metadata with multiple key-values
      */
-    void metadata(Map<String, Object> values);
+    void metadata(Map<String, String> values);
 
     /**
      * Expect a file attachment
--- a/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r813/CustomTestMetadataEventsCrossVersionSpec.groovy
+++ b/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r813/CustomTestMetadataEventsCrossVersionSpec.groovy
@@ -219,6 +219,64 @@ class CustomTestMetadataEventsCrossVersionSpec extends ToolingApiSpecification i
         }
     }
 
+
+    @TargetGradleVersion(">=8.13 <9.4.0")
+    def "reports custom test events with non-String key-values"() {
+        given:
+        buildFile("""
+            import java.time.Instant
+
+            abstract class CustomTestTask extends DefaultTask {
+                @Inject
+                abstract TestEventReporterFactory getTestEventReporterFactory()
+
+                @Inject
+                abstract ProjectLayout getLayout()
+
+                @TaskAction
+                void runTests() {
+                    try (def reporter = testEventReporterFactory.createTestEventReporter(
+                        "Custom test root",
+                        getLayout().getBuildDirectory().dir("test-results/Custom test root").get(),
+                        getLayout().getBuildDirectory().dir("reports/tests/Custom test root").get()
+                    )) {
+                        reporter.started(Instant.now())
+                        try (def myTest = reporter.reportTest("MyTestInternal", "My test!")) {
+                            myTest.started(Instant.now())
+                            myTest.metadata(Instant.now(), "mykey1", Collections.singletonList("value"))
+                            myTest.metadata(Instant.now(), "mykey2", 10)
+                            myTest.succeeded(Instant.now())
+                        }
+                        reporter.succeeded(Instant.now())
+                    }
+                }
+            }
+
+            tasks.register("customTest", CustomTestTask)
+        """)
+
+        when:
+        withConnection {
+            ProjectConnection connection ->
+                connection.newBuild()
+                    .addProgressListener(events, OperationType.TASK, OperationType.TEST, OperationType.TEST_METADATA)
+                    .forTasks('customTest')
+                    .run()
+        }
+
+        then:
+        testEvents {
+            task(":customTest") {
+                nested("Test suite 'Custom test root'") {
+                    test(if813OrOlderTestDisplayName()) {
+                        metadata("mykey1", "[value]")
+                        metadata("mykey2", "10")
+                    }
+                }
+            }
+        }
+    }
+
     def "reports custom test events at root, group and nested group levels"() {
         given:
         buildFile("""


big-guy:
@bot-gradle test this

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/ClientForwardingTestMetadataOperationListener.java
```diff
@@ -57,11 +61,24 @@ public void progress(OperationIdentifier buildOperationId, OperationProgressEven
...
+            }
         }
     }
 
     @Override
     public void finished(BuildOperationDescriptor buildOperation, OperationFinishEvent finishEvent) {
     }
+
+
 }
```
donat:
```suggestion
}
```

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/internal/DefaultTestMetadataEvent.java
```diff
@@ -61,4 +65,9 @@ public String toString() {
     public Map<String, Object> getValues() {
         return values;
     }
+
+    @Override
+    public <T> T get(Class<T> viewType) {
+        return new ProtocolToModelAdapter().adapt(viewType, payload);
```
donat: ❌  I think we should not use proxy objects for types that may be known on the daemon and on the TAPI consumer sides. Instead, we should add FileEntry-specific classes to the protocol and use it in the TAPI consumer object creation. 

The API looks nice and extensible, but we should avoid using proxies when possible. They are the source of may problems we faced with the Tooling API (e.g. performance and memory leaks).

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestFileAttachmentDataEvent.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+import java.nio.file.Path;
+
+@NullMarked
+public class DefaultTestFileAttachmentDataEvent extends AbstractTestDataEvent {
+    private final Path path;
+    private final @Nullable String mediaType;
```
donat, line 33: Instead of nullable, can we provide a sane type by default, for example `application/octet-stream`?

donat (changes requested):
I have a proposal for avoiding using proxies (and a couple of  💅), otherwise lgtm.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/FileAttachment.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * A file attachment was published by an executed test.
+ *
+ * @since 9.4.0
+ */
+@Incubating
+@NullMarked
+public interface FileAttachment {
+    /**
+     * Path to the published file attachment.
```
tresat: ❓ Is this meant to be a `File` or a `Path`?  If a path, relative to what, or absolute?

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/FileAttachment.java
```diff
@@ -0,0 +1,51 @@
...
+     *
+     * @return path to the file
+     * @since 9.4.0
+     */
+    File getPath();
+
+    /**
+     * <a href="https://www.iana.org/assignments/media-types/media-types.xhtml">Media type</a> of the file.
+     *
+     * @return media type of the file attachment or {@code null} if the file attachment represents a directory
```
tresat: ❓ JUnit [says](https://docs.junit.org/5.13.0/api/org.junit.platform.engine/org/junit/platform/engine/reporting/FileEntry.html) the `mediaType` may be `null`.  Is it only to support directories, or will they accept `null` on files, too?  If they accept `null` on files, then this might be wrong.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/consumer/parameters/BuildProgressListenerAdapter.java
```diff
@@ -635,7 +638,12 @@ private TestOutputEvent transformTestOutput(InternalTestOutputEvent event, Inter
 
     private TestMetadataEvent transformTestMetadata(InternalTestMetadataEvent event, InternalTestMetadataDescriptor descriptor) {
         OperationDescriptor clientDescriptor = addDescriptor(event.getDescriptor(), toDescriptor(descriptor));
-        return new DefaultTestMetadataEvent(event.getEventTime(), clientDescriptor, event.getValues());
+        return new DefaultTestMetadataEvent(event.getEventTime(), clientDescriptor, event.getValues(), null);
+    }
+
+    private TestMetadataEvent transformTestMetadata2(InternalTestMetadataEvent2 event, InternalTestMetadataDescriptor descriptor) {
```
tresat: 🤔 Is it okay to leave all these `2`s in here?  Should we call these `WithPayload` or something instead?

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/protocol/events/InternalTestMetadataEvent2.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.gradle.tooling.internal.protocol.events;
+
+/**
+ * DO NOT CHANGE THIS INTERFACE. It is part of the cross-version protocol.
+ *
+ * @since 9.4.0
+ */
+public interface InternalTestMetadataEvent2 extends InternalProgressEvent {
```
tresat: ❗ It seems wrong to be adding a type with a `2` suffix with a note saying it can't be changed.  Can we name it something better?  `WithPayload`?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestFileAttachmentDataEvent.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.gradle.api.internal.tasks.testing;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+import java.nio.file.Path;
+
+@NullMarked
+public class DefaultTestFileAttachmentDataEvent extends AbstractTestDataEvent {
```
tresat, line 31: 💅 Javadoc on this and `DefaultTestKeyValueDataEvent`.  Perhaps relating them to the JUnit types they will transport?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/JUnitXmlResultWriter.java
```diff
@@ -28,22 +30,26 @@
...
 
 public class JUnitXmlResultWriter {
 
+    private final Path rootDir;
     private final String hostName;
     private final TestResultsProvider testResultsProvider;
     private final JUnitXmlResultOptions options;
 
-    public JUnitXmlResultWriter(String hostName, TestResultsProvider testResultsProvider, JUnitXmlResultOptions options) {
+    public JUnitXmlResultWriter(Path rootDir, String hostName, TestResultsProvider testResultsProvider, JUnitXmlResultOptions options) {
```
tresat: 💅 rootDir = Root of what?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/JUnitXmlResultWriter.java
```diff
@@ -98,14 +106,19 @@ public void write(TestClassResult result, OutputStream output) {
...
-    private void writeOutputs(SimpleXmlWriter writer, long classId, boolean allClassOutput, TestOutputEvent.Destination destination) throws IOException {
-        writer.startCDATA();
-        if (allClassOutput) {
-            testResultsProvider.writeAllOutput(classId, destination, writer);
-        } else {
-            testResultsProvider.writeNonTestOutput(classId, destination, writer);
+    private void writeOutputs(SimpleXmlWriter writer, long classId, boolean includeRawOutput, boolean allClassOutput, TestOutputEvent.Destination destination, List<DefaultTestFileAttachmentDataEvent> fileAttachments) throws IOException {
+        boolean mayWrite = includeRawOutput || !fileAttachments.isEmpty();
+
+        if (mayWrite) {
```
tresat: ❗ This seems wrong.  File Attachments aren't output.  Yet if there is a File Attachment, we always write the output?  Seems like these 2 things should remain independent?

platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/junit/result/JUnitXmlResultWriterMergeRerunSpec.groovy
```diff
@@ -26,7 +26,7 @@ class JUnitXmlResultWriterMergeRerunSpec extends Specification {
     def outputPerTestCase = true
 
     protected JUnitXmlResultWriter getGenerator() {
-        new JUnitXmlResultWriter("localhost", provider, new JUnitXmlResultOptions(outputPerTestCase, true, true, true))
+        new JUnitXmlResultWriter(new File("").toPath(), "localhost", provider, new JUnitXmlResultOptions(outputPerTestCase, true, true, true))
```
tresat, line 29: 💭 Should this use a file created from `testDirectory` just in case?

subprojects/build-events/src/main/java/org/gradle/internal/build/event/types/DefaultFileAttachment.java
```diff
@@ -0,0 +1,51 @@
...
+public class DefaultFileAttachment implements Serializable {
+    private final File file;
+    private final @Nullable String mediaType;
+
+    public DefaultFileAttachment(Path path, @Nullable String mediaType) {
+        this.file = path.toFile();
+        this.mediaType = mediaType;
+    }
+
+    public File getPath() {
```
tresat: 💅 getFile()?

subprojects/build-events/src/main/java/org/gradle/internal/build/event/types/DefaultPayloadDataEvent.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.build.event.types;
+
+import org.gradle.tooling.internal.protocol.events.InternalTestMetadataDescriptor;
+import org.gradle.tooling.internal.protocol.events.InternalTestMetadataEvent2;
+
+public class DefaultPayloadDataEvent extends AbstractProgressEvent<InternalTestMetadataDescriptor> implements InternalTestMetadataEvent2 {
```
tresat: 💅 Javadoc?  I'd expect it to explain why is this implementing `2`.

subprojects/core/src/main/java/org/gradle/internal/html/SimpleHtmlWriter.java
```diff
@@ -52,6 +55,16 @@ public SimpleMarkupWriter startElement(String name) throws IOException {
         return super.startElement(name);
     }
 
+    /**
+     * Create a reference to another file that is relative to this HTML file.
```
tresat: 💅 This isn't accurate.

tresat (dismissed):

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestFileAttachmentDataEvent.java
big-guy: Originally, I was using `application/octet-stream` if we didn't know, but the underlying `FileEntry` allows for null:
https://docs.junit.org/5.13.0/api/org.junit.platform.engine/org/junit/platform/engine/reporting/FileEntry.html#from(java.nio.file.Path,java.lang.String)

JUnit Jupiter suggests using `application/octet-stream`:
https://docs.junit.org/5.13.0/api/org.junit.jupiter.api/org/junit/jupiter/api/TestReporter.html#publishFile(java.nio.file.Path,org.junit.jupiter.api.extension.MediaType)

But this only applies to regular files. It still uses null for directories.  

When I thought this was going to just be null for directories, I looked on IANA's list of registered media types to see if there was a suggestion for directories. There _is_ a `text/directory`: https://www.iana.org/assignments/media-types/media-types.xhtml#:~:text=directory%20%2D%20DEPRECATED%20by%20RFC6350

But this is deprecated and actually referred to a file that listed the contents of a directory (and not a directory itself).  Directories don't appear to have a sensible/conventional media type. 

I think that means mediaType needs to be nullable:
- JUnit Jupiter will usually provide a media type, except for directories.
- Other test engines on JUnit Platform can provide null for any reason

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/FileAttachment.java
big-guy: JUnit Jupiter appears to only support null for directories, but the underlying JUnit Platform allows it for any reason: see also https://github.com/gradle/gradle/pull/35534#discussion_r2542827620

I'll clarify this.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/FileAttachment.java
big-guy: This was originally a NIO Path, I'll use File since that's the type now

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/consumer/parameters/BuildProgressListenerAdapter.java
big-guy: I should probably follow the convention we use elsewhere: V2 or Version2 and not a plain `2`.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/protocol/events/InternalTestMetadataEvent2.java
big-guy: The convention is to use `Version#`:

InternalBasicProblemDetailsVersion2.java
InternalBasicProblemDetailsVersion3.java
InternalBasicProblemDetailsVersion4.java
InternalBuildActionVersion2.java
InternalBuildControllerVersion2.java
InternalProblemAggregationDetailsVersion3.java
InternalProblemAggregationVersion2.java
InternalProblemAggregationVersion3.java
InternalProblemDetailsVersion2.java
InternalProblemEventVersion2.java

I'll add some docs about what the difference between Version1 and Version2 are.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/JUnitXmlResultWriter.java
big-guy: This is one of the quirks of how this is crammed into the existing format.

Azure pipelines uses this: https://devblogs.microsoft.com/devops/junit-attachments-support-for-publish-test-results/
Gitlab seems to support it: https://docs.gitlab.com/ci/testing/unit_test_reports/#xml-structure-example
Jenkins is the originator of it: https://kohsuke.org/2012/03/13/attaching-files-to-junit-tests/

The alternative is to use `<properties>` as suggested here: https://github.com/testmoapp/junitxml#attachments-files

But I haven't seen any references to any tools that look there for attachments.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestFileAttachmentDataEvent.java
donat: Got it. Now, nullable makes sense to me.

donat (approved):
Lgtm

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/JUnitXmlResultWriter.java
tresat: I'm not saying we should put the attachments anywhere else in the XML.

I'm saying, if a successful test outputs "hello world" using an output event, and publishes a file entry, and does NOT opt-in to writing output, shouldn't the `<system-out>` element **only** contain the attachment, and **not** the "hello world"?

big-guy:
@bot-gradle test this

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/JUnitXmlResultWriter.java
big-guy: Ah, I understand. I'll add a test for that

big-guy:
@bot-gradle test this

big-guy:
@bot-gradle test this

big-guy:
@bot-gradle help

big-guy:
@bot-gradle test AllCrossVersionTests PullRequestFeedback

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/protocol/events/InternalTestMetadataEventVersion2.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * DO NOT CHANGE THIS INTERFACE. It is part of the cross-version protocol.
+ *
+ * This is a new message sent from Gradle 9.4.0 and newer with an arbitrary payload.
+ * Older clients (before 9.4.0) ignore this message.
+ *
+ * @since 9.4.0
+ */
+@NullMarked
+public interface InternalTestMetadataEventVersion2 extends InternalProgressEvent {
```
tresat, line 33: ❓ Should this be named `InternalFileAttachmentEvent` or something similar instead?

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/protocol/events/InternalTestMetadataEventVersion2.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+import java.io.File;
+
+/**
+ * DO NOT CHANGE THIS INTERFACE. It is part of the cross-version protocol.
+ *
+ * This is a new message sent from Gradle 9.4.0 and newer with an arbitrary payload.
```
tresat, line 27: ❗ This is incorrect - the payload here is fixed.

big-guy:
@bot-gradle test AllCrossVersionTests

big-guy:
@bot-gradle test ACT PRF

subprojects/core/src/main/java/org/gradle/internal/html/SimpleHtmlWriter.java
```diff
@@ -52,6 +55,19 @@ public SimpleMarkupWriter startElement(String name) throws IOException {
         return super.startElement(name);
     }
 
+    /**
+     * Construct a path to another file that is relative to the directory of this HTML file.
+     *
+     * This can be used to create links to other files without requiring absolute paths to be
+     * embedded in the HTML file.
+     */
+    public String relativeLink(Path otherFile) {
```
tresat, line 64: 💭 Should the Simple HTML writer be doing the link relativization?  That doesn't seem like its responsibility, vs. just getting relativized links to write.  Not a big deal if this is easier.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -385,111 +393,151 @@ protected void render(PerRootInfo info, SimpleHtmlWriter htmlWriter) throws IOEx
...
+            htmlWriter.endElement();
         }
 
-        private String formatLogTime(long logTime) {
-            Instant instant = Instant.ofEpochMilli(logTime);
-
-            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS z")
-                .withZone(ZoneId.systemDefault());
+        private static void renderLink(SimpleHtmlWriter htmlWriter, DefaultTestFileAttachmentDataEvent metadata) throws IOException {
+            htmlWriter.startElement("a").attribute("href", htmlWriter.relativeLink(metadata.getPath())).characters(htmlWriter.relativeLink(metadata.getPath())).endElement();
```
tresat, line 529: ❓ What do you think about rendering just the file name here rather than the relative path?  I wonder if the relative path from the report page would be meaningful at all.  Keep the links like this, but the text displayed in the report might be better as just the file name.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -385,111 +393,151 @@ protected void render(PerRootInfo info, SimpleHtmlWriter htmlWriter) throws IOEx
...
             for (int metadataIdx = 0; metadataIdx < metadatas.size(); metadataIdx++) {
-                TestMetadataEvent metadata = metadatas.get(metadataIdx);
-                renderFirstMetadataElement(metadata, metadataIdx, htmlWriter);
-                if (metadata.getValues().size() > 1) {
-                    Map<String, String> additionalEntries = metadata.getValues();
-                    renderAdditionalMetadataElements(additionalEntries, metadataIdx, htmlWriter);
+                DefaultTestKeyValueDataEvent metadata = metadatas.get(metadataIdx);
+                Map<String, String> elements = metadata.getValues();
+
+                htmlWriter.startElement("tr").attribute("class", metadataIdx % 2 == 0 ? "even" : "odd");
```
cobexer, line 438: ❓ Could we replace the code complexity of generating `even` and `odd` classes with using `table > tbody >tr:nth-child(even)` and `table > tbody >tr:nth-child(odd)` in the CSS file instead?

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:nth-child

cobexer: Benefits would be: 
* Reduced code complexity for generating the HTML
* Correct even/odd coloring of table rows even if tables get filtered with JavaScript (not sure how our tabs work TBH)

blindpirate:
@bot-gradle test AllCrossVersionTests

blindpirate:
@bot-gradle test AllCrossVersionTests

blindpirate:
@bot-gradle test AllCrossVersionTests

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
tresat: My worry is that 2 identical file names (on different test's pages) won't automatically be assumed to be the same file.  2 identical "shortened" relative URLs are probably far more likely to be confused for being the same file.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
tresat: Seems worth trying, maybe not now though.  I'll add a followup item to our list, thanks.

Issue #35342, opened 2025-10-16
Gradle sync fails with IP during settings with EmptyStackException


alllex:
When dogfooding IP in the gradle/gradle build and doing a sync in IDEA, the build sometimes immediately fails with
```
Settings file '/Users/asemin/projects/gradle/build-tool/quality6/settings.gradle.kts' line: 30

* What went wrong:
Error resolving plugin [id: 'gradlebuild.build-environment']
> A problem occurred configuring project ':build-logic-settings:build-environment'.
   > java.util.EmptyStackException (no error message)
```

No build scan is published, because the build fails too early.

---

The problem is reproducible in the Gradleception sync test
https://ge.gradle.org/s/t265objk5aceq/tests/task/:smoke-ide-test:smokeIdeTest/details/org.gradle.ide.sync.IsolatedProjectsGradleceptionSyncTest/can%20sync%20gradle%2Fgradle%20incrementally%20without%20error?focused-execution=1&page=eyJvdXRwdXQiOnsiMCI6NH19&top-execution=1#L738

Stacktrace:
```
 Caused by: java.util.EmptyStackException	
 	at org.gradle.internal.cc.impl.DefaultDynamicCallProblemReporting.leaveDynamicCall(DynamicCallProblemReporting.kt:74)	
 	at org.gradle.internal.cc.impl.DefaultBuildModelControllerServices$ConfigurationCacheIsolatedProjectsProvider$createDynamicCallProjectIsolationProblemReporting$1$2.accept(DefaultBuildModelControllerServices.kt:186)	
 	at org.gradle.configuration.internal.DefaultDynamicCallContextTracker.lambda$leaveDynamicCall$2(DefaultDynamicCallContextTracker.java:51)	
 	at org.gradle.configuration.internal.DefaultDynamicCallContextTracker.leaveDynamicCall(DefaultDynamicCallContextTracker.java:51)	
 	at org.gradle.internal.cc.impl.TrackingDynamicLookupRoutine.withDynamicCall(TrackingDynamicLookupRoutine.kt:76)	
 	at org.gradle.internal.cc.impl.TrackingDynamicLookupRoutine.property(TrackingDynamicLookupRoutine.kt:34)	
 	at org.gradle.api.internal.project.DefaultProject.property(DefaultProject.java:1152)	
 	at org.gradle.api.internal.project.DefaultProject.getProperty(DefaultProject.java:1129)	
[...]

PR #35535, opened 2025-11-01
Move `DynamicCallContextTracker` to `Build` scope

abstratt:
Issue: #35342

DynamicCallContextTracker keeps state (enter/leave listeners) that is relevant in the context of a build. We should either change it to manage said listeners separately per build, or we should make it a `Build` scope service. 

This change set does that.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


abstratt:
@bot-gradle test this

PR #35558, opened 2025-11-04
Extract HttpClient interface

jvandort:
Extract an interface from HttpClientHelper to allow us to use HTTP client implementations other than Apache http client. This will allow us to selectively use implementations such as the http client in the JVM standard library since Java 11, if we are running on 11+ The new HttpClient interface does not expose any types related to the apache http client.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

PR #35425, opened 2025-10-22
Refactor ResolveTestFixture for IP compatibility

jvandort:
We use this fixture in many places to verify the shape of a dependency graph.
There were a number of annoying things with this fixture, including not being IP compatible.

In this commit, we:
- Remove buildscript block from ResolveTestFixture, allowing it to be used alongside the plugins block with less hassle
- Update ResolveTestFixture to generate string containing fixture instead of injecting that string directly into buildscripts, allowing tests themselves to choose where in file the fixture lives
- Remove special logic in ResolveTestFixture to avoid building artifacts, instead updating the relevant tests to avoid task dependency cycles.
- Replace ResolveTestFixture.addDefaultVariantDerivationStrategy by applying jvm-ecosystem plugin in relevant tests
- Use ResolveTestFixture consistently across tests
- Remove suspicious PluginDslSupport fixture which functioned as a way to use the plugins block alonside the resolve test fixture. Now, the plugins block can be used without workarounds
- Enhance BuildTestFixture to make it easier to use the plugins block, by configuring group and version in settings instead of in build files.

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #35559, opened 2025-11-04
Address test failures from #35425

jvandort:
The first failure stems from the interaction between gradle.lifecycle.beforeProject and allprojects. We update the build using allprojects to instead write directly to the build files so it can override the beforeProject defaults The second is due to a removal of a mustRunAfter, affecting parallel tests. The test verified that only a single call to version listing is made, however when executed in parallel multiple calls are made. We add the dependency back to ensure only one task runs at once. Eventually we might want to make the actual implementation resilient to parallel execution, so one request blocks on the other

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
Tests failing on master currently should now pass:
https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_AllVersionsCrossVersion_10_bucket9/105056623
https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_Parallel_7_bucket1/105056625

PR #35569, opened 2025-11-05
Fix link on deprecation warning for TAPI

reinsch82:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/extensibility/test-kit/src/main/java/org/gradle/testkit/runner/internal/feature/BuildResultOutputFeatureCheck.java
+++ b/platforms/extensibility/test-kit/src/main/java/org/gradle/testkit/runner/internal/feature/BuildResultOutputFeatureCheck.java
@@ -20,6 +20,8 @@ import org.gradle.internal.deprecation.DeprecationLogger;
 import org.gradle.testkit.runner.UnsupportedFeatureException;
 import org.gradle.util.GradleVersion;
 
+import java.util.function.Function;
+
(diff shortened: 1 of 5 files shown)


platforms/extensibility/test-kit/src/main/java/org/gradle/testkit/runner/internal/feature/BuildResultOutputFeatureCheck.java
```diff
@@ -35,15 +37,22 @@ public BuildResultOutputFeatureCheck(GradleVersion targetGradleVersion, boolean
...
-                    .willBecomeAnErrorInGradle10()
-                    .withUserManual("third_party_integration", "sec:embedding_compatibility")
-                    .nagUser();
-            }
+            checkDeprecationWarning(targetGradleVersion, version -> String.format("Capturing build output in debug mode with the GradleRunner for the version of Gradle you are using (%s) is deprecated with TestKit.",
+                version));
+        }
+    }
+
+    public static void checkDeprecationWarning(GradleVersion targetGradleVersion, Function<String, String> messageSupplier) {
```
jbartok: Totally optional, but could we find a better name for these methods (instead of `checkDeprecationWarning`)? It sounds stupid; we are not "checking the correctness of a deprecation warning" or anything of the sort. I understand that that was never the intention, but that's how it sounds to me.

PR #35577, opened 2025-11-06
Tooling API support for non-class based testing

donat:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestExecutionBuildConfigurationAction.java
+++ b/platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestExecutionBuildConfigurationAction.java
@@ -20,6 +20,8 @@ import org.gradle.api.Task;
 import org.gradle.api.internal.GradleInternal;
 import org.gradle.api.internal.project.ProjectState;
 import org.gradle.api.internal.tasks.testing.filter.DefaultTestFilter;
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
 import org.gradle.api.specs.Specs;
 import org.gradle.api.tasks.testing.AbstractTestTask;
 import org.gradle.api.tasks.testing.Test;
@@ -33,11 +35,16 @@ import org.gradle.execution.plan.QueryableExecutionPlan;
 import org.gradle.internal.build.event.types.DefaultTestDescriptor;
 import org.gradle.process.internal.DefaultJavaDebugOptions;
 import org.gradle.tooling.internal.protocol.events.InternalJvmTestDescriptor;
+import org.gradle.tooling.internal.protocol.events.InternalSourceAwareTestDescriptor;
 import org.gradle.tooling.internal.protocol.events.InternalTestDescriptor;
 import org.gradle.tooling.internal.protocol.test.InternalDebugOptions;
 import org.gradle.tooling.internal.protocol.test.InternalJvmTestRequest;
 import org.gradle.tooling.internal.protocol.test.InternalTaskSpec;
 import org.gradle.tooling.internal.protocol.test.InternalTestSpec;
+import org.gradle.tooling.internal.protocol.test.source.InternalClassSource;
+import org.gradle.tooling.internal.protocol.test.source.InternalFilesystemSource;
+import org.gradle.tooling.internal.protocol.test.source.InternalMethodSource;
+import org.gradle.tooling.internal.protocol.test.source.InternalTestSource;
 import org.gradle.tooling.internal.provider.action.TestExecutionRequestAction;
 import org.jspecify.annotations.NullMarked;
 
@@ -50,6 +57,9 @@ import java.util.function.Consumer;
 
 @NullMarked
 class TestExecutionBuildConfigurationAction implements EntryTaskSelector {
+
+    private static final Logger LOG = Logging.getLogger(TestExecutionBuildConfigurationAction.class);
+
     private final TestExecutionRequestAction testExecutionRequest;
 
     public TestExecutionBuildConfigurationAction(TestExecutionRequestAction testExecutionRequest) {
@@ -167,6 +177,7 @@ class TestExecutionBuildConfigurationAction implements EntryTaskSelector {
 
     private void configureTestTasksInBuild(Context context) {
         final Collection<InternalTestDescriptor> testDescriptors = testExecutionRequest.getTestExecutionDescriptors();
+        warnIfUnsupportedTestRerunningForResourceBasedTests(testDescriptors);
         for (final InternalTestDescriptor descriptor : testDescriptors) {
             final String testTaskPath = taskPathOf(descriptor);
             for (AbstractTestTask testTask : queryTestTasks(context, testTaskPath)) {
@@ -174,6 +185,22 @@ class TestExecutionBuildConfigurationAction implements EntryTaskSelector {
                 for (InternalTestDescriptor testDescriptor : testDescriptors) {
                     if (taskPathOf(testDescriptor).equals(testTaskPath)) {
                         includeTestMatching((InternalJvmTestDescriptor) testDescriptor, testTask);
+                    }
+                }
+            }
+        }
+    }
+
+    private static void warnIfUnsupportedTestRerunningForResourceBasedTests(Collection<InternalTestDescriptor> testDescriptors) {
+        Set<String> seenTasks = new LinkedHashSet<>();
+        for (InternalTestDescriptor descriptor : testDescriptors) {
+            if (descriptor instanceof InternalSourceAwareTestDescriptor) {
+                InternalSourceAwareTestDescriptor sd  = (InternalSourceAwareTestDescriptor) descriptor;
+                if (sd.getSource() instanceof InternalFilesystemSource) {
+                    String taskPath = taskPathOf(descriptor);
+                    if (!seenTasks.contains(taskPath)) {
+                        LOG.warn("Re-running resource-based tests is not supported via TestLauncher API. The '{}' task will be scheduled without further filtering.", taskPath);
+                        seenTasks.add(taskPath);
                     }
                 }
             }
@@ -199,10 +226,23 @@ class TestExecutionBuildConfigurationAction implements EntryTaskSelector {
     private static void includeTestMatching(InternalJvmTestDescriptor descriptor, AbstractTestTask testTask) {
         String className = descriptor.getClassName();
         String methodName = descriptor.getMethodName();
-        if (className == null && methodName == null) {
-            testTask.getFilter().includeTestsMatching("*");
+        // for resource-based tests, don't apply any class-based filtering
+        if (isClassBasedTestDescriptor(descriptor)) {
+            if (className == null && methodName == null) {
+                testTask.getFilter().includeTestsMatching("*");
+            } else {
+                testTask.getFilter().includeTest(className, methodName);
+            }
+        }
+    }
+
+    private static boolean isClassBasedTestDescriptor(InternalJvmTestDescriptor descriptor) {
+        if (descriptor instanceof InternalSourceAwareTestDescriptor) {
+            InternalTestSource source = ((InternalSourceAwareTestDescriptor) descriptor).getSource();
+            return source instanceof InternalClassSource || source instanceof InternalMethodSource;
         } else {
-            testTask.getFilter().includeTest(className, methodName);
+            // assume class-based when no source information is available
+            return true;
         }
     }
 
--- a/platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestOperationMapper.java
+++ b/platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestOperationMapper.java
@@ -23,9 +23,18 @@ import org.gradle.api.internal.tasks.testing.operations.ExecuteTestBuildOperatio
 import org.gradle.api.tasks.testing.TestDescriptor;
 import org.gradle.api.tasks.testing.TestFailure;
 import org.gradle.api.tasks.testing.TestResult;
+import org.gradle.api.tasks.testing.source.ClassSource;
+import org.gradle.api.tasks.testing.source.ClasspathResourceSource;
+import org.gradle.api.tasks.testing.source.DirectorySource;
+import org.gradle.api.tasks.testing.source.FilePosition;
+import org.gradle.api.tasks.testing.source.FileSource;
+import org.gradle.api.tasks.testing.source.MethodSource;
+import org.gradle.api.tasks.testing.source.NoSource;
+import org.gradle.api.tasks.testing.source.TestSource;
 import org.gradle.internal.build.event.BuildEventSubscriptions;
 import org.gradle.internal.build.event.types.AbstractTestResult;
 import org.gradle.internal.build.event.types.DefaultFileComparisonTestAssertionFailure;
+import org.gradle.internal.build.event.types.DefaultFilePosition;
 import org.gradle.internal.build.event.types.DefaultTestAssertionFailure;
 import org.gradle.internal.build.event.types.DefaultTestDescriptor;
 import org.gradle.internal.build.event.types.DefaultTestFailureResult;
@@ -34,6 +43,13 @@ import org.gradle.internal.build.event.types.DefaultTestFrameworkFailure;
 import org.gradle.internal.build.event.types.DefaultTestSkippedResult;
 import org.gradle.internal.build.event.types.DefaultTestStartedProgressEvent;
 import org.gradle.internal.build.event.types.DefaultTestSuccessResult;
+import org.gradle.internal.build.event.types.test.source.DefaultClassSource;
+import org.gradle.internal.build.event.types.test.source.DefaultClasspathResourceSource;
+import org.gradle.internal.build.event.types.test.source.DefaultDirectorySource;
+import org.gradle.internal.build.event.types.test.source.DefaultFileSource;
+import org.gradle.internal.build.event.types.test.source.DefaultMethodSource;
+import org.gradle.internal.build.event.types.test.source.DefaultNoSource;
+import org.gradle.internal.build.event.types.test.source.DefaultOtherSource;
 import org.gradle.internal.operations.BuildOperationDescriptor;
 import org.gradle.internal.operations.OperationFinishEvent;
 import org.gradle.internal.operations.OperationIdentifier;
@@ -41,15 +57,18 @@ import org.gradle.internal.operations.OperationStartEvent;
 import org.gradle.tooling.events.OperationDescriptor;
 import org.gradle.tooling.events.OperationType;
 import org.gradle.tooling.internal.protocol.InternalFailure;
+import org.gradle.tooling.internal.protocol.events.InternalFilePosition;
 import org.gradle.tooling.internal.protocol.events.InternalJvmTestDescriptor;
 import org.gradle.tooling.internal.protocol.events.InternalOperationFinishedProgressEvent;
 import org.gradle.tooling.internal.protocol.events.InternalOperationStartedProgressEvent;
+import org.gradle.tooling.internal.protocol.events.InternalTestDescriptor;
+import org.gradle.tooling.internal.protocol.test.source.InternalTestSource;
 import org.jspecify.annotations.Nullable;
 
 import java.util.ArrayList;
 import java.util.List;
 
-class TestOperationMapper implements BuildOperationMapper<ExecuteTestBuildOperationType.Details, DefaultTestDescriptor> {
+class TestOperationMapper implements BuildOperationMapper<ExecuteTestBuildOperationType.Details, InternalTestDescriptor> {
     private final TaskForTestEventTracker taskTracker;
 
     TestOperationMapper(TaskForTestEventTracker taskTracker) {
@@ -72,26 +91,25 @@ class TestOperationMapper implements BuildOperationMapper<ExecuteTestBuildOperat
     }
 
     @Override
-    public DefaultTestDescriptor createDescriptor(ExecuteTestBuildOperationType.Details details, BuildOperationDescriptor buildOperation, @Nullable OperationIdentifier parent) {
+    public InternalTestDescriptor createDescriptor(ExecuteTestBuildOperationType.Details details, BuildOperationDescriptor buildOperation, @Nullable OperationIdentifier parent) {
         TestDescriptor testDescriptor = details.getTestDescriptor();
         return testDescriptor.isComposite() ? toTestDescriptorForSuite(buildOperation.getId(), parent, testDescriptor) : toTestDescriptorForTest(buildOperation.getId(), parent, testDescriptor);
     }
 
     @Override
-    public InternalOperationStartedProgressEvent createStartedEvent(DefaultTestDescriptor descriptor, ExecuteTestBuildOperationType.Details details, OperationStartEvent startEvent) {
+    public InternalOperationStartedProgressEvent createStartedEvent(InternalTestDescriptor descriptor, ExecuteTestBuildOperationType.Details details, OperationStartEvent startEvent) {
         return new DefaultTestStartedProgressEvent(details.getStartTime(), descriptor);
     }
 
     @Override
-    public InternalOperationFinishedProgressEvent createFinishedEvent(DefaultTestDescriptor descriptor, ExecuteTestBuildOperationType.Details details, OperationFinishEvent finishEvent) {
+    public InternalOperationFinishedProgressEvent createFinishedEvent(InternalTestDescriptor descriptor, ExecuteTestBuildOperationType.Details details, OperationFinishEvent finishEvent) {
         TestResult testResult = ((ExecuteTestBuildOperationType.Result) finishEvent.getResult()).getResult();
         return new DefaultTestFinishedProgressEvent(testResult.getEndTime(), descriptor, adapt(testResult));
     }
 
-    private DefaultTestDescriptor toTestDescriptorForSuite(OperationIdentifier buildOperationId, OperationIdentifier parentId, TestDescriptor suite) {
+    private InternalTestDescriptor toTestDescriptorForSuite(OperationIdentifier buildOperationId, OperationIdentifier parentId, TestDescriptor suite) {
         String methodName = null;
         String operationDisplayName = suite.toString();
-
         TestDescriptor originalDescriptor = getOriginalDescriptor(suite);
         if (originalDescriptor instanceof AbstractTestDescriptor) {
             methodName = ((AbstractTestDescriptor) originalDescriptor).getMethodName();
@@ -99,19 +117,61 @@ class TestOperationMapper implements BuildOperationMapper<ExecuteTestBuildOperat
         } else {
             operationDisplayName = getLegacyOperationDisplayName(operationDisplayName, originalDescriptor);
         }
-        return new DefaultTestDescriptor(buildOperationId, suite.getName(), operationDisplayName, suite.getDisplayName(), InternalJvmTestDescriptor.KIND_SUITE, suite.getName(), suite.getClassName(), methodName, parentId, taskTracker.getTaskPath(buildOperationId));
+        InternalTestSource testSource = toInternalTestSource(suite.getSource());
+        return new DefaultTestDescriptor(buildOperationId, suite.getName(), operationDisplayName, suite.getDisplayName(), InternalJvmTestDescriptor.KIND_SUITE, suite.getName(), suite.getClassName(), methodName, parentId, taskTracker.getTaskPath(buildOperationId), testSource);
     }
 
-    private DefaultTestDescriptor toTestDescriptorForTest(OperationIdentifier buildOperationId, OperationIdentifier parentId, TestDescriptor test) {
+    private InternalTestDescriptor toTestDescriptorForTest(OperationIdentifier buildOperationId, OperationIdentifier parentId, TestDescriptor test) {
         String operationDisplayName = test.toString();
-
         TestDescriptor originalDescriptor = getOriginalDescriptor(test);
         if (originalDescriptor instanceof AbstractTestDescriptor) {
             operationDisplayName = adjustOperationDisplayNameForIntelliJ(operationDisplayName, (AbstractTestDescriptor) originalDescriptor);
         } else {
             operationDisplayName = getLegacyOperationDisplayName(operationDisplayName, originalDescriptor);
         }
-        return new DefaultTestDescriptor(buildOperationId, test.getName(), operationDisplayName, test.getDisplayName(), InternalJvmTestDescriptor.KIND_ATOMIC, null, test.getClassName(), test.getName(), parentId, taskTracker.getTaskPath(buildOperationId));
+        InternalTestSource testSource = toInternalTestSource(test.getSource());
+        return new DefaultTestDescriptor(
+            buildOperationId,
+            test.getName(),
+            operationDisplayName,
+            test.getDisplayName(),
+            InternalJvmTestDescriptor.KIND_ATOMIC,
+            null,
+            test.getClassName(),
+            test.getName(),
+            parentId,
+            taskTracker.getTaskPath(buildOperationId),
+            testSource
+        );
+    }
+
+    private static InternalTestSource toInternalTestSource(TestSource source) {
+        if (source instanceof FileSource) {
+            FileSource fileSource = (FileSource) source;
+            return new DefaultFileSource(fileSource.getFile(), toFilePosition(fileSource.getPosition()));
+        } else if (source instanceof DirectorySource) {
+            return new DefaultDirectorySource(((DirectorySource) source).getFile());
+        } else if (source instanceof ClassSource) {
+            ClassSource classSource = (ClassSource) source;
+            return new DefaultClassSource(classSource.getClassName());
+        } else if (source instanceof MethodSource) {
+            MethodSource methodSource = (MethodSource) source;
+            return new DefaultMethodSource(methodSource.getClassName(), methodSource.getMethodName());
+        } else if (source instanceof ClasspathResourceSource) {
+            ClasspathResourceSource classpathResourceSource = (ClasspathResourceSource) source;
+            return new DefaultClasspathResourceSource(classpathResourceSource.getClasspathResourceName(), toFilePosition(classpathResourceSource.getPosition()));
+        } else if (source instanceof NoSource) {
+            return DefaultNoSource.getInstance();
+        } else {
+            return DefaultOtherSource.getInstance();
+        }
+    }
+
+    private static @Nullable InternalFilePosition toFilePosition(@Nullable FilePosition position) {
+        if (position == null) {
+            return null;
+        }
+        return new DefaultFilePosition(position.getLine(), position.getColumn());
     }
 
     /**
--- a/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/AbstractResourceBasedTestingCrossVersionTest.groovy
+++ b/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/AbstractResourceBasedTestingCrossVersionTest.groovy
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.integtests.tooling.r940
+
+import org.gradle.integtests.fixtures.TestResources
+import org.gradle.integtests.tooling.TestEventsFixture
+import org.gradle.integtests.tooling.fixture.ProgressEvents
+import org.gradle.integtests.tooling.fixture.TextUtil
+import org.gradle.integtests.tooling.fixture.ToolingApiSpecification
+import org.gradle.test.fixtures.file.TestDirectoryProvider
+import org.gradle.test.fixtures.file.TestNameTestDirectoryProvider
+import testengines.TestEnginesFixture
+
+abstract class AbstractResourceBasedTestingCrossVersionTest extends ToolingApiSpecification implements TestEventsFixture {
+
+    protected static final DEFAULT_DEFINITIONS_LOCATION = "src/test/definitions"
+    private static File engineJarLib
+
+    ProgressEvents events = ProgressEvents.create()
+
+    abstract List<TestEnginesFixture.TestEngines> getEnginesToSetup()
+
+    def setup() {
+        if (!engineJarLib) {
+            TestDirectoryProvider testClassDirectoryProvider = new TestNameTestDirectoryProvider.UniquePerTestClassDirectoryProvider(this.getClass())
+            TestResources resources = new TestResources(testClassDirectoryProvider, TestEnginesFixture.TestEngines.class, this.getClass())
+
+            testClassDirectoryProvider.getTestDirectory().getParentFile()
+
+            resources.maybeCopy("shared")
+            enginesToSetup.forEach {
+                resources.maybeCopy(it.name)
+            }
+
+            String engineCopyToDirName = "test-engine-build"
+            File engineBuildDir = testClassDirectoryProvider.testDirectory.file(engineCopyToDirName)
+
+            withConnection(connector().forProjectDirectory(engineBuildDir)) {
+                it.newBuild().forTasks("build").run()
+            }
+            engineJarLib = engineBuildDir.file("build/libs/${engineCopyToDirName}.jar")
+        }
+    }
+
+    protected String enableEngineForSuite() {
+        return """
+                useJUnitJupiter()
+
+                dependencies {
+                    implementation files("${TextUtil.normaliseFileSeparators(engineJarLib.absolutePath)}")
+                }
+        """
+    }
+
+    protected void writeTestDefinitions(String path = "src/test/definitions") {
+        file("$path/SomeTestSpec.rbt") << """<?xml version="1.0" encoding="UTF-8" ?>
+            <tests>
+                <test name="foo" />
+                <test name="bar" />
+            </tests>
+        """
+        file("$path/subSomeOtherTestSpec.rbt") << """<?xml version="1.0" encoding="UTF-8" ?>
+            <tests>
+                <test name="other" />
+            </tests>
+        """
+    }
+
+    protected void assertTestsFromAllDefinitionsExecuted() {
+        testEvents {
+            task(':test') {
+                nested('Gradle Test Run :test') {
+                    nested('Gradle Test Executor') {
+                        test('Test SomeTestSpec.rbt - foo')
+                        test('Test SomeTestSpec.rbt - bar')
+                        test('Test subSomeOtherTestSpec.rbt - other')
+                    }
+                }
+            }
+        }
+    }
+
+    protected void assertTestsFromSecondDefinitionsExecuted() {
+        testEvents {
+            task(':test') {
+                nested('Gradle Test Run :test') {
+                    nested('Gradle Test Executor') {
+                        test('Test subSomeOtherTestSpec.rbt - other')
+                    }
+                }
+            }
+        }
+    }
+}
--- a/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/FetchIdeaModel.java
+++ b/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/FetchIdeaModel.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.integtests.tooling.r940;
+
+import org.gradle.tooling.BuildAction;
+import org.gradle.tooling.BuildController;
+import org.gradle.tooling.model.idea.IdeaProject;
+
+/**
+ * Example build action used in the r940 cross-version tooling API tests.
+ */
+public class FetchIdeaModel implements BuildAction<IdeaProject> {
+    public IdeaProject execute(BuildController controller) {
+        return controller.getModel(IdeaProject.class);
+    }
+}
--- a/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResourceBasedTestingCrossVersionTest.groovy
+++ b/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResourceBasedTestingCrossVersionTest.groovy
@@ -1,0 +1,231 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.integtests.tooling.r940
+
+import org.gradle.integtests.tooling.fixture.TargetGradleVersion
+import org.gradle.integtests.tooling.fixture.ToolingApiVersion
+import org.gradle.tooling.ProjectConnection
+import org.gradle.tooling.events.OperationType
+import org.gradle.tooling.events.test.JvmTestOperationDescriptor
+import org.gradle.tooling.events.test.TestOperationDescriptor
+import org.gradle.tooling.events.test.source.FileSource
+import org.gradle.tooling.events.test.source.FilesystemSource
+import org.gradle.tooling.model.gradle.GradleBuild
+import testengines.TestEnginesFixture
+
+@TargetGradleVersion(">=9.4.0")
+class  ResourceBasedTestingCrossVersionTest extends AbstractResourceBasedTestingCrossVersionTest {
+
+    @Override
+    List<TestEnginesFixture.TestEngines> getEnginesToSetup() {
+        return [TestEnginesFixture.TestEngines.BASIC_RESOURCE_BASED]
+    }
+
+    def "can launch resource-based test with #entryPoint"() {
+        given:
+        buildFile << """
+            plugins {
+                id 'java-library'
+            }
+
+            ${mavenCentralRepository()}
+
+            testing.suites.test {
+                ${enableEngineForSuite()}
+
+                targets.all {
+                    testTask.configure {
+                        testDefinitionDirs.from("$DEFAULT_DEFINITIONS_LOCATION")
+                    }
+                }
+            }
+        """
+        writeTestDefinitions()
+
+        when:
+        withConnection {
+            entryPointConfiguration(it).addProgressListener(events, OperationType.TASK, OperationType.TEST)."$execMethod"()
+        }
+
+        then:
+        assertTestsFromAllDefinitionsExecuted()
+
+        where:
+        entryPoint             | entryPointConfiguration                                                    | execMethod
+        'BuildLauncher'        | { ProjectConnection p -> p.newBuild().forTasks("test") }                   | 'run'
+        'TestLauncher'         | { ProjectConnection p -> p.newTestLauncher().forTasks("test") }            | 'run'
+        'ModelBuilder'         | { ProjectConnection p -> p.model(GradleBuild).forTasks("test") }           | 'get'
+        'BuildActionExecuter'  | { ProjectConnection p -> p.action(new FetchIdeaModel()).forTasks("test") } | 'run'
+    }
+
+    def "can filter resource-based tests"() {
+        given:
+        buildFile << """
+         plugins {
+             id 'java-library'
+         }
+
+         ${mavenCentralRepository()}
+
+         testing.suites.test {
+             ${enableEngineForSuite()}
+
+             targets.all {
+                 testTask.configure {
+                     testDefinitionDirs.from("$DEFAULT_DEFINITIONS_LOCATION")
+
+                     options {
+                         excludeEngines("junit-jupiter")
+                     }
+                 }
+             }
+         }
+
+         // Ensure the definitions directory exists even if no definitions are added; otherwsie the task will fail with "Test definitions directory does not exist"
+         project.layout.projectDirectory.file("$DEFAULT_DEFINITIONS_LOCATION").getAsFile().mkdirs()
+     """
+        writeTestDefinitions()
+
+        when:
+        withConnection {
+            it.newTestLauncher()
+                .addProgressListener(events, OperationType.TASK, OperationType.TEST)
+                .withTestsFor { tests ->
+                    tests.forTaskPath(':test').includePattern("$DEFAULT_DEFINITIONS_LOCATION/subSomeOtherTestSpec.rbt")
+                }
+                .run()
+        }
+
+        then:
+        assertTestsFromSecondDefinitionsExecuted()
+    }
+
+    @ToolingApiVersion(">=9.4.0")
+    def "can rerun resource-based test task based on descriptors but cannot filter for individual scenarios"() {
+        given:
+        buildFile << """
+            plugins {
+                id 'java-library'
+            }
+
+            ${mavenCentralRepository()}
+
+            testing.suites.test {
+                ${enableEngineForSuite()}
+
+                targets.all {
+                    testTask.configure {
+                        testDefinitionDirs.from("$DEFAULT_DEFINITIONS_LOCATION")
+                    }
+                }
+            }
+        """
+        writeTestDefinitions()
+        withConnection {
+            it.newTestLauncher().addProgressListener(events, OperationType.TASK, OperationType.TEST).forTasks("test").run()
+        }
+
+        when:
+        assertTestsFromAllDefinitionsExecuted()
+        def test1 = events.operation('Test SomeTestSpec.rbt - foo')
+
+        then:
+        (test1.descriptor as JvmTestOperationDescriptor).source instanceof FilesystemSource
+
+        when:
+        events.clear()
+        withConnection {
+            it.newTestLauncher().addProgressListener(events, OperationType.TASK, OperationType.TEST).withTests(test1.descriptor as TestOperationDescriptor).run()
+        }
+
+        then:
+        assertTestsFromAllDefinitionsExecuted()
+        result.output.contains("Re-running resource-based tests is not supported via TestLauncher API. The ':test' task will be scheduled without further filtering.")
+    }
+
+
+    @ToolingApiVersion(">=8.0 <9.4.0")
+    def "old client can run resource-based tests"() {
+        given:
+        buildFile << """
+            plugins {
+                id 'java-library'
+            }
+
+            ${mavenCentralRepository()}
+
+            testing.suites.test {
+                ${enableEngineForSuite()}
+
+                targets.all {
+                    testTask.configure {
+                        testDefinitionDirs.from("$DEFAULT_DEFINITIONS_LOCATION")
+                    }
+                }
+            }
+        """
+        writeTestDefinitions()
+
+        when:
+        withConnection {
+            it.newBuild().addProgressListener(events, OperationType.TASK, OperationType.TEST).forTasks("test").run()
+        }
+
+        then:
+        assertTestsFromAllDefinitionsExecuted()
+    }
+
+    @ToolingApiVersion(">=9.4.0")
+    def "receives detailed information about resource-based test execution"() {
+        given:
+        buildFile << """
+            plugins {
+                id 'java-library'
+            }
+
+            ${mavenCentralRepository()}
+
+            testing.suites.test {
+                ${enableEngineForSuite()}
+
+                targets.all {
+                    testTask.configure {
+                        testDefinitionDirs.from("$DEFAULT_DEFINITIONS_LOCATION")
+                    }
+                }
+            }
+        """
+        writeTestDefinitions()
+
+        when:
+        withConnection {
+            it.newBuild().addProgressListener(events, OperationType.TASK, OperationType.TEST).forTasks("test").run()
+        }
+
+        then:
+        assertTestsFromAllDefinitionsExecuted()
+        def test1 = events.operation('Test SomeTestSpec.rbt - foo')
+        def test2 = events.operation('Test SomeTestSpec.rbt - bar')
+        def test3 = events.operation('Test subSomeOtherTestSpec.rbt - other')
+        (test1.descriptor as JvmTestOperationDescriptor).source instanceof FileSource
+        ((test1.descriptor as JvmTestOperationDescriptor).source as FileSource).file == file('src/test/definitions/SomeTestSpec.rbt')
+        (test2.descriptor as JvmTestOperationDescriptor).source instanceof FileSource
+        ((test2.descriptor as JvmTestOperationDescriptor).source as FileSource).file == file('src/test/definitions/SomeTestSpec.rbt')
+        (test3.descriptor as JvmTestOperationDescriptor).source instanceof FileSource
+        ((test3.descriptor as JvmTestOperationDescriptor).source as FileSource).file == file('src/test/definitions/subSomeOtherTestSpec.rbt')
+    }
+}
--- a/platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/source/ClasspathResourceSource.java
+++ b/platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/source/ClasspathResourceSource.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.tooling.events.test.source;
+
+import org.gradle.api.Incubating;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Classpath resource-based test source.
+ *
+ * @since 9.4.0
+ */
+@Incubating
+public interface ClasspathResourceSource extends TestSource {
+
+    /**
+     * Returns the name of the source classpath resource.
+     *
+     * @since 9.4.0
+     */
+    String getClasspathResourceName();
+
+    /**
+     * Returns the position within the file, or null if not available.
+     *
+     * @since 9.4.0
+     */
+    @Nullable
+    FilePosition getPosition();
+}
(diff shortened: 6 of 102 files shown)


donat:
@bot-gradle test this

donat:
@bot-gradle test this

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/internal/DefaultReportMetadataEvent.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.Map;
+
+/**
+ * Implementation of the {@code TestMetadataEvent} interface.
+ */
+@NullMarked
+public class DefaultReportMetadataEvent extends DefaultTestMetadataEvent implements ReportMetadataEvent {
+
+    private final long timestamp;
```
tresat: 💅 I think we should name this something a little clearer, like `metadataTimestamp`, so it isn't confused with `eventTime`.  We should clarify this is the time supplied to the metadata event, _not_ the time the event emitted.

I know this is confusing.  The value here is the time reported by `TestMetadataEvent#getLogTime()`, which we've _said_ could be set to the start time (not the actual time the event happened) to indicate a "timeless" metadata.  See `"reports custom test events with metadata reusing test start timestamp"`.

subprojects/build-events/src/main/java/org/gradle/internal/build/event/types/DefaultReportMetadataEvent.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Map;
+
+/**
+ * Default implementation of the {@code InternalTestMetadataEvent} interface.
+ *
+ * This is created by the provider side of the tooling API.
+ */
+@NullMarked
+public class DefaultReportMetadataEvent extends DefaultTestMetadataEvent implements InternalReportMetadataEvent {
+    private final long timestamp;
```
tresat: 💅 Let's clarify the meaning of the timestamp here, too.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/TestResources.java
```diff
@@ -86,11 +86,11 @@ public boolean maybeCopy(String resource) {
         }
         TestFile dir = resources.findResource(resource);
         if (dir != null) {
-            logger.debug("Copying test resource '{}' from {} to test directory.", resource, dir);
+            logger.warn("Copying test resource '{}' from {} to test directory.", resource, dir);
```
tresat: 💅 I don't think these should be left as warnings, the expectation is that these succeed.

subprojects/build-events/src/main/java/org/gradle/internal/build/event/types/DefaultResourceBasedTestDescriptor.java
```diff
@@ -0,0 +1,66 @@
...
+import org.gradle.tooling.internal.protocol.events.InternalOperationDescriptor;
+import org.gradle.tooling.internal.protocol.events.InternalResourceBasedTestDescriptor;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+import java.io.File;
+import java.io.Serializable;
+
+@NullMarked
+public class DefaultResourceBasedTestDescriptor extends DefaultTestDescriptor implements Serializable, InternalResourceBasedTestDescriptor, InternalOperationDescriptor {
```
tresat: ❓ Do we have any option for not extending `DefaultTestDescriptor` here?  This really shouldn't need a `suite/class/methodName` - which I see are nullable, but still, the inheritance itself is misleading.

I see on `InternalJvmTestDescriptor` the note "DO NOT CHANGE THIS INTERFACE. It is part of the cross-version protocol."

What do you think about doing something like this instead: https://github.com/gradle/gradle/pull/35620/commits/3509f6c7e5ad3e749d52c3335919c897b73b6a0e

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/tasks/testing/FileSource.java
```diff
@@ -0,0 +1,39 @@
...
+import java.io.File;
+
+/**
+ * The file defining the test.
+ *
+ * @since 9.3.0
+ */
+@Incubating
+@NullMarked
+public interface FileSource extends TestSource {
```
tresat: 🤔 I wonder if we should be even more explicit and call this `SingleFileSource`.

In `JunitPlatformTestExecutionListener#sourceOf` it looks we're already being cautious and using the `UnknownSource` if we get a `DirectorySource`.  We should probably test for this case directly with a `TestEngine` that executes directory-based descriptors.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestDescriptor.java
```diff
@@ -44,7 +49,13 @@ public DefaultTestDescriptor(Object id, @Nullable String className, String name,
     @Override
     public String toString() {
         String className = getClassName();
-        return "Test " + getName() + (className == null ? "" : ("(" + className + ")"));
+        String classNameSuffix = "";
+        // TODO (donat) store whether class name is not available for tests  (is the case for resource-based testing)
+        // should be deleted once https://github.com/gradle/gradle/pull/35575 is merged
```
tresat: Should merge soon, let's be sure to remove this after rebasing.

platforms/ide/tooling-api/src/testFixtures/resources/testengines/rbt-engine/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/ResourceBasedTestEngine.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.TestEngine;
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.engine.UniqueId;
+import org.junit.platform.engine.support.descriptor.EngineDescriptor;
+import org.junit.platform.engine.support.discovery.EngineDiscoveryRequestResolver;
+
+import java.util.stream.Collectors;
+
+public class ResourceBasedTestEngine implements TestEngine {
```
tresat: ❗ I really think we should find a way to avoid duplicating all these sample `TestEngine`s.  Can we do something very simple like have `AbstractResourceBasedTestingCrossVersionTest` just `../.../` into `testing-jvm`'s fixtures dir to copy from there?

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestOperationMapper.java
```diff
@@ -99,6 +102,12 @@ private DefaultTestDescriptor toTestDescriptorForSuite(OperationIdentifier build
         } else {
             operationDisplayName = getLegacyOperationDisplayName(operationDisplayName, originalDescriptor);
         }
+
+        TestSource source = suite.getSource();
+        if (source instanceof FileSource) {
```
tresat: ❓ Should this ever happen for a Suite?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResourceBasedTestingCrossVersionTest.groovy
```diff
@@ -0,0 +1,238 @@
...
+        events.tests.size() == 5 // task + executor + 3 test
+        result.output.contains("Warning: Re-running resource-based tests is not supported via TestLauncher API. The ':test' task will be scheduled without further filtering.")
+
+        where:
+        excludingJupiter << [true, false]
+    }
+
+
+    @ToolingApiVersion(">=9.3.0")
+    def "Receives detailed information about non-class-based test execution (excluding jupiter engine = #excludingJupiter)"() {
```
tresat: 🤔 I don't think you need to test this with and without excluding jupiter.  That was just me being paranoid when setting up initial testing.

tresat (changes requested):

subprojects/build-events/src/main/java/org/gradle/internal/build/event/types/DefaultResourceBasedTestDescriptor.java
donat: I'll look into this.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/tasks/testing/FileSource.java
donat: That make sense; I'll adjust the code.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestDescriptor.java
donat: I see it's already merged.

platforms/ide/tooling-api/src/testFixtures/resources/testengines/rbt-engine/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/ResourceBasedTestEngine.java
donat: 💯 We should not duplicate the test engines.
I started your proposed approach and copied the resources from `../..` but it failed on CI (file not found). I'll take a deeper look on how we can remove the duplication.

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestOperationMapper.java
donat: good point, probably not.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResourceBasedTestingCrossVersionTest.groovy
donat: Makes sense. This was simply a shameless copy-paste. I'll clean that up.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/TestResources.java
donat: This was a temporary debug change. Removed it.

subprojects/build-events/src/main/java/org/gradle/internal/build/event/types/DefaultResourceBasedTestDescriptor.java
donat: I was able to integrate your proposal from https://github.com/gradle/gradle/commit/3509f6c7e5ad3e749d52c3335919c897b73b6a0e

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/internal/DefaultReportMetadataEvent.java
donat: For me the meaning was clear, but I understand why you want to apply better naming. I've adjusted the naming and added a bit of javadoc.

donat:
@bot-gradle test this

donat:
@bot-gradle test this

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestExecutionBuildConfigurationAction.java
```diff
@@ -160,26 +176,41 @@ private void configureTestTasksForInternalJvmTestRequest(QueryableExecutionPlan
...
                     if (taskPathOf(testDescriptor).equals(testTaskPath)) {
-                        includeTestMatching((InternalJvmTestDescriptor) testDescriptor, testTask);
+                        includeTestMatching(testDescriptor, testTask);
                     }
                 }
             }
         }
     }
 
+    private static void warnUnsupportedTestRerunningForResourceBasedTests(Collection<InternalTestDescriptor> testDescriptors) {
```
tresat:
```suggestion
    private static void warnIfUnsupportedTestRerunningForResourceBasedTests(Collection<InternalTestDescriptor> testDescriptors) {
```

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/FetchIdeaModel.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package org.gradle.integtests.tooling.r940;
+
+import org.gradle.tooling.BuildAction;
+import org.gradle.tooling.BuildController;
+import org.gradle.tooling.model.idea.IdeaProject;
+
+public class FetchIdeaModel implements BuildAction<IdeaProject> {
```
tresat, line 26: 💅 Javadoc?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestDescriptor.java
```diff
@@ -27,6 +28,7 @@ public class DefaultTestDescriptor extends AbstractTestDescriptor {
     @Nullable
     private final String className;
     private final String classDisplayName;
+    private final boolean isKnownClass;
```
tresat: ❓ Is this field used solely for toString debugging?  Could we remove it?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestDescriptor.java
tresat, line 26: 🤔 Since we probably can't change the name of this type, could we note in a common that this is only used for class-based tests?

subprojects/build-events/src/main/java/org/gradle/internal/build/event/types/DefaultTestDescriptor.java
```diff
@@ -25,7 +25,7 @@
 import java.io.Serializable;
 
 @NullMarked
-public class DefaultTestDescriptor implements Serializable, InternalJvmTestDescriptor, InternalOperationDescriptor {
+public class DefaultTestDescriptor extends BaseTestDescriptor implements Serializable, InternalJvmTestDescriptor, InternalOperationDescriptor {
```
tresat: 🤔 Can we change the name of this type to `DefaultClassBasedTestDescriptor` or something?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestDescriptor.java
donat: It is not only for debugging. When we create the object, we set the className to a special `UnknownClass` string, which makes no sense for non-class-based tests. I introduced `isKnownClass` field to avoid string comparison in the `toString` method.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestDescriptor.java
donat: I'm fairly certain that we create instances of this class for non-class-based tests as well.

subprojects/build-events/src/main/java/org/gradle/internal/build/event/types/DefaultTestDescriptor.java
donat: Yes, we can.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/FetchIdeaModel.java
donat: I don't think we need that. This is a sample build action used in `ResourceBasedTestingCrossVersionTest`. We have a few copies already in the crossVersion source set with the same exact content. It should be static inner class in `ResourceBasedTestingCrossVersionTest`, but the TAPI build action class loading is not capable of loading inner classes from Spock tests.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestDescriptor.java
donat: This is a bit hacky, but has a reason: I did not want to do further refactoring because the class is used by `test-distribution`.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/FetchIdeaModel.java
tresat: My mistake here, this is a test class.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestDescriptor.java
tresat: You're right, we do create this directly from `JUnitPlatformTestExecutionLister` regardless of the type of test.

donat:
@bot-gradle help

donat:
@bot-gradle test QuickFeedback

donat:
@bot-gradle test QuickFeedback

donat:
@bot-gradle test QuickFeedback

cobexer:
@bot-gradle test BD

donat:
@bot-gradle test QuickFeedback

donat:
@bot-gradle test this

donat:
Some test coverage is missing to verify that the TAPI client can receive all JUnit test source types.

subprojects/build-events/src/main/java/org/gradle/internal/build/event/types/test/source/DefaultCompositeTestSource.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.gradle.tooling.internal.protocol.events.InternalCompositeTestSource;
+import org.gradle.tooling.internal.protocol.events.InternalTestSource;
+import org.jspecify.annotations.NullMarked;
+
+import java.io.Serializable;
+import java.util.List;
+
+@NullMarked
+public class DefaultCompositeTestSource implements InternalCompositeTestSource, Serializable {
```
tresat: 🤔 Could we move all these source types to a new sub-package, and add a `package-info` file explaining that the hierarchy here is meant to mirror the hierarchy of `org.junit.platform.engine.TestSource`?

platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/junit/result/JUnitXmlResultWriterSpec.groovy
```diff
@@ -35,6 +35,7 @@ import static org.gradle.api.tasks.testing.TestResult.ResultType.SKIPPED
 import static org.gradle.api.tasks.testing.TestResult.ResultType.SUCCESS
 import static org.hamcrest.CoreMatchers.equalTo
 
+@Ignore
```
tresat: 💅 Ignore?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/tasks/testing/source/TestSource.java
```diff
@@ -0,0 +1,34 @@
...
+package org.gradle.api.tasks.testing;
+
+import org.gradle.api.Incubating;
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Representation of the source of a test or container used to navigate to its location by IDEs and build tools.
+ * This is a marker interface. Clients need to check instances for concrete subclasses or subinterfaces.
+ * <p>
+ * This is an API aligned to test sources representation of the JUnit platform.
```
tresat, line 25: 💭 Can you add the FQN of the JUnit type here?  It's a little confusing that we have a type with the same name and hierarchy, easy to miss.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/testsource/TestSources.java
```diff
@@ -0,0 +1,74 @@
...
+import org.gradle.api.tasks.testing.TestSource;
+import org.gradle.api.tasks.testing.UnknownSource;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+import java.io.File;
+import java.util.List;
+
+@NullMarked
+public class TestSources {
```
tresat: 💭 This type doesn't seem to be necessary.  I suppose it's nice to create all the different types of sources through this one factory type.  Could be `final` and should have javadoc.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestDescriptor.java
```diff
@@ -35,16 +37,19 @@ public DefaultTestDescriptor(Object id, @Nullable String className, String name)
 
     @UsedByScanPlugin("test-distribution")
     public DefaultTestDescriptor(Object id, @Nullable String className, String name, @Nullable String classDisplayName, String displayName) {
-        super(id, name);
+        this(id, className, name, classDisplayName, displayName, TestSources.missing());
```
tresat: 🤔 IDK enough about the scan plugin to know if it supplying `missing` is going to be an issue.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -243,12 +243,12 @@ class NonClassBasedTestingIntegrationTest extends AbstractNonClassBasedTestingIn
...
-        containsLine(result.getOutput(), "FINISH [Test SomeTestSpec.rbt : bar()] [SomeTestSpec.rbt : bar] [SUCCESS] [1] [null]")
-        containsLine(result.getOutput(), "START [Test subSomeOtherTestSpec.rbt : other()] [subSomeOtherTestSpec.rbt : other]")
-        containsLine(result.getOutput(), "FINISH [Test subSomeOtherTestSpec.rbt : other()] [subSomeOtherTestSpec.rbt : other] [SUCCESS] [1] [null]")
+        containsLine(result.getOutput(), "START [Test SomeTestSpec.rbt : foo] [SomeTestSpec.rbt : foo]")
+        containsLine(result.getOutput(), "FINISH [Test SomeTestSpec.rbt : foo] [SomeTestSpec.rbt : foo] [SUCCESS] [1] [null]")
+        containsLine(result.getOutput(), "START [Test SomeTestSpec.rbt : bar] [SomeTestSpec.rbt : bar]")
+        containsLine(result.getOutput(), "FINISH [Test SomeTestSpec.rbt : bar] [SomeTestSpec.rbt : bar] [SUCCESS] [1] [null]")
+        containsLine(result.getOutput(), "START [Test subSomeOtherTestSpec.rbt : other] [subSomeOtherTestSpec.rbt : other]")
+        containsLine(result.getOutput(), "FINISH [Test subSomeOtherTestSpec.rbt : other] [subSomeOtherTestSpec.rbt : other] [SUCCESS] [1] [null]")
```
tresat: 💭 You'll probably have to update this again next time you rebase.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -339,7 +347,41 @@ private DefaultTestSuiteDescriptor createNestedTestSuite(TestIdentifier node, St
...
         String classDisplayName = node.getDisplayName();
-        return new DefaultTestClassDescriptor(idGenerator.generateId(), name, classDisplayName);
+        return new DefaultTestClassDescriptor(idGenerator.generateId(), name, classDisplayName, sourceOf(node));
+    }
+
+    private static TestSource sourceOf(TestIdentifier node) {
+        return node.getSource().map(s -> sourceOf(s)).orElse(TestSources.missing());
+    }
+
+    public static TestSource sourceOf(org.junit.platform.engine.TestSource source) {
```
tresat, line 361: 🤔 Why not move all this to `TestSources`?  Is the JUnit dependency a problem?

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/protocol/test/source/InternalMethodSource.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * DO NOT CHANGE THIS INTERFACE. It is part of the cross-version protocol.
+ *
+ * @since 9.4.0
+ */
+@NullMarked
+public interface InternalMethodSource extends InternalTestSource {
```
tresat, line 24: 💭 All these types could probably due with a sub-package as well.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/internal/DefaultResourceBasedJvmTestOperationDescriptor.java
```diff
@@ -0,0 +1,36 @@
...
+import org.gradle.tooling.internal.protocol.events.InternalTestDescriptor;
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public final class DefaultResourceBasedJvmTestOperationDescriptor extends DefaultTestOperationDescriptor implements ResourceBasedJvmTestOperationDescriptor {
+
+    public DefaultResourceBasedJvmTestOperationDescriptor(
+        InternalTestDescriptor internalTestDescriptor,
+        OperationDescriptor parent,
+        TestSource testSource
```
tresat: 🤔 Should this be stronger-typed to `FilesystemSource`?

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestExecutionBuildConfigurationAction.java
```diff
@@ -160,26 +176,41 @@ private void configureTestTasksForInternalJvmTestRequest(QueryableExecutionPlan
...
         }
     }
 
+    private static void warnIfUnsupportedTestRerunningForResourceBasedTests(Collection<InternalTestDescriptor> testDescriptors) {
+        Set<String> seenTasks = new LinkedHashSet<>();
+        for (InternalTestDescriptor descriptor : testDescriptors) {
+            if (descriptor instanceof InternalResourceBasedTestDescriptor) {
+                String taskPath = taskPathOf(descriptor);
+                if (!seenTasks.contains(taskPath)) {
+                    LOG.warn("Re-running resource-based tests is not supported via TestLauncher API. The '{}' task will be scheduled without further filtering.", taskPath);
```
tresat: 🤔 Could you create an issue now for what is needed to support this?  We'll probably want to address this soon.

subprojects/build-events/src/main/java/org/gradle/internal/build/event/types/DefaultClassBasedTestDescriptor.java
```diff
@@ -48,8 +48,10 @@ public DefaultTestDescriptor(
         @Nullable String className,
         @Nullable String methodName,
         OperationIdentifier parentId,
-        String taskPath
+        String taskPath,
+        InternalTestSource testSource
```
tresat, line 54: 🤔 Could this be stronger-typed?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResourceBasedTestingCrossVersionTest.groovy
```diff
@@ -0,0 +1,264 @@
...
+
+        where:
+        entryPoint             | entryPointConfiguration                                                    | execMethod
+        'BuildLauncher'        | { ProjectConnection p -> p.newBuild().forTasks("test") }                   | 'run'
+        'TestLauncher'         | { ProjectConnection p -> p.newTestLauncher().forTasks("test") }            | 'run'
+        'ModelBuilder'         | { ProjectConnection p -> p.model(GradleBuild).forTasks("test") }           | 'get'
+        'BuildActionExecuter'  | { ProjectConnection p -> p.action(new FetchIdeaModel()).forTasks("test") } | 'run'
+    }
+
+    def "filtering class-based filters with #filterType in TestLauncher will prevent scanning for resource-based tests"() {
```
tresat: 🤔 Could the case when `testDefinitionDirs` is present warn?  Then we could update the test once filtering is finished.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestDescriptor.java
donat: Should this be `null` instead?

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
donat: >  Is the JUnit dependency a problem?

Exactly. `TestSources` is in `base-infrastructure` and doesn't see JUnit platform classes.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestDescriptor.java
tresat: IDK.  @wolfs - Can you answer this?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
donat: In fact I adjusted the test output a bit so that the extra () is removed from the output. I think `Test SomeTestSpec.rbt : foo()` would imply a foo() method, but there are no test methods here, only a test run called  `foo` (without the parenthesis).

platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/junit/result/JUnitXmlResultWriterSpec.groovy
donat: Typo

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/internal/DefaultResourceBasedJvmTestOperationDescriptor.java
donat: I'll check

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResourceBasedTestingCrossVersionTest.groovy
donat: I'll check

subprojects/build-events/src/main/java/org/gradle/internal/build/event/types/DefaultClassBasedTestDescriptor.java
donat: Probably, I'll check

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
tresat: The "subSomeOtherTestSpec.rbt" is also incorrect and fixed in `master`.  Just letting you know to expect this when you rebase.

donat:
@bot-gradle test this

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/testsource/TestSources.java
donat: That's right, I think we should remove it.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
donat: Interesting. The test passes locally. I'll keep an eye on CI.

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestExecutionBuildConfigurationAction.java
donat: https://github.com/gradle/gradle/issues/35800

donat:
@bot-gradle test this

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestExecutionBuildConfigurationAction.java
```diff
@@ -130,23 +137,32 @@ private void configureTestTasksForTestDescriptors(Context context) {
...
                     for (String pkg : testSpec.getPackages()) {
                         commandLineIncludePatterns.add(pkg + ".*");
+                        disableResourceBasedTests(task);
                     }
                 }
             }
         }
     }
 
+    private static void disableResourceBasedTests(AbstractTestTask testTask) {
```
big-guy: Why do we need to disable these tests?  Can we avoid this change until Tom's filtering stuff lands?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/tasks/testing/testsource/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Defines different test sources that can be reported by test engines.
+ * The test source data is serialized from the test worker to the daemon.
+ * The class hierarchy is meant to mirror the one in the JUnit Platform
+ * (see {@code org.junit.platform.engine.TestSource}).
+ */
+package org.gradle.api.tasks.testing.testsource;
```
big-guy: Maybe drop the redundant `test`?
```suggestion
package org.gradle.api.tasks.testing.source;
```

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/tasks/testing/source/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Defines different test sources that can be reported by test engines.
+ * The test source data is serialized from the test worker to the daemon.
+ * The class hierarchy is meant to mirror the one in the JUnit Platform
+ * (see {@code org.junit.platform.engine.TestSource}).
```
big-guy: I would refrain from connecting this too closely to JUnit platform and its engines.  This is a generic Gradle testing feature.


```suggestion
 * Defines different test sources that can be associated to tests.
 * The test source data is serialized from the test worker to the daemon.
 * The class hierarchy is meant to mirror the one in the JUnit Platform
 * (see {@code org.junit.platform.engine.TestSource}), but applies to any kind of test.
```

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/source/MissingSource.java
```diff
@@ -0,0 +1,30 @@
...
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * The test engine did not declare any sources for the test.
+ *
+ * @since 9.4.0
+ */
+@Incubating
+@NullMarked
+public interface MissingSource extends TestSource {
```
big-guy: What's the difference between a missing source and an unknown source?  Can we describe that in the docs?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResourceBasedTestingCrossVersionTest.groovy
donat: I'll add this tomorrow.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
tresat: Oh, I see, I fixed these in my branch - the changes haven't made it to `master` yet.  No need to do anything yet then, sorry for the false alarm.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/internal/source/DefaultPackageSource.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package org.gradle.tooling.events.test.internal.testsource;
+
+import org.gradle.tooling.events.test.testsource.PackageSource;
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public class DefaultPackageSource implements PackageSource {
```
big-guy: I think if you've marked the package, you don't need to mark each file

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/source/ClasspathResourceSource.java
```diff
@@ -0,0 +1,46 @@
...
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Classpath resource-based test source.
+ *
+ * @since 9.4.0
+ */
+@Incubating
+@NullMarked
```
big-guy: If we've marked the package, I don't think we need to mark each class

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestMethodDescriptor.java
```diff
@@ -16,9 +16,11 @@
 
 package org.gradle.api.internal.tasks.testing;
 
+import org.gradle.api.internal.tasks.testing.testsource.DefaultMethodSource;
+
 public class DefaultTestMethodDescriptor extends DefaultTestDescriptor {
     public DefaultTestMethodDescriptor(Object id, String className, String methodName) {
-        super(id, className, methodName);
+        super(id, className, methodName, null, methodName, new DefaultMethodSource(className, methodName, ""));
```
big-guy: Does this mean that all method sources never have the parameters?

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/protocol/events/InternalReportMetadataEvent.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * DO NOT CHANGE THIS INTERFACE. It is part of the cross-version protocol.
+ *
+ * @since 9.3.0
+ */
+@NullMarked
+public interface InternalReportMetadataEvent extends InternalTestMetadataEvent {
```
big-guy: Unused? 9.3 -> 9.4?

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/TestOperationDescriptor.java
```diff
@@ -34,4 +35,12 @@ public interface TestOperationDescriptor extends OperationDescriptor {
      */
     @Incubating
     String getTestDisplayName();
+
+    /**
+     * Returns the information about where the test was defined.
+     * @return the test source.
+     * @since 9.4.0
```
big-guy: We should describe what this means for older Gradle versions

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiClassLoaderProvider.groovy
```diff
@@ -104,6 +105,7 @@ class ToolingApiClassLoaderProvider {
         sharedSpec.allowClass(JavaVersionParser)
         sharedSpec.allowClass(DebugUtil)
         sharedSpec.allowClass(Jvm)
+        sharedSpec.allowClass(TestEnginesFixture.TestEngines)
```
big-guy, line 108: Can we not poke this through or move `TestEnginesFixture` into one of the other packages that are already poked through?

platforms/jvm/testing-jvm/src/testFixtures/groovy/testengines/TestEnginesFixture.groovy
big-guy, line 1: I didn't notice this before, but this class is in the wrong package.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/ResourceBasedJvmTestOperationDescriptor.java
```diff
@@ -0,0 +1,40 @@
...
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Describes a resource-based test that runs on the JVM and for which an event has occurred.
+ *
+ * @since 9.4.0
+ */
+@NullMarked
+@Incubating
+public interface ResourceBasedJvmTestOperationDescriptor extends TestOperationDescriptor {
```
big-guy: Do we actually care about this distinction (resource v class)?  How can we reliably determine if something is a class vs a resource?

e.g., `TestSource` is extensible, I think, so any type we do not recognize will fall into our Unknown test source.  So we could have "resource tests" that have unknown test sources and then we'd report those as JvmTestOperationDescriptors.

Is the `JvmTestKind` important here?

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestOperationMapper.java
```diff
@@ -72,46 +97,90 @@ public List<? extends BuildOperationTracker> getTrackers() {
...
             operationDisplayName = adjustOperationDisplayNameForIntelliJ(operationDisplayName, (AbstractTestDescriptor) originalDescriptor);
         } else {
             operationDisplayName = getLegacyOperationDisplayName(operationDisplayName, originalDescriptor);
         }
-        return new DefaultTestDescriptor(buildOperationId, test.getName(), operationDisplayName, test.getDisplayName(), InternalJvmTestDescriptor.KIND_ATOMIC, null, test.getClassName(), test.getName(), parentId, taskTracker.getTaskPath(buildOperationId));
+
+        InternalTestSource testSource = toInternalTestSource(test.getSource());
+        if (testSource instanceof InternalFilesystemSource) {
+            return new DefaultResourceBasedTestDescriptor(buildOperationId, test.getName(), operationDisplayName, test.getDisplayName(), parentId, taskTracker.getTaskPath(buildOperationId), (InternalFilesystemSource) testSource);
+        }
```
big-guy: If I'm following this right... I think this means you could have:

- A "container" test descriptor (e.g., test.xml)
- A "atomic test" test descriptor (e.g., lines 1-10 of test.xml)

The container test descriptor would be reported as a DefaultClassBasedTestDescriptor, even though it's a resource. And the atomic test descriptor would be reported as a DefaultResourceBasedTestDescriptor (which seems closer).

I think you could have deeper/more elaborate things where a container (directory?) has other containers under it (each file).

I wonder if we don't care about this at all and just need a single type of "JVM test descriptor" that can be container/atomic test or classes/resources.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -339,7 +355,42 @@ private DefaultTestSuiteDescriptor createNestedTestSuite(TestIdentifier node, St
...
+        } else if (source instanceof PackageSource) {
+            PackageSource packageSource = (PackageSource) source;
+            return new DefaultPackageSource(packageSource.getPackageName());
+        } else if (source instanceof CompositeTestSource) {
+            CompositeTestSource compositeTestSource = (CompositeTestSource) source;
+            List<TestSource> mappedSources = compositeTestSource.getSources().stream().map(JUnitPlatformTestExecutionListener::sourceOf).collect(Collectors.toList());
+            return new DefaultCompositeTestSource(mappedSources);
+        } else {
+            return DefaultUnknownSource.getInstance();
+        }
```
big-guy, line 380: TestSource is extensible, so I think this list cannot be exhaustive. 

Can users actually see CompositeTestSource, PackageSource, ClasspathResourceSource, etc today with the way we do discovery?

What would we lose if we didn't support all of these initially?

big-guy: I think we should evaluate each of the test sources and make sure they make sense for what we want to expose.

There are a few weird things:
1. ClassSource provides the class name, but not where to find the source file itself. I could see us reporting either the location of the class file or the location of the source file (or both).  This leaves it up to consumers to figure out how to map from class name -> file name. 
2. ClassSource also provides a FilePosition, which I assume applies to the position in a source file (that we don't provide)
3. Similarly, ClasspathResourceSource provides a resource path, but you need to know how to find this in the right classpath. I'd sort of expect this to give you a more exact location.
4. MethodSource provides a `methodParameterTypes`, which is just a comma separated string of all the parameters to the underlying method concatenated together. It's weird that this isn't a `Collection<String>` or something like that instead.
5. MethodSource provides a `className`, but again very little to link it back to a physical location, like ClassSource.

Is CompositeTestSource ever used in JUnit Jupiter?

I wonder if we can scale this back to 3 things:
1. Filesystem sources
2. Class and method sources
3. Unknown sources

FIlesystem sources would provide a File (and maybe something else for non-directories)
Class and method sources would be a marker only (no other information is available)
Unknown is everything else.

That way we're not commiting to providing exactly what JUnit Platform provides.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/AbstractResourceBasedTestingCrossVersionTest.groovy
```diff
@@ -0,0 +1,93 @@
...
+import org.gradle.test.fixtures.file.TestNameTestDirectoryProvider
+import testengines.TestEnginesFixture
+
+abstract class AbstractResourceBasedTestingCrossVersionTest extends ToolingApiSpecification {
+
+    private static final String ENGINE_COPY_TO_DIR_NAME = "test-engine-build"
+
+    private static File engineJarLib
+    private static File engineBuildDir
+    private static File perClassDir
```
big-guy: These can't be shared, I believe.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/AbstractResourceBasedTestingCrossVersionTest.groovy
```diff
@@ -0,0 +1,93 @@
...
+            enginesToSetup.forEach {
+                resources.maybeCopy(it.name)
+            }
+
+            engineBuildDir = testClassDirectoryProvider.testDirectory.file(ENGINE_COPY_TO_DIR_NAME)
+
+            withConnection(connector().forProjectDirectory(engineBuildDir)) {
+                it.newBuild().forTasks("build").run()
+            }
+            engineJarLib = engineBuildDir.file("build/libs/${ENGINE_COPY_TO_DIR_NAME}.jar")
```
big-guy: I know you're just mimicking the integ tests, but I don't think we need the extra complexity.  This is getting called on every test already, so just copy the test engine into the test and build it as part of the rest of the build-under-test.

big-guy (changes requested):
Not as part of this PR, but I think we need to emit test sources for JUnit4 and TestNG so that consumers can rely on a single source of truth and not need to understand which testing framework is used underneath.

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestExecutionBuildConfigurationAction.java
donat: I kept the original TestLauncher design intact. if the clients selects a test class or method, it expects to run only that. It would be surprising to see unrelated tests running in the same invocation, no matter if they are class or resource based.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/tasks/testing/testsource/package-info.java
donat: Done.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/source/MissingSource.java
donat: I've added some explanation. Ptal.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/protocol/events/InternalReportMetadataEvent.java
donat: Unused. I'll delete it.

donat:
@bot-gradle test this

platforms/jvm/testing-jvm/src/testFixtures/groovy/testengines/TestEnginesFixture.groovy
tresat: What do you mean?  It's in `package testengines` in `/src/testFixtures/groovy/testengines`.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/ResourceBasedJvmTestOperationDescriptor.java
tresat: Isn't checking if the JUnit `TestSource` in the `TestIdentifier` extends `Class/Method/PackageSource` in our execution listener enough?  Even though it's extensible, why would we treat anything that _doesn't_ extend one of these as class-based?

Then anything else becomes non-class-based.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
donat: Looking at the current state I think scaling back the test sources list is the way to go. Also, test sources for class-based test descriptors and test events are weird as they duplicate information. I'll simplify the PR to to only support  Filesystem sources only (ie no class and method sources).

donat:
@bot-gradle test this

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
donat: I've removed the unused test source classes.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/ResourceBasedJvmTestOperationDescriptor.java
big-guy: > Isn't checking if the JUnit TestSource in the TestIdentifier extends Class/Method/PackageSource in our execution listener enough?

I'm not sure if those represent all possible TestSources that could be used by what we call a "class based descriptor".  TestSource is kind of optional, so there could be test engines that do not provide a test source.  e.g., Vintage sometimes provides a null source: https://github.com/junit-team/junit-framework/blob/464022d6c497463f9801791cbe98141ec15f33db/junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/TestSourceProvider.java#L41

I couldn't find anything obvious in the Cucumber or TestNG JUnit Platform test engines that suggested they had other types of TestSource.  Depending on how it's supposed to be used, I could see a CompositeTestSource representing something class-like and non-class like at the same time.

> Then anything else becomes non-class-based.

This isn't currently how we have it.  Everything is class-based unless it's a leaf node and has a  `FilesystemSource` test source.

That means in the hierarchy, you could have something like:
```
root
   class-based test descriptor with a FileSource test source
      resource-based test descriptor with a FileSource + FilePosition test source
```

We also make the assumption that all resource-based tests must be on the filesystem, which ignores something like ClasspathResourceSource, which is a resource file that could be contained inside a jar. 

I'm thinking about how we present tests as an abstract thing without having to explode the number of test descriptor types.

Our TestOperationDescriptor hierarchy is currently:
```
TestOperationDescriptor
    JvmTestOperationDescriptor
    ResourceBasedJvmTestOperationDescriptor
```

I understand why we don't want to nest under `JvmTestOperationDescriptor` because this already carries some baggage WRT class name, method name, etc.

I think our options are:
1. Every test event generated by Gradle is a TestOperationDescriptor and we should phase out JvmTestOperationDescriptor. The test display name and test source is all that's needed.
2. Everything is a JvmTestOperationDescriptor that comes a Test task. The class name, suite name and method name should be phased out. The test display name and test source is all that's needed.
3. Clients can expect TestOperationDescriptor, JvmTestOperationDescriptor, ResourceBasedJvmTestOperationDescriptor and possibly more as we recognize more test sources as special. 

I'm questioning if ResourceBasedJvmTestOperationDescriptor is reliable or useful for clients if they need to look at test source anyways.

big-guy: My suggestion is that we think about the types as separating roles (this is a container or not) vs sources. Since we have JvmTestKind, I'm not sure we even need separate types for containers vs tests. We might be able to get away with everything being a JvmTestOperationDescriptor with nullable fields.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/source/UriSource.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.net.URI;
+
+/**
+ * A test source that can be identified by a URI.
+ *
+ * @since 9.4.0
+ */
+@Incubating
+public interface UriSource extends TestSource {
```
big-guy: I'm not sure we need this one right now either since we don't have a way of selecting URI-based tests, so the only way we'd see this is if a test engine created it directly

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/source/FilePosition.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    int getLine();
+
+    /**
+     * The column number of this FilePosition, if available.
+     *
+     * @since 9.4.0
+     */
+    @Nullable
+    Integer getColumn();
```
big-guy, line 45: I don't have a strong feeling about this, but it looks a little strange to see int + Integer next to each other.

A couple of options:
- Make `column` an int, but the convention is to use `-1` or something to indicate an unknown column instead of null
- Make `line` an `Integer` too for symmetry 
- Ignore me :)

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/tasks/testing/source/MissingSource.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * The test engine did not declare any sources for the test at all.
+ *
+ * This is different from {@link UnknownSource} which indicates that the test engine declared sources, but the sources were not recognized by Gradle.
+ *
+ * @since 9.4.0
+ */
+@Incubating
+public interface MissingSource extends TestSource {
```
big-guy: Ah, I understand better what you mean now. Maybe a slightly different name here would work:
MissingSource sounds more like we were told what the source was, but it didn't exist when we looked for it instead of the test source was never provided.  "UnknownSource" might be a better name.  Or `NoSource`, but this might sound strong/imply that the test definitely doesn't have a source. 

And the current UnknownSource sounds more like "it could have source, but we don't know what it is" but what we're trying to say is that we know there is a source, but we don't understand what kind of source it is. Maybe a name like UnrecognizedSource or OtherSource would be better. OtherSource is more neutral and fits in with something you'd see in an enum.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResourceBasedTestingCrossVersionTest.groovy
```diff
@@ -0,0 +1,264 @@
...
+        writeTestDefinitions()
+
+        when:
+        withConnection {
+            entryPointConfiguration(it).addProgressListener(events, OperationType.TEST)."$execMethod"()
+        }
+
+        then:
+        events.tests.size() == 5 // task + executor + 3 tests
+        events.operation('Test SomeTestSpec.rbt : foo')
```
big-guy: Can we use `TestEventsFixture` for these and maybe add something to assert that particular test sources were found?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/AbstractTestDescriptor.java
```diff
@@ -27,10 +28,12 @@
 public abstract class AbstractTestDescriptor implements TestDescriptorInternal {
     private final Object id;
     private final String name;
+    private final TestSource source;
 
-    public AbstractTestDescriptor(Object id, String name) {
+    public AbstractTestDescriptor(Object id, String name, @Nullable TestSource source) {
```
big-guy: I'm not sure test source can be null.  Our TestEventSerializer requires it to be non-null.  I think it's a good idea for us to try to not let this be null. It would be OK if that means defaulting to Missing or Unknown (I'm not sure which is better).

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestExecutionBuildConfigurationAction.java
big-guy: If I understand correctly...There's a Test task that has both classes and test definitions. If you try running a single class, it'll still run all test definitions + that class. You can't select an individual test definition yet, so the reverse isn't true. This works around that by disabling test definitions if there are class filters.

I wouldn't worry about this as part of this work. Tom's adding filtering and that's part of what he would need to make work.  We don't need a workaround like this at the TAPI level.

donat:
@bot-gradle test this

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/AbstractTestDescriptor.java
donat: I'm leaning towards `NoSource` as a default value. It's not that Gradle cannot tell what's the source, it's more like, "This test has no associated test sources".

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResourceBasedTestingCrossVersionTest.groovy
donat: Right. I've refactored the tests.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/tasks/testing/source/MissingSource.java
donat: I've renamed the relevant test sources according your comments. Now we have `NoSource` and `OtherSource`. I kept it simple for now, but later we could provide some data with `OtherSource`, like the class name of the unrecognised test source type.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/source/UriSource.java
donat: I got rid of it.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/AbstractResourceBasedTestingCrossVersionTest.groovy
donat: There's some extra complexity around this because TestResources does not integrate well with the cross-version test engine. maybeCopy simply fails I do what you suggested. I can give you more details but I think we should tackle that separately.

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestOperationMapper.java
donat: I've removed DefaultResourceBasedTestDescriptor and the related types from the branch.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/events/test/ResourceBasedJvmTestOperationDescriptor.java
donat: Done, I think.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DefaultTestMethodDescriptor.java
donat: I simplified the codebase to align what we currently have in Gradle, ie MethodSource only has a class and a method name.

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TestExecutionBuildConfigurationAction.java
donat: Makes sense. I removed the special handling for TAPI filtering.

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/ClassAndNonClassBasedTestingIntegrationTest.groovy
```diff
@@ -169,4 +169,39 @@ class ClassAndNonClassBasedTestingIntegrationTest extends AbstractNonClassBasedT
...
+                    }
+                }
+            }
+        """
+
+        writeTestClasses()
+        writeTestDefinitions()
+
+        when:
+        succeeds("test", "--tests", "src/test/definitions/sub/SomeOtherTestSpec.rbt", "--tests", "!")
```
tresat: ❓ Why involve class-based testing here at all?  I think just show non-class-based filtering.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junit/JUnitTestEventAdapter.java
```diff
@@ -446,15 +450,26 @@ public void testFinished(Description description) {
...
+        String methodName = methodName(description);
+        TestSource source;
+        if (className != null && methodName != null) {
+            source = new DefaultMethodSource(className, methodName);
+        } else if (className != null && methodName == null) {
+            source = new DefaultClassSource(className);
+        } else {
+            source = DefaultOtherSource.getInstance();
+        }
+        return new DefaultTestDescriptor(id, className, methodName, source);
```
tresat, line 466: 🤔 The different between this and `nullSafeDescriptor` seems to be in the possibility of providing an `OtherSource`.  Why do we need 2 methods here?  Can we consolidate these, or document when each should be called?

platforms/jvm/testing-jvm/src/testFixtures/resources/testengines/custom-test-source/test-engine-build/src/main/java/org/gradle/testing/testengine/engine/ResourceBasedTestEngine.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.TestEngine;
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.engine.UniqueId;
+import org.junit.platform.engine.support.descriptor.EngineDescriptor;
+import org.junit.platform.engine.support.discovery.EngineDiscoveryRequestResolver;
+
+import java.util.stream.Collectors;
+
+public class ResourceBasedTestEngine implements TestEngine {
```
tresat: 🤔 We should give this engine a unique name.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/source/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@NullMarked
+package org.gradle.api.internal.tasks.testing.source;
```
tresat, line 18: 💅 It would be nice to have a brief description of the purpose of the types in each of these new packages.  Just something about what they're used for.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/ClassAndNonClassBasedTestingIntegrationTest.groovy
donat: I think this is not needed. The one you suggest already exists.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/source/package-info.java
donat: There's a bit of description added to the packages containing the interfaces. Should we add more content? Should we reference them in the 'impl' packages?

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junit/JUnitTestEventAdapter.java
donat: Right, I'll missed the duplication. I'll take a stab at it.

donat:
@bot-gradle merge

donat:
@bot-gradle cancel

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/ClassAndNonClassBasedTestingIntegrationTest.groovy
tresat: I was unclear.  I was suggesting deleting this.  But I suppose there's no harm in leaving it.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/source/package-info.java
tresat: The purpose is probably more obvious based on the location to you, but I would appreciate a sentence or two on all of the new packages, even the impls, when I'm reading through them.

PR #35579, opened 2025-11-06
Fixes the formatting of the metadata in HTML test reports

tresat:
Proper row background alternation, aligned data in cells.


tresat:
After: 
image

tresat:
Before:
image

PR #35588, opened 2025-11-07
Restore failed and skipped test tabs

### LINKED ISSUE #35574: New test reports no longer filter for failed tests
### Current Behavior

The new test reports show all test results, even if there are failures. 

### Expected Behavior

In previous versions of Gradle, test reports would only show passing tests if there were no failures. If there were any failures, only the failures would be shown by default.

Now, I have to search through a wall of links (and potentially many levels of them) to find the tests that need to be addressed. 

This is especially disruptive when running many tests using test distribution, as it is easy to have very large test reports that contain many successful test executions. 

Image

### Context (optional)

_No response_

1. Run tests for a project with both passing and failing tests
2. Observe test report containing both passing and failing tests

### Gradle version

9.3.0-milestone-1

9.2.0

### Build scan URL (optional)

_No response_

_No response_

octylFractal:
Fixes #35574

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
@@ -16,18 +16,22 @@
 
 package org.gradle.api.internal.tasks.testing.report.generic;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import org.apache.commons.lang3.stream.Streams;
 import org.gradle.api.internal.tasks.testing.report.generic.MetadataRendererRegistry.MetadataRenderer;
-import org.gradle.api.internal.tasks.testing.results.serializable.TestOutputReader;
 import org.gradle.api.internal.tasks.testing.results.serializable.SerializableFailure;
 import org.gradle.api.internal.tasks.testing.results.serializable.SerializableTestResult;
 import org.gradle.api.internal.tasks.testing.results.serializable.SerializedMetadata;
+import org.gradle.api.internal.tasks.testing.results.serializable.TestOutputReader;
 import org.gradle.api.tasks.testing.TestOutputEvent;
 import org.gradle.api.tasks.testing.TestResult;
+import org.gradle.internal.Pair;
 import org.gradle.internal.html.SimpleHtmlWriter;
 import org.gradle.internal.time.TimeFormatting;
 import org.gradle.reporting.ReportRenderer;
+import org.gradle.reporting.TabsRenderer;
+import org.gradle.util.Path;
 import org.jspecify.annotations.Nullable;
 
 import java.io.IOException;
@@ -36,9 +40,11 @@ import java.math.RoundingMode;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
 import static org.gradle.reporting.HtmlWriterTools.addClipboardCopyButton;
@@ -83,9 +89,53 @@ public abstract class PerRootTabRenderer extends ReportRenderer<TestTreeModel, S
             if (info.getChildren().isEmpty()) {
                 renderLeafDetails(info, htmlWriter);
             } else {
-                renderChildren(htmlWriter);
+                renderContainerDetails(htmlWriter);
             }
             htmlWriter.endElement();
+        }
+
+        private void renderContainerDetails(SimpleHtmlWriter htmlWriter) throws IOException {
+            List<Pair<String, ChildTableRenderer>> childTableRenderers = getChildTableRenderers();
+
+            ReportRenderer<TestTreeModel, SimpleHtmlWriter> renderer;
+            if (childTableRenderers.size() == 1) {
+                // No need to render tabs if there is only one tab
+                renderer = Objects.requireNonNull(childTableRenderers.get(0).right);
+            } else {
+                TabsRenderer<TestTreeModel> tabsRenderer = new TabsRenderer<>();
+                childTableRenderers.forEach(p -> tabsRenderer.add(p.left, p.right));
+                renderer = tabsRenderer;
+            }
+
+            renderer.render(getCurrentModel(), htmlWriter);
+        }
+
+        private List<Pair<String, ChildTableRenderer>> getChildTableRenderers() {
+            List<Pair<Path, TestTreeModel.PerRootInfo>> children = Streams.of(getCurrentModel().getChildrenOf(rootIndex))
+                .flatMap(t ->
+                    t.getPerRootInfo().get(rootIndex).stream()
+                        .map(p -> Pair.of(t.getPath(), p))
+                )
+                .collect(Collectors.toList());
+            ImmutableList.Builder<Pair<String, ChildTableRenderer>> childTableRenderers = ImmutableList.builder();
+            addResultTabIfNeeded("Failed", TestResult.ResultType.FAILURE, children, childTableRenderers);
+            addResultTabIfNeeded("Skipped", TestResult.ResultType.SKIPPED, children, childTableRenderers);
+            childTableRenderers.add(Pair.of("All", new ChildTableRenderer(children)));
+            return childTableRenderers.build();
+        }
+
+        private static void addResultTabIfNeeded(
+            String name,
+            TestResult.ResultType resultType,
+            List<Pair<Path, TestTreeModel.PerRootInfo>> children,
+            ImmutableList.Builder<Pair<String, ChildTableRenderer>> childListRenderers
+        ) {
+            List<Pair<Path, TestTreeModel.PerRootInfo>> matchedChildren = children.stream()
+                .filter(p -> p.right.getResult().getResultType() == resultType)
+                .collect(Collectors.toList());
+            if (!matchedChildren.isEmpty()) {
+                childListRenderers.add(Pair.of(name, new ChildTableRenderer(matchedChildren)));
+            }
         }
 
         private static void renderSummary(TestTreeModel.PerRootInfo info, SimpleHtmlWriter htmlWriter, SerializableTestResult testResult) throws IOException {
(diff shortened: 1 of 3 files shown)


platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -83,11 +88,50 @@ protected void render(TestTreeModel.PerRootInfo info, SimpleHtmlWriter htmlWrite
...
+            renderer.render(getCurrentModel(), htmlWriter);
+        }
+
+        private static void addResultTabIfNeeded(
+            String name,
+            TestResult.ResultType resultType,
+            List<Pair<Path, TestTreeModel.PerRootInfo>> children,
+            List<Pair<String, ChildListRenderer>> childListRenderers
+        ) {
+            List<Pair<Path, TestTreeModel.PerRootInfo>> failedChildren = children.stream()
```
tresat: 💅 Maybe `matchedChildren`, as these are not always failures.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -200,48 +244,76 @@ private void renderFailure(SerializableFailure failure, SimpleHtmlWriter htmlWri
...
+
+        private static String getFormattedSuccessRate(TestTreeModel.PerRootInfo info) {
+            if (info.getTotalLeafCount() == 0) {
+                return "-";
+            }
+
+            BigDecimal runTests = BigDecimal.valueOf(info.getTotalLeafCount());
+            BigDecimal successful = BigDecimal.valueOf(info.getTotalLeafCount() - info.getFailedLeafCount());
+
+            return successful.divide(runTests, 2, RoundingMode.DOWN).multiply(BigDecimal.valueOf(100)).intValue() + "%";
```
tresat: 💭 What's up with the fractional tests here?  Shouldn't these be ints?

platforms/software/testing-base/src/main/resources/org/gradle/api/internal/tasks/testing/report/generic/style.css
```diff
@@ -11,6 +12,14 @@
     vertical-align: top;
 }
 
+div.tab table.thin-cols td {
+    padding-left: 0.5em;
+}
+
+div.tab table.thin-cols th {
+    padding-left: 0.5em;
+}
```
tresat: 💅 Instead of `thin-cols` (which only describes the intended format), it would be better to add a class that describes the purpose of the element, and then apply the style to it.  Maybe `test-results`?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -83,11 +88,50 @@ protected void render(TestTreeModel.PerRootInfo info, SimpleHtmlWriter htmlWrite
...
+            ReportRenderer<TestTreeModel, SimpleHtmlWriter> renderer;
+            if (childListRenderers.size() == 1) {
+                // No need to render tabs if there is only one tab
+                renderer = Objects.requireNonNull(childListRenderers.get(0).right);
+            } else {
+                TabsRenderer<TestTreeModel> tabsRenderer = new TabsRenderer<>();
+                childListRenderers.forEach(p -> tabsRenderer.add(p.left, p.right));
+                renderer = tabsRenderer;
+            }
+            renderer.render(getCurrentModel(), htmlWriter);
```
tresat, line 110: 🤔 I think it would be clearer to split this method.  It's named `renderChildTabs`, but it sometimes doesn't result in child tabs being rendered.  It would be confusing to debug a report that you can see lacks tabs and find yourself stepping through this method.

Maybe something like:

```java
renderNonLeafSummary() {
	List renderers = getChildTabRenderers(); // lines 97-107
	... // continue with line 110
}
```

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -200,48 +244,76 @@ private void renderFailure(SerializableFailure failure, SimpleHtmlWriter htmlWri
...
-            htmlWriter.startElement("th").characters("Child").endElement();
-            if (anyNameAndDisplayNameDiffer) {
-                htmlWriter.startElement("th").characters("Name").endElement();
-            }
-            htmlWriter.startElement("th").characters("Tests").endElement();
-            htmlWriter.startElement("th").characters("Failures").endElement();
-            htmlWriter.startElement("th").characters("Skipped").endElement();
-            htmlWriter.startElement("th").characters("Duration").endElement();
-            htmlWriter.startElement("th").characters("Success rate").endElement();
+                htmlWriter.startElement("th").characters("Child").endElement();
```
tresat, line 274: 💅 Is it worth even having "Child" appear as a title in the HTML vs. just leaving this header cell blank?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -200,48 +244,76 @@ private void renderFailure(SerializableFailure failure, SimpleHtmlWriter htmlWri
...
 
-            List<TestTreeModel> sortedByName = Streams.of(getCurrentModel().getChildrenOf(rootIndex))
-                .sorted(Comparator.comparing(TestTreeModel::getPath))
-                .collect(Collectors.toList());
+                List<Pair<Path, TestTreeModel.PerRootInfo>> sortedByName = new ArrayList<>(children);
+                sortedByName.sort(Comparator.comparing(Pair::left));
 
-            for (TestTreeModel child : sortedByName) {
-                for (TestTreeModel.PerRootInfo perRootInfo : child.getPerRootInfo().get(rootIndex)) {
+                for (Pair<Path, TestTreeModel.PerRootInfo> pair : sortedByName) {
```
tresat, line 290: 💅 Why not use a Stream pipeline here?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -200,48 +244,76 @@ private void renderFailure(SerializableFailure failure, SimpleHtmlWriter htmlWri
...
+            BigDecimal successful = BigDecimal.valueOf(info.getTotalLeafCount() - info.getFailedLeafCount());
+
+            return successful.divide(runTests, 2, RoundingMode.DOWN).multiply(BigDecimal.valueOf(100)).intValue() + "%";
+        }
 
-            boolean anyNameAndDisplayNameDiffer = Iterables.any(
-                getCurrentModel().getChildrenOf(rootIndex),
-                childModel -> Iterables.any(
-                    childModel.getPerRootInfo().get(rootIndex),
+        private static final class ChildListRenderer extends ReportRenderer<TestTreeModel, SimpleHtmlWriter> {
```
tresat: 💅 Maybe `TestResultsTableRenderer`?  I think if someone came to this file looking ot adjust the table, `ChildListRenderer` is not obvious.  

Also, in the context of this class alone, "child" and "children" aren't the clearest.  It's just rendering a table of test results - it doesn't know or care that they are in any sort of hierarchical relationship with anything.

tresat (changes requested):
This didn't seem to work for me when I ran all the tests in a package through the IDE and introduced some artificial failures.  I don't see any new tabs:

Image

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
octylFractal: This is not part of this change (I might move the inner class because this diff is really bad), but it follows the behavior of the original test report. You can't do a `divide` without converting to `BigDecimal` first, so I don't see an issue with this.

Maybe `(runTests * 100) / successful` would be OK even if they were just `long` values, but that's a separate change to make.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
octylFractal: I think the clarity is worth it, since there's no explicit "Children" header outside of the table.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
octylFractal: Actually, it does kind of care about that, because `getUrlTo` is based on that hierarchical relationship. Given this class is an inner class of `ForSummary` and contains a table header of "Child", I don't think it's that surprising. I should call it `ChildTableRenderer` though.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
octylFractal: The writer throws `IOException`, which makes this a bit uglier since it would need a try/catch. I don't think it's worth it.

octylFractal:
> I noticed you have to click on the word "All" or "Failed" directly to switch the tabs - clicking inside the tab but outside the text does nothin.

That's annoying but not introduced by these changes -- I'll have to make another PR to improve that.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -83,11 +89,55 @@ protected void render(TestTreeModel.PerRootInfo info, SimpleHtmlWriter htmlWrite
...
+        }
+
+        private List<Pair<String, ChildTableRenderer>> getChildTableRenderers() {
+            List<Pair<Path, TestTreeModel.PerRootInfo>> children = Streams.of(getCurrentModel().getChildrenOf(rootIndex))
+                .flatMap(t ->
+                    t.getPerRootInfo().get(rootIndex).stream()
+                        .map(p -> Pair.of(t.getPath(), p))
+                )
+                .collect(Collectors.toList());
+            ImmutableList.Builder<Pair<String, ChildTableRenderer>> childTableRenderers = ImmutableList.builder();
```
tresat, line 120: 💅 Why bother with an immutable list here?

tresat (approved):

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
octylFractal: I prefer returning immutable results from methods by default, since it avoids exposing the mutable list as a contract of the method even by accident.

PR #35593, opened 2025-11-07
Rename TestClassProcessor to TestDefinitionProcessor

tresat:
Rename the processor classes - these are no longer limited to just processing test classes, but process any type of `TestDefinition`.


tresat:
@bot-gradle test this

octylFractal (approved):

PR #35601, opened 2025-11-07
Remove the testing-jvm-platform project

tresat:
Move contents to `testing-jvm-infrastructure` project.  This removes the need to use reflection to create a `JUnitPlatformTestDefinitionProcessor`.


tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

big-guy (approved):
The original reason that JUnit platform was split out on its own was for Java compatibility.  JUnit platform required Java 8+ and JUnit4/TestNG could run on Java 6 or 7.

We will probably want to introduce this kind of split again, but it looks like the way it was done originally didn't plan for how to split out JUnit4 and TestNG too.

Issue #35005, opened 2025-09-16
Retire/Remove `SoftwareType` annotation and functionality


ghale:
### Issue type

Polishing

### Problem description

We no longer want to support the SoftwareType annotation and its related functionality. Everything should use the new binding API. This was all experimental/incubating, so we should be able to be aggressive about yanking it out.

PR #35619, opened 2025-11-10
Remove legacy software type support

ghale:
This removes the support for the legacy `@SoftwareType` annotation.  Various tests that previously used plugins with this annotation have been updated to use the binding API instead.  The project report task has also been updated to capture plugins applied with the binding API instead of the legacy API.

This PR also moves a number of classes that are intended as future public API from `project-features` to `project-features-api`.

Note that the smoke test for the experimental AGP declarative plugin has been disabled as this was built with the legacy API.  We'll reenable this once the AGP team has created plugin for the binding API.

Fixes https://github.com/gradle/gradle/issues/35005

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

ghale:
@bot-gradle test this

PR #35621, opened 2025-11-10
Read internal properties from gradle.properties files

alllex:
Fixes https://github.com/gradle/gradle/issues/31243

This PR introduces the ability to read **internal options** from `gradle.properties` files in the usual locations, as previously they could only be supplied via system properties.

Internal properties are now sourced from the following places (with priority from highest to lowest):

- System properties aka `-D`
- `<Gradle User Home>/gradle.properties`
- `<build tree root>/gradle.properties`
- `<Gradle Home>/gradle.properties`

Internal options are also sometimes required very early in the Gradle invocation lifecycle (even before the root build tree is set up), like the `org.gradle.internal.operations.trace` property. In order to accommodate that, the reading of internal properties also now happens early -- in the `CrossBuildSession` scope.

As a result, this provides a consistent and reliable way for Gradle developers to introduce internal properties for their needs without worrying about their availability. On the user side, during troubleshooting, this provides a familiar way to set internal properties that would also work for Tooling API scenarios like IDE Sync.

### DIFF
--- a/testing/internal-testing/src/main/groovy/org/gradle/test/fixtures/file/TestFile.java
+++ b/testing/internal-testing/src/main/groovy/org/gradle/test/fixtures/file/TestFile.java
@@ -189,22 +189,6 @@ public class TestFile extends File {
         }
     }
 
-    /**
-     * Appends the given content to this file, creating parent directories if necessary.
-     */
-    public TestFile append(Object content) {
-        getParentFile().mkdirs();
-        try {
-            ResourceGroovyMethods.append(this, content);
-            return this;
-        } catch (IOException e) {
-            throw new RuntimeException(String.format("Could not append to test file '%s'", this), e);
-        }
-    }
-
-    /**
-     * Replaces the text of this file, creating parent directories if necessary.
-     */
     public TestFile setText(String content) {
         getParentFile().mkdirs();
         try {
@@ -995,25 +979,6 @@ public class TestFile extends File {
             uri += "/";
         }
         return uri;
-    }
-
-    /**
-     * Appends content to the {@code gradle.properties} file in this directory.
-     * <p>
-     * Creates parent directories, if necessary.
-     * <p>
-     * Example usage:
-     * <pre>
-     * file("sub").propertiesFile """
-     *     org.gradle.internal.foo=bar
-     * """
-     * </pre>
-     * Appends the text to {@code sub/gradle.properties} file.
-     *
-     * @return gradle.properties file
-     */
-    public TestFile propertiesFile(@Language("properties") String append) {
-        return file("gradle.properties").append(append);
     }
 
     public static class Snapshot {
--- a/subprojects/core/src/integTest/groovy/org/gradle/api/internal/options/InternalOptionsResolutionIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/api/internal/options/InternalOptionsResolutionIntegrationTest.groovy
@@ -49,7 +49,7 @@ class InternalOptionsResolutionIntegrationTest extends AbstractIntegrationSpec {
 
         settingsFile << printInternalOption()
 
-        executer.gradleUserHomeDir.propertiesFile """
+        executer.gradleUserHomeDir.file("gradle.properties") << """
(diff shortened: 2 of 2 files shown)


alllex:

alllex:

alllex:
@bot-gradle test this

alllex:

testing/internal-testing/src/main/groovy/org/gradle/test/fixtures/file/TestFile.java
```diff
@@ -975,6 +997,25 @@ public String getDisplayUriForDir() {
...
+     * <pre>
+     * file("sub").propertiesFile """
+     *     org.gradle.internal.foo=bar
+     * """
+     * </pre>
+     * Appends the text to {@code sub/gradle.properties} file.
+     *
+     * @return gradle.properties file
+     */
+    public TestFile propertiesFile(@Language("properties") String append) {
```
mlopatkin: ❓ Does `propertiesFile` really belong to `TestFile`? The latter is rather generic.

I'm also not sure if the proposed API shape makes sense, I'd expect the function to write the contents into `this`.

alllex:
> We're speaking internal options may be required very early in Gradle lifecycle - is it possible to test it also? Maybe via init script?

@6hundreds, this is tested via the property for the Build Operations Trace, which is read even before we have a build tree. See this test in the PR:
https://github.com/gradle/gradle/pull/35621/files#diff-847b271875a932551ad8be787c045a1db4ccd5f1f4684840dc19fbb7322203c0R113

subprojects/core/src/main/java/org/gradle/api/internal/options/InternalOptionsFactory.java
```diff
@@ -0,0 +1,90 @@
...
+        File rootBuildDir
+    ) {
+        Map<String, String> properties = new HashMap<>();
+        putAllProperties(properties, startParameter.getGradleHomeDir());
+        putAllProperties(properties, rootBuildDir);
+        putAllProperties(properties, startParameter.getGradleUserHomeDir());
+        properties.putAll(startParameter.getSystemPropertiesArgs());
+
+        Map<String, String> internalProperties = filterInternalOptions(properties);
+        return new DefaultInternalOptions(internalProperties);
```
mlopatkin, line 70: 🤔 We're somewhat duplicating this logic here and when creating the non-internal. I don't think we want different orderings between those, but I'm not sure how feasible it is to remove this duplication.

mlopatkin (changes requested):
Looks good overall except for TestFile changes.

testing/internal-testing/src/main/groovy/org/gradle/test/fixtures/file/TestFile.java
alllex: I personally think it does, together with the rest of the helper functions that allow you to refer to a common Gradle file _in some directory_. The helpers we have in `LanguageSpecificTestFileFixture` are all only working for simple builds, providing files in the build root. As soon as you need any other directory, the test code has to specify files by hand and there is no highlighting. You can use `xxxSnippet` functions, but that only makes test bodies look bloated, imo.

At the same time, I don't want to push hard for this, especially if this goes again the intuition of other people. We don't want friction when writing tests that can arise from misinterpreting helper method semantics.

I removed these changes from the PR.

subprojects/core/src/main/java/org/gradle/api/internal/options/InternalOptionsFactory.java
alllex: We definitely don't want them to diverge. However, the tests cover the semantics, so I wouldn't be too concerned about this, given the amount of duplicated code is relatively small.

It would probably be hard to unify this with the logic of `GradlePropertiesController`, which comes into picture much later in the build lifecycle. It's probably better to start with the unification with `org.gradle.launcher.cli.converter.LayoutToPropertiesConverter#convert`.

At the same time, now that we have tests, they would support the unification in the later refactoring. I don't think it belongs in this PR.

alllex:
The change is entirely internal and does not require release notes

PR #35628, opened 2025-11-11
Restructure UnitOfWork for better readability

lptr:
`UnitOfWork` was created quite organically, and is due for some cleanup.

- Move inner classes out to top-level types to reduce clutter.
- Reorganize methods to move more important ones to the top and group related methods.
- Move methods and inner-types meaningful only in a mutable context to `MutableUnitOfWork`.
- Rename **identity inputs** --> immutable inputs
- Rename **regular inputs** --> mutable inputs
- Add some TODOs for future improvements.

### DIFF
--- a/platforms/core-configuration/model-reflect/src/main/java/org/gradle/internal/reflect/annotations/TypeAnnotationMetadataStore.java
+++ b/platforms/core-configuration/model-reflect/src/main/java/org/gradle/internal/reflect/annotations/TypeAnnotationMetadataStore.java
@@ -54,6 +54,7 @@ import org.gradle.internal.service.scopes.ServiceScope;
  * </p>
  */
 @ServiceScope(Scope.Global.class)
+// TODO TypeAnnotationStore
 public interface TypeAnnotationMetadataStore {
     TypeAnnotationMetadata getTypeAnnotationMetadata(Class<?> type);
 }
--- a/subprojects/core/src/main/java/org/gradle/api/internal/tasks/properties/TaskProperties.java
+++ b/subprojects/core/src/main/java/org/gradle/api/internal/tasks/properties/TaskProperties.java
@@ -91,5 +91,6 @@ public interface TaskProperties {
     /**
      * Validations for the properties.
      */
+    // TODO Rename to validateProperties
     void validate(PropertyValidationContext validationContext);
 }


lptr:
@bot-gradle test this

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/Stage1BlocksAccessorClassPath.kt
```diff
@@ -155,7 +156,7 @@ abstract class AbstractStage1BlockAccessorsUnitOfWork(
     }
 
     override fun identify(identityInputs: MutableMap<String, ValueSnapshot>, identityFileInputs: MutableMap<String, CurrentFileCollectionFingerprint>) =
-        UnitOfWork.Identity { "$classLoaderHash-$identitySuffix" }
+        Identity { "$classLoaderHash-$identitySuffix" }
```
bamboo, line 159: Just a minor note but I did find `UnitOfWork.Identity` more expressive than just `Identity` since that's quite an overloaded term.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/fingerprint/ConfigurationCacheFingerprintWriter.kt
```diff
@@ -66,9 +66,9 @@ import org.gradle.internal.configuration.problems.PropertyProblem
 import org.gradle.internal.configuration.problems.PropertyTrace
 import org.gradle.internal.configuration.problems.StructuredMessage
 import org.gradle.internal.configuration.problems.StructuredMessageBuilder
+import org.gradle.internal.execution.InputVisitor
```
bamboo, line 69: 👍

subprojects/core/src/main/java/org/gradle/execution/DefaultWorkValidationWarningRecorder.java
```diff
@@ -36,10 +37,10 @@ public class DefaultWorkValidationWarningRecorder implements ValidateStep.Valida
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultWorkValidationWarningRecorder.class);
 
     // TODO If we can ensure that the recorder is only called once per work execution, then we can demote this to a simple counter
-    private final Set<UnitOfWork.Identity> workWithWarnings = ConcurrentHashMap.newKeySet();
+    private final Set<Identity> workWithWarnings = ConcurrentHashMap.newKeySet();
 
     @Override
-    public void recordValidationWarnings(UnitOfWork.Identity identity, UnitOfWork work, Collection<? extends InternalProblem> warnings) {
+    public void recordValidationWarnings(Identity identity, UnitOfWork work, Collection<? extends InternalProblem> warnings) {
         workWithWarnings.add(identity);
```
bamboo, line 44: 🤔 Looking at this code makes me wonder whether all relevant `Identity` implementations are proper values wrt `equals` and `hashCode`. I know some aren't. Why not record the unique id instead?

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/Stage1BlocksAccessorClassPath.kt
lptr: How about `WorkIdentity`?

subprojects/core/src/main/java/org/gradle/execution/DefaultWorkValidationWarningRecorder.java
lptr: As discussed today, this can definitely be improved. We should probably move the whole deferred logic out of the execution engine, and into the way we build/traverse the execution graph.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/Stage1BlocksAccessorClassPath.kt
bamboo: `WorkIdentity` reads well 👍

PR #35629, opened 2025-11-11
Prevent partially initialized annotation handlers from being exposed

### LINKED ISSUE #35526: org.gradle.internal.instantiation.ClassGenerationException: Could not generate a decorated class for type DirectoryProperty
### Current Behavior

This is a link to a mid stacktrace on my build https://github.com/xenoterracide/gradle-semver/actions/runs/18941346345/job/54080662841?pr=494#step:5:437

strangely locally I got a concurrent timeout lock error (see build scans for both behaviors)

### Expected Behavior

tests should have continued to pass

### Context (optional)

unable to upgrade or had to turn off parallel tests. Removing this code allowed me to workaround the issue

```kts
  maxParallelForks =
    Runtime
      .getRuntime()
      .availableProcessors()
      .div(2)
      .or(1)
  systemProperties(
    "junit.jupiter.execution.parallel.enabled" to "true",
    "junit.jupiter.execution.parallel.mode.default" to "concurrent",
    "junit.jupiter.execution.parallel.mode.classes.default" to "concurrent",
  )
```

https://github.com/xenoterracide/gradle-semver/actions/runs/18941346345/job/54080662841?pr=494#step:5:437

I've only seen this on CI but that might be because of #35592

Instructions in README. If instructions don't work then let me know

https://github.com/xenoterracide/gradle-semver/

### Gradle version

9.2.0

9.1.0

### Build scan URL (optional)

https://scans.gradle.com/s/ugjlok7t2bkma

```

Gradle 9.2.0

Build time:    2025-10-29 13:53:23 UTC
Revision:      d9d6bbce03b3d88c67ef5a0ff31f7ae5e332d6bf

Kotlin:        2.2.20
Groovy:        4.0.28
Ant:           Apache Ant(TM) version 1.10.15 compiled on August 25 2024
Launcher JVM:  21.0.9 (Eclipse Adoptium 21.0.9+10-LTS)
Daemon JVM:    /home/xeno/.asdf/installs/java/temurin-21.0.9+10.0.LTS (no JDK specified, using current Java home)
OS:            Linux 6.12.48-1-MANJARO amd64

```

mlopatkin:
The initialization of an annotation handler consists of two parts:
1. Creating an instance with the constructor
2. Feeding registered annotation services to `whenRegistered`

Before the change, we were exposing the reference to the annotation handler instance after step (1). When two threads request the same annotation handler (`ManagedObjectRegistry` in the linked issue) concurrently, the second one was able to see the instance created by the first one before the latter finished the second step. In the `ManagedObjectRegistry` case, the second thread asked it to create the `RegularFileProperty` before the first thread told it that this property type is actually handled.

The fix is simple: now we publish the reference only after the initialization is complete, including all `whenRegistered`. Concurrent requests will hit the slow path of the double-checked locking, waiting for the initialization, but this is mandatory for correctness.

Fixes #35526

### DIFF
--- a/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
+++ b/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
@@ -42,6 +42,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
 
 import static java.util.Collections.singletonList;
 import static org.gradle.util.internal.CollectionUtils.collect;
@@ -508,10 +509,10 @@ public class DefaultServiceRegistry implements CloseableServiceRegistry, Contain
             }
         }
 
-        public void instanceRealized(List<Class<?>> declaredServiceTypes, String displayName, Object instance) {
+        public void instanceRealized(List<Class<?>> declaredServiceTypes, Supplier<String> displayName, Object instance) {
             if (instance instanceof AnnotatedServiceLifecycleHandler && !isAssignableFromAnyType(AnnotatedServiceLifecycleHandler.class, declaredServiceTypes)) {
                 throw new IllegalStateException(String.format("%s implements %s but is not declared as a service of this type. This service is declared as having %s.",
-                    displayName, AnnotatedServiceLifecycleHandler.class.getSimpleName(), format("type", declaredServiceTypes)));
+                    displayName.get(), AnnotatedServiceLifecycleHandler.class.getSimpleName(), format("type", declaredServiceTypes)));
             }
             if (instance instanceof AnnotatedServiceLifecycleHandler) {
                 annotationHandlerCreated((AnnotatedServiceLifecycleHandler) instance);
@@ -522,7 +523,7 @@ public class DefaultServiceRegistry implements CloseableServiceRegistry, Contain
                     boolean declaredWithAnnotation = anyTypeHasAnnotation(annotation, declaredServiceTypes);
                     if (implementationHasAnnotation && !declaredWithAnnotation) {
                         throw new IllegalStateException(String.format("%s is annotated with @%s but is not declared as a service with this annotation. This service is declared as having %s.",
-                            displayName, format(annotation), format("type", declaredServiceTypes)));
+                            displayName.get(), format(annotation), format("type", declaredServiceTypes)));
                     }
                 }
             }
@@ -609,7 +610,7 @@ public class DefaultServiceRegistry implements CloseableServiceRegistry, Contain
         abstract List<Class<?>> getDeclaredServiceTypes();
 
         protected void instanceRealized(Object instance) {
-            owner.ownServices.instanceRealized(getDeclaredServiceTypes(), getDisplayName(), instance);
+            owner.ownServices.instanceRealized(getDeclaredServiceTypes(), this::getDisplayName, instance);
         }
 
         protected void setInstance(Object instance) {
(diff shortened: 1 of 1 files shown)


mlopatkin:
@bot-gradle test this without PTS, APT

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -598,8 +598,9 @@ protected ManagedObjectServiceProvider(DefaultServiceRegistry owner) {
         abstract List<Class<?>> getDeclaredServiceTypes();
 
         protected void setInstance(Object instance) {
-            this.instance = instance;
             owner.ownServices.instanceRealized(this, instance);
+            // Only expose the instance after we're done with initialization.
+            this.instance = instance;
```
bamboo, line 619: 👍

mlopatkin:
@bot-gradle test QFL without PTS

mlopatkin:
@bot-gradle test this without PTS, APT

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -496,11 +496,10 @@ private void putServiceType(Class<?> type, ServiceProvider serviceProvider) {
             }
         }
 
-        public void instanceRealized(ManagedObjectServiceProvider serviceProvider, Object instance) {
-            List<Class<?>> declaredServiceTypes = serviceProvider.getDeclaredServiceTypes();
-            if (instance instanceof AnnotatedServiceLifecycleHandler && !isAssignableFromAnyType(AnnotatedServiceLifecycleHandler.class, serviceProvider.getDeclaredServiceTypes())) {
+        public void instanceRealized(List<Class<?>> declaredServiceTypes, String displayName, Object instance) {
+            if (instance instanceof AnnotatedServiceLifecycleHandler && !isAssignableFromAnyType(AnnotatedServiceLifecycleHandler.class, declaredServiceTypes)) {
                 throw new IllegalStateException(String.format("%s implements %s but is not declared as a service of this type. This service is declared as having %s.",
-                    serviceProvider.getDisplayName(), AnnotatedServiceLifecycleHandler.class.getSimpleName(), format("type", declaredServiceTypes)));
```
mlopatkin, line 515: 💭 We cannot use `getDisplayName()` here anymore, because for `FixedInstanceService` it relies on the fact that the `instance` is already set, and this method is actually called from `setInstance`. As the proper display name is very useful, I rearranged stuff in a way that we no longer rely on the `serviceProvider` in this method at all.
Hopefully, JIT can inline it and avoid allocating the `displayName` when exception doesn't happen.

alllex: The display name computation is probably cheap, but also non-trivial. This change means we'll be computing the names for all instances of all services, which can create more memory pressure for larger builds.

I think the intention would be clearer if we pass the display name as a lazy computation instead, e.g. a `Supplier`.

alllex:
If the issue is indeed a regression in 9.2.0, this change seems relatively small to qualify for the backport to 9.2.1.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
mlopatkin: >I think the intention would be clearer if we pass the display name as a lazy computation instead, e.g. a Supplier.

Thanks!

I was initially reluctant because we'll have to allocate the `Supplier`, but in the end it may be cheaper than computing the unnecessary String upfront. May be easier for JIT to tackle as well.

mlopatkin:
@bot-gradle test RfR

PR #35633, opened 2025-11-11
Support relative path for build operations trace

alllex:
It is now possible to use a relative path value for the `org.gradle.internal.operations.trace` property, which will be resolved against the "current directory" as determined by Gradle. In most cases, it's the root directory of the build, which is the most expected location.

Previously, the output location was relative to the working directory of the daemon in the Gradle User Home. Apart from being inconvenient, it's also a shared space used by all daemons of the same version, which can result in contention or invalid output.

As a nice side effect, when the property value is empty and the default of `operations` is assumed, it's also resolved against the root directory of the build:
```
-Dorg.gradle.internal.operations.trace=
```

now produces an `operations-log.txt` directly in the root of the build.

Relative directories are also supported, so you can do things like `-Dorg.gradle.internal.operations.trace=build/trace` if you want.

### DIFF
--- a/platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationTraceIntegrationTest.groovy
+++ b/platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationTraceIntegrationTest.groovy
@@ -18,6 +18,7 @@ package org.gradle.internal.operations.trace
 
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
+import org.gradle.integtests.fixtures.BuildOperationTreeFixture
 import org.gradle.test.fixtures.file.TestNameTestDirectoryProvider
 import org.junit.Rule
 
@@ -25,6 +26,23 @@ class BuildOperationTraceIntegrationTest extends AbstractIntegrationSpec {
 
     @Rule
     TestNameTestDirectoryProvider tmpDir = new TestNameTestDirectoryProvider(getClass())
+
+    def "produces a valid trace"() {
+        when:
+        run "help", "-D${BuildOperationTrace.SYSPROP}=trace"
+
+        then:
+        file("trace-log.txt").exists()
+
+        and:
+        postBuildOutputContains("Build operation trace:")
+
+        and:
+        def tree = BuildOperationTrace.readTree(file("trace").path)
+        def fixture = new BuildOperationTreeFixture(tree)
+        fixture.roots.first().displayName == "Run build"
+        fixture.only("Configure project :")
+    }
 
     def "no tree files are produced by default"() {
         when:
@@ -34,9 +52,6 @@ class BuildOperationTraceIntegrationTest extends AbstractIntegrationSpec {
         file("trace-log.txt").exists()
         !file("trace-tree.txt").exists()
         !file("trace-tree.json").exists()
-
-        and:
-        postBuildOutputContains("Build operation trace:")
     }
 
     def "no trace files are produced when trace parameter is false"() {


alllex:

platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationTraceIntegrationTest.groovy
```diff
@@ -0,0 +1,76 @@
...
+        when:
+        run "help", "-D${BuildOperationTrace.TREE_SYSPROP}=true", "-D${BuildOperationTrace.SYSPROP}=$absolutePath"
+
+        then:
+        tmpDir.file("custom-trace-log.txt").exists()
+        tmpDir.file("custom-trace-tree.txt").exists()
+        tmpDir.file("custom-trace-tree.json").exists()
+    }
+
+    def "trace files are relative to the current directory when parameter is #description"() {
```
bamboo, line 82: ❓ Perhaps the test should change the working directory for more precise coverage?

alllex: Our test infrastructure already sets the working directory to something custom.

Without the change in the behavior, this test with the forking executor would have produced files in this directory:
```
<gradle/gradle checkout>/build/daemon/distributions-core/9.3.0-20251111230000+0000/
```

which I believe is the working directory of the daemon that the integration test allocates.

Moreover, if you run the test with the old behavior with an embedded executor, then the working directory is different again, it's `platforms/core-runtime/build-operations-trace`.

Do you think we need to do more?

platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationTraceIntegrationTest.groovy
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+package org.gradle.internal.operations.trace
+
+
+import org.gradle.integtests.fixtures.AbstractIntegrationSpec
+import org.gradle.test.fixtures.file.TestNameTestDirectoryProvider
+import org.junit.Rule
+
+class BuildOperationTraceIntegrationTest extends AbstractIntegrationSpec {
```
abstratt, line 25: 👍

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationTrace.java
```diff
@@ -265,6 +276,7 @@ public void write(SerializedOperation serializedOperation) {
         @Override
         public void complete(boolean writeTree) {
             try {
+                System.out.println("Build operation trace: " + logFile(basePath));
```
abstratt, line 279: 💅 

This message and the message on line 301 could be better aligned

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationTrace.java
```diff
@@ -496,12 +508,12 @@ private static List<BuildOperationRecord> readLogToTreeRoots(final File logFile,
...
 
-    private static File logFile(String basePath) {
-        return file(basePath, "-log.txt");
+    private static Path logFile(Path basePath) {
+        return withSuffix(basePath, "-log.txt");
     }
 
-    private static File file(@Nullable String base, String suffix) {
-        return new File((base == null || base.trim().isEmpty() ? "operations" : base) + suffix).getAbsoluteFile();
+    private static Path withSuffix(Path base, String suffix) {
```
abstratt, line 515: 👍

abstratt:
The [examples](https://bt-internal-docs.grdev.net/gbt/how-to/develop/internal-options-with-idea-sync/?h=org.gradle.internal.operations.trace) in the internal BT docs strongly suggests absolute paths should be used. 

I thought of suggesting they should be updated to leverage the new behavior, however they are still correct, it is not really documentation for this option, and having to add a disclaimer about relative paths being relative to the build, but only starting in 9.3 etc seems unnecessary noise.

platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationTraceIntegrationTest.groovy
bamboo: In the spirit of "explicit is better than implicit" I'd make the working directory explicit in the test and different from the one set by the test infrastructure.

alllex: Updated the test to change the working directory. PTAL!

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationTrace.java
alllex: I was reluctant to change those, since tree printing is now disabled by default. But I guess it makes sense

alllex:
@bot-gradle test this

platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationTraceIntegrationTest.groovy
```diff
@@ -0,0 +1,99 @@
...
+
+        then:
+        tmpDir.file("custom-trace-log.txt").exists()
+        tmpDir.file("custom-trace-tree.txt").exists()
+        tmpDir.file("custom-trace-tree.json").exists()
+    }
+
+    def "trace files are relative to the current directory when parameter is #description"() {
+        when:
+        inDirectory"sub"
```
bamboo: 💅 Formatting seems off

bamboo (approved):

platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationTraceIntegrationTest.groovy
alllex: Good catch!

PR #35634, opened 2025-11-11
Move mutable work out of the immutable execution pipeline

lptr:
This is a follow-up to:

- https://github.com/gradle/gradle/pull/35628

The aim is to further clear up the difference between the mutable and immutable pipelines of the execution engine; the immutable pipeline should not know about input changes, the execution history, or overlapping outputs.

To move closer to this goal, this PR splits `ExecuteStep` and `ValidateStep` and introduces their `Mutable` and `Immutable` variants. This required the previously shared tail-end of the pipeline to be inlined into each pipeline.

Only mutable work should have mutable inputs. Mutable inputs have been replaced with immutable for `DependencyAccessorUnitOfWork`.

A performance hack was dropped around immutable transforms that used a cheaper hash in creating the identity of the transform based on a snapshot of the input artifact, and later was re-fingerprinted with normalization. After running all performance tests it seems the hack was not necessary anymore.

Implementation types are now captured in `IdentifyStep`. This later simplifies capturing the before execution state in the immutable pipeline, as we now don't have to fingerprint any new inputs there.

There is no further need for `NoInputChangesStep`, because `InputChangesContext` is gone from the immutable pipeline, and thus the step was removed.

Since `AbstractSkipEmptyWorkStep` had only `SkipEmptyMutableWorkStep` remaining as its sole implementation, the two classes have been merged.

Some context types are also split:
- `BeforeExecutionContext` now has a simple `ImmutableBeforeExecutionContext` variant for immutable execution, and a `MutableBeforeExecutionContext` variant with previous execution context and overlapping output handling.
- `ValidationFinishedContext` is also split (and is reintroduced as an interface) to allow for different inheritance of the mutable and immutable variants to reflect the difference in the pipelines.

Also dropped was the `ChangingOutputsContext` -- a context used purely to mark the point(s) in the pipeline(s) where outputs might be changed. While it would be nice to mark such a point via the type system, it added quite some complexity, and it was also not really true for mutable execution where we delete stale files etc. earlier in the pipeline anyway.

There is also more standardization on "mutable" vs. "immutable" naming for steps and contexts where relevant. The term "incremental" should now only be used when considering the kind of execution of a mutable unit of work, or that of the behavior of its mutable inputs.

## Next steps

- Make `Step` generic over `UnitOfWork` and build the mutable/immutable pipelines only out of steps that can handle the respective types of work.
- Move `UnitOfWork.visitMutableInputs()` to `MutableUnitOfWork`. Immutable work should only have immutable inputs.

- `WorkInputListener` is used in continuous mode to report which are the "interesting" inputs. There is some complexity around "interesting"; doing things this way lets continuous mode ignore changes to the classpath of a `JavaCompile` task that has no sources to compile. It is probably complexity not really worth it, and dropping `relevantBehaviors` and simply checking all input files in continuous mode always would simplify things a bunch.

   `WorkInputListener` is also used in CC, but `relevantBehaviors` from the previous point makes it hard to understand if this is doing the right thing. Again, capturing all inputs here and dropping `relevantBehaviors` altogether is probably the right approach.

- It would be nice to disambiguate the overloaded verb "execute" as it currently refers to (at least) two different things:

    1. Actually executing the action of the unit of work, producing its outputs locally.
    2. Making sure that the outputs of a unit of work are manifested locally. This might involve executing the work in the first sense of the word, but can also mean reusing a previous output (up-to-date checks, build cache).
    3. The invocation of potentially the same work (=same identity) which happens only for project-defined transforms today.

    The role of the execution engine is to "execute" in the 2) sense. We should also move 3) out of the engine to simplify operation and remove the need for `IdentityCache` and the whole deferred mode in the early pipeline.

### DIFF
--- a/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/ImmutableUnitOfWork.java
+++ b/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/ImmutableUnitOfWork.java
@@ -26,4 +26,13 @@ public interface ImmutableUnitOfWork extends UnitOfWork {
      * Returns the {@link ImmutableWorkspaceProvider} to allocate a workspace to execution this work in.
      */
     ImmutableWorkspaceProvider getWorkspaceProvider();
+
+    /**
+     * @deprecated Immutable work should only have immutable inputs.
+     *
+     * TODO Move this method to {@link MutableUnitOfWork}
+     */
+    @Override
+    @Deprecated
+    default void visitMutableInputs(InputVisitor visitor) {}
 }
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/catalog/DefaultDependenciesAccessors.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/catalog/DefaultDependenciesAccessors.java
@@ -433,10 +433,6 @@ public class DefaultDependenciesAccessors implements DependenciesAccessors {
             visitor.visitInputProperty(IN_VERSIONS, model::getVersionAliases);
             visitor.visitInputProperty(IN_PLUGINS, model::getPluginAliases);
             visitor.visitInputProperty(IN_MODEL_NAME, model::getName);
-        }
-
-        @Override
-        public void visitMutableInputs(InputVisitor visitor) {
             visitor.visitInputFileProperty(IN_CLASSPATH, InputBehavior.NON_INCREMENTAL,
                 new InputVisitor.InputFileValueSupplier(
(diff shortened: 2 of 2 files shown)


lptr:
@bot-gradle help

lptr:
@bot-gradle test AllPerformanceTest

lptr:
All performance tests passed:

https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_AllPerformanceTests_Trigger/105323894

asodja:
@bot-gradle test this APT

PR #35640, opened 2025-11-12
Isolate project's extensions

6hundreds:
This PR make it possible for projects access to extensions of parent project(up to the root project) in IP mode.
This intends to cover `Access to configuration time data` cross-project data sharing use case.

Access to extensions considered safe due to extensions isolation done via CC serialization machinery.
[Special Project-scope service](https://github.com/gradle/gradle/blob/778cc5c640ad794a3130b93647619345b82def05/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/IsolatedExtensionsProvider.kt#L50) is responsible for keeping serialized extensions of the particular project and deserialize it on demand.
However, we _eagerly_ serialize all the extensions of the project. This is required to have a consistent view of the extensions during access across different projects. See [the test](https://github.com/gradle/gradle/blob/fd763e767dc0a5d3916211b33fd0eb526a170c25/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsExtensionsAccessIntegrationTest.groovy#L187).
Serializibility of an extension is limited by CC-serialization constrains. So, for example, we can't serialize `GradleScript` object, which is may be captured accidentely. During extensions serialization we [store any exception](https://github.com/gradle/gradle/blob/778cc5c640ad794a3130b93647619345b82def05/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/IsolatedExtensionsProvider.kt#L112) thrown during serialization, and emit IP problem when(if) not-serializable extension was accessed. 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsExtensionsAccessIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsExtensionsAccessIntegrationTest.groovy
@@ -201,10 +201,10 @@ class IsolatedProjectsExtensionsAccessIntegrationTest extends AbstractIsolatedPr
                     }
                 }
             }
-            extensions.add("broken", value)
+            extensions.add("mutable", value)
         """
         buildFile("a/build.gradle", """
-            println("Broken is \${rootProject.extensions.getByName('broken')}")
+            println("Mutable is \${rootProject.extensions.getByName('mutable')}")
             def deps = configurations.dependencyScope("deps") {
                 dependencies.add(dependencyFactory.create(rootProject))
             }
@@ -217,15 +217,20 @@ class IsolatedProjectsExtensionsAccessIntegrationTest extends AbstractIsolatedPr
             res.get().getFiles() // Realize a dependency on the parent project at configuration time
         """)
         buildFile("a/b/build.gradle", """
-            println("Broken is \${rootProject.extensions.getByName('broken')}")
+            println("Mutable is \${rootProject.extensions.getByName('mutable')}")
         """)
 
         when:
         isolatedProjectsRun "help"
 
         then:
-        outputContains("Broken is before")
-        outputContains("Broken is before")
+        outputContains """
+> Configure project :a
+Mutable is before
+Setting value to after
+
+> Configure project :a:b
+Mutable is before"""
     }
 
     //TODO Test isolation of ExtensionContainer.getExtensionSchema?


platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/IsolatedExtensionsContainer.kt
```diff
@@ -0,0 +1,128 @@
...
+    override fun findByName(name: String): Any? {
+        val serialized = serializedExtensionsByName.compute(name) { _, value ->
+            value ?: delegate.findByName(name)?.let { serializer.serialize(it) }
+        }
+        return serialized?.let { deserializer.deserialize(it) }
+    }
+
+    override fun getExtraProperties(): ExtraPropertiesExtension {
+        val properties = deserializer.deserialize(serializedExtraProperties)
+        return DefaultExtraPropertiesExtension(properties).apply { setGradleProperties(gradleProperties) }
```
jvandort: I think we should probably use an immutable version of `ExtraPropertiesExtension` here

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/IsolatedExtensionsContainer.kt
```diff
@@ -0,0 +1,128 @@
...
+        throw InvalidUserCodeException("Mutation of the isolated extensions is prohibited")
+    }
+
+    override fun <T : Any> configure(type: TypeOf<T>, action: Action<in T>) {
+        throw InvalidUserCodeException("Mutation of the isolated extensions is prohibited")
+    }
+
+    override fun <T : Any> configure(name: String, action: Action<in T>) {
+        throw InvalidUserCodeException("Mutation of the isolated extensions is prohibited")
+    }
```
jvandort, line 105: These configure methods _technically_ don't perform any mutation. They just call the action with the extension of the given type. 

Now, I think in practice this method is likely almost always used to mutate the extension, so I am fine with this throwing, but I thought it was worth mentioning that in general we should technically be able to implement this method

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/IsolatedExtensionsContainer.kt
```diff
@@ -0,0 +1,128 @@
...
+        return serialized?.let { deserializer.deserialize(it) }
+    }
+
+    override fun getExtraProperties(): ExtraPropertiesExtension {
+        val properties = deserializer.deserialize(serializedExtraProperties)
+        return DefaultExtraPropertiesExtension(properties).apply { setGradleProperties(gradleProperties) }
+    }
+
+    override fun <T : Any> add(publicType: Class<T>, name: String, extension: T) {
+        throw InvalidUserCodeException("Mutation of the isolated extensions is prohibited")
```
jvandort, line 72: Thought on something like
```
Mutation of another project's `ExtensionContainer` is prohibited when IsolatedProjects is enabled
```

Or better yet, can we make this emit an IP violation?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/IsolatedExtensionsContainer.kt
```diff
@@ -0,0 +1,128 @@
...
+import org.gradle.api.reflect.TypeOf
+import org.gradle.api.reflect.TypeOf.typeOf
+import org.gradle.internal.extensibility.DefaultExtensionsSchema
+import org.gradle.internal.extensibility.DefaultExtraPropertiesExtension
+import org.gradle.internal.extensions.stdlib.uncheckedCast
+import org.gradle.internal.isolate.graph.IsolatedActionDeserializer
+import org.gradle.internal.isolate.graph.IsolatedActionSerializer
+import org.gradle.internal.isolate.graph.SerializedIsolatedActionGraph
+
+internal class IsolatedExtensionsContainer(
```
jvandort, line 31: I think we are still missing synchronization here. 

Even if we are only reading properties (serializing) concurrently, the serialization process can do things like evaluate providers, which can run any number of non-thread-safe operations.

I think we need to graph the project lock before doing any interaction with `delegate`.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/IsolatedExtensionsContainer.kt
```diff
@@ -0,0 +1,128 @@
...
+    private val serializer: IsolatedActionSerializer,
+    private val deserializer: IsolatedActionDeserializer,
+    private val delegate: ExtensionContainerInternal,
+    private val gradleProperties : GradleProperties
+) : ExtensionContainerInternal {
+
+    private val serializedExtensionsByType = mutableMapOf<TypeOf<*>, SerializedIsolatedActionGraph<Any>>()
```
jvandort: Since we always return a new instance of this extension container whenever a use calls `getExtensions` on another project, I am not sure I see the need to perform the caching here. I think something like this would be more useful if we always return the same instance of this container for the same project

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/IsolatedExtensionsContainer.kt
```diff
@@ -0,0 +1,128 @@
...
+    private val serializedExtensionsByName = mutableMapOf<String, SerializedIsolatedActionGraph<Any>>()
```
jvandort: The name `SerializedIsolatedActionGraph` doesn't really make much sense anymore now that we're serializing more than just actions.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/IsolatedExtensionsContainer.kt
jvandort: Does `properties` not already contain the gradle properties?

Looking at `DefaultExtraPropertiesExtension.getProperties` it looks like it already contains them

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/IsolatedExtensionsContainer.kt
```diff
@@ -0,0 +1,128 @@
...
+    private val serializedExtensionsAsMap: SerializedIsolatedActionGraph<Map<String, Any>> by lazy {
```
jvandort: Having three caches here seems inefficient. If the user requests an extension by name, and then by type, and then calls getAsMap, we can theoretically serialize and cache the same extension three times

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
```diff
@@ -164,6 +170,22 @@ class ProblemReportingCrossProjectModelAccess(
         private val dynamicCallProblemReporting: DynamicCallProblemReporting,
     ) : MutableStateAccessAwareProject(delegate, referrer) {
 
+        override fun getExtensions(): ExtensionContainerInternal {
```
abstratt, line 172: 💭 

Given than half of the protocol is not available for immutable access, it may make sense to  refactor `ExtensionContainer` to extend an `ExtensionContainerView` (working name), move the read-only methods there and expose `ExtensionContainerView` in `Project`, or even better, expose it via `IsolatedProject` instead.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheServices.kt
```diff
@@ -90,6 +98,35 @@ class ConfigurationCacheServices : AbstractGradleModuleServices() {
...
+            calculatedValueContainerFactory: CalculatedValueContainerFactory,
+            isolationCodecsProvider: IsolationCodecsProvider
+        ): IsolatedExtensionsProvider {
+            val isolateOwner = object : IsolateOwner {
+                override val delegate: Any = project
+                override fun <T : Any> service(type: Class<T>): T = project.services.get(type)
+            }
+            return DefaultIsolatedExtensionsProvider(
+                project.owner,
+                IsolatedActionSerializer(isolateOwner, beanStateWriterLookup, isolationCodecsProvider),
```
6hundreds, line 123: `IsolatedActionSerializer` must be renamed then, since it's not only about `IsolatedAction` serialization anymore

Issue #34957, opened 2025-09-11
Investigate marking inner classes generated by ASM instantiators as 'static' to avoid memory leaks


aSemy:
### Expected Behavior

To create instances of 'managed' objects Gradle uses ASM to generate subtypes, and additionally creates inner classes to instantiate new instances.

The inner classes are marked as 'public' and 'synthetic'. E.g. in NamedObjectInstantiator and AsmBackedClassGenerator:

- https://github.com/gradle/gradle/blob/v9.0.0/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/model/NamedObjectInstantiator.java#L238
- https://github.com/gradle/gradle/blob/v9.0.0/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/model/NamedObjectInstantiator.java#L154
- https://github.com/gradle/gradle/blob/v9.0.0/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AsmBackedClassGenerator.java#L630

My understanding is inner classes should always be static, where possible, to avoid memory leaks. Should `ACC_PUBLIC | ACC_SYNTHETIC` be replaced with `ACC_PUBLIC | ACC_SYNTHETIC | ACC_STATIC`?

### Context

A non-static inner class keeps an implicit reference to its enclosing instance, which can cause memory leaks. See:

- https://www.jetbrains.com/help/inspectopedia/InnerClassMayBeStatic.html
- https://stackoverflow.com/questions/10864853/when-exactly-is-it-leak-safe-to-use-anonymous-inner-classes?noredirect=1&lq=1

---

I think these non-static classes might be causing classloader leaks when using the Worker API https://github.com/gradle/gradle/issues/18313. When running [this reproducer](https://gist.github.com/martinbonnin/ef7a88a7f31324203b48ee879c640904) every time the Worker API was used Gradle would create a new VisitableURLClassLoader, which appeared to be referenced by the generated inner classes.

I saw the YourKit 'Objects retained by inner class back references' inspection reports a lot of these generated classes:

<img width="2626" height="1172" alt="Image" src="https://github.com/user-attachments/assets/305bd37d-bc22-4a75-b421-f4601a7b241b" />

bamboo:
The issue is in the backlog of the relevant team and is prioritized by them.

***

Thanks, @aSemy, that seems very relevant.

vlsi:
[...]

PR #35643, opened 2025-11-12
Use ClassValue in DefaultClassMap to avoid leaking classloaders

### LINKED ISSUE #35028: Consider using ClassValue for Class -> Value map
### Expected Behavior

`org.gradle.cache.internal.ClassCacheFactory#newClassMap` should probably use [`ClassValue`](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassValue.html)

Currently, `newClassMap` uses `ConcurrentHashMap`, and it is likely less efficient than a Java-optimized `ClassValue`.

I wonder if retrofitting `ClassValue` into `newClassMap` or replacing the return type of `newClassMap` from `Cache<K, V>` to `ClassValue<V>` could be accepted.

Currently `org.gradle.internal.instantiation.generator.AsmBackedClassGenerator` uses `newClassMap`, thus it might retain extra objects when they could be released otherwise.

For instance, the reproducer from https://github.com/gradle/gradle/issues/34957 yields the following 

Image

It looks like the only strong reference that holds `VisitableURLClassLoader` is the `DefaultClassMap` in question.

### Context

See:
* https://github.com/gradle/gradle/issues/34957

vlsi:
`DefaultClassMap` stored a value for every key class in a static `ConcurrentHashMap` (`leakyValues`). For a class loaded by a classloader other than a `VisitableURLClassLoader` — for example an ASM-generated type — that strong reference lived for the whole daemon and kept the class, and its classloader, from being garbage collected. #35028 asked for `ClassValue` here.

The storage scope now follows the lifetime of the key class:

- bootstrap classes (`null` classloader): a `ConcurrentHashMap`, since these classes are never unloaded;
- `VisitableURLClassLoader`: the classloader's user data, as before, so the entries are released with the classloader;
- any other classloader: a `ClassValue`, so each entry is collected together with its class.

Fixes #35028

### Scope

This is a targeted change to one branch of `DefaultClassMap`. It does **not** address the Worker API classloader growth discussed in #34957. The worker classloaders there are themselves `VisitableURLClassLoader`s, handled by the unchanged user-data branch, and they are retained through soft references rather than this map.

### Verification

- [x] Unit test in `DefaultCrossBuildInMemoryCacheFactoryTest`: a class loaded by a throwaway classloader is stored in the cache, the session ends, and after GC the class is unloaded. The test fails on the old `leakyValues` map and passes with `ClassValue`.

### Note on #34957

I reproduced the #34957 Worker API scenario on `master` and on this branch — 8 builds with distinct worker classpaths against a single daemon. The live `VisitableURLClassLoader` count grows identically on both (so this change has no effect there), and the growth disappears under `-XX:SoftRefLRUPolicyMSPerMB=0`. The accumulation is therefore soft-reference caching, reclaimed under memory pressure, not a strong leak.

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/cache/internal/DefaultCrossBuildInMemoryCacheFactory.java
+++ b/subprojects/core/src/main/java/org/gradle/cache/internal/DefaultCrossBuildInMemoryCacheFactory.java
@@ -32,6 +32,7 @@ import java.util.Set;
 import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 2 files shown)


mlopatkin:
@bot-gradle test QF, PT

mlopatkin:
@bot-gradle test APT

vlsi:
@mlopatkin , I have rebased the PR and added a unit test.
The fix should be safe, and it simplifies at last in a unit-test case, so it is worth considering.

It would be great if you could push it through your test pipelines.

vlsi:
Closing in favour of #38527, which supersedes this change.

That PR backs both `newClassCache()` and `newClassMap()` with a `ClassValue`-based cache and, crucially, stops extending `AbstractCrossBuildInMemoryCache`. This PR kept `DefaultClassMap` on that base, so every key `Class` stayed pinned in `valuesForThisSession` for the whole build session — the within-session retention that #38527 removes and this PR did not. It also covers the case this PR targeted: a class loaded by a non-`VisitableURLClassLoader` is stored via `ClassValue` and collected together with its class, verified there by a test that loads a throwaway class in an isolated `URLClassLoader`.

Since #38527 resolves what #35028 asked for (use `ClassValue` for `newClassMap`) and fixes the underlying leak more completely, there is nothing left worth keeping here. Thanks @bamboo.

bamboo:
Thank you, @vlsi!

Incidentally I seem to have missed this PR otherwise it could have been the basis for #38527.

Issue #35625, opened 2025-11-11
Resolution failures during capability conflicts


jvandort:
### Current Behavior

Failures like
```
Caused by: java.lang.NullPointerException: Cannot invoke "org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.SelectorState.getTargetModule()" because the return value of "org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.EdgeState.getSelector()" is null |  
-- | --
at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.NodeState.disconnectOutgoingEdge(NodeState.java:1023) |  
at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.NodeState.removeOutgoingEdges(NodeState.java:1002) |  
at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.NodeState.deselect(NodeState.java:1063) |  
at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.ComponentState.removeOutgoingEdges(ComponentState.java:399) |  
at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.ModuleResolveState.changeSelection(ModuleResolveState.java:221) |  
at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.ModuleResolveState.maybeUpdateSelection(ModuleResolveState.java:481) |  
at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.DependencyGraphBuilder.performSelection(DependencyGraphBuilder.java:259) |  
at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.DependencyGraphBuilder.performSelectionSerially(DependencyGraphBuilder.java:238) |  
at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.DependencyGraphBuilder.resolveEdges(DependencyGraphBuilder.java:221) |  
at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.DependencyGraphBuilder.traverseGraph(DependencyGraphBuilder.java:199) |  
at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.DependencyGraphBuilder.resolve(DependencyGraphBuilder.java:160)
```

### Expected Behavior

No NPEs

### Context (optional)

[...]

PR #34893, opened 2025-09-04
Reimplement strict version constraints

jvandort:
WIP

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #35652, opened 2025-11-13
Resolve NPE in resolution engine

jvandort:
Since refactoring strict version constraints in https://github.com/gradle/gradle/pull/34893 some graph structures can cause NPEs in the resolution engine.

Prior to 34893, there was a bug which erroneously allowed an edge to be added as an outgoing edge to a node more than once. This was due to improper state tracking in removeOutgoingEdges, causing us to retain state in `upcomingNoLongerPendingConstraints` even when a node has been removed from the graph.

This state remained even after being added back to the graph, causing the constraint edge to be added twice to a given node. In the past, this bug silently caused no problem, however after our changes to strict versions, we cleared out an edge's selector when the edge was released. Having two duplicate edges out of a node would cause the duplicate edge to be released twice, causing the NPE.

In practice, an edge should never be added as an outgoing edge to a node more than once. This commit resolves the duplicate edge bug by ensuring we clean up `upcomingNoLongerPendingConstraints` in `removeOutgoingEdges`, and document `removeOutgoingEdges` as effectively the inverse of `visitOutgoingDependenciesAndCollectEdges`.

Since we call cleanupConstraints in removeOutgoingEdges, we refactor existing code which called this method standalone. Now nodes have two simple operations 'from the outside'. you either want to attach all outgoing edges, or remove all outgoing edges. It is up to the node to track the internal state, and no additional public state tracking methods like cleanupConstraints should be necessary.

We additionally strengthen guarantees in the graph, by verifying the reason a component is not selected when a node is processed. Previously, we had a TODO stating we should determine when a component is not selected but a node in that component is selected/has incoming edges. This case happens when a node is part of a module conflict. In these circumstances, we assert that the node is indeed not connected to the graph, as the conflict handlers are responsible for disconnecting a node's outgoing edges when it enters a conflict.

Fixes https://github.com/gradle/gradle/issues/35625

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesConflictResolutionIssuesIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesConflictResolutionIssuesIntegrationTest.groovy
@@ -857,99 +857,51 @@ class CapabilitiesConflictResolutionIssuesIntegrationTest extends AbstractIntegr
     }
 
     @Issue("https://github.com/gradle/gradle/issues/35625")
-    def "can detect Slf4J logger implementation conflicts"() {
+    def "can use manual alignment and capability conflict detection together"() {
+        mavenRepo.module("org.slf4j", "slf4j-simple", "1.7.36").publish()
+
+        mavenRepo.module("org.slf4j", "slf4j-simple", "1.7.27")
+            .dependsOn(mavenRepo.module("org.slf4j", "slf4j-api", "1.7.27").publish())
+            .publish()
+
+        mavenRepo.module("org.apache.logging.log4j", "log4j-slf4j2-impl", "2.20.0")
+            .dependsOn(mavenRepo.module("org.slf4j", "slf4j-api", "1.7.36").publish())
+            .publish()
+
         given:
         buildFile << """
-            plugins {
-                id("org.gradlex.jvm-dependency-conflict-resolution").version("2.4")
-            }
-
             ${common}
-            ${mavenCentralRepository()}
 
             dependencies {
                 implementation("org.slf4j:slf4j-simple:1.7.27")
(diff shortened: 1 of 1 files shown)


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test rfr

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesConflictResolutionIssuesIntegrationTest.groovy
```diff
@@ -856,6 +856,102 @@ class CapabilitiesConflictResolutionIssuesIntegrationTest extends AbstractIntegr
         }
     }
 
+    @Issue("https://github.com/gradle/gradle/issues/35625")
+    def "can detect Slf4J logger implementation conflicts"() {
+        given:
+        buildFile << """
+            plugins {
+                id("org.gradlex.jvm-dependency-conflict-resolution").version("2.4")
```
ljacomet: ❓ Do we need all those tests with a dependency on a published plugin and real dependencies?
This might mean flakiness if the test runs and there are network issues.

ljacomet (approved):
Code changes are ✅ 

Just a question about tests, see below

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesConflictResolutionIssuesIntegrationTest.groovy
jvandort: This was just me being lazy. I've removed all but one of the tests, and updated the remaining test to inline the relevant logic from the plugin and use local dependencies.

PR #35659, opened 2025-11-13
Move InternalFetchAwareBuildController to org.gradle.tooling.internal.protocol package

asodja:
Internal discussion: https://gradle.slack.com/archives/C092LN2SS3C/p1761856479909159

A general guideline is that such Tooling API interfaces should be in `org.gradle.tooling.internal.protocol` and not in any subpackage.


asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

PR #35666, opened 2025-11-13
Restrict TestEventReporter.metadata to String-only values

big-guy:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/TestEventReporter.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/TestEventReporter.java
@@ -61,7 +61,7 @@ public interface TestEventReporter extends AutoCloseable {
      * @param logTime the time the metadata was logged, should be between the start and end times of the test (inclusive)
      * @param key a key to identify the metadata
      * @param value the metadata value
-     * @since 9.3.0
+     * @since 9.4.0
      */
     void metadata(Instant logTime, String key, String value);
 
@@ -76,8 +76,8 @@ public interface TestEventReporter extends AutoCloseable {
      * Keys should usually be unique within the scope of a single test, but this is not enforced.
      *
      * @param logTime the time the metadata was logged, should be between the start and end times of the test (inclusive)
-     * @param values the metadata values, containing in a map which must be serializable by the Tooling API
-     * @since 9.3.0
+     * @param values the key-value pairs
+     * @since 8.13
      */
     void metadata(Instant logTime, Map<String, String> values);
 
--- a/platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/tasks/testing/TestMetadataEvent.java
+++ b/platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/tasks/testing/TestMetadataEvent.java
@@ -42,7 +42,7 @@ public interface TestMetadataEvent {
      * Retrieves the recorded metadata values for this event.
      *
      * @return the event metadata values
-     * @since 9.3.0
+     * @since 8.13
      */
     Map<String, String> getValues();
 }


big-guy:
@bot-gradle test this

big-guy:
@bot-gradle test QF

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/ClientForwardingTestMetadataOperationListener.java
```diff
@@ -56,7 +57,7 @@ public void progress(OperationIdentifier buildOperationId, OperationProgressEven
             ExecuteTestBuildOperationType.Metadata metadata = (ExecuteTestBuildOperationType.Metadata) details;
             InternalTestMetadataDescriptor descriptor = new DefaultTestMetadataDescriptor(new OperationIdentifier(idFactory.nextId()), buildOperationId);
             TestMetadataEvent metadataMetadataEvent = metadata.getMetadata();
-            eventConsumer.progress(new DefaultTestMetadataEvent(progressEvent.getTime(), descriptor, metadataMetadataEvent.getValues()));
+            eventConsumer.progress(new DefaultTestMetadataEvent(progressEvent.getTime(), descriptor, Cast.uncheckedNonnullCast(metadataMetadataEvent.getValues())));
```
big-guy, line 60: From the TAPI side, we still expose a `Map<String, Object>`, but it'll only contain string values now.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResult.java
```diff
@@ -134,8 +135,9 @@ public SerializableTestResult build() {
 
     public static final class Serializer {
         private Serializer() { /* static util class is not instantiable */ }
+        private final static org.gradle.internal.serialize.Serializer<TestMetadataEvent> METADATA_EVENT_SERIALIZER = TestEventSerializer.create().build(TestMetadataEvent.class);
```
big-guy, line 138: When I add file attachments, I'll re-evaluate how this should be passed around and why we're propagating Exception vs IOException.

big-guy:
@bot-gradle test this

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestMethodResult.java
```diff
@@ -31,12 +31,12 @@ public class TestMethodResult {
     private final TestResult.ResultType resultType;
     private final long duration;
     private final long endTime;
-    private final List<SerializedMetadata> metadatas;
+    private final List<TestMetadataEvent> metadatas;
     private final List<SerializableFailure> failures;
```
tresat, line 35: 🤔 I'm unsure of re-using `TestEventMetadata` to replace `SerializedMetadata`.

It seems fine _now_, but it also:
- Restricts how we can evolve this class in the future, since now it must be directly serializable.
- Seems misaligns with how we treat failures here (which makes sense, failures are more complex and this is a simple "value type").

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
```diff
@@ -246,7 +239,7 @@ public BuildOperationDescriptor.Builder description() {
 
         @Override
         public void run(BuildOperationContext context) {
-            GenericPageRenderer renderer = new GenericPageRenderer(outputReaders, rootDisplayNames, metadataRendererRegistry);
+            GenericPageRenderer renderer = new GenericPageRenderer(outputReaders, rootDisplayNames);
```
tresat, line 242: ❓ Are we concerned about losing the ability to easily render different kinds of metadata differently in the reports with this change?  

Everything is a `String` now, so we'd need a different approach for basics like "render URLs as clickable links".

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericPageRenderer.java
```diff
@@ -156,7 +153,7 @@ protected ReportRenderer<TestTreeModel, SimpleHtmlWriter> getContentRenderer() {
                     }
                 }
                 if (!info.getMetadatas().isEmpty()) {
-                    perRootInfoTabsRenderer.add("metadata", new PerRootTabRenderer.ForMetadata(rootIndex, perRootInfoIndex, metadataRendererRegistry));
+                    perRootInfoTabsRenderer.add("metadata", new PerRootTabRenderer.ForMetadata(rootIndex, perRootInfoIndex));
```
tresat, line 156: ❔ Didn't we want to change the name of the tab, too?  Is that worth doing in this PR, too?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/MetadataRendererRegistry.java
```diff
@@ -16,169 +16,10 @@
...
-import java.io.IOException;
-import java.net.URI;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Set;
-
 /**
- * A registry of {@link MetadataRenderer} instances, used to determing how to render custom HTML
- * for a given type of test metadata when generating a test report.
+ * Kept for backwards compatibility
```
tresat, line 20: ❓ Isn't this still needed for use in the classic reports?  

I think it might be better to limit this PR to just changing the metadata types, and address updating the custom rendering (or the removal of it) in another.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -425,13 +424,13 @@ private SimpleHtmlWriter renderMetadataTableHeader(SimpleHtmlWriter htmlWriter)
...
+        private SimpleHtmlWriter renderMetadataTableBody(List<TestMetadataEvent> metadatas, SimpleHtmlWriter htmlWriter) throws IOException {
             htmlWriter.startElement("tbody");
             for (int metadataIdx = 0; metadataIdx < metadatas.size(); metadataIdx++) {
-                SerializedMetadata metadata = metadatas.get(metadataIdx);
+                TestMetadataEvent metadata = metadatas.get(metadataIdx);
                 renderFirstMetadataElement(metadata, metadataIdx, htmlWriter);
-                if (metadata.getEntries().size() > 1) {
-                    List<SerializedMetadata.SerializedMetadataElement> additionalEntries = metadata.getEntries().subList(1, metadata.getEntries().size());
+                if (metadata.getValues().size() > 1) {
+                    Map<String, String> additionalEntries = metadata.getValues();
```
tresat, line 433: ❓ Why change the rendering iteration logic here (except for updating the type)?  Was there a problem with it?

I'm admittedly partial here, but I think the old loops were clearer.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResult.java
```diff
@@ -170,7 +172,7 @@ public static SerializableTestResult deserialize(Decoder decoder) throws IOExcep
             }
 
             ImmutableList<SerializableFailure> failures = deserializeFailures(decoder);
-            ImmutableList<SerializedMetadata> metadatas = deserializeMetadatas(decoder);
+            ImmutableList<TestMetadataEvent> metadatas = deserializeMetadatas(decoder);
```
tresat, line 175: 🤔 Looking at the code here, it seems almost like an oversight that `TestMetadataEvent`s are both an input and output to serialization.

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/TestEventReporter.java
```diff
@@ -77,9 +77,9 @@ public interface TestEventReporter extends AutoCloseable {
      *
      * @param logTime the time the metadata was logged, should be between the start and end times of the test (inclusive)
      * @param values the metadata values, containing in a map which must be serializable by the Tooling API
-     * @since 8.13
+     * @since 9.3.0
```
tresat: 🤔 Are we required to change the "since" on these when changing the generic type arg?  If no, I don't think we should.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestMethodResult.java
big-guy: > Restricts how we can evolve this class in the future, since now it must be directly serializable.

It has to be serializable already because we have to send it from the TestWorker to the daemon.  We have a serializer for it in `TestEventsSerializer`. 

> Seems misaligns with how we treat failures here (which makes sense, failures are more complex and this is a simple "value type").

This is a fair point.  Part of the reason failures are treated differently is that we don't need most of the information contained in a full `TestFailure` (like the exception) and this is trying to minimize the data written.

I'd like to find some way of getting rid of most of this serialization because there are still two paths here:
- Serialization needed for HTML report generation
- Serialization needed for JUnit XML report generation

TestClassResult and TestMethodResult are only used by the JUnit XML side of things. SerializableTestResult is used for HTML reports and TestTreeModelResultsProvider adapts this into TestClassResult/TestMethodResult needed by the JUnit XML report.  IOW, this looks much more complicated and hard to change because we have so many similar copies of the same thing and we're converting between each of them. 

I think there are three paths:
- Continue to have parallel copies of the same structure because they're used for different purposes
- Eliminate most/all duplication by reusing the same structures everywhere
- Replace more things with something like open test reporting

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
big-guy: Yes, I think this is OK. 

Currently, only TestEventReporter would have been able to produce a value other than a String, so there are very few (if any) users of this functionality.  We're going to have two types of data:
- Simple key-value maps (String->String)
- Structured types 

Someone asked JUnit how they would support non-String values and what they talked about is very similar to what I was thinking:
https://github.com/junit-team/junit-framework/issues/2296

If you just want to report some data, without worrying about the structure, you use the map and all strings.

If you want to report some structured data, you'd create a particular object and Gradle/consumers would recognize it as special (e.g., file attachments).

What could be convenient is if you want to report just a single simple thing (like a URL), a structured object is kind of overkill and a Map is awkward.  We could recognize more "simple" types as structured types, e.g., URL, Path, FIle, primitives, etc.  If we did that, I would probably add a requirement for structured types to also report a name.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericPageRenderer.java
big-guy: I have this in the second PR that adds the second tab

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/MetadataRendererRegistry.java
big-guy: No, this was never used in classic reports.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
big-guy: This was mainly to make the signatures work because we went from a List of Entry-like things to a Map and `subList` was a List only thing. 

I went back to a simple iteration in the other PR: https://github.com/gradle/gradle/pull/35534/files#diff-fc7a5972bba2f79fd8fd47802204eb29f4172aac4977439b9a5eef670fe50164R432-R452

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResult.java
big-guy: I'm not sure what you mean by this. Is it because we have a serializer and deserializer in the same scope?

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/TestEventReporter.java
big-guy: I could be convinced either way.  I wasn't sure what's more confusing when there's a method that you could call in 8.13, but now it doesn't compile or work (because the signature changed).  

1. If it says since 8.13, does that mean the signature was the same in 8.13?
2. If it says since 9.3.0, does that mean the method did not exist in any way before then?

(1) is confusing if you expected to use the API across multiple versions of Gradle.
(2) is confusing if you find the API in older versions of Gradle.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestMethodResult.java
tresat: I think the split between XML/HTML reports is the bigger concern here, re: duplication.  

I misunderstood the serialization requirement when reviewing this.  Given that, I think you might have a point about simplifying this.

Can we do it in a separate PR?  It doesn't seem like a necessary change for now, but a possible refactoring, and it makes this more complex.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
tresat: But that JUnit issue was never implemented (it is the same as what you were suggesting re: payloads), so it seems we are still taking functionality away from any users of `TestEventReporter` (if there are any).

I think it's not a matter of single vs. multiple data points, or convenience, but of knowing that a `String` represents something.  I'm less concerned with the awkwardness of using a structured object for a single value and more concerned with us losing one extra data point when it comes to making sense of the values.

If you can provide an `Object`, you can do a type test, figure out it's a `URL`, then treat it differently (in reports, or wherever).  If all you have is a `String`, you're not going to test every one of those just in case it happens to be a `URL`, so you're going to have to use some agreed-upon key that lets Gradle know "this is a URL" if you want special handling.

I'm not saying we shouldn't change to `String` values, I just want to make sure we've thought about:
- Actually taking functionality/UX nicety away from users 
- Doing this outside of a major version (it's still `@Incubating`, but this isn't just tweaking an API, it's removing a feature)

tresat: Though I suppose if we were never actually doing the special-case rendering, we're not changing the UX, just removing the potential for a better UX.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
tresat: Got it.

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/TestEventReporter.java
tresat: We could also do 3. Change the method name.  This would be clearer, but even more annoying.

I lean slightly towards 1.  Plugins using things across multiple versions know to expect problems.  Devs reading docs expect "since" to mean creation time, not last modification time.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResult.java
tresat: I mean, when I see `TestEventMetadata` going into `serialize()` I think it's strange to see `TestEventMetadata` coming out of deserialization right next to `SerializableFailure`s coming out of deserialization.  It's just another piece of finding it odd that failures and metadatas aren't treated similarly.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestMethodResult.java
big-guy: The reason I did this here is that it gets rid of the special serialization that the next PR would need to remove, so either it's being removed here or it's being removed in the other PR.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
big-guy: The special case rendering was pretty limited: File, RegularFile and URI. Everything else was just a toString.  The only other thing I think we lose is the shortening that we would sometimes do for really long values. 

The main thing lost here is that if you reported a URI/URL, it would be made clickable.  With these changes, there's no way to do that.  If you're in JUnit, you could emit a FileEntry, but it can only reference local files. 

I think this is OK for now.

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/TestEventReporter.java
big-guy: The uneven thing is:
`metadata(Instant, String, Object)` -> `metadata(Instant, String, String)` is a creation of a new method
`metadata(Instant, Map<String, Object>)` -> `metadata(Instant, Map<String, String>)` is a modification that erases to the same thing

I'll put these back to 8.13

tresat (approved):

big-guy:
@bot-gradle test this

PR #35674, opened 2025-11-13
Resolve operations trace relative to root directory

alllex:
Follow up for https://github.com/gradle/gradle/pull/35633

The previous PR introduced the ability to specify operations trace relative to the "current directory". While this is an improvement of the base directory being the working directory of the Gradle daemon, the "current directory" is sensitive to where Gradle is invoked. This can lead to rare but nonetheless confusing situations. For instance, when the `trace` parameter is specified in the gradle.properties files (via `jvmargs`), despite being a fixed path, it will produce output in different locations if Gradle is invoked from a subproject directory.

This PR makes the behavior stable and makes the base directory always be the root of the build. While this may seem like a minor change, it actually provides a natural way to introduce the useful concept of "user action root directory." See javadoc for a detailed description:
https://github.com/gradle/gradle/pull/35674/files#diff-e0739e7956e7ae52dfd4649812318bd126b50d9e7a5a1d409caa6e4f36e3040eR37-R54

### DIFF
--- a/platforms/core-runtime/build-state/build.gradle.kts
+++ b/platforms/core-runtime/build-state/build.gradle.kts
@@ -33,6 +33,7 @@ dependencies {
     api(projects.serviceRegistryBuilder)
     api(projects.stdlibJavaExtensions)
 
+    implementation(projects.buildDiscoveryImpl)
     implementation(projects.buildOperationsTrace)
     implementation(projects.concurrent)
     implementation(projects.coreApi)
--- a/platforms/extensibility/unit-test-fixtures/build.gradle.kts
+++ b/platforms/extensibility/unit-test-fixtures/build.gradle.kts
@@ -34,6 +34,7 @@ dependencies {
 
     api(libs.jspecify)
 
+    implementation(projects.buildDiscoveryImpl)
     implementation(projects.buildProcessServices)
     implementation(projects.buildState)
     implementation(projects.classloaders)


alllex:
@bot-gradle test this

platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationTraceIntegrationTest.groovy
```diff
@@ -79,21 +87,27 @@ class BuildOperationTraceIntegrationTest extends AbstractIntegrationSpec {
...
+    def "when running from subdirectory, trace files are relative to the root directory for #description parameter"() {
+        // Explicit settings file to ensure test directory is the root directory of the build
+        settingsFile """
+            rootProject.name = "root"
+            include("sub")
+        """
+        createDirs("sub")
+
         when:
         inDirectory "sub"
```
alllex, line 99: ℹ️  The test was not exactly testing what I intended initially. I didn't realize that if you call `inDirectory` without creating an explicit settings file, then the settings file is never created. Gradle pretends there is an empty settings file, but in the `inDirectory` argument, not in the test directory.

Now we provide the settings file explicitly.

platforms/core-runtime/build-state/src/main/java/org/gradle/internal/buildprocess/execution/BuildSessionLifecycleBuildActionExecutor.java
```diff
@@ -77,6 +82,15 @@ public BuildActionResult execute(BuildAction action, BuildActionParameters actio
         }
     }
 
+    /**
+     * @see CrossBuildSessionParameters#getUserActionRootDirectory()
+     */
+    private File getUserActionRootDirectory(StartParameterInternal startParameter) {
+        BuildLayoutFactory buildLayoutFactory = globalServices.get(BuildLayoutFactory.class);
+        BuildLayout layout = buildLayoutFactory.getLayoutFor(startParameter.toBuildLayoutConfiguration());
```
alllex, line 90: ℹ️  While we do a similar computation in multiple places, the most important part here is what `startParameter` we are using. This start parameter here is the OG. It is the one that actually starts all work required to process the `BuildAction`, provided by the user.

There are many other flavors of the start parameters that may or may not appear later during the build action execution. We create slightly different start parameters for included builds. There can be an entirely different start parameter for a nested build tree. Maybe there are more examples. Importantly, some of those can resolve to a different `BuildLayout` or to a different "root directory".

mlopatkin (approved):

PR #35680, opened 2025-11-14
Allow java executables that cannot be probed to be used with JavaExec

### LINKED ISSUE #25345: No longer possible to specify a executable which is not part of a Java toolchain for a JavaExec task
### Expected Behavior

In 7.6.1 you were able to do this and it ran:

```
tasks.register<JavaExec>("LaunchWithCustomLauncher")
{
    group = "run"
    executable = file("customLauncher").absolutePath. // This is some launcher executable which loads a java runtime over JNI but it not part of that runtime
}

```
### Current Behavior

in 8.0 to 8.1.1 I get the following error message:

Execution failed for task ':LaunchWithCustomLauncher'.
> Failed to calculate the value of task ':LaunchWithCustomLauncher' property 'javaLauncher'.
   > Toolchain installation '/Users/claes/IdeaProjects' could not be probed: A problem occurred starting process 'command '/Users/claes/IdeaProjects/bin/java''

It looks like it's caused by the changes in this PR: https://github.com/gradle/gradle/issues/21386

### Context (optional)

It's a bit of an edge case, but because of various macOS + JNI reasons, we need to use a custom launcher to launch our (mostly) Java application for debugging (in IntelliJ). This launcher loads a JRE via JNI and passes on various command line arguments to it. This way we are able to use the functionality of the JavaExec interface (IDEA connecting up the debugger, passing the right class path & java arguments and so on).

So everything was fine for our use-case until 8.0 decided to be more strict. Which probably makes sense in most cases, but there should be some way to at least opt-out of this behaviour. I wasn't able to find a way to override it either by parameters or by subclassing JavaExec.

### Steps to Reproduce

git clone https://github.com/kurasu/customlauncher

./gradlew LaunchWithCustomLauncher

### Gradle version

8.1.1

7.6.1

### Build scan URL (optional)

_No response_

_No response_

ghale:
This allows the `executable` property of `JavaExec` to be set to a java executable that cannot be probed for some reason.  Since it cannot be probed (and we cannot derive metadata about the java version being used) we disable incrementality/cacheability for the task when such an executable is used.  However, if it is set to a valid java installation (and probing is successful) incrementality/cacheability will work as before (assuming outputs have also been added to the task).

Fixes #25345 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
+++ b/platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
@@ -152,9 +152,7 @@ public abstract class JavaExec extends ConventionTask implements JavaExecSpec {
         // The task will only be up-to-date if it has outputs, those outputs are up-to-date,
         // and the Java launcher can be probed (i.e. javaLanguageVersion is not UNKNOWN)
         doNotTrackStateIf("Java launcher cannot be probed",
-            task -> {
-                return javaLauncher.map(launcher -> launcher.getMetadata().getLanguageVersion()).get() == DefaultJavaLanguageVersion.UNKNOWN;
-            });
+            task -> javaLauncher.map(launcher -> launcher.getMetadata().getLanguageVersion()).get() == DefaultJavaLanguageVersion.UNKNOWN);
     }
 
     @TaskAction
--- a/platforms/jvm/toolchains-jvm-shared/src/main/java/org/gradle/jvm/toolchain/internal/SpecificExecutableToolchainSpec.java
+++ b/platforms/jvm/toolchains-jvm-shared/src/main/java/org/gradle/jvm/toolchain/internal/SpecificExecutableToolchainSpec.java
@@ -78,6 +78,9 @@ public class SpecificExecutableToolchainSpec extends DefaultToolchainSpec {
     }
 
     public File getJavaHome() {
+        // This allows for a "normal" JDK layout where the 'java' executable is in JAVA_HOME/bin/java.
+        // If such an executable is provided, and Gradle can probe the executable, the resulting toolchain
+        // could potentially be used to provide other tools as well (e.g. a compiler or javadoc tool).
         return javaExecutable.getParentFile().getParentFile();
     }
 
--- a/subprojects/core/src/main/java/org/gradle/api/internal/TaskInternal.java
+++ b/subprojects/core/src/main/java/org/gradle/api/internal/TaskInternal.java
@@ -78,7 +78,6 @@ public interface TaskInternal extends Task, Configurable<Task> {
      * @param reason - the reason for not tracking the task state when the condition matches
      * @param spec - the condition to evaluate
      */
-    // TODO We could consider making this a part of the public API
     void doNotTrackStateIf(String reason, Spec<? super TaskInternal> spec);
 
     @Internal


ghale:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/api/internal/AbstractTask.java
```diff
@@ -205,6 +205,7 @@ private AbstractTask(TaskInfo taskInfo) {
         taskExecutionAccessChecker = services.get(TaskExecutionAccessChecker.class);
 
         this.timeout = project.getObjects().property(Duration.class);
+        this.reasonNotToTrackState = project.getObjects().property(String.class);
```
big-guy: I think you can make `getReasonNotToTrackState` an abstract `Property<String>` and it'll be treated as a managed property

platforms/jvm/jvm-services/src/main/java/org/gradle/internal/jvm/inspection/MetadataProbe.java
```diff
@@ -40,14 +40,14 @@
 import static org.objectweb.asm.Opcodes.RETURN;
 import static org.objectweb.asm.Opcodes.V1_1;
 
-class MetadataProbe {
-
+public class MetadataProbe {
+    public static final String PROBE_CLASS_NAME = "JavaProbe";
```
alllex, line 44: 👍

platforms/jvm/jvm-services/src/main/java/org/gradle/jvm/toolchain/JavaLanguageVersion.java
```diff
@@ -94,4 +94,32 @@ static JavaLanguageVersion current() {
...
+            return false;
+        }
+
+        @Override
+        public boolean canCompileOrRun(int otherVersion) {
+            return false;
+        }
+
+        @Override
+        @SuppressWarnings("")
```
alllex: ❓  Is this intentional?

platforms/jvm/jvm-services/src/main/java/org/gradle/jvm/toolchain/JavaLanguageVersion.java
```diff
@@ -94,4 +94,32 @@ static JavaLanguageVersion current() {
      */
     boolean canCompileOrRun(int otherVersion);
 
+    /**
+     * A constant representing an unknown Java language version.
```
alllex: 💭  Given this will show up in auto-complete, we should provide some guidance on whether or how this can be used directly.

Why does this have to be part of the public API?

---

From the current usage, it seems it is not exactly intended for public use. However, the failure mode is unclear if this gets set as part of `JavaToolchainSpec.languageVersion = JavaLanguageVersion.UNKNOWN`. We should make this explicit that selecting toolchains with an unknown version is not supported. We already have `JavaToolchainSpecInternal.isValid()`, which could be helpful.

If we can't make illegal states unrepresentable, we should at least fail fast.

platforms/jvm/toolchains-jvm-shared/src/main/java/org/gradle/jvm/toolchain/internal/SpecificExecutableJavaToolchain.java
```diff
@@ -0,0 +1,44 @@
...
+package org.gradle.jvm.toolchain.internal;
+
+import org.gradle.api.file.RegularFile;
+import org.gradle.api.internal.file.FileFactory;
+import org.gradle.internal.jvm.inspection.JvmInstallationMetadata;
+
+import java.io.File;
+
+/**
+ * A Java toolchain that uses a specific 'java' executable.  This toolchain does not support finding other tools.
```
alllex, line 26: 💅 
```suggestion
 * A Java toolchain that uses a specific 'java' executable.
 * <p>
 * This toolchain does not support finding other tools.
```

subprojects/core/src/main/java/org/gradle/api/internal/AbstractTask.java
```diff
@@ -403,13 +404,28 @@ public void doNotTrackState(String reasonNotToTrackState) {
...
-        return Optional.ofNullable(reasonNotToTrackState);
+    public Provider<String> getReasonNotToTrackState() {
+        return reasonNotToTrackState;
+    }
+
+    @Override
+    public void doNotTrackStateIf(String reason, Spec<? super TaskInternal> spec) {
+        taskMutator.mutate("Task.doNotTrackStateIf(String, Spec)", new Runnable() {
+            @Override
+            public void run() {
```
alllex: 💅  Can this be a lambda?

subprojects/core/src/test/groovy/org/gradle/api/internal/changedetection/changes/DefaultTaskExecutionModeResolverTest.groovy
```diff
@@ -47,7 +48,7 @@ class DefaultTaskExecutionModeResolverTest extends Specification {
 
         then:
         state == DefaultTaskExecutionMode.untracked("For testing")
-        _ * task.getReasonNotToTrackState() >> Optional.of("For testing")
+        _ * task.getReasonNotToTrackState() >> TestUtil.providerFactory().provider { "For testing" }
```
alllex: 💅 
```suggestion
        _ * task.getReasonNotToTrackState() >> Providers.of("For testing")
```

platforms/jvm/toolchains-jvm-shared/src/main/java/org/gradle/jvm/toolchain/internal/JavaToolchainQueryService.java
```diff
@@ -236,12 +240,16 @@ private JavaToolchain downloadToolchain(JavaToolchainSpec spec, Set<JavaInstalla
     private JavaToolchain asToolchainOrThrow(InstallationLocation javaHome, JavaToolchainSpec spec, Set<JavaInstallationCapability> requiredCapabilities, boolean isFallback) {
         final JvmInstallationMetadata metadata = detector.getMetadata(javaHome);
 
-        if (!metadata.isValidInstallation()) {
+        if (!metadata.isValidInstallation() && !(spec instanceof SpecificExecutableToolchainSpec)) {
             throw new GradleException("Toolchain installation '" + javaHome.getLocation() + "' could not be probed: " + metadata.getErrorMessage(), metadata.getErrorCause());
         }
         if (!metadata.getCapabilities().containsAll(requiredCapabilities)) {
```
alllex, line 248: ❔ How does the capabilities check work in case of the specific executable? If the specific executable cannot be probed, doesn't that mean we'll always fail this check?

platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
```diff
@@ -147,6 +148,13 @@ public JavaExec() {
             .orElse(javaToolchainService.launcherFor(it -> {}));
         javaLauncher = objectFactory.property(JavaLauncher.class).convention(javaLauncherConvention);
         javaLauncher.finalizeValueOnRead();
+
+        // The task will only be up-to-date if it has outputs, those outputs are up-to-date,
+        // and the Java launcher can be probed (i.e. javaLanguageVersion is not UNKNOWN)
+        doNotTrackStateIf("Java launcher cannot be probed",
```
alllex, line 154: 🤔There is some sort of logical contention between this call and potential calls to `AbstractTask.doNotTrackState(reason)`.

If we were to make `doNotTrackStateIf` properly a part of the public API. Then would probably make it work like `onlyIf` works today -- appending predicates, and not replacing them.

However, I am doubtful that putting the new `doNotTrackStateIf` method on `TaskInternal` is enough to ensure it won't be used. User-defined tasks extend `AbstractTask`, so this is directly accessible. I would not be surprised if plugin authors started using it, even if accidentally. We should really spend time to get rid of this "internal" mess in tasks definitions in Gradle 10. But that's beside the point.

It might be better to make the new method more resilient from the get-go, even if we keep the new method on the `TaskInternal` for now.

platforms/jvm/language-java/src/integTest/groovy/org/gradle/api/tasks/JavaExecToolchainIntegrationTest.groovy
```diff
@@ -172,6 +175,91 @@ class JavaExecToolchainIntegrationTest extends AbstractIntegrationSpec implement
         "assigned tool" | "when configured"                    | "other"  | null           | "other"
     }
 
+    @Requires(UnitTestPreconditions.Unix)
+    def "can specify executable that cannot be probed"() {
```
alllex, line 180: ❓  Is there a test we should add to demonstrate how this integrates with Build Scans?

Most test related to that live in `JavaToolchainBuildOperationsIntegrationTest`

subprojects/core/src/main/java/org/gradle/api/internal/AbstractTask.java
ghale: Yeah, but that would mean it would be exposed as a property, which means it could be set directly.  I would rather have it be a private property and mediate access to the value through `doNotTrackState()` and `doNotTrackStateIf()` so that we can use the task mutator, etc.

platforms/jvm/toolchains-jvm-shared/src/main/java/org/gradle/jvm/toolchain/internal/JavaToolchainQueryService.java
ghale: Yeah, we'll fail the check if you request capabilities other than a launcher.  A `SpecificExectuableToolchainSpec` by definition returns a toolchain that can only be used for the `java` tool (i.e. a launcher).  If you provide one of these specs and ask for a different capability in the query (e.g. a compiler) this check will fail.  If you use `SpecificExectuableToolchainSpec` and don't specify any capabilities at query time (i.e. you only want a launcher), this check passes.  If you then try to ask the resultant toolchain for a different tool (e.g. `javac`) you'll get an error from the toolchain.

platforms/jvm/jvm-services/src/main/java/org/gradle/jvm/toolchain/JavaLanguageVersion.java
ghale: Nope.  Copilot hallucination that was folded by the IDE so I didn't notice it.  Thanks for catching this.

platforms/jvm/jvm-services/src/main/java/org/gradle/jvm/toolchain/JavaLanguageVersion.java
ghale: > Why does this have to be part of the public API?

It doesn't.  I've moved it to `DefaultJavaLanguageVersion`.

> We should make this explicit that selecting toolchains with an unknown version is not supported.

Good point.  I've made this a condition for making the spec invalid.

subprojects/core/src/main/java/org/gradle/api/internal/AbstractTask.java
ghale: Yep, done.

subprojects/core/src/test/groovy/org/gradle/api/internal/changedetection/changes/DefaultTaskExecutionModeResolverTest.groovy
ghale: I've refactored this so these still get `Optional<String>` and ultimately these changes aren't needed anymore.

platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
ghale: This is a fair point.  I've refactored this into a set of reasons that accumulate rather than replacing each other.  If we get multiple reasons for not tracking state, we will concatenate them into a single String joined with "; " and provide this similar to how it was provided before.

platforms/jvm/language-java/src/integTest/groovy/org/gradle/api/tasks/JavaExecToolchainIntegrationTest.groovy
ghale: Yeah, we should.  I've added a test to `JavaToolchainBuildOperationsIntegrationTest` and made some changes to the `DefaultJavaToolchainUsageProgressDetails` such that toolchains that have invalid metadata will display as "unknown unknown (unknown)".  This is not great (it would be better if it was something like "user-specified java executable" or even something showing the path).  We could potentially hijack something like `javaVendor` and set it to something we want displayed, but that feels pretty hacky.  If we want to improve this, I think we should tackle it as a coordinated change across build tool, build scan plugin and develocity in a follow up effort.

Note that executables that are set, but can be probed, will show up in build operations and build scans just like other toolchains.  So we see it displayed as "unknown" _only_ when a specific executable is provided _and_ we cannot probe the executable for some reason.

ghale:
@bot-gradle test this

platforms/jvm/toolchains-jvm-shared/src/main/java/org/gradle/jvm/toolchain/internal/JavaToolchainQueryService.java
ghale: A nuance of this is that if you provide a `SpecificExectuableToolchainSpec` that _can_ be probed, then it will behave like a "normal" toolchain (i.e. you can request capabilities and it will query the metadata for those capabilities and if you ask the resulting toolchain for a different tool, it could potentially provide it, if it has that capability).

You could imagine us building off of this to allow tasks that use different capabilities to specify executables that can't be probed (e.g. JavaCompile or Javadoc tasks).  In other words, they could return toolchains that allow you to request the "compiler" or "javadoc" capabilities and can only provide those tools.  However, I think JavaExec is a more common case and the only one we should worry about right now.  If we want to expand this non-probe-able functionality to other tools, I think we should consider providing a way to specify custom toolchains (along with the capabilities they provide) and retire the `executable` and `javaHome` properties on these tasks (i.e. you specify the custom toolchain some place like `settings.gradle` and then select it like any other toolchain).

platforms/jvm/language-java/src/integTest/groovy/org/gradle/api/tasks/JavaExecToolchainIntegrationTest.groovy
alllex: That sounds like a good enough approach for now. I think it makes sense to check in with the Build Scans team even before we merge this to share how this is going to work and if they have any input. But as you said, we don't have to address it immediately, since it's a corner case.

platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
```diff
@@ -147,6 +148,13 @@ public JavaExec() {
...
+            task -> {
+                return javaLauncher.map(launcher -> launcher.getMetadata().getLanguageVersion()).get() == DefaultJavaLanguageVersion.UNKNOWN;
```
alllex: 💅  Could be a body-less lambda

platforms/jvm/toolchains-jvm-shared/src/main/java/org/gradle/jvm/toolchain/internal/SpecificExecutableToolchainSpec.java
```diff
@@ -0,0 +1,97 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean isValid() {
+        return true;
+    }
+
+    public File getJavaHome() {
+        return javaExecutable.getParentFile().getParentFile();
```
alllex, line 84: 💭 This is probably worth a comment about expecting the `<JAVA_HOME>/bin/java` layout.

Do we validate that expectation somewhere?

platforms/jvm/toolchains-jvm-shared/src/main/java/org/gradle/jvm/toolchain/internal/JavaToolchainQueryService.java
alllex: I agree with all your points. I am still missing some bit of understanding, which is probably me not understanding the implementation enough.

If the user provided `SpecificExectuableToolchainSpec` and it cannot be probed, the `metadata` in this method will be an instance of `FailureInstallationMetadata`. This means it has `isValidInstallation() == false` and also `getCapabilities() == Sets.empty()`.

If I read this right, the first `if` will not trigger, because we safe guarded against the `SpecificExectuableToolchainSpec` explicitly. That means we get to the second `if`, which will have to always trigger, since the `metadata.getCapabilities()` must be empty, and throw an exception.

I guess, another option is that the `requiredCapabilities` is also empty. Is that what we are relying on here?

platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
alllex: > we will concatenate them into a single String joined with "; " and provide this similar to how it was provided before.

This would work, but feels hacky in terms of the resulting string. What do we do in other places, when user-provided reasons need to be concatenated?

This is a corner case and would rarely trigger, so we don't have to figure it out in this PR. Though, I would love us to eventually find a solution that is not just `;` concatenation, but something more clean.

platforms/jvm/language-java/src/integTest/groovy/org/gradle/api/tasks/JavaToolchainBuildOperationsIntegrationTest.groovy
```diff
@@ -556,6 +556,83 @@ class JavaToolchainBuildOperationsIntegrationTest extends AbstractIntegrationSpe
         output.contains(jdkMetadata.javaHome.toString())
     }
 
+    @Requires(UnitTestPreconditions.Unix)
+    def "emits toolchain usages for JavaExec task with configured executable that cannot be probed"() {
```
alllex, line 560: 👍  It's good that we capture this explicitly.

Having `unknown` for all Toolchain properties in a Build Scan is not ideal, but we can improve it in a follow-up.

subprojects/core/src/main/java/org/gradle/api/internal/TaskInternal.java
```diff
@@ -64,6 +64,23 @@ public interface TaskInternal extends Task, Configurable<Task> {
...
+    Set<String> getReasonsNotToTrackState();
+
+    /**
+     * Sets a condition to not track the task state when the given spec matches.  This condition will be evaluated when the return
+     * value of {@link #getReasonNotToTrackState()} is queried.
+     *
+     * @param reason - the reason for not tracking the task state when the condition matches
+     * @param spec - the condition to evaluate
+     */
+    // TODO We could consider making this a part of the public API
```
alllex: 💅  We can probably drop this TODO from the source code

subprojects/core/src/main/java/org/gradle/execution/ProjectExecutionServices.java
```diff
@@ -127,8 +127,8 @@ TaskExecuter createTaskExecuter(
             missingTaskDependencyDetector
         );
         executer = new ProblemsTaskPathTrackingTaskExecuter(executer);
-        executer = new FinalizePropertiesTaskExecuter(executer);
         executer = new ResolveTaskExecutionModeExecuter(repository, executer);
+        executer = new FinalizePropertiesTaskExecuter(executer);
```
alllex, line 131: ❔  Why is this required? Why side effects does this change have?

alllex (approved):
Looks good to me overall!

Left some more suggestions and comments

platforms/jvm/toolchains-jvm-shared/src/main/java/org/gradle/jvm/toolchain/internal/JavaToolchainQueryService.java
ghale: The second `if` doesn't trigger when `metatadata.getCapabilities()` is empty - it triggers because it doesn't contain all of the _requested_ capabilities.  The way this works now, you _cannot_ request additional capabilities when the jvm can't be probed.  The only thing you can do is request a toolchain with no capabilities (i.e. that only provides a launcher).  So in the case where you are just requesting a launcher (the only valid case with an executable that cannot be probed), your requested capabilities are empty and the capabilities from the metadata are empty, so this condition does not trigger.

If your question is "why isn't launching also a capability?", I don't know, this is how I found it.  I suspect it's more due to organic growth than a concrete reason.

alllex: Now it clicked, thank you for the detailed explanations!

subprojects/core/src/main/java/org/gradle/execution/ProjectExecutionServices.java
ghale: I don't think it's _required_, but it was a concern raised by @big-guy that the conditional might be querying the property before it's finalized (e.g. I suppose you could do something sketchy like change the value of a property in the `doNotTrackStateIf()` spec).  It doesn't seem to have side effects that I can tell, and it seems technically more "correct" that we should finalize any properties before resolving the execution mode, so I changed it.  

Does this change concern you at all?

alllex: I don't have specific concerns, I'm not deeply familiar with these specific executer steps. But the change stood out as seemingly unrelated to the rest of the PR, so I wanted to make sure it was intentionally part of the changeset.

platforms/jvm/toolchains-jvm-shared/src/main/java/org/gradle/jvm/toolchain/internal/SpecificExecutableToolchainSpec.java
ghale: It's not an expectation, but if the executable is in a valid JDK, this allows us to get the java home and resolve it to a "normal" toolchain.  I've added a comment to this effect.

platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
ghale: Yep.  Changed.

subprojects/core/src/main/java/org/gradle/api/internal/TaskInternal.java
ghale: I only added it to answer the future question of "why isn't this a part of the public api?".  I suppose it's noisy.  Removed.

PR #35684, opened 2025-11-14
Decouple ProjectExecutionServices from mutable Project instance

jvandort:
Currently, nodes in an execution graph can provide their owning project, as context. Services from this owning project are used to build a ProjectExecutionServices, which then provides services that nodes can use during execution.
    
In practice, most services that are required for task execution do not depend on project state. Only a small subset of services are actually derived from project state.
    
As part of Isolated Projects, we want to avoid passing around and handling mutable project instances. By having the task execution services derive from the project's service registry, we run the risk of access project-scoped services without capturing the correct resources.
    
This commit updates ProjectExecutionServices to explicitly accept the project-scoped state it needs to be initialized. This way, we can build the execution services based off of the build scoped services, and allow the execution services to be customized using the small subset of services that a project uses.
    
Eventually, we may be able to avoid caching ProjectExecutionServices instances on the project itself, improving scalability, as we might not necessarily need an instance of these services per project.
    
In any case, this commit moves us towards the goal of removing references to the Project interface from the execution engine, so that the engine can live outside of the context of a "project".

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/internal/service/scopes/BuildScopeServices.java
+++ b/subprojects/core/src/main/java/org/gradle/internal/service/scopes/BuildScopeServices.java
@@ -138,6 +138,7 @@ import org.gradle.execution.plan.DefaultNodeExecutor;
 import org.gradle.execution.plan.DefaultNodeValidator;
 import org.gradle.execution.plan.ExecutionNodeAccessHierarchies;
 import org.gradle.execution.plan.ExecutionPlanFactory;
+import org.gradle.execution.plan.MissingTaskDependencyDetector;
 import org.gradle.execution.plan.NodeValidator;
 import org.gradle.execution.plan.OrdinalGroupFactory;
 import org.gradle.execution.plan.PlanExecutor;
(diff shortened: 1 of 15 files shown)


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/internal/service/scopes/BuildScopeServices.java
```diff
@@ -342,6 +343,11 @@ ExecutionNodeAccessHierarchies createExecutionNodeAccessHierarchies(FileSystem f
         return new ExecutionNodeAccessHierarchies(fileSystem.isCaseSensitive() ? CaseSensitivity.CASE_SENSITIVE : CaseSensitivity.CASE_INSENSITIVE, stat);
     }
 
+    @Provides
+    MissingTaskDependencyDetector createMissingTaskDependencyDetector(ExecutionNodeAccessHierarchies hierarchies) {
+        return new MissingTaskDependencyDetector(hierarchies.getOutputHierarchy(), hierarchies.createInputHierarchy());
```
alllex, line 348: ❌ The instance of the input hierarchy is now implicitly build-scoped and lives across projects, but it's not cleared like output and destroyable hierarchies. See `MissingTaskDependencyDetector` implementation.

PR #35703, opened 2025-11-17
Extract 'testJavaVersion' param

blindpirate:
Closes https://github.com/gradle/gradle-private/issues/4931

This PR creates a param `testJavaVersion`, which could be overwritten via TeamCity parameter.

### DIFF
--- a/.teamcity/src/main/kotlin/configurations/FunctionalTestsPass.kt
+++ b/.teamcity/src/main/kotlin/configurations/FunctionalTestsPass.kt
@@ -17,6 +17,7 @@
 package configurations
 
 import common.applyDefaultSettings
+import jetbrains.buildServer.configs.kotlin.ParameterDisplay
 import model.CIBuildModel
 import model.TestCoverage
 import projects.FunctionalTestProject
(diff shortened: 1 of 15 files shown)


.teamcity/src/main/kotlin/util/AdHocPerformanceScenario.kt
```diff
@@ -104,7 +97,6 @@ abstract class AdHocPerformanceScenario(
                             os,
                             arch,
                             "%testJavaVersion%",
-                            "%testJavaVendor%",
```
cobexer, line 107: ❓ Is this dropping the wrong arg here?

blindpirate: Right, that's a typo, fixed!

blindpirate:
@bot-gradle squash

blindpirate:
@bot-gradle test

.teamcity/src/main/kotlin/configurations/FunctionalTestsPass.kt
```diff
@@ -34,6 +35,16 @@ class FunctionalTestsPass(
         dependencies {
             snapshotDependencies(functionalTestProject.functionalTests)
         }
+
+        params {
+            text(
+                "reverse.dep.*.testJavaVersion",
```
blindpirate, line 41: Added this to allow trigger a set of builds with overwritten `testJavaVersion`.

cobexer:
@blindpirate looks like getting this to work is way harder than moving the additional arguments to a better place in the command lines, let's stop trying to get this to work and close this PR

PR #35712, opened 2025-11-17
Additional support for Develocity URL injection

ljacomet:
This PR builds on top of #35059

In addition, it adds optional support for doing `BuildOption` configuration from an environment variable.

### DIFF
--- a/subprojects/core/src/integTest/groovy/org/gradle/initialization/GradleHelpIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/initialization/GradleHelpIntegrationTest.groovy
@@ -51,7 +51,7 @@ USAGE: gradle [option...] [task...]
 -d, --debug                        Log in debug mode (includes normal stacktrace).
 --daemon                           Uses the Gradle daemon to run the build. Starts the daemon if not running.
 --no-daemon                        Do not use the Gradle daemon to run the build. Useful occasionally if you have configured Gradle to always run with the daemon by default.
---develocity-url                   Specify the URL of the Develocity server to use for Build Scans.
+--develocity-default-url           Specify the default URL of the Develocity server to use for Build Scans. If the Develocity plugin is configured, this value is ignored.
 --export-keys                      Exports the public keys used for dependency verification.
 -F, --dependency-verification      Configures the dependency verification mode. Values are 'strict', 'lenient' or 'off'.
 --foreground                       Starts the Gradle daemon in the foreground.
--- a/subprojects/core/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
+++ b/subprojects/core/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
@@ -363,13 +363,13 @@ public class StartParameterBuildOptions extends BuildOptionSet<StartParameterInt
     }
 
     public static class DevelocityUrlOption extends StringBuildOption<StartParameterInternal> {
-        public static final String LONG_OPTION = "develocity-url";
-        public static final String GRADLE_PROPERTY = "com.gradle.develocity.url";
-        public static final String ENVIRONMENT_VARIABLE = "COM_GRADLE_DEVELOCITY_URL";
+        public static final String LONG_OPTION = "develocity-default-url";
+        public static final String GRADLE_PROPERTY = "com.gradle.develocity.default.url";
+        public static final String ENVIRONMENT_VARIABLE = "COM_GRADLE_DEVELOCITY_DEFAULT_URL";
 
         public DevelocityUrlOption() {
             super(GRADLE_PROPERTY, CommandLineOptionConfiguration.create(LONG_OPTION,
-                "Specify the URL of the Develocity server to use for Build Scans."));
+                "Specify the default URL of the Develocity server to use for Build Scans. If the Develocity plugin is configured, this value is ignored."));
         }
 
         @Override
--- a/platforms/enterprise/enterprise/src/integTest/groovy/org/gradle/internal/enterprise/DevelocityPluginConfigIntegrationTest.groovy
+++ b/platforms/enterprise/enterprise/src/integTest/groovy/org/gradle/internal/enterprise/DevelocityPluginConfigIntegrationTest.groovy
@@ -84,7 +84,7 @@ class DevelocityPluginConfigIntegrationTest extends AbstractIntegrationSpec {
 
     def "is configured when passing a Develocity URL via environment variable"() {
         given:
-        executer.withEnvironmentVars(['COM_GRADLE_DEVELOCITY_URL': 'https://example.com'])
+        executer.withEnvironmentVars(['COM_GRADLE_DEVELOCITY_DEFAULT_URL': 'https://example.com'])
 
         when:
         succeeds "t"
@@ -99,7 +99,7 @@ class DevelocityPluginConfigIntegrationTest extends AbstractIntegrationSpec {
     def "is configured when passing a Develocity URL via properties"() {
         given:
         propertiesFile << """
-            com.gradle.develocity.url=https://example.com
+            com.gradle.develocity.default.url=https://example.com
         """
 
         when:
@@ -114,7 +114,7 @@ class DevelocityPluginConfigIntegrationTest extends AbstractIntegrationSpec {
 
     def "is configured when passing a Develocity URL on the CLI"() {
         when:
-        succeeds "t", "-Dcom.gradle.develocity.url=https://example.com"
+        succeeds "t", "--develocity-default-url=https://example.com"
 
         then:
         plugin.assertBuildScanRequest(output, NONE)
@@ -126,7 +126,7 @@ class DevelocityPluginConfigIntegrationTest extends AbstractIntegrationSpec {
     def "is suppressed when configuring a Develocity URL via properties and passing --no-scan"() {
         given:
         propertiesFile << """
-            com.gradle.develocity.url=https://example.com
+            com.gradle.develocity.default.url=https://example.com
         """
 
         when:
@@ -144,7 +144,7 @@ class DevelocityPluginConfigIntegrationTest extends AbstractIntegrationSpec {
         settingsFile << plugin.plugins()
 
         when:
-        succeeds "t", "-Dcom.gradle.develocity.url=https://example.com"
+        succeeds "t", "--develocity-default-url=https://example.com"
 
         then:
         plugin.assertAutoApplied(output, false)
@@ -175,7 +175,7 @@ class DevelocityPluginConfigIntegrationTest extends AbstractIntegrationSpec {
         """
 
         when:
-        succeeds "t", "-Dcom.gradle.develocity.url=https://example.com", "--init-script", initScript.absolutePath
+        succeeds "t", "--develocity-default-url=https://example.com", "--init-script", initScript.absolutePath
 
         then:
         plugin.assertAutoApplied(output, true)
@@ -206,7 +206,7 @@ class DevelocityPluginConfigIntegrationTest extends AbstractIntegrationSpec {
         """
 
         when:
-        succeeds "t", "-Dcom.gradle.develocity.url=https://example.com", "--init-script", initScript.absolutePath
+        succeeds "t", "--develocity-default-url=https://example.com", "--init-script", initScript.absolutePath
 
         then:
         plugin.assertAutoApplied(output, false)


ljacomet:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
```diff
@@ -360,6 +362,30 @@ public void applyTo(boolean value, StartParameterInternal settings, Origin origi
         }
     }
 
+    public static class DevelocityUrlOption extends StringBuildOption<StartParameterInternal> {
+       public static final String LONG_OPTION = "develocity-url";
+        public static final String GRADLE_PROPERTY = "org.gradle.develocity.url";
+        public static final String ENVIRONMENT_VARIABLE = "ORG_GRADLE_DEVELOCITY_URL";
```
big-guy: I think the open questions here are:

- Do we want a CLI option at all? I don't think we thought deeply about it
- Do we want to use `org.gradle` or `com.gradle` as the prefix for a non-OSS option?
- (Same question) Do we want to use `ORG_GRADLE` or `COM_GRADLE` as the prefix?

If we do keep a CLI option (for discoverability), maybe we would want to segregate the DV options

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DefaultGradleEnterprisePluginAdapter.java
```diff
@@ -112,6 +113,11 @@ public void buildFinished(@Nullable Throwable buildFailure, List<org.gradle.inte
     }
 
     private void createPluginService() {
+        String injectedDevelocityUrl = config.getDevelocityUrl();
+        if (injectedDevelocityUrl != null) {
+            // We might want to stop setting this system property if the Develocity plugin is new enough to read the URL from the configuration directly.
+            System.setProperty("com.gradle.scan.server", injectedDevelocityUrl);
```
big-guy: Do we need to check if this is already set and do something else (warn or fail)?

platforms/enterprise/enterprise/src/integTest/groovy/org/gradle/internal/enterprise/core/BuildScanAutoApplyIntegrationTest.groovy
big-guy, line 1: Do we need tests that show how settings this property interacts with `com.gradle.scan.server`, an init script that enables DV and a build that has DV applied and configured?

subprojects/core/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
ljacomet: * I was hesitant, but having a documented option on the CLI is nice IMO.
* We should use `COM_GRADLE_` and `com.gradle` prefixes indeed

I don't think we should look into segregating the options. We did not do it for scan.

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DefaultGradleEnterprisePluginAdapter.java
ljacomet: I think we should warn but not fail. The new option takes precedence over the old one IMO, so we should just warn if they are not aligned.

platforms/enterprise/enterprise/src/integTest/groovy/org/gradle/internal/enterprise/core/BuildScanAutoApplyIntegrationTest.groovy
ljacomet: I guess we could, but we did not have tests looking at combining `--scan` and that internal URL server property.
And since our additional setup goes through the same code path, I am tempted to leave it as-is for now.

subprojects/core/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
```diff
@@ -360,6 +362,30 @@ public void applyTo(boolean value, StartParameterInternal settings, Origin origi
         }
     }
 
+    public static class DevelocityUrlOption extends StringBuildOption<StartParameterInternal> {
+       public static final String LONG_OPTION = "develocity-url";
```
alllex: 💅  whitespace

ljacomet:
@bot-gradle test this

platforms/enterprise/enterprise/src/integTest/groovy/org/gradle/internal/enterprise/core/BuildScanAutoApplyIntegrationTest.groovy
alllex: Ideally, we should add the tests. One more case with `--no-scan` should be clear as well.

But also a broader case of when a Develocity URL overrides the `server` url in the build logic as set via the Settings Develocity Extension.

ljacomet: > when a Develocity URL overrides the server url in the build logic as set via the Settings Develocity Extension

The feature added with this PR works the other way around: it cannot take precedence over the build scan being configured in the `Settings` of the build.
We need that behavior for different reasons

ljacomet: And testing proves I am wrong ... digging deeper.

subprojects/core/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
big-guy: > I don't think we should look into segregating the options. We did not do it for scan.

When there was only a single option, it was OK to be mixed in with the rest, but now that we have two options and they're related, I think this is something we could consider. 

We have ~60 options now (counting "no" options, so closer to 30). We might want to do this for other options too. 

Imagine if `gradle -h` is like this:

<details>
<pre>
USAGE: gradlew [option...] [task...]

-?, -h, --help                     Shows this help message.
-V, --show-version                 Print version info and continue.
-v, --version                      Print version info and exit.
--status                           Shows status of running and recently stopped Gradle daemon(s).

--                                 Signals the end of built-in options. Gradle parses subsequent parameters as only 

Execution
    -a, --no-rebuild                   Do not rebuild project dependencies.
    --continue                         Continue task execution after a task failure.
    --no-continue                      Stop task execution after a task failure.
    --rerun-tasks                      Ignore previously cached task results.
    -m, --dry-run                      Run the builds with all task actions disabled.
    -t, --continuous                   Enables continuous build. Gradle does not exit and will re-execute tasks when task file inputs change.
    -x, --exclude-task                 Specify a task to be excluded from execution.

Configuration
    -D, --system-prop                  Set system property of the JVM (e.g. -Dmyprop=myvalue).
    -g, --gradle-user-home             Specifies the Gradle user home directory. Defaults to ~/.gradle
    -I, --init-script                  Specify an initialization script.
    --include-build                    Include the specified build in the composite.
    --offline                          Execute the build without accessing network resources.
    -P, --project-prop                 Set project property for the build script (e.g. -Pmyprop=myvalue).
    -p, --project-dir                  Specifies the start directory for Gradle. Defaults to current directory.
    --project-cache-dir                Specify the project-specific cache directory. Defaults to .gradle in the root project directory.

Performance
    --build-cache                      Enables the Gradle build cache. Gradle will try to reuse outputs from previous builds.
    --configuration-cache              Enables the configuration cache. Gradle will try to reuse the build configuration from previous builds.
    --configure-on-demand              Configure necessary projects only. Gradle will attempt to reduce configuration time for large multi-project builds. [incubating]
    --no-configure-on-demand           Disables the use of configuration on demand. [incubating]
    --daemon                           Uses the Gradle daemon to run the build. Starts the daemon if not running.
    --no-daemon                        Do not use the Gradle daemon to run the build. Useful occasionally if you have configured Gradle to always run with the daemon by default.
    --max-workers                      Configure the number of concurrent workers Gradle is allowed to use.
    --parallel                         Build projects in parallel. Gradle will attempt to determine the optimal number of executor threads to use.
    --no-parallel                      Disables parallel execution to build projects.
    --priority                         Specifies the scheduling priority for the Gradle daemon and all processes launched by it. Values are 'normal' (default) or 'low'
    --watch-fs                         Enables watching the file system for changes, allowing data about the file system to be re-used for the next build.
    --no-watch-fs                      Disables watching the file system.

Security
    -F, --dependency-verification      Configures the dependency verification mode. Values are 'strict', 'lenient' or 'off'.
    -M, --write-verification-metadata  Generates checksums for dependencies used in the project (comma-separated list)
    --refresh-keys                     Refresh the public keys used for dependency verification.
    --update-locks                     Perform a partial update of the dependency lock, letting passed in module notations change version. [incubating]
    --write-locks                      Persists dependency resolution for locked configurations, ignoring existing locking information if it exists tasks or task options.

Diagnostics
    --configuration-cache-problems     Configures how the configuration cache handles problems (fail or warn). Defaults to fail.
    --problems-report                  (Experimental) enables HTML problems report
    --no-problems-report               (Experimental) disables HTML problems report
    --profile                          Profile build execution time and generates a report in the <build_dir>/reports/profile directory.
    --property-upgrade-report          (Experimental) Runs build with experimental property upgrade report.
    --warning-mode                     Specifies which mode of warnings to generate. Values are 'all', 'fail', 'summary'(default) or 'none'
    --task-graph                       (Experimental) Print task graph instead of executing tasks.
    -U, --refresh-dependencies         Refresh the state of dependencies.

Logging
    --console                          Specifies which type of console output to generate. Values are 'plain', 'colored', 'auto' (default), 'rich' or 'verbose'.
    -d, --debug                        Log in debug mode (includes normal stacktrace).
    -i, --info                         Set log level to info.
    -q, --quiet                        Log errors only.
    -S, --full-stacktrace              Print out the full (very verbose) stacktrace for all exceptions.
    -s, --stacktrace                   Print out the stacktrace for all exceptions.
    -w, --warn                         Set log level to warn.

Develocity
    --scan                             Generate a Build Scan (powered by Develocity).

Build Scan and Develocity are registered trademarks of Gradle, Inc.
For more information, please visit https://gradle.com/develocity/product/build-scan/.

To see help contextual to the project, use gradlew help

To see more detail about a task, run gradlew help --task <task>

To see a list of available tasks, run gradlew tasks
</pre>
</details>

This isn't a blocking issue.

ljacomet: I could have been more clear: I don't think I would mix arranging the options with the goal of this PR.

But effectively the output concept you share is compelling.

platforms/enterprise/enterprise/src/integTest/groovy/org/gradle/internal/enterprise/core/BuildScanAutoApplyIntegrationTest.groovy
big-guy: > The feature added with this PR works the other way around: it cannot take precedence over the build scan being configured in the Settings of the build.

I think this is OK. We don't have that many places where there's a property that can be set in the DSL and on the CLI and the DSL is overridden by the CLI option

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DefaultGradleEnterprisePluginAdapter.java
big-guy: I agree it should not fail

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheKey.kt
```diff
@@ -70,6 +70,12 @@ class ConfigurationCacheKey(
         putBoolean(startParameter.isOffline)
         putBoolean(startParameter.isIsolatedProjects)
         putBuildScan()
+        val develocityUrl = startParameter.develocityUrl
+        if (develocityUrl != null) {
+            putString(develocityUrl)
+        } else {
+            putNull()
+        }
```
eskatos: 💅 Introduce explaining function `putDevelocityUrl()`

platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/ProviderConnection.java
```diff
@@ -347,7 +347,13 @@ private Parameters initParams(ProviderOperationParameters operationParameters) {
...
         AllProperties properties = new LayoutToPropertiesConverter(buildLayoutFactory).convert(initialProperties, buildLayoutResult);
 
-        DaemonParameters daemonParams = new DaemonParameters(buildLayoutResult.getGradleUserHomeDir(), fileCollectionFactory, Collections.emptyMap(), operationParameters.getEnvironmentVariables(null));
+        // Either use the environment variables from the operation parameters, or inherit the current process environment
+        Map<String, String> environmentVariables = operationParameters.getEnvironmentVariables(null);
+        if (environmentVariables == null) {
+            environmentVariables = new HashMap<>(System.getenv());
+        }
+
+        DaemonParameters daemonParams = new DaemonParameters(buildLayoutResult.getGradleUserHomeDir(), fileCollectionFactory, Collections.emptyMap(), environmentVariables);
```
eskatos, line 356: ❌ The `DaemonParameters` CTOR already uses `System.getenv()` if it gets a null `environmentVariables` parameter.

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DefaultGradleEnterprisePluginAdapter.java
ljacomet: Turns out we should not use that option at all as it fully overrides whatever the user configured in their `Settings`, so pulling this out to use a different approach.

subprojects/core/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
ljacomet: Created #35771

platforms/enterprise/enterprise/src/integTest/groovy/org/gradle/internal/enterprise/core/BuildScanAutoApplyIntegrationTest.groovy
ljacomet: PR updated with tests that should match the requests, minus the ones that are no longer relevant given the changes.

ljacomet:
@bot-gradle test this

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DevelocityAutoAppliedPluginConfigurationAction.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public void execute(AppliedPlugin plugin) {
+        GradleEnterprisePluginAutoAppliedStatus autoAppliedStatus = settings.getServices().get(GradleEnterprisePluginAutoAppliedStatus.class);
+        String develocityUrl = ((StartParameterInternal) settings.getStartParameter()).getDevelocityUrl();
+        if (autoAppliedStatus.isAutoApplied() && develocityUrl != null) {
+            Object develocityConfiguration = settings.getExtensions().getByName("develocity");
+            try {
+                @SuppressWarnings("unchecked")
+                Property<String> server = (Property<@NonNull String>) develocityConfiguration.getClass().getMethod("getServer").invoke(develocityConfiguration);
```
big-guy: Cast.uncheckedCast?

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DefaultGradleEnterprisePluginConfig.java
```diff
@@ -47,10 +52,16 @@ public boolean isAutoApplied() {
...
+    @Nullable
+    public String getDevelocityUrl() {
+        return develocityUrl;
+    }
+
     private BuildScanRequest buildScanRequest(StartParameter startParameter) {
         if (startParameter.isNoBuildScan()) {
             return BuildScanRequest.SUPPRESSED;
-        } else if (startParameter.isBuildScan()) {
+        } else if (startParameter.isBuildScan() || ((StartParameterInternal) startParameter).getDevelocityUrl() != null) {
```
wolfs: 🤔 I don't think we should change this here. Why is a build scan requested when you set the Develocity URL? It is also not requested when you apply the Develocity plugin in your build script, so I'd just leave this out and rely on the Develocity plugin publishing by default.

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DevelocityAutoAppliedPluginConfigurationAction.java
```diff
@@ -0,0 +1,63 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(DevelocityAutoAppliedPluginConfigurationAction.class);
+
+    private final SettingsInternal settings;
+
+    public DevelocityAutoAppliedPluginConfigurationAction(Settings settings) {
+        this.settings = (SettingsInternal) settings;
+    }
+
+    @Override
+    public void execute(AppliedPlugin plugin) {
```
wolfs, line 45: Why are we doing this here? Use the system property I used before and add the URL to the structured data we pass to the Develocity plugin.

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/GradleEnterpriseAutoAppliedPluginRegistry.java
```diff
@@ -50,14 +50,19 @@ public PluginRequests getAutoAppliedPlugins(Settings target) {
         if (((StartParameterInternal) target.getStartParameter()).isUseEmptySettings() || !shouldApplyDevelocityPlugin(target)) {
             return PluginRequests.EMPTY;
         } else {
+            // We are going with an auto-application request, let's configure the URL
+            // TODO Remove this once the default applied version supports DefaultGradleEnterprisePluginConfig.getDevelocityUrl()
+            target.getPluginManager().withPlugin(AutoAppliedDevelocityPlugin.ID.getId(), new DevelocityAutoAppliedPluginConfigurationAction(target));
```
wolfs, line 55: Like I said above, use the system property. If you do this here you would also configure the plugin when the build applies and configures it already.

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/GradleEnterpriseAutoAppliedPluginRegistry.java
```diff
@@ -50,14 +50,19 @@ public PluginRequests getAutoAppliedPlugins(Settings target) {
...
     }
 
     private static boolean shouldApplyDevelocityPlugin(Settings settings) {
         Gradle gradle = settings.getGradle();
-        StartParameter startParameter = gradle.getStartParameter();
-        return startParameter.isBuildScan() && gradle.getParent() == null;
+        StartParameterInternal startParameter = (StartParameterInternal) gradle.getStartParameter();
+        return (startParameter.isBuildScan()
+                || (!Strings.isNullOrEmpty(startParameter.getDevelocityUrl()) && !startParameter.isNoBuildScan()))
+            && gradle.getParent() == null;
```
wolfs, line 65: I think we should still apply the Develocity plugin even if you pass --no-scan. So we keep the classpath the same and don't have to re-configure everything just because we pass --no-scan.

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DevelocityAutoAppliedPluginConfigurationAction.java
ljacomet: No, because that property is a full overwrite. If you configure the DV plugin, the value set there is ignored in favor of the value put in the system property.

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/GradleEnterpriseAutoAppliedPluginRegistry.java
ljacomet: My testing (and the integration tests) show this differently: because this gets executed only if we "auto-applied", it is safer.

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/GradleEnterpriseAutoAppliedPluginRegistry.java
ljacomet: AFAICT both options have the same user facing behavior.
In both cases, you get a CC miss when the only different in an invocation is `--no-scan` being present or not.

But I understand the point about applying the plugin since you are effectively requesting it when passing the Develocity URL option.

Will update the PR.

ljacomet:
@bot-gradle test this

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DevelocityAutoAppliedPluginConfigurationAction.java
wolfs: Yes, and that is the point of the environment variable, right?

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/GradleEnterpriseAutoAppliedPluginRegistry.java
wolfs: > In both cases, you get a CC miss when the only different in an invocation is --no-scan being present or not.

Why, is that something built into CC?

platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/ProviderStartParameterConverter.java
```diff
@@ -55,7 +56,7 @@ private List<TaskExecutionRequest> unpack(final List<InternalLaunchable> launcha
         return requests;
     }
 
-    public StartParameterInternal toStartParameter(ProviderOperationParameters parameters, BuildLayoutResult buildLayout, AllProperties properties) {
+    public StartParameterInternal toStartParameter(ProviderOperationParameters parameters, BuildLayoutResult buildLayout, AllProperties properties, Map<String, String> environmentVariables) {
```
big-guy, line 59: Not now, but should AllProperties be carrying environment variables too?

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DevelocityAutoAppliedPluginConfigurationAction.java
big-guy: We only wanted to use the DV URL from the env var if the build had DV auto-applied and the user had not configured the URL.  Otherwise, working builds could suddenly start publishing their build scans elsewhere

ljacomet: I have renamed the property and its description to clarify this.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheKey.kt
```diff
@@ -91,6 +92,16 @@ class ConfigurationCacheKey(
         )
     }
 
+    private
+    fun Hasher.putDevelocityUrl() {
+        val develocityUrl = startParameter.develocityUrl
+        if (develocityUrl != null) {
+            putString(develocityUrl)
+        } else {
+            putNull()
```
bamboo: 🚀 `putNull` is unnecessary, the absence of `putString` is enough to change the hash

platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/ProviderStartParameterConverter.java
bamboo: If not now, when? 😉

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DevelocityAutoAppliedPluginConfigurationAction.java
```diff
@@ -0,0 +1,62 @@
...
+import org.gradle.api.plugins.AppliedPlugin;
+import org.gradle.api.provider.Property;
+import org.gradle.internal.Cast;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Configures the Develocity plugin if it was auto-applied.
+ * <p>
+ * This is a temporary solution until the default applied version understands the added develocity URL configuration option.
```
bamboo, line 32: ❓ Do we have an issue tracking that?

platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/ProviderStartParameterConverter.java
ljacomet: I created #35850. There are things to clarify since there is `AllProperties.merge` so I don't think it needs to happen right now.

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DevelocityAutoAppliedPluginConfigurationAction.java
ljacomet: I started the conversation with the scans team on this and will follow up. Created #35851 to track this.

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/BuildOptionSet.java
```diff
@@ -67,4 +67,16 @@ public T convert(Map<String, String> properties, T target) throws CommandLineArg
             }
         };
     }
+
+    public PropertiesConverter<T> envVarConverter() {
+        return new PropertiesConverter<T>() {;
+            @Override
+            public T convert(Map<String, String> envVars, T target) throws CommandLineArgumentException {
+                for (BuildOption<? super T> option : getAllOptions()) {
+                        option.applyFromEnvVar(envVars, target);
```
bamboo, line 76: 💅 Formatting seems off here

subprojects/core/src/integTest/groovy/org/gradle/initialization/GradleHelpIntegrationTest.groovy
```diff
@@ -51,6 +51,7 @@ USAGE: gradle [option...] [task...]
 -d, --debug                        Log in debug mode (includes normal stacktrace).
 --daemon                           Uses the Gradle daemon to run the build. Starts the daemon if not running.
 --no-daemon                        Do not use the Gradle daemon to run the build. Useful occasionally if you have configured Gradle to always run with the daemon by default.
+--develocity-default-url           Specify the default URL of the Develocity server to use for Build Scans. If the Develocity plugin is configured, this value is ignored.
```
wolfs, line 54: ❌  This doesn't make any sense to me. The whole point is here IMO to definitely publish to that Develocity server when it is configured, no matter what the build says. I don't see a good use-case where you would want to overwrite the URL in the build. As soon as the Develocity plugin picks up the location from the configuration, it will also overwrite the URL configured in the DSL, won't it? I don't see why it shouldn't.

ljacomet: After discussion, both paths have merits but they also have impacts on feature evolution. As such, we are postponing adding this feature until we commit on a path based on how we see it evolving.

ljacomet:
This is reverted by #35861

PR #35718, opened 2025-11-17
Polish `CompositeStoppable`

bamboo:
- Avoid adding NO_OP stoppables
- Modernize code

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/CompositeStoppable.java
+++ b/platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/CompositeStoppable.java
@@ -70,7 +70,7 @@ public class CompositeStoppable implements Stoppable {
 
     public CompositeStoppable add(Object closeable) {
         Stoppable stoppable = toStoppable(closeable);
-        if (null != stoppable) {
+        if (stoppable != null) {
             add(stoppable);
         }
         return this;


platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/CompositeStoppable.java
```diff
@@ -75,11 +68,20 @@ public CompositeStoppable add(Object... elements) {
         return this;
     }
 
-    public synchronized CompositeStoppable add(Object closeable) {
-        this.elements.add(toStoppable(closeable));
+    public CompositeStoppable add(Object closeable) {
+        Stoppable stoppable = toStoppable(closeable);
+        if (null != stoppable) {
```
bamboo: ℹ️ This is the main change

alllex: 💅 
```suggestion
        if (stoppable != null) {
```

feels more conventional

alllex (approved):

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/CompositeStoppable.java
bamboo: Right, funny side-effect of inlining

bamboo:
@bot-gradle test QFL

Issue #35570, opened 2025-11-05
Gradle 9.3.0-milestone-1 JUnit Suite fails


bmorris591:
### Current Behavior

Running Gradle 9.3.0-milestone-1 with `org.junit.platform:junit-platform-suite:6.0.1` and `io.cucumber:cucumber-bom:7.31.0` I get the following exception 

```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':volvooncall-mysql:integrationTest'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:135)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:288)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:133)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:121)
        at org.gradle.api.internal.tasks.execution.ProblemsTaskPathTrackingTaskExecuter.execute(ProblemsTaskPathTrackingTaskExecuter.java:41)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
        at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:74)
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
[...]

PR #35720, opened 2025-11-17
Support multiple display names for a container

octylFractal:
This apparently comes up in some cases, so we need to handle it gracefully instead of giving a hard fail. See #35570

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
@@ -48,6 +48,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 /**
@@ -329,6 +330,34 @@ public final class SerializableTestResultStore {
          * @throws IOException if an error occurs while processing the result
          */
         void process(long id, @Nullable Long parentId, SerializableTestResult result, OutputRanges outputRanges) throws IOException;
+    }
+
+    /**
+     * Exists for backwards compatibility with TestFilesCleanupService.
+     *
+     * Mirrors the old structure of OutputTrackedResult that was used by TestFilesCleanupService.
+     * Doesn't need to be the original class since the type wasn't explicitly referenced.
+     */
+    public static final class FacadeForOutputTrackedResult {
+        private final SerializableTestResult innerResult;
+
+        public FacadeForOutputTrackedResult(SerializableTestResult innerResult) {
+            this.innerResult = innerResult;
+        }
+
+        public SerializableTestResult getInnerResult() {
+            return innerResult;
+        }
+    }
+
+    /**
+     * Exists for backwards compatibility with TestFilesCleanupService.
+     */
+    @SuppressWarnings("unused")
+    public void forEachResult(Consumer<FacadeForOutputTrackedResult> consumer) throws IOException {
+        forEachResult((id, parentId, result, outputRanges) ->
+            consumer.accept(new FacadeForOutputTrackedResult(result))
+        );
     }
 
     /**


octylFractal:
@bot-gradle test this

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericPageRenderer.java
```diff
@@ -146,16 +147,28 @@ protected ReportRenderer<TestTreeModel, SimpleHtmlWriter> getContentRenderer() {
...
+                    }
+                }
+                if (hasStdout) {
+                    perRootInfoTabsRenderer.add("standard output", new PerRootTabRenderer.ForOutput(rootIndex, perRootInfoIndex, outputReader, TestOutputEvent.Destination.StdOut));
+                }
+                if (hasStderr) {
+                    perRootInfoTabsRenderer.add("error output", new PerRootTabRenderer.ForOutput(rootIndex, perRootInfoIndex, outputReader, TestOutputEvent.Destination.StdErr));
                 }
-                if (!info.getMetadatas().isEmpty()) {
+                if (info.getMetadatas().iterator().hasNext()) {
```
tresat: 💅 Why not `!isEmpty()`?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResult.java
```diff
@@ -37,6 +37,13 @@
  * </p>
  */
 public final class SerializableTestResult {
+    public static String getCombinedDisplayName(List<SerializableTestResult> testResults) {
+        return testResults.stream()
+            .map(SerializableTestResult::getDisplayName)
+            .distinct()
+            .collect(Collectors.joining(" / "));
```
tresat, line 44: 🤔 This seems like the resulting name could get long and unwieldy pretty quickly.

Since we're only calling this when we render, it seems like we've got better UX options.

We could do something like: return a collection of these names, display the first one initially along with have a `[...]` button next to it that would toggle show/hide the others somehow.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -199,33 +203,44 @@ private static void renderSummaryGroup(PerRootInfo info, SimpleHtmlWriter htmlWr
...
-        private static String getFormattedDuration(SerializableTestResult testResult) {
-            return TimeFormatting.formatDurationVeryTerse(testResult.getDuration());
+        private static String getFormattedDuration(PerRootInfo info) {
+            long minStartTime = Long.MAX_VALUE;
+            long maxEndTime = Long.MIN_VALUE;
+            for (SerializableTestResult result : info.getResults()) {
+                minStartTime = Math.min(minStartTime, result.getStartTime());
+                maxEndTime = Math.max(maxEndTime, result.getEndTime());
+            }
+            return TimeFormatting.formatDurationVeryTerse(maxEndTime - minStartTime);
```
tresat: ❗ This doesn't seem right.  If there are multiple results, and one started at t=1 and took 5 seconds (running until t=6), and another started at t=2 and ran 8 seconds (until t=10), this would result in a displayed duration of 9 seconds...which isn't anything.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
```diff
@@ -330,6 +348,33 @@ public void render(TestTreeModel model, SimpleHtmlWriter htmlWriter) throws IOEx
...
 
+        private static TestResult.ResultType getResultType(PerRootInfo info) {
+            if (info.getChildren().isEmpty()) {
+                // There should only be one result for leaf nodes
+                if (info.getResults().size() > 1) {
+                    throw new IllegalStateException("Leaf nodes should only have one result");
+                }
+                return info.getResults().get(0).getResultType();
+            }
+            // For container nodes, merge result types, with FAILURE > SUCCESS > SKIPPED.
```
tresat, line 355: 👍

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/TestTreeModel.java
```diff
@@ -120,7 +120,7 @@ public static TestTreeModel loadModelFromStores(List<SerializableTestResultStore
         return rootBuilder.build();
     }
 
-    private static final class StoreLoader implements Consumer<OutputTrackedResult> {
+    private static final class StoreLoader implements SerializableTestResultStore.ResultProcessor {
```
tresat, line 123: 👍

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/OutputRanges.java
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Opaque object used to search for test output associated with a specific test result.
+ *
+ * <p>
+ * Internally stores information that allows narrowing the location of output of a specific test result in an
+ * output events file.
+ * </p>
+ */
+public final class OutputRanges {
```
tresat, line 34: 🤔 I think this name is confusing - it sounds like a range type, like `[1-5]` or something.  

How about `OutputIndices`?  `OutputListing`?  `OutputCatalog`?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/TestTreeModelResultsProvider.java
```diff
@@ -176,25 +188,51 @@ private static TestClassResult createEmptyClassResult(TestTreeModel groupingNode
...
+                );
+            }
+            // This can happen for a variety of cases, e.g. parameterized tests or multiple test frameworks.
+            // Therefore, we shouldn't fail, as a slightly broken report is better than a task failure.
+            if (!result.getDisplayName().equals(displayName)) {
+                LOGGER.info(
+                    "Expected all results for grouping node {} to have the same display name, but found: {} and {}",
+                    groupingNode.getPath(), displayName, result.getDisplayName()
+                );
+            }
```
tresat, line 212: 🤔 What do you think about logging this as a warning instead?  No one will see it as info.  It does seem to indicate a single test task is "doing too much" to me - the build should split the tests and TestEngines into multiple tasks.  The preferable situation is no duplicates.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/TestTreeModelResultsProvider.java
```diff
@@ -176,25 +188,51 @@ private static TestClassResult createEmptyClassResult(TestTreeModel groupingNode
...
+            leaf.getId(),
             result.getName(),
             result.getDisplayName(),
             result.getResultType(),
             result.getDuration(),
             result.getEndTime(),
-            perRootInfo.getMetadatas()
+            ImmutableList.copyOf(leaf.getMetadatas())
         );
         methodResult.getFailures().addAll(result.getFailures());
```
tresat, line 237: 💅 Failures and metadatas should probably be treated the same re: defensive copying.

tresat (changes requested):

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericPageRenderer.java
octylFractal: It returns an `Iterable`, not a `Collection`, so that's not possible. I can use `!Iterables.isEmpty` if you want?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResult.java
octylFractal: I don't think I care that much about the UX when this is a very niche situation. Adding the button would be quite a lot of complexity, and possibly look weird in title situations.

I don't expect there to be more than one name 99% of the time.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
octylFractal: What would you prefer? I could use the longest duration, but that doesn't seem any better. I could do a `/`-joined thing like with display names. Or I could sum all durations?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/OutputRanges.java
octylFractal: It is a range type? It happens to have two start locations instead of one, but that is kind of an optimization over creating two `OutputRange` objects, one for stdout and one for stderr. So as a type that encapsulates two ranges (stdout and stderr), I think it is "output ranges".

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/TestTreeModelResultsProvider.java
octylFractal: Sure, I can make it a warning. The test I have in mind for this wouldn't be splittable though, so it's possible someone will ask us to tone it down / turn it off in the future. I'm fine with pushing it up until then.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/TestTreeModelResultsProvider.java
octylFractal: I'd prefer to do that in a separate PR, it's more involved and requires refactoring tests.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericPageRenderer.java
octylFractal: I did so.

octylFractal:
I've added a test using Cucumber that demonstrates the result.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
tresat: I think the /-joined string makes the most sense.  These runs could have overlapped on parallel workers, so summing them is possibly misleading.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/OutputRanges.java
tresat: If keeping the name, then perhaps revise the javadoc?  It doesn't seem right to call it opaque when the fields are accessible.  

It's a little weird that this stores the offsets only, and not the file those offsets index into, making this type on its own unusable without knowledge of _which_ file it references.  IDK, I guess I just found this type a little confusing when I was first reading this PR.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResult.java
tresat: Yeah, I guess it's not worth optimizing for such an unlikely case then.

Can you test with a few absurdly long names and make sure the page is rendered well enough, with appropriate scrolling, to read them all?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/TestTreeModelResultsProvider.java
tresat: Making an immutable copy of the failures list would invalidate tests?

octylFractal: I would need to move the list to the constructor; adjust any tests that are using `getFailures().add[All]`, `addFailure`, or `setAssumptionFailure` (to make `TestMethodResult` fully immutable); and finally either add a testing-only constructor which doesn't require a list for failures (and metadata?) or add the list to each test that creates it.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/PerRootTabRenderer.java
octylFractal: Done.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/OutputRanges.java
octylFractal: I rewrote it, WDYT?

tresat: Much better!

tresat (approved):

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResult.java
octylFractal: It ends up looking like this, which I think is fine.

cobexer:
@bot-gradle test RfN

octylFractal:
@bot-gradle test RfN

octylFractal:
@bot-gradle test this

PR #35724, opened 2025-11-18
Disable OOM check for docsTest on all OSes

blindpirate:


blindpirate:
This is observed in https://builds.gradle.org/buildConfiguration/Gradle_Release_Check_DocsTest_Linux/105531471

Issue #35545, opened 2025-11-03
KotlinDslBaseScriptModel#getScriptTemplatesClassPath() is insufficient to load script templates in Intellij IDEA


imbananko:
### Current Behavior

```
gradle-core-api-9.2.0.jar
gradle-kotlin-dsl-9.2.0.jar
gradle-kotlin-dsl-shared-runtime-9.2.0.jar
gradle-kotlin-dsl-tooling-models-9.2.0.jar
kotlin-script-runtime-2.2.20.jar
``` 
When running [loadDefinitionsFromTemplates](https://github.com/JetBrains/intellij-community/blob/feccb219c4e9bf36b3b4e76bbb3e7c93f097556e/plugins/kotlin/base/scripting/scripting.shared/src/org/jetbrains/kotlin/idea/core/script/shared/definition/scriptDefinitionLoading.kt#L43) with jars mentioned above - I do get the following exception:
```
"java.lang.NoClassDefFoundError: org/gradle/api/HasImplicitReceiver"
```

### Expected Behavior

Successfully loaded script templates

### Context (optional)

Seems like `gradle-api-9.2.0-rc-1.jar` is missing. But it's still presented in `KotlinDslBaseScriptModel#getCompileClassPath()`

### Self-contained Reproducer Project

I'm gonna use the workaround with `KotlinDslBaseScriptModel#getCompileClassPath()` and push the change into IDEA master branch, so I'll update this section later.

### Gradle version

9.2.0-rc-1

ov7a:
This issue needs a decision from the team responsible for that area. They have been informed. Response time may vary.

reinsch82:
The issue is in the backlog of the relevant team and is prioritized by them.

***

eskatos:
@imbananko, this will be delivered in 9.3.0-rc1

PR #34781, opened 2025-08-25
Improve/extend Kotlin base script models

jbartok:
### Contributor Checklist
- [ ] [Review Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md).
- [ ] Make sure that all commits are [signed off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff) to indicate that you agree to the terms of [Developer Certificate of Origin](https://developercertificate.org/).
- [ ] Make sure all contributed code can be distributed under the terms of the [Apache License 2.0](https://github.com/gradle/gradle/blob/master/LICENSE), e.g. the code was written by yourself or the original code is licensed under [a license compatible to Apache License 2.0](https://apache.org/legal/resolved.html).
- [ ] Check ["Allow edit from maintainers" option](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) in pull request so that additional changes can be pushed by Gradle team.
- [ ] Provide integration tests (under `<subproject>/src/integTest`) to verify changes from a user perspective.
- [ ] Provide unit tests (under `<subproject>/src/test`) to verify logic.
- [ ] Update User Guide, DSL Reference, and Javadoc for public-facing changes.
- [ ] Ensure that tests pass sanity check: `./gradlew sanityCheck`.
- [ ] Ensure that tests pass locally: `./gradlew <changed-subproject>:quickTest`.

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #35727, opened 2025-11-18
Fix Kotlin DSL script templates classpath in `GradleDslBaseScriptModel`

eskatos:
* Fixes https://github.com/gradle/gradle/issues/35545

This issue prevents IDE vendors to leverage our TAPI models to load the Kotlin DSL script templates.

The problem was introduced in https://github.com/gradle/gradle/commit/5a99a8a2f056f68d4093601f90d71b94352c6f01 by arbitrarily filtering the complete working classpath because it is too large, hindering the IDE performance. This commit was part of:
* https://github.com/gradle/gradle/pull/34781

The arbitrary filtering that was added to `GradleDslBaseScriptModelBuilder` missed the `org.gradle.api.HasImplicitReceiver` class that is referenced by fqName in the script templates.

This PR is a quick fix, not changing the way the classpath is computed. Long term this should be enhanced as described in this follow up issue:
* https://github.com/gradle/gradle/issues/35728

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/kotlin-dsl-tooling-builders/src/crossVersionTest/groovy/org/gradle/kotlin/dsl/tooling/builders/r93/GradleDslBaseScriptModelCrossVersionSpec.groovy
+++ b/platforms/core-configuration/kotlin-dsl-tooling-builders/src/crossVersionTest/groovy/org/gradle/kotlin/dsl/tooling/builders/r93/GradleDslBaseScriptModelCrossVersionSpec.groovy
@@ -14,13 +14,13 @@
  * limitations under the License.
  */
 
-package org.gradle.kotlin.dsl.tooling.builders.r92
+package org.gradle.kotlin.dsl.tooling.builders.r93
 
 import org.gradle.integtests.tooling.fixture.TargetGradleVersion
 import org.gradle.kotlin.dsl.tooling.builders.AbstractKotlinScriptModelCrossVersionTest
 import org.gradle.tooling.model.dsl.GradleDslBaseScriptModel
 
-@TargetGradleVersion(">=9.2 <9.3")
+@TargetGradleVersion(">=9.3")
 class GradleDslBaseScriptModelCrossVersionSpec extends AbstractKotlinScriptModelCrossVersionTest {
 
     def "GradleDslBaseScriptModel is obtained without configuring projects"() {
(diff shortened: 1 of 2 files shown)


platforms/core-configuration/kotlin-dsl-tooling-builders/src/crossVersionTest/groovy/org/gradle/kotlin/dsl/tooling/builders/r92/GradleDslBaseScriptModelCrossVersionSpec.groovy
```diff
@@ -20,7 +20,7 @@ import org.gradle.integtests.tooling.fixture.TargetGradleVersion
 import org.gradle.kotlin.dsl.tooling.builders.AbstractKotlinScriptModelCrossVersionTest
 import org.gradle.tooling.model.dsl.GradleDslBaseScriptModel
 
-@TargetGradleVersion(">=9.2")
+@TargetGradleVersion(">=9.2 <9.3")
```
asodja, line 23: ❓ Do we still need this test? I see there is one line difference between this and r93. Can we merge them or just delete this one?

eskatos: We should not merge them. With the added check it fails with Gradle 9.2.
If we remove the `.r92` one it means there would not be coverage for the 9.2 version of this model builder anymore.

asodja: I see, I wonder if we need to care about 9.2. Maybe we could ask IntelliJ to use this model only for 9.3+

eskatos: I'll leave that in your team's hands then.

PR #35730, opened 2025-11-18
Remove non-class-based testing and report entry APIs

tresat:
This work is incomplete, and will not be announced and delivered in Gradle 9.3.


tresat:
@bot-gradle test this

tresat:
@bot-gradle test and merge

PR #35737, opened 2025-11-18
Address performance regression caused by strict version constraint refactoring

jvandort:
Fixes performance regression introduced by https://github.com/gradle/gradle/pull/34893 by tracking the current value of the ancestors strict versions for every node, as it changes, instead of recomputing it each time a node is visited.  

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/alignment/ForcingUsingStrictlyPlatformAlignmentTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/alignment/ForcingUsingStrictlyPlatformAlignmentTest.groovy
@@ -164,7 +164,7 @@ class ForcingUsingStrictlyPlatformAlignmentTest extends AbstractAlignmentSpec {
         then:
         failure.assertHasCause """Cannot find a version of 'org:databind' that satisfies the version constraints:
    Dependency path: 'root project :' (conf) --> 'org:databind:{strictly 2.7.9}'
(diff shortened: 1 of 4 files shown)


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test rfr

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/strict/StrictVersionConstraints.java
```diff
@@ -105,7 +106,19 @@ public StrictVersionConstraints intersect(StrictVersionConstraints other) {
         if (other == EMPTY) {
             return EMPTY;
         }
-        return of(ImmutableSet.copyOf(Sets.intersection(modules, other.modules)));
+
+        Set<ModuleIdentifier> smaller = (modules.size() < other.modules.size()) ? modules : other.modules;
```
ljacomet, line 110: 🤔 This might be good to put in its own method, with a name that indicates why it exists.

ljacomet (approved):
One minor comment, see below. Otherwise this is good to go!

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/strict/StrictVersionConstraints.java
jvandort: IMO the comment below describes the reasoning here 
`// Iterating over the smaller set to minimize the number of contains() checks`

PR #35750, opened 2025-11-19
Correct UnknownClass in test logging for file-based tests

tresat:
Updates test logging to stop incorrectly reporting UnknownClass as the class name for non-class-based tests, improving clarity in console output.

It now uses an empty string instead of "UnknownClass" for file-based tests.

Instead of ouput like this:

```text
UnknownClass.tests.xml : foo STANDARD_OUT
    12:36:01.497 [Test worker] INFO org.gradle.rbt.engine.ResourceBasedTestEngine -- Executing test: Test [file=tests.xml, name=foo]

UnknownClass.tests.xml : bar STANDARD_OUT
    12:36:01.499 [Test worker] INFO org.gradle.rbt.engine.ResourceBasedTestEngine -- Executing test: Test [file=tests.xml, name=bar]
```

We now get output like this:

```text
tests.xml : foo STANDARD_OUT
    12:36:01.497 [Test worker] INFO org.gradle.rbt.engine.ResourceBasedTestEngine -- Executing test: Test [file=tests.xml, name=foo]

tests.xml : bar STANDARD_OUT
    12:36:01.499 [Test worker] INFO org.gradle.rbt.engine.ResourceBasedTestEngine -- Executing test: Test [file=tests.xml, name=bar]
```

### DIFF
--- a/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
+++ b/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
@@ -306,7 +306,7 @@ public class JUnitPlatformTestExecutionListener implements TestExecutionListener
         Optional<MethodSource> methodSource = getMethodSource(node);
         if (methodSource.isPresent()) {
             TestDescriptorInternal parentDescriptor = findTestParentDescriptor(node);
-            String className = createUnknownClassName(node, parentDescriptor);
+            String className = determineClassName(node, parentDescriptor);
             return new DefaultParameterizedTestDescriptor(idGenerator.generateId(), node.getLegacyReportingName(), className, displayName, candidateId);
         } else {
             return new DefaultNestedTestSuiteDescriptor(idGenerator.generateId(), node.getLegacyReportingName(), displayName, candidateId);
@@ -328,20 +328,20 @@ public class JUnitPlatformTestExecutionListener implements TestExecutionListener
 
     private TestDescriptorInternal createTestDescriptor(TestIdentifier test, String name, String displayName) {
         TestDescriptorInternal parentDescriptor = findTestParentDescriptor(test);
-        String className = createUnknownClassName(test, parentDescriptor);
-        String classDisplayName = createUnknownClassDisplayName(test, parentDescriptor);
+        String className = determineClassName(test, parentDescriptor);
+        String classDisplayName = determineClassDisplayName(test, parentDescriptor);
         return new DefaultTestDescriptor(idGenerator.generateId(), className, name, classDisplayName, displayName);
     }
 
-    private String createUnknownClassName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor) {
-        return createUnknownName(node, parentDescriptor, TestDescriptorInternal::getName);
+    private String determineClassName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor) {
+        return determineName(node, parentDescriptor, TestDescriptorInternal::getName);
     }
 
-    private String createUnknownClassDisplayName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor) {
-        return createUnknownName(node, parentDescriptor, TestDescriptorInternal::getClassDisplayName);
+    private String determineClassDisplayName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor) {
+        return determineName(node, parentDescriptor, TestDescriptorInternal::getClassDisplayName);
     }
 
-    private String createUnknownName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor, Function<TestDescriptorInternal, @Nullable String> nameGetter) {
+    private String determineName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor, Function<TestDescriptorInternal, @Nullable String> nameGetter) {
         TestSource source = node.getSource().orElse(null);
         if (source instanceof ClassSource || source instanceof MethodSource) {
             if (parentDescriptor == null) {


tresat:
@bot-gradle test this

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -328,11 +327,32 @@ private TestDescriptorInternal createSyntheticTestDescriptorForContainer(TestIde
...
+        return createUnknownName(node, parentDescriptor, TestDescriptorInternal::getName);
+    }
+
+    private String createUnknownClassDisplayName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor) {
+        return createUnknownName(node, parentDescriptor, TestDescriptorInternal::getClassDisplayName);
+    }
+
+    private String createUnknownName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor, Function<TestDescriptorInternal, @Nullable String> nameGetter) {
+        if (node.getSource().orElse(null) instanceof FileSource) {
+            return JUnitPlatformSupport.UNKNOWN_FILE;
```
octylFractal: ❓ Why can't we use the name from the file here instead?

tresat: I think we should expect that to be already present as the `displayName` from the `TestIdentifier` for a `FileSource` type test.  How would you name a test in the file without the file name?  This is an assumption though, as we can't know for sure how every `TestEngine` will structure display names.

If we did this, currently, we would get output like:

```text
tests.xml.tests.xml : foo STANDARD_OUT
    12:36:01.497 [Test worker] INFO org.gradle.rbt.engine.ResourceBasedTestEngine -- Executing test: Test [file=tests.xml, name=foo]

```

which I don't think makes sense.

I don't think there's a need to use these class-based fields at all with non-class-based testing.

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
octylFractal: Ah, I see. Makes sense.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformSupport.java
```diff
@@ -13,8 +13,15 @@
...
-public abstract class JUnitSupport {
-    public static final String UNKNOWN_CLASS = "UnknownClass";
+/**
+ * This interfaces defines static values used when creating JUnit Platform test descriptors.
+ * <p>
+ * Changes here affect how tests are displayed in the console.
+ */
+public interface JUnitPlatformSupport {
+    String UNKNOWN_CLASS = "UnknownClass";
+    String UNKNOWN_FILE = ""; // Empty string for file-based tests
```
octylFractal: :x: Convert to Javadoc.
```suggestion
    /**
     * An empty string for file-based tests.
     */
    String UNKNOWN_FILE = "";
```

octylFractal: 💭 Should this be called something else? We know what the file is, we just don't want to have a value here. Perhaps `EMPTY_CLASS_VALUE`?

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -328,11 +327,32 @@ private TestDescriptorInternal createSyntheticTestDescriptorForContainer(TestIde
...
+        } else {
+            if (parentDescriptor == null) {
+                return JUnitPlatformSupport.UNKNOWN_CLASS;
```
octylFractal, line 348: 💭 Why not use `UNKNOWN` here too? I don't quite understand what makes it so that we know it's a class here, but not in the other case.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/logging/AbstractTestLogger.java
octylFractal, line 71: ❓ Should we adjust this comment if this branch is being used for non-class-based testing?

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformSupport.java
tresat: I just wanted to keep the naming pattern of `UNKNOWN_CLASS`, but I think you're right, that's not what this represents.

I'm going to go with `NON_CLASS`.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
tresat: We're in the `else`-branch for (node.getSource().orElse(null) instanceof FileSource).  

I guess we don't know it's a class, but this is what the prior code used.  I reworked this to be a little more clear/correct.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/logging/AbstractTestLogger.java
tresat: Done.

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -328,11 +328,33 @@ private TestDescriptorInternal createSyntheticTestDescriptorForContainer(TestIde
...
 
+    private String createUnknownClassName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor) {
+        return createUnknownName(node, parentDescriptor, TestDescriptorInternal::getName);
+    }
+
+    private String createUnknownClassDisplayName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor) {
+        return createUnknownName(node, parentDescriptor, TestDescriptorInternal::getClassDisplayName);
+    }
+
+    private String createUnknownName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor, Function<TestDescriptorInternal, @Nullable String> nameGetter) {
```
octylFractal: :x: I think this has a weird name, now that I read it more. Shouldn't these methods be called something like `determineClassName`, `determineClassDisplayName`, and `determineName`? They're called unconditionally, not only when we don't know the name.

tresat: You know, I had that at first, then I switched away from it.  I don't have strong feelings here, I'll switch back.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -328,11 +328,33 @@ private TestDescriptorInternal createSyntheticTestDescriptorForContainer(TestIde
...
 
+    private String determineUnknownClassName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor) {
+        return determineUnknownName(node, parentDescriptor, TestDescriptorInternal::getName);
+    }
+
+    private String determineUnknownClassDisplayName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor) {
+        return determineUnknownName(node, parentDescriptor, TestDescriptorInternal::getClassDisplayName);
+    }
+
+    private String determineUnknownName(TestIdentifier node, @Nullable TestDescriptorInternal parentDescriptor, Function<TestDescriptorInternal, @Nullable String> nameGetter) {
```
octylFractal: :x: I still don't think it should have `Unknown` in the name. It returns a valid, known, value with class-based tests.

tresat: Yeah, I suppose that makes sense.  How does it look now?

PR #35753, opened 2025-11-20
Add parallel non-class-based testing test class

tresat:


tresat:
@bot-gradle test and merge

PR #35755, opened 2025-11-20
Create ToolingModelContext and propagate isFetch to BuildToolingModelController

reinsch82:
The goal is to avoid use of the `isResilientModelBuilding` property, so we can get rid of it.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultBuildToolingModelControllerFactory.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultBuildToolingModelControllerFactory.kt
@@ -30,9 +30,9 @@ internal
 class DefaultBuildToolingModelControllerFactory(
     private val modelParameters: BuildModelParameters
 ) : BuildToolingModelControllerFactory {
-    override fun createController(owner: BuildState, lifecycleController: BuildLifecycleController): BuildToolingModelController {
+    override fun createController(owner: BuildState, lifecycleController: BuildLifecycleController, inResilientContext: Boolean): BuildToolingModelController {
         val modelBuilderLookup = lifecycleController.gradle.services.get(ToolingModelBuilderLookup::class.java)
-        val toolingModelController = if (modelParameters.isResilientModelBuilding) {
+        val toolingModelController = if (inResilientContext) {
             ResilientBuildToolingModelController(owner, lifecycleController, modelBuilderLookup)
         } else {
             DefaultBuildToolingModelController(owner, lifecycleController, modelBuilderLookup)
--- a/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.groovy
+++ b/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.groovy
@@ -73,8 +73,8 @@ class ResilientKotlinDslScriptsModelBuilderCrossVersionSpec extends ToolingApiSp
         when:
         def resilientModels = succeeds {
             action(KotlinModelAction.resilientModel(ROOT_PROJECT_FIRST))
-                    .withArguments("-Dorg.gradle.internal.resilient-model-building=true")
-                    .run()
+                .withArguments("-Dorg.gradle.internal.resilient-model-building=true")
+                .run()
         }
 
         then:
@@ -110,14 +110,14 @@ class ResilientKotlinDslScriptsModelBuilderCrossVersionSpec extends ToolingApiSp
         when:
         def model = succeeds {
             action(KotlinModelAction.resilientModel(ROOT_PROJECT_FIRST))
-                    .withArguments("-Dorg.gradle.internal.resilient-model-building=true")
-                    .run()
+                .withArguments("-Dorg.gradle.internal.resilient-model-building=true")
+                .run()
         }
 
         then:
         assertHasScriptModelForFiles(model/*, "settings.gradle.kts"*/)
         assertHasErrorsInScriptModels(model,
-                Pair.of(".", ".*Settings file.*settings\\.gradle\\.kts.*Script compilation error.*"))
+            Pair.of(".", ".*Settings file.*settings\\.gradle\\.kts.*Script compilation error.*"))
         // assertHasJarsInScriptModelClasspath(model, "settings.gradle.kts", "gradle-kotlin-dsl-plugins")
     }
 
@@ -140,14 +140,14 @@ class ResilientKotlinDslScriptsModelBuilderCrossVersionSpec extends ToolingApiSp
         when:
         def model = succeeds {
             action(KotlinModelAction.resilientModel(ROOT_PROJECT_FIRST))
-                    .withArguments("-Dorg.gradle.internal.resilient-model-building=true")
-                    .run()
+                .withArguments("-Dorg.gradle.internal.resilient-model-building=true")
+                .run()
         }
 
         then:
         assertHasScriptModelForFiles(model, "settings.gradle.kts", "build.gradle.kts")
         assertHasErrorsInScriptModels(model,
-                Pair.of(".", ".*Build file.*build\\.gradle\\.kts.*Script compilation error.*"))
+            Pair.of(".", ".*Build file.*build\\.gradle\\.kts.*Script compilation error.*"))
         assertHasJarsInScriptModelClasspath(model, "build.gradle.kts", "gradle-kotlin-dsl-plugins")
     }
 
@@ -163,14 +163,14 @@ class ResilientKotlinDslScriptsModelBuilderCrossVersionSpec extends ToolingApiSp
         when:
         def model = succeeds {
             action(KotlinModelAction.resilientModel(ROOT_PROJECT_FIRST))
-                    .withArguments("-Dorg.gradle.internal.resilient-model-building=true")
-                    .run()
+                .withArguments("-Dorg.gradle.internal.resilient-model-building=true")
+                .run()
         }
 
         then:
         assertHasScriptModelForFiles(model, "settings.gradle.kts", "build.gradle.kts")
         assertHasErrorsInScriptModels(model,
-                Pair.of(".", ".*Build file.*build\\.gradle\\.kts.*Script compilation error.*"))
+            Pair.of(".", ".*Build file.*build\\.gradle\\.kts.*Script compilation error.*"))
         assertHasJarsInScriptModelClasspath(model, "build.gradle.kts", "gradle-api")
     }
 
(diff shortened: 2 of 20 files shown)


platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultBuildToolingModelControllerFactory.kt
```diff
@@ -30,11 +30,12 @@ internal
 class DefaultBuildToolingModelControllerFactory(
     private val modelParameters: BuildModelParameters
 ) : BuildToolingModelControllerFactory {
-    override fun createController(owner: BuildState, lifecycleController: BuildLifecycleController): BuildToolingModelController {
+    override fun createController(owner: BuildState, lifecycleController: BuildLifecycleController, isFetch: Boolean): BuildToolingModelController {
         val modelBuilderLookup = lifecycleController.gradle.services.get(ToolingModelBuilderLookup::class.java)
-        val toolingModelController = if (modelParameters.isResilientModelBuilding) {
+        val toolingModelController = if(isFetch) {
```
alllex: 💅  formatting

subprojects/core/src/main/java/org/gradle/internal/ToolingModelContext.java
```diff
@@ -14,23 +14,28 @@
...
-    public void I_ask_it_to_say_hi() {
-        System.out.println("When");
+    public boolean hasParameter() {
+        return hasParameter;
     }
 
-    @Then("^it should answer with Howdy World")
-    public void it_should_answer_with() {
-        System.out.println("Then");
+    public boolean isFetch() {
```
alllex: 🤔  It's very unclear what this means without context. I have the same feeling about all `isFetch` parameters we now pass around. Can we use a more descriptive name that describes the purpose or an intended effect?

The `fetch` terminology is fine for the API layer, where we want to be concise. But internally, we can switch it for something more self-explanatory.

cc @asodja, since the `fetch` terminology is now somewhat cross-cutting for TAPI implementation

subprojects/core/src/main/java/org/gradle/internal/ToolingModelContext.java
```diff
@@ -14,23 +14,28 @@
...
-import io.cucumber.java.en.Given;
-import io.cucumber.java.en.Then;
-import io.cucumber.java.en.When;
+package org.gradle.internal;
 
-public class HelloStepdefs {
-    @Given("^I have a hello app with Howdy and /four")
-    public void I_have_a_hello_app_with() {
-        System.out.println("Given");
+public class ToolingModelContext {
```
alllex: 💭  Would it be fair to say this is about the model _request_? Maybe we can add this bit to the class name as well.

subprojects/core/src/main/java/org/gradle/internal/ToolingModelContext.java
```diff
@@ -14,23 +14,28 @@
  * limitations under the License.
  */
 
-import io.cucumber.java.en.Given;
-import io.cucumber.java.en.Then;
-import io.cucumber.java.en.When;
+package org.gradle.internal;
```
alllex: ❌  This is not the right package for this class

alllex (dismissed):
Is there any behavior change or is it a pure refactoring?
If it's the latter, it would be helpful to include some context in the PR description.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultBuildToolingModelControllerFactory.kt
reinsch82: builds used to fail on wrong kotlin formatting, right?
Can't we have that again?

subprojects/core/src/main/java/org/gradle/internal/ToolingModelContext.java
asodja: ➕  Good catch

subprojects/core/src/main/java/org/gradle/internal/ToolingModelContext.java
asodja: 💅  Good point, let's use "isResilientMode", since we use "resilient mode" also in docs and in other class names

subprojects/core/src/main/java/org/gradle/internal/buildtree/DefaultBuildTreeModelCreator.java
```diff
@@ -69,19 +70,19 @@ public <T> T fromBuildModel(BuildTreeModelAction<? extends T> action) {
...
         @Override
         public GradleInternal getConfiguredModel() {
-            return defaultTarget.withToolingModels(BuildToolingModelController::getConfiguredModel);
+            return defaultTarget.withToolingModels(false, BuildToolingModelController::getConfiguredModel);
         }
 
         @Override
         @Nullable
-        public Object getModel(BuildTreeModelTarget target, String modelName, @Nullable Object parameter) throws UnknownModelException {
+        public Object getModel(BuildTreeModelTarget target, String modelName, @Nullable Object parameter, boolean isFetch) throws UnknownModelException {
```
asodja: 💅 We could already wrap these `String modelName, @Nullable Object parameter, boolean isFetch` in a context object, and context object and then we could pass context to `getModelForScope()` too

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.groovy
```diff
@@ -64,7 +64,9 @@ class ResilientKotlinDslScriptsModelBuilderCrossVersionSpec extends ToolingApiSp
 
         when:
         def original = succeeds {
-            action(KotlinModelAction.originalModel(ROOT_PROJECT_FIRST)).run()
+            action(KotlinModelAction.originalModel(ROOT_PROJECT_FIRST))
+                .withArguments("-Dorg.gradle.internal.resilient-model-building=true")
```
asodja: ❌ This is normal request that should not have:
```
.withArguments("-Dorg.gradle.internal.resilient-model-building=true")
```

subprojects/core/src/main/java/org/gradle/internal/build/BuildLifecycleController.java
```diff
@@ -108,7 +108,7 @@ public interface BuildLifecycleController {
     /**
      * Runs an action against the tooling model creators of this build. May configure the build as required.
      */
-    <T> T withToolingModels(Function<? super BuildToolingModelController, T> action);
+    <T> T withToolingModels(boolean isFetch, Function<? super BuildToolingModelController, T> action);
```
asodja: 💅   Let's not forget to rename this to `isResilientMode` too

subprojects/core/src/main/java/org/gradle/internal/build/BuildState.java
```diff
@@ -120,7 +120,7 @@ default BuildIdentifier getBuildIdentifier() {
     /**
      * Runs an action against the tooling model creators of this build. May configure the build as required.
      */
-    <T> T withToolingModels(Function<? super BuildToolingModelController, T> action);
+    <T> T withToolingModels(boolean isFetch, Function<? super BuildToolingModelController, T> action);
```
asodja: 💅   Let's not forget to rename this to `isResilientMode` too

subprojects/core/src/main/java/org/gradle/internal/build/BuildToolingModelControllerFactory.java
```diff
@@ -21,5 +21,5 @@
 
 @ServiceScope(Scope.BuildTree.class)
 public interface BuildToolingModelControllerFactory {
-    BuildToolingModelController createController(BuildState owner, BuildLifecycleController controller);
+    BuildToolingModelController createController(BuildState owner, BuildLifecycleController controller, boolean isFetch);
```
asodja: 💅   Let's not forget to rename this to `isResilientMode` too

subprojects/core/src/main/java/org/gradle/internal/buildtree/BuildTreeModelController.java
```diff
@@ -38,7 +38,7 @@ public interface BuildTreeModelController {
      * @throws UnknownModelException when the model builder cannot be found
      */
     @Nullable
-    Object getModel(BuildTreeModelTarget target, String modelName, @Nullable Object parameter) throws UnknownModelException;
+    Object getModel(BuildTreeModelTarget target, String modelName, @Nullable Object parameter, boolean isFetch) throws UnknownModelException;
```
asodja: 💅 As mentioned before, I think we could already pass context object here, would make this signature nicer. And context object would have `getParameter` and `hasParameter` method

subprojects/core/src/main/java/org/gradle/internal/ToolingModelContext.java
asodja: I think we should move it to `org.gradle.internal.buildtree` if we will use it in [BuildTreeModelController.java‎](https://github.com/gradle/gradle/pull/35755/files#diff-a5c8a711b6aad7915b24b7ed63e508ac0eefeb27ce3ad14d396c1db5c140c795)
or in `org.gradle.internal.build` otherwise.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultBuildToolingModelControllerFactory.kt
alllex: It's worth asking in Slack about the state of this.

subprojects/core/src/main/java/org/gradle/internal/ToolingModelContext.java
alllex: I've seen this happen a couple of times recently. I created an ArchUnit test for this. PTAL!
- https://github.com/gradle/gradle/pull/35809

subprojects/core/src/main/java/org/gradle/internal/build/DefaultBuildToolingModelController.java
```diff
@@ -48,34 +49,34 @@ public GradleInternal getConfiguredModel() {
     }
 
     @Override
-    public ToolingModelScope locateBuilderForTarget(String modelName, boolean param) {
+    public ToolingModelScope locateBuilderForTarget(ToolingModelRequestContext toolingModelContext) {
         // Look for a build scoped builder
-        ToolingModelBuilderLookup.Builder builder = buildScopeLookup.maybeLocateForBuildScope(modelName, param, buildState);
+        ToolingModelBuilderLookup.Builder builder = buildScopeLookup.maybeLocateForBuildScope(toolingModelContext.getModelName(), toolingModelContext.getParameter().isPresent(), buildState);
```
asodja, line 54: 🎉 `toolingModelContext.getParameter().isPresent()` is actually very expressive and clear, I like it

subprojects/core/src/main/java/org/gradle/tooling/provider/model/internal/DefaultToolingModelBuilderRegistry.java
```diff
@@ -126,6 +126,8 @@ public Builder maybeLocateForBuildScope(String modelName, boolean parameter, Bui
 
     private Builder locateForClientOperation(String modelName, ProjectInternal project, boolean parameter) throws UnknownModelException {
         Registration registration = get(modelName);
+//        if (isFetch && registration.getBuilder() instance ResilientModelBuiolder) {}
```
asodja: 💅 Do we need this in production code?

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
```diff
@@ -59,18 +61,17 @@ private static void rethrowExceptionIfNotResilientModel(ProjectState target, Str
...
-    protected ToolingModelScope doLocate(ProjectState target, String modelName, boolean param) {
-        return new ResilientProjectToolingScope(target, modelName, param);
+    protected ToolingModelScope doLocate(ProjectState target, ToolingModelRequestContext toolingModelContext) {
+        return new ResilientProjectToolingScope(target, toolingModelContext);
     }
 
-    private static class ResilientProjectToolingScope extends ProjectToolingScope {
-        public ResilientProjectToolingScope(ProjectState target, String modelName, boolean parameter) {
-            super(target, modelName, parameter);
+    private class ResilientProjectToolingScope extends ProjectToolingScope {
```
asodja: ❓ Is there a reason we made this class non-static? Generally static classes are prefered since they have no reference to a parent (and all it's properties), so it's harder to make some memory or context leak if we pass them around.

reinsch82: we don't, this is a leftover from previous changes.

asodja (approved):
LGTM

Issue #19788, opened 2022-02-01
ResolvedVariantResult getResolvedVariant() is sometimes null


cmwarrington:
### Expected Behavior
ResolvedVariantResult getResolvedVariant() is never null.

### Current Behavior
The project in https://issuetracker.google.com/214259374#comment1 hits a corner case where ResolvedVariantResult getResolvedVariant() returns null.

### Context
Android Studio switched to requesting a rewritten dependency model from Android Gradle Plugin from version 7.2.0-alpha05, where the model builder walks the dependency graph and now depends on ResolvedVariantResult.getResolvedVariant, and this broke sync for projects that hit this issue in Gradle.

### Steps to Reproduce 
Open the project described in https://issuetracker.google.com/214259374#comment1 (with Android Studio C canary 5).

### Your Environment
See linked issue

eskatos:
Sorry for the late reply.
The issue is in the backlog of the relevant team and is prioritized by them.

big-guy:
@cmwarrington could you try with the latest 8.x or 9?

PR #35758, opened 2025-11-20
Ensure all edges without failures have selected variants

jvandort:
We add additional invariants to verify this. Previously, we had a few hacks that would work around a few bugs that would cause resolved edges without target nodes. These hacks included picking some random other selected node from the same component, or if that fails, returning null in a public API that should never return null.

Along the way, we fixed two bugs that would have triggered this invariant:
1. Allow nodes in the same component to conflict with each other. Previously this would break the graph when we tried to resolve the conflict. Now, we properly track replaced nodes in the same component allowing us to properly resolve these conflicts when selecting variants.
2. Attach unattached edges when performing initial selection. This can occur when a selection is deferred once due to a constraint edge being added in the future, but after selection is deferred, another constraint comes in selecting a component. Previously, we only attached unattached edges when changing a selection of a module, but now we always do this

Fixes https://github.com/gradle/gradle/issues/19788

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
Superseded by https://github.com/gradle/gradle/pull/35801 and https://github.com/gradle/gradle/pull/36292

PR #35759, opened 2025-11-20
Add filtering for non-class-based tests

tresat:


tresat:
@bot-gradle test this

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcher.java
```diff
@@ -36,28 +38,69 @@
...
-    private final List<TestPattern> buildScriptIncludePatterns;
-    private final List<TestPattern> buildScriptExcludePatterns;
-    private final List<TestPattern> commandLineIncludePatterns;
+    private final List<ClassTestPattern> buildScriptIncludePatterns;
+    private final List<ClassTestPattern> buildScriptExcludePatterns;
+    private final List<ClassTestPattern> commandLineIncludePatterns;
+
+    private final List<FileTestPattern> fileBuildScriptIncludePatterns;
+    private final List<FileTestPattern> fileBuildScriptExcludePatterns;
+    private final List<FileTestPattern> fileCommandLineIncludePatterns;
```
tresat: I went with a simple duplication approach here.  There is probably a way to consolidate some of the code here, but I'm not too concerned with the 3 extra fields - it helps ensure I haven't changed the other behavior.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcher.java
```diff
@@ -137,13 +178,42 @@ private boolean matchesClassAndMethod(List<TestPattern> patterns, String classNa
...
+                } else {
+                    if (pattern.length() > 0) {
+                        pattern.append(".*"); //replace wildcard '*' with '.*'
+                    }
+                    pattern.append(Pattern.quote(s)); //quote everything else
+                }
+            }
+            return Pattern.compile(pattern.toString());
+        }
+    }
```
tresat, line 221: I _think_ this is the only bit of tweaking we need to turn the filter pattern into a useful regex.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/DirectoryBasedTestDefinition.java
```diff
@@ -52,7 +52,10 @@ public String getDisplayName() {
 
     @Override
     public boolean matches(TestSelectionMatcher matcher) {
-        // TODO: Implement filtering
+        // The only place this method is called is from PatternMatchTestDefinitionProcessor,
+        // where we want to let all candidate test definition dirs that have already
+        // been filtered in the Test task itself through to the workers.
+        // So we'll just let everything through here.
```
tresat, line 58: This is Gate (B) filtering from the Spec, which isn't necessary or beneficial.

It probably means we could move `matches()` out of `TestDefinition` entirely, but I kept it simple for now.

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/FilePathFilter.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.File;
+import java.nio.file.Path;
+import java.util.Optional;
+
+/**
+ * A JUnit Platform {@link PostDiscoveryFilter} filter that includes or excludes
+ * file or directory based tests based on their relative paths to the root directory
+ * of the project that contains them.
+ */
+@NullMarked
```
big-guy: I think these don't need to be null marked if the package is null marked.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestDefinitionProcessor.java
```diff
@@ -212,10 +211,20 @@ private LauncherDiscoveryRequest createLauncherDiscoveryRequest() {
...
+                ClassMethodNameFilter classFilter = new ClassMethodNameFilter(matcher);
+                FilePathFilter fileFilter = new FilePathFilter(matcher, spec.getBaseDefinitionsDir());
+
+                DelegatingByTypeFilter delegatingFilter = new DelegatingByTypeFilter();
+                delegatingFilter.addDelegate(ClassSource.class, classFilter);
+                delegatingFilter.addDelegate(MethodSource.class, classFilter);
+                delegatingFilter.addDelegate(FileSource.class, fileFilter);
+                delegatingFilter.addDelegate(DirectorySource.class, fileFilter);
+
+                requestBuilder.filters(delegatingFilter);
```
big-guy, line 231: What's the significance of using `DelegatingByTypeFilter` vs just passing both classFilter and fileFilter here? We have other filters we create after this (engine filters).

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/FilePathFilter.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public FilterResult apply(TestDescriptor descriptor) {
+        return FilterResult.includedIf(shouldRun(descriptor), () -> "File match", () -> "File mismatch");
+    }
+
+    private boolean shouldRun(TestDescriptor descriptor) {
+        Optional<TestSource> source = descriptor.getSource();
+        if (!source.isPresent()) {
+            throw new IllegalArgumentException("No test source found for " + descriptor);
```
big-guy: This seems like an overly strict check.  We can just return true here since we don't know how to filter something without a source?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -297,7 +297,211 @@ class NonClassBasedTestingIntegrationTest extends AbstractNonClassBasedTestingIn
...
+
+        writeTestDefinitions()
+
+        when:
+        succeeds("test", "--info")
+
+        then:
+        outputDoesNotContain("INFO: Executing resource-based test: Test[file=SomeTestSpec.rbt, name=foo]")
+        outputDoesNotContain("INFO: Executing resource-based test: Test[file=SomeTestSpec.rbt, name=bar]")
+        outputContains("INFO: Executing resource-based test: Test[file=SomeOtherTestSpec.rbt, name=other]")
```
big-guy, line 379: Maybe not as part of this, but we should not be using console output to assert that certain tests have executed.  We should be checking the reports.  `outputDoesNotContain` and `outputContains` are not reliable because `outputDoesNotContain` will allow typos to pass through and `outputContains` will allow extra tests to execute.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -297,7 +297,211 @@ class NonClassBasedTestingIntegrationTest extends AbstractNonClassBasedTestingIn
...
+                        filter {
+                            includeTestsMatching "*subdir1/SomeTestSpec*"
+                        }
+                    }
+                }
+            }
+        """
+
+        writeTestDefinitions()
+        writeTestDefinitions()
```
big-guy, line 498: typo?

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
```diff
@@ -167,6 +172,8 @@
 @NullMarked
 @CacheableTask
 public abstract class Test extends AbstractTestTask implements JavaForkOptions, PatternFilterable {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractTestTask.class);
```
big-guy: Wrong logger?

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/internal/tasks/testing/detection/DefaultTestExecuter.java
```diff
@@ -68,7 +69,7 @@ public class DefaultTestExecuter implements TestExecuter<JvmTestExecutionSpec> {
     public DefaultTestExecuter(
         WorkerProcessFactory workerFactory, ActorFactory actorFactory, ModuleRegistry moduleRegistry,
         WorkerLeaseService workerLeaseService, int maxWorkerCount,
-        Clock clock, DefaultTestFilter testFilter
+        Clock clock, DefaultTestFilter testFilter, ProjectLayout projectLayout
```
big-guy: This is unused?

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestDefinitionProcessor.java
tresat: Each `PostDiscoveryFilter`'s `apply(TestDescriptor)` method returns a `FilterResult`.  That result must say included or excluded.

I think it would be odd for the class-based filter to be run on a non-class-based test and have to report "included" (and vice-versa), when really that filter is n/a and can't render a meaningful opinion.

Using the delegating filter avoids the improper type filter even running.  I'll note this on the javadoc.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/FilePathFilter.java
tresat: We won't even get here, since the delegating filter won't find anything by type, so I'd rather leave this as an error.  I'll consolidate this.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
tresat: Yes.  Updating all these tests to use the assertions was just waiting on finishing some changes to the reporting.  

We'll switch everything in this package to use the asssertions and add checking the XML to the assertions soon.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
tresat: Yes.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
tresat: Yes.

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/internal/tasks/testing/detection/DefaultTestExecuter.java
tresat: It is not.  Removing.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -297,7 +312,407 @@ class NonClassBasedTestingIntegrationTest extends AbstractNonClassBasedTestingIn
...
+        when:
+        succeeds("test", "--info")
+
+        then:
+        outputContains("INFO: Executing resource-based test: Test[file=SampleTest.rbt, name=foo]")
+        resultsFor().assertTestPathsNotExecuted(":definitions.SampleTest:foo()")
+    }
+
+    @Ignore
+    def "can filter resource-based test with #entryPoint"() {
```
big-guy: This needs to be in crossVersionTest, but if we need to ignore it, I'd just remove it for now

tresat:
@bot-gradle test and merge

tresat:
@bot-gradle test and merge

Issue #2688, opened 2017-08-09
Gradle logging hygiene


rmorrise:
Implement log hygiene in Gradle's default configuration, as described in:
https://discuss.gradle.org/t/gradle-daemon-produces-a-lot-of-logs/9905

### Expected Behavior

Gradle should be configured out-of-the-box with a rolling file appender or similar mechanism so that the daemon logs are periodically cleaned up.

### Current Behavior

Gradle produces daemon logs in a hidden user directory. These logs accumulate over time and are never cleaned up.

### Context

My Grails build was leaking daemon processes, contributing to low disk space on C:
(The daemon process leak is beyond the scope of this feature request.)

### Steps to Reproduce (for bugs)

Not a bug.

### Your Environment

 * Build scan URL: 

Win7 x64, Grails 3.3.0, Gradle 3.4.1, Groovy 2.4.7, JVM 1.8.0_66

staffanf:
I just found 20 Gb worth of deamon logs in ~/.gradle. Some of these were 4Gb full of System.out output some debug task I ran some month ago.
Would be nice with some kind of log rotation at least.

agilob:
I just removed 89Gb of text files in `.gradle/daemon/4.7` -.-'

lostcart:
I just deleted 160GB after Android studio complained I didn't have enough disk space - 1 day later and log files are at 2.4GB - this is insane . (I also have over 9GB of memory used from AS, but that's another issue :D)

fbe:
+1

petarmarijanovicfive:
+1

I deleted 100+GB. 

This is a big issue. Gradle shouldn't log anything to a separate file unless it is specifically configured to do so.

coderkun:
It looks like the log level in these files is `DEBUG`. If it is possible to configure the log level the Gradle daemon is using, it will reduce the amount of logs greatly.

chaosky:
> It looks like the log level in these files is `DEBUG`. If it is possible to configure the log level the Gradle daemon is using, it will reduce the amount of logs greatly.

It looks like it's hardcoded in [DaemonMain.java#L192](https://github.com/gradle/gradle/blob/v5.6.4/subprojects/launcher/src/main/java/org/gradle/launcher/daemon/bootstrap/DaemonMain.java#L192), probably providing a `System.property` such as `-Dorg.gradle.daemon.logging.level=(quiet,warn,lifecycle,info,debug)` to control it.

@fbe @rmorrise

[...]

PR #35765, opened 2025-11-21
Delete daemon logs older than 2 weeks on shutdown

reinsch82:
This PR adds deletion of files on Gradle daemon shut down.

Also some small refactorings in adjacent code.

Fixes 
 - #2688

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/base-services/src/main/java/org/gradle/util/internal/DefaultGradleVersion.java
+++ b/platforms/core-runtime/base-services/src/main/java/org/gradle/util/internal/DefaultGradleVersion.java
@@ -35,7 +35,7 @@ import static java.lang.String.format;
 import static org.gradle.internal.IoActions.uncheckedClose;
 
 public final class DefaultGradleVersion extends GradleVersion {
-    private static final Pattern VERSION_PATTERN = Pattern.compile("((\\d+)(\\.\\d+)+)(-(\\p{Alpha}+)-(\\w+))?(-(SNAPSHOT|\\d{14}([-+]\\d{4})?))?");
+    public static final Pattern VERSION_PATTERN = Pattern.compile("((\\d+)(\\.\\d+)+)(-(\\p{Alpha}+)-(\\w+))?(-(SNAPSHOT|\\d{14}([-+]\\d{4})?))?");
     private static final int STAGE_MILESTONE = 0;
     private static final int STAGE_UNKNOWN = 1;
     private static final int STAGE_PREVIEW = 2;
--- a/platforms/core-runtime/daemon-server/src/main/java/org/gradle/launcher/daemon/bootstrap/DaemonMain.java
+++ b/platforms/core-runtime/daemon-server/src/main/java/org/gradle/launcher/daemon/bootstrap/DaemonMain.java
@@ -40,6 +40,7 @@ import org.gradle.launcher.daemon.configuration.DaemonServerConfiguration;
 import org.gradle.launcher.daemon.configuration.DefaultDaemonServerConfiguration;
 import org.gradle.launcher.daemon.context.DaemonContext;
 import org.gradle.launcher.daemon.logging.DaemonMessages;
+import org.gradle.launcher.daemon.registry.DaemonDir;
 import org.gradle.launcher.daemon.server.Daemon;
 import org.gradle.launcher.daemon.server.DaemonLogFile;
 import org.gradle.launcher.daemon.server.DaemonProcessState;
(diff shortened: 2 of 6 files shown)


platforms/core-runtime/daemon-server/src/main/java/org/gradle/launcher/daemon/bootstrap/DaemonMain.java
```diff
@@ -143,15 +127,81 @@ protected void doAction(String[] args, ExecutionListener listener) {
...
+                LOGGER.debug("Could not list daemon log directories for cleanup in: {}", daemonBaseDir.getAbsolutePath());
+                return;
+            }
+            for (File daemonLogDirectory : daemonLogDirectories) {
+                File[] logFiles = daemonLogDirectory.listFiles(f -> f.isFile() && f.getName().endsWith(DAEMON_LOG_SUFFIX) && f.getName().startsWith(DAEMON_LOG_PREFIX));
+                if (logFiles == null) {
+                    LOGGER.debug("Could not list log files for cleanup in: {}", daemonLogDirectory.getAbsolutePath());
+                    return;
+                }
+                long maxAge = System.currentTimeMillis() - FOURTEEN_DAYS_MILLIS;
```
eskatos: ❌ This should be pulled up outside the outer for loop.

platforms/core-runtime/daemon-server/src/main/java/org/gradle/launcher/daemon/bootstrap/DaemonMain.java
```diff
@@ -143,15 +127,81 @@ protected void doAction(String[] args, ExecutionListener listener) {
...
+
+    /**
+     * Removes all log files in the given folder that haven't been modified for at least 2 weeks
+     *
+     * @param daemonBaseDir The currently used log file
+     */
+    public static void cleanupOldLogFiles(File daemonBaseDir) {
+        try {
+            File[] daemonLogDirectories = daemonBaseDir.listFiles(file ->
+                file.isDirectory() && DefaultGradleVersion.VERSION_PATTERN.matcher(file.getName()).matches());
```
eskatos, line 176: ❔ Do we need to match the sub-directories names given we then match the daemon log file names?

platforms/core-runtime/daemon-server/src/main/java/org/gradle/launcher/daemon/bootstrap/DaemonMain.java
```diff
@@ -143,15 +127,81 @@ protected void doAction(String[] args, ExecutionListener listener) {
...
+            if (daemonLogDirectories == null) {
+                LOGGER.debug("Could not list daemon log directories for cleanup in: {}", daemonBaseDir.getAbsolutePath());
```
eskatos: 💅 After reading the contract of `File.listFiles(filter)` again, I think this should be a `warn` log statement. Same for the other one below.

platforms/core-runtime/daemon-server/src/main/java/org/gradle/launcher/daemon/bootstrap/DaemonMain.java
reinsch82: I guess we don't need to. But it narrows down to minimize unintended impact.
A scenario I could imagine: For some reason a user might want to copy a logging directory for investigative purposes. we'd also delete this copy. I guess we don't give any guarantees about what happens to stuff in this directory, but I still think it would be odd if we cleaned a copy too.

eskatos: If the copied directory name happens to match it will have the same effect though.
If a user wants to copy for investigation they should move things out of the Gradle User Home.
There is no guarantee about anything but user init scripts and `gradle.properties` file in the GUH.
I think exposing the `DefaultGradleVersion` constant and the extra processing aren't worth it. Not a strong opinion though.

eskatos (approved):

platforms/core-runtime/base-services/src/main/java/org/gradle/util/internal/DefaultGradleVersion.java
```diff
@@ -35,7 +35,7 @@
 import static org.gradle.internal.IoActions.uncheckedClose;
 
 public final class DefaultGradleVersion extends GradleVersion {
-    private static final Pattern VERSION_PATTERN = Pattern.compile("((\\d+)(\\.\\d+)+)(-(\\p{Alpha}+)-(\\w+))?(-(SNAPSHOT|\\d{14}([-+]\\d{4})?))?");
+    public static final Pattern VERSION_PATTERN = Pattern.compile("((\\d+)(\\.\\d+)+)(-(\\p{Alpha}+)-(\\w+))?(-(SNAPSHOT|\\d{14}([-+]\\d{4})?))?");
```
big-guy: Can we not make this public?

PR #35775, opened 2025-11-21
Provide a useful error message when starting tests under a JUnit TEST

octylFractal:
This is likely incorrect behavior
(https://github.com/junit-team/junit-framework/issues/5168) and previously resulted in an opaque error from TestEventReporterAsListener.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
+++ b/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
@@ -48,6 +48,8 @@ import org.gradle.api.tasks.testing.TestResult.ResultType;
 import org.gradle.api.tasks.testing.source.FilePosition;
 import org.gradle.api.tasks.testing.source.TestSource;
 import org.gradle.internal.MutableBoolean;
+import org.gradle.internal.UncheckedException;
+import org.gradle.internal.exceptions.DefaultMultiCauseException;
 import org.gradle.internal.id.CompositeIdGenerator;
 import org.gradle.internal.id.IdGenerator;
 import org.gradle.internal.time.Clock;
@@ -84,6 +86,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import static org.gradle.api.tasks.testing.TestResult.ResultType.SKIPPED;
 import static org.junit.platform.engine.TestExecutionResult.Status.ABORTED;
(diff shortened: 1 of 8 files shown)


platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -141,6 +148,12 @@ public JUnitPlatformTestExecutionListener(TestResultProcessor resultProcessor, C
         this.baseDefinitionsDir = baseDefinitionsDir;
     }
 
+    public void throwAnyFatalExceptions() {
+        for (Throwable fatalException : fatalExceptions) {
+            throw UncheckedException.throwAsUncheckedException(fatalException);
+        }
```
big-guy: Doesn't this only throw the first one?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/TestProcessingStartupFailureIntegrationTest.groovy
```diff
@@ -99,7 +99,7 @@ class TestProcessingStartupFailureIntegrationTest extends AbstractIntegrationSpe
         then: "Task failure is reported"
 
         failure.assertHasFailure("Execution failed for task ':test'.") {
-            it.assertHasCause("Could not execute test class 'MyOtherTest'.")
+            it.assertHasCause("Could not execute test class 'MyOtherTest': Incompatible magic value 1668248178 in class file MyOtherTest")
```
big-guy: Is this reported as the cause of "Could not execute test class" too?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/SuiteTestDefinitionProcessor.java
```diff
@@ -35,15 +35,22 @@ public SuiteTestDefinitionProcessor(TestDescriptorInternal suiteDescriptor, Test
         this.clock = clock;
     }
 
+    private static TestSuiteExecutionException createTestSuiteExecutionException(String baseMessage, Throwable cause) {
+        return new TestSuiteExecutionException(
+            // Include the cause message because we're a GradleException and by default the console will not show the cause message
+            baseMessage + (cause.getMessage() != null ? ": " + cause.getMessage() : "."),
```
big-guy: I wouldn't do this.  What does this look like without including the cause?

There are ways to make the console show the useful thing if it's a cause. Have you tried marking TestSuiteExecutionException `@Contextual`?

octylFractal: I had done this since it was how it was before, but if you'd rather make it `@Contextual`, I made https://github.com/gradle/gradle/pull/36265. I think it's an improvement, if you agree then let's merge https://github.com/gradle/gradle/pull/36265 first and then I'll fix this one.

PR #35777, opened 2025-11-22
Optimize zip->zip copies when data is DEFLATED

octylFractal:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


octylFractal:
@bot-gradle test PT

octylFractal:
It seems like without better size information this won't work out well. Plus, it didn't seem to get as much perf wins as I hoped.

PR #35794, opened 2025-11-24
[Investigation] Simulate IntelliJ sync

asodja:


jbartok:
Incorporated into 
* https://github.com/gradle/gradle/pull/35804

PR #35796, opened 2025-11-24
Simplify `AnnotatedServiceLifecycleHandler` infrastructure

bamboo:
By implementing service scope validation directly as part of `ScopedServiceRegistry`.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/ServiceScopeValidator.java
+++ b/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/ServiceScopeValidator.java
@@ -20,12 +20,9 @@ import org.gradle.internal.service.scopes.Scope;
 import org.gradle.internal.service.scopes.ServiceScope;
 import org.jspecify.annotations.Nullable;
 
-import java.lang.annotation.Annotation;
 import java.util.ArrayDeque;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
-import java.util.List;
 import java.util.Queue;
 import java.util.Set;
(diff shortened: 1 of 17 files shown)


platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/ServiceScopeValidator.java
```diff
@@ -50,19 +45,8 @@ public ServiceScopeValidator(Class<? extends Scope> scope, boolean strict) {
...
-
-    @Override
-    public Class<? extends Annotation> getImplicitAnnotation() {
-        return ServiceScope.class;
-    }
-
-    @Override
-    public void whenRegistered(Class<? extends Annotation> annotation, Registration registration) {
-        for (Class<?> declaredType : registration.getDeclaredTypes()) {
+    void validate(DefaultServiceRegistry.SingletonService registration) {
```
alllex, line 48: 💭 I am not sure I like the type of argument we end up having here. There is a lot of similarity with `AnnotatedServiceLifecycleHandler.Registration`. Though the latter is part of the public API of the service framework, while the validator is internal.

Maybe a better approach is to make SingletonService an interface that provides the declared types, and merge whatever implementation the SingletonService class has today into the `ManagedObjectServiceProvider`, since the former is the only child of the latter in the class hierarchy.

alllex (commented):
Looks good overall!

bamboo:
@bot-gradle test RFN

PR #35801, opened 2025-11-24
Allow two nodes in the same component to capabiltiy conflict

jvandort:
Capability conflict resolution was based on prior code for module conflict resolution. The nature of module conflict resolution means that if one module wins, all nodes in that module wins. However, capability conflict detection is by definition node-based. Two nodes in the same component may conflict with each other. We had a broken attempt at handling this, the 'evicted' field for a node, but we never consulted that field when selecting variants. So, what would end up happening is we would resolve the conflict and mark a node as evicted. However, we would then clear any remaining edges pointing to it without proper bookkeeping. This lead to edges in the graph without a target node, as we never connected the evicted node's incoming edge.

In this commit, we ensure edges that would point to an evicted node properly point to the replacement node.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
@@ -92,7 +92,7 @@ public class NodeState implements DependencyGraphNode {
     // In opposite to outgoing edges, virtual edges are for now pretty rare, so they are created lazily
     private @Nullable List<EdgeState> virtualEdges;
     private boolean queued;
-    private boolean evicted;
+    private @Nullable NodeState replacement;
     private int transitiveEdgeCount;
     private @Nullable Set<ModuleIdentifier> upcomingNoLongerPendingConstraints;
 
@@ -648,13 +648,13 @@ public class NodeState implements DependencyGraphNode {
 
     void addIncomingEdge(EdgeState dependencyEdge) {
         if (!incomingEdges.contains(dependencyEdge)) {
+            cachedModuleResolutionFilter = null;
             incomingEdges.add(dependencyEdge);
             incomingHash += dependencyEdge.hashCode();
             if (dependencyEdge.isTransitive()) {
                 transitiveEdgeCount++;
             }
             requeueChildrenOfEndorsingParent(dependencyEdge);
-            cachedModuleResolutionFilter = null;
 
             if (incomingEdges.size() == 1) {
                 updateAncestorsStrictVersions(getStrictVersionsForEdge(dependencyEdge));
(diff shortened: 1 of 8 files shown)


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
```diff
@@ -1252,15 +1260,6 @@ boolean isSelectedByVariantAwareResolution() {
         return selectedByVariantAwareResolution && isSelected();
     }
 
-    void makePending(EdgeState edgeState) {
```
jvandort: The `removeOutgoingEdge` method just a few lines up is identical to this one.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
```diff
@@ -1123,36 +1158,32 @@ public void restart(ComponentState selected) {
...
             for (EdgeState edge : new ArrayList<>(incomingEdges)) {
                 edge.retarget();
             }
         }
-        // TODO: Restarting incoming edges should ensure they are pointing to the correct node.
-        // If they end up pointing to us after restart, we should not remove them.
-        clearIncomingEdges();
-    }
 
-    private void clearIncomingEdges() {
```
jvandort, line 1154: Renamed to `removeAllIncomingEdges`, and move up next to `addIncomingEdge` and `removeIncomingEdge`

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
```diff
@@ -1174,29 +1205,6 @@ void markForVirtualPlatformRefresh() {
...
-        if (incomingEdges.isEmpty()) {
-            return;
-        }
-        // Cleaning has to be done on a copied collection because of the recompute happening on selector removal
-        List<EdgeState> remainingIncomingEdges = ImmutableList.copyOf(incomingEdges);
-        clearIncomingEdges();
-        for (EdgeState incomingEdge : remainingIncomingEdges) {
-            assert incomingEdge.isConstraint();
-            NodeState from = incomingEdge.getFrom();
-            if (from != backToPendingSource) {
```
jvandort, line 1195: This existing logic was duplicated in `ModuleResolveState`.

jvandort:
@bot-gradle test this

PR #35811, opened 2025-11-25
Refine daemon logs cleanup to provide progress, build ops and configurability

### LINKED ISSUE #35803: Use existing infrastructure (like `VersionSpecificCacheCleanupAction`) to clean up daemon logs
This is a follow up of 

-  https://github.com/gradle/gradle/pull/35765

We want to use existing infrastructure to properly delete those logs.

Gnanasaikiran:
Fixes #35803

### Context

This PR follows up on #35765 (Delete daemon logs older than 2 weeks on shutdown) by refining the daemon logs cleanup to use existing Gradle cleanup infrastructure.

As mentioned in the TODO comment in DaemonMain.java, the cleanup should use `GradleUserHomeCleanupService` instead of being done directly in the daemon bootstrap code.

### Changes

1. **Created `DaemonLogsCleanupAction`** - A new cleanup action that implements `MonitoredCleanupAction` interface, following the same pattern as `VersionSpecificCacheCleanupAction`. This provides:
   - Progress reporting via `CleanupProgressMonitor`
   - Build operations integration for tracking
   - Configurable retention period through cache configurations
   - Proper error handling with `Deleter`

2. **Integrated with `GradleUserHomeCleanupService`** - Added daemon logs cleanup to the existing cleanup workflow, which runs during Gradle User Home cleanup.

### Benefits

- Consistent cleanup behavior across all Gradle cache/log cleanup operations
- Progress visibility during cleanup
- Build operation tracking for debugging and monitoring
- Configurable retention period through standard cache configuration

### Note

The old cleanup code in `DaemonMain.java` (the `cleanupOldLogFiles` method and related code) can be removed in a follow-up commit once this approach is validated.

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/cache/internal/GradleUserHomeCleanupService.java
+++ b/subprojects/core/src/main/java/org/gradle/cache/internal/GradleUserHomeCleanupService.java
@@ -34,6 +34,7 @@ import java.io.File;
 
 @ServiceScope(Scope.UserHome.class)
 public class GradleUserHomeCleanupService implements Stoppable {
+
     private final Deleter deleter;
     private final GradleUserHomeDirProvider userHomeDirProvider;
     private final GlobalScopedCacheBuilderFactory cacheBuilderFactory;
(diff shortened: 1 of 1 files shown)


6hundreds (changes requested):
> As mentioned in the TODO comment in DaemonMain.java, the cleanup should use GradleUserHomeCleanupService instead of being done directly in the daemon bootstrap code.

Given the above, we expect this PR takes care about removal the todo in `DaemonMain.java` and corresponding code. Also the PR should add some tests for the new code, to prove the behavior.

PR #35825, opened 2025-11-26
Only have ResilientGradleBuildBuilder

reinsch82:
This pull request refactors the Gradle IDE tooling model builders to simplify the codebase and improve resiliency by merging the resilient build logic into the main `GradleBuildBuilder`. The most important changes are:

### Refactoring and Code Simplification

* Merged the functionality of `ResilientGradleBuildBuilder` into `GradleBuildBuilder`, removing the separate resilient builder class and consolidating all logic for resilient model building into a single implementation. (`platforms/ide/ide/src/main/java/org/gradle/plugins/ide/internal/tooling/ResilientGradleBuildBuilder.java`, `platforms/ide/ide/src/main/java/org/gradle/plugins/ide/internal/tooling/GradleBuildBuilder.java`) [[1]](diffhunk://#diff-0a4335256e2aacc6daf6e6968b8a32894ec1bc0c036838d6938268922c528b23L1-L205) [[2]](diffhunk://#diff-6399537d11f81b0ebe7362c96a005b5f86223dbb118d4ede86050b62ac40e1c9R19-R67) [[3]](diffhunk://#diff-6399537d11f81b0ebe7362c96a005b5f86223dbb118d4ede86050b62ac40e1c9L51-R206)
* Updated the construction and registration of `GradleBuildBuilder` to always use the resilient logic, removing the conditional registration and the now-unnecessary factory method. (`platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/internal/tooling/ToolingModelServices.java`) [[1]](diffhunk://#diff-cc1bf80614864fc2c4d9c781c4da9a1398dd2fd2e24f5c95eac9a2c3d81d8378L76-R75) [[2]](diffhunk://#diff-cc1bf80614864fc2c4d9c781c4da9a1398dd2fd2e24f5c95eac9a2c3d81d8378L87-L92) [[3]](diffhunk://#diff-cc1bf80614864fc2c4d9c781c4da9a1398dd2fd2e24f5c95eac9a2c3d81d8378L33)

### Improved Resiliency and Failure Handling

* Enhanced `GradleBuildBuilder` to always handle failed builds and collect failures, ensuring the tooling model is robust even when some included builds fail to load. (`platforms/ide/ide/src/main/java/org/gradle/plugins/ide/internal/tooling/GradleBuildBuilder.java`)

### Test and Import Updates

* Updated imports in test files to reflect the new structure and dependencies of the refactored builder. (`platforms/ide/ide/src/test/groovy/org/gradle/plugins/ide/internal/tooling/GradleBuildBuilderTest.groovy`)

These changes make the IDE tooling model more maintainable and resilient by unifying the build logic and ensuring consistent handling of partial or failed builds.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
+++ b/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
@@ -165,6 +165,10 @@ class CustomPlugin implements Plugin<Project> {
             pluginManagement {
                 $repositoriesBlock
             }
+
+            dependencyResolutionManagement {
+                $repositoriesBlock
+            }
         """
         included.file("build.gradle.kts") << """
             plugins {
@@ -218,7 +222,8 @@ class CustomPlugin implements Plugin<Project> {
 
         @Override
         ModelResult execute(BuildController controller) {
-            GradleBuild gradleBuild = controller.getModel(GradleBuild.class)
+            GradleBuild gradleBuild = controller.fetch(GradleBuild.class).model
+            assert gradleBuild != null
             List<String> successfulQueriedProjects = []
             List<String> failedQueriedProjects = []
             if (queryStrategy == ROOT_BUILD_FIRST) {


reinsch82:
@bot-gradle test this

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/DefaultBuildController.java
```diff
@@ -125,16 +125,26 @@ private ToolingModelBuilderResultInternal doGetModel(@Nullable Object target, To
...
+            return ToolingModelBuilderResultInternal.of(model);
+        }
+        ToolingModelBuilderResultInternal resultInternal = (ToolingModelBuilderResultInternal) model;
+        if (modelRequestContext.inResilientContext()) {
+            return resultInternal;
+        }
+        if (resultInternal.getFailures().isEmpty()) {
+            return resultInternal;
+        }
+        throw resultInternal.throwOriginal();
```
asodja: ❓ This looks weird, can we avoid this where we return a RuntimeException and we try to trick compiler?

We if we make `throwOriginal()` return `void` rather switch `if` around, e.g.:
```
        if (!resultInternal.getFailures().isEmpty()) {
            resultInternal.throwOriginal();
        }
        return resultInternal;
```

subprojects/core/src/main/java/org/gradle/tooling/provider/model/internal/ToolingModelBuilderResultInternal.java
```diff
@@ -16,30 +16,45 @@
...
     private final List<Failure> failures;
 
-    private ToolingModelBuilderResultInternal(Object model, List<Failure> failures) {
+    private ToolingModelBuilderResultInternal(@Nullable Object model, List<Failure> failures) {
         this.model = model;
         this.failures = failures;
     }
 
+    public RuntimeException throwOriginal() {
+        if(failures.isEmpty()) {
```
asodja: 💅 Formatting of `if(`

subprojects/core/src/main/java/org/gradle/tooling/provider/model/internal/ToolingModelBuilderResultInternal.java
```diff
@@ -16,30 +16,45 @@
...
+    private final @Nullable Object model;
     private final List<Failure> failures;
 
-    private ToolingModelBuilderResultInternal(Object model, List<Failure> failures) {
+    private ToolingModelBuilderResultInternal(@Nullable Object model, List<Failure> failures) {
         this.model = model;
         this.failures = failures;
     }
 
+    public RuntimeException throwOriginal() {
```
asodja: 💭   We could also change this method to `throwFailureIfPresent()`

And logic would be:
```
if (failures.isEmpty()) {
   return;
}
throw failure logic...
```

That way you wouldn't need to do a check `if (!result.failures.isEmpty())` in places where you want to throw if there is a failure.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/FetchBuildActionCrossVersionSpec.groovy
```diff
@@ -0,0 +1,255 @@
...
+import org.gradle.integtests.tooling.r16.CustomModel
+import org.gradle.util.GradleVersion
+import org.gradle.integtests.tooling.r930.FetchBuildActionCrossVersionSpec.FetchCustomModelPerProjectAction
+import org.gradle.integtests.tooling.r930.FetchBuildActionCrossVersionSpec.FetchCustomModelAction
+import org.gradle.integtests.tooling.r930.FetchBuildActionCrossVersionSpec.FetchGradleBuildAction
+import org.gradle.integtests.tooling.r930.FetchBuildActionCrossVersionSpec.FetchUnknownModelAction
+
+@ToolingApiVersion(">=9.3.0")
+@TargetGradleVersion(">=9.4.0")
+class FetchBuildActionCrossVersionSpec extends ToolingApiSpecification {
```
asodja, line 26: ❓ Is that similar test as `org.gradle.integtests.tooling.r930.FetchBuildActionCrossVersionSpec`?

What is the difference between this and `r930.FetchBuildActionCrossVersionSpec`` version? 
Is that the recommended approach when testing TAPI and there is small difference between version behaviour?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/ResilientGradleBuildSyncCrossVersionSpec.groovy
```diff
@@ -43,7 +44,9 @@ class ResilientGradleBuildSyncCrossVersionSpec extends ToolingApiSpecification {
     BuildActionResult runFetchModelAction(Consumer<BuildActionExecuter<BuildActionResult>> configurer = {}) {
         succeeds {
             def action = action(new FetchModelAction())
-                .withArguments(RESILIENT_MODEL_TRUE)
+            if (targetVersion < GradleVersion.version("9.4.0")){
+                action.withArguments(RESILIENT_MODEL_TRUE)
```
asodja, line 50: 💅 Let's maybe add a comment why we need flag before 9.4

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/DefaultBuildController.java
asodja: I think even better would be, to  handle this `!resultInternal.getFailures().isEmpty()` in `resultInternal` as suggested before and then simplify logic here to:
```
        resultInternal.throwFailureIfPresent();
        return resultInternal;
```

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/BuildModelActionRunner.java
```diff
@@ -80,7 +81,16 @@ public void beforeTasks(BuildTreeModelController controller) {
         public Object fromBuildModel(BuildTreeModelController controller) {
             String modelName = buildModelAction.getModelName();
             try {
-                return controller.getModel(BuildTreeModelTarget.ofDefault(), new ToolingModelRequestContext(modelName, null, false));
+                ToolingModelRequestContext modelRequestContext = new ToolingModelRequestContext(modelName, null, false);
+                Object model = controller.getModel(BuildTreeModelTarget.ofDefault(), modelRequestContext);
+                if (model instanceof ToolingModelBuilderResultInternal) {
```
asodja: ❌ As discussed on Slack: We should either move the logic and the logic in `DefaultBuildController` to `AbstractToolingScope` or reuse this logic between `BuildModelActionRunner` and `DefaultBuildController`

platforms/ide/ide/src/main/java/org/gradle/plugins/ide/internal/tooling/GradleBuildBuilder.java
```diff
@@ -110,8 +120,90 @@ protected void addAllImportableBuilds(BuildState targetBuild, GradleInternal gra
...
+
+            ensureProjectsLoaded(targetBuild);
+
+            GradleInternal gradle = targetBuild.getMutableModel();
+            addProjectsAndBuildIdentifier(targetBuild, model);
+            try {
+                addFailedBuilds(targetBuild, model);
+                addIncludedBuilds(gradle, model);
+            } catch (IllegalStateException e) {
+                //ignore, happens when included builds are not accessible, but we need this for resiliency
```
asodja: ❓ Do you know in which cases these exception happen? Should we add these exceptions to `failures`?

asodja (changes requested):

platforms/ide/ide/src/main/java/org/gradle/plugins/ide/internal/tooling/GradleBuildBuilder.java
reinsch82: They happen when the included builds list is not yet set, because evaluating settings (or maybe the build file) didn't work

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/FetchBuildActionCrossVersionSpec.groovy
reinsch82: not a lot, I removed the plain duplicates here.

subprojects/core/src/main/java/org/gradle/internal/buildtree/BuildTreeModelTarget.java
```diff
@@ -44,13 +44,22 @@ public static Project ofProject(File buildRootDir, String projectPath) {
         return new Project(buildRootDir, Path.path(projectPath));
     }
 
+    public String describeTargetScope(){
```
asodja: 💯 That is a nice touch

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/DefaultBuildController.java
```diff
@@ -177,15 +173,23 @@ public void dispatch(Object value) {
...
-            ToolingModelBuilderResultInternal model = doGetModel(target, new ToolingModelRequestContext(modelIdentifier.getName(), parameter, true));
-            List<InternalFailure> failures = toInternalFailures(model.getFailures());
-            return new DefaultInternalFetchModelResult<>(uncheckedNonnullCast(model.getModel()), failures);
+            Object model = doGetModel(target, new ToolingModelRequestContext(modelIdentifier.getName(), parameter, true));
+            if (model instanceof ToolingModelBuilderResultInternal) {
+                ToolingModelBuilderResultInternal resultInternal = (ToolingModelBuilderResultInternal) model;
+                List<InternalFailure> failures = toInternalFailures(resultInternal.getFailures());
+                return new DefaultInternalFetchModelResult<>(uncheckedNonnullCast(resultInternal.getModel()), failures);
+            }
+            return createFetchModelResult(new IllegalStateException("Expected a ToolingModelBuilderResultInternal, but got " + model)); //!?
```
asodja: ❌  This logic looks a bit funny. I think that type checking is a good argument we change return type of `ToolingModelScope.getModel` to `ToolingModelBuilderResultInternal`. And we do a transformation from Object to `ToolingModelBuilderResultInternal` already in AbstractToolingModelScope (we anyway already do that there).
That way we will always have `ToolingModelBuilderResultInternal` type here and no more "any" type. And we will have nicer logic here.

WDYT?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheAwareBuildToolingModelController.kt
```diff
@@ -50,10 +50,9 @@ class ConfigurationCacheAwareBuildToolingModelController(
         private val cache: BuildTreeConfigurationCache
     ) : ToolingModelScope {
         override fun getTarget() = delegate.target
-
-        override fun getModel(modelName: String, parameter: ToolingModelParameterCarrier?): Any? {
-            return cache.loadOrCreateIntermediateModel(target?.identity, modelName, parameter) {
-                delegate.getModel(modelName, parameter)
+        override fun getModel(modelRequestContext: ToolingModelRequestContext, parameter: ToolingModelParameterCarrier?): Any? {
```
asodja: ❌   That made me realize: we should also take into account "in resilient context" for caching. Otherwise caching won't be correct, since `getModel()` and `fetchModel()` behave a bit differently.

We need to fix that. We can do this in this PR or we do it in a separate PR after.

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/DefaultBuildController.java
reinsch82: I tried, this leads to a `ToolingModelBuilderResultInternal` instance in `BuildModelActionRunner`. 
We want to avoid that, so we don't need to do any special handling for non resilient mode there.

asodja: Would we just need to call `ToolingModelBuilderResultInternal.getModel()` In `BuildModelActionRunner`?

reinsch82:
@bot-gradle test this

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/BuildModelActionRunner.java
```diff
@@ -47,8 +49,8 @@ public Result run(BuildAction action, final BuildTreeLifecycleController buildCo
         ModelCreateAction createAction = new ModelCreateAction(buildModelAction);
         try {
             if (buildModelAction.isCreateModel()) {
-                Object result = buildController.fromBuildModel(buildModelAction.isRunTasks(), createAction);
-                SerializedPayload serializedResult = payloadSerializer.serialize(result);
+                ToolingModelBuilderResultInternal result = buildController.fromBuildModel(buildModelAction.isRunTasks(), createAction);
```
asodja, line 52: 🤔 Alternatively we could also call `.getModel()` already in `ModelCreateAction`, so return value of `ModelCreateAction` would be the same as before, and here we wouldn't deal with `ToolingModelBuilderResultInternal`

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/BuildModelActionRunner.java
```diff
@@ -77,10 +79,11 @@ public void beforeTasks(BuildTreeModelController controller) {
         }
 
         @Override
-        public Object fromBuildModel(BuildTreeModelController controller) {
+        public @Nullable ToolingModelBuilderResultInternal fromBuildModel(BuildTreeModelController controller) {
```
asodja, line 82: ❓ Is `ToolingModelBuilderResultInternal` really Nullable here?

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/DefaultBuildController.java
```diff
@@ -124,12 +124,7 @@ private ToolingModelBuilderResultInternal doGetModel(@Nullable Object target, To
 
         BuildTreeModelTarget scopedTarget = resolveTarget(target);
         try {
-            Object model = controller.getModel(scopedTarget, modelRequestContext);
-            if (model instanceof ToolingModelBuilderResultInternal) {
-                return (ToolingModelBuilderResultInternal) model;
-            } else {
-                return ToolingModelBuilderResultInternal.of(model);
-            }
+            return controller.getModel(scopedTarget, modelRequestContext);
```
asodja, line 127: 🎉 that looks nicer

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/FetchBuildActionCrossVersionSpec.groovy
```diff
@@ -0,0 +1,51 @@
...
+package org.gradle.integtests.tooling.r940
+
+import org.gradle.integtests.tooling.fixture.TargetGradleVersion
+import org.gradle.integtests.tooling.fixture.ToolingApiSpecification
+import org.gradle.integtests.tooling.fixture.ToolingApiVersion
+import org.gradle.integtests.tooling.r930.FetchBuildActionCrossVersionSpec.FetchGradleBuildAction
+
+@ToolingApiVersion(">=9.3.0")
+@TargetGradleVersion(">=9.4.0")
+class FetchBuildActionCrossVersionSpec extends ToolingApiSpecification {
```
asodja, line 26: ❓ This looks suspicious. For Gradle 9.4.0+ we have less coverage for fetch API

asodja: Ok, we have coverage in `FetchBuildActionCrossVersionSpec`. This just asserts a different behaviour from Gradle 9.4.0 compared to 9.3.0

asodja:
@bot-gradle test this

reinsch82:
@bot-gradle test this

PR #35827, opened 2025-11-26
Add simple TAPI filtering test

tresat:


platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -638,4 +646,57 @@ class NonClassBasedTestingIntegrationTest extends AbstractNonClassBasedTestingIn
         outputContains("INFO: Executing resource-based test: Test[file=SampleTest.rbt, name=foo]")
         resultsFor().assertTestPathsNotExecuted(":definitions.SampleTest:foo()")
     }
+
+    def "can filter resource-based test using TAPI with #entryPoint"() {
```
octylFractal, line 682: ❓ How is this filtering using TAPI? It looks like the only filter is in the build script?

tresat: Yes, but the test is that filtering works when running via TAPI, and makes the resource-based test events available.

octylFractal: I think the test name should be `can filter resource-based test when using TAPI with #entryPoint` then, because we do have filtering in TAPI and it's confusing the way it is.

PR #35829, opened 2025-11-26
Always group tests using the highest non-root node

### LINKED ISSUE #15075: The Gradle test runner's XML report lumps together unrelated @Nested tests under the same test/@classname
(Originally reported with the JUnit 5 project as https://github.com/junit-team/junit5/issues/2472:

> Since Gradle writes that XML report, we cannot fix that. Please open an issue on the Gradle side.

So here it goes:

## Steps to reproduce

1. Check out the `reproducer` branch from https://github.com/sewe/junit5-samples:
    ```sh
    git clone https://github.com/sewe/junit5-samples -b reproducer
    ```
2. ```sh
    pushd junit5-samples/junit5-jupiter-starter-maven
    ./mvnw test
    ls target/surefire-reports/TEST-*.xml
    ```
3. This produces a `TEST-*.xml` files for each of the three `Example?Test` classes (which all inherit from a common `BaseClass`). This is **good**.
4. ```sh
    popd
    pushd junit5-samples/junit5-jupiter-starter-gradle
    ./gradlew test
    ls build/test-results/test/TEST-*.xml
    ```
5. This produces, in addition to files for each of the three `Example?Test` classes, a file `TEST-com.example.BaseClass$NestedClass.xml` which lumps together the test results of the `@Nested` test inherited from `BaseClass` by both `Example1Test` and `Example2Test`. This is **bad**.
   Looking at the XML report you will see two **indistinguishable** test executions (`<failure>`s in this case, but the problem occurs for succesful tests as well):
   ```sh
   grep -r failure -B 1 'build/test-results/test/TEST-com.example.BaseClass$NestedClass.xml'
   ```

As these two executions are different, the should be communicated as such to upstream tools. But based on `TEST-com.example.BaseClass$NestedClass.xml`, it looks like the **same** test was executed twice.

## Context

 - Used versions (Jupiter/Vintage/Platform): JUnit 5.7, as used by `junit5-starter` (https://github.com/sewe/junit5-samples/tree/reproducer)

 - Build Tool/IDE:

```sh
./mvnw --version
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /home/sewe/.m2/wrapper/dists/apache-maven-3.6.3-bin/1iopthnavndlasol9gbrbg6bf2/apache-maven-3.6.3
Java version: 1.8.0_201, vendor: Oracle Corporation, runtime: /usr/lib/jvm/java-8-oracle/jre
Default locale: en_CA, platform encoding: UTF-8
OS name: "linux", version: "4.15.0-122-generic", arch: "amd64", family: "unix"
```

```sh
./gradlew --version

Gradle 6.2.2

Build time:   2020-03-04 08:49:31 UTC
Revision:     7d0bf6dcb46c143bcc3b7a0fa40a8e5ca28e5856

Kotlin:       1.3.61
Groovy:       2.5.8
Ant:          Apache Ant(TM) version 1.10.7 compiled on September 1 2019
JVM:          1.8.0_201 (Oracle Corporation 25.201-b09)
OS:           Linux 4.15.0-122-generic amd64
```

octylFractal:
This behavior more closely matches surefire.

Fixes #15075

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/cucumberjvm/CucumberJVMReportIntegrationTest.groovy
+++ b/platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/cucumberjvm/CucumberJVMReportIntegrationTest.groovy
@@ -20,6 +20,7 @@ import org.gradle.api.internal.tasks.testing.report.VerifiesGenericTestReportRes
 import org.gradle.api.internal.tasks.testing.report.generic.GenericTestExecutionResult
 import org.gradle.api.tasks.testing.TestResult
 import org.gradle.integtests.fixtures.AbstractSampleIntegrationTest
+import org.gradle.integtests.fixtures.JUnitXmlTestExecutionResult
 import org.gradle.integtests.fixtures.TestResources
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.UnitTestPreconditions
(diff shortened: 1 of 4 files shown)


octylFractal:
@bot-gradle test this

PR #35849, opened 2025-11-27
Make `DefaultServiceRegistry.OwnServices` partially thread-safe

### LINKED ISSUE #35385: Thread-unsafe code in DefaultServiceRegistry$OwnServices leads to CME
### Current Behavior

A CME occurs occasionally in this class, see this scan from our build: https://ge.gradle.org/s/b7uxt7lw6ckws/failures?expanded-stacktrace=WyIwLTEiXQ#1

### Expected Behavior

No CME should occur.

### Context (optional)

Failed to run a build locally while doing normal work.

This doesn't seem to be a regression as the relevant code hasn't been touched in a while.

This should be fixable by just making `lifecycleHandlers` a `CopyOnWriteArrayList`, if it's indeed low-write high-read like I think it is.

I don't know if I can make one of these as the issue is a race condition I only got lucky to hit. The fix seems obvious regardless.

### Gradle version

9.2.0-rc-2

### Build scan URL (optional)

https://ge.gradle.org/s/b7uxt7lw6ckws

_No response_

bamboo:
By serializing the related state transitions of `services` and `lifecycle handlers` via an atomic reference.

Operations on `OwnServices#providersByType` remain thread-unsafe and will be dealt with in a subsequent PR.

Resolves #35385

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
+++ b/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
@@ -518,7 +518,7 @@ public class DefaultServiceRegistry implements CloseableServiceRegistry, Contain
 
         void annotationHandlerCreated(AnnotatedServiceLifecycleHandler annotationHandler) {
             ServicesSnapshot snapshot = services.updateAndGet(it -> it.addLifecycleHandler(annotationHandler));
-            @Nullable ServiceList list = snapshot.services;
+            ServiceList list = snapshot.services;
             while (list != null) {
                 notifyAnnotationHandler(annotationHandler, list.service);
                 list = list.next;
@@ -1419,7 +1419,16 @@ public class DefaultServiceRegistry implements CloseableServiceRegistry, Contain
         }
     }
 
-    static class ServicesSnapshot {
+    /**
+     * Carries a snapshot of the current set of services and lifecycle handlers so they can change together.
+     *
+     * Lifecycle handlers are maintained in a copy-on-write array since there are at most 3 lifecycle handler instances
+     * per registry, and they are iterated frequently (for every service registration).
+     *
+     * Services are maintained in a linked list since there are many, they are frequently written and iterated very
+     * rarely (once per lifecycle handler).
+     */
+    private static class ServicesSnapshot {
         static final ServicesSnapshot EMPTY = new ServicesSnapshot(null, new AnnotatedServiceLifecycleHandler[0]);
 
         final @Nullable ServiceList services;
@@ -1452,7 +1461,7 @@ public class DefaultServiceRegistry implements CloseableServiceRegistry, Contain
         }
     }
 
-    static class ServiceList {
+    private static class ServiceList {
         final SingletonService service;
         final @Nullable ServiceList next;
 


bamboo:
@bot-gradle test RFN

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -461,8 +460,8 @@ public void add(SingletonService serviceProvider) {
             assertMutable();
             stoppable.add(serviceProvider);
             collectProvidersForClassHierarchy(inspector, serviceProvider.getDeclaredServiceTypes(), serviceProvider);
```
alllex, line 462: ❓ The change fixes the services list update, but the `providersByType` update remains unsynchronized.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -518,9 +517,11 @@ public void instanceRealized(List<Class<?>> declaredServiceTypes, Supplier<Strin
         }
 
         void annotationHandlerCreated(AnnotatedServiceLifecycleHandler annotationHandler) {
-            lifecycleHandlers.add(annotationHandler);
-            for (SingletonService candidate : services) {
-                notifyAnnotationHandler(annotationHandler, candidate);
+            ServicesSnapshot snapshot = services.updateAndGet(it -> it.addLifecycleHandler(annotationHandler));
+            @Nullable ServiceList list = snapshot.services;
```
alllex: 💅 IntelliJ says that Nullable annotations on local variables are not applicable. Do we need to keep it here?

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -1417,4 +1418,47 @@ private boolean locateAnnotation(Class<? extends Annotation> annotation) {
             }
         }
     }
+
+    static class ServicesSnapshot {
```
alllex: 💅 Class visibility can be private

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -393,8 +393,7 @@ public void visit(Service service) {
     private class OwnServices implements ServiceProvider {
         private final Map<Class<?>, List<ServiceProvider>> providersByType = new HashMap<Class<?>, List<ServiceProvider>>(16, 0.5f);
         private final CompositeStoppable stoppable = CompositeStoppable.stoppable();
-        private final List<SingletonService> services = new ArrayList<SingletonService>();
-        private final List<AnnotatedServiceLifecycleHandler> lifecycleHandlers = new ArrayList<AnnotatedServiceLifecycleHandler>();
+        private final AtomicReference<ServicesSnapshot> services = new AtomicReference<>(ServicesSnapshot.EMPTY);
```
asodja, line 396: 🤔 An alternative approach with similar behaviour could be to use Java's `ConcurrentLinkedQueue` for `services` and `lifecycleHandlers`. With that we could avoid having extra `ServicesSnapshot` class and keep the same structure

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -1417,4 +1418,47 @@ private boolean locateAnnotation(Class<? extends Annotation> annotation) {
             }
         }
     }
+
+    static class ServicesSnapshot {
+        static final ServicesSnapshot EMPTY = new ServicesSnapshot(null, new AnnotatedServiceLifecycleHandler[0]);
+
+        final @Nullable ServiceList services;
+        final AnnotatedServiceLifecycleHandler[] lifecycleHandlers;
```
asodja, line 1435: ❓  Why we use `ServiceList` for services but we use arrays and `copy-on-write` for `lifecycleHandlers`?

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -518,9 +517,11 @@ public void instanceRealized(List<Class<?>> declaredServiceTypes, Supplier<Strin
         }
 
         void annotationHandlerCreated(AnnotatedServiceLifecycleHandler annotationHandler) {
-            lifecycleHandlers.add(annotationHandler);
-            for (SingletonService candidate : services) {
-                notifyAnnotationHandler(annotationHandler, candidate);
+            ServicesSnapshot snapshot = services.updateAndGet(it -> it.addLifecycleHandler(annotationHandler));
+            @Nullable ServiceList list = snapshot.services;
+            while (list != null) {
```
Pankraz76, line 522: would like to consider raw recursion up front to point this **delicacy** out. this woud wipe out the nullable as this is the start param to start it.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
asodja: Although thinking about it more, it's not obvious that using `ConcurrentLinkedQueue` will result in correct behaviour if we add a service and a lifecycleHandler concurrently

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -461,8 +460,8 @@ public void add(SingletonService serviceProvider) {
             assertMutable();
             stoppable.add(serviceProvider);
             collectProvidersForClassHierarchy(inspector, serviceProvider.getDeclaredServiceTypes(), serviceProvider);
-            services.add(serviceProvider);
-            for (AnnotatedServiceLifecycleHandler annotationHandler : lifecycleHandlers) {
+            ServicesSnapshot snapshot = services.updateAndGet(it -> it.addService(serviceProvider));
+            for (AnnotatedServiceLifecycleHandler annotationHandler : snapshot.lifecycleHandlers) {
```
Pankraz76, line 464:
```suggestion
            for (AnnotatedServiceLifecycleHandler annotationHandler : snapshot.services.updateAndGet(it -> it.addService(serviceProvider))) {
```
could inline to keep high cohesion promise reduce coupling context. Considering snapshot service will probably provide some snapshot.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
Pankraz76, line 396:
```suggestion
        private final AtomicReference<ServicesSnapshot> services = new AtomicReference<>(EMPTY);
```
having ServicesSnapshot already read in first part, again cohesion low coupling.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
bamboo: There are at most 3 lifecycle handler instances per registry and they are iterated on frequently (for every registration).

There are many more services and they are written a lot but iterated very rarely (3 times at most).

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
bamboo: > Although thinking about it more, it's not obvious that using `ConcurrentLinkedQueue` will result in correct behaviour if we add a service and a lifecycleHandler concurrently

Yes. That setup could violate the "once-and-only-once" contract for `AnnotatedServiceLifecycleHandler#whenRegistered`.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
bamboo: I tend to think of `@Nullable` as part of the type declaration, even if the tooling won't help, it seems more precise to me.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
bamboo: Indeed. I focused on the concrete reports but I agree it would be better and more reassuring to have complete thread-safety. 

I'll reassess my solution.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
Pankraz76: could leverage prone to have this automate giving freedom and security at once.

- https://errorprone.info/bugpattern/EffectivelyPrivate

like how many effort have you invested and want to increase in the future on this topic, or kind of fix this global. This is kind of the real question in general, imho.

sorry should have checked before, discovered:

Still it could be reconsidered, if its easy to fix. (default) Scope private is next to immutability (final), a good thing to have for robust software, avoiding to much envy and side effects.

- https://github.com/gradle/gradle/blob/1757939b3bb4bebb6a854ea775db5040244df746/build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.code-quality.gradle.kts#L57

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
alllex: I would agree if the tooling stayed neutral, but IDEA highlights the `@Nullable` in yellow, making me think there is a problem, where there is none. So tooling is actually actively "unhelpful".

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
mlopatkin: > There are at most 3 lifecycle handler instances per registry and they are iterated on frequently (for every registration).
> 
> There are many more services and they are written a lot but iterated very rarely (3 times at most).

I stumbled upon this as well, so this should be probably a comment.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -1417,4 +1418,47 @@ private boolean locateAnnotation(Class<? extends Annotation> annotation) {
...
+
+        private static AnnotatedServiceLifecycleHandler[] append(AnnotatedServiceLifecycleHandler[] array, AnnotatedServiceLifecycleHandler annotationHandler) {
+            AnnotatedServiceLifecycleHandler[] newArray = new AnnotatedServiceLifecycleHandler[array.length + 1];
+            System.arraycopy(array, 0, newArray, 0, array.length);
+            newArray[array.length] = annotationHandler;
+            return newArray;
+        }
+    }
+
+    static class ServiceList {
```
mlopatkin: 🤔 Why not use a [PersistentList](https://github.com/gradle/gradle/blob/HEAD/platforms/core-runtime/functional/src/main/java/org/gradle/internal/collect/PersistentList.java)?

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
mlopatkin: >could leverage prone to have this automate giving freedom and security at once.

AFAIK, this diagnostic goes in other direction: it complains about public stuff on private classes, so it won't help here.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
mlopatkin: >I would agree if the tooling stayed neutral, but IDEA highlights the @Nullable in yellow, making me think there is a problem, where there is none. So tooling is actually actively "unhelpful".

FWIW, JSpecify folks explicitly advice against nullness annotations on local variables: https://jspecify.dev/docs/user-guide/#local-variables. I don't know the motivation, perhaps they wanted to reduce visual clutter.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
bamboo: > I stumbled upon this as well, so this should be probably a comment.

Will do.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
bamboo: The `functional` module is currently not available in workers. In its current form it'd also bring `guava` and `fastutil` dependencies so I thought a custom node class was a good compromise.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
bamboo: Ok.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
bamboo: @alllex any objection to extending the solution to `providersByType` in a subsequent PR?

bamboo (commented):
Thanks for the feedback. Applied.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
alllex: No objection, but then we should update the PR name to note that this is only a partial thread-safety fix, and add a note to the PR description on what is still missing.

bamboo:
> If we merge this as is, we should update the PR name and description to clarify this.

Done.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
alllex: For posterity, this is a follow-up issue that was eventually addressed:
- https://github.com/gradle/gradle/issues/35923

PR #35859, opened 2025-11-28
Making setting resolution report problems when multiple scripts are present

### LINKED ISSUE #35840: Settings scripts should report a warning when multiple candidates are present

hegyibalint:
Fixes #35840


hegyibalint:
@bot-gradle  test this

hegyibalint:
@bot-gradle test this

hegyibalint:
@bot-gradle cancel

platforms/core-runtime/build-discovery/src/test/groovy/org/gradle/internal/scripts/DefaultScriptFileResolverTest.groovy
```diff
@@ -0,0 +1,153 @@
...
+        when:
+        def resolver = new DefaultScriptFileResolver()
+        def result = resolver.resolveScriptFile(testDir, "custom")
+
+        then:
+        result.selectedCandidate == selected
+        result.ignoredCandidates == recognizedButIgnored
+    }
+
+    def "listener should only be notified once for custom-named build files"() {
```
eskatos: 💭 It would be good to add more tests where the default extension file does not exist and assert on the listener notification count and files.

For example, for a test with `a.gradle.kts` and `a.gradle.dcl` but no `a.gradle` I would expect 2 notifications for `a.gradle` and `a.gradle.kts`.

The rationale is that if `a.gradle` gets created, CC should be invalidated.

I think we should cover all possible cases here to be sure we address them all properly.

hegyibalint: This is a good point! 
In addition, I was very happy-path focused, and forgot to check what happens when no file is present; the `*.gradle` file should still be notified in that case.

hegyibalint: Made a more consolidated test that now should cover the concerns above. PTAL

hegyibalint:
@bot-gradle test APT

blindpirate:
This test is failing: https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_Quick_1_Trigger/107577584?hideTestsFromDependencies=true

hegyibalint:
@blindpirate I have no idea why that test started to break.
I cannot reproduce it either locally, or by [running the same pipeline on the feature branch](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_Quick_1_bucket1/107590394)

Also the failure mode is quite strange:
```
groovy.lang.GroovyRuntimeException: Could not find matching constructor for: org.gradle.internal.scripts.DefaultScriptFileResolver(org.gradle.internal.scripts.DefaultScriptFileResolverTest$CountingListener)
    at org.gradle.internal.scripts.DefaultScriptFileResolverTest.resolution finds and notifies the correct script file(DefaultScriptFileResolverTest.groovy:61)
```

But the constructor is only called [here](https://github.com/gradle/gradle/pull/35859/changes#diff-7a51fc6bc0c2f57f0b792c5332c32bb5d0c74851f7a409eab676e6a6c3b171e3R58), and it doesn't seem to be a problem for me.

Any ideas?

hegyibalint:
Lack of a better idea, I will just try to merge it in again to see if this problem persist or not.

hegyibalint:
Found the issue, there wa a nice listener architecture implemented since I last merged `master`.

hegyibalint:
@bot-gradle test this

subprojects/core/src/integTest/groovy/org/gradle/initialization/MultipleSettingsFilesIntegrationTest.groovy
```diff
@@ -0,0 +1,205 @@
...
+            definition.id.fqid == 'scripts:multiple-scripts'
+            definition.id.displayName == 'Multiple scripts'
+            contextualLabel.contains("Multiple settings script files were found in directory")
+            contextualLabel.contains("included")
+            details.contains("Selected 'settings.gradle'")
+            details.contains("ignoring 'settings.gradle.kts'")
+        }
+    }
+
+    def "warns even when only using Kotlin settings file in practice"() {
```
alllex, line 184: 🤔 I'm confused about this test. The comments hint that it should test an invalid Groovy settings file, but the file is perfectly valid and the build succeeds.

Overall, the logic of this test seems to be fully covered by `uses the selected settings file (Groovy over Kotlin)`.

platforms/core-runtime/build-discovery/src/test/groovy/org/gradle/internal/scripts/DefaultScriptFileResolverTest.groovy
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    /**
+     * Generate a list of filenames with all known scripting language extensions.
+     *
+     * @param basename the base name of the file, e.g `build`
+     * @return a list of filenames with all known extensions, e.g. `['build.gradle', 'build.gradle.kts', 'build.gradle.dcl']`
+     */
+    static List<String> withExtensions(String basename) {
```
alllex, line 112: 🔎 unused method

platforms/core-runtime/build-discovery-impl/src/main/java/org/gradle/initialization/SettingsLocation.java
```diff
@@ -37,12 +38,22 @@ public File getSettingsDir() {
         return settingsDir;
     }
 
+    /**
+     * Returns the resolution result for the settings file.
+     */
+    @Nullable
```
alllex, line 44: 💭 We should clarify what `null` means for this. Is it the exact same as for the settings file getter? Slightly different?

.teamcity/subprojects.json
```diff
@@ -132,6 +132,13 @@
     "functionalTests": false,
     "crossVersionTests": false
   },
+  {
+    "name": "build-discovery-reporting",
+    "path": "platforms/core-runtime/build-discovery-reporting",
```
alllex, line 137: 🤔 What is the advantage of introducing a new subproject, instead of reusing `build-discovery-impl`?

PR #35863, opened 2025-11-28
Make ErrorProne suppressions local

ov7a:
1. It makes suppressions more visible
2. It prevents adding more code that triggers suppressed warnings.

I also addressed some (hopefully) obvious violations 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ov7a:
@bot-gradle test this

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/dsl/DefaultComponentMetadataProcessor.java
```diff
@@ -95,25 +94,9 @@ private static ModuleComponentResolveMetadata realizeMetadata(ModuleComponentRes
...
-            }
-        }
-        try {
-            ModuleComponentResolveMetadata forceRead = serializer.read(new InputStreamBackedDecoder(new ByteArrayInputStream(bytes)));
-            // TODO: CC cannot enable this assertion because moduleSource is not serialized, so doesn't appear in the deserialized form
-            //assert metadata.equals(forceRead);
-            metadata = forceRead;
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
+            serializer.write(new OutputStreamBackedEncoder(baos), metadata);
+            metadata = serializer.read(new InputStreamBackedDecoder(new ByteArrayInputStream(baos.toByteArray())));
```
reinsch82, line 101: 🎉 🎉

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/parser/MetaDataParser.java
```diff
@@ -36,11 +38,12 @@ public interface MetaDataParser<T extends MutableModuleComponentResolveMetadata>
      * moved from publishing `pom` only to also publish `module` files.
      */
     String GRADLE_6_METADATA_MARKER = "do_not_remove: published-with-gradle-metadata";
-    String[] GRADLE_METADATA_MARKER_COMMENT_LINES = {
+    List<String> GRADLE_METADATA_MARKER_COMMENT_LINES = ImmutableList.of(
             "This module was also published with a richer model, Gradle metadata, ",
             "which should be used instead. Do not delete the following line which ",
             "is to indicate to Gradle or any Gradle module metadata file consumer ",
-            "that they should prefer consuming it instead."};
+        "that they should prefer consuming it instead."
```
reinsch82, line 45: 💅 
```suggestion
            "that they should prefer consuming it instead."
```

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/oldresult/TransientConfigurationResultsBuilder.java
```diff
@@ -137,14 +137,10 @@ public void visitArtifacts(DependencyGraphNode from, LocalFileDependencyMetadata
...
             return cache.load(() -> {
-                try {
-                    return binaryData.read(decoder -> deserialize(decoder, artifactResults, buildOperationExecutor, resolutionHost));
-                } finally {
-                    try {
-                        binaryData.close();
-                    } catch (IOException e) {
-                        throw throwAsUncheckedException(e);
-                    }
+                try (BinaryStore.BinaryData reader = binaryData) {
```
reinsch82, line 140: 🎉

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/attributes/matching/AttributeSelectionSchema.java
```diff
@@ -37,7 +37,7 @@ public interface AttributeSelectionSchema {
     boolean hasAttribute(Attribute<?> attribute);
 
     /**
-     * Given a set of {@code candidate} attribute values for a given {@code attribute}, produce
+     * Given a set of {@code candidates} attribute values for a given {@code attribute}, produce
```
reinsch82: 💅 I actually think the original is correct here, it is used as an adjective here, therefore it should be singular.
Maybe one of our native speakers should way in here.

```suggestion
     * Given a set of {@code candidate} attribute values for a given {@code attribute}, produce
```

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/consumer/SynchronizedLogging.java
```diff
@@ -29,7 +29,7 @@
  * Provides logging services per thread.
  */
 public class SynchronizedLogging implements LoggingProvider {
-    private final ThreadLocal<ThreadLoggingServices> services = new ThreadLocal<ThreadLoggingServices>();
+    private final static ThreadLocal<ThreadLoggingServices> SERVICES = new ThreadLocal<>();
```
Pankraz76: is this naming some prone rule?


only picnic has this rule to fix once and globally.

- https://error-prone.picnic.tech/bugpatterns/ConstantNaming/

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/attributes/matching/AttributeSelectionSchema.java
ov7a: Ok, I'll rephrase to have proper ref to the param

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/consumer/SynchronizedLogging.java
alllex: ❓  @ov7a, I haven't looked into the implementation, but adding `static` is a semantic change that has implications. I would be skeptical about changing this as part of a large PR without futher clarification.

subprojects/composite-builds/src/main/java/org/gradle/composite/internal/DefaultIncludedBuildTaskGraph.java
```diff
@@ -65,7 +65,7 @@ private enum State {
     private final int monitoringPollTime;
     private final TimeUnit monitoringPollTimeUnit;
     private final ManagedExecutor executorService;
-    private final ThreadLocal<DefaultBuildTreeWorkGraph> current = new ThreadLocal<>();
+    private final static ThreadLocal<DefaultBuildTreeWorkGraph> CURRENT = new ThreadLocal<>();
```
alllex: ❓  Same question here: are we sure it's the right change to make it `static`?

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/consumer/SynchronizedLogging.java
ov7a: Ok, I'll revert that if we don't have confidence

subprojects/composite-builds/src/main/java/org/gradle/composite/internal/DefaultIncludedBuildTaskGraph.java
Pankraz76: Just my 2cts and question, I definitely don't know, but it will hopefully make a Single Source of Truth across all instances, which seems kind of fair—to align all of them to obtain one and the same field value named `current`.

If this is not a problem, or if the intent is just that, then this should add up to the story, imho.

subprojects/core/src/main/java/org/gradle/api/internal/DelegatingDomainObjectSet.java
```diff
@@ -171,7 +171,7 @@ public Object[] toArray() {
     }
 
     @Override
-    public <T> T[] toArray(T[] a) {
+    public <AT> AT[] toArray(AT[] a) {
```
alllex: 🤔  It's a very unconventional name for a generic type. Let's maybe use `S`, since it is used in other methods?

platforms/native/platform-native/src/main/java/org/gradle/nativeplatform/toolchain/internal/AbstractPlatformToolProvider.java
```diff
@@ -157,7 +157,7 @@ public <T extends CompileSpec> org.gradle.language.base.internal.compile.Compile
         if (StripperSpec.class.isAssignableFrom(spec)) {
             return CompilerUtil.castCompiler(createStripper());
         }
-        throw new IllegalArgumentException(String.format("Don't know how to compile from a spec of type %s.", spec.getClass().getSimpleName()));
+        throw new IllegalArgumentException(String.format("Don't know how to compile from a spec of type %s.", spec.getSimpleName()));
```
mlopatkin, line 160: 🎉 Nice

platforms/software/ivy/src/main/java/org/gradle/api/publish/ivy/internal/tasks/IvyDescriptorFileGenerator.java
```diff
@@ -309,11 +308,11 @@ public OptionalAttributeXmlWriter comment(String comment) throws IOException {
     public static void insertGradleMetadataMarker(XmlProvider xmlProvider) {
         String comment = Joiner.on("").join(
```
mlopatkin, line 309: 🤔 Not sure why Guava is necessary here, but that's out of scope

mlopatkin (approved):

Pankraz76:
this seems to be an obsolete leftover now, not referring to this anymore, right?

Goal was to apply convention principle, which is imho, well done, achieved.

- https://github.com/gradle/gradle/blob/1468e41e448ec0c31f6f5b6999825da00e22c931/build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.code-quality.gradle.kts#L132

Pankraz76:
This would be one way to reduce manual effort in the contribution workflow and align with the initiative started in:

- https://github.com/gradle/gradle/pull/35878

Maybe the previous decision against adopting the Picnic extension and Spotless integration could be reconsidered, as this would improve overall robustness and reinforce the convention principle—similar to what has been assumed in the JUnit integration efforts:

- https://github.com/junit-team/junit-framework/pull/5193
- https://github.com/junit-team/junit-framework/pull/5191

Issue #35584, opened 2025-11-06
Add IP Gradleception tests to synthetically schedule all tasks


jvandort:
### Expected Behavior

There are Gradleception tests with IP enabled that automatically schedules all tasks.

The goal here would be to verify all task inputs are seen by CC, and are therefore realized during serialization causing IP violations if they are present. 

### Context

At a minimum, a test that does something like:
1. Checks out the gradle/gradle build
2. Adds the following (or equivalent) to settings.gradle `gradleLifecycle.afterProject { tasks.named("build") { dependsOn(tasks) } `, and any other included builds settings files.
3. Verifies `build` succeeds `--dry-run` without IP violations

PR #35867, opened 2025-11-28
Add IP Gradleception tests to synthetically schedule all tasks

abstratt:
Fixes: 

#35584

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/schedule-all-tasks.gradle
+++ b/testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/schedule-all-tasks.gradle
@@ -49,8 +49,8 @@ def excludedTaskTypes = [
 
 gradle.lifecycle.afterProject { Project p ->
     p.tasks.register(SCHEDULE_ALL_TASK).configure { anchor ->
-        anchor.dependsOn tasks.matching { t ->
-            !excludedTaskTypes.any { it.isAssignableFrom(toVeto.class) } && !excludedTaskNames.contains(toVeto.name)
+        anchor.dependsOn p.tasks.matching { t ->
+            !excludedTaskTypes.any { it.isAssignableFrom(t.class) } && !excludedTaskNames.contains(t.name)
         }
     }
 }


abstratt:
@bot-gradle test PRF

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
```diff
@@ -49,4 +49,49 @@ class GradleBuildIsolatedProjectsSmokeTest extends AbstractGradleBuildIsolatedPr
...
         }
     }
+
+    def "can realize all Gradle build tasks with isolated projects enabled"() {
+        def realizeAllTasksScript = "realize-all-tasks.gradle"
+
+        File realizeAllTasksScriptFile = new File(testProjectDir, realizeAllTasksScript)
+        realizeAllTasksScriptFile << getClass().getResource(realizeAllTasksScript).text
+        def fixture = new ConfigurationCacheProblemsFixture(testProjectDir)
+        println(testProjectDir.absolutePath)
```
6hundreds: 🤔 I think it's a debug leftover?

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
```diff
@@ -49,4 +49,49 @@ class GradleBuildIsolatedProjectsSmokeTest extends AbstractGradleBuildIsolatedPr
...
+        def tasks = [
+            "--init-script",
+            realizeAllTasksScriptFile.absolutePath,
+            "-DscheduleTasks=true",
+            // see https://github.com/gradle/gradle-org-conventions-plugin/blob/185ed5cd4923c061a1c70d77c27758df4c80c6d9/src/main/java/io/github/gradle/conventions/customvalueprovider/GitInformationCustomValueProvider.java#L24
+            "--no-scan"
+        ] + requiredGradleProperties
+
+        expect:
+        testProjectDir.directory
```
6hundreds: 🤔 What the assert does it correspond to? `testProjectDir.directory != null` ?

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/realize-all-tasks.gradle
```diff
@@ -0,0 +1,237 @@
...
+private void log(String message, Closure lazyValue, Consumer<String> consumer) {
+    consumer.accept("*** " + message)
+    lazyValue?.tap {
+        consumer.accept("*** " + it.call())
+    }
+}
+
+class Filters {
+    def isTaskAccepted(TaskProvider task) {
+        if (shouldExcludeTask(task.name, ((ProviderInternal) task).type)) {
```
6hundreds: 💅 return `!shouldExcludeTask(task.name, ((ProviderInternal) task).type)`

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/realize-all-tasks.gradle
```diff
@@ -0,0 +1,237 @@
...
+        projectAnchor.dependsOn(filteredTasks)
+    }
+    taskGraph.whenReady { graph ->
+        def allTasks = graph.allTasks
+        info("Disabling task execution on ${allTasks.size()} tasks")
+        allTasks.each {
+            debug("Disabling ${it.path}")
+            it.setEnabled(false)
+        }
+    }
```
6hundreds: 🤔 Shouldn't this be controlled by `--dry-run` flag?

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/realize-all-tasks.gradle
```diff
@@ -0,0 +1,237 @@
...
+    debug("Realized in $eachProject: ${realizedInProject.size()}")
+    realizedInProject.each { projectTask ->
+        debug("Realized: ${projectTask.name} - ${projectTask.class.simpleName}")
+        realizedTasks << projectTask
+    }
+}
+
+gradle.projectsEvaluated { Gradle gradle ->
+    if (!scheduleTasks) {
```
6hundreds: 🤔 I see we always pass `-DscheduleTasks=true` in the test, do we want to test it with `false` also? If not, do we need this flag at all?

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/realize-all-tasks.gradle
```diff
@@ -0,0 +1,237 @@
...
+
+    info("Looking for tasks to schedule - considering ${allProjects.size()} projects")
+    // Build the list of requested Task objects (from names)
+    def requestedTasks= toRequestedTasks(gradle.rootProject, requestedTaskNames)
+    debug("Resolved requested tasks: ${requestedTasks.size()} of ${requestedTasks}")
+    // Build the list of requested Task objects (from names)
+    List<String> requestedTaskPaths = requestedTasks*.path
+
+    // create a root anchor task
+    def rootAnchor = gradle.rootProject.tasks.register("__anchor").get()
```
6hundreds: 🤔 Do you want to use `ANCHOR` constant?

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/realize-all-tasks.gradle
```diff
@@ -0,0 +1,237 @@
...
+            return
+        }
+        def projectAnchor= project.tasks.register(ANCHOR, {
+            if (anyRequests) {
+                rootAnchor.dependsOn(it)
+            }
+        }).get()
+        debug("Registering ${projectAnchor.path} under ${project.path} on ${System.identityHashCode(anchorsPerProject)}")
+        anchorsPerProject[project.path] = projectAnchor
+    }
```
6hundreds: 💭 Don't feel strong, but maybe the better place for this logic is `beforeProject` callback, not `projectsEvaluated`?

6hundreds (commented):
Left some comments, but honestly it's not 100% clear to me that the content of `realize-all-tasks.gradle` is completely justified

bamboo:
> not 100% clear to me that the content of `realize-all-tasks.gradle` is completely justified

Can you expand on that, @6hundreds?

6hundreds:
> Can you expand on that, @6hundreds?

A lot of configuration is there, I'd like to be confident that all the things are required for the goal we aim.

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
abstratt: `File.isDirectory()` checks for the presence of a corresponding file system entry of the directory type.  But this was a leftover from a debugging session, will remove.

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/realize-all-tasks.gradle
abstratt: This script is useful in standalone way, and one may want to exercise realization but not scheduling (which is more costly - and comprehensive). I found it useful to be able to only check the impact of realization without scheduling.

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/realize-all-tasks.gradle
abstratt: I looked into it. There are some data dependencies that make things quite more complicated than if we do it altogether from here.

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/realize-all-tasks.gradle
abstratt: `--dry-run` is recommended to make things faster as it avoids going through execution, but this ensures we never try to execute tasks if the user forgot to specify `--dry-run`. I wish there was a way of enabling dry-run from an init script.

abstratt:
> > Can you expand on that, @6hundreds?
> 
> A lot of configuration is there, I'd like to be confident that all the things are required for the goal we aim.


@6hundreds  Do you have any specific points you think are problematic?

6hundreds:
> @6hundreds Do you have any specific points you think are problematic?

I don't. It's surprising how much need to be done just to schedule all tasks

6hundreds (approved):
LGTM!

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/schedule-all-tasks.gradle
```diff
@@ -0,0 +1,92 @@
...
+    ]
+)
+
+Task rootAnchor = null
+gradle.afterProject { Project p ->
+    def created = p.tasks.register(SCHEDULE_ALL_TASK)
+    created.configure { anchor ->
+        if (p.rootProject != p) {
+            assert rootAnchor
+            rootAnchor.dependsOn(anchor)
```
jvandort: Why do we need to have the root anchor depend on the subproject's anchors?

When we invoke gradle, we include the `scheduleAll` CLI arg, which should request the `scheduleAll` task from all projects. 

I find it suspicious that we are calling `dependsOn` using task instances from different projects like this. I'd like to see us get to a point where this test can be implemented with `gradle.lifecycle` instead of the legacy mechanism.

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/schedule-all-tasks.gradle
```diff
@@ -0,0 +1,92 @@
...
+        }
+        anchor.dependsOn(dependencies)
+    }
+    if (p.rootProject == p) {
+        rootAnchor = created.get()
+    }
+}
+
+class Filters {
+    List<String> excludedTaskNames = []
```
jvandort: I'm not sure I understand the difference between `excludedTaskNames` and `excludedTaskDependencies`. 

IIUC If you simplify all the filtering logic here you get this:
```
!excludedTaskTypes.any { it.isAssignableFrom(t.class) } && !excludedTaskNames.contains(t.name) && !excludedTaskDependencies.contains(t.name)
```
and these two sets become functionally identical. I'm not sure I see the point of having different sets, and IMO this whole `Filters` class hides the logic thats actually happening here

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
```diff
@@ -49,4 +49,47 @@ class GradleBuildIsolatedProjectsSmokeTest extends AbstractGradleBuildIsolatedPr
             )
         }
     }
+
+    def "can schedule all Gradle build tasks with isolated projects enabled"() {
+        def scheduleAllTasksScript = "schedule-all-tasks.gradle"
+
+        File scheduleAllTasksScriptFile = new File(testProjectDir, scheduleAllTasksScript)
+        scheduleAllTasksScriptFile << getClass().getResource(scheduleAllTasksScript).text
```
jvandort, line 57: How come we are defining this as a resource instead of inline? 
IMO its pretty standard to define scripts inside the tests themselves. 
This pattern adds to indirection and then makes my have to go search for the init script somewhere else

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
```diff
@@ -49,4 +49,47 @@ class GradleBuildIsolatedProjectsSmokeTest extends AbstractGradleBuildIsolatedPr
...
+        maxIsolatedProjectProblems = 200000
+        run(isolatedProjectsRunner(tasks).withEnvironment(requiredEnvironmentVars))
+
+        then:
+        fixture.assertHtmlReportHasProblems(result.output) {
+            withUniqueProblems(
+                "Project : cannot access Project.plugins functionality on subprojects via allprojects",
+                "Project : cannot access Project.extensions functionality on subprojects via allprojects",
+            )
+            // maximum number of problems we collect (should be 86520)
```
jvandort, line 91: Am I reading this correctly that there are `86520` IP violations emitted?

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/schedule-all-tasks.gradle
abstratt: Yeah, I was thinking about that as well. Will remove that edge. Will use `gradle.lifecycle` as well.

testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/schedule-all-tasks.gradle
abstratt: `excludedTaskDependencies` can indeed be merged with `excludedTaskNames`. Will do that. Will inline the filtering as well. They were a bit more complicated before, the explicit modeling made more sense than now.

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
abstratt: It is pretty standard, but it is handy to be able to use this script outside of the test (we may end up evolving it to a Gradle feature btw). Testing it by running it directly (as opposed to via the smoke test) is much easier as well. 

We have done this for at least one other smoke test, btw: https://github.com/gradle/gradle/blob/66f77eb97d8ff79058e97f9679d224c46f12b8c7/testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/validate-external-gradle-plugin.gradle.kts

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
abstratt: Yeah... it is only 2 problems, probably 43260 instances each.

abstratt:
@bot-gradle test PRF

Issue #35843, opened 2025-11-27
Implement equals and hashCode for Identity type


asodja:
Execution Work [Identity](https://github.com/gradle/gradle/blob/29b0442cd98e3d36a778d0e05ea0b85bea59c769/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/Identity.java#L19) is used as a key for cache in some cases:
https://github.com/gradle/gradle/blob/034bf8134a2ce7bd13b8403bc6f4cca39ed792ec/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/transform/MutableTransformWorkspaceServices.java#L31

But not all Identity implementations implement `hashCode` and `equals`. That is currently not an issue, since we use identity as a cache key just for transforms which identities has `hashCode` and `equals` implemented.  But it could still be error prone.

Let's implement equals and hashcode for Identity implementations.

bamboo:
🤔 I wonder if the unique id returned by `Identity#getUniqueId()` wouldn't be a suitable cache key.

Pankraz76:
expected this to be covered by prone already, as not finding suppression for this check.

- https://errorprone.info/bugpattern/EqualsHashCode

PR #35879, opened 2025-12-01
Introduce specialized IdentityCache for DeferredResults

asodja:
To avoid having issues if Identity doesn't implement hashCode/equals.

Fixes https://github.com/gradle/gradle/issues/35843

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/execution/DefaultWorkValidationWarningRecorder.java
+++ b/subprojects/core/src/main/java/org/gradle/execution/DefaultWorkValidationWarningRecorder.java
@@ -37,11 +37,11 @@ public class DefaultWorkValidationWarningRecorder implements ValidateStep.Valida
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultWorkValidationWarningRecorder.class);
 
     // TODO If we can ensure that the recorder is only called once per work execution, then we can demote this to a simple counter
-    private final Set<Identity> workWithWarnings = ConcurrentHashMap.newKeySet();
+    private final Set<String> workIdsWithWarnings = ConcurrentHashMap.newKeySet();
 
     @Override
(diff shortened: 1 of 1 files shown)


asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/IdentityCache.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package org.gradle.internal.execution;
+
+import org.jspecify.annotations.Nullable;
+
+import java.util.function.Supplier;
+
+public interface IdentityCache<T> {
```
asodja, line 23: 💭 Maybe a better name would be: DeferredWorkCache

asodja:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/execution/DefaultWorkValidationWarningRecorder.java
```diff
@@ -37,11 +37,11 @@ public class DefaultWorkValidationWarningRecorder implements ValidateStep.Valida
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultWorkValidationWarningRecorder.class);
 
     // TODO If we can ensure that the recorder is only called once per work execution, then we can demote this to a simple counter
-    private final Set<Identity> workWithWarnings = ConcurrentHashMap.newKeySet();
+    private final Set<String> workIdsWithWarnings = ConcurrentHashMap.newKeySet();
```
asodja, line 40: 💭 Since we also have to fix this case, maybe we should just make `Identity` abstract class, so we can force implementing equals and hashcode.

PR #35893, opened 2025-12-01
Move CC build operation helpers to own package

alllex:
It was the only file in `:configuration-cache` that located outside of `org.gradle.internal.cc` package.


bamboo:
Yes, there was some concern that the type names might have been depended on by some DV logic.

alllex:
I've [confirmed](https://gradle.slack.com/archives/C03PH5M446L/p1768382860082259) with the Scans team that there are currently no references to this package in the DV code, and that Gradle's cross-version tests should have enough coverage. If the tests are successful, we can proceed with the merge.

alllex:

alllex:

PR #35902, opened 2025-12-02
Split up `BuildTreeModelControllerServices`

alllex:
Splits up `BuildTreeModelControllerServices` implementation into multiple more focused pieces.

That service existed to implement the lifecycle step of running the build action over a build tree. This can happen multiple times in a build session, so every build session creates a fresh build tree, and the build action is getting re-executed.

The main change of this PR is to separate the concerns previously combined in a single service.

There is now a new service `BuildModelParametersFactory`, which is solely dedicated to computing the build model parameters based on the build action requirements (those include build options to enable features and the state of the `StartParameter` in general).

Additionally, a custom indirection in the form of `BuildTreeModelControllerServices.Supplier` has been replace with a more explicit build model parameter computation and a corresponding service registration provider that is able to register the needed services in a usual manner, while still taking the parameters into account.

---

There are still more complications with the current implementation. The new version of `BuildTreeModelControllerServices` still lives in `:configuration-cache`. The module is supposed to contain the _implementation_ of Configuration Cache, but it also carries the responsibility of registering Vintage services. There is a number of classes/services in that module that don't belong here, but don't have a better place to go yet.

In addition, the previous state of things made it difficult to extract services and implementations related to Isolated Projects. Some of them are currently mechanically coupled to Configuration Cache implementation, even though there is no fundamental reason for it.

In general, the motivation of this PR is to prepare the codebase for easier decoupling of Configuration Cache and Isolated Projects.


alllex:

PR #35926, opened 2025-12-03
Add methods for declaring bindings with unsafe definitions

ghale:
Allows project features to be bound declaring the definition as "unsafe".  If not declared unsafe, the assumption is that a definitions is safe and we'll throw an exception if we find that it isn't.  

Safe means:
- Implemented as an interface
- Does not have a separate implementation type
- Does not have injected services or contain nested injected services

Registering a definition implementation type implicitly declares the definition as unsafe.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/software/internal/DefaultProjectFeatureDeclarations.java
+++ b/platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/software/internal/DefaultProjectFeatureDeclarations.java
@@ -48,6 +48,7 @@ import org.jspecify.annotations.Nullable;
 
 import javax.inject.Inject;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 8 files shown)


ghale:
@bot-gradle test this

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/software/internal/DefaultProjectFeatureDeclarations.java
```diff
@@ -151,6 +169,69 @@ private void registerTypeIfPresent(RegisteringPluginKey registeringPluginKey, Cl
         }
     }
 
+    private void validateDefinitionSafety(ProjectFeatureBindingDeclaration<?, ?> binding) {
+        List<InternalProblem> problems = new ArrayList<>();
+        if (binding.getDefinitionImplementationType().isPresent() && !binding.getDefinitionImplementationType().get().equals(binding.getDefinitionType())) {
+            problems.add(problemReporter.internalCreate(builder -> builder
+                .id("unsafe-definition-implementation-type", "Definition implementation type specified for safe definition", GradleCoreProblemGroup.configurationUsage())
+                .details("Project feature '" + binding.getName() + "' has a definition with type '" + binding.getDefinitionType().getSimpleName() + "' which was declared safe but has an implementation type '" + binding.getDefinitionImplementationType().get().getSimpleName() + "'.  " +
+                    "Safe definitions must not specify an implementation type.")
```
eskatos: 💅 This should use a problem `contextualLabel`. Other problem creations below should do the same.

```suggestion
                .details("Safe definitions must not specify an implementation type.")
                .contextualLabel("Project feature '" + binding.getName() + "' has a definition with type '" + binding.getDefinitionType().getSimpleName() + "' which was declared safe but has an implementation type '" + binding.getDefinitionImplementationType().get().getSimpleName() + "'")
```

ghale: Yep, much better now.

PR #35931, opened 2025-12-04
Isolate TAPI tests validating resource-based test execution

donat:
The cross-version tests are not working well with `TestResources` implemented for integration tests, causing flaky test executions, especially on Windows. The root cause is the custom test engine running the cross version tests: it it runs the tests in the context of different class loaders, multiple times (with the same name). This PR embeds the sources the test engine into the base class to make the test stable.

The PR also simplifies the cross-version tests as the two test engine variants are merged into one.

Example failure: https://ge.gradle.org/s/yqea7dydnknku

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


donat:
@bot-gradle test ACT

jbartok (approved):

PR #35915, opened 2025-12-03
Reduce the need for custom services for ProjectBuilder

alllex:
The implementation of `ProjectBuilder` today requires an injection point for custom Build-scoped services. This makes reasoning about and refactoring the services more complicated.
https://github.com/gradle/gradle/blob/master/platforms/extensibility/unit-test-fixtures/src/main/java/org/gradle/testfixtures/internal/ProjectBuilderImpl.java#L285

This PR is a step towards removing the need for that injection point.

PR #35933, opened 2025-12-04
Use immutable project descriptors after settings

alllex:
This is a follow-up for

*   [Reduce the need for custom services for ProjectBuilder #35915](https://github.com/gradle/gradle/pull/35915)

### Context

After the settings phase, the project hierarchy is fully defined — yet the mutable `ProjectDescriptor` continues to be passed around through `ProjectState` and into project creation. Downstream code must reason about mutation contracts that no longer apply, and the `ProjectDescriptorRegistry` (a settings-scope service) leaks into build-scope concerns like `ProjectBuilder`.

The goal is to reduce complexity, simplify reasoning about mutable state contracts, and reduce coupling between the common project-registration path and the special-cased `ProjectBuilder`. This PR doesn't fully achieve it, but once we resolve the build file fingerprinting issue (build script resolution currently gets attributed to build-scoped CC fingerprint too early), the mutable descriptor registry can be discarded entirely after the settings phase.

### Approach

Introduce `ImmutableProjectDescriptor` — a read-only snapshot of a project's identity, directory, build file, and children — and thread it through `ProjectState` and project creation in place of the mutable `ProjectDescriptor`.

The mutable registry remains the source of truth during settings evaluation. At the transition into the configuration phase, each mutable descriptor is converted to an immutable one and handed to `ProjectState`. From that point on, nothing downstream needs access to the mutable registry.

For `ProjectBuilder`, which has no settings phase and creates projects incrementally, a dedicated `ProjectBuilderProjectDescriptor` allows late addition of children while still presenting an immutable-style API.

### Changes

*   Introduce `ImmutableProjectDescriptor` and thread it through `ProjectState`, project registration, and project creation
*   Add `ProjectBuilderProjectDescriptor` for `ProjectBuilder`'s incremental lifecycle
*   Refactor `ProjectBuilderImpl` to construct immutable descriptors directly, removing its dependency on `ProjectDescriptorRegistry`
*   Small cleanups: make `ProjectDescriptorRegistry` field final in `DefaultProjectDescriptor`, fix nullability on `ProjectDescriptorInternal.getParent()`, extract `DEFAULT_BUILD_FILE` constant to `BuildLogicFiles`


alllex:

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
```diff
@@ -284,7 +317,7 @@ public ProjectInternal getMutableModel(ProjectState project) {
     @NullMarked
     private class ProjectStateImpl implements ProjectState, Closeable {
 
-        private final ProjectDescriptorInternal descriptor;
+        private final ImmutableProjectDescriptor descriptor;
```
alllex, line 321: ℹ️  This change represents the main goal of the PR

subprojects/core/src/main/java/org/gradle/initialization/DefaultProjectDescriptor.java
```diff
@@ -41,6 +40,7 @@ public class DefaultProjectDescriptor implements ProjectDescriptorInternal {
     public static final String INVALID_NAME_IN_INCLUDE_HINT = "Set the 'rootProject.name' or adjust the 'include' statement (see "
         + new DocumentationRegistry().getDslRefForProperty(Settings.class, "include(java.lang.String[])") + " for more details).";
 
+    @Deprecated // Do not immediately remove, since it's an old API that could have been used despite being internal
```
jvandort: Do we have any evidence that this is being used? I don't usually deprecate internals for removal unless we know plugins are actually using it

platforms/core-configuration/kotlin-dsl-provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/precompiled/tasks/GeneratePrecompiledScriptPluginAccessors.kt
```diff
@@ -467,6 +471,14 @@ abstract class GeneratePrecompiledScriptPluginAccessors @Inject internal constru
         }
     }
 
+    private fun descriptorForRoot(settings: SettingsInternal): ImmutableProjectDescriptor {
+        val descriptor = settings.rootProject as ProjectDescriptorInternal
```
jvandort, line 475: Can we override `getRootProject` on `SettingsInternal` to return `ProjectDescriptorInternal` so we don't need to cast here?

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
```diff
@@ -110,26 +144,26 @@ public void discardProjectsFor(BuildState build) {
...
-        projectsByPath.put(identity.getBuildTreePath(), projectState);
+        ProjectStateImpl projectState = new ProjectStateImpl(buildState, descriptor, projectFactory, stateTransitionControllerFactory, buildServices);
+        projectsByPath.put(descriptor.getIdentity().getBuildTreePath(), projectState);
         projectsById.put(projectState.getComponentIdentifier(), projectState);
-        projectRegistry.add(projectPath, projectState);
+        projectRegistry.add(descriptor.getIdentity().getProjectPath(), projectState);
         return projectState;
     }
 
+    private static ProjectIdentity projectIdentityFor(Path buildPath, ProjectDescriptorInternal descriptor) {
```
jvandort: Nitpick, but if we're pulling out this logic into a separate method IMO it should live closer to the method it is used in

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
```diff
@@ -68,18 +73,47 @@ public DefaultProjectStateRegistry(WorkerLeaseService workerLeaseService) {
...
+    private ProjectIdentity registerAllProjectsRecursively(
+        BuildState buildState,
+        DefaultBuildProjectRegistry projectRegistry,
+        @Nullable ProjectIdentity parentIdentity,
+        ProjectDescriptorInternal descriptor
+    ) {
+        Path buildPath = buildState.getIdentityPath();
+        ProjectIdentity identity = projectIdentityFor(buildPath, descriptor);
+
+        ArrayList<ProjectIdentity> children = new ArrayList<>();
```
jvandort:
```suggestion
        List<ProjectIdentity> children = new ArrayList<>();
```

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
```diff
@@ -68,18 +73,47 @@ public DefaultProjectStateRegistry(WorkerLeaseService workerLeaseService) {
...
+            ProjectIdentity childIdentity = registerAllProjectsRecursively(buildState, projectRegistry, identity, childDescriptor);
+            children.add(childIdentity);
+        }
+
+        children.sort(Comparator.comparing(ProjectIdentity::getBuildTreePath));
+
+        // TODO:isolated avoid preserving a reference to the original mutable descriptor
+        // We avoid getting the build file early here, because otherwise it will be registered as a build-scoped configuration input,
+        // which makes the project invalidation less-fine grained. See `ProjectScopedScriptResolution.resolveScriptsForProject(...)`
+        Supplier<File> buildFileSupplier = Lazy.unsafe().of(descriptor::getBuildFile);
```
jvandort: These descriptors are accessible from `ProjectState`, meaning you do not need a lock to access them. I would say we should make this a locking lazy

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
```diff
@@ -68,18 +73,47 @@ public DefaultProjectStateRegistry(WorkerLeaseService workerLeaseService) {
 
     @Override
     public void registerProjects(BuildState owner, ProjectDescriptorRegistry projectRegistry) {
```
jvandort, line 76: I think this is fine for now (unless you want to address it here), but I don't think `DefaultProjectStateRegistry` should "know" about the mutable versions of the project descriptors at all. I think this registry should accept _only_ immutable representations. If some source decided to construct a project hierarchy immutably, this registry should accept that. IMO the mutable project descriptor (and project descriptor registry) are implementation details of the  `Settings` DSL

jvandort (approved):
LGTM. 

I wonder if we should eventually deprecate allowing you to use subprojects in `ProjectBuilder`, and what use cases child projects with PB fills today.

subprojects/core/src/main/java/org/gradle/initialization/DefaultProjectDescriptor.java
alllex: Not that I know of. I got burned with something similar before, so I wanted to avoid a similar hustle. Though I can't remember exactly what that was. There is `SettingsInternal.BUILD_SRC`, but it's not deprecated.

This field seems more obscure, so I agree that we should probably just remove it, and only consider keeping it longer if it breaks something more important.

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
alllex: > meaning you do not need a lock to access them. I would say we should make this a locking lazy

This seems like a contradiction. Can you expand?

jvandort: Currently you can access the new `ProjectState#getDescriptor` method from a `ProjectState` without a project lock.

You can then access the build file from the descriptor without a project lock.

Therefore, anything exposed on ProjectState should be thread-safe, so this _should_ be a locking Lazy

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
alllex: I generally agree. I think I even tried it, but the implementation got tangled. Though I don't recall why

I'd rather we get the current refactoring merged, and then proceed with further simplifications.

platforms/core-configuration/kotlin-dsl-provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/precompiled/tasks/GeneratePrecompiledScriptPluginAccessors.kt
alllex: I tried that, and for some reason hell breaks loose in many Declarative tests 🤷‍♂️
Let's figure it out separately

- https://github.com/gradle/gradle/pull/37734

PR #35940, opened 2025-12-05
Split build model parameters into modes

alllex:
to make it easier to reason about how specific behaviors are enabled/disabled for each mode.

Modes:
- Vintage
- CC (only for tasks)
- IP

This will also make it easier to implement upcoming flags for additional IP modes and the explicit Parallel Tooling Actions flag.

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParameters.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParameters.kt
@@ -22,19 +22,19 @@ import org.gradle.internal.buildtree.BuildModelParameters
 internal sealed class AbstractBuildModelParameters : BuildModelParameters {
 
     override fun toDisplayMap(): Map<String, Any?> = mapOf(
-        "requiresToolingModels" to isRequiresToolingModels,
-        "parallelProjectExecution" to isParallelProjectExecution,
-        "configureOnDemand" to isConfigureOnDemand,
         "configurationCache" to isConfigurationCache,
         "configurationCacheDisabledReason" to configurationCacheDisabledReason,
-        "configurationCacheParallelStore" to isConfigurationCacheParallelStore,
         "configurationCacheParallelLoad" to isConfigurationCacheParallelLoad,
-        "isolatedProjects" to isIsolatedProjects,
-        "parallelProjectConfiguration" to isParallelProjectExecution,
+        "configurationCacheParallelStore" to isConfigurationCacheParallelStore,
+        "configureOnDemand" to isConfigureOnDemand,
         "intermediateModelCache" to isIntermediateModelCache,
-        "parallelToolingApiActions" to isParallelToolingApiActions,
         "invalidateCoupledProjects" to isInvalidateCoupledProjects,
+        "isolatedProjects" to isIsolatedProjects,
         "modelAsProjectDependency" to isModelAsProjectDependency,
+        "parallelProjectConfiguration" to isParallelProjectExecution,
+        "parallelProjectExecution" to isParallelProjectExecution,
+        "parallelToolingApiActions" to isParallelToolingApiActions,
+        "requiresToolingModels" to isRequiresToolingModels,
         "resilientModelBuilding" to isResilientModelBuilding,
     )
 }


alllex:

alllex:

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
```diff
@@ -0,0 +1,255 @@
...
+        )
+    }
+
+    private
+    fun vintageMode(
+        requirements: BuildActionModelRequirements,
+        startParameter: StartParameterInternal,
+        options: InternalOptions,
+        ccDisabledReason: String? = null
+    ): GradleVintageMode {
```
abstratt, line 133: 💅 

I'd go with the interface type here, same for the other methods for CC and IP. AFAIU, The mode classes are just an internal implementation decision on how to implement `BuildModelParameters`

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    private
+    fun warnOnPreviouslyExistingOptions(options: InternalOptions) {
+        val replacements = mapOf(
+            "org.gradle.internal.isolated-projects.configure-on-demand.tooling" to isolatedProjectsConfigureOnDemand.propertyName,
+            "org.gradle.internal.isolated-projects.configure-on-demand.tasks" to isolatedProjectsConfigureOnDemand.propertyName,
+        )
+        for ((previous, current) in replacements) {
+            if (options.getOption(StringInternalOption.of(previous)).isExplicit) {
```
abstratt, line 229: 💅 

The fact you need to create (potentially mistyped) option to check for the presence of an explicit property value suggests `InternalOptions`  could have a `@Nullable String getExplicitPropertyValue(String propertyName)` method.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
```diff
@@ -0,0 +1,255 @@
+/*
```
abstratt, line 1: ❓ 

@alllex Any changes here, or just a module/package move?

alllex: It's a move, but the implementation of the `parameters()` method has been refactored significantly. You can also go commit-by-commit, since I put the moves between packages in separate commits compared to code changes.

abstratt: Actually, the instantiation of the model parameters changed to use the sealed class hierarchy. Anything else?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParameters.kt
```diff
@@ -0,0 +1,140 @@
...
+
+package org.gradle.internal.buildtree.control
+
+import org.gradle.internal.buildtree.BuildModelParameters
+
+
+internal sealed class AbstractBuildModelParameters : BuildModelParameters {
+
+    override fun toDisplayMap(): Map<String, Any?> = mapOf(
+        "requiresToolingModels" to isRequiresToolingModels,
```
abstratt: 💅 If order does not matter, I'd go with alphabetical order here

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
alllex: No that's all, only the change in the method implementation

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
```diff
@@ -0,0 +1,255 @@
...
+        val options = DefaultInternalOptions(startParameter.systemPropertiesArgs)
+        warnOnPreviouslyExistingOptions(options)
+
+        val ccDisabledReason = getConfigurationCacheDisabledReason(startParameter)
+        return when {
+            // TODO:isolated should this also disable IP?
+            ccDisabledReason != null -> vintageMode(requirements, startParameter, options, ccDisabledReason)
+            startParameter.isolatedProjects.get() -> isolatedProjectsMode(requirements, startParameter, options)
+            startParameter.configurationCache.get() ->
+                if (requirements.isCreatesModel)
```
abstratt: 💭 

It would be nice to see an explanation why `creates-model` disables CC, like:

`// model creation is not supported by CC yet`

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParameters.kt
alllex: The order does not matter. The properties are somewhat grouped, but I guess the grouping is too loose anyway. Will change to alphabetical.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
alllex: I think it's fine overall, since these methods are private. However, I intentionally put the implementation types as the return type to ensure we don't accidentally lose invariants. That is because now some assumptions are backed into the sealed hierarchy itself (some of the methods return constants). That's why I wanted to have the methods here correspond 1-to-1 to the implementations, and the routing to happen only at the top level, where we decide on the mode.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
alllex: This check here is a curtesy to some individual early adopters with whom we shared these internal properties. We'll remove this method in one of the upcoming minors. Because of this I think it's okay to have a hacky solution of looking up an "on demand" internal property.

PR #35941, opened 2025-12-05
Add Daemon log clean up action

reinsch82:
Fixes
* https://github.com/gradle/gradle/issues/35803

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
+++ b/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
@@ -92,7 +92,8 @@ class GradleBuildIsolatedProjectsSmokeTest extends AbstractGradleBuildIsolatedPr
                 "Project ':' cannot access 'Project.extensions' functionality on subprojects via 'allprojects'",
             )
             // TODO-RC if this ends up making this test too brittle, we should use a looser/range-based expectation
-            totalProblemsCount = 87354
+            // This value needs to be increased whenever a new project is added to gradle/gradle
+            totalProblemsCount = 88192
         }
         result.assertNoConfigurationCache()
     }


reinsch82:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/cache/internal/GradleUserHomeCleanupService.java
```diff
@@ -83,16 +87,16 @@ public void stop() {
     }
 
     private boolean execute(MonitoredCleanupAction action) {
-        return buildOperationRunner.call(new CallableBuildOperation<Boolean>() {
+        return Boolean.TRUE.equals(buildOperationRunner.call(new CallableBuildOperation<Boolean>() {
```
jbartok, line 91: This equality check looks unnecessary.

subprojects/core/src/main/java/org/gradle/cache/internal/DaemonLogCleanupAction.java
```diff
@@ -0,0 +1,113 @@
...
+        if (logFiles == null) {
+            LOGGER.warn("Could not list log files for cleanup in: {}", daemonLogDirectory.getAbsolutePath());
+            return;
+        }
+
+        for (File logFile : logFiles) {
+            if (logFile.lastModified() < maxAge) {
+                try {
+                    LOGGER.debug("Deleting old daemon log file at {}", logFile);
+                    deleter.deleteRecursively(logFile);
```
jbartok: Why do we "delete recursively"? Based on the filter above, it's a file, not a directory.

subprojects/core/src/main/java/org/gradle/cache/internal/DaemonLogCleanupAction.java
```diff
@@ -0,0 +1,113 @@
...
+
+        if (logFiles == null) {
+            LOGGER.warn("Could not list log files for cleanup in: {}", daemonLogDirectory.getAbsolutePath());
+            return;
+        }
+
+        for (File logFile : logFiles) {
+            if (logFile.lastModified() < maxAge) {
+                try {
+                    LOGGER.debug("Deleting old daemon log file at {}", logFile);
```
jbartok: I would say remove the "at", to be consistent with the warning message below. It's a file, not a directory, anyways, right?

subprojects/core/src/main/java/org/gradle/cache/internal/GradleUserHomeCleanupService.java
reinsch82: Since `call` returns a `Nullable` this guarantees that unboxing doesn't throw an NPE.
It's mostly compiler appeasement though.

platforms/documentation/docs/src/snippets/bestPractices/useBuildCache-do/tests/useBuildCache-do.firstRun.out
```diff
@@ -2,3 +2,4 @@
 
 BUILD SUCCESSFUL in 0s
 1 actionable task: 1 executed
+Could not list daemon log directories for cleanup in: /home/user/gradle/samples/user-home/daemon
```
eskatos: ❓ Could you explain why we have to add this here?

This might make running the samples locally problematic.

reinsch82: I could run the sample locally, I'm not aware of a scenario where it doesn't work yet.
But I don't exactly know why it shows up.
This implies that in this scenario we don't create any daemon logs, I have no idea why that would be.

eskatos: If we emit this warning in the case where there's simply no log, then it's something we should fix.
There's no point in warning the user because there's no daemon log.

blindpirate:
This is failing on branch: https://github.com/gradle/gradle/pull/35941

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

PR #35947, opened 2025-12-05
Fail-fast if improper filter type is configured in test task

tresat:
The filter format for class-based and non-class-based tests is different.

Configuring a filter using the format for a type of test that is not present is never correct.  In this case, Gradle will now fail-fast the `Test` task, rather than silently doing nothing and producing probably undesired test executions.


platforms/jvm/testing-jvm/src/test/groovy/org/gradle/api/tasks/testing/TestTaskSpec.groovy
```diff
@@ -318,6 +319,11 @@ class TestTaskSpec extends AbstractProjectBuilderSpec {
             it.includePatterns = "Foo"
         }
 
+        def testDir = temporaryFolder.createDir("testClasses")
+        testDir.mkdir()
+        new File(testDir, "Foo.java").createNewFile()
+        task.setTestClassesDirs(TestFiles.fixed(testDir))
```
tresat, line 325: Necessary, otherwise this test falls with the new message.

PR #35948, opened 2025-12-05
Introduce parallel tooling model building option

alllex:
Implements a [spec](https://docs.google.com/document/d/1DgDKp1UyYKRZ5psPIXOCnk55yngWvp6bSHSlIvjBVKM) regarding a better parallel model fetching (PMF) setup in IDEA, in particular, unlocking PMF for Isolated Projects.

This is a required step to fix:
- https://github.com/gradle/gradle/issues/35427

### New `org.gradle.tooling.parallel` option

Previously, PMF has been implicitly enabled with the Parallel Execution (`--parallel` or `org.gradle.parallel`) flag. Users who wanted parallel task execution didn't have a way to disable PMF if they've seen problems with. Vice versa, if parallel task execution didn't work in the build, users didn't have a way to only enable PMF.

The new `org.gradle.tooling.parallel` provides explicit control over PMF. It can be added to the `gradle.properties` file, where it also becomes easily sharable via VCS with other users working on the same project.

### Default of `org.gradle.tooling.parallel`

When no explicit value is provided for `org.gradle.tooling.parallel`, it defaults to the value of `org.gradle.parallel`. This is done to preserve backward-compatibilty for the observable behavior of PMF. Most importantly, this is required because for a long time Android Studio users have been observing PMF by having `org.gradle.parallel` in their builds.

However, that historical implication (and the above default behavior) is what prevents IDEA from safely migrating to delegate PMF decision to a Gradle user and why there is a UI option in IDEA to enable this explicitly.

In order to mitigate this, a new **system property** is introduced: `org.gradle.tooling.parallel.ignore-legacy-default`

That system property is `false` by default, which means it does nothing with respect to the logic above. However, when set to `true`, it removes the historical implication and starts to require users to explicitly enable `org.gradle.tooling.parallel`, if they want to observe PMF.

This allows new IDEA versions to start passing the `ignore-legacy-default` property during sync, making sure that users who have `org.gradle.parallel` already enabled don't get (potentially unsafe) PMF behavior. Due to this, for new Gradle versions that support these new properties, IDEA would finally be able to delegate PMF decision to the user.

### DIFF
--- a/platforms/documentation/docs/src/docs/release/notes.md
+++ b/platforms/documentation/docs/src/docs/release/notes.md
@@ -85,6 +85,20 @@ void main() {
     }
 }
 ```
+
+### New property for Tooling API parallelism control
+
+A new Gradle property `org.gradle.tooling.parallel` allows explicitly controlling whether Tooling API clients can run actions against the build in parallel.
+This is particularly relevant for the IDE Sync scenarios, where IDEs can take advantage of the parallelism to improve performance.
+
+```properties
+# gradle.properties
+org.gradle.tooling.parallel=true
+```
+
+Historically, this was only controlled by the `org.gradle.parallel` property, which is often used to get parallel task execution.
+However, previously it was not possible to enable or disable one without affecting the other.
+
 <!--
 
 ================== TEMPLATE ==============================


platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
```diff
@@ -154,6 +162,19 @@ object BuildModelParametersProvider {
...
+    private
+    fun parallelModelBuildingForVintage(startParameter: StartParameterInternal): Boolean {
+        val parallelModelBuildingOption = startParameter.parallelToolingModelBuilding
+        val parallelModelBuildingIgnoreLegacyDefault =
+            java.lang.Boolean.parseBoolean(startParameter.systemPropertiesArgs[parallelModelBuildingIgnoreLegacyDefault])
+        val parallelModelBuilding = when {
+            parallelModelBuildingOption.isExplicit -> parallelModelBuildingOption.get()
+            parallelModelBuildingIgnoreLegacyDefault -> false
+            else -> startParameter.isParallelProjectExecutionEnabled
+        }
```
alllex, line 174: ℹ️ This is the main change. We now have a more elaborate strategy on deciding when PMF is enabled for vintage.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
```diff
@@ -45,9 +45,6 @@ object BuildModelParametersProvider {
     private
     val configurationCacheParallelLoad = InternalFlag("org.gradle.configuration-cache.internal.parallel-load", true)
 
-    @JvmStatic
-    val parallelBuilding = InternalFlag("org.gradle.internal.tooling.parallel", true)
```
alllex, line 49: ℹ️  The internal property is now replaced with a public one.

FTR, the internal property was `true` by default, because it acted as a guard, rather than an enabler of the behavior. It had the same semantics as the new `ignore-legacy-default` system property has.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
```diff
@@ -74,6 +71,16 @@ object BuildModelParametersProvider {
     val resilientModelBuilding =
         InternalFlag("org.gradle.internal.resilient-model-building", false)
 
+    /**
+     * A public *system property* that allows removing the implication that
+     * `org.gradle.parallel` also controls parallel model building for Vintage.
+     *
+     * It exists as a transitionary measure to allow IDEs to effectively require a separate user opt-in
+     * into parallel model building via the explicit `org.gradle.tooling.parallel` property.
+     */
```
donat, line 80: I assume we don't want to advertise the new property any more than this. So, no mention of it in the user guide is fine.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ParallelActionExecutionCrossVersionSpec.groovy
```diff
@@ -0,0 +1,130 @@
...
+            class CustomModel implements Serializable {
+                String path
+            }
+
+            class CustomBuilder implements ToolingModelBuilder {
+                boolean canBuild(String modelName) { modelName.endsWith("CustomModel") }
+
+                Object buildAll(String modelName, Project project) {
+                    // Do some dependency resolution
+                    project.configurations.runtimeClasspath.files.each { }
```
donat, line 51: 🤔 I probably miss some context here, but do this model builder actually need to do dependency resolution?

alllex: I preserved the structure of the test as it was for `r68` (see the one-line change there to exclude 9.4+ versions)

The resolution here is used to trigger a dependency resolution callback, which in turn calls the `BlockingHttpServer`, through which we validate the parallelism. See here:
https://github.com/gradle/gradle/pull/35948/files#diff-f439321512ba54e29f137f75971a0cd3e559ee8a917938b3d74fc9b6ed9b3895R117-R119

I don't know if there are additional shenanigans for doing the HTTP calls directly inside the model builder. But since it works, and it does not matter much, I opted for keeping the new test class here to be very similar to the one in `r68`.

alllex:
@bot-gradle test BD

PR #35949, opened 2025-12-05
Using JarFile instead of JarInputStream to scan for module information

### LINKED ISSUE #30589: Automatic modules are not detected when MANIFEST.MF is not the 1st or 2nd entry
### Current Behavior

If a jar doesn't have its MANIFEST.MF as the 1st or 2nd entry, then [JavaModuleDetector](https://github.com/gradle/gradle/blob/8716158d3ec8c59e38f87a67f1f311f297b79576/subprojects/core/src/main/java/org/gradle/internal/jvm/JavaModuleDetector.java#L153) will not detect its `Automatic-Module-Name` which in turn will make Gradle believe it is a non-modular jar.

The consequence of this is that you can't use the jar in a modular context.

### Expected Behavior

Gradle should detect the `Automatic-Module-Name` even if the MANIFEST.MF is not the 1st or the 2nd entry.

I believe if the `containsAutomaticModuleName(jarStream)` would be after the loop, then it would solve this issue (though I haven't tested it).

### Context (optional)

_No response_

See <https://github.com/kelemen/gradle-automatic-module-bug>

And run `./gradlew compileJava`

Which will fail with "error: module not found: org.jdbi.v3.core" even though the manifest entry is there in the jar.

### Gradle version

8.10.1

### Build scan URL (optional)

_No response_

_No response_

kelemen:
The current behavior of `JavaModuleDetector` is suboptimal, because it uses `JarInputStream` to scan through the entries in the jar file, which might require decompressing the whole archive for no reason. `JarFile` is more efficient since it just checks the entries. Also, the JDK uses `JarFile` to read jars and not `JarInputStream`, so the change also makes jar scanning consistent with the JDK (even if it is very unlikely to find a jar file where the entries read by `JarFile` are not the same as with `JarInputStream`). Additionally, `JarInputStream` is incorrect, because it can't find the `MANIFEST.MF` if it is not the first or the second entry, despite the fact that the spec doesn't have such constraint on jar files.

This also mostly fixes #30589. Partially, because the fact if a jar is modularized or not is cached, and if it is already cached, then this change has no effect. However, the two methods can only differ for jar files which are currently improperly handled by Gradle. If you want me to address the caching issue by invalidating cache (say, saving the property for a different key), then I would be happy to apply the changes.

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/jvm-services/src/main/java/org/gradle/internal/jvm/JavaModuleDetector.java
+++ b/platforms/jvm/jvm-services/src/main/java/org/gradle/internal/jvm/JavaModuleDetector.java
@@ -26,13 +26,13 @@ import org.gradle.internal.service.scopes.ServiceScope;
 import org.jspecify.annotations.Nullable;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.Collection;
-import java.util.jar.JarInputStream;
+import java.util.Enumeration;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
 import java.util.jar.Manifest;
 import java.util.regex.Pattern;
-import java.util.zip.ZipEntry;
 
 @ServiceScope(Scope.UserHome.class)
 public class JavaModuleDetector {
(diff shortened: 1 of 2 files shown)


ljacomet:

***

If we accept those changes, then the linked issue should be re-opened and possibly closed as completed unless there is further work needed to cover all cases.

kelemen:
I don't know what was the complete reason to close the issue, because the stated one in the comment was basically that lot of other tools have similar bugs, so Gradle can have it too.

That aside, I think the change in this PR is good regardless of that particular issue, because it is more efficient, and reads the jar the same way as the JVM does (in theory there could be other differences than in the referenced issue), which simply leaves less room for problems in the future. Also, the current implementation is more efficient, and has the exact same complexity.

big-guy:
This seems all right to us. We'll take a look for 9.5

> if a jar is modularized or not is cached, and if it is already cached, then this change has no effect.

If I'm following things correctly, the cache isn't a problem because it's Gradle-version specific, so the cache will only see one implementation of JavaModuleDetector at a time.

kelemen:
> If I'm following things correctly, the cache isn't a problem because it's Gradle-version specific, so the cache will only see one implementation of JavaModuleDetector at a time.

To be honest, I haven't checked how and where the cache is stored exactly (just the fact that it is cached in files), but I trust that you know this better than I do.

Issue #35207, opened 2025-10-02
Gradle 7.6.6 & 8.14.3: Problems reading data from Binary store


vicky17d:
### Current Behavior

Running into `Problems reading data from Binary store in /Users/<user>/.gradle/.tmp/gradle14622350702072981960.bin offset 213645 exists? true` when opening the project in IntelliJ IDEA

Able to reproduce this issue with the following snippet in one of the modules' build.gradle file and running `./gradlew` without any arguments
```
afterEvaluate {
    configurations.runtimeClasspath.resolvedConfiguration.lenientConfiguration.allModuleDependencies.each { println(it) }
}
```

Tried the following:
- Upgrade from 7.6.2 to 7.6.6 and then to 8.14.3 -> Issue exists at all 3 versions
- `./gradlew --stop` and cleaned up project's .gradle folder and also `~/.gradlew/.tmp` 

We are running into this problem in a large project at a specific commit where some dependencies are updated. After another commit, we don't see the problem but this is fairly a large project and used by a big team. Not able to sync the project in IntelliJ causes productivity loss for many engineers.

Here are the stacktrace:
```

### Expected Behavior

No such issue

### Context (optional)

_No response_

### Self-contained Reproducer Project

Not able to provide as the project is fairly large and issue happens at a specific commit and involves a lot of inhouse plugins.

### Gradle version

8.14.3

### Build scan URL (optional)

_No response_

### Your Environment (optional)

```
Failure 1 of 1A problem occurred configuring project ':module1:nested-module-fcap'
Build file '/Users/<username>/oncall/module1/module1/nested-module-fcap/build.gradle' line: 20
Fetching failure history...
Problems reading data from Binary store in /Users/<username>/.gradle/.tmp/gradle9800953731057484843.bin offset 185449 exists? true
[...]

PR #35957, opened 2025-12-06
Avoid detached subgraphs in resolved dependency graph

jvandort:
There were a few scenarios where we would short-circuit cleaning up a graph, leading to graphs that contained detached subgraphs with missing connections between nodes. This graph state surfaced as a corrupt serialized graph in the legacy ResolvedConfiguration serialized graph. This is because the TransientConfigurationResultsBuilder serializes all nodes of all components that are part of the graph, even if those nodes are not reachable from any other node in the graph.

We add additional validation to graphs, ensuring that the source node of all incoming edges to a node is itself part of the graph, ensuring the final graph of a resolution does not contain disconnected subgraphs.

We also attempted to add additional validation to the graph state, ensuring that all edges are connected to their source and target nodes correctly, however there are existing bugs in the engine preventing these validations from passing. We will need to address and un-comment these assertions as part of upcoming work

Fixes https://github.com/gradle/gradle/issues/35207

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
@@ -16,6 +16,7 @@
 
 package org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.LinkedHashMultimap;
@@ -138,12 +139,14 @@ public class NodeState implements DependencyGraphNode {
     /**
      * The transitive strict versions from inherited from parents.
      */
-    private StrictVersionConstraints ancestorsStrictVersions = StrictVersionConstraints.EMPTY;
+    @VisibleForTesting
+    StrictVersionConstraints ancestorsStrictVersions = StrictVersionConstraints.EMPTY;
 
     /**
      * Our own strict version constraints, from the previous graph traversal.
      */
-    private @Nullable StrictVersionConstraints ownStrictVersions;
+    @VisibleForTesting
+    @Nullable StrictVersionConstraints ownStrictVersions;
 
     /**
      * Cached copy of all endorsed strict versions. Must be invalidated whenever
@@ -856,7 +859,8 @@ public class NodeState implements DependencyGraphNode {
         return edgeExclusions;
     }
 
-    private void collectOwnStrictVersions(ExcludeSpec moduleResolutionFilter) {
+    @VisibleForTesting
+    void collectOwnStrictVersions(ExcludeSpec moduleResolutionFilter) {
         List<DependencyState> dependencies = dependencies(moduleResolutionFilter);
         Set<ModuleIdentifier> constraintsSet = null;
         for (DependencyState dependencyState : dependencies) {
@@ -887,6 +891,13 @@ public class NodeState implements DependencyGraphNode {
         StrictVersionConstraints existingOwnStrictVersions = this.ownStrictVersions;
         this.ownStrictVersions = newStrictVersions;
 
+        if (existingOwnStrictVersions == null) {
+            // If our existing strict versions are null, nobody else has observed them,
+            // so their value being initialized for the first time will no invalidate
+            // any existing calculated strict versions.
+            return;
+        }
+
         if (!newStrictVersions.equals(existingOwnStrictVersions)) {
             for (EdgeState incomingEdge : incomingEdges) {
                 if (incomingEdge.getDependencyMetadata().isEndorsingStrictVersions()) {
@@ -911,7 +922,8 @@ public class NodeState implements DependencyGraphNode {
      * Recompute the strict versions inherited from ancestors,
      * propagating the new value to all descendants.
      */
-    private void recomputeAncestorsStrictVersions() {
+    @VisibleForTesting
+    void recomputeAncestorsStrictVersions() {
         updateAncestorsStrictVersions(collectAncestorsStrictVersions());
     }
 
@@ -950,11 +962,19 @@ public class NodeState implements DependencyGraphNode {
         }
 
         if (incomingEdges.size() == 1) {
-            return getStrictVersionsForEdge(incomingEdges.get(0));
+            EdgeState dependencyEdge = incomingEdges.get(0);
+            if (dependencyEdge.getFrom().isSelected()) {
+                return getStrictVersionsForEdge(dependencyEdge);
+            } else {
+                return StrictVersionConstraints.EMPTY;
+            }
         }
 
         StrictVersionConstraints ancestorsStrictVersions = null;
         for (EdgeState dependencyEdge : incomingEdges) {
+            if (!dependencyEdge.getFrom().isSelected()) {
+                continue;
+            }
             StrictVersionConstraints allEdgeStrictVersions = getStrictVersionsForEdge(dependencyEdge);
 
             ancestorsStrictVersions = ancestorsStrictVersions == null
@@ -966,7 +986,7 @@ public class NodeState implements DependencyGraphNode {
                 break;
             }
         }
-        return ancestorsStrictVersions;
+        return ancestorsStrictVersions != null ?  ancestorsStrictVersions : StrictVersionConstraints.EMPTY;
     }
 
     /**
@@ -1011,7 +1031,7 @@ public class NodeState implements DependencyGraphNode {
 
         for (EdgeState outgoingEdge : outgoingEdges) {
             for (NodeState targetNode : outgoingEdge.getTargetNodes()) {
-                // The endorsed strict versions of this node contribute to the
+                // The endorsed strict versions of this node contributes to the
                 // ancestors strict versions of our children.
                 targetNode.recomputeAncestorsStrictVersions();
             }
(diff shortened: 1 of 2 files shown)


jvandort:
@bot-gradle test rfr

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/platforms/PlatformResolveIntegrationTest.groovy
```diff
@@ -717,6 +717,26 @@ class PlatformResolveIntegrationTest extends AbstractHttpDependencyResolutionTes
         //Shape of the graph is not checked as bug was failing resolution altogether
     }
 
+    def foo() {
```
jvandort: Something about these dependencies would cause the engine to absolutely blow up. This dependency graph would cause reentrant selection behavior. Now this no longer happens but we should probably minimize this

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/alignment/AlignmentIntegrationTest.groovy
```diff
@@ -451,7 +451,7 @@ class AlignmentIntegrationTest extends AbstractAlignmentSpec {
         expectAlignment {
             module('core') alignsTo('2.9.4') byPublishedPlatform()
             module('databind') tries('2.7.9') alignsTo('2.9.4') byPublishedPlatform()
-            module('annotations') tries('2.7.9') alignsTo('2.9.4') byPublishedPlatform()
+            module('annotations') tries('2.7.9') tries('2.9.0') alignsTo('2.9.4') byPublishedPlatform()
```
jvandort, line 454: These changes are caused by no longer eagerly removing outgoing edges from losing nodes of version/module/capability conflicts. We still remove the outgoing edges of these nodes when they are processed when they're removed from the queue (we onFewerSelected them when we retarget edges away from them), but new ordering causes us to try a few different versions here.

The advantage of this new approach is that we no longer are susceptible to reentrant module selection, preventing concurrent modification exceptions during module selection.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/ModuleResolveState.java
```diff
@@ -419,6 +403,7 @@ boolean isVirtualPlatform() {
     }
 
     void disconnectIncomingEdge(NodeState removalSource, EdgeState incomingEdge) {
+        incomingEdge.clearSelector();
```
jvandort, line 405: Previously, we cleared our selector _after_ we removed incoming attached and unattached constraints. 

In some cases like when a constraint was already "controlling" the version, this would cause the removal of constraints to trigger selection, which then caused us to select the hard edge that we are currently trying to remove `incomingEdge`. So, we make sure to clear our hard selector first so that removing the incoming constraints does not cause us to try to select the version of the hard edge we're removing. 

In practice, incomingEdge.clearSelector, and any edge removal in `clearIncomingAttachedConstraints` and `clearIncomingUnattachedConstraints` can cause selection to occur, but we don't want to try to trigger selection at all in these cases. We know we are removing all incoming edges if we get down to clearing constraints. At the end of this method if `pendingDependencies.isPending()`, the module should have no selectors, so there is no point in performing selection at any point in this. method (except for if there are multiple hard edges targeting this module)

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
```diff
@@ -347,11 +346,17 @@ private void doVisitDependencies(ExcludeSpec resolutionFilter, StrictVersionCons
...
         if (upcomingNoLongerPendingConstraints != null && potentiallyActivatedConstraints != null) {
-            for (ModuleIdentifier module : upcomingNoLongerPendingConstraints) {
-                Collection<DependencyState> dependencyStates = potentiallyActivatedConstraints.get(module);
+            for (ModuleIdentifier moduleId : upcomingNoLongerPendingConstraints) {
+                Collection<DependencyState> dependencyStates = potentiallyActivatedConstraints.get(moduleId);
                 if (!dependencyStates.isEmpty()) {
-                    for (DependencyState dependencyState : dependencyStates) {
-                        createAndLinkEdgeState(dependencyState, discoveredEdges, resolutionFilter, ancestorsStrictVersions, false);
+                    ModuleResolveState module = resolveState.getModule(moduleId);
+                    if (module.isPending()) {
```
jvandort, line 356: See `dependency constraint can be not pending, then pending, then not pending and still participate in resolution`. 

We would notify this node that one of its outgoing constraints needs to be activated, but in between that time and when this node comes up in the queue, the target module of the potential constraint goes back to pending. We previously didn't detect this and then went all the way to trying to attach the edge to the target node in the pending module before realizing this. 

We detect this much earlier and no longer need to try to detach this stale edge during `attachToTargetNodes`

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
```diff
@@ -1255,15 +1236,6 @@ boolean isSelectedByVariantAwareResolution() {
         return selectedByVariantAwareResolution && isSelected();
     }
 
-    void makePending(EdgeState edgeState) {
```
jvandort, line 1255: Was identical to `removeOutgoingEdge` above

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/ModuleResolveState.java
```diff
@@ -217,16 +212,10 @@ private void changeSelection(ComponentState newSelection) {
         assert this.selected != newSelection;
         assert newSelection.getModule() == this;
 
-        changingSelection = true;
-
-        // Remove any outgoing edges for the current selection
-        selected.removeOutgoingEdges();
```
jvandort, line 222: Removing outgoing edges from a node at any time can cause module version selection (as we clear selectors, which may potentially release them, which may trigger selection). 

Below, in doRestart, we take any incoming edge to this module and retarget them to the new selected component. As part of this, any incoming edges to the selected node are removed from that node, causing it to be `onFewerSelected`'d. This puts it back in the queue, and _then_ we will remove its outgoing edges (since it at that point no longer has incoming edges, as they were all retargeted away).

This allows us to avoid reentrant behavior in module version selection.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
```diff
@@ -1078,28 +1081,14 @@ public void removeOutgoingEdges() {
...
-        boolean alreadyRemoving = removingOutgoingEdges;
-        removingOutgoingEdges = true;
-        if (!outgoingEdges.isEmpty() && !alreadyRemoving) {
+        if (!outgoingEdges.isEmpty()) {
             for (EdgeState outgoingEdge : outgoingEdges) {
-                outgoingEdge.markUnused(); // Track that these edges have been removed from outgoing but maybe not from incoming, in case we hit one of these continues and do not call disconnectOutgoingEdge
-                ComponentState targetComponent = outgoingEdge.getTargetComponent();
-                if (targetComponent == component) {
-                    // if the same component depends on itself: do not attempt to cleanup the same thing several times
-                    continue;
```
jvandort, line 1087: This continue was the original cause of the bug we are trying to fix. We detect some edge case (the original reason was to avoid some CME due to reentrant behavior), and then just skip the work we need to do. 

We can't just skip the work that has to be done here, as that leads to incorrectly connected graph states, thus causing the bug we observed.

The second `continue` below was not observed in the original reproducer that drove this PR, but it can likely cause the same behavior (skipping removing an outgoing edge when we were supposed to). In order to resolve _this_ instance, we needed to make additional changes to avoid reentrant behavior specifically in module version selection.

jvandort:
@bot-gradle test rfr apt

jvandort:
@bot-gradle test rfr

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
```diff
@@ -948,11 +963,19 @@ private StrictVersionConstraints collectAncestorsStrictVersions() {
         }
 
         if (incomingEdges.size() == 1) {
-            return getStrictVersionsForEdge(incomingEdges.get(0));
+            EdgeState dependencyEdge = incomingEdges.get(0);
```
jvandort, line 965: We might need these changes in 9.3

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/platforms/PlatformResolveIntegrationTest.groovy
jvandort: Lets remove this from the PR

jvandort:
@bot-gradle test rfr

ljacomet:
@jvandort From looking at the reason this PR was removed from the merge queue, it looks like there is a snippet build failure that can be seen [here](https://builds.gradle.org/buildConfiguration/Gradle_Release_Check_DocsTest_Windows_virtual_Batch_4_1/106659365?buildTab=overview&focusLine=NaN). Unfortunately no scan because the chain got cancelled.

PR #35958, opened 2025-12-06
Polish `BuildModelParameters` methods naming and javadocs

alllex:
Aligns the naming of parameters that are related to Tooling Model Building. Also adds some javadocs to explain parameters and their applicability.


alllex:

PR #35962, opened 2025-12-07
Share test file helpers between integ and cross-version tests

alllex:
Extracts common file locations used during testing into a trait that can be shared between Integration Tests (`AbstractIntegrationSpec`) and Cross-Version Tooling API tests (`ToolingApiSpecification`).

Since the locations are now shared, it's also possible to bring the snippet syntax-highlighting goodness of integration tests into the cross-version TAPI tests.


alllex:

PR #35963, opened 2025-12-07
Finding Gradle sources intelligently in SourceDistributionProvider

### LINKED ISSUE #27863: SourceDistributionProvider contains hardcoded distributionUrl referring to https://services.gradle.org
### Current Behavior

I reset distributionUrl ,but it  not work correctly in IDEA, but it will build successfully in terminal

image

### Expected Behavior

all to  donwload gradle-x.zip will use my set up distributionUrl  in IDEA 

### Context (optional)

https://youtrack.jetbrains.com/issue/IDEA-344183/Ive-set-up-a-proxy-image-for-gradle-but-IDEA-doesnt-actually-use-proxies-everywhere.

### Steps to Reproduce

https://github.com/aohanhongzhi/JavalinDemo-Kotlin

### Gradle version

8.0 and 8.5

### Build scan URL (optional)

https://gradle.com/s/eqdlih3yq5elk

ubuntu , window , macos

kelemen:
Currently Gradle has the `https://services.gradle.org/distributions(-snapshots)` URL hardwired when the IDE is downloading the sources of Gradle. This is a problem for people sitting behind a strict proxy. This PR aims to solve this problem using the following strategy (in preference order):

1. The URL, artifact pattern and credentials can be specified in new Gradle properties which are always used when set.
2. The wrapper URL is read from `gradle-wrapper.properties`, and when its hostname is `services.gradle.org`, then we fallback to the old behavior. This is to take the least risk with currently working setups.
3. If the filename in the wrapper URL matches the same pattern as in the official repository, then we will use the base URL (URL without the filename) as the repository and expect the sources under that URL.

The change also recognizes the credentials set in `gradle.wrapperUser` and `gradle.wrapperPassword`.

* Fixes #27863. 

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ljacomet:

eskatos:
Hi Attila,
Thank for the PR. This is looking good.
Adding new system/gradle properties raises questions. We should be careful not adding too many knobs.
Could you consider reducing this PR to all the other changes as a first step?
Namely 2/ and 3/. They make sense on their own and provide a good enhancement already.

kelemen:
@eskatos Sure, I will remove the system properties from this PR. Do you want me to open a new PR on top of these changes to include system properties so that you can independently consider them?

Also, I have seen that bearer token support was added in the mean time for the Gradle wrapper, so I will add the bearer token support here as well for consistency.

kelemen:
I have updated the PR with the following changes to the previous state:

1. It is not longer possible to customize the source distribution URL via Gradle properties. Either we can automatically transform the distribution URL or fallback to the default.
2. Supports the same authorization mechanism as the wrapper (that is, supports bearer tokens as well). The shared wrapper code was refactored to rely on the same code (as much as possible) to configure the authorization mechanism.
3. While making the authorization configuration shared, I have removed the reflection hack from the wrapper, because the wrapper now requires Java 8 anyway, so `Base64` is always available.
4. I was forced to add some `jspecify` annotations into `wrapper-shared`, I hope that is fine.

As previously mentioned, I can open a separate PR for making the URLs fully customizable on top of these changes if needed.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/internal/GradleDistRepoDescriptorLocator.kt
```diff
@@ -0,0 +1,173 @@
...
+            val header = credentials.authorizationHeader()
+            name = header.key
+            value = header.value
+        }
+    }
+}
+
+data class GradleDistVersion(
+    val versionStr: String,
+    val snapshot: Boolean = versionStr.contains('+')
```
eskatos: 💅 Let's name this `isSnapshot`

eskatos:

kelemen:
I have updated the wrapper jar hash in the tests in a separate commit. Tell me if you want to squash the commits.

eskatos (approved):
LGTM!

eskatos:
No need to squash, thanks.
I amended with a few refinements.
I'll send this for merge now.

eskatos:
Thank you Attila, this will be in Gradle 9.4.0

PR #35983, opened 2025-12-09
Add names to Task's onlyIf, cacheIf, actions, etc. in CC report

### LINKED ISSUE #35721: Provide better attribution for task-captured closures in the configuration cache report
### Expected Behavior

Task-captured closures, like `doLast`, `onlyIf`, `upToDateWhen` should be clearly attributed in the configuration cache report if they capture unsupported types.

Only some task fields are reported to hint what captures unsupported types:

Image

### Context

It is hard to understand what captures the references when the task contains multiple lambdas:
```kotlin
fun myFalse() = false

tasks.register("myTask") {
    outputs.cacheIf {
        myFalse()
    }
    outputs.doNotCacheIf("reason") {
        myFalse()
    }
    outputs.upToDateWhen {
        myFalse()
    }

    onlyIf {
        myFalse()
    }

    doLast {
        myFalse()
    }
}
```

mlopatkin:
Fixes #35721

- [x] Update the report library version to include gradle/configuration-cache-report#44

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
@@ -464,7 +464,8 @@ class ConfigurationCacheTaskSerializationIntegrationTest extends AbstractConfigu
         configurationCacheFails "myTask"
 
         then:
-        problems.htmlReport(failure).assertContents {
+        // test HTML report directly as console output has only one unique problem as it ignores deep property traces
+        problems.htmlReport(failure.error).assertContents {
             totalProblemsCount = 5
             problemsWithStackTraceCount = 0
 
--- a/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
+++ b/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
@@ -71,7 +71,8 @@ class ConfigurationCacheProblemsFixture {
     }
 
     /**
-     * Creates a fixture to assert on the report based on the file URL written in the build output. The report URL may be absent in the output.
+     * Creates a fixture to assert on the report based on the file URL written in the build output. The report URL may be absent in the output
+     *  (happens when there are inputs but no problems, unless <code>--info</code> is enabled).
      *
      * @param output the output of the build
      */
--- a/testing/internal-integ-testing/src/test/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixtureTest.groovy
+++ b/testing/internal-integ-testing/src/test/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixtureTest.groovy
@@ -39,12 +39,12 @@ class ConfigurationCacheProblemsFixtureTest extends Specification {
     private TestFile reportFile = reportDir.file("configuration-cache-report.html")
     private def fixture = new ConfigurationCacheProblemsFixture(rootDir)
 
-    def report() {
-        fixture.htmlReport()
+    ConfigurationCacheReportFixture report() {
+        return fixture.htmlReport()
     }
 
-    def report(TestFile reportFile) {
-        fixture.htmlReport("See the complete report at " + new ConsoleRenderer().asClickableFileUrl(reportFile))
+    ConfigurationCacheReportFixture report(TestFile reportFile) {
+        return fixture.htmlReport("See the complete report at " + new ConsoleRenderer().asClickableFileUrl(reportFile))
     }
 
     def "findReportFile finds a single report dir"() {
@@ -391,7 +391,7 @@ class ConfigurationCacheProblemsFixtureTest extends Specification {
     private TestFile generateReportFile(List<String> problemMessages, List<List<String>> problemsAndPropertyTraces = [], List<List<String>> problemsAndStacktraceParts = []) {
         List<String> problemMarkup = []
         problemMessages.eachWithIndex { String problemText, int index ->
-            problemMarkup << """
+            def fragment = """
             {
                     "problem": [{
                         "text": "${problemText}"
@@ -412,6 +412,7 @@ class ConfigurationCacheProblemsFixtureTest extends Specification {
             """
             }
             """
+            problemMarkup << fragment.toString()
         }
         def jsonData = """
 // begin-report-data


mlopatkin:

mlopatkin:
@bot-gradle test this

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -96,12 +96,16 @@ class TaskNodeCodec(
             writeLong(task.taskIdentity.id)
             writeNullableString(task.reasonTaskIsIncompatibleWithConfigurationCache.orElse(null))
 
+            // Interesting fact: changing the order in which stuff is written may break builds,
+            // if writing something triggers a side effect that modifies stuff written later.
+            // For example, serializing a file collection may trigger dependency resolution callbacks.
+            // At least one dependency management test was relying on the callback modifying a list referenced by the task action.
```
abstratt, line 102: An example:

https://github.com/gradle/gradle/blob/d2662472fe09519aefab4e2699727c83c3988556/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/rules/ComponentMetadataRulesIntegrationTest.groovy#L373

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -110,6 +114,7 @@ class TaskNodeCodec(
                             )
                         }
                     }
+                    writeTaskActions(task)
```
abstratt, line 117: ❓ 

How were we storing task actions before? (were we?)

abstratt: (and if not, why do we need to do that now?)

subprojects/core/src/main/java/org/gradle/api/internal/AbstractTask.java
```diff
@@ -278,6 +278,11 @@ public List<InputChangesAwareTaskAction> getTaskActions() {
         return actions;
     }
 
+    @Override
+    public void setTaskActions(List<InputChangesAwareTaskAction> taskActions) {
```
abstratt: ❓ 

Shouldn't we set `hasCustomActions` to true if not empty?

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
mlopatkin: > (and if not, why do we need to do that now?)

It is somewhat magical, or at least "action-at-distance". We serialize task instances as beans, but except for fields defined in `AbstractTask`, like `project` or `identity`. However, there is an exception to exception defined there:
https://github.com/gradle/gradle/blob/2789575f585c97dab5aebcaf24b9c6f7341b372f/platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/BeanSchema.kt#L167

`actions` used to be there, but I replaced it with a Virtual Field instead.

subprojects/core/src/main/java/org/gradle/api/internal/AbstractTask.java
abstratt: Hmm, so that is covered by regular bean serialization of task instances?

mlopatkin: > Shouldn't we set `hasCustomActions` to true if not empty?

The array also contains `@TaskAction` methods, so it isn't possible. We didn't do that before as well, and the only place the flag is used is to trigger some logging in `org.gradle.api.internal.tasks.execution.TaskExecution#shouldDisableCaching`.

mlopatkin: > Hmm, so that is covered by regular bean serialization of task instances?

No, as it isn't part of the `abstractTaskRelevantFields` allowlist. Just nobody cared about the flag, I guess.

mlopatkin:
The report is in.

subprojects/core/src/main/java/org/gradle/api/internal/TaskInternal.java
```diff
@@ -47,12 +47,26 @@ public interface TaskInternal extends Task, Configurable<Task> {
     @Internal
     List<InputChangesAwareTaskAction> getTaskActions();
 
+    /**
+     * Sets the task actions for this task. Used by the configuration cache when restoring the task.
+     *
+     * @param taskActions the actions to restore
+     */
+    void setTaskActions(List<InputChangesAwareTaskAction> taskActions);
```
abstratt: 💭 

I liked the "restore" pattern you used for `onlyIf` for setters that are really only meant to be invoked by CC.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -253,6 +253,61 @@ ${text}
...
+     * <p>
+     * Calling this method multiple times adds other locations to the expected list.
+     * The location expectations are order-sensitive.
+     * <p>
+     * This doesn't check the stacktrace, but the property trace instead, so it is better suited for serialization failures.
+     *
+     * @param location the initial location (e.g. task name)
+     * @return the spec to further refine the expected location.
+     */
+    TraceSpec at(String location) {
```
abstratt, line 281: 💭 

~Maybe use `withLocation` to follow the pattern we have for other specs~

abstratt: Never mind, this returns a different object

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
```diff
@@ -429,4 +432,49 @@ class ConfigurationCacheTaskSerializationIntegrationTest extends AbstractConfigu
...
+        when:
+        configurationCacheFails "myTask"
+
+        then:
+        problems.htmlReport(failure).assertContents {
+            totalProblemsCount = 5
+            problemsWithStackTraceCount = 0
+
+            withProblem("cannot serialize Gradle script object references as these are not supported with the configuration cache.") {
+                at(":myTask").at("upToDate specs")
```
abstratt, line 473: 💭 

Took me a while to understand these two methods are different things (`ProblemSpec.at()` creates a new `TraceSpec`, `TraceSpec.at()` changes the target `TraceSpec`). I think using different names would help.

abstratt: Ideas:

* `ProblemSpec.locatedAt()` and `TraceSpec.at()`
* `ProblemSpec.at()` and `TraceSpec.plus()`

abstratt:
I am rebasing this to catch up with https://github.com/gradle/gradle/pull/36047

testing/internal-integ-testing/src/test/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixtureTest.groovy
```diff
@@ -305,55 +379,47 @@ class ConfigurationCacheProblemsFixtureTest extends Specification {
...
+        def problemMessages = (0..<problems).collect { "Some problem ${it+1}".toString() }
+        def problemStacktraces = (0..<problemsWithStacktrace).collect { index ->
+            ["somePart"]
+        }
+        generateReportFile(problemMessages, [], problemStacktraces)
     }
 
-    private TestFile generateReportFile(Map<String, List<String>> problemsAndStacktraces) {
-
+    private TestFile generateReportFile(List<String> problemMessages, List<List<String>> problemsAndPropertyTraces = [], List<List<String>> problemsAndStacktraceParts = []) {
```
abstratt, line 391: Change to lists because problem messages don't need to be unique

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
6hundreds: I also in favour of having different names for these, like `of(":myTask").at("upToDate specs")`

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredBuildInputsIntegrationTest.groovy
```diff
@@ -868,7 +868,7 @@ internal property sun.java.command changes.
         outputContains("Configuration: some.property.1 = 1")
         outputDoesNotContain("Configuration: some.property.2 = 2")
         problems.assertResultHasProblems(result) {
-            withInput("Gradle runtime: Gradle property 'some.property.*'")
+            withInput("Project ':': Gradle property 'some.property.*'")
```
6hundreds, line 871: 👍

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummary.kt
```diff
@@ -214,14 +238,18 @@ class Summary(
      * This should exclude problems with severity [ProblemSeverity.SuppressedSilently], which are still
      * accounted for in [totalProblemCount] and shown in the HTML report, but intentionally omitted
      * from the console to keep output noise low during graceful degradation.
+     *
+     * Also, these problems are console-unique (but not necessarily report unique, which take the full property trace chain into account).
```
6hundreds: ❓ 
The part `are console-unique (but not necessarily report unique` is not 100% clear to me :
Console problems are the subset of report problems, so if problems are different for the console, they are certainly different for the report. Isn't it?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummary.kt
```diff
@@ -214,14 +238,18 @@ class Summary(
...
      */
     private
     val consoleProblemCauses: Map<ProblemCause, ProblemSeverity>,
 
     /**
      * Count of problems that should appear in the HTML report.
      */
-    val uniqueProblemCount: Int,
+    @get:VisibleForTesting
```
6hundreds, line 250: ❓ 
You can access internal field in tests, why do we need `@VisibleForTesting`?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummary.kt
```diff
@@ -380,14 +409,29 @@ internal
...
-    val documentationSection: DocumentationSection?
+    val documentationSection: DocumentationSection?,
+    val traceHash: Int?
 ) {
+    /**
+     * A problem cause may reflect the full [org.gradle.internal.configuration.problems.PropertyTrace] stack (as required by the CC report) or
+     * (if they are shallow) just the top location (as required by the console).
+     */
+    private
+    val shallowTrace get() = traceHash == null
```
6hundreds: 💅 I think it's more readable to name `boolean-returning` properties starting from `is/are`
`isShallowTrace`

platforms/core-configuration/configuration-cache/src/test/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummaryTest.kt
```diff
@@ -55,7 +56,7 @@ class ConfigurationCacheProblemsSummaryTest {
             subject.onProblem(buildLogicProblem("build.gradle", "failure"), ProblemSeverity.Deferred)
         )
         assertThat(
-            subject.get().uniqueProblemCount,
```
6hundreds, line 58: ❓ `uniqueProblemCount` became `reportUniqueProblemCount` in this PR. Did we change our test assertions to console problems intentionally?

platforms/core-configuration/configuration-cache/src/test/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummaryTest.kt
```diff
@@ -268,32 +270,87 @@ See the complete report at $REPORT_URL
...
+        val unknown = ProblemCause.of(PropertyProblem(PropertyTrace.Unknown, message))
+        val project1 = ProblemCause.of(PropertyProblem(PropertyTrace.Project(":p1", PropertyTrace.Gradle), message))
+        val project1b = ProblemCause.of(PropertyProblem(PropertyTrace.Project(":p1", PropertyTrace.Gradle), message))
+        val project2 = ProblemCause.of(PropertyProblem(PropertyTrace.Project(":p2", PropertyTrace.Gradle), message))
+        val project2ThenUnknown = ProblemCause.of(PropertyProblem(PropertyTrace.Project(":p2", PropertyTrace.Unknown), message))
+        assertEquals(gradle, gradle2)
+        assertNotEquals(gradle, unknown)
+        assertEquals(project1, project1b)
+        assertNotEquals(project1, project2)
+        assertNotEquals(project2, project2ThenUnknown)
```
6hundreds, line 304: 🤔 ProblemCause is a data class, here I have feeling we're testing data class's equals method, what is not something really useful. Am I miss something?

platforms/core-configuration/configuration-cache/src/test/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummaryTest.kt
```diff
@@ -268,32 +270,87 @@ See the complete report at $REPORT_URL
...
+        assertThat(gradleRuntime.message, equalTo("failure"))
+        assertThat(gradleRuntime.userCodeLocation, equalTo("Gradle runtime"))
+
+        val unknown = ProblemCause.of(PropertyProblem(PropertyTrace.Unknown, StructuredMessage.forText("failure")))
+        assertThat(unknown.message, equalTo("failure"))
+        assertThat(unknown.userCodeLocation, equalTo("unknown location"))
+
+        val projectAtUnknown = ProblemCause.of(PropertyProblem(PropertyTrace.Project(":p1", PropertyTrace.Unknown), StructuredMessage.forText("failure")))
+        assertThat(projectAtUnknown.message, equalTo("failure"))
+        assertThat(projectAtUnknown.userCodeLocation, equalTo("unknown location"))
```
6hundreds, line 287: 🤔 I might be wrong, but we're testing here just values to be set. How useful this test is?

platforms/core-configuration/configuration-cache/src/test/kotlin/org/gradle/internal/cc/impl/problems/PropertyTraceTest.kt
```diff
@@ -58,4 +59,41 @@ class PropertyTraceTest {
...
+                PropertyTrace.Property(PropertyKind.PropertyUsage, "prop1", PropertyTrace.Unknown).fullHash
+            )
+        )
+        assertThat(
+            PropertyTrace.Property(PropertyKind.PropertyUsage, "prop1", PropertyTrace.Unknown).fullHash,
+            not(
+                equalTo(
+                    PropertyTrace.Property(PropertyKind.PropertyUsage, "prop1", PropertyTrace.Gradle).fullHash
+                )
+            )
```
6hundreds: 💅 Not a blocker, but how verbose `not(equalTo())` is. `assertEquals` would fit well here.

subprojects/core/src/main/java/org/gradle/api/internal/TaskInternal.java
6hundreds: 💅 I think `set` naming convention should be consistent: either `set` or `restore` works, but it seems like `set` is more common.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummary.kt
abstratt: For documentation ("why isn't this thing private?")

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummary.kt
abstratt: Yes, but not the other way around. Report problems are unique considering deep property traces. Console problems are unique considering only the root property trace (two different problems with the same root property trace but different "tail" traces are redundant for the console).

We could call them "shallow-unique" and "deep-unique" (or something like that - but I like these alternatives less than console-unique and report-unique).

platforms/core-configuration/configuration-cache/src/test/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummaryTest.kt
abstratt: This ensures problem causes are build as expected given different types of problems and their traces.  

It shows the user code location comes from the final trace, not the first one.

It also shows structured messages are collected as rendered strings.

platforms/core-configuration/configuration-cache/src/test/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummaryTest.kt
abstratt: We are explaining to the reader the impact trace chains have in (comparing deep) problem causes. I will make the `traceHash` property private (which was an oversight). I will add another test case (or assertions) for comparing problem causes with shallow traces.

subprojects/core/src/main/java/org/gradle/api/internal/TaskInternal.java
abstratt: The thing is that these are not regular setters. They are internal/"friends"-only raw setters, should not result in additional computation being performed (validation, events fired etc). Mike had to choose a slightly different base property name "taskActions" because the simpler "actions"-based name was already taken by the proper/more general getter. Maybe we should change it to `doSet`.

platforms/core-configuration/configuration-cache/src/test/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummaryTest.kt
abstratt: Good catch. I changed the tests to inspect both report and console unique counts

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
abstratt: I will leave this for a future improvement, I am not confident on a better replacement myself

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummary.kt
6hundreds: I'm good with `console-unique` and `report-unique`. 
What I'm trying to say is `"but not necessarily report unique"` part of the comment is a bit miss leading - as console operates over root traces, uniqueness at the console level gives you uniqueness on the report level automatically, but not the other way around. So under this condition the statement `"but not necessarily report unique"` is not correct, because those problems 100% unique on the report level.

It's not a blocker to me, I just wanted to verify my understanding

6hundreds (approved):
LGTM!

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummary.kt
abstratt: Ah, got it now. I changed it to:

```
* Also, these problems are console-unique (which is different from report-unique, which take the full property trace chain into account).
```

abstratt:
Will address release notes in separate

abstratt:
See #36153 for release notes

subprojects/core/src/main/java/org/gradle/api/internal/TaskInternal.java
mlopatkin: I really like the consistent `restore` pattern we ended up with.

PR #35986, opened 2025-12-09
Make `ServiceRegistry` thread-safe via persistent collections

### LINKED ISSUE #35923: ServiceRegistry service registration is not thread-safe
Specifically, operations on `DefaultServiceRegistry.OwnServices#providersByType` that happen during service registration.

For instance, here:
https://github.com/gradle/gradle/blob/ab44028f8acabef3ba8a8d682b0bab3060839f3b/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java#L462

Some recent related improvements in the area:
- https://github.com/gradle/gradle/pull/35849

bamboo:
Fixes #35923.

This change-set brings three fully persistent collection implementations:

- PersistentArray
- PersistentSet
- PersistentMap

The implementations are suitable for use everywhere including in workers.

They are faster or as fast as the state of the art implementations in Clojure, Scala and Capsule.

PersistentArray is implemented via a bit-array mapped trie with a branching factor of 32, a.k.a, Clojure persistent vectors.

See Chapter 2 of [Improving RRB-Tree Performance through Transience](https://hypirion.com/thesis.pdf) for a full account.

PersistentSet and PersistentMap are implemented via Compressed Hash-Array Mapped Prefix-trees, a.k.a., CHAMP tries.

See [Optimizing Hash-Array Mapped Tries for Fast and Lean Immutable JVM Collections](https://michael.steindorfer.name/publications/oopsla15.pdf).

These collections will unlock many optimizations to dependency resolution logic, the virtual file system and more. They will also allow many simplifications to code implementing custom immutable collections (e.g. `ImmutableActionSet`).

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/collections/src/test/groovy/org/gradle/internal/collect/PersistentMapTest.groovy
+++ b/platforms/core-runtime/collections/src/test/groovy/org/gradle/internal/collect/PersistentMapTest.groovy
@@ -16,6 +16,7 @@
 
 package org.gradle.internal.collect
 
+import org.gradle.util.internal.ToBeImplemented
 import spock.lang.Specification
 
 class PersistentMapTest extends Specification {
@@ -454,5 +455,31 @@ class PersistentMapTest extends Specification {
         and: 'Verify the maps actually have different values'
         map1.get(key2) == "value2"
         map2.get(key2) == "DIFFERENT_VALUE"
+    }
+
+    @ToBeImplemented
+    def "maps with different values have different hash codes"() {
+        when:
+        def m1 = PersistentMap.of("a", 1)
+        def m2 = PersistentMap.of("a", 2)
+
+        then:
+        m1.hashCode() != m2.hashCode()
+        m1.assoc("a", 2).hashCode() == m2.hashCode()
+        m1.hashCode() == m2.assoc("a", 1).hashCode()
+
+        when:
+        def m3 = m1.assoc("b", 2).assoc("c", 3)
+        def m4 = m1.assoc("b", 3).assoc("c", 3)
+
+        then:
+        // Undesirable behavior (hash code collisions)
+        m3.hashCode() == m4.hashCode()
+        m3.assoc("b", 3).hashCode() == m4.hashCode()
+        // Desirable behavior
+//        m3.hashCode() != m4.hashCode()
+//        m3.assoc("b", 3).hashCode() != m4.hashCode()
+//        m3.hashCode() != m4.assoc("b", 2).hashCode()
+//        m3.dissoc("b").hashCode() == m4.dissoc("b").hashCode()
     }
 }


bamboo:
@bot-gradle test QFL

bamboo:
@bot-gradle test this

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -391,17 +391,24 @@ public void visit(Service service) {
...
         private final CompositeStoppable stoppable = CompositeStoppable.stoppable();
-        private final AtomicReference<ServicesSnapshot> services = new AtomicReference<>(ServicesSnapshot.EMPTY);
+        private final AtomicReference<ServicesSnapshot> services;
 
         public OwnServices() {
-            providersByType.put(ServiceRegistry.class, Collections.<ServiceProvider>singletonList(new ThisAsService(ServiceAccess.getPublicScope())));
+            final ThisAsService thisServiceRegistry = new ThisAsService(ServiceAccess.getPublicScope());
+            services = new AtomicReference<>(
+                ServicesSnapshot.of(
+                    PersistentMap.of(
```
bamboo, line 401: ℹ️ This is the main functional change. The service provider mappings were moved from a mutable `HashMap<Class<?>, List<ServiceProvider>>` to an immutable `PersistentMap<Class<?>, PersistentArray<ServiceProvider>` whose changes are coordinated under an `AtomicReference`.

bamboo:
@bot-gradle test RFN

jvandort:
I worry about introducing a bunch of code we didn't write to our codebase and the burden of maintaining it. 

Is there not a modern performant collections library out there that is actively maintained?

bamboo:
> code we didn't write

What do you mean "code we didn't write"? I wrote every single line of code here.

jvandort:
Huh guess I saw a bunch of code and the links to the papers on persistent collections and assumed we forked the library you brought up the other day. 

In any case I have at least some minor concerns about maintaining such a collections library.

If we wrote all of this from scratch and will be maintaining this ourselves, why not just fork the other library? How does this implementation differ from what is available?

bamboo:
> Huh guess I saw a bunch of code and the links to the papers on persistent collections and assumed we forked the library you brought up the other day.

Ok, that's not what happened.

> How does this implementation differ from what is available?

These implementations are suitable for use everywhere including in workers, i.e., we can use them with JDK 8.

They are faster or as fast as the state of the art implementations in Clojure, Scala and Capsule (the library I brought up the other day). I carefully built them that way.

And we can add exactly the functionality we need the way we need it.

> I have at least some minor concerns about maintaining such a collections library.

I understand not being familiar with the algorithms but these are very well documented. We have more mysterious code in the codebase.

bamboo:
> I wrote every single line of code here.

FTR: with c80c631ba939836058bdb6bb3b9e10ba7ec51d0f and e9db14ae4d868fbb19700f56d75813d255f670a3, the statement is no longer true 😉

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/MutableBoolean.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.collect;
+
+/// Guess what?
```
jvandort: What?

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/ChampIterator.java
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+package org.gradle.internal.collect;
+
+import org.jspecify.annotations.Nullable;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+// ✅ Iterates over CHAMP trie using an explicit stack (linked list of IteratorState).
```
jvandort: The emojis in the comments are an interesting touch

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/ArrayCopy.java
```diff
@@ -0,0 +1,140 @@
...
+
+package org.gradle.internal.collect;
+
+import java.util.Arrays;
+
+/// Different array copying operations required by the persistent collection
+/// implementations.
+///
+/// In general, these should be as efficient as possible, contain
+/// no branching code, and minimize [System] calls.
```
jvandort: What's wrong with System calls specifically compared to other code?

I see a lot of `System` usages in this class

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/ArrayCopy.java
```diff
@@ -0,0 +1,140 @@
...
+
+    static Object[] insertAt(int index, Object[] array, Object e1, Object e2) {
+        Object[] newArray = Arrays.copyOf(array, array.length + 2);
+        System.arraycopy(array, index, newArray, index + 2, array.length - index);
+        newArray[index] = e1;
+        newArray[index + 1] = e2;
+        return newArray;
+    }
+
+    // 🤔 Note: This has different behavior based on payload (0 vs 1).
```
jvandort: Seems like this should be two different methods then? Why is payload an integer if we only allow 0 and 1?

jvandort:
What's with the emojis in the comments?

bamboo:
> What's with the emojis in the comments?

Those comments came from the AI review and I decided to leave them in as a reminder.

jvandort:
A reminder of our AI overlords?

PR #36004, opened 2025-12-11
Make test more resilient to temporary server busyness

tresat:


subprojects/core/src/test/groovy/org/gradle/util/internal/DistributionLocatorIntegrationTest.groovy
```diff
@@ -46,12 +48,26 @@ class DistributionLocatorIntegrationTest extends Specification {
         urlExist(locator.getDistributionFor(distributions.mostRecentReleaseSnapshot.version))
     }
 
-    void urlExist(URI url) {
+    private void urlExist(URI url) {
+        int responseCode = 0
+        for (int attempt = 1; attempt <= NUM_RETRIES; attempt++) {
```
jvandort: I would expect we already have a utility somewhere for retries/backoff. It's pretty simple here so I don't really have a problem duplicating but if it's easy to find an existing fixture id rather we use that

jvandort (approved):

subprojects/core/src/test/groovy/org/gradle/util/internal/DistributionLocatorIntegrationTest.groovy
bamboo: We have `ExponentialBackoff`.

tresat: Good idea, thanks for pointing that out.

I've reworked the test to use exponential backoff and clean it up some.  Please take another look.

tresat:
I'm satisfied with this.  If anyone looks at it while I'm away for the holidays and has no objections, please press merge.

blindpirate (commented):
Hi @tresat , thanks for making this PR!

After carefully reading `DistributionLocator`, here's my 2c:

1. This test should verify that `DistributionLocator` generates the correct URLs, not that the URLs are reachable. `DistributionLocator` only constructs URIs; it doesn't download. Adding retry logic shifts the test toward integration/connectivity, which is out of scope for a unit test.

2. Retries don't help when the server is down: If the server is unavailable (503/504), retries won't fix it. The test will still fail, just slower. Retries help with transient network issues, not server outages.

3. The network connectivity is covered by promotion job (there is an integration test there to make sure the distribution is download-able).

I've made another PR (well, credit for AI) in https://github.com/gradle/gradle/pull/36140

tresat:
@blindpirate I agree this test acts as more than a unit test, as it accesses the network.  It is named `*IntegrationTest`, which is also confusing.

I have no strong feelings on which change is more appropriate.  If everything here is already covered by the promotion job, then let's just do your change (and rename the file to `*Test` instead of `*IntegrationTest`).  My only concern here was to prevent getting bitten by this specific flakiness again, not to comprehensively handle all failure codes.

PR #36014, opened 2025-12-12
Revert "Print failed scenario's url for troubleshooting"

blindpirate:
Reverts gradle/gradle#36012


cobexer:
Looks like we don't need this currently any more

PR #36015, opened 2025-12-12
[Investigation] Run tapi builders always

asodja:


asodja:
Replaced by https://github.com/gradle/gradle/pull/36037

PR #36016, opened 2025-12-12
Mark performance as successful unconditionally

blindpirate:
As a mitigation for https://github.com/gradle/gradle-private/issues/4975


cobexer:
@blindpirate Performance tests are working again almost all the time, merging this is thus not critical currently.

I leave it as a draft for now in case we need it urgently again until the root cause is understood and resolved.

blindpirate:
Closed as not needed any more.

PR #35402, opened 2025-10-21
Jansi: migrate from 1.18 to 2.4.2 for Windows ARM64 support

cobexer:
Especially picking up:
* https://github.com/fusesource/jansi/pull/240
* https://github.com/fusesource/jansi/pull/292

Relates to discussions on:
* https://github.com/gradle/gradle/issues/29807
* https://github.com/gradle/gradle/issues/29807#issuecomment-3426812702 (& later)

See also:
* https://github.com/gradle/gradle/pull/35415
* https://github.com/gradle/gradle/issues/35419

PR #36017, opened 2025-12-12
Fix CMD shell output by installing jansi `AnsiConsole`

### LINKED ISSUE #35953: Colored terminal detection broken on Windows in 9.3.0-rc-1
### Current Behavior

Image

### Expected Behavior

Image

### Context (optional)

_No response_

Just any project.

### Gradle version

9.3.0-rc-1

9.2.1

### Build scan URL (optional)

_No response_

```
gradlew --version
Picked up JAVA_TOOL_OPTIONS: -Xms32M -Xmx256M -XX:+UseG1GC -D-XX:+PrintCommandLineFlags

Gradle 9.3.0-rc-1

Build time:    2025-11-28 16:38:38 UTC
Revision:      7103de78a66088931f83b1eea19b299af989c114

Kotlin:        2.2.21
Groovy:        4.0.29
Ant:           Apache Ant(TM) version 1.10.15 compiled on August 25 2024
Launcher JVM:  21.0.7 (Oracle Corporation 21.0.7+8-LTS-245)
Daemon JVM:    P:\tools\lang\java-21.0.7-x64-jdk (no Daemon JVM specified, using current Java home)
OS:            Windows 10 10.0 amd64
```

The problem shows up in
* Standalone `cmd` window
* IDEA Terminal view "Command Prompt" (`cmd`)
* IDEA Terminal view "Windows PowerShell"
* Didn't test more, probably more terminals are broken, and/or maybe other OSs too.

When changing the version back to 9.2.1 in the same session (git checkout), it "suddenly" works again.

reinsch82:
This seemingly wasn't necessary in Jansi 1.x (https://github.com/fusesource/jansi/issues/281) despite the doc always claiming it was necessary (https://github.com/fusesource/jansi#example-usage)

Jansi 1.x did eager registration, but with less checks and guard rails involved. There would stll be a deprecated option to eagerly install the streams, but since it is deprecated we don't want to use it.

* Fixes #35953

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/InProcessGradleExecuter.java
+++ b/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/InProcessGradleExecuter.java
@@ -16,7 +16,6 @@
 
 package org.gradle.integtests.fixtures.executer;
 
-import com.google.common.collect.ImmutableList;
 import org.apache.commons.io.output.TeeOutputStream;
 import org.gradle.BuildResult;
 import org.gradle.StartParameter;
@@ -290,12 +289,7 @@ public class InProcessGradleExecuter extends DaemonGradleExecuter {
             builder.setWorkingDir(getWorkingDir());
             builder.setExecutable(new File(getJavaHomeLocation(), "bin/java"));
             builder.classpath(getExecHandleFactoryClasspath());
-            List<String> launcherJvmArgs = invocation.launcherJvmArgs;
-            ImmutableList.Builder<Object> launchArgsBuilder = ImmutableList.builderWithExpectedSize(invocation.launcherJvmArgs.size() + 1);
-            launchArgsBuilder.addAll(launcherJvmArgs);
-
-//            launchArgsBuilder.add("-D" + AnsiConsole.JANSI_MODE + "=" + AnsiConsole.JANSI_MODE_FORCE);
-            builder.jvmArgs(launchArgsBuilder.build());
+            builder.jvmArgs(invocation.launcherJvmArgs);
             // Apply the agent to the newly created daemon. The feature flag decides if it is going to be used.
             for (File agent : GLOBAL_SERVICES.get(ModuleRegistry.class).getModule(AgentUtils.AGENT_MODULE_NAME).getClasspath().getAsFiles()) {
                 builder.jvmArgs("-javaagent:" + agent.getAbsolutePath());
--- a/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/NoDaemonGradleExecuter.java
+++ b/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/NoDaemonGradleExecuter.java
@@ -30,6 +30,7 @@ import org.gradle.process.internal.DefaultClientExecHandleBuilder;
 import org.gradle.test.fixtures.file.TestDirectoryProvider;
 import org.gradle.test.fixtures.file.TestFile;
 import org.gradle.testfixtures.internal.NativeServicesTestFixture;
+import org.gradle.util.DebugUtil;
 import org.gradle.util.GradleVersion;
 
 import java.io.File;
@@ -55,6 +56,11 @@ public class NoDaemonGradleExecuter extends AbstractGradleExecuter {
 
     public NoDaemonGradleExecuter(GradleDistribution distribution, TestDirectoryProvider testDirectoryProvider, GradleVersion gradleVersion, IntegrationTestBuildContext buildContext) {
         super(distribution, testDirectoryProvider, gradleVersion, buildContext);
+    }
+
+    @Override
+    protected boolean isDebuggerAttached() {
+        return DebugUtil.isDebuggerAttached();
     }
 
     @Override
--- a/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/DaemonGradleExecuter.java
+++ b/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/DaemonGradleExecuter.java
@@ -17,7 +17,6 @@ package org.gradle.integtests.fixtures.executer;
 
 import org.gradle.internal.nativeintegration.services.NativeServices;
 import org.gradle.test.fixtures.file.TestDirectoryProvider;
-import org.gradle.util.DebugUtil;
 import org.gradle.util.GradleVersion;
 
 import static org.gradle.integtests.fixtures.executer.AbstractGradleExecuter.CliDaemonArgument.NO_DAEMON;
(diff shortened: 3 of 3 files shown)


eskatos (approved):
Tested live on a Windows machine, confirmed it fixes the regression

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/sink/ConsoleConfigureAction.java
```diff
@@ -122,6 +122,7 @@ private static void configureRichConsole(OutputEventRenderer renderer, ConsoleMe
     }
 
     private static Console consoleFor(OutputStream stream, Supplier<OutputStream> jansiFallback, ConsoleMetaData consoleMetaData, ColorMap colourMap) {
+        org.fusesource.jansi.AnsiConsole.systemInstall();
```
big-guy: I'm not sure about this because I think this gets called twice (once for stdout and once for stderr). systemInstall does nothing if you call it twice, but it also modifies System.setOut/setErr.

If you just want to make sure the `install()` is called on the underlying print streams, maybe we can call that ourselves?

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/InProcessGradleExecuter.java
```diff
@@ -289,7 +290,11 @@ protected Factory<BaseExecHandleBuilder> getExecHandleFactory() {
             builder.setWorkingDir(getWorkingDir());
             builder.setExecutable(new File(getJavaHomeLocation(), "bin/java"));
             builder.classpath(getExecHandleFactoryClasspath());
-            builder.jvmArgs(invocation.launcherJvmArgs);
+            List<String> launcherJvmArgs = invocation.launcherJvmArgs;
+            ImmutableList.Builder<Object> launchArgsBuilder = ImmutableList.builderWithExpectedSize(invocation.launcherJvmArgs.size() + 1);
+            launchArgsBuilder.addAll(launcherJvmArgs);
+            launchArgsBuilder.add("-Djansi.mode=force");
```
reinsch82: force jansi to use ansi console in embedded execution.

this makes embedded "console output" behave the same as in external execution.

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/sink/ConsoleConfigureAction.java
```diff
@@ -63,16 +63,19 @@ private static ConsoleMetaData getConsoleMetaData() {
     }
 
     private static void configureAutoConsole(OutputEventRenderer renderer, ConsoleMetaData consoleMetaData, OutputStream stdout, OutputStream stderr) {
-        if (consoleMetaData.isStdOut() && consoleMetaData.isStdErr()) {
+        if (consoleMetaData.isStdOutATerminal() || consoleMetaData.isStdErrATerminal()) {
+            org.fusesource.jansi.AnsiConsole.systemInstall();
```
reinsch82: as partially suggested by @big-guy only call `systemInstall()` once.

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/sink/ConsoleConfigureAction.java
reinsch82: I don't think I understand what we gain by calling it ourselves.
But this only gets called and used in the launcher, to my understanding. The launcher goes away anyway, so I don't think it matters how high the `installed` count is at the end.

reinsch82:
@bot-gradle test RFR

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/sink/ConsoleConfigureAction.java
big-guy: ❌ I don't think we should be doing this at all.

My concern was more about the calls to setErr and setOut that systemInstall does. We should not be calling `systemInstall` because the only thing it does that we weren't doing before is calling setErr/setOut directly.  AnsiConsole::err and AnsiConsole::out below call `initStreams`, so I don't think there's anything else that systemInstall does that we weren't already doing. 

The docs that say you should call systemInstall are just talking about the easy way to enable this where jansi is responsible for System.out/System.err. Inside Gradle, our logging manager is responsible for this setting these.

IIUC, calling systemInstall all the time is just masking another problem by letting jansi write directly to stdout/stderr. Christophe alludes to some differences here: https://github.com/gradle/gradle/pull/35402/files#r2452421805

Have we reproduced this on our Windows agents?

I don't immediately know what's wrong here, but my suspicion is that we're not wrapping the same outputstream as we were before. e.g., we replaced AnsiConsoleUtil.wrapOutputStream with `org.fusesource.jansi.AnsiConsole#out` and I'm not 100% sure these are equivalent.

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle  test RFN

reinsch82:
@bot-gradle test RFN

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/source/TeePrintStream.java
```diff
@@ -0,0 +1,349 @@
...
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    @Override
+    protected void setError() {
+        throw new UnsupportedOperationException("setError");
```
reinsch82: 🤔 I assume this might be a behavior change?! How bad is it?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/NoDaemonGradleExecuter.java
```diff
@@ -178,6 +184,7 @@ public File getWorkingDir() {
                 builder.environment("JAVA_HOME", "");
                 builder.environment("GRADLE_OPTS", "");
                 builder.environment("JAVA_OPTS", "");
+//                builder.environment("JAVA_OPTS", "-D" + AnsiConsole.JANSI_MODE + "=" + AnsiConsole.JANSI_MODE_FORCE);
```
reinsch82: I will remove this after tests are run

big-guy (approved):

PR #36020, opened 2025-12-12
Refactor declarative test fixtures to be less error-prone

ghale:
This refactors some of the declarative gradle test fixtures to be more composable and less error-prone.  

Specifically, it does the following:

- Moves the display of property values to be part of the definition builder so that definitions are not required to conform to any predetermined structure (i.e. each definition builder controls both the properties it exposes as well as how those properties are displayed)
- Moves the mapping of definition to build model to be part of the the definition builder (i.e. each definition can describe how its properties should be mapped to the build model)
- Adds display of model values by default (also declared by the definition builder)
- Simplifies handling of definitions with both public and implementation types
- Makes all definition-specific values in the plugin builder to be derived from a definition object provided at construction rather than needing to be set separately/additionally.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

PR #36022, opened 2025-12-12
Improve performance of dependency resolution involving strict version constraints

bamboo:
By replacing expensive `ImmutableSet` copying operations with `PersistentSet` updates in `StrictVersionConstraints`.

This change saves `1s` of configuration time (~7%) in the `gradle/gradle` build running `codeQuality`: 

strict-version-constraints-benchmark

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #36026, opened 2025-12-13
Fix exclude simplification logic

### LINKED ISSUE #32945: Fix exclude simplification logic
### Current Behavior

Discovered by error prone, the following PR had a major flaw: part of the exclude information is lost:
* #13825

### Expected Behavior

The flattened `anyOf` should be added to the list, not discarded

### Context (optional)

_No response_

Have a private one, from a Slack conversation

### Gradle version

6.6+

### Build scan URL (optional)

_No response_

_No response_

bamboo:
The previous `ImmutableSet` based implementation caused incorrect simplification.

We now leverage `PersistentSet` semantics to properly compute the fixed-point of the union simplification within the `exclude` algebra.

Fixes #32945

We benchmarked this code against the original version in the [exclude-merging performance comparison scenario](https://github.com/gradle/performance-comparisons/tree/master/exclude-merging) and it showed no regression but a slight improvement:

exclude

The 1st and 3rd rows are from the original version while the 2nd and 4th show the new version. Only the 3rd and 4th runs are charted.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
@@ -211,26 +211,28 @@ public class NormalizingExcludeFactory extends DelegatingExcludeFactory {
      * @param specs the initial set of elements to simplify
      * @return the fixed point set where another full simplification pass yields the same set
      */
-    @SuppressWarnings("ConstantValue") // wrongly accused by the IDE
     private static PersistentSet<ExcludeSpec> fixedPointOf(Simplification function, PersistentSet<ExcludeSpec> specs) {
-        iteration:
-        while (specs.size() > 1) {
-            PersistentSet<ExcludeSpec> original = specs;
-
-            for (ExcludeSpec left : original) {
-                for (ExcludeSpec right : original) {
-                    if (left == right) {
-                        continue;
-                    }
-                    specs = function.apply(left, right, original);
-                    if (specs != original) {
-                        // Restart iteration with new specs
-                        continue iteration;
-                    }
-                }
-            }
-            if (original == specs) {
+        PersistentSet<ExcludeSpec> current = specs;
+        while (current.size() > 1) {
+            PersistentSet<ExcludeSpec> simplified = simplifyOnce(function, current);
+            if (simplified == current) {
                 break;
+            }
+            current = simplified;
+        }
+        return current;
+    }
+
+    private static PersistentSet<ExcludeSpec> simplifyOnce(Simplification function, PersistentSet<ExcludeSpec> specs) {
+        for (ExcludeSpec left : specs) {
+            for (ExcludeSpec right : specs) {
+                if (left == right) {
+                    continue;
+                }
+                PersistentSet<ExcludeSpec> simplified = function.apply(left, right, specs);
+                if (simplified != specs) {
+                    return simplified;
+                }
             }
         }
         return specs;


bamboo:
@bot-gradle test RFN

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
```diff
@@ -137,137 +107,163 @@ private ExcludeSpec doUnion(Set<ExcludeSpec> specs) {
...
+        PersistentSet<ExcludeSpec> elements = PersistentSet.<ExcludeSpec>of()
+            .union(moduleIdExcludes)
+            .union(groupExcludes)
+            .union(moduleExcludes)
+            .union(moduleIdSetsExcludes)
+            .union(groupSetExcludes)
+            .union(moduleSetExcludes)
+            .union(other);
+
+        elements = fixedPointOf(this::simplifyUnion, elements);
```
bamboo, line 173: ℹ️ This is the fix.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
```diff
@@ -288,80 +281,66 @@ private <T extends ExcludeSpec> FlattenOperationResult flatten(Class<T> flattenT
...
-                                asArray[j] = null;
-                                simplified = true;
-                            }
-                        }
-                    }
-                }
-            }
-            if (simplified) {
-                result = Arrays.stream(asArray).filter(Objects::nonNull).collect(toSet());
+        result = fixedPointOf(this::simplifyIntersect, result);
```
bamboo, line 323: ℹ️ This is new behavior that seems to make sense for symmetry with the behavior for unions.

bamboo:
This is a follow-up to #36022 and should be rebased after that one is accepted (or not).

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
bamboo: @ljacomet Please check whether this makes sense since I added this after our session together.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/CachingExcludeFactory.java
```diff
@@ -50,13 +50,13 @@ public ExcludeSpec allOf(ExcludeSpec one, ExcludeSpec two) {
     }
 
     @Override
-    public ExcludeSpec anyOf(Set<ExcludeSpec> specs) {
-        return caches.getAnyOf(new ExcludesKey(specs), key -> delegate.anyOf(key.specs));
+    public ExcludeSpec anyOf(PersistentSet<ExcludeSpec> specs) {
+        return caches.getAnyOf(specs, delegate::anyOf);
```
Pankraz76, line 54: nice, assuming this would be covered by prone as well. 

- https://errorprone.info/bugpattern/UnnecessaryLambda

Pankraz76: actually seems not, as related to new API usage.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/simple/DefaultExcludeFactory.java
```diff
@@ -60,22 +58,22 @@ public ModuleIdExclude moduleId(ModuleIdentifier id) {
...
     public ExcludeSpec allOf(ExcludeSpec one, ExcludeSpec two) {
-        return DefaultExcludeAllOf.of(ImmutableSet.of(one, two));
+        return DefaultExcludeAllOf.of(PersistentSet.of(one, two));
     }
 
     @Override
-    public ExcludeSpec anyOf(Set<ExcludeSpec> specs) {
-        return DefaultExcludeAnyOf.of(ImmutableSet.copyOf(specs));
+    public ExcludeSpec anyOf(PersistentSet<ExcludeSpec> specs) {
+        return DefaultExcludeAnyOf.of(PersistentSet.copyOf(specs));
```
jvandort: copyOf should not be necessary here and below

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
```diff
@@ -815,38 +810,34 @@ private ExcludeSpec joinEdgeAndNodeExclusionsThenCacheResult(ExcludeSpec nodeExc
...
     }
 
     @Nullable
-    private ExcludeSpec joinNodeExclusions(@Nullable ExcludeSpec nodeExclusions, @Nullable Set<ExcludeSpec> excludedByEither) {
-        if (excludedByEither != null) {
+    private ExcludeSpec joinNodeExclusions(@Nullable ExcludeSpec nodeExclusions, PersistentSet<ExcludeSpec> excludedByEither) {
+        if (excludedByEither.isNotEmpty()) {
             if (nodeExclusions != null) {
-                excludedByEither.add(nodeExclusions);
+                excludedByEither = excludedByEither.plus(nodeExclusions);
```
jvandort: Both this method and the one below previously mutated their input parameters (`excludedByEither`, and `excludedByBoth` respectively). Now, they new implementations do not. 

It might make sense to just inline this method and the one below since they both effectively have two output values.

platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/simple/DefaultCompositeExcludeTest.groovy
```diff
@@ -117,23 +118,23 @@ class DefaultCompositeExcludeTest extends Specification {
     }
 
     ModuleSetExclude nextModuleSet() {
-        FACTORY.moduleSet((0..(1 + random.nextInt(5))).collect { randomModuleName() } as Set<String>)
+        FACTORY.moduleSet(PersistentSet.copyOf((0..(1 + random.nextInt(5))).collect { randomModuleName() } as Set<String>))
```
jvandort: I don't think the `as Set...` is necessary on these lines anymore

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
```diff
@@ -137,137 +107,163 @@ private ExcludeSpec doUnion(Set<ExcludeSpec> specs) {
...
+
+            for (ExcludeSpec left : original) {
+                for (ExcludeSpec right : original) {
+                    if (left == right) {
+                        continue;
+                    }
+                    specs = function.apply(left, right, original);
+                    if (specs != original) {
+                        // Restart iteration with new specs
+                        continue iteration;
```
jvandort: Wow. TIL you can use labels along with continue statements in Java. I have no idea how I haven't come across this yet.

In fact, it seems this pattern is generally avoided in the Gradle codebase. Maybe my regex skills aren't up to par, but I can't find a single other instance of labels in any of our Java code.

Granted I haven't found this in any of our ADRs or style guides, but I've generally accepted as an unwritten rule that GOTOs/labels are worth avoiding over more traditional control flow. Can we refactor this into separate helper method to avoid the label/continue?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
```diff
@@ -137,137 +107,163 @@ private ExcludeSpec doUnion(Set<ExcludeSpec> specs) {
...
+
+    /**
+     * Computes the fixed point of a simplification over a set of {@link ExcludeSpec} elements.
+     * <p>
+     * The supplied {@code function} takes two candidate elements and the current set of
+     * elements, and may return either the unchanged set (no simplification possible) or a
+     * new set where a simplification has been applied (for example, by replacing both inputs
+     * with a simplified/merged element, or by removing redundant elements).
+     * </p>
+     * <p>
```
jvandort: The style guide says we should be delimiting paragraphs with a single `<p>` and avoiding the additional `</p>` https://github.com/gradle/gradle/blob/f7e3ecef852528afb9b08923be90cb4dd463e10b/contributing/JavadocStyleGuide.md?plain=1#L66

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
bamboo: > Can we refactor this into separate helper method to avoid the label/continue?

Do you find the code unclear as currently written?

bamboo:
> should be rebased after that one is accepted

Done.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
bamboo: Interesting. In both cases the mutated argument was not used after the method call.

jvandort: Ah, not important then. I missed that these method calls were not in the loop

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
jvandort: With three nested loops and multiple control flow statements jumping between them, I find this method relatively complex to parse compared to others.

I felt it strange enough to comment on it and had to go google how label-based continues worked.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/simple/DefaultExcludeFactory.java
bamboo: Right!

platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/simple/DefaultCompositeExcludeTest.groovy
bamboo: Indeed!

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
bamboo: > I felt it strange enough to comment on it and had to go google how label-based continues worked.

I understand the lack of familiarity but now that you learned about labels, do you still find it unclear?

I tried it with a nested method call and I actually found it harder to understand with the required jump.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
bamboo: Fixed.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
bamboo: In any case, I rewrote the methods to avoid mutating the parameters. It's clearer that way.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
jvandort: Yeah and I'm not trying to be difficult here. I find this method complicated. 

Something like this makes it much more clear to me that there are two separate operation here:

```
    private static PersistentSet<ExcludeSpec> fixedPointOf(Simplification function, PersistentSet<ExcludeSpec> specs) {
        PersistentSet<ExcludeSpec> current = specs;
        while (current.size() > 1) {
            PersistentSet<ExcludeSpec> simplified = trySimplify(function, current);
            if (simplified != null) {
                current = simplified;
            } else {
                break;
            }
        }
        return current;
    }

    private static @Nullable PersistentSet<ExcludeSpec> trySimplify(Simplification function, PersistentSet<ExcludeSpec> specs) {
        for (ExcludeSpec left : specs) {
            for (ExcludeSpec right : specs) {
                if (left == right) {
                    continue;
                }
                PersistentSet<ExcludeSpec> result = function.apply(left, right, specs);
                if (result != specs) {
                    return result;
                }
            }
        }

        return null;
    }
```

The inner operation takes a simplification function, and applies it to pairs of specs, returning a simplified result if it can find one. 

The outer function iteratively calls the inner until no simplification can be found. 

What about this version do you find more difficult to understand?

bamboo: > What about this version do you find more difficult to understand?

For me it's just a little more work to understand the whole algorithm in its full glory due to the indirection.

It's ok though.

bamboo: The `null` is uncalled for though.

bamboo: Done.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/OptimizingExcludeFactory.java
```diff
@@ -40,18 +40,19 @@ public ExcludeSpec allOf(ExcludeSpec one, ExcludeSpec two) {
...
     @Override
-    public ExcludeSpec anyOf(Set<ExcludeSpec> specs) {
-        return Optimizations.optimizeCollection(this, specs, list -> {
-            if (list.size() == 2) {
-                Iterator<ExcludeSpec> it = list.iterator();
+    public ExcludeSpec anyOf(PersistentSet<ExcludeSpec> specs) {
+        return Optimizations.optimizeCollection(this, specs, set -> {
+            if (set.size() == 2) {
+                Iterator<ExcludeSpec> it = set.iterator();
+                // TODO: Return anyOf(set) to preserve the original set when it fails to optimize
```
ljacomet, line 47: 🤔  Any reason to not address this TODO now?

AFAICT this whole `if` is no longer needed given the properties of a `PersistentSet`.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/NormalizingExcludeFactory.java
ljacomet: This makes sense indeed, thanks for extracting the commonality.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/excludes/factories/OptimizingExcludeFactory.java
bamboo: IIRC the way I understood the code, it is giving `delegate` a chance to optimize `anyOf(x, y)`.

The problem is that when `delegate` fails to find a suitable optimization it will wastefully recreate a `anyOf(PersistentSet.of(x, y))` when that was already the original representation.

I added the TODO because I thought that required a deeper refactoring into the delegating structure.

bamboo:
Thanks, @ljacomet!

blindpirate:
Have to revert because of https://github.com/gradle/gradle-private/issues/4996

PR #36036, opened 2025-12-15
Support for definitions with no build model

### LINKED ISSUE #35869: DCL: bindProjectFeature does not support BuildModel.NONE for the Feature Definition
### Current Behavior

Error: `Cannot determine build model type for interface KfxDefinition`

### Expected Behavior

No error

### Context (optional)

ModelTypeUtils throws an exception if ownBuildModel is null, but DefaultTypeParameterInspection returns null `if (parametersType == noParamsType) { return null; }`

```java
Class<OwnBuildModel> ownBuildModel = inspection.parameterTypeFor(definition);
if (ownBuildModel == null) {
    throw new IllegalArgumentException("Cannot determine build model type for " + definition);
}
```

```kotlin
@BindsProjectFeature(KfxProjectFeaturePlugin.Binding::class)
class KfxProjectFeaturePlugin : Plugin<Project> {
    override fun apply(target: Project) {}

    class Binding : ProjectFeatureBinding {
        override fun bind(builder: ProjectFeatureBindingBuilder) {
            builder.bindProjectFeature(
                "kfx",
            ) { kfxDefinition: KfxDefinition, _: BuildModel.NONE, projectDefinition: ToolDefinitions ->
            }
        }
    }
}
interface KfxDefinition : Definition<BuildModel.NONE>
```

### Gradle version

9.3.0-rc-1

### Build scan URL (optional)

_No response_

_No response_

ghale:
This allows project features to be registerd with no build model (i.e. with `BuildModel.None`).  Other features can still bind to the feature's definition.  Any feature that attempts to bind to a build model of `None` will get an error.

Fixes #35869 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/internal/declarativedsl/settings/ProjectFeatureFixture.groovy
+++ b/platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/internal/declarativedsl/settings/ProjectFeatureFixture.groovy
@@ -306,6 +306,17 @@ trait ProjectFeatureFixture extends ProjectTypeFixture {
         return withProjectFeature(projectTypeDefinition, projectType, projectFeatureDefinition, projectFeature, settingsBuilder)
     }
 
+    PluginBuilder withKotlinProjectFeaturePluginsThatHasNoBuildModel() {
+        def projectTypeDefinition = new ProjectTypeDefinitionClassBuilder()
+        def projectType = new ProjectTypePluginClassBuilder(projectTypeDefinition)
+        def projectFeatureDefinition = new ProjectFeatureDefinitionWithNoBuildModelClassBuilder()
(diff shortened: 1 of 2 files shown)


ghale:
@bot-gradle test this

ghale:
> Can you please also add a Kotlin integration test using the reified generics method with BuildModel.None? Will it work out if the box 

@hfhbd Yes, this should work just fine.  I've added a test that demonstrates it.  We use a fixture to generate the class, so it's kind of hard to see what it looks like just looking at the code, but it generates a binding class that looks like this:
```
class Binding : ProjectFeatureBinding {
    override fun bind(builder: ProjectFeatureBindingBuilder) {
        builder.bindProjectFeature<
                FeatureDefinition,
                TestProjectTypeDefinition,
                BuildModel.None
        >("feature") { definition, model, parent ->
            ...
        }
    }
}
```

PR #36040, opened 2025-12-15
Polish benchmarks for `collections` module

bamboo:
Add explaining comments.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/collections/build.gradle.kts
+++ b/platforms/core-runtime/collections/build.gradle.kts
@@ -33,10 +33,14 @@ dependencies {
 
     implementation(projects.stdlibJavaExtensions)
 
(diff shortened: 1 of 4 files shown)


jvandort:
Thinking about it now I wonder if it's actually necessary to comment all this out.

In practice these dependencies should not be resolved unless the user tries to run that require the configurations files.

Do we end up trying to compile/check the jmh source sets during sanityCheck or something?

bamboo:
> Do we end up trying to compile/check the jmh source sets during sanityCheck or something?

Yes. `codeQuality` does it

PR #36043, opened 2025-12-16
DCL Schema Definition API Part I

h0tk3y:


h0tk3y:
@bot-gradle test RFN

h0tk3y:
@bot-gradle test RFN

h0tk3y:
@bot-gradle test RFN

h0tk3y:
@bot-gradle test RFN

h0tk3y:
@bot-gradle test RFN

platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/internal/declarativedsl/settings/ProjectFeatureFixture.groovy
```diff
@@ -689,24 +688,21 @@ trait ProjectFeatureFixture extends ProjectTypeFixture {
...
-                    @${Restricted.class.simpleName}
                     Property<String> getText();
 
                     ${getMaybeInjectedServiceDeclaration()}
 
                     @Nested
                     Fizz getFizz();
 
-                    @${Configuring.class.simpleName}
+                    @${HiddenInDefinition.class.simpleName}
```
ghale, line 705: ❓ Is this because we handle `fizz { }` implicitly now?  Is the implication that to have an action-taking method on the definition object, it _must_ be `HiddenInDefinition`?

platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/internal/declarativedsl/settings/ProjectFeatureFixture.groovy
```diff
@@ -870,24 +859,19 @@ trait ProjectFeatureFixture extends ProjectTypeFixture {
...
                 public abstract class ${publicTypeClassName} implements ${Definition.class.simpleName}<${publicTypeClassName}.FeatureModel> {
-                    @${Restricted.class.simpleName}
                     public abstract Property<String> getText();
 
                     ${maybeInjectedServiceDeclaration}
 
                     @Nested
                     abstract Fizz getFizz();
 
-                    @${Configuring.class.simpleName}
```
ghale, line 890: ❓ Is `HiddenInDefinition` not required here?

platforms/core-runtime/base-services/src/main/java/org/gradle/api/Named.java
```diff
@@ -15,9 +15,12 @@
  */
 package org.gradle.api;
 
+import org.gradle.declarative.dsl.model.annotations.HiddenInDefinition;
+
 /**
  * Types can implement this interface and use the embedded {@link Namer} implementation, to satisfy API that calls for a namer.
  */
+@HiddenInDefinition
```
ghale, line 23: 🤔  It seems possible you might want to extend interfaces that are not Gradle-aware (and not marked hidden).  If so, you're only option would be to mark the extending type as `HiddenInDefinition` and then mark every property visible, which feels really awkward, or override the methods from the parent and mark those hidden, which could be just as awkward.  I wonder if we should have something better you could do on the extending class, for instance, something like:
```
@HiddenInDefinition(Named.class) // Hide all elements coming from Named, but only those from Named
interface Foo extends Named {
    ...
}
```

ghale (approved):
It feels like the hidden stuff might still have some room for improvement, but I think this is probably good as it is and we can continue polishing it in follow up PRs.

platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/internal/declarativedsl/settings/ProjectFeatureFixture.groovy
h0tk3y: Yes, the schema builder now interprets a getter `getFizz()` of a non-trivial type as a nested model. It will generate a function `fizz { }` from the getter.

So, on the one hand, there is no need to define an explicit `fizz { }` on the side, but on the other hand, it is still needed for Kotlin DSL to work, as the schema builder's synthetic configuring function is not yet exposed to Kotlin DSL.

However, if you define both the getter and the configuring function manually in the source, the schema will be ambiguous, and DCL will fail to resolve `fizz { }`. Hence `@HiddenInDefinition` on the method.

platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/internal/declarativedsl/settings/ProjectFeatureFixture.groovy
h0tk3y: Ah, it was not required because `getFizz` was non-public.

UPD: fixed in another commit.

platforms/core-runtime/base-services/src/main/java/org/gradle/api/Named.java
h0tk3y: Yep, I was thinking about something like that.
I think we can even add support for this use case with a separate annotation, so we don't add more responsibility to `HiddenInDefinition`.
And one more workaround for now is to introduce an intermediate supertype that inherits from the type you want to hide and mark the intermediate type as hidden.

platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/internal/declarativedsl/settings/ProjectFeatureFixture.groovy
ghale: I was thinking you might have build logic where you want to call an action taking method.  But I guess if we didn't need this for Kotlin, it would probably be an uncommon case that you would need this for anything else.

PR #36044, opened 2025-12-16
Deny repositories on unknown host exception

ljacomet:
Also adjust test to clarify when Gradle denies and when it does not.


ljacomet:
@bot-gradle test RfR

ljacomet:
Replaced by #36059

PR #36045, opened 2025-12-16
Minify jnr constants to reduce distribution size by ~1Mb

ov7a:
Also, refactor the minify action to avoid referencing exact versions.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs-asciidoctor-extensions-base/build.gradle.kts
+++ b/platforms/documentation/docs-asciidoctor-extensions-base/build.gradle.kts
@@ -1,5 +1,3 @@
-import gradlebuild.basics.classanalysis.Attributes.minified
-
 plugins {
     id("gradlebuild.internal.java")
     groovy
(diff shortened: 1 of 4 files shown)


ov7a:
@bot-gradle test this

platforms/documentation/docs-asciidoctor-extensions-base/build.gradle.kts
```diff
@@ -16,6 +18,18 @@ dependencies {
     testImplementation("org.spockframework:spock-core")
 }
 
+// ascii-doctor depends on JRuby, which has a dependency on jnr-constants, unit tests fail because of that, but this shouldn't happen when it's used as a plugin due to classloader isolation
+configurations.all {
```
ov7a: I'm not sure if it's the correct approach. Please advise.

ov7a:
@bot-gradle test RFN

build-logic-commons/basics/src/main/kotlin/gradlebuild.minify.gradle.kts
```diff
@@ -23,9 +23,12 @@ import gradlebuild.basics.transforms.Minify
  * classes and the classes they depend on. The classes are not relocated, they all
  * remain in their original namespace. This reduces the final Gradle distribution
  * size and makes us more conscious of which parts of a library we really need.
+ *
+ * WARNING: these changes might break things in subtle ways.
+ * Existing tests might not cover all use cases.
```
blindpirate: This comment doesn't add much value. What shall we do to verify the change if current tests are not enough? Build some local distribution and run some local builds?

ov7a: My intent here was to highlight that relying on tests is not enough, since some of the classes might be constructed via reflection, and some can be accessed in scenarios/platforms that are not covered by tests. Manual class filtering requires either a full understanding of how the library works and which classes are used or extensive test coverage.

Does this clarify things? If yes, what part should I add to the comment?

ov7a:
@bot-gradle test RFN

PR #36054, opened 2025-12-17
Make virtual outgoing edges and declared outgoing edges more similar

jvandort:
Removes `PendingDependenciesVisitor`, inlining its implementation. Then update virtual outgoing edges to reuse the inlined implementation, so that virtual edges and normal edges can be treated mostly the same.

Some minor other changes:
* We mark an edge unattached immediately at the same time we compute its selector and adjust module hard edges count
* Allow `LenientPlatformDependencyMetadata` to act as a hard edge. Edges to the virtual platform were previously constraints, but are now hard edges. This lets us get rid of some special-case logic.
* potentiallyActivatedConstraints now holds entire EdgeState instances. We only "link" that state if it is used. 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test this

PR #36055, opened 2025-12-17
Profile SQL executions

blindpirate:
Closes https://github.com/gradle/gradle-private/issues/4977#event-21563433464

Performance DB queries might be inefficient but we have no metrics to track them. This PR does:

1. Print profiling data around every SQL operation with the format `[Profiling] {name=aaa, sql=bbb, params=[], result=ddd, durationMs=123}`. The reason we use JSON is that SQL include many commas and it would make CSV output hard to read.
2. When we run performance report generation task, it captures the lines of stdout, strips the `[Profiling] ` prefix, and writes them into `profiling.txt`. It's part of performance report dir so it would be automatically collected by TeamCity artifact.

The [artifact publish mode is also set to "always"](https://www.jetbrains.com/help/teamcity/2025.07/configuring-general-settings.html?Configuring%20General%20Settings#General+Build+Configuration+Settings) so that the artifact will be uploaded even when build runs into timeout.

Example of the profiling data can be found [here](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_PerformanceTestTestLinux_Trigger/106728180?buildTab=artifacts#%2Fperformance-test-results.zip;%2Fperformance-test-results.zip!%2Freport).

### DIFF
--- a/.teamcity/src/main/kotlin/configurations/PerformanceTestsPass.kt
+++ b/.teamcity/src/main/kotlin/configurations/PerformanceTestsPass.kt
@@ -19,6 +19,7 @@ package configurations
 import common.Os
 import common.applyDefaultSettings
 import jetbrains.buildServer.configs.kotlin.ParameterDisplay
+import jetbrains.buildServer.configs.kotlin.PublishMode
 import jetbrains.buildServer.configs.kotlin.ReuseBuilds
 import model.CIBuildModel
 import model.PerformanceTestType
@@ -66,6 +67,7 @@ class PerformanceTestsPass(
                     "performanceTestReport"
                 }
 
+            publishArtifacts = PublishMode.ALWAYS
             artifactRules = """
 testing/$performanceProjectName/build/performance-test-results.zip
 """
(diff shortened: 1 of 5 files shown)


testing/internal-performance-testing/src/main/groovy/org/gradle/performance/fixture/CrossVersionPerformanceTestRunner.groovy
```diff
@@ -194,8 +194,8 @@ class CrossVersionPerformanceTestRunner extends PerformanceTestSpec {
         def builder = GradleBuildExperimentSpec.builder()
             .projectName(testProject)
             .displayName(displayName)
-            .warmUpCount(warmUpRuns)
-            .invocationCount(runs)
+            .warmUpCount(1)
```
cobexer: ❌ We don't keep that, right?

cobexer:
The remainder LGTM

testing/internal-performance-testing/src/main/groovy/org/gradle/performance/fixture/CrossVersionPerformanceTestRunner.groovy
blindpirate: Right. I remember I did comment "this is for testing purpose only and will be removed before merging", but it's somehow gone.

PR #36059, opened 2025-12-17
Disable repositories on unknown host exception

ljacomet:
Also adjust test to clarify when Gradle disables and when it does not.


big-guy (approved):

ljacomet:
Superseded by #36060

PR #36060, opened 2025-12-17
Disable repositories more often and add property to continue when disabled

ljacomet:
This builds on top of 36059

~It is a draft because there is still some work that needs to be done in disabling repos more aggressively.~

Repositories are now also disabled if the max retry is reached for transient connection errors. This improves build reproducibility.
A new property can be configured per repository to indicate if a disabled repository should still allow resolution to continue.

### DIFF
--- a/platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/repositories/DefaultIvyArtifactRepositoryTest.groovy
+++ b/platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/repositories/DefaultIvyArtifactRepositoryTest.groovy
@@ -605,7 +605,7 @@ class DefaultIvyArtifactRepositoryTest extends Specification {
             TestUtil.checksumService, providerFactory, new VersionParser()
         )
         repo.name = 'repo'
-        repo.allowInsecureContinueOnConnectionFailure.convention(false)
+        repo.allowInsecureContinueWhenDisabled.convention(false)
         return repo
     }
 
--- a/platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/repositories/DefaultMavenArtifactRepositoryTest.groovy
+++ b/platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/repositories/DefaultMavenArtifactRepositoryTest.groovy
@@ -360,7 +360,7 @@ class DefaultMavenArtifactRepositoryTest extends Specification {
             Mock(FileResourceRepository), mavenMetadataFactory, SnapshotTestUtil.isolatableFactory(),
             TestUtil.objectFactory(), urlArtifactRepositoryFactory, TestUtil.checksumService, providerFactory, new VersionParser())
         repo.name = 'repo'
-        repo.allowInsecureContinueOnConnectionFailure.convention(false)
+        repo.allowInsecureContinueWhenDisabled.convention(false)
         return repo
     }
 


ljacomet:
@bot-gradle test RfR

ljacomet:
@bot-gradle test RfR

subprojects/core-api/src/main/java/org/gradle/api/artifacts/repositories/UrlArtifactRepository.java
```diff
@@ -73,4 +74,12 @@ public interface UrlArtifactRepository {
...
      */
     void setAllowInsecureProtocol(boolean allowInsecureProtocol);
+
+    /**
+     * Specifies whether to continue downloading artifacts from other repositories if connection to this repository fails.
+     *
+     * @return A Gradle property for the continueOnConnectionFailure flag.
+     * @since 9.3
+     */
+    Property<Boolean> getAllowInsecureContinueOnConnectionFailure();
```
big-guy: Need to add this to UrlArtifactRepository.xml too

subprojects/core-api/src/main/java/org/gradle/api/artifacts/repositories/UrlArtifactRepository.java
```diff
@@ -73,4 +74,12 @@ public interface UrlArtifactRepository {
      * @see #isAllowInsecureProtocol()
      */
     void setAllowInsecureProtocol(boolean allowInsecureProtocol);
+
+    /**
+     * Specifies whether to continue downloading artifacts from other repositories if connection to this repository fails.
+     *
+     * @return A Gradle property for the continueOnConnectionFailure flag.
+     * @since 9.3
```
big-guy: This needs to be 9.3.0

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/RepositoryFailureCollector.java
```diff
@@ -0,0 +1,47 @@
...
+
+    Collection<Throwable> getFailures() {
+        return failures;
+    }
+
+    boolean hasFatalError() {
+        return fatalError;
+    }
+
+    void toggleFatalError() {
```
big-guy: toggle implies that calling this twice would return the state back to the original state. `markFatalError` or `markAsFatal` would be a better name.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/RepositoryDisabler.java
```diff
@@ -40,7 +40,7 @@ public interface RepositoryDisabler {
      * (<strong>Be sure to note the ambiguity in this value</strong>)
      * @implSpec implementations <strong>MUST</strong> return {@code false} if the repository is not disabled by this call
      */
-    boolean tryDisableRepository(String repositoryId, Throwable throwable);
+    boolean tryDisableRepository(String repositoryId, Throwable throwable, boolean b);
```
big-guy: What's `b`?  can you add some docs?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/DynamicVersionResolver.java
```diff
@@ -213,6 +207,15 @@ private RepositoryChainModuleResolution findLatestModule(LinkedList<RepositoryRe
         return best;
     }
 
+    private static void handleFailure(List<RepositoryResolveState> queue, RepositoryResolveState request, RepositoryFailureCollector failures, Exception t) {
+        failures.addFailure(t);
+        if (!request.isContinueOnConnectionFailure() && request.isRepositoryDisabled()) {
```
big-guy: This is a nit, but thinking about these conditions in the opposite order might be better.

e.g., "if (the repository is disabled AND we can't continue to the next one)" vs "if (we can't continue to the next repository AND the repository is disabled)"

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/ErrorHandlingModuleComponentRepository.java
```diff
@@ -57,12 +57,14 @@ public class ErrorHandlingModuleComponentRepository implements ModuleComponentRe
...
+    private final RepositoryDisabler remoteRepositoryDisabler;
     private final ErrorHandlingModuleComponentRepositoryAccess remote;
 
     public ErrorHandlingModuleComponentRepository(ModuleComponentRepository<ExternalModuleComponentGraphResolveState> delegate, RepositoryDisabler remoteRepositoryDisabler) {
         this.delegate = delegate;
-        local = new ErrorHandlingModuleComponentRepositoryAccess(delegate.getLocalAccess(), getId(), RepositoryDisabler.NoOpDisabler.INSTANCE, getName());
-        remote = new ErrorHandlingModuleComponentRepositoryAccess(delegate.getRemoteAccess(), getId(), remoteRepositoryDisabler, getName());
+        this.remoteRepositoryDisabler = remoteRepositoryDisabler;
+        local = new ErrorHandlingModuleComponentRepositoryAccess(delegate.getLocalAccess(), getId(), RepositoryDisabler.NoOpDisabler.INSTANCE, getName(), false);
+        remote = new ErrorHandlingModuleComponentRepositoryAccess(delegate.getRemoteAccess(), getId(), remoteRepositoryDisabler, getName(), isContinueOnConnectionFailure());
```
big-guy: For consistency...
```suggestion
        this.local = new ErrorHandlingModuleComponentRepositoryAccess(delegate.getLocalAccess(), getId(), RepositoryDisabler.NoOpDisabler.INSTANCE, getName(), false);
        this.remote = new ErrorHandlingModuleComponentRepositoryAccess(delegate.getRemoteAccess(), getId(), remoteRepositoryDisabler, getName(), isContinueOnConnectionFailure());
```

subprojects/core-api/src/main/java/org/gradle/api/artifacts/repositories/UrlArtifactRepository.java
```diff
@@ -73,4 +74,14 @@ public interface UrlArtifactRepository {
...
+
+    /**
+     * Specifies whether to continue downloading artifacts from other repositories if connection to this repository gets disabled.
+     * <p>
+     * The default value for this property is {@code false}.
+     *
+     * @return A Gradle property for the allowInsecureContinueWhenDisabled flag.
+     * @since 9.3
+     */
+    Property<Boolean> getAllowInsecureContinueWhenDisabled();
```
big-guy, line 87: This needs to be documented in UrlArtifactRepository.xml too

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/ErrorHandlingModuleComponentRepository.java
```diff
@@ -57,12 +57,14 @@ public class ErrorHandlingModuleComponentRepository implements ModuleComponentRe
...
     private final ErrorHandlingModuleComponentRepositoryAccess local;
+    private final RepositoryDisabler remoteRepositoryDisabler;
     private final ErrorHandlingModuleComponentRepositoryAccess remote;
 
     public ErrorHandlingModuleComponentRepository(ModuleComponentRepository<ExternalModuleComponentGraphResolveState> delegate, RepositoryDisabler remoteRepositoryDisabler) {
         this.delegate = delegate;
-        local = new ErrorHandlingModuleComponentRepositoryAccess(delegate.getLocalAccess(), getId(), RepositoryDisabler.NoOpDisabler.INSTANCE, getName());
-        remote = new ErrorHandlingModuleComponentRepositoryAccess(delegate.getRemoteAccess(), getId(), remoteRepositoryDisabler, getName());
+        this.remoteRepositoryDisabler = remoteRepositoryDisabler;
+        local = new ErrorHandlingModuleComponentRepositoryAccess(delegate.getLocalAccess(), getId(), RepositoryDisabler.NoOpDisabler.INSTANCE, getName(), false);
```
big-guy: Does it make sense to say that the local part never continues (false) or that it always continues (true) or should it match the remote part? I'm not entirely sure myself since the local side can never be disabled.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/DynamicVersionResolver.java
ljacomet: Agreed, did it that way because one is a simple `boolean` while the other is a `map.containsKey`

platforms/documentation/docs/src/docs/dsl/org.gradle.api.artifacts.repositories.UrlArtifactRepository.xml
```diff
@@ -39,6 +39,9 @@
                     <td>Name</td>
                 </tr>
             </thead>
+            <tr>
+                <td>getAllowInsecureContinueWhenDisabled</td>
```
big-guy: This should be allowInsecureContinueWhenDisabled and in the properties section

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/RepositoryChainComponentMetaDataResolver.java
```diff
@@ -157,17 +154,21 @@ private RepositoryChainModuleResolution findBestMatch(List<ComponentMetaDataReso
...
             BuildableModuleComponentMetaDataResolveResult<ExternalModuleComponentGraphResolveState> metaDataResolveResult;
             metaDataResolveResult = request.resolve();
             switch (metaDataResolveResult.getState()) {
                 case Failed:
-                    failures.add(metaDataResolveResult.getFailure());
-                    if (isCriticalFailure(metaDataResolveResult.getFailure())) {
+                    ModuleVersionResolveException failure = metaDataResolveResult.getFailure();
+                    assert failure != null; // Failure cannot be null in Failed state
+                    failures.addFailure(failure);
+                    if (!request.isContinueOnConnectionFailure() && request.isRepositoryDisabled()) {
```
big-guy: Like you did in the other file, does the order make more sense the other way?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/RepositoryDisabler.java
```diff
@@ -56,7 +57,7 @@ public Optional<Throwable> getDisabledReason(String repositoryId) {
         }
 
         @Override
-        public boolean tryDisableRepository(String repositoryId, Throwable throwable) {
+        public boolean tryDisableRepository(String repositoryId, Throwable throwable, boolean b) {
```
big-guy:
```suggestion
        public boolean tryDisableRepository(String repositoryId, Throwable throwable, boolean retriesExceeded) {
```

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -327,6 +327,15 @@ These PGP signatures allow users and build systems to verify that a downloaded G
...
 For details on verifying Gradle distributions and JARs, see the [Signing key for Gradle artifacts](https://gradle.org/keys/) page.
 
+### Dependency repositories can be disabled for more reasons
+
+When Gradle fails to retrieve information from a repository, it disables the repository for the remainder of the build and fails all dependency resolution.
+This behavior ensures reproducibility.
+
+In this release, more failures will cause the repository to be disabled.
+
+See [the documentation](/userguide/graph_resolution.html#sec:repository-disabling) for details, including ways to continue resolution even if a repository is disabled.
```
big-guy:
```suggestion
### Dependency repositories can be disabled and skipped for more reasons

When Gradle fails to retrieve information from a repository after multiple retries or specific fatal errors, the repository will no longer be checked for the remainder of the build. This will usually fail dependency resolution because Gradle will not continue to the next repository in the list unless told otherwise.

This behavior ensures reproducibility.

In this release, more failures will cause the repository to be disabled, such as an incorrect hostname.

See [the documentation](userguide/graph_resolution.html#sec:repository-disabling) for details, including ways to continue resolution even if a repository is disabled.
```

subprojects/core-api/src/main/java/org/gradle/api/artifacts/repositories/UrlArtifactRepository.java
```diff
@@ -73,4 +74,14 @@ public interface UrlArtifactRepository {
...
      */
     void setAllowInsecureProtocol(boolean allowInsecureProtocol);
+
+    /**
+     * Specifies whether to continue downloading artifacts from other repositories if connection to this repository gets disabled.
+     * <p>
+     * The default value for this property is {@code false}.
+     *
+     * @return A Gradle property for the allowInsecureContinueWhenDisabled flag.
+     * @since 9.3.0
```
big-guy:
```suggestion
     * Specifies whether to continue checking other repositories if this repository is disabled due to connection or communication errors.
     * <p>
     * The conventional value for this property is {@code false}, which means to not continue to check other repositories after this one.
     *
     * @return a property that control the behavior to continue or not
     * @since 9.3.0
     * @see <a href="https://docs.gradle.org/current/userguide/graph_resolution.html#sec:repository-disabling">Disabled repositories</a>
```

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/ErrorHandlingModuleComponentRepository.java
```diff
@@ -100,6 +102,16 @@ public InstantiatingAction<ComponentMetadataSupplierDetails> getComponentMetadat
...
     }
 
+    @Override
+    public boolean isContinueOnConnectionFailure() {
+        return delegate.isContinueOnConnectionFailure();
+    }
+
+    @Override
+    public boolean isRepositoryDisabled() {
+        return remoteRepositoryDisabler.isDisabled(getId()) || delegate.isRepositoryDisabled();
```
big-guy, line 112: If I'm following this correctly, `delegate.isRepositoryDisabled` is always false.  The only way for it to not be false is if `remoteRepositoryDisabler.isDisabled` is true. 

It looks like `ExternalResourceResolver` is the only real implementation.  Could it get a `RepositoryDisabler` and have it check isDisabled in `isRepositoryDisabled`?

platforms/documentation/docs/src/docs/release/notes.md
lkasso: +1

lkasso (approved):
Approved once @big-guy's changes are introduced.

platforms/documentation/docs/src/docs/dsl/org.gradle.api.artifacts.repositories.UrlArtifactRepository.xml
ljacomet: That makes sense, will remember!

platforms/documentation/docs/src/docs/release/notes.md
ljacomet: I also copied the first paragraph over in the docs in place of the one that was already there.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/ErrorHandlingModuleComponentRepository.java
ljacomet: I think leaving it as `false` is good for now. But the whole error handling thing should be redefined for the local part. While there are common `IOException`, most of the logic makes little sense for local. And with proper handling, we might be able to do better errors. But let's do that in a different PR.

ljacomet:
@bot-gradle test RfR

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/ErrorHandlingModuleComponentRepository.java
big-guy: I didn't test this at all: https://github.com/gradle/gradle/commit/01c9ac929faf5eecc45b8b1309aeb901363d407a

big-guy:
@bot-gradle test RfR

PR #36065, opened 2025-12-18
Default plugin ID to the name of the declaration

### LINKED ISSUE #35649: Default plugin id to the name of registration container
### Expected Behavior

```
gradlePlugins {
    plugins {
        register("com.foo") {
            implementationClass = "com.foo.FooPlugin"
        }
    }
}
```

This should be enough 

```
gradlePlugins {
    plugins {
        register("com.foo") {
            id = "com.foo"
            implementationClass = "com.foo.FooPlugin"
        }
    }
}
```

### Context

It would be nice not to have to repeat the same string twice.

urdak:
Removes redundancy when registering plugins

Fixes #35649

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs/src/docs/release/notes.md
+++ b/platforms/documentation/docs/src/docs/release/notes.md
@@ -125,6 +125,23 @@ tasks.validatePlugins {
 }
 ```
 
+### Simpler plugin registration
+
+Plugin builds that use the `java-gradle-plugin` can now register each plugin with less ceremony.
+The plugin ID is now set to the registration's name by default:
+
+```kotlin
+gradlePlugin {
+    plugins {
+        register("my.plugin-id") {
+            implementationClass = "my.PluginClass"
+        }
+    }
+}
+```
+
+See the [Java Gradle Plugin](userguide/java_gradle_plugin.html#sec:gradle_plugin_dev_usage) plugin documentation for more information.
+
 ## Tooling integration improvements
 
 This release adds a few enhancements to the built-in Tooling API models:


platforms/extensibility/plugin-development/src/main/java/org/gradle/plugin/devel/PluginDeclaration.java
```diff
@@ -44,9 +45,14 @@ public String getName() {
         return name;
     }
 
+    /**
+     * Returns the plugin ID.
+     * Defaults to the name of this plugin declaration.
+     */
     @ToBeReplacedByLazyProperty
+    @NonNull
     public String getId() {
```
urdak, line 48: Would it make sense to also migrate to the lazy property API in this PR?

hfhbd: That would be a breaking change. IMHO the `id` should be deprecated in Gradle 9 and removed in Gradle 10, I never understood the use-case of `name` and `id`.

ljacomet:

eskatos:
A more future proof approach would be to let the ID have the name value by default.

These javabean properties will be migrated to Gradle's rich properties. Then, the good way to implement this fix would be to let ID have a convention of the value of the name:
```java
extension.getPlugins().configureEach(pluginDeclaration -> 
    pluginDeclaration.getId().convention(pluginDeclaration.getName())
);
```

We can already use that approach, in `JavaGradlePluginPlugin.java`
```java
extension.getPlugins().configureEach(pluginDeclaration ->
    pluginDeclaration.setId(pluginDeclaration.getName())
);
```

And keep the validation that ID is not null, because in any case the ID could be set back to null.

platforms/extensibility/plugin-development/src/main/java/org/gradle/plugin/devel/PluginDeclaration.java
eskatos: The migration to lazy property would be a breaking change by itself. These migrations will be done in a Gradle major.

Removing the ID property (lazy or not) would be another breaking change because there are builds out there that don't use their proper plugin ID as the name.

Let's keep this PR focused on making it possible to omit setting the ID by using the value of the name by default.

urdak:
> And keep the validation that ID is not null, because in any case the ID could be set back to null.

I missed this part. I will add it back and add another test.

eskatos:
@bot-gradle test this

eskatos (approved):

PR #36075, opened 2025-12-18
Fix parallel tests using new fixture

mlopatkin:
The nice API doesn't really work. It turns out that the result of the successful parallel execution, `ParallelExecutionResult`, also implements `ExecutionFailure`, which causes Groovy to select a wrong overload that looks up the URL in the stderr.

The easiest solution is to remove the overloads and use the correct stream explicitly.

Fixes gradle/gradle-private#4979


mlopatkin:
@bot-gradle test RfN

PR #36087, opened 2025-12-19
[Investigation] Use PersistenSet instead of ImmutableSet in DependentsSet

asodja:


asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

PR #36090, opened 2025-12-19
Revert Jenkins plugin version change in DV smoke test

ljacomet:
The version that was used is no longer available. Going back to previous
 release.


guylabs:
Revert here https://github.com/gradle/gradle/pull/36095

PR #36096, opened 2025-12-19
Deprecate Closure-taking methods on Test

big-guy:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


big-guy:
@bot-gradle test this

big-guy:
@bot-gradle test this

big-guy:
@bot-gradle test this

big-guy:
@bot-gradle test this

jvandort (commented):
Code changes lgtm but were missing the upgrade guide adoc sections

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
```diff
@@ -1036,8 +1039,16 @@ public TestFramework getTestFramework() {
...
     }
 
+    /**
+     * Do not call this method.
+     * @deprecated This will be removed in Gradle 10
+     */
+    @Deprecated
     public TestFramework testFramework(@Nullable Closure testFrameworkConfigure) {
-        // TODO: Deprecate and remove this method
+        DeprecationLogger.deprecateMethod(Test.class, "testFramework(Closure)")
```
jvandort, line 1048: Can we direct the user to call the options method in this deprecation?

jvandort (approved):

big-guy:
@bot-gradle cancel

big-guy:
@bot-gradle test this

vlsi:
I see you deprecate methods that were the only API available.
I have Gradle plugin which makes output much more accessible to users and agents: https://github.com/vlsi/vlsi-release-plugins/blob/master/plugins/gradle-extensions-plugin

Now that you deprecate the methods and plan removing them, the plugin will break for the users who did not upgrade.

Hey, this is really sad ecosystem to have. Please consider going with "deprecated level=hidden" instead so previously compiled plugins continue to work.

Frankly speaking, I am sure `afterTest -> TestListener` transition can be handled by Gradle itself, and it could support existing plugins with literally zero effort.

I am sure you have all the resources to implement `@deprecated(level=hidden)` for java code via bytecode preprocessor (= `access_modifier+=synthetic`). It would reduce a lot of "Gradle has broken again" all over the place. It is drop of the old APIs that disturbs the ecosystem.

A concrete ask: can add a time window at least 3-5 years in-between deprecation and drop of the APIs?
Sure there might be cases when the support of old APIs is complicated. I'm fine if they are removed sooner. However, there are a LOT of cases when Gradle just drops methods which cost zero in support.

PR #36098, opened 2025-12-20
Adjust ArchUnit tests to properly verify NullMarked annotations

octylFractal:
I got tired of having to annotate inner classes that don't need to be annotated due to our ArchUnit rules being incorrect. As described [here](https://jspecify.dev/docs/api/org/jspecify/annotations/NullMarked.html), `@NullMarked` applies to all enclosed elements, so inner classes don't need the annotation if the outer class has it.

This PR is broken up into 4 commits:
1. Adjust the check to account for enclosing elements properly.
2. Renames the check separately so that the changes in output from the previous commit are more easily viewable.
3. Adds a new test for unnecessarily annotated classes.
4. Removes all violations of the new rule from the previous commit.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


testing/architecture-test/src/test/java/org/gradle/architecture/test/ApiNullabilityTest.java
```diff
@@ -70,4 +71,12 @@ public class ApiNullabilityTest {
     public static final ArchRule public_api_methods_are_not_null_unmarked = freeze(methods()
         .that(are(public_api_methods))
         .should(not(beNullUnmarkedMethod())));
+
+    @ArchTest
+    public static final ArchRule classes_are_not_unnecessarily_null_marked = freeze(classes()
+        .that(are(inGradlePublicApiPackages()).or(inGradleInternalApiPackages()))
```
jvandort, line 77: To say a class is either in public or in internal API seems it would apply no filter. Is this `then` necessary at all?

octylFractal: Not quite, our internal API excludes `gradleMaintainedExternalDependency()` and `inTestFixturePackages()`, so this still ends up excluding some classes. It might be more optimal to just directly exclude those though, and have a `inGradleAnyApiPackages()`?

jvandort: Guess my assumption was wrong I think this is fine the way it is then

PR #36105, opened 2025-12-21
Let CC offload string encoding to a separate thread

bamboo:
So parallel CC and IP writers can proceed in parallel.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/ParallelStringEncoder.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/ParallelStringEncoder.kt
@@ -19,10 +19,14 @@ package org.gradle.internal.cc.impl.serialize
 import com.esotericsoftware.kryo.io.Output
 import org.gradle.internal.cc.impl.serialize.ParallelStringEncoder.Companion.EMPTY_STRING_ID
 import org.gradle.internal.cc.impl.serialize.ParallelStringEncoder.Companion.NULL_STRING_ID
+import org.gradle.internal.collect.PersistentArray
 import org.gradle.internal.serialize.Encoder
 import org.gradle.internal.serialize.graph.StringEncoder
 import java.io.OutputStream
 import java.util.concurrent.ConcurrentHashMap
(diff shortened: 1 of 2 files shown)


bamboo:
@bot-gradle test ACC

bamboo:
@bot-gradle test APT

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/ParallelStringEncoder.kt
```diff
@@ -46,13 +50,59 @@ class ParallelStringEncoder(stream: OutputStream) : StringEncoder {
...
+                var writing = true
+                while (writing) {
+                    val pending = writeQueue.getAndUpdate(WriteQueue::flush).pending
+                    if (pending.isEmpty) {
+                        Thread.onSpinWait()
+                        continue
+                    }
+                    for (s in pending) {
+                        output.writeString(s)
+                        if (s.isEmpty()) {
```
abstratt, line 92: 💅 

An EOF string constant would make things a little more explicit:
```
if (s === EOF) {
```

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/ParallelStringEncoder.kt
```diff
@@ -46,13 +50,59 @@ class ParallelStringEncoder(stream: OutputStream) : StringEncoder {
...
     private
-    val output = Output(stream)
+    val writeThread = thread(priority = Thread.MIN_PRIORITY) {
+        try {
+            Output(stream).use { output ->
+                var writing = true
+                while (writing) {
+                    val pending = writeQueue.getAndUpdate(WriteQueue::flush).pending
+                    if (pending.isEmpty) {
+                        Thread.onSpinWait()
```
abstratt, line 87: ❓ 

Curious on how this compares to some sort of signal in terms of wastefully  "busy waiting"

PR #36107, opened 2025-12-22
Update timeout for FlakyTestQuarantine builds

blindpirate:
Currently, all flaky test quarantine builds inherit old timeout config, which is 180 min. After we split them, most builds can finish in 30m, with some slow builds (like Windows NoDaemon) need slightly more than 30m to finish. 

This PR changes:

1. Set the default timeout for FlakyTestQuarantine to be 60 min.
2. Set smoke test under FlakyTestQuarantine mode with timeout=30m and parallelism=1.
3. Update all artifact rules publish mode, because we have need to get timeout logs uploaded as artifacts even when the builds run into timeout.

### DIFF
--- a/.teamcity/src/main/kotlin/configurations/BuildDistributions.kt
+++ b/.teamcity/src/main/kotlin/configurations/BuildDistributions.kt
@@ -3,6 +3,7 @@ package configurations
 import common.Os.LINUX
 import common.buildScanTagParam
 import common.getBuildScanCustomValueParam
+import common.setArtifactRules
 import model.CIBuildModel
 import model.Stage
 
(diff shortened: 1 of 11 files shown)


.teamcity/src/main/kotlin/configurations/BuildDistributions.kt
```diff
@@ -32,9 +33,10 @@ class BuildDistributions(
             publishBuildStatusToGithub(model)
         }
 
-        artifactRules =
+        setArtifactRules(
             """$artifactRules
 packaging/distributions-full/build/distributions/*.zip => distributions
 platforms/core-runtime/base-services/build/generated-resources/build-receipt/org/gradle/build-receipt.properties
-"""
+""",
```
blindpirate, line 40: This is formatted by ktlint automatically, which adds trailing comma for mutlti-line method call.

PR #36110, opened 2025-12-22
Enable NullAway in multiple projects

reinsch82:
### Context

Part of the ongoing effort to enable NullAway across the codebase project by project (in dependency-topological order), with one commit per project.

This PR enables NullAway (`errorprone { nullawayEnabled = true }`) in 11 projects:

- `:internal-instrumentation-processor`
- `:service-registry-builder`
- `:java-api-extractor`
- `:snapshots`
- `:wrapper-shared`
- `:wrapper-main`
- `:build-cache-packaging`
- `:normalization-java`
- `:file-watching`
- `:native`
- `:build-cache`

### Approach

Annotations-only — no behavior changes (one deliberate exception below):

- `@Nullable` on genuinely optional fields, parameters, and returns (matching already-annotated interfaces where overrides were missing it).
- `requireNonNull(...)` where a value is non-null in correct usage but NullAway cannot prove it (e.g. set-after-construction fields, map lookups known to be present).
- Targeted `@SuppressWarnings("NullAway")` with a justifying comment only where neither fits.

### Notable points for review

- **Wrapper nullability** (`:wrapper-shared`): `WrapperConfiguration.getDistribution()` stays `@Nullable` — a null distribution is a legitimate, documented state (`WrapperExecutor.forProjectDirectory()` without a wrapper properties file; `DistributionFactory` relies on it to fall back to the current Gradle version). `requireNonNull` moved into the callers that genuinely require a configured distribution (`Install.createDist()`, `PathAssembler.getDistribution()`). The setter accepts null to satisfy the accessor-symmetry rule in `KotlinCompatibilityTest`.
- **Wrapper JAR hash**: `EXPECTED_WRAPPER_JAR_HASH` in `WrapperGenerationIntegrationTest` is updated — the `@Nullable` annotations added to wrapper classes intentionally change the `gradle-wrapper.jar` bytecode.
- **NullAway #681 cleanup**: NullAway now infers that `AtomicReference` holds a non-null type (uber/NullAway#681), so the pre-existing workarounds (`requireNonNull` calls and suppressions) are removed in `DefaultBuildCacheController`, `BaseLocalBuildCacheServiceHandle`, `FileWatchingFilter`, `DefaultBuildOperationListenerManager`, `DefaultWorkerLeaseService`, `DefaultServiceRegistry` (×2), and `InstrumentedInputs`.
- **New `@Initializer` annotation** (`:java-api-extractor`): `ApiMemberSelector.classMember` is populated by the ASM `visit()` callback; NullAway treats any annotation named `Initializer` as marking a field initializer, which replaces a blanket suppression.
- **`FieldMember.compareTo` suppression restored**: the Linux CI toolchain no longer flags it, but the Windows ARM toolchain still rejects passing the `@Nullable` `value` fields to `Comparator.compare()` without an explicit `@Nullable` type argument — the previously reviewed suppression is kept.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/ApiMemberSelector.java
+++ b/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/ApiMemberSelector.java
@@ -54,6 +54,9 @@ public class ApiMemberSelector extends ClassVisitor {
     private final boolean apiIncludesPackagePrivateMembers;
 
     private boolean isInnerClass;
+    // TODO: Replace the suppression with an @Initializer-style annotation
+    //  (https://github.com/uber/NullAway/wiki/Supported-Annotations#initialization).
+    @SuppressWarnings("NullAway")
     private ClassMember classMember;
     private boolean thisClassIsPrivateInnerClass;
 
@@ -69,7 +72,7 @@ public class ApiMemberSelector extends ClassVisitor {
     }
 
     @Override
-    public void visit(int version, int access, String name, @Nullable String signature, @Nullable String superName, @Nullable String[] interfaces) {
+    public void visit(int version, int access, String name, @Nullable String signature, @Nullable String superName, String @Nullable [] interfaces) {
         super.visit(version, access, name, signature, superName, interfaces);
         classMember = new ClassMember(version, access, name, signature, superName, interfaces);
         isInnerClass = (access & ACC_SUPER) == ACC_SUPER;
@@ -96,7 +99,7 @@ public class ApiMemberSelector extends ClassVisitor {
 
     @Nullable
     @Override
-    public MethodVisitor visitMethod(int access, String name, String desc, @Nullable String signature, @Nullable String[] exceptions) {
+    public MethodVisitor visitMethod(int access, String name, String desc, @Nullable String signature, String @Nullable [] exceptions) {
         if ("<clinit>".equals(name)) {
             // discard static initializers
             return null;
--- a/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/ApiMemberSelector.java
+++ b/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/ApiMemberSelector.java
@@ -54,9 +54,6 @@ public class ApiMemberSelector extends ClassVisitor {
     private final boolean apiIncludesPackagePrivateMembers;
 
     private boolean isInnerClass;
-    // TODO: Replace the suppression with an @Initializer-style annotation
-    //  (https://github.com/uber/NullAway/wiki/Supported-Annotations#initialization).
-    @SuppressWarnings("NullAway")
     private ClassMember classMember;
     private boolean thisClassIsPrivateInnerClass;
 
@@ -71,6 +68,7 @@ public class ApiMemberSelector extends ClassVisitor {
         return thisClassIsPrivateInnerClass;
     }
 
+    @Initializer
     @Override
     public void visit(int version, int access, String name, @Nullable String signature, @Nullable String superName, String @Nullable [] interfaces) {
         super.visit(version, access, name, signature, superName, interfaces);
--- a/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/FieldMember.java
+++ b/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/FieldMember.java
@@ -33,8 +33,12 @@ public class FieldMember extends TypedMember implements Comparable<FieldMember> 
     }
 
     @Override
+    // The explicit @Nullable type argument is required to compare the nullable 'value' fields, but it
+    // makes NullAway reject Ordering.arbitrary() (a Comparator<Object>) for a @Nullable type variable,
+    // so the suppression is needed as well.
+    @SuppressWarnings("NullAway")
     public int compareTo(FieldMember o) {
-        return super.compare(o).compare(value, o.value, Ordering.arbitrary()).result();
+        return super.compare(o).<@Nullable Object>compare(value, o.value, Ordering.arbitrary()).result();
     }
 
     @Override
--- a/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/SortingAnnotationVisitor.java
+++ b/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/SortingAnnotationVisitor.java
@@ -20,6 +20,8 @@ import org.jspecify.annotations.Nullable;
 import org.objectweb.asm.AnnotationVisitor;
 import org.objectweb.asm.Opcodes;
 
+import static java.util.Objects.requireNonNull;
+
 import java.util.LinkedList;
 import java.util.List;
 
@@ -28,6 +30,7 @@ public class SortingAnnotationVisitor extends AnnotationVisitor {
     private final List<AnnotationValue<?>> annotationValues = new LinkedList<>();
     private final AnnotationMember annotation;
 
+    @Nullable
     private SortingAnnotationVisitor parentVisitor;
     @Nullable
     private String annotationValueName;
@@ -72,7 +75,7 @@ public class SortingAnnotationVisitor extends AnnotationVisitor {
     public void visitEnd() {
         if (annotationValueName != null) {
             AnnotationAnnotationValue value = new AnnotationAnnotationValue(annotationValueName, annotation);
-            parentVisitor.annotationValues.add(value);
+            requireNonNull(parentVisitor).annotationValues.add(value);
             annotationValueName = null;
         } else if (arrayValueName != null) {
             ArrayAnnotationValue value = new ArrayAnnotationValue(
--- a/platforms/core-execution/build-cache-packaging/src/main/java/org/gradle/caching/internal/packaging/impl/TarBuildCacheEntryPacker.java
+++ b/platforms/core-execution/build-cache-packaging/src/main/java/org/gradle/caching/internal/packaging/impl/TarBuildCacheEntryPacker.java
@@ -72,6 +72,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static java.util.Objects.requireNonNull;
 import static org.gradle.internal.file.FileMetadata.AccessType.DIRECT;
 import static org.gradle.internal.snapshot.DirectorySnapshotBuilder.EmptyDirectoryHandlingStrategy.INCLUDE_EMPTY_DIRS;
 
@@ -144,7 +145,7 @@ public class TarBuildCacheEntryPacker implements BuildCacheEntryPacker {
         entity.visitOutputTrees((treeName, type, root) -> {
             FileSystemSnapshot treeSnapshots = snapshots.get(treeName);
             try {
-                long entryCount = packTree(treeName, type, treeSnapshots, tarOutput);
+                long entryCount = packTree(treeName, type, requireNonNull(treeSnapshots), tarOutput);
                 entries.addAndGet(entryCount);
             } catch (Exception ex) {
                 throw new RuntimeException(String.format("Could not pack tree '%s': %s", treeName, ex.getMessage()), ex);
--- a/platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/DefaultBuildCacheController.java
+++ b/platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/DefaultBuildCacheController.java
@@ -70,6 +70,7 @@ import java.time.Duration;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
+import static java.util.Objects.requireNonNull;
 
 public class DefaultBuildCacheController implements BuildCacheController {
     @VisibleForTesting
@@ -145,7 +146,9 @@ public class DefaultBuildCacheController implements BuildCacheController {
                 result.set(remoteResult);
             }
         });
-        return result.get();
+        // TODO: Drop requireNonNull once https://github.com/uber/NullAway/issues/681 is fixed
+        //  (the AtomicReference is initialized with Optional.empty() and only set to non-null values).
+        return requireNonNull(result.get());
     }
 
     @Override
--- a/platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/DefaultBuildCacheController.java
+++ b/platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/DefaultBuildCacheController.java
@@ -70,7 +70,6 @@ import java.time.Duration;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
-import static java.util.Objects.requireNonNull;
 
 public class DefaultBuildCacheController implements BuildCacheController {
     @VisibleForTesting
@@ -146,9 +145,7 @@ public class DefaultBuildCacheController implements BuildCacheController {
                 result.set(remoteResult);
             }
         });
-        // TODO: Drop requireNonNull once https://github.com/uber/NullAway/issues/681 is fixed
-        //  (the AtomicReference is initialized with Optional.empty() and only set to non-null values).
-        return requireNonNull(result.get());
+        return result.get();
     }
 
     @Override
--- a/platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/service/BaseLocalBuildCacheServiceHandle.java
+++ b/platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/service/BaseLocalBuildCacheServiceHandle.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
+import static java.util.Objects.requireNonNull;
 
 public class BaseLocalBuildCacheServiceHandle implements LocalBuildCacheServiceHandle {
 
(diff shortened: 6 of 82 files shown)


mlopatkin:
@reinsch82 sorry, I did a rebase of the target branch and messed up the state of this PR. I don't have anything pending in the `pt-next`.

reinsch82:
@mlopatkin rebased this one too.

platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/ApiMemberSelector.java
```diff
@@ -54,6 +54,7 @@ public class ApiMemberSelector extends ClassVisitor {
     private final boolean apiIncludesPackagePrivateMembers;
 
     private boolean isInnerClass;
+    @SuppressWarnings("NullAway")
```
mlopatkin: 🤔 This can probably be addressed by an [`@Initializer`-style annotation](https://github.com/uber/NullAway/wiki/Supported-Annotations#initialization).

Could you please add a TODO here?

platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/FieldMember.java
```diff
@@ -24,17 +24,17 @@
...
     public FieldMember(int access, String name, @Nullable String signature, String typeDesc, @Nullable Object value) {
         super(access, name, signature, typeDesc);
         this.value = value;
     }
 
     @Override
+    @SuppressWarnings("NullAway")
     public int compareTo(FieldMember o) {
-        return super.compare(o).compare(value, o.value, Ordering.arbitrary()).result();
+        return super.compare(o).<@Nullable Object>compare(value, o.value, Ordering.arbitrary()).result();
```
mlopatkin, line 41: ❓ Do we need both the explicit type and the suppression? Suppression may benefit from a justification too.

platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/SortingAnnotationVisitor.java
```diff
@@ -28,6 +28,7 @@ public class SortingAnnotationVisitor extends AnnotationVisitor {
     private final List<AnnotationValue<?>> annotationValues = new LinkedList<>();
     private final AnnotationMember annotation;
 
+    @SuppressWarnings("NullAway")
```
mlopatkin: ❓ Why is it suppressed rather than made `@Nullable`?

platforms/core-execution/build-cache-packaging/src/main/java/org/gradle/caching/internal/packaging/impl/TarBuildCacheEntryPacker.java
```diff
@@ -144,7 +145,7 @@ private long pack(CacheableEntity entity, Map<String, ? extends FileSystemSnapsh
         entity.visitOutputTrees((treeName, type, root) -> {
             FileSystemSnapshot treeSnapshots = snapshots.get(treeName);
             try {
-                long entryCount = packTree(treeName, type, treeSnapshots, tarOutput);
+                long entryCount = packTree(treeName, type, checkNotNull(treeSnapshots), tarOutput);
```
mlopatkin: 💅 `Objects.requireNonNull` is probably better to avoid having more Guava.

platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/service/BaseLocalBuildCacheServiceHandle.java
```diff
@@ -46,7 +47,7 @@ public LocalBuildCacheService getService() {
     public Optional<BuildCacheLoadResult> maybeLoad(BuildCacheKey key, Function<File, BuildCacheLoadResult> unpackFunction) {
         AtomicReference<Optional<BuildCacheLoadResult>> result = new AtomicReference<>(Optional.empty());
         service.loadLocally(key, file -> result.set(Optional.ofNullable(unpackFunction.apply(file))));
-        return result.get();
+        return Cast.unsafeStripNullable(result.get());
```
mlopatkin: 💅 Could you please add a TODO referencing https://github.com/uber/NullAway/issues/681? I believe it should be fixed in the newer versions of nullaway.

platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/service/OpFiringRemoteBuildCacheServiceHandle.java
```diff
@@ -156,6 +157,7 @@ public synchronized Throwable fillInStackTrace() {
             return this;
         }
 
+        @Nullable
```
mlopatkin: 💅  Is it really nullable? The only constructor accepts a non-nullable cause, so it should be available. We can `requireNonNull` instead.

platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/DefaultBuildCacheController.java
```diff
@@ -133,18 +134,21 @@ private Optional<BuildCacheLoadResult> loadRemoteAndStoreResultLocally(BuildCach
...
-                throw new BuildCacheOperationException("Could not load from remote cache: " + e.getMessage(), e);
-            }
+            Optional<BuildCacheLoadResult> remoteResult = loadFromRemote(key, entity, file);
             if (remoteResult.isPresent()) {
                 local.maybeStore(key, file);
                 result.set(remoteResult);
             }
         });
-        return result.get();
+        return Cast.unsafeStripNullable(result.get());
```
mlopatkin: 💅 Could you please add a TODO referencing https://github.com/uber/NullAway/issues/681? I believe it should be fixed in the newer versions of nullaway.

platforms/core-execution/file-watching/src/main/java/org/gradle/internal/watch/registry/impl/NonHierarchicalFileWatcherUpdater.java
```diff
@@ -159,7 +159,7 @@ private void updateWatchedDirectories(Map<String, Integer> changedWatchDirectori
                 fileWatcher.startWatching(directoriesToStartWatching);
             }
         } catch (NativeException e) {
-            if (e.getMessage().contains("Already watching path: ")) {
+            if (e.getMessage() != null && e.getMessage().contains("Already watching path: ")) {
```
mlopatkin, line 164: 🤔 Is NullAway smart/dumb enough to believe that `getMessage` always returns the same result?

platforms/core-execution/file-watching/src/main/java/org/gradle/internal/watch/vfs/impl/FileWatchingFilter.java
```diff
@@ -57,6 +57,14 @@ public void locationsWritten(Iterable<String> locations) {
...
     }
 
+    /**
+     * Returns whether the given location should be watched for file system changes.
+     *
+     * <p>Invariant: locationsWrittenByCurrentBuild is never null - it's initialized with
+     * FileHierarchySet.empty() and always set to non-null values in locationsWritten()
+     * and resetLocationsWritten().</p>
+     */
+    @SuppressWarnings("NullAway")
```
mlopatkin: 💅 Could you please add a TODO referencing https://github.com/uber/NullAway/issues/681?

platforms/core-runtime/internal-instrumentation-processor/src/main/java/org/gradle/internal/instrumentation/processor/modelreader/impl/TypeUtils.java
```diff
@@ -118,6 +118,7 @@ public static List<ExecutableElement> getExecutableElementsFromElements(Stream<?
 
     public static String elementQualifiedName(Element element) {
         if (element instanceof ExecutableElement) {
+            @SuppressWarnings("NullAway")
             String enclosingTypeName = ((TypeElement) element.getEnclosingElement()).getQualifiedName().toString();
```
mlopatkin: ❓ Can we replace suppression with `requireNonNull`?

mlopatkin (changes requested):
Great work! I left a few comments. It also looks like some tests are now legitimately failing.

I think the PR can target master now, and we can merge it as soon as it's green.

platforms/core-execution/file-watching/src/main/java/org/gradle/internal/watch/registry/impl/NonHierarchicalFileWatcherUpdater.java
reinsch82: It seems so :), the issue goes away.

mlopatkin: > It seems so :), the issue goes away.

Ah, they even have [this documented](https://github.com/uber/NullAway/wiki/How-NullAway-Works#:~:text=//%20NullAway%20infers%20this.next.getNext()%20to%20be%20%40NonNull%20here). Their assumption of mutations being local extends to the return value of getters. TIL. Quite practical.

I'm still suffering from CheckerFramework flashbacks, apparently.

reinsch82:
@mlopatkin reopened and addressed your review comments in 7989d9bb668:

- **ApiMemberSelector** (`classMember`): added a TODO to replace the suppression with an `@Initializer`-style annotation.
- **SortingAnnotationVisitor** (`parentVisitor`): made it `@Nullable` and `requireNonNull` it at the single dereference, instead of suppressing the field.
- **FieldMember.compareTo**: kept the suppression but documented why both it and the explicit `@Nullable` type argument are needed — the type argument is required to compare the nullable `value` fields, but it makes NullAway reject `Ordering.arbitrary()` (a `Comparator<Object>`) for a `@Nullable` type variable.
- **TarBuildCacheEntryPacker**: replaced Guava's `checkNotNull` with `Objects.requireNonNull`.
- **OpFiringRemoteBuildCacheServiceHandle.getIOException**: the cause is never null (the only constructor takes a non-null cause), so it now `requireNonNull`s instead of returning `@Nullable`.
- **DefaultBuildCacheController** & **BaseLocalBuildCacheServiceHandle**: added TODOs referencing https://github.com/uber/NullAway/issues/681.
- **FileWatchingFilter**: added a TODO referencing NullAway#681.
- **TypeUtils.elementQualifiedName**: replaced the suppression with `requireNonNull`.
- **NonHierarchicalFileWatcherUpdater**: already resolved in our earlier discussion (plain `getMessage() != null` check, no suppression).

reinsch82:
@mlopatkin I need to rebase this...

reinsch82:
Rebased onto current `master` and retargeted the PR base to `master` (it was on `ml/nullaway-pt-next`, ~4874 commits behind). A literal replay wasn't viable — `master` had diverged and done overlapping nullability work — so I re-derived the enablement against current `master`, one commit per module. All 11 modules compile clean under NullAway and pass checkstyle. This also picks up `master`'s Gradle wrapper, so no separate Gradle upgrade is needed.

Your earlier review comments are incorporated:
- **ApiMemberSelector.classMember**: suppression + TODO for an `@Initializer`-style annotation.
- **SortingAnnotationVisitor.parentVisitor**: `@Nullable` + `requireNonNull` at the dereference (not a suppression).
- **TarBuildCacheEntryPacker**: `Objects.requireNonNull` instead of Guava `checkNotNull`.
- **OpFiringRemoteBuildCacheServiceHandle.getIOException**: `requireNonNull` on the (always-present) cause.
- **DefaultBuildCacheController / BaseLocalBuildCacheServiceHandle**: `requireNonNull` with TODOs referencing NullAway#681 (the `AtomicReference.get()` case).
- **TypeUtils**: `requireNonNull` instead of suppression.
- **NonHierarchicalFileWatcherUpdater**: plain `getMessage() != null` check, no suppression.
- **FileWatchingFilter**: suppression kept with a TODO referencing NullAway#681.
- **FieldMember**: no longer flagged on current master, so the suppression is gone entirely.

platforms/core-runtime/native/src/main/java/org/gradle/internal/nativeintegration/EnvironmentModificationResult.java
```diff
@@ -16,20 +16,25 @@
...
+
 /**
  * Encapsulates what happened when we tried to modify the environment.
  */
 public enum EnvironmentModificationResult {
     SUCCESS(null),
     UNSUPPORTED_ENVIRONMENT("There is no native integration with this operating environment.");
 
+    @Nullable
     private final String reason;
```
mlopatkin, line 26: :nail_care: Let's make it non-nullable, we can have `"Environment updated successfully"` as a value for `SUCCESS`.

platforms/core-execution/snapshots/src/main/java/org/gradle/internal/snapshot/impl/DirectorySnapshotter.java
```diff
@@ -521,7 +523,7 @@ private String getInternedFileName(Path path) {
         }
 
         public FileSystemLocationSnapshot getResult() {
-            return builder.getResult();
+            return requireNonNull(builder.getResult());
```
mlopatkin: :question: Why is it safe to assume that `getResult()` returns non-null here?

platforms/core-execution/normalization-java/src/main/java/org/gradle/api/internal/changedetection/state/ZipHasher.java
```diff
@@ -100,6 +100,8 @@ private List<FileSystemLocationFingerprint> fingerprintZipEntries(String zipFile
         }
     }
 
+    // The withInputStream callback below discards its result; the explicit null return is intentional.
```
mlopatkin: :thinking: making `<T exteds @Nullable Object> withInputStream(...)` is probably a better solution here.

mlopatkin (changes requested):
Great to see another chunk of projects fixed! I have one concern about a suppression in DirectorySnapshotter, which I believe must be addressed before merging, and a couple of style nits.

platforms/core-runtime/internal-instrumentation-processor/src/main/java/org/gradle/internal/instrumentation/extensions/property/PropertyUpgradeClassSourceGenerator.java
```diff
@@ -118,13 +122,14 @@ private static MethodSpec mapToMethodSpec(CallInterceptionRequest request, Consu
             onProcessedRequest.accept(request);
             return spec;
         } catch (Exception e) {
-            onFailure.accept(new HasFailures.FailureInfo(request, e.getMessage()));
+            onFailure.accept(new HasFailures.FailureInfo(request, requireNonNull(e.getMessage())));
```
mlopatkin: ⚠️ `getMessage()` is nullable for many exceptions (NPE and friends), and this is a generic `catch (Exception e)`. If a caught exception has no message, `requireNonNull` will throw here — from inside the failure handler — masking the original error with an unrelated NPE. Since we only need a message string, let's not require non-null: `e.toString()` (keeps the exception type) or `String.valueOf(e.getMessage())` would both be safe.

platforms/core-runtime/internal-instrumentation-processor/src/main/java/org/gradle/internal/instrumentation/extensions/property/PropertyUpgradeAnnotatedMethodReader.java
```diff
@@ -187,7 +191,7 @@ public Collection<Result> readRequest(ExecutableElement method, ReadRequestConte
                 .map(Success::new)
                 .collect(Collectors.toList());
         } catch (IllegalArgumentException failure) {
-            return Collections.singletonList(new InvalidRequest(failure.getMessage()));
+            return Collections.singletonList(new InvalidRequest(requireNonNull(failure.getMessage())));
```
mlopatkin: ⚠️ Same concern as in `PropertyUpgradeClassSourceGenerator`: this only holds if every `IllegalArgumentException` reaching here was thrown by our own code with a non-null message. If one ever comes from a library, `requireNonNull` turns a reportable failure into an NPE. Safer to use `failure.toString()` / `String.valueOf(failure.getMessage())` rather than assert non-null.

platforms/core-execution/snapshots/src/main/java/org/gradle/internal/snapshot/impl/DirectorySnapshotter.java
reinsch82: I changed it a bit, but in general the assumption was safe, since [this](https://github.com/gradle/gradle/blob/e8615a2d9a4a998cd23eade6980d915ca602e731/platforms/core-execution/snapshots/src/main/java/org/gradle/internal/vfs/impl/DefaultFileSystemAccess.java#L196) also assumed `NonNull`. Now we through a dedicated exception in case it is null.

mlopatkin: > I changed it a bit, but in general the assumption was safe, since [this](https://github.com/gradle/gradle/blob/e8615a2d9a4a998cd23eade6980d915ca602e731/platforms/core-execution/snapshots/src/main/java/org/gradle/internal/vfs/impl/DefaultFileSystemAccess.java#L196) also assumed `NonNull`. Now we through an dedicated exception in case it is null.

IDEA shows [another callsite](https://github.com/gradle/gradle/blob/eef51ada5d868db32de254833a051ec69296d212/platforms/core-execution/snapshots/src/main/java/org/gradle/internal/snapshot/impl/DirectorySnapshotter.java#L403), which is `@Nullable` all the way up, so I was a bit reluctant about having a hard failure. Your current approach is fine by me, thank you for looking into it.

mlopatkin (approved):

Issue #35914, opened 2025-12-03
Improve error message on using @InptuFile on a ConfigurableFileCollection


remcomokveld:
### Issue type

Polishing

### Problem description

Thanks for all the great work you are all doing! This is super minor but I figured it could still be helpful for others in the future to submut. 

I just worked on something where I changed the following

```kotlin
@get:InputFile 
abstract val myInput: RegularFileProperty
```
 to
```kotlin
@get:InputFile 
@get:SkipWhenEmpty 
abstract val myInput: ConfigurableFileCollection
```

Combined with `myInput.singleFile`.

I missed changing `@get:InputFile` to `@get:InputFiles` which gives the following error message.

```
A problem was found with the configuration of task ':app:myTask' (type 'MyTask').
  - Type 'com.example.MyTask' property 'myInput' has unsupported value 'task ':app:myTask' property 'myInput''.

    Reason: Type 'DefaultConfigurableFileCollection' cannot be converted to a file.

    Possible solutions:
      1. Use a String or CharSequence path, for example 'src/main/java' or '/usr/include'.
      2. Use a String or CharSequence URI, for example 'file:/usr/include'.
      3. Use a File instance.
      4. Use a Path instance.
      5. Use a Directory instance.
      6. Use a RegularFile instance.
      7. Use a URI or URL instance of file.
      8. Use a TextResource instance.

    For more information, please refer to https://docs.gradle.org/8.14.3/userguide/validation_problems.html#unsupported_notation in the Gradle documentation.

* Try:
> Use a String or CharSequence path, for example 'src/main/java' or '/usr/include'
> Use a String or CharSequence URI, for example 'file:/usr/include'
> Use a File instance
> Use a Path instance
> Use a Directory instance
> Use a RegularFile instance
> Use a URI or URL instance of file
> Use a TextResource instance
```

Although it is a very nicely detailed with lots of suggestions, is missing the correct suggestion, which would be to change the annotation.. 

It would be nice if this error was tweaked to suggest changing the annotation if the type of the input is a `ConfigurableFileCollection`

tresat:
The issue is in the backlog of the relevant team and is prioritized by them.

---

We agree this error message is confusing, contains a lot of duplication, and misses the most important point - this can also be fixed by changing the annotation type, not only the property type.

Issue #36091, opened 2025-12-19
Simple problem rendering impacts rendering of problems that fail a build


ljacomet:
### Current Behavior

The rendering issue (ignoring the missing solution) shown in:
* #35914

is actually a consequence of:
* #35101

The `ProblemBodyWriter` renders solutions but is also used to render problems in `BuildExceptionReporter#getMessage` which results in duplicate output. Solutions are printed in the problem body AND in the `Try` section.

### Expected Behavior

Solutions should not appear in the problem body when a problem is rendered as a build failure. Solutions should only appear in the `Try` section there.

### Self-contained Reproducer Project

See https://github.com/gradle/gradle/blob/d185c97c3b9dbdd5ad3909e27f67e301a872cb39/platforms/core-runtime/build-configuration/src/integTest/groovy/org/gradle/interal/buildconfiguration/tasks/UpdateDaemonJvmIntegrationTest.groovy#L64-L75

### Gradle version

9.3.0-rc-1

### Gradle version that used to work

9.2.1

PR #36111, opened 2025-12-22
Remove duplication from build failure message

donat:
Fixes
- https://github.com/gradle/gradle/issues/36091

Relates to 
- https://github.com/gradle/gradle/issues/35914

### Context

This PR addresses two duplication issues in build exception reporting.

1) There's a general issue with builds failing because of a `ProblemReporter.throwing()` call. The build failure message contains a rendered version of the problem, which accidentally includes the data in the `Problem.solutions` property. This causes a duplication as the same information is displayed in `* Try:` section of the build failure. The PR adjusts the rendering implementation such that the problem rendering for build failures shouldn't include `Problem.solutions`.

2) Type validation failures are integrated with the problems API but not with the problems rendering. The validation infrastructure builds a complex message string describing the problem, the reason, and - most importantly - the solutions. To remove the duplication, the PR removes `Problem.solutions` from the build failure message if the associated problem's group is `GradleCoreProblemGroups.validation()`.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/ide/problems-api/src/integTest/groovy/org/gradle/api/problems/ProblemsServiceIntegrationTest.groovy
+++ b/platforms/ide/problems-api/src/integTest/groovy/org/gradle/api/problems/ProblemsServiceIntegrationTest.groovy
@@ -484,6 +484,25 @@ Problem found: Project is a prototype (id: sample-problems:prototype-project)
         }
     }
 
+    @Issue("https://github.com/gradle/gradle/issues/35914")
+    def "build failure message does not contain duplicate information"() {
+        given:
+        disableProblemsApiCheck()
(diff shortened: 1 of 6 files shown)


donat:
@bot-gradle test QFL

donat:
Thanks for the review!

> While this addresses duplication, I don't think it fixes https://github.com/gradle/gradle/issues/35914 entirely, because it still lacks the proper guidance. But this issue may be fixed later (and it's not blocking for RC2).

 That's right. I'll adjust the description accordingly.

PR #36113, opened 2025-12-22
Reduce reflection overhead when resolving serialization methods

bamboo:
By building only the required list of methods instead of the whole class details.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


bamboo:
@bot-gradle test ACC APT

PR #36117, opened 2025-12-22
Limit engines used for archTest to just ArchUnit

octylFractal:
Other engines were being invoked for discovery, especially Spock, which cost about 0.4s of time for each archTest on my machine. We don't use these for this task, which only runs PackageCycleTest.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


cobexer:
@blindpirate can you think of any issues this might create?

blindpirate:
No, LGTM.

PR #36120, opened 2025-12-23
Add export DEVELOCITY_ACCESS_KEY

blindpirate:
The following code not enough to set the `DEVELOCITY_ACCESS_KEY` for current bash process, so we need an extra `export DEVELOCITY_ACCESS_KEY`.

```
echo "##teamcity[setParameter name='env.DEVELOCITY_ACCESS_KEY' value='%ge.gradle.org.access.key.us%;%gbt-td.grdev.net.access.key%']"
```

Example: https://ge.gradle.org/s/p7lyu2svbwfui/custom-values

### DIFF
--- a/.teamcity/scripts/configure_build_env_on_ec2.sh
+++ b/.teamcity/scripts/configure_build_env_on_ec2.sh
@@ -34,6 +34,7 @@ AWS_REGION=$(curl -s "http://169.254.169.254/latest/meta-data/placement/region")
 if [[ "$AWS_REGION" == us-* ]]; then
   echo "For $AWS_REGION switching to user teamcityus access token"
   echo "##teamcity[setParameter name='env.DEVELOCITY_ACCESS_KEY' value='%ge.gradle.org.access.key.us%;%gbt-td.grdev.net.access.key%']"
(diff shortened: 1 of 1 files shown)


.teamcity/scripts/configure_build_env_on_ec2.sh
```diff
@@ -34,6 +34,7 @@ AWS_REGION=$(curl -s "http://169.254.169.254/latest/meta-data/placement/region")
 if [[ "$AWS_REGION" == us-* ]]; then
   echo "For $AWS_REGION switching to user teamcityus access token"
   echo "##teamcity[setParameter name='env.DEVELOCITY_ACCESS_KEY' value='%ge.gradle.org.access.key.us%;%gbt-td.grdev.net.access.key%']"
+  export DEVELOCITY_ACCESS_KEY="${DEVELOCITY_ACCESS_KEY_US}"
```
blindpirate, line 37: Can't use `DEVELOCITY_ACCESS_KEY=%ge.gradle.org.access.key.us%` because TeamCity won't replace that value inside file.

cobexer (approved):

PR #36132, opened 2025-12-24
Rebaseline performance tests and re-enable the disabled DCL perf test

h0tk3y:
#36043 brought incompatibility in the DCL schema building. Therefore, the DCL performance tests
could no longer run against the older baseline.

To fix this, bump the baseline to the recent master commit.


h0tk3y:
@bot-gradle test RFN

PR #36133, opened 2025-12-25
Generate thread dumps for FlakyTestQuarantine after 30m

blindpirate:
FlakyTestQuarantine builds usually run very fast (10~20m). If a build runs for >30m, it's likely to be hanging. This PR generates thread dumps for all FlakyTestQuarantine builds if they're longer than 30m.


blindpirate:
@bot-gradle test

cobexer (approved):

PR #36136, opened 2025-12-25
Trigger legacy release branches nightly and weekly

blindpirate:
In the past, the scheduled trigger rule is:

- For `master`/`release`, we trigger them at midnight, but only when there is pending changes.
- For old release branches, we trigger them at weekend unconditionally, but not midnight.

This PR updates the rule to be:

- For old release branches, we trigger them at:
  - Weekend unconditionally
  - Midnight if there is pending change

This is helpful when old release branch is pushed and we'd like to see a nightly next day.


cobexer (approved):

blindpirate:
@bot-gradle cherry-pick to release

PR #36161, opened 2025-12-30
Remove reasons from SelectorState

jvandort:
Now a selector does not keep track of all the reasons from edges that use the selector. The previous implementation did not account for when an edge clears its selector. A selector's forced, fromLock, and changing properties also suffer from a similar bug where they are never updated when an edge releases a selector.

Separately:
- We slightly adjust the logic for adding the forced reason. Now we declare the forced reason when a dependency is equivalent to a force. Now, the reason reflects actual forcing behavior.
- We fix a bug where once an edge contributed its reasons to a selector, it didn't subsequently apply its reasons to any new selector it used. This caused some reasons to be 'missed' previously

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test this

PR #36163, opened 2025-12-30
Re-enable the custom build init template test

ghale:
This test was disabled in https://github.com/gradle/gradle/pull/35926 due to breaking changes.  The custom build init template projects have now been updated, so we update the versions and re-enable the test.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

PR #36169, opened 2025-12-31
Refactor module registry

jvandort:
Previously, the ModuleRegistry treated Project modules -- those derived from projects in the gradle build -- differently from external modules. External modules would not have a module in the registry, and therefore we were not able to determine their dependencies or other metadata about the dependency at runtime from the ModuleRegistry.
    
This commit unifies external dependencies and project dependencies in the ModuleRegistry, treating them as the same concept. Previously, external dependencies were tracked as part of the 'runtime' classpath of a module and were opaque to the module registry. Now, external
dependencies get a module of their own.
    
There are two major changes in this commit. First, we change the way we generate the module registry structure. Second, we change the format of the ModuleRegistry, and thus the structure of the Gradle distribution.
    
Previously, the metadata of a module was saved in a .properties file within the jar that implemented the module. Since we want to treat external modules as distribution modules, we want to avoid modifying the jar of external dependencies. So, we now place the properties file corresponding to a jar next to that jar, where the name of the properties file is the name of the module with the .properties extension. This provides two efficiency bonuses for the ModuleRegistry. First, the registry does not need to unzip jars in order to determine the structure of the module graph, and second we can directly lookup modules by name. Previously, we were relying on regexes to map from module name to jar name, as the jar name
includes the version of the module.
    
The name of the jar is now included in the module properties file. So, previously the ModuleRegsitry would find a jar by regex + search, but now it can directly lookup a module's metadata by name, and the metadata includes the name of the jar.
    
To generate these properties files for each module, we change from individually generating a metadata file in each jar to generating a set of metadata files all at once. Previously, the ClasspathManifest task would live in each project, inspect that project's classpath, and write the properties file to the current project's jar. Now, GenerateClasspathModuleProperties task lives in a distribution project, inspects the entire distribution's classpath at once, and generates the properties files for all modules in the distribution. This results in significantly fewer tasks, while also allowing the metadata writing task to consider the entire distribution's classpath at once instead of many sub-classpaths individually. This change is necessary, as we need to be able to write metadata files for external dependencies, and the previous method did not allow us to do this.
    
At runtime, when loading modules, we update the ModuleRegistry implementation to be much more efficient. The previous implementation relied on a recursive graph traversal algorithm which allocated collections at each node in the graph. These allocations were non-negligible for performance, and we even had separate purpose-built build logic to 'trim' transitive dependencies to make this graph traversal process faster. Now, we implement an iterative graph traversal algorithm which is more efficient. Additionally, it allows traversing starting from multiple root nodes at once, a common use case in the gradle runtime, improving traversal efficiency, as these roots often contain overlapping transitive subgraphs.
    
The final notable change is we now store the module coordinates of each distribution module inside its properties file. This allows us to load the coordinates of the dependency at runtime. So, now the coordinates of our external dependencies are available at runtime, opening the door to exposing modules from the distribution to users. This was the primary motivation behind this commit. These changes open the door to future changes that will improve the way dependencies like gradleApi, localGroovy, and toolingApi are loaded from the distribution.
    
As part of these changes, we extensively reviewed existing logic and determined that the structure of the gradle distribution does not need to be stable across gradle versions. We did not find evidence of any cross-version bootstrap logic in the tooling API relying on the module structure of different gradle versions. One exception is DistributionFactory.InstalledDistribution, where the tooling API loads all jars
from the 'lib' directory of a distribution. However, this mechanism does not depend on the module structure -- the properties files -- of that directory. This commit does not change which jar files are present in the distribution, or where those jar files are located. It just changes where and what metadata we write about those jar files.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/GenerateClasspathModuleProperties.kt
+++ b/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/GenerateClasspathModuleProperties.kt
@@ -43,12 +43,12 @@ import java.io.File
 import java.util.Properties
 
 /**
- * Given a configuration, produces a .properties for each artifact in the configuration.
- * The properties file describes the artifact's dependencies, so that its runtime classpath
- * can be analyzed and reconstructed later.
+ * Given a configuration, produces a .properties for each node in the configuration's resolved
+ * graph. The properties file describes the node's dependencies, and stores its associated artifact,
+ * so that classpaths derived from the graph can be analyzed and reconstructed later.
  *
  * This task assumes each component in the graph has a single variant and each variant
- * has a single file.
+ * has at most one artifact.
  */
 @DisableCachingByDefault(because = "Unable to snapshot ComponentIdentifier") // Can be made cacheable after https://github.com/gradle/gradle/pull/36174
 abstract class GenerateClasspathModuleProperties : DefaultTask() {


subprojects/core/build.gradle.kts
```diff
@@ -8,14 +8,6 @@ configurations {
     register("reports")
 }
 
-tasks.classpathManifest {
-    optionalProjects.add("gradle-kotlin-dsl")
```
jvandort, line 12: Optional dependencies no longer exist as a concept, so we instead model this as a `runtimeOnly` dependency. Both the old optional and new runtimeOnly approaches cause `gradle-kotlin-dsl` to appear as a dependency of the `gradle-core` module, leading to the same classpaths.

subprojects/core/build.gradle.kts
```diff
@@ -8,14 +8,6 @@ configurations {
...
-    // The gradle-runtime-api-info.jar is added by a 'distributions-...' project if it is on the (integration test) runtime classpath.
-    // It contains information services in ':core' need to reason about the complete Gradle distribution.
-    // To allow parts of ':core' code to be instantiated in unit tests without relying on this functionality, the dependency is optional.
-    optionalProjects.add("gradle-runtime-api-info")
```
jvandort, line 16: There is no need to mark this as a dependency of core. Code that needs this module load it from within a custom classloader here: https://github.com/gradle/gradle/blob/8f2e5f54c17ccf04456df0bbfd245a68347577f2/subprojects/core/src/main/java/org/gradle/internal/service/scopes/GlobalScopeServices.java#L246

This loading mechanism has changed since this line here was originally added, which is why we don't need to make any associated changes for this in this PR

platforms/core-runtime/build-process-services/src/main/java/org/gradle/api/internal/classpath/Module.java
```diff
@@ -16,40 +16,51 @@
...
 
     /**
-     * Returns the transitive closure of all modules required by this module, including the module itself.
+     * An optional set of Maven module coordinates that identifies this module outside
+     * a Gradle distribution. These are the coordinates that would traditionally be used
+     * to resolve this module from a repository if this module were to live in an external
+     * repository.
      */
-    Set<Module> getAllRequiredModules();
+    @Nullable ModuleAlias getAlias();
```
jvandort, line 51: TODO: Add unit tests for getAlias. We don't use alias in this PR but the follow-up PR will use it.

subprojects/core/build.gradle.kts
```diff
@@ -143,6 +135,10 @@ dependencies {
         because("it needs to forward calls from instrumented code to the Kotlin standard library")
     }
 
+    runtimeOnly(projects.kotlinDsl) {
```
jvandort: kotlinDsl has a dependency on `org.gradle:kotlin-dsl-shared-runtime`, which lives in an included build. This is completely breaking configuration cache right now. See sanityCheck output

subprojects/core/build.gradle.kts
jvandort: Due to https://github.com/gradle/gradle/issues/36177 we cannot add this as a dependency to core. Instead we add it as a dependency to `launcher`. See separate workaround commit for more details

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test rfr

subprojects/core/build.gradle.kts
jvandort: Circling back to this comment, there were a number of places we still relied on runtime-api-info being available on the Gradle runtime classpath. These have been migrated to `RuntimeApiInfo` in this PR

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/GenerateClasspathModuleProperties.kt
```diff
@@ -0,0 +1,271 @@
...
+
+/**
+ * Given a configuration, produces a .properties for each artifact in the configuration.
+ * The properties file describes the artifact's dependencies, so that its runtime classpath
+ * can be analyzed and reconstructed later.
+ *
+ * This task assumes each component in the graph has a single variant and each variant
+ * has a single file.
+ */
+@DisableCachingByDefault(because = "Unable to snapshot ComponentIdentifier") // Can be made cacheable after https://github.com/gradle/gradle/pull/36174
```
eskatos, line 53: ❔ The referenced PR has been merged. Can it be made cacheable now?

subprojects/core/build.gradle.kts
jvandort: We worked around this issue.

eskatos (approved):
🚀

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/GenerateClasspathModuleProperties.kt
jvandort: We need to update the gradle/gradle build to use a release of Gradle with these changes. Since we don't have a milestone (or RC/GA) with these changes yet we cannot yet make this change

eskatos:
@jvandort, now that this is merged, the performance tests should be rebaselined to lock in the improvements.

jvandort:
@eskatos PTAL https://github.com/gradle/gradle/pull/36206

Issue #36180, opened 2026-01-03
CC report summary shows misleading information regarding unique vs considered problems


abstratt:
### Current Behavior

With the changes from https://github.com/gradle/gradle/issues/36046, the CC report now shows misleading information re: unique problems. 

The reason is that it used to 

The console shows this:

```
87354 problems were found storing the configuration cache, 2 of which seem unique.
- Unknown location: Project ':' cannot access 'Project.extensions' functionality on subprojects via 'allprojects'
- Unknown location: Project ':' cannot access 'Project.plugins' functionality on subprojects via 'allprojects'
```

which is correct. The HTML report, though, states this:

```87354 problems were found, only the first 2 were included in this report```

which is incorrect. 

### Expected Behavior

Instead, the report should say something like:

```87354 problems were found; of those, only 2 were unique and appear in this report```

### Context (optional)

The reason for this issue is that, before the fix for #36046, the report was responsible for eliminating duplicates, and had access to all problems (up to 4096 problems).

Now the report only includes unique problems. The report UI has no means to know whether the number of problems is smaller than the total number of problems due to duplication or hitting the maximum number of problems to be collected (note this is not the same was [`o.g.configuration-cache.max-problems`](https://docs.gradle.org/current/userguide/configuration_cache_enabling.html#config_cache:usage:max_problems), as that other setting is about the build status, not the number of problems to collect in the report).

### Self-contained Reproducer Project

Running this under `gradle/gradle` shows the issue:

```
GRADLE_INTERNAL_REPO_URL=file:///dev/null \
./gradlew \
-I testing/smoke-test/src/smokeTest/resources/org/gradle/smoketests/schedule-all-tasks.gradle \
-Dorg.gradle.internal.isolated-projects.parallel=false  \
-Dorg.gradle.configuration-cache.internal.recreate-cache=true \
-Dorg.gradle.configuration-cache.problems=warn \
-Dorg.gradle.configuration-cache.max-problems=200000 \
-Pgradle_installPath=/dev/null \
-PartifactoryUserName=foo \
-PartifactoryUserPassword=bar \
-PtoolingApiShadedJarInstallPath=/tmp \
--dry-run scheduleAll
```

### Gradle version

a build from master towards 9.4

PR #36181, opened 2026-01-03
Include enough info in CC report data to handle problem deduping and limiting the number of problems collected

abstratt:
Issue: #36180

Depends on: https://github.com/gradle/configuration-cache-report/pull/48

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblemsSummary.kt
```diff
@@ -64,24 +64,33 @@ class ConfigurationCacheProblemsSummary(
...
+     */
     private
     var suppressedSilentlyProblemCount: Int = 0
 
+    /**
+     * Problems reported beyond what could be collected.
+     */
+    private
+    var overflownProblemCount: Int = 0
+
```
mlopatkin, line 93: 🎉 I really like the comments.

PR #36192, opened 2026-01-05
Reorder system property priority for the wrapper

### LINKED ISSUE #36189: Wrapper: system property priority order is incorrect
### Current Behavior

According to [the documentation](https://docs.gradle.org/current/userguide/build_environment.html#setting_a_system_property), system properties provided through the command-line should take priority over system properties coming in through user-level and project-level `gradle.properties` files.

In practice, this order is the inverse (and seems to have been for years) in the specific case of the gradle wrapper:
https://github.com/gradle/gradle/blob/4b9891be14e4835a1b9c09dfd45687241f4bee47/platforms/core-runtime/wrapper-main/src/main/java/org/gradle/wrapper/GradleWrapperMain.java#L48-L58

Can this behavior be made consistent?

### Expected Behavior

Correct priority of system properties handling in the wrapper: command-line arguments should take precedence over gradle properties files.

### Context (optional)

We are trying to migrate the resolution of our custom gradle distribution to a new remote repository, which requires a way to manually override authentication during this migration.
The old authentication is defined on our CI infrastructure as secrets injected into the agents, so would instantly apply everywhere, which prevents us from executing initial verification and gradual migration.

Any gradle project that uses the wrapper can be used to reproduce.

### Gradle version

9.2.1

### Build scan URL (optional)

_No response_

_No response_

NielsDoucet:
Reorder system property priority for the wrapper to respect global and documented priorities.

Fixes #36189

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs/src/docs/release/notes.md
+++ b/platforms/documentation/docs/src/docs/release/notes.md
@@ -22,6 +22,7 @@ Include only their name, impactful features should be called out separately belo
 -->
 
 We would like to thank the following community members for their contributions to this release of Gradle:
+[Niels Doucet](https://github.com/NielsDoucet),
 [Ujwal Suresh Vanjare](https://github.com/usv240).
 
 Be sure to check out the [public roadmap](https://roadmap.gradle.org) for insight into what's planned for future releases.


NielsDoucet:
I'd like to add a test for this change in priority ordering, but I'm not sure where to add it.
Given that there's no unit test available for the `GradleWrapperMain` class, I assume there should be an integration test where this could be verified, but I didn't immediately find a suitable candidate.

tresat:

Please add tests to cover your changes.  

***

You can find some example tests to copy in [WrapperUpgradeIntegrationTest](https://github.com/gradle/gradle/blob/master/platforms/core-runtime/wrapper-main/src/integTest/groovy/org/gradle/integtests/WrapperUpgradeIntegrationTest.groovy) or [WrapperProjectIntegrationTest](https://github.com/gradle/gradle/blob/master/platforms/core-runtime/wrapper-main/src/integTest/groovy/org/gradle/integtests/WrapperProjectIntegrationTest.groovy), please add some similar test coverage to this PR to demonstrate the behavior.

NielsDoucet:
I added a test for this, but really struggled to
a) create a dedicated test class for this and have it actually run, so I "hijacked" the `WrapperIntegrationTest` class
b) create a proper integration test for this, as I couldn't see a good way to get access to the system properties inside the wrapper main function. So it's a unit test that validates the private internals, which I'm not really a fan of

Is this sufficient and acceptable? Because otherwise I'll need some guidance on how to improve on this.
I assume the reason there's currently no test for this class can be attributed (at least partially) to the issues I encountered.

ov7a:
The relevant team for this area will confirm the design of the implementation choices.

eskatos:
@bot-gradle test this

donat:
@NielsDoucet We took a look and the solution (along with the problem description) makes sense. I'll investigate if there's a historical reason why the order of system properties are reversed. 
In the meanwhile can you please look into the CI failures, such as this one: https://ge.gradle.org/s/bmzrujo3ung6

donat:
@bot-gradle test this

donat:
@NielsDoucet I've fixed the build failure. See: https://github.com/gradle/gradle/pull/36192/changes/182dded8e4920a8ab225e8d99e19ae266ef2f4f1

NielsDoucet:
> @NielsDoucet I've fixed the build failure. See: [182dded](https://github.com/gradle/gradle/commit/182dded8e4920a8ab225e8d99e19ae266ef2f4f1)

Thank you very much. Seems like there's some undocumented state management in the `CommandLineParser`.
Perhaps the parser should throw an exception if it gets configured after `parse` has been called?

NielsDoucet:
@donat Finally figured out why system properties were "eagerly" read from the command-line.
You can define gradle user home using its system property, which is required to find the user's `gradle.properties`.

I now simply initialize the command-line system properties a 2nd time in order to make sure they get highest priority again. Not pretty, but functional.

donat:
@NielsDoucet  Nice! I've added a bit of refactoring to extend the covered edge cases. I've also figured out how to define proper integration tests.

platforms/core-runtime/wrapper-main/src/integTest/groovy/org/gradle/integtests/WrapperUserHomeIntegrationTest.groovy
```diff
@@ -50,4 +51,54 @@ class WrapperUserHomeIntegrationTest extends AbstractWrapperIntegrationSpec {
...
+        prepareWrapper().run()
+        def gradleUserHome = testDirectory.file('some-custom-user-home')
+        Properties props = new Properties();
+        props.setProperty('systemProp.gradle.user.home', gradleUserHome.absolutePath)
+        try (OutputStream out = new FileOutputStream(testDirectory.file('gradle.properties'))) {
+            props.store(out, null);
+        }
+
+        when:
+        wrapperExecuter.withGradleUserHomeDir(null).withArguments('-g', gradleUserHome.absolutePath).run()
```
eskatos: ❔ It seems to me that we are defining the GUH path twice in this test, once in the `<settings-dir>/gradle.properties` file, and once on the command line using `-g`. I would have expected only the former given the test name, correct?

platforms/core-runtime/wrapper-main/src/integTest/groovy/org/gradle/integtests/WrapperUserHomeIntegrationTest.groovy
```diff
@@ -50,4 +51,54 @@ class WrapperUserHomeIntegrationTest extends AbstractWrapperIntegrationSpec {
...
+        def gradleUserHome = testDirectory.file('some-custom-user-home')
+        gradleUserHome.mkdirs()
+        def userGradleProperties = new File(gradleUserHome, 'gradle.properties')
+        userGradleProperties.text = 'systemProp.gradle.user.home=missingLocation'
+
+        when:
+        def res = wrapperExecuter.withGradleUserHomeDir(null).withArguments('-g', gradleUserHome.absolutePath).run()
+
+        then:
+        res.assertOutputContains("WARNING Ignored custom Gradle user home location configured in Gradle user home: ${userGradleProperties.absolutePath}")
```
eskatos, line 100: 🍳 🐔

platforms/core-runtime/wrapper-main/src/integTest/groovy/org/gradle/integtests/WrapperUserHomeIntegrationTest.groovy
donat: That is correct, the `-g` configuration was incorrect. I removed it.

donat:
@bot-gradle test this

platforms/core-runtime/wrapper-main/src/integTest/groovy/org/gradle/integtests/WrapperSystemPropertyPrecedenceIntegrationTest.groovy
```diff
@@ -0,0 +1,58 @@
...
+import spock.lang.Issue
+
+@Requires(value = IntegTestPreconditions.NotEmbeddedExecutor, reason = NOT_EMBEDDED_REASON)
+class WrapperSystemPropertyPrecedenceIntegrationTest extends AbstractWrapperIntegrationSpec {
+
+    @Issue('https://github.com/gradle/gradle/issues/36189')
+    def 'system properties respect project < user < cli precedence'() {
+        given:
+        prepareWrapper().run()
+        buildKotlinFile << 'println("foo=" + System.getProperty("foo", "<null>"))'
```
NielsDoucet, line 34: Are we sure we're testing and validating the system properties as read by the wrapper?
Wouldn't the gradle process itself have updated these at this point?

This is the line of thinking that made me take the ugly route by testing the wrapper's internals, but I might've just been completely wrong. Just want to make sure this is actually testing the state of the wrapper itself.

donat: You are correct. I was too fixated on the end result, but there's definitely merit in providing a unit test asserting `GradleWrapperMain`. In fact I think we should have both. I'll restore your the unit tests today.

platforms/core-runtime/wrapper-main/src/main/java/org/gradle/wrapper/GradleWrapperMain.java
```diff
@@ -69,10 +77,32 @@ public static void main(String[] args) throws Exception {
...
+        // If there are duplicate keys, the values from p2 take precedence.
+        Map<String, String> result = new HashMap<>(p1);
+        result.putAll(p2);
+        return result;
+    }
+
+    private static Map<String, String> combine(Map<String, String> p1, Map<String, String> p2, Map<String, String> p3) {
+        // If there are duplicate keys, the values from p3 take precedence over p2, which take precedence over p1.
+        Map<String, String> result = new HashMap<>(p1);
+        result.putAll(p2);
```
reinsch82:
```suggestion
        Map<String, String> result = combine(p1, p2);
```

reinsch82 (approved):
Looks mostly good :)

platforms/core-runtime/wrapper-main/src/main/java/org/gradle/wrapper/GradleWrapperMain.java
donat: Fixed.

donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@NielsDoucet thank you for the contribution!

Issue #29049, opened 2024-05-07
Improve dependency resolution for localGroovy()


donat:
https://github.com/gradle/gradle/pull/26743#issuecomment-2003121025

> ... we actually want to get rid of the property altogether. The Gradle libs URLs was a workaround because the localGroovy dependency is handled in Gradle in a special way. We aim to simplify Gradle and making localGroovy resolvable/configurable via dependency management.

akiraly:
@donat thanks for scoping this into 8.9. One more idea/question: would it be possible to not do the automatic/implicit localGroovy resolution at all when a project is only using the Kotlin DSL?

donat:
This is not my area of expertise, but my understanding is that some Groovy bits are part of the Gradle API. For example, we still have a few methods taking `Closure` parameter (example: [Project.java](https://github.com/gradle/gradle/blob/master/subprojects/core-api/src/main/java/org/gradle/api/Project.java#L1559)). Also, the Kotlin DSL offers interoperability with Groovy scripts (example: [withGroovyBuiler](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.kotlin.dsl/with-groovy-builder.html)). Because of this, I assume we need the localGroovy even when only the the Kotlin DSL is being used by the project, at least for the time being.

big-guy:
We want the `localGroovy()` dependency to behave like a regular dependency with capabilities and participation in dependency conflict handling.

As it exists today, it acts like a file collection dependency and is invisible to dependency resolution. 

A couple of ideas for a solution:
- Make file collection dependencies a part of the graph
- Make localGroovy add a regular dependency and use special handling to use the Gradle distribution like a repository

ysb33r:
@big-guy Do you think this will get into 9.1 perhaps?

big-guy:
We might be able to just deprecate localGroovy and recommend that people use external coordinates for Groovy instead.  Groovy Gradle Plugin could make sure you get the right version of Groovy.

We still have a need to do capability conflicts with the bundled version of Groovy and any external Groovy dependencies when they appear on certain classpaths.

PR #36205, opened 2026-01-07
Introduce gradleDistribution repository

jvandort:
This repository allows users to resolve select dependencies from the gradle distribution. This is intended as a replacement for the existing localGroovy dependency, and will eventually be used to provide a local copy of the new gradle public api jar

Related: https://github.com/gradle/gradle/issues/29049

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs/src/docs/release/notes.md
+++ b/platforms/documentation/docs/src/docs/release/notes.md
@@ -42,6 +42,36 @@ For Java, Groovy, Kotlin, and Android compatibility, see the [full compatibility
 
 <!-- Do not add breaking changes or deprecations here! Add them to the upgrade guide instead. -->
 
+### Gradle Distribution Repository
+
+This release introduces the `gradleDistribution()` repository, which allows select bundled dependencies to be resolved directly from the Gradle distribution without requiring network access.
+The `gradleDistribution()` repository can be used like any other repository:
+
+```kotlin
+plugins {
+    id("java-library")
+}
+
+repositories {
+    gradleDistribution()
+}
+
+dependencies {
+    implementation("org.apache.groovy:groovy:4+")
+}
+```
+
+By running `./gradlew dependencies --configuration runtimeClasspath`, we can see this resolves
+to the Groovy version bundled with Gradle:
+
+```
+runtimeClasspath - Runtime classpath of source set 'main'.
+\--- org.apache.groovy:groovy:4+ -> 4.0.29
+```
+
+The versions of dependencies bundled with Gradle may change between minor releases.
+For this reason, dynamic versions should always be used when resolving dependencies from the `gradleDistribution()` repository.
+
 ### Bearer Token Authentication for Wrapper Download
 
 When downloading Gradle distributions from an HTTPS backend (or even from an HTTP one, but the secure version is preferred), the Wrapper now also supports Bearer token authentication.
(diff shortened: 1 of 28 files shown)


jvandort:
TODO: Add release notes

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

bamboo:
Perhaps this could also be used to resolve the embedded Kotlin dependencies and avoid the downloads on first use of `embedded-kotlin` and/or `kotlin-dsl`.

jvandort:
Which downloads are you talking about in particular? Not sure im familiar with these

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/distribution/DefaultAvailableDistributionModules.java
```diff
@@ -0,0 +1,205 @@
...
+ */
+public class DefaultAvailableDistributionModules implements AvailableDistributionModules {
+
+    /**
+     * All top-level modules we have decided to allow users to depend on. All modules in this list,
+     * and all of their transitive dependencies, will be available to users via the Gradle
+     * distribution repository.
+     */
+    private static final List<String> EXPOSED_TOP_LEVEL_MODULES = ImmutableList.<String>builder()
+        .addAll(DependencyClassPathProvider.GROOVY_MODULES)
```
jvandort, line 51: We should determine if we want to expose all of these modules to start (which are part of localGroovy) or some subset instead

bamboo:
> Which downloads are you talking about in particular? Not sure im familiar with these

I'm talking about an included build logic project applying the `kotlin-dsl` plugin. That will cause the project to depend on the Kotlin standard libraries.

jvandort:
Yeah. Sterling and I have already talked about how this mechanism could be used on the plugin classpath to resolve plugins and their dependencies. We were talking about the Java plugin and breaking up the existing monolith classpath we have now. But the same pattern should work with the kotlin dsl plugin

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -42,6 +42,36 @@ For Java, Groovy, Kotlin, and Android compatibility, see the [full compatibility
...
+By running `./gradlew dependencies --configuration runtimeClasspath`, we can see this resolves
+to the Groovy version bundled with Gradle:
+
+```
+runtimeClasspath - Runtime classpath of source set 'main'.
+\--- org.apache.groovy:groovy:4+ -> 4.0.29
+```
+
+The versions of dependencies bundled with Gradle may change between minor releases.
+For this reason, dynamic versions should always be used when resolving dependencies from the `gradleDistribution()` repository.
```
hfhbd, line 73: What about a POM using the Gradle version to not specifying a dynamic version at all?

jvandort: So you're suggesting a new synthetic dependency you can declare that acts as a platform? 

IMO the annoying part there is you have two steps now. 1. Declare distribution repo 2. Declare distribution platform

hfhbd: Yeah, that's true, but on the other hand, it would easier when writing a Gradle plugin targeting different Gradle versions, wouldn't it?
The platform POM could be added automatically when you specify the target Gradle version for your plugin.

jvandort: We plan to have the Gradle API embedded in the distribution as well, so if you declare a gradle api dependency it will already have transitive dependencies on groovy, etc, with the proper versions

PR #36207, opened 2026-01-07
Increase timeout for Linux DocsTest

blindpirate:
While we are investigating https://github.com/gradle/gradle-private/issues/4994, increase timeout to 90m to unblock merge.


cobexer:
Let's set up the timeout so that the tests can succeed (slowly) even if Test Distribution is disabled or unavailable.

cobexer (approved):

PR #36211, opened 2026-01-07
Fix broken test where we now erroneously find the properties file instead of the jar

jvandort:
Caused by changes from https://github.com/gradle/gradle/pull/36169

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/extensibility/test-kit/src/integTest/groovy/org/gradle/testkit/runner/enduser/GradleRunnerMiscEndUserIntegrationTest.groovy
```diff
@@ -75,7 +75,7 @@ class GradleRunnerMiscEndUserIntegrationTest extends BaseTestKitEndUserIntegrati
             }
         }
 
-        def testKitJar = jarsDir.listFiles().find { it.name.contains "test-kit" }
+        def testKitJar = jarsDir.listFiles().find { it.name.contains("test-kit") && it.name.endsWith(".jar") }
```
big-guy: Maybe instead of checking for jar here, we should only copy jars above

jvandort:
@bot-gradle cancel

PR #36214, opened 2026-01-07
Add Best Practice: `map` versus `flatmap` when wiring properties

lkasso:
Tested:
`./gradlew :docs:docsTest --tests "*.snippet-best-practices-be-careful-with-map-flatmap*"`


platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,78 @@
...
+val generator = tasks.register<GeneratorTask>("generate") {
+    outputFile.set(layout.buildDirectory.file("out.txt"))
+}
+
+val consumer = tasks.register<ConsumerTask>("consume") {
+    // CORRECT: Map the property directly.
+    // The resulting Provider<String> still "knows" it comes from the 'generator' task.
+    inputContent.set(generator.flatMap { it.outputFile }.map { it.asFile.readText() })
+}
+----
```
tresat: ❗ I think this is a very subtle footgun, and deserves a full running example.  This is a good start at it.

platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,78 @@
...
+The primary difference between `map` and `flatMap` in the <<lazy_configuration.adoc#lazy_configuration_faqs,Provider API>> is how they handle the producer information.
+When you use `map`, Gradle treats the transformation as a continuation of the existing provider; the task dependency (the _producer_) remains attached.
+
+In contrast, `flatMap` is designed to transform one provider into a completely different provider.
+If you use `flatMap` to extract a property from a task (e.g., `task.flatMap { it.someProperty }`), Gradle often loses the connection to the original task that owns that property.
+The code will look logically correct but fails at execution time because the producer task is never added to the task graph.
+
+Key rules:
+
+- Don't ever call .`get()` inside a closure. Calling `.get()` realizes the value immediately, breaking lazy configuration and task dependency tracking.
```
tresat:
```suggestion
- It is usually a mistake to call .`get()` on a `Provider` inside a closure. Calling `.get()` realizes the value immediately, breaking lazy configuration and task dependency tracking.
```

platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,78 @@
...
+- If you need to transform a file's content, ensure the file's provider is what you are mapping over, not the task itself.
```
tresat: ❓ Why not the task itself?  That wouldn't give you file content.  I'm not sure what this is trying to say.

tresat (changes requested):
Also needs an index entry.

platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,78 @@
...
+[source,kotlin]
+----
+val generator = tasks.register<GeneratorTask>("generate") {
+    outputFile.set(layout.buildDirectory.file("out.txt"))
+}
+
+val consumer = tasks.register<ConsumerTask>("consume") {
+    // FOOTGUN: flatMap here loses the dependency on the 'generator' task
+    // Gradle will try to run 'consume' without running 'generate' first
+    inputContent.set(generator.flatMap { it.outputFile.map { file -> file.asFile.readText() } })
```
mlopatkin: :x: I tried your example and it works just fine, assuming a reasonably written Generator (with `@OutputFile` - but without that it won't work for the correct solution too). What I see in the bugreports is typically something along the lines of `generator.map { it.outputFile.get().asFile.readText }`, which indeed doesn't really work, at least with CC.

platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+Using `flatMap` on a `TaskProvider` to extract an input property will sever the automatic task dependency chain, causing build failures or incorrect incremental results.
+
+=== Explanation
+
+The primary difference between `map` and `flatMap` in the <<lazy_configuration.adoc#lazy_configuration_faqs,Provider API>> is how they handle the producer information.
+When you use `map`, Gradle treats the transformation as a continuation of the existing provider; the task dependency (the _producer_) remains attached.
+
+In contrast, `flatMap` is designed to transform one provider into a completely different provider.
+If you use `flatMap` to extract a property from a task (e.g., `task.flatMap { it.someProperty }`), Gradle often loses the connection to the original task that owns that property.
```
mlopatkin: >Gradle often loses the connection to the original task that owns that property.

:x: While there are many corner cases where Gradle may miss the dependency, I wouldn't call them happening "often". Maybe it is a good idea to spell the proper way, which is to have an `abstract Property` annotated  with some flavor of `@Output`, in the best practices too.

Dependency propagation doesn't work when your flatmap returns something not annotated as `@Output`, but that's probably fine, nobody should expect that. Then the minefield starts:
- `@Nested` beans with `@Output` properties may or may not work, depending on how you get the bean
- If the annotated getter returns a `Provider`, it doesn't work either - this is probably the most popular footgun here.
- Providers created inside `flatMap` closure with `provider {}` do not work.


So in the end, I think we should encourage people to have just:
```kotlin
@get:OutputFile
abstract val outputFile: RegularFileProperty
```

platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_laziness.adoc
mlopatkin: :+1: This is a good point, and a source of many issues people face.

mlopatkin (changes requested):
I'm sorry for being negative again, but this looks off. I read the advice as "avoid `flatMap`", but this shouldn't be our message, at least not with today's API surface. We actually want people to build dependency chains with `flatMap`.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices.adoc
```diff
@@ -54,6 +54,7 @@ To view all Gradle Best Practices at a glance, along with the Gradle version the
 - <<best_practices_tasks.adoc#best_practices_for_tasks,Best Practices for Tasks>>
 - <<best_practices_performance.adoc#best_practices_for_performance,Best Practices for Performance>>
 - <<best_practices_security.adoc#best_practices_for_security,Best Practices for Security>>
+- <<best_practices_laziness.adoc#best_practices_for_laziness,Best Practices for Laziness>>
```
tresat: ❗ I'm unsure about having "Laziness" as a top-level category.

Laziness is just a mechanism, the goal of using laziness is Performance, or Correctness or Comprehensibility.  If the Laziness is used in a particular section of the DSL (like tasks, or testing) we should put lazy items there.  If it is a more general recommendation, I think Performance is more appropriate.

For this page, I think the first sub-item ("Use flatMap to Extract...") could be put under Tasks, as the purpose of using the proper mapping method is to preserve task inputs.  But then the other sub-items really are "generic Provider API things"...

But I get that there are a lot of Provider API gotchas, so maybe it's worth grouping them together in a new section.  I think if we do continue to go this route, we should definitely separate these 3 sub-items into full standalone recommendations with separate Explanations, index entries, etc.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,132 @@
...
+// limitations under the License.
+
+[[best_practices_for_laziness]]
+= Best Practices for Laziness
+
+[[map_versus_flatmap]]
+== Be Careful with `map` vs. `flatMap` when Wiring Properties
+
+When wiring task outputs to inputs, several common patterns break task dependency tracking or cause configuration-time failures.
+Understanding when to use `map`, `flatMap`, and how to avoid creating "naked providers" is essential for reliable builds.
```
tresat: ❗ "Naked" meaning unintentionally unattached to task dependencies.  Do we use this language anywhere else?  "Naked" makes me think more in the direction of undecorated or unlabeled, and not unattached or disconnected.  I think there must be a better word for this kind of mistake.

It's bad that if you go to the linked reference page and search for "naked" there are 0 hits.  If we do use this term, we should align the language on that page too.  I think these examples will lead many people to reading that full reference page - there's just too much about Provider API to easily summarize.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+=== Explanation
+
+Gradle's Provider API uses `map` and `flatMap` to transform and connect task properties.
```
tresat: 💅 
```suggestion
Well-designed builds use Gradle's Provider API's `map` and `flatMap` methods to transform and connect properties across tasks.
```

The Provider API is a general purpose API - it's not specific to task properties.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,132 @@
...
+However, certain patterns break the connection between tasks, causing builds to fail or produce incorrect results.
+
+When wiring task outputs to inputs:
+
+1. **Use `flatMap` to extract Provider properties** — this is correct and necessary for the Provider API
```
tresat: 💅 We shouldn't say "properties" here, it's overloaded in this context.

We could say:

```suggestion
1. **Use `flatMap` to reference task properties from providers of taskss** — this is correct and necessary to preserve task dependency information
```

I don't like saying "dependencies" though, that makes me think of Dependency Management stuff.  Maybe also mention that providers of tasks are the result of `task.register` calls?

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,132 @@
...
+2. **Chain `map` to transform values** — don't create new providers with `provider {}`
+3. **Use `map` to extract eager properties** — extract plain values like `File` or `String`, not Provider properties
```
tresat: 🤔 I think "eager properties" is confusing.  Also, we should avoid using the word properties here since this is general provider advice.


```suggestion
3. **Use `map` to eagerly extract values from provider* — extract plain values like `File` or `String`, not return different Providers
```

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,132 @@
...
+4. **Never call `.get()` to read file content in callbacks** — the file doesn't exist yet at configuration time
```
tresat: 🤔 This is a very specific bit of advice about File/Dir valued Properties/Providers and configuration vs. execution time.  I think it definitely belongs as a separate best practice.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,132 @@
...
+5. **Use directly annotated output properties** — avoid deriving properties through `.map {}`
```
tresat: 🤔 I'm not sure people will understand what "directly annotated output properties" means, and this advice _seems_ to clash with 2 and 3.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+NOTE: For dependency tracking to work, task output properties must be properly annotated with `@OutputFile`, `@OutputDirectory`, or similar annotations.
```
tresat: 💅 This is the explanation of the language in 5 - why keep as a separate NOTE rather than inline in 5?

Could we link to the package here?  Or some list of all of the possible output annotations - we must have that somewhere?

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,132 @@
...
+Without these annotations, Gradle cannot track task dependencies.
+
+=== Use `flatMap` to Extract Lazy Provider Properties and `map` to Transform Values
```
tresat: 🤔 I think "Lazy Provider Properties" is going to 🤯 readers.
 
```suggestion
=== Use `flatMap` to lazily extract new Providers from Task Properties and `map` to Transform Values
```

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+When wiring task outputs to inputs, use `flatMap` to extract a Provider property from a task, then chain `map` to transform the value.
```
tresat: 💅 We need to do a pass on this doc and the examples for consistent capitalization of provider, task, property, etc.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,132 @@
...
+==== Don't Do This
+
+Creating a new provider inside `flatMap` loses all dependency information:
+
+====
+include::sample[dir="snippets/bestPractices/beCarefulWithMapFlatmap-avoid/kotlin",files="build.gradle.kts[tags=naked-provider]"]
+include::sample[dir="snippets/bestPractices/beCarefulWithMapFlatmap-avoid/groovy",files="build.gradle[tags=naked-provider]"]
+====
+
+The `provider {}` factory creates a provider with no connection to any task, so Gradle cannot determine that it depends on the `generate` task.
```
tresat: Can we turn this and the inline comments in the example into callouts?

platforms/documentation/docs/src/snippets/bestPractices/beCarefulWithMapFlatmap-avoid/groovy/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+}
+// end::common-tasks[]
+
+// tag::naked-provider[]
+def generator = tasks.register('generate', GeneratorTask) {
+    outputFile = layout.buildDirectory.file('output.txt')
+}
+
+tasks.register('consumeNaked', ConsumerTask) {
+    // BROKEN: Creating provider {} inside flatMap creates a "naked provider"
```
tresat: ❗ This comment seems to be on the wrong line - the code on line 37 isn't doing this, it's done later in lines 40-42, right?

platforms/documentation/docs/src/snippets/bestPractices/beCarefulWithMapFlatmap-avoid/groovy/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+    // The resulting provider has NO knowledge of the 'generator' task
+    inputFile = generator.flatMap { it.outputFile }
```
tresat: ❗ This isn't true - look at the Provider javadoc.

platforms/documentation/docs/src/snippets/bestPractices/beCarefulWithMapFlatmap-avoid/groovy/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+
+    // BROKEN: Don't chain without map
+    inputContent = generator.flatMap {
+        provider { it.outputFile.get().asFile.text }
```
tresat: 🤔 This is doing 2 separate things wrong - the `provider {}` and the `get()` inside of the `flatMap {}`.  Maybe these should be separated.

platforms/documentation/docs/src/snippets/bestPractices/beCarefulWithMapFlatmap-avoid/groovy/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+
+// tag::derived-property[]
+abstract class ProducerTask extends DefaultTask {
+    @Internal
+    abstract DirectoryProperty getSomeDirectory()
+
+    // This property is DERIVED via map - not directly annotated
+    @OutputFile
+    Provider<RegularFile> getOutputFile() {
+        someDirectory.map { it.file('output.txt') }
```
tresat: 💅 We should always use `return` even though Groovy doesn't require it.

platforms/documentation/docs/src/snippets/bestPractices/beCarefulWithMapFlatmap-do/groovy/build.gradle
```diff
@@ -0,0 +1,128 @@
...
+}
+
+tasks.register('consume', ConsumerTask) {
+    // CORRECT: flatMap extracts the Provider property
+    // The dependency on 'generate' is preserved
+    inputFile = generator.flatMap { it.outputFile }
+
+    // CORRECT: Chain map on the Provider to read content lazily
+    inputContent = generator.flatMap { it.outputFile }
+        .map { it.asFile.text }
```
tresat: While correct, what I'm unclear about looking at the example is why we need separate properties for `inputFile` and `inputContent`.  Wouldn't a task just read the contents of the input file in the `@TaskAction` using: `inputFile.get().asFile.text`?

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_laziness.adoc
```diff
@@ -0,0 +1,132 @@
...
+include::sample[dir="snippets/bestPractices/beCarefulWithMapFlatmap-do/kotlin",files="build.gradle.kts[tags=flatmap-provider]"]
+include::sample[dir="snippets/bestPractices/beCarefulWithMapFlatmap-do/groovy",files="build.gradle[tags=flatmap-provider]"]
+====
+
+This preserves the task dependency because the provider chain remains connected to the original task.
+
+=== Use `map` for Eager Properties
+
+When a task uses old-style eager properties (like `File`) instead of Provider properties, use `map` instead of `flatMap` to extract the value.
+The `flatMap` operator expects a Provider return type, so it cannot be used with plain values:
```
tresat: ❗ `flatMap` is a method, not an operator.  It's also weird to say it **expects** a Provider return type - it **has** a Provider return type.

tresat (changes requested):
I think we need to separate this into separate items, with their own examples.

One of the (newly emerged) goals is to have these items and example be easily interpretable, specific recommendations that AI Agents could read, understand, match and use to fix a build.  I think there is too much here, and the lack of a separate explanation for each sub-item will hurt the usability of this.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices.adoc
lkasso: Moved to Tasks

platforms/documentation/docs/src/snippets/bestPractices/beCarefulWithMapFlatmap-do/groovy/build.gradle
lkasso: It's just for the purpose of the example.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_index.adoc
```diff
@@ -86,6 +86,8 @@ Use this as a quick reference to track newly added recommendations, check adopti
 | <<best_practices_security.adoc#validate_wrapper_checksum,Validate the Gradle Wrapper on every Upgrade>> | Security | 9.3.0
 
 | <<best_practices_testing.adoc#test_custom_types_with_testkit,Test your custom Task and Plugins with TestKit>> | Testing | 9.4.0
+
+| <<best_practices_tasks.adoc#map_versus_flatmap,Wiring Task Outputs with `map` and `flatMap`>> | Tasks | 9.5.0
```
lkasso:
```suggestion
| <<best_practices_tasks.adoc#map_versus_flatmap,Wiring Task Outputs with `map` and `flatMap`>> | Tasks | 9.7.0
```

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
```diff
@@ -698,3 +698,131 @@ Additionally, because the `Project` object is not accessed after configuration,
...
+
+[source,kotlin]
+----
+consumer.inputContent.set(
+    producer.flatMap { it.outputFile }       // extract the Provider, keeping the task dependency
+        .map { it.asFile.readText() }        // transform the value lazily at execution time
+)
+----
+
+For dependency tracking to work, task outputs should be directly annotated abstract getters:
```
tresat, line 738: 🤔 Probably worth showing the Groovy syntax too here, since it's quite different in this case.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
```diff
@@ -698,3 +698,131 @@ Additionally, because the `Project` object is not accessed after configuration,
...
+abstract val outputFile: RegularFileProperty
+----
+
+With this pattern, `flatMap` reliably carries the task dependency and `map` safely transforms the result.
+
+A few specific patterns break this chain:
+
+1. **Do not call `.get()` inside `map` or `flatMap`** — the file doesn't exist yet at configuration time, breaking lazy configuration and task dependency tracking
+2. **Do not use standalone `provider {}` to wrap task outputs** — `provider {}` produces a disconnected provider with no task dependency information
+3. **Use directly annotated abstract getters for task outputs** — `flatMap` can only track dependencies through directly annotated abstract getters, not outputs computed via `.map {}`. Getters that return `Provider` instead of `Property` also break tracking, even when annotated. Dependency tracking through `@Nested` objects is unreliable and should be avoided.
```
tresat: 🤔 "Dependency tracking through `@Nested` objects is unreliable and should be avoided." - is this still true?  Do we have a test to confirm it?

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
```diff
@@ -698,3 +698,131 @@ Additionally, because the `Project` object is not accessed after configuration,
...
+4. **Use `map` for plain values, `flatMap` for `Provider`-typed outputs** — `flatMap` requires its closure to return a `Provider`; use `map` to extract `File` or `String` values
+
+=== 1. Do not call `.get()` inside `map` or `flatMap`
+
+Calling `.get()` inside `map` to read file content fails because the file does not exist yet at configuration time — the producer task has not run.
+This pattern also breaks the Configuration Cache.
+
+==== Don't Do This
```
tresat: ❗ For this example, It might be good to also show the `GeneratorTask`.  Or at least mention explicitly that the same task is used in both the don't and do examples.  The difference is subtle here - on first read I thought the task itself differed.

Probably shouldn't name it `eagerGenerator` vs. `lazyGenerator`, too.  Just `generatorTask`.

tresat: Also, the:

```
// BROKEN: Tries to read the file at configuration time
// The file doesn't exist yet - generator hasn't run!
```

Should probably go directly above the `inputContent` configuration.  It seems to apply to the `inputFile.set()` call, which was another point of confusion, as this is the exact same code in both examples (just with a different task name).

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
```diff
@@ -698,3 +698,131 @@ Additionally, because the `Project` object is not accessed after configuration,
...
+====
+
+This preserves the task dependency because the provider chain remains connected to the original task.
+
+=== 2. Do not use standalone `provider {}` to wrap task outputs
+
+A standalone `provider {}` creates a provider with no connection to any task.
+Gradle cannot determine that it depends on the producer task, so the producer will not be added to the task graph.
+
+==== Don't Do This
```
tresat: Same comment re: generator task names in example 2.  If only the consumer task is changing, only the consumer task should be named differently.

tresat: I'd also consider moving the declaration of `val content` directly _inside_ the `consumeNaked` task declaration + configuration block, to keep the examples as similar as possible.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
```diff
@@ -698,3 +698,131 @@ Additionally, because the `Project` object is not accessed after configuration,
 === Tags
 
 `<<tags_reference.adoc#tag:tasks,#tasks>>`, `<<tags_reference.adoc#tag:inputs-and-outputs,#inputs-and-outputs>>`
+
+[[map_versus_flatmap]]
+== Wiring Task Outputs with `map` and `flatMap`
```
tresat:
```suggestion
== Properly wire lazy Task `@Output`s using `map` and `flatMap`
```

I think the title should match the imperative tone of the other BPs here.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
```diff
@@ -698,3 +698,131 @@ Additionally, because the `Project` object is not accessed after configuration,
...
+==== Do This Instead
+
+Use `flatMap` to extract the `Provider`-typed output, then chain `map` to transform the value:
+
+====
+include::sample[dir="snippets/bestPractices/beCarefulWithMapFlatmap-do/kotlin",files="build.gradle.kts[tags=flatmap-provider]"]
+include::sample[dir="snippets/bestPractices/beCarefulWithMapFlatmap-do/groovy",files="build.gradle[tags=flatmap-provider]"]
+====
+
+=== 3. Use directly annotated abstract getters for task outputs
```
tresat: For all the examples, why get away from annotations like `//<1>` in favor of comments like `//BROKEN` or `//WORKAROUND`?

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
```diff
@@ -698,3 +698,131 @@ Additionally, because the `Project` object is not accessed after configuration,
...
+====
+
+The preferred approach is to use directly annotated abstract getters instead of derived ones, so that `flatMap` works reliably:
+
+====
+include::sample[dir="snippets/bestPractices/beCarefulWithMapFlatmap-do/kotlin",files="build.gradle.kts[tags=direct-annotation]"]
+include::sample[dir="snippets/bestPractices/beCarefulWithMapFlatmap-do/groovy",files="build.gradle[tags=direct-annotation]"]
+====
+
+=== 4. Use `map` for plain values, `flatMap` for `Provider`-typed outputs
```
tresat: I don't think it is valid to say `it cannot be used with plain values`.  You _could_ create a new provider inside of the `flatMap` call - but then you lose the task dependency.

This lacks a Don't example.  I think if it had one, showing a standalone `provider` call, it would be more clear.

But at that point... it would be pretty much a duplicate of item 2.

I think after reading 3 items with the familiar Avoid - Do pattern, I was expecting the same here.  But there is only a Do.  Perhaps a header "Do This" could make that less surprising.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
```diff
@@ -698,3 +698,131 @@ Additionally, because the `Project` object is not accessed after configuration,
...
+With this pattern, `flatMap` reliably carries the task dependency and `map` safely transforms the result.
+
+A few specific patterns break this chain:
+
+1. **Do not call `.get()` inside `map` or `flatMap`** — the file doesn't exist yet at configuration time, breaking lazy configuration and task dependency tracking
+2. **Do not use standalone `provider {}` to wrap task outputs** — `provider {}` produces a disconnected provider with no task dependency information
+3. **Use directly annotated abstract getters for task outputs** — `flatMap` can only track dependencies through directly annotated abstract getters, not outputs computed via `.map {}`. Getters that return `Provider` instead of `Property` also break tracking, even when annotated. Dependency tracking through `@Nested` objects is unreliable and should be avoided.
+4. **Use `map` for plain values, `flatMap` for `Provider`-typed outputs** — `flatMap` requires its closure to return a `Provider`; use `map` to extract `File` or `String` values
+
+=== 1. Do not call `.get()` inside `map` or `flatMap`
```
tresat: Can we use `====` here, to indent each of the 4 items one notch under the `Explanation` in the sidebar ToC?  It is less readable with all of these items at the same level of indentation:

Image

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
tresat: "outputs computed via `.map {}`" - this was flagged by @big-guy as unclear.  Reading this again, I agree it should be rephrased.

I think the word "computed" is the crux of this for me.  How about something like: 

"task outputs that exist only to define transforms of other properties"?

Wordier, but it makes it clear to me what we're talking about.

tresat (changes requested):
I gave this a close reading, in light of the review thread in Slack, and have a few clarifications I think are worth making.  Hopefully none of this came up before.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
lkasso: Done.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
lkasso: Added

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
lkasso: `subprojects/core/src/integTest/groovy/org/gradle/api/tasks/NestedInputIntegrationTest.groovy` (lines 122–162) shows the tracking is working, `@Nested` + `@OutputFile` does propagate task dependencies.
I think this came from a discussion with Mike that I misunderstood. I'm removing this, thanks for the catch.

lkasso: Sounds good. I used your suggestion.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
lkasso: Done.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
lkasso: I updated the name to `generatorTask` but I will run into an issue because the task is used in two different examples. I might have to use a different name or different files.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
lkasso: Done. Had to create two `-avoid` folders for this to work.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
lkasso: Force of habit from "regular" docs, switching to numbered callouts.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
lkasso: Done.

tresat (approved):
I think this is ready.  @big-guy - do you want to give this one more read through?  @mlopatkin - are your concerns addressed?

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
```diff
@@ -698,3 +698,175 @@ Additionally, because the `Project` object is not accessed after configuration,
...
+
+[[map_versus_flatmap]]
+== Wire lazy task outputs using `map` and `flatMap`
+
+Use `flatMap` to extract a `Provider`-typed output from a task, and `map` to transform the resulting value.
+Together, they preserve the task dependency chain so Gradle knows which tasks must run first.
+
+=== Explanation
+
+When wiring task outputs to inputs, `flatMap` extracts a `Provider`-typed output from a task while preserving the task dependency.
```
mlopatkin: 🤔 This is a very tricky topic and I think it is worth it to be extra precise here:
>while preserving the task dependency.

should be

>while preserving the task dependency that the output `Provider` carries.

`taskProvider.flatMap { it.outputFileProperty }` doesn't preserve the task dependency of `taskProvider`. It is `outputFile` that has an associated dependency on its owner.

Being precise can help people build the intuition why `taskProvider.flatMap { provider { it.getOutputFile() } }` doesn't work.

mlopatkin (commented):
Looks good, but I think we should be slightly more precise about flatMap semantics.

platforms/documentation/docs/src/docs/userguide/best-practices/best_practices_tasks.adoc
lkasso: Done!

mlopatkin (approved):

PR #36219, opened 2026-01-08
Ensure that all new packages are null-marked

mlopatkin:
This adds an architecture test that ensures every package with Java code has `@NullMarked` in it.  Existing violations are frozen, so it gives us an idea where we are in terms of markings.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


testing/architecture-test/src/test/java/org/gradle/architecture/test/ArchUnitFixture.java
```diff
@@ -350,6 +351,27 @@ static ArchCondition<JavaMethod> beNullUnmarkedMethod() {
         return beAnnotatedWith(NullUnmarked.class);
     }
 
+    /**
+     * Verifies that the class is in a package with {@code @NullMarked} annotation. This condition emits violations once per package.
+     *
+     * @return the class condition
+     */
+    static ArchCondition<JavaClass> beInANullMarkedPackage() {
+        return new ArchCondition<>("be in a @NullMarked package") {
```
mlopatkin, line 351: 💭 It looks easier to just implement the condition directly rather than trying to come up with a combination of existing pieces. I'm also not sure if we can combine the pieces to have a per-package diagnostics rather than per-class.

mlopatkin:
Obviously, something is fishy with the implementation, but I cannot figure it yet from the test failure and I have other priorities to attend to. I move the PR to draft for now.

mlopatkin:

mlopatkin:
> Obviously, something is fishy with the implementation, but I cannot figure it yet from the test failure and I have other priorities to attend to. I move the PR to draft for now.

So the reason was a split package that had `@NullMarked` in one instance and `@GroovyNullMarked` in another. My local builds (and Teamcity) picked up the latter, so there was a suppression in the freeze file, but the GH Action picked up the former and complained about unneeded suppression in a most cryptic way possible.

Issue #34705, opened 2025-08-18
Unify locking logic in AssignImmutableWorkspaceStep


asodja:
In https://github.com/gradle/gradle/blob/15bc562deb35590c4ca4e8e0b244c7c34f454aca/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java#L144-L157 we now have two strategies to handle locking, one for Unix and one for Windows. Maybe we could unify this, to use the same strategy as for Windows also for Unix, since Unix logic has some issues, e.g.: https://github.com/gradle/gradle/issues/34585,  https://github.com/gradle/gradle/issues/28974

This requires some optimization, since otherwise we would see some performance regression. We probably need to implement logic like:

```
if (!(cleanup will be scheduled) && workspaceExists()) {
   loadWorkspace()
} else {
   lock (workspace) {
      if (workspaceExists()) {
         loadWorkspace()
      } else {
         execute()
      }
    }
}
```

The only problem is, that we also have cleanup.

delgurth:
> since Unix logic has some issues, e.g.: https://github.com/gradle/gradle/issues/34585, https://github.com/gradle/gradle/issues/28974

Not sure how you came to the conclusion that [the second issue 28974](https://github.com/gradle/gradle/issues/28974) is specific for Unix. The issue was originally reported for Windows, but in the comments people also say that it occurs on Linux build agents.

PR #36220, opened 2026-01-08
Introduce fine grained cache

asodja:
This PR introduces a new caching infrastructure called FineGrainedPersistentCache. Unlike the standard persistent cache which often relies on coarse-grained (all-or-nothing) locking of the entire cache directory, this implementation allows entry-specific locking. This will be used for the execution engine, where multiple parallel processes may need to access different parts of the cache simultaneously without blocking one another.

Key Changes
1. Core Infrastructure
Base Interface: Introduced the FineGrainedPersistentCache interface to define the contract for fine-grained locking and storage.

2. Implementation: Developed DefaultFineGrainedPersistentCache, which provides the logic for handling entry-specific locks and persistent storage.

3. Cache Factory Integration
Integrated the new fine-grained cache into the DefaultCacheFactory, allowing it to be managed and instantiated through Gradle's standard cache lifecycle.

4. Cleanup Strategy
Mark-and-Sweep Cleanup: Introduced FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup. This strategy allows for cleaning up individual cache entries that haven't been used recently without needing to lock or invalidate the entire cache.

5. Lock Validation 
Added lock validation logic to FileLock, that is used by the cache to check if acquired lock is still valid even if deleted

This PR is a base to [Unify locking logic in AssignImmutableWorkspaceStep](https://github.com/gradle/gradle/issues/34705).

Design spec:
https://docs.google.com/document/d/1E46c3M5InKqPLZZBWmEsjvMO2uz_4WOgff-pdtRsBfk

### DIFF
--- a/platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFileLockManagerContentionTest.groovy
+++ b/platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFileLockManagerContentionTest.groovy
@@ -225,7 +225,7 @@ class DefaultFileLockManagerContentionTest extends ConcurrentSpec {
         lock2.isLockFile(lockFile)
         recreationState
         lock2.state
-        recreationState != lock2.state
+        lock2.state.hasBeenUpdatedSince(recreationState) == expectedUpdated
         1 * lock1ContentionAction.accept(_) >> { FileLockReleasedSignal signal ->
             // Simulate recreate of the lock file with a different state just after lock2 signals that it wants to access the lock
             assert lockFile.delete()
@@ -251,9 +251,26 @@ class DefaultFileLockManagerContentionTest extends ConcurrentSpec {
         }
 
         where:
-        description | expectedLock2ContentionCalls | lockOptionsDescription                  | lockOptions
-        "does NOT"  | 0                            | "WITHOUT ensure file system check flag" | DefaultLockOptions.mode(Exclusive)
-        "does"      | 1                            | "WITH ensure file system check flag"    | DefaultLockOptions.mode(Exclusive).ensureAcquiredLockRepresentsStateOnFileSystem()
+        description | expectedLock2ContentionCalls | expectedUpdated | lockOptionsDescription                  | lockOptions
+        "does NOT"  | 0                            | true            | "WITHOUT ensure file system check flag" | DefaultLockOptions.mode(Exclusive)
+        "does"      | 1                            | false           | "WITH ensure file system check flag"    | DefaultLockOptions.mode(Exclusive).ensureAcquiredLockRepresentsStateOnFileSystem()
+    }
+
+    @Requires(UnitTestPreconditions.Windows)
+    def "lock file cannot be deleted while lock is held on Windows"() {
+        given:
+        def file = tmpDir.file("held.bin")
+        def lockFile = tmpDir.file("held.bin.lock")
+
+        when:
+        FileLock lock = createLock(DefaultLockOptions.mode(Exclusive), file, manager, null)
+
+        then:
+        lock
+        lock.isLockFile(lockFile)
+        lockFile.exists()
+        !lockFile.delete()
+        lockFile.exists()
     }
 
     FileLock createLock(FileLockManager.LockMode lockMode, File file, FileLockManager lockManager = manager, Consumer<FileLockReleasedSignal> whenContended = null) {


asodja:
@bot-gradle test QF

asodja:
@bot-gradle test this

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/internal/os/OperatingSystem.java
```diff
@@ -263,6 +260,16 @@ private String resolveNativePrefix() {
         public String getPathVar() {
             return "Path";
         }
+
+        private static String withExtension(String filePath, String extension) {
```
asodja, line 264: 💭 I moved this class to `stdlib-java-extensions`. This method was originally in FileUtils in base-services. Since I can't move `FileUtils` to `stdlib-java-extensions` I decided to duplicate it.

As discussed we could get rid of methods like:
- `getExecutableName`
- `getExecutableSuffix`
- `getSharedLibraryName`
- `getSharedLibrarySuffix`
- `getStaticLibraryName`
etc.

and with that also remove this method. But after starting working on this, it became bigger then this whole PR. So I can do that in a separate PR

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/FileUtils.java
```diff
@@ -164,22 +164,10 @@ public static String withExtension(String filePath, String extension) {
...
-     */
-    public static String removeExtension(String filePath) {
-        int fileNameStart = Math.max(filePath.lastIndexOf('/'), filePath.lastIndexOf('\\'));
-        int extensionPos = filePath.lastIndexOf('.');
-
-        if (extensionPos > fileNameStart) {
-            return filePath.substring(0, extensionPos);
-        }
-        return filePath;
+        int lastFileSeparator = Math.max(filePath.lastIndexOf('/'), filePath.lastIndexOf('\\'));
```
asodja, line 167: 💭 This just inlines`removeExtension` method. It's not needed anywhere else

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/filelock/DefaultLockStateSerializer.java
```diff
@@ -49,11 +49,16 @@ public void write(DataOutput dataOutput, LockState lockState) throws IOException
...
     public LockState read(DataInput dataInput) throws IOException {
-        long creationNumber = dataInput.readLong();
+        long creationNumber = readLockId(dataInput);
         long sequenceNumber = dataInput.readLong();
         return new SequenceNumberLockState(creationNumber, sequenceNumber, sequenceNumber);
     }
 
+    @Override
+    public long readLockId(DataInput dataInput) throws IOException {
+        return dataInput.readLong();
```
asodja: 💭 We could change this and use `<timestamp, random long>` or `<UUID>` as id which adds additional long write and read. But since it seems that probability of two longs colliding is very very small (5.42 * 10^-20) I didn't change that for now, I just changed so we use SplittableRandom instead of Random.

I am thinking of changing that to <timestamp, random long> just to be safe (since it's faster than UUID and still very very safe).

What do you think?

alllex: Instead of storing and reading the timestamp, can we read the file creation timestamp? It will roughly correspond to the timestamp you are proposing, because we will be generating it only on file creation.

alllex: After thinking more about it, the single long should be more than enough.

We are protecting against a rare case of lock acquisition and lock deletion happening simultaneously in different processes. More so, contending on the same cache key. This already seems very rare, would you agree?

If that's the case, then differentiating by an additional random 64-bit number seems more than enough.

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/internal/os/OperatingSystem.java
alllex: It might be good to add a note that this is a copy of another method and that it should be removed in the future in favor of using a shared helper method, once it can be moved to an accessible location.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup.java
```diff
@@ -0,0 +1,260 @@
...
+            return path.getName();
+        }
+    }
+
+    private static class MarkAndSweepCacheEntrySoftDeleter implements FineGrainedCacheEntrySoftDeleter {
+
+        private final File gcDir;
+
+        public MarkAndSweepCacheEntrySoftDeleter(FineGrainedPersistentCache cache) {
+            this.gcDir = new File(cache.getBaseDir(), ".internal/gc");
```
alllex: 💭 It's important that GC uses the same `.internal` directory that the fine-grained cache defines, so it might better to have the `.internal` as a named constant somewhere like the fine-grained cache interface.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
```diff
@@ -0,0 +1,166 @@
...
+            action.run();
+            return null;
+        });
+    }
+
+    private FileLock acquireLock(String key) {
+        FileLock lock = null;
+        while (lock == null) {
+            File lockFile = getLockFile(key);
+            lock = fileLockManager.lock(lockFile, EXCLUSIVE_LOCKING_MODE, displayName, "");
```
alllex: 💭 Should we pass some descriptive operation name for debugging purposes? I'm mostly guessing it would be useful, since the method takes the parameter. I don't know if it's actually useful in practice or if it might incur a noticeable performance penalty

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/FineGrainedPersistentCache.java
```diff
@@ -0,0 +1,78 @@
...
+ * <p>
+ * IMPORTANT: If an implementation uses file locking then it must place ALL per-key lock files under a dedicated directory located at
+ * {@code <base-dir>/.internal/locks} (see {@link #LOCKS_DIR_RELATIVE_PATH}). Each lock file must be named using the cache key plus the
+ * {@code ".lock"} suffix, i.e. {@code "<base-dir>/.internal/locks/<key>.lock"}. Cleanup strategies can then rely on this convention to locate and safely
+ * remove orphaned lock files without requiring additional API calls.
+ * Directory {@code ".internal"} can be used to store any internal metadata required by the cache implementation.
+ * </p>
+ */
+@NullMarked
+public interface FineGrainedPersistentCache extends Closeable, CleanableStore, HasCleanupAction {
```
alllex, line 68: 💭 It would be nice to have the layout of the cache on disk described in a single place. Possibly, the javadoc of this class is a good spot

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/FineGrainedPersistentCache.java
```diff
@@ -0,0 +1,78 @@
...
+package org.gradle.cache;
+
+import org.jspecify.annotations.NullMarked;
+
+import java.io.Closeable;
+import java.util.function.Supplier;
+
+/**
+ * A persistent cache that locks fine-grained: it has one lock per key instead of one lock for the whole cache.
+ *
```
alllex, line 27:
```suggestion
 * <p>
```

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultCacheFactoryTest.groovy
```diff
@@ -54,111 +57,149 @@ class DefaultCacheFactoryTest extends Specification {
 
     void "creates directory backed cache instance"() {
         when:
-        def cache = factory.open(tmpDir.testDirectory, "<display>", [prop: 'value'], mode(Shared), null, null)
+        def coarseGrainedCache = factory.open(coarseGrainedCacheDir, "<coarse>", [prop: 'value'], mode(Shared), null, null)
+        def fineGrainedCache = factory.openFineGrained(fineGrainedCacheDir, "<fine>", {  })
```
alllex: 💭 While the current approach works, I'd much prefer the tests either be separated for the two types of caches, or if we used the `where:` approach, testing different methods of the factory. The latter seems especially appealing since many tests seem to be checking generic cache behavior.

It also seems that multiple tests check for the implemtation types (like `DefaultPersistentDirectoryStore`), while it would be enough for only a single test to do so. This might help keeping most test method bodies agnostic to the type of cache they are checking.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
```diff
@@ -0,0 +1,166 @@
...
+        File baseDir,
+        String displayName,
+        FileLockManager fileLockManager,
+        FineGrainedCacheCleanupStrategy cleanupStrategy
+    ) {
+        this.baseDir = baseDir;
+        this.displayName = displayName;
+        this.fileLockManager = fileLockManager;
+        this.guard = ProducerGuard.adaptive();
+        this.internalDir = new File(baseDir, ".internal");
```
alllex: 💭 The magic constant should be part of the interface

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
```diff
@@ -0,0 +1,166 @@
...
+        this.locksDir = new File(baseDir, LOCKS_DIR_RELATIVE_PATH);
+        File gcFile = new File(internalDir, "gc.properties");
```
alllex, line 74: 💭 Since this constant is shared with the clean-up strategy, it should also be part of the interface

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
```diff
@@ -0,0 +1,166 @@
...
+        this.cleanupExecutor = new DefaultCacheCleanupExecutor(this, gcFile, cleanupStrategy.getCleanupStrategy());
+    }
+
+    @Override
+    public FineGrainedPersistentCache open() {
```
alllex, line 79: 💭 I'm getting a soft warning in IDEA that this overrides null-marked stuff, but is not null-marked. Ideally, we should mark this and all other new classes as null-marked.

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFineGrainedPersistentCacheTest.groovy
```diff
@@ -0,0 +1,209 @@
...
+        0 * cleanupAction.clean(cache, _)
+        gcFile.assertIsFile()
+
+        when:
+        markCacheForCleanup(gcFile)
+        cache.open()
+        cache.close()
+
+        then:
+        1 * cleanupAction.clean(cache, _) >> { throw new RuntimeException("Boom") }
```
alllex, line 225: ❔ Would the gc file still be modified?

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFineGrainedPersistentCacheTest.groovy
```diff
@@ -0,0 +1,209 @@
...
+        0 * _
+        gcFile.assertDoesNotExist()
+    }
+
+    def "rejects invalid keys"() {
+        when:
+        cache.useCache(key) { }
+
+        then:
+        thrown(IllegalArgumentException)
```
alllex: 💭 It's always a good idea to check for at least a part of the exception message to avoid situations when the method starts failing for a wrong reason, but the tests don't catch that.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
```diff
@@ -0,0 +1,166 @@
...
+import java.io.File;
+import java.util.Collection;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
+
+import static org.gradle.cache.FileLockManager.LockMode.Exclusive;
+
+/**
+ * An implementation of {@link FineGrainedPersistentCache}.
+ *
```
alllex, line 42:
```suggestion
 * <p>
```

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFineGrainedPersistentCacheTest.groovy
```diff
@@ -0,0 +1,209 @@
...
+    def "useCache acquires per-key exclusive lock and releases it"() {
+        given:
+        def key = "entryKey"
+        def expectedLockFile = cacheDir.file("${LOCKS_DIR_RELATIVE_PATH}/${key}.lock")
+
+        when:
+        cache.open()
+        def result = cache.useCache(key, {
+            return 42
+        } as Supplier<Integer>)
```
alllex, line 121: 💭 Would be nice to have a test for the `useCache()` with the `Runnable` argument, for completeness

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFineGrainedPersistentCacheTest.groovy
```diff
@@ -0,0 +1,209 @@
...
+
+import java.time.Instant
+import java.util.concurrent.TimeUnit
+import java.util.function.Supplier
+
+import static org.gradle.cache.FileLockManager.LockMode.Exclusive
+import static org.gradle.cache.FineGrainedPersistentCache.LOCKS_DIR_RELATIVE_PATH
+import static org.gradle.cache.internal.filelock.DefaultLockOptions.mode
+
+class DefaultFineGrainedPersistentCacheTest extends Specification {
```
alllex, line 40: ❓ Do we have tests that directly or indirectly validate that the fine-grained cache is not blocking for different keys and blocking for the same keys if multiple threads or even processes try to mutate it?

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
```diff
@@ -0,0 +1,166 @@
...
+    public DefaultFineGrainedPersistentCache(
+        File baseDir,
+        String displayName,
+        FileLockManager fileLockManager,
+        FineGrainedCacheCleanupStrategy cleanupStrategy
+    ) {
+        this.baseDir = baseDir;
+        this.displayName = displayName;
+        this.fileLockManager = fileLockManager;
+        this.guard = ProducerGuard.adaptive();
```
alllex, line 71: 🤔 The docs for this guard say that it allocates an individual lock per key. This means we'll be creating a lock for each new transform we are running or for any other immutable work. We can have tens of thousands of those in large Android builds, IIRC. Can this lead to significant additional memory pressure?

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/filelock/DefaultLockStateSerializer.java
asodja: Yes, I think it's rare case

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
asodja: We already pass displayName which is the name of a cache, but I can pass `"acquireLock"` as operation.

I think passing key might not be that useful

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup.java
asodja: Makes sense

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
asodja: Done

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
asodja: Do you mean "gc.properties"? We don't need to use the same name in clean-up, so they are not really connected directly. I just used the same name for familiarity

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/FineGrainedPersistentCache.java
asodja: Done

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/FineGrainedPersistentCache.java
asodja: Done

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultCacheFactoryTest.groovy
asodja: Change the approach and now we use `where:`

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
asodja: Fixed and marked also other classes

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFineGrainedPersistentCacheTest.groovy
asodja: It seems that is not the case, we touch gc file only after cleanup. I added a check

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFineGrainedPersistentCacheTest.groovy
asodja: Good point, added check for the message

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
asodja: Added

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
asodja: See implementation:
https://github.com/gradle/gradle/blob/b3a8557bf0bd98a6a8247dfbf046fdd11c147fdd/platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/ProducerGuard.java#L82-L105

It locks a key only for the time of execution of supplier. And then it removes the key. So we will never have `tens of thousands` locks/entries at a time, since we can't execute that many transforms at once. There is no memory issue for this.

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFineGrainedPersistentCacheTest.groovy
asodja: Added thread concurrency tests as DefaultFineGrainedPersistentCacheConcurrencyTest

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFineGrainedPersistentCacheTest.groovy
asodja: Done

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/LockOptions.java
```diff
@@ -21,6 +21,27 @@ public interface LockOptions {
 
     boolean isUseCrossVersionImplementation();
 
+    /**
+     * Ensures that the acquired lock represents a file state on the file system.
+     *
+     * This safeguard is necessary when a Gradle process is allowed to delete a
+     * lock file while holding it.
```
alllex: 💭 I'd add that this is useful for the safe lock clean-up scenarios. Otherwise, without context, it's not immediately clear why would we do this.

Maybe also useful to note that this is not possible on all operating systems.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/LockOptions.java
```diff
@@ -21,6 +21,27 @@ public interface LockOptions {
...
+     *
+     * <p>
+     * Example Scenario:
+     * <ol>
+     * <li>Process 1 acquires the lock. Process 2 attempts to acquire it and waits.</li>
+     * <li>Process 1 deletes the lock file from the disk, then releases the lock.</li>
+     * <li>Process 2 serves the wait and successfully acquires the lock.</li>
```
alllex: 💅 I am not sure if I understand the word `serves` in this sentence

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFileLockManager.java
```diff
@@ -172,7 +210,8 @@ public DefaultFileLock(File target, LockOptions options, String displayName, Str
             }
 
             LockStateSerializer stateProtocol = options.isUseCrossVersionImplementation() ? new Version1LockStateSerializer() : new DefaultLockStateSerializer();
-            lockFileAccess = new LockFileAccess(lockFile, new LockStateAccess(stateProtocol));
+            this.lockStateAccess = new LockStateAccess(stateProtocol, () -> isFirstLockAccess.set(true));
```
alllex, line 213: 💭 I think we can simplify this code further by avoiding the need for `isFirstLockAccess`.

Since we actually perform the full locking logic immediately below, we get back the initial `LockState`. That state can already tell us whether it was the first access via `LockState#isInInitialState`.

The place where we currently call the new `onFirstFileLockAccessAction` callback explicitly creates `protocol.createInitialState()`. That means calling `isInInitialState` will return `true` as we expect. Further, after the full operation under the lock has been performed, it seems we are [unconditionally calling](https://github.com/gradle/gradle/blob/44672211e8e6a2c340cedbd0369e8811342483c2/platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFileLockManager.java#L273) `markClean()`, which updates the state to indicate it's not longer initial. Thus, subsequent reads of the lock should return `isInInitialState == false`. Overall, this seems to be the intent behind the method anyway.

---

We can have a final boolean field instead, which we populate at the end of this constructor call:

```java
isFirstLockAccess = lockState.isInInitialState();
```

---

There is, technically, one case when `isInIntialState()` will return `true`, when it, perhaps, shouldn't. It's when Process 1 created the lock, started the operation, but then the operation failed. It means `markClean` wasn't called. So when Process 2 reads the state from disk, it will read `sequenceNumber=0`, which in the current implementation of `readState()` means `originalSequenceNumber` will also be `0`, which means `isInInitialState()` will return `true` even though the lock was not created but read. 

Having said that, this is not a problem for my proposal above. We only intended to use that boolean as a small optimization, so it's okay that it is not accurate in that very rare case of the Process 1 action failure. The lock read from disk will be `isDirty` in that case.

TL;DR let's avoid the need for the new callback.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFileLockManager.java
```diff
@@ -302,6 +342,32 @@ public LockMode getMode() {
...
+                // On Windows, if we hold the lock it means lock was not updated on the file system,
+                // since OS won't allow to modify the state region or delete the file due to lock
+                return lock == null;
+            }
+
+            if (lock == null || !lockFile.exists()) {
+                return true;
+            }
+
+            try (RandomAccessFile randomAccessFile = new RandomAccessFile(lockFile, "r")) {
```
alllex, line 361: ❔ Is it worth adding a comment that it's important we create a new `RandomAccessFile` instead of using the existing `lockFileAccess` to re-read the lock state? It would be helpful for me personally, but I'm also okay with not adding it, if you think it's trivial

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/FineGrainedMarkAndSweepCacheCleanupStrategy.java
```diff
@@ -0,0 +1,50 @@
...
+ * A {@link FineGrainedCacheCleanupStrategy} that uses a {@link FineGrainedCacheEntrySoftDeleter} to determine which entries are soft deleted.
+ *
+ * This strategy first marks entries as soft deleted and after a {@link FineGrainedMarkAndSweepCacheCleanupStrategy#SOFT_DELETION_DURATION} period of time hard deletes them.
+ * The process that reads/writes to cache, can use {@link FineGrainedCacheEntrySoftDeleter#isSoftDeleted(String)} to determine which entries are soft deleted.
+ * Soft deleted entries should then not be read without the lock.
+ * The process can remove the soft delete marker via {@link FineGrainedCacheEntrySoftDeleter#removeSoftDeleteMarker(String)} if it determines the entry should not be hard deleted.
+ */
+public interface FineGrainedMarkAndSweepCacheCleanupStrategy extends FineGrainedCacheCleanupStrategy {
+
+    Duration SOFT_DELETION_DURATION = Duration.ofHours(6);
```
alllex, line 52: 💭Since it's a hardcoded constant, I think it's best if we add reasoning about why this value was chosen

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup.java
```diff
@@ -0,0 +1,263 @@
...
+                // And finally delete also a lock file
+                FileUtils.deleteQuietly(getLockFile(key));
+                return true;
+            }
+            return false;
+        }
+
+        private void cleanupOrphanGcDirsAndFileLocks() {
+            Set<String> entryKeys = listEntryKeys(cache.getBaseDir(),
+                file -> !cache.getReservedCacheFiles().contains(file),
```
alllex: 💅 We are re-creating the reserved cache files collection for the check of every file.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup.java
```diff
@@ -0,0 +1,263 @@
...
+                Function.identity()
+            );
+            Set<String> locksKeys = listEntryKeys(getLocksDir(),
+                file -> file.getName().endsWith(".lock"),
+                name -> name.substring(0, name.lastIndexOf(".lock"))
+            );
+            Set<String> gcDirsKeys = listEntryKeys(softDeleter.getGcDir(), File::isDirectory, Function.identity());
+
+            Set<String> orphanKeys = Sets.difference(Sets.union(locksKeys, gcDirsKeys), entryKeys);
```
alllex, line 157: ❔ We find orphan locks and gc-dirs. Can there be orphan key entries without existing locks?

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup.java
```diff
@@ -0,0 +1,263 @@
...
+                // Then delete gc information
+                FileUtils.deleteQuietly(softDeleter.getKeyGcDir(key));
+                // And finally delete also a lock file
+                FileUtils.deleteQuietly(getLockFile(key));
+                return true;
+            }
+            return false;
+        }
+
+        private void cleanupOrphanGcDirsAndFileLocks() {
```
alllex, line 145: 🤔 This clean up runs after the main build finishes, but the current implementation requires to materialize the collection of all existing cache entries, which can be pretty large. Is it worth considering making this whole clean up in a streaming fashion?

We could stream through existing locks, extract the key and check if the corresponding entry exists. If it doesn't we can immediately delete the orphaned lock. Same with gc dirs.

I don't suggest we do this in the current PR, but this can be a follow-up to consider.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFineGrainedPersistentCache.java
alllex: I see now. With shared constants extracted, it's now clearer that these are indeed independent from the clean-up strategy.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup.java
```diff
@@ -0,0 +1,263 @@
...
+            Set<String> entryKeys = listEntryKeys(cache.getBaseDir(),
+                file -> !cache.getReservedCacheFiles().contains(file),
+                Function.identity()
```
alllex, line 149: ❓ The `FineGrainedPersistentCache` javadoc says that the entries key be both files and directories, and it says that if they are files they can be called `<key.bin>`. I am not sure if that's correct, but if it is, then we are removing the `.bin` suffices, while we should

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/FineGrainedPersistentCache.java
alllex: Very nice docs!

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup.java
asodja: Yeah, I think I can remove this and I can say that it's expected that entries are directories. And if we ever need just files as entries we can change that

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/LockOptions.java
asodja: I changed it to only say:
```
This safeguard is necessary for safe lock cleanup.
```

I think + example should make it clear why it's used.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/LockOptions.java
asodja: Changed just to `Process 2 then acquires the lock.`

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFileLockManager.java
asodja: > Further, after the full operation under the lock has been performed, it seems we are [unconditionally calling](https://github.com/gradle/gradle/blob/44672211e8e6a2c340cedbd0369e8811342483c2/platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFileLockManager.java#L273) markClean(), which updates the state to indicate it's not longer initial.

We actually don't call `FileLock.writeFile()` in our case, that is why I added a new callback. So if we would want to use this `isInIntialState()` we would need to call `FileLock.writeFile()` or a similar method somewhere. I am not sure if we can safely do that from multiple threads

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFileLockManager.java
asodja: Added a comment

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/FineGrainedMarkAndSweepCacheCleanupStrategy.java
asodja: Added reasoning

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup.java
asodja: Good point, moved to a variable

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup.java
asodja: There can be, but that is not a problem. File lock is always created if needed automatically

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup.java
asodja: Right now this solution optimizes  "I/O" performance, since listing all directories at once is faster than calling file.exists() for every directory.

But if there is a problem we can also optimize it for memory. So we keep less things in memory. So we can do changes if there will be issues.

asodja:
@bot-gradle test this

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFileLockManagerContentionTest.groovy
```diff
@@ -147,8 +150,118 @@ class DefaultFileLockManagerContentionTest extends ConcurrentSpec {
...
+
+        when:
+        def lock2 = createLock(lockOptions, file, manager2, lock2ContentionAction)
+
+        then:
+        lock2
+        lock2.isLockFile(lockFile)
+        recreationState
+        lock2.state
+        recreationState != lock2.state
```
alllex: ❓ Is this meant to test referential equality? The `SequenceNumberLockState` class does not implement `equals`. Running the `WITH` variant of the test, I see that the creation numbers for the states in this line are the same. Is this expected?

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFineGrainedPersistentCacheConcurrencyTest.groovy
```diff
@@ -0,0 +1,207 @@
...
+                allowFinish.await(5, TimeUnit.SECONDS)
+            }
+        }, "fg-cache-t1")
+
+        Thread t2 = new Thread({
+            cache.useCache(key2) {
+                bothStarted.countDown()
+                // wait until both threads started
+                bothStarted.await(5, TimeUnit.SECONDS)
+                overlapped.set(true)
```
alllex, line 84: 💅 I find it strange that both threads saturate the same flag. Maybe we should have two booleans to indicate both threads observed the overlap?

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFineGrainedPersistentCacheConcurrencyTest.groovy
```diff
@@ -0,0 +1,207 @@
...
+import spock.lang.AutoCleanup
+import spock.lang.Specification
+import spock.lang.Subject
+
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
+
+class DefaultFineGrainedPersistentCacheConcurrencyTest extends Specification {
```
alllex, line 33: 👏 Great tests!

alllex (approved):
I believe all important concerns were addressed. Great work!

The test coverage is particularly nice 🧑‍🍳

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/DefaultFileLockManagerContentionTest.groovy
asodja: Good catch. Behaviour is correct but this check is not.

If some other process recreates the file lock, lock2 should pick up the recreated file lock. So for "WITH" case this state should be equal. We can probably check with `!recreationState.hasBeenUpdatedSince(lock2.state)`

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FineGrainedMarkAndSweepLeastRecentlyUsedCacheCleanup.java
alllex: Ah, because the clean strategy would attempt to delete any old enough file/directory in the base dir, even if there are no corresponding locks/gc-files, right?

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFileLockManager.java
alllex: I'd love to follow up on that. If we don't call updateFile or writeFile after acquiring the lock, then the lock we've just created will remain in the dirty state. That is because `createInitialState()` for our type of lock always returns the state that would be identified as `isDirty()`.

Maybe we don't use the dirtiness property in this context anywhere, but this still sounds fishy. Something that I think is at least worth an explanation in the code (as comments, probably).

---

The other thing we can replace the new callback with (I'm really not a fan of it, though it does the job! 😅) is another getter on the LockState that basically checks if the `originalSequenceNumber < 0` (note - strictly). This will be equivalent to the new callback, because the initial state, created at runtime, always has `originalSequenceNumber == -1`, while anything that is written to disk will have a value of at least `0` or higher.

asodja: As far as I understand this logic, you have to call fileLock.writeFile/fileLock.updateFile to update this sequence or clean dirtiness. This seems tightly related to indexed cache and some other file based caches, where we call such methods.
But in lots of cases this is not even called (even for coarse grained caches). And our case is one of such cases. 

So I will this for now, I would need more time and understanding to figure out how to reuse this logic.

asodja:
@bot-gradle test this

PR #36221, opened 2026-01-08
Reintroduce "Fix exclude simplification logic (#36026)""

bamboo:
Reverts gradle/gradle#36217

The original PR (#36026) exposed a defect in `PersistentMap#modify` which has been address in #36224.


bamboo:
@bot-gradle test APT

Issue #34585, opened 2025-08-07
Docker: Could not move inconsistent immutable workspace


akharlamov:
### Current Behavior

I observe an issue similar to https://github.com/gradle/gradle/issues/31438 but inside a docker container on Ubuntu 22. 

```
m@b95281bd2cd3:~$ ./gradlew --offline --stacktrace mcpServer -PtargetProjectDir=/app/sources/jingle
Calculating task graph as no cached configuration is available for tasks: mcpServer

[Incubating] Problems report is available at: file:///app/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* Where:
Build file '/app/build.gradle.kts' line: 138

* What went wrong:
Failed to query the value of task ':prepareSandbox' property 'sandboxConfigDirectory'.
> Failed to query the value of task ':prepareSandbox' property 'sandboxDirectory'.
   > A build operation failed.
         Could not move inconsistent immutable workspace (/app/.gradle/caches/8.14/transforms/3b2d455af66406b61ef3b1a0ba7fde4e) to temporary location (/app/.gradle/caches/8.14/transforms/3b2d455af66406b61ef3b1a0ba7fde4e-e9295ee8-c0d8-42ee-bb4a-cb05bb560ad4)
      > Could not move inconsistent immutable workspace (/app/.gradle/caches/8.14/transforms/3b2d455af66406b61ef3b1a0ba7fde4e) to temporary location (/app/.gradle/caches/8.14/transforms/3b2d455af66406b61ef3b1a0ba7fde4e-e9295ee8-c0d8-42ee-bb4a-cb05bb560ad4)

* Try:
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

* Exception is:
org.gradle.api.internal.provider.AbstractProperty$PropertyQueryException: Failed to query the value of task ':prepareSandbox' property 'sandboxConfigDirectory'.
        at org.gradle.api.internal.provider.AbstractProperty.doCalculateValue(AbstractProperty.java:175)
        at org.gradle.api.internal.provider.AbstractProperty.calculateOwnValue(AbstractProperty.java:165)
        at org.gradle.api.internal.provider.AbstractMinimalProvider.calculateValue(AbstractMinimalProvider.java:117)
        at org.gradle.api.internal.provider.TransformBackedProvider.calculateOwnValue(TransformBackedProvider.java:82)
[...]

Issue #28974, opened 2024-04-26
Transforms cache problems "Could not read workspace metadata"


loetifuss:
### Current Behavior

On our Jenkins agents running on Windows we're having trouble with corrupted build cache entries. Every couple of builds agents start to complain about missing "metadata.bin" files in the build cache. Looking at the cache directory the files are indeed missing, there is just a subdirectory "**transformed/instrumented**" containing a Jar.

Subsequent builds will fail once this error shows up. The only solution is to delete the cache directory. Then it works for a while until the error shows up again. So far we have only seen this issue on our Windows agents.

Searching online, we found other people seem to have the same issue since updating to 8.7: https://github.com/ankidroid/Anki-Android/issues/16161

```
08:38:23  Starting a Gradle Daemon, 2 incompatible and 2 stopped Daemons could not be reused, use --status for details
08:38:27  Calculating task graph as no cached configuration is available for tasks: unitTest dashboardAll
08:38:29  
08:38:29  FAILURE: Build failed with an exception.
08:38:29  
08:38:29  * What went wrong:
08:38:29  Multiple build operations failed.
08:38:29      Could not read workspace metadata from C:\Users\jenkins\.gradle\caches\transforms-4\488376c3749de4c34a681b5fd5a06aa8\metadata.bin
08:38:29      Could not read workspace metadata from C:\Users\jenkins\.gradle\caches\transforms-4\e2dc2f44e335b5e104801cc73d245ee1\metadata.bin
08:38:29  > Could not read workspace metadata from C:\Users\jenkins\.gradle\caches\transforms-4\488376c3749de4c34a681b5fd5a06aa8\metadata.bin
08:38:29  > Could not read workspace metadata from C:\Users\jenkins\.gradle\caches\transforms-4\e2dc2f44e335b5e104801cc73d245ee1\metadata.bin
08:38:29  
08:38:29  * Try:
08:38:29  > Run with --info or --debug option to get more log output.
08:38:29  > Run with --scan to get full insights.
08:38:29  > Get more help at https://help.gradle.org.
08:38:29  
08:38:29  * Exception is:
08:38:29  org.gradle.internal.operations.MultipleBuildOperationFailures: Multiple build operations failed.
08:38:29      Could not read workspace metadata from C:\Users\jenkins\.gradle\caches\transforms-4\488376c3749de4c34a681b5fd5a06aa8\metadata.bin
[...]

PR #36227, opened 2026-01-09
Unify AssignImmutableWorkspaceStep locking logic by using FineGrainedPersistentCache

asodja:
This PR unifies the locking logic within the AssignImmutableWorkspaceStep by utilizing the FineGrainedPersistentCache.

Prior to this change, we used the "atomic-move" semantics on Unix and not most performant locking solution on Windows. While atomic-move semantics was fast we had some issues like https://github.com/gradle/gradle/issues/34585, https://github.com/gradle/gradle/issues/28974 and with Windows solution there was a performance hit.

This solution tries to be as fast as atomic-move (no file lock is used when workspace is there) while tries to be as safe as Windows solution (uses file locking when workspace is not there).

Key Changes
- Unified Locking: It migrates the locking logic of AssignImmutableWorkspaceStep  to use FineGrainedPersistentCache. This ensures that when multiple processes attempt to compute a workspace with the same unique ID, the file lock is used. Reads are done without file locks and writes are done with a file lock 

- Concurrency Optimization: It introduces `getIfRunningOrCompute` logic. This ensures that only one thread performs the actual work for a specific workspace key at any given time, while other threads wait for the result or map to the concurrent result.

- Integration Test Fixes: Because this change touches core execution and caching logic, several integration tests had to be updated to account for the new locking behavior and cleanup, including:
    - ConcurrentBuildsArtifactTransformIntegrationTest
    - KotlinScriptCacheCleanupIntegrationTest
    - CrossBuildScriptCachingIntegrationSpec
    - AssignImmutableWorkspaceStepConcurrencyTest

Fixes https://github.com/gradle/gradle/issues/34705, https://github.com/gradle/gradle/issues/34585, https://github.com/gradle/gradle/issues/28974

### DIFF
--- a/platforms/core-configuration/kotlin-dsl/src/integTest/groovy/org/gradle/kotlin/dsl/KotlinScriptCacheCleanupIntegrationTest.groovy
+++ b/platforms/core-configuration/kotlin-dsl/src/integTest/groovy/org/gradle/kotlin/dsl/KotlinScriptCacheCleanupIntegrationTest.groovy
@@ -16,7 +16,6 @@
 
 package org.gradle.kotlin.dsl
 
-
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.integtests.fixtures.UnsupportedWithConfigurationCache
 import org.gradle.integtests.fixtures.cache.FileAccessTimeJournalFixture
--- a/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
+++ b/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
@@ -16,7 +16,6 @@
 
 package org.gradle.internal.execution.steps;
 
-import com.google.common.base.Function;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableSortedMap;
@@ -30,6 +29,7 @@ import org.gradle.internal.execution.history.ImmutableWorkspaceMetadata;
 import org.gradle.internal.execution.history.ImmutableWorkspaceMetadataStore;
 import org.gradle.internal.execution.history.impl.DefaultExecutionOutputState;
 import org.gradle.internal.execution.workspace.ImmutableWorkspaceProvider;
+import org.gradle.internal.execution.workspace.ImmutableWorkspaceProvider.ConcurrentResult;
 import org.gradle.internal.execution.workspace.ImmutableWorkspaceProvider.ImmutableWorkspace;
 import org.gradle.internal.file.Deleter;
 import org.gradle.internal.hash.HashCode;
@@ -106,44 +106,51 @@ public class AssignImmutableWorkspaceStep<C extends IdentityContext> implements 
         String uniqueId = context.getIdentity().getUniqueId();
 
         ImmutableWorkspace workspace = workspaceProvider.getWorkspace(uniqueId);
-        // Gets a workspace result or compute it,
-        // but only computes it once across all threads that try to compute it at the same time
-        return workspace.getIfRunningOrCompute(concurrentResultMapper(work, workspace), () -> {
-            // If a workspace already exists, just return the result
-            WorkspaceLoad initialLoad = loadImmutableWorkspaceIfExists(work, workspace);
-            if (initialLoad.isSuccess()) {
-                return initialLoad.getResult();
-            }
+        // Loads a workspace result or creates it,
+        // but only execute loadOrCreateWorkspace() once across all threads that try to run it at the same time
+        ConcurrentResult<WorkspaceResult> result = workspace.getOrCompute(() -> loadOrCreateWorkspace(work, workspace, context));
 
-            // If the workspace doesn't exist or is soft-deleted or broken, execute the work with file lock
-            return workspace.withFileLock(() -> {
-                // We need to invalidate snapshots in case another process populated the workspace just before us
-                Runnable invalidateSnapshots = () -> fileSystemAccess.invalidate(ImmutableList.of(workspace.getImmutableLocation().getAbsolutePath()));
-                // First, try to load it again if another process populated it just before us (double-checked locking)
-                WorkspaceLoad load = loadImmutableWorkspaceIfComplete(work, workspace, invalidateSnapshots);
-
-                WorkspaceResult result = load.isSuccess()
-                    ? load.getResult()
-                    : executeInWorkspace(work, context, workspace);
-
-                if (initialLoad.isSoftDeleted()) {
-                    workspace.ensureUnSoftDeleted();
-                }
-                return result;
-            });
-        });
+        // If a result is produced by the current thread, return it, otherwise map it as up-to-date
+        return result.isProducedByCurrentThread()
+            ? result.get()
+            : mapConcurrentResultToUpToDate(result.get(), work, workspace);
     }
 
-    private static Function<WorkspaceResult, WorkspaceResult> concurrentResultMapper(UnitOfWork work, ImmutableWorkspace workspace) {
-        return result -> {
-            if (result.getExecution().isSuccessful()) {
-                @SuppressWarnings("OptionalGetWithoutIsPresent")
-                ExecutionOutputState executionOutputState = result.getAfterExecutionOutputState().get();
-                return getUpToDate(work, workspace.getImmutableLocation(), executionOutputState.getOutputFilesProducedByWork(), executionOutputState.getOriginMetadata());
-            } else {
-                return new WorkspaceResult(result, null);
+    private static WorkspaceResult mapConcurrentResultToUpToDate(WorkspaceResult result, UnitOfWork work, ImmutableWorkspace workspace) {
+        if (result.getExecution().isSuccessful()) {
+            @SuppressWarnings("OptionalGetWithoutIsPresent")
+            ExecutionOutputState executionOutputState = result.getAfterExecutionOutputState().get();
+            return getUpToDate(work, workspace.getImmutableLocation(), executionOutputState.getOutputFilesProducedByWork(), executionOutputState.getOriginMetadata());
+        } else {
+            return new WorkspaceResult(result, null);
+        }
+    }
+
+    private WorkspaceResult loadOrCreateWorkspace(UnitOfWork work, ImmutableWorkspace workspace, C context) {
+        // First try to load a workspace and if it's already there and complete return the result
+        WorkspaceLoad initialLoad = loadImmutableWorkspaceIfExists(work, workspace);
+        if (initialLoad.isSuccess()) {
+            return initialLoad.getResult();
+        }
+
+        // If the workspace doesn't exist or is soft-deleted or broken, execute the work with a file lock
+        return workspace.withFileLock(() -> {
+            // We need to invalidate snapshots in case another process populated the workspace just before us
+            Runnable invalidateSnapshots = () -> fileSystemAccess.invalidate(ImmutableList.of(workspace.getImmutableLocation().getAbsolutePath()));
+            // First, try to load it again if another process populated it just before us (double-checked locking)
+            WorkspaceLoad load = loadImmutableWorkspaceIfComplete(work, workspace, invalidateSnapshots);
+
+            WorkspaceResult result = load.isSuccess()
+                ? load.getResult()
+                : executeInWorkspace(work, context, workspace);
+
+            // Only un soft-delete if soft-deleted at initial load,
+            // if an entry was soft-deleted while acquiring lock, we leave it to the next read to handle it
+            if (initialLoad.isSoftDeleted()) {
+                workspace.ensureUnSoftDeleted();
             }
-        };
+            return result;
+        });
     }
 
     private WorkspaceLoad loadImmutableWorkspaceIfExists(UnitOfWork work, ImmutableWorkspace workspace) {
--- a/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/ImmutableWorkspaceProvider.java
+++ b/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/ImmutableWorkspaceProvider.java
@@ -17,18 +17,16 @@
 package org.gradle.internal.execution.workspace;
 
 import java.io.File;
-import java.util.function.Function;
 import java.util.function.Supplier;
 
 public interface ImmutableWorkspaceProvider {
-    ImmutableWorkspace getWorkspace(String path);
+    ImmutableWorkspace getWorkspace(String workId);
 
     interface ImmutableWorkspace {
         /**
          * Immutable workspace outputs location.
          *
-         * For {@link LockingImmutableWorkspace} this will normally be $GRADLE_USER_HOME/caches/transforms/[gradle-version]/[hash]/workspace/
-         *     and for {@link AtomicMoveImmutableWorkspace} this will normally be $GRADLE_USER_HOME/caches/transforms/[gradle-version]/[hash]/
+         * This will normally be $GRADLE_USER_HOME/caches/[gradle-version]/[cache-name]/[hash]/
          */
         File getImmutableLocation();
 
@@ -40,13 +38,9 @@ public interface ImmutableWorkspaceProvider {
         /**
          * Gets a result from the workspace if it's already running or computes it otherwise.
          *
-         * This method makes sure only one thread is executing the given action for Workspace at a time.
-         *
-         * <p>
-         * If a result from another thread already exists then {@code concurrentResultMapper} is called, otherwise {@code action} is executed.
-         * </p>
+         * This method makes sure only one thread is executing the given action for a workspace at a time.
          */
-        <T> T getIfRunningOrCompute(Function<T, T> concurrentResultMapper, Supplier<T> action);
+        <T> ConcurrentResult<T> getOrCompute(Supplier<T> action);
 
         /**
          * Returns true if the workspace has been soft deleted.
@@ -54,37 +48,34 @@ public interface ImmutableWorkspaceProvider {
         boolean isSoftDeleted();
 
         /**
-         * Remove soft deletion marker, which means entry won't be deleted anymore.
+         * Remove a soft deletion marker, which means entry won't be deleted anymore.
          */
         void ensureUnSoftDeleted();
     }
 
-    /**
-     * A workspace that relies on locking to ensure that only one process can access it at a time.
-     * Used on Windows where atomic moves cause issues with file locking.
-     */
-    interface LockingImmutableWorkspace extends ImmutableWorkspace {
+    class ConcurrentResult<T> {
 
-        /**
-         * Executes the given action under the global scoped lock.
-         */
-        <T> T withWorkspaceLock(Supplier<T> supplier);
-    }
+        private final T value;
+        private final boolean isProducedByCurrentThread;
 
-    /**
-     * A workspace that relies on atomic moves of immutable workspace directory.
-     * Used on Unix-like systems where atomic moves are supported.
-     */
-    interface AtomicMoveImmutableWorkspace extends ImmutableWorkspace {
+        private ConcurrentResult(T value, boolean isProducedByCurrentThread) {
+            this.value = value;
+            this.isProducedByCurrentThread = isProducedByCurrentThread;
+        }
 
-        /**
-         * Provides a temporary workspace and executes the given action in it.
-         */
-        <T> T withTemporaryWorkspace(TemporaryWorkspaceAction<T> action);
+        public T get() {
+            return value;
+        }
+        public boolean isProducedByCurrentThread() {
+            return isProducedByCurrentThread;
+        }
 
-        @FunctionalInterface
-        interface TemporaryWorkspaceAction<T> {
-            T executeInTemporaryWorkspace(File temporaryWorkspaceLocation);
+        public static <T> ConcurrentResult<T> producedByCurrentThread(T value) {
+            return new ConcurrentResult<>(value, true);
+        }
+
+        public static <T> ConcurrentResult<T> producedByOtherThread(T value) {
+            return new ConcurrentResult<>(value, false);
         }
     }
 }
--- a/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/impl/CacheBasedImmutableWorkspaceProvider.java
+++ b/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/impl/CacheBasedImmutableWorkspaceProvider.java
@@ -16,6 +16,7 @@
 
 package org.gradle.internal.execution.workspace.impl;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.gradle.api.internal.cache.CacheConfigurationsInternal;
 import org.gradle.cache.FineGrainedCacheBuilder;
 import org.gradle.cache.FineGrainedCacheCleanupStrategyFactory;
@@ -25,6 +26,7 @@ import org.gradle.cache.FineGrainedPersistentCache;
 import org.gradle.internal.Cast;
 import org.gradle.internal.execution.workspace.ImmutableWorkspaceProvider;
 import org.gradle.internal.file.FileAccessTimeJournal;
+import org.gradle.internal.file.FileAccessTracker;
 import org.gradle.internal.file.impl.SingleDepthFileAccessTracker;
 
 import java.io.Closeable;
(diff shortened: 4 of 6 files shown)


asodja:
@bot-gradle test QF

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
```diff
@@ -66,197 +60,167 @@
...
-            );
-        } else {
-            AtomicMoveImmutableWorkspace workspace = workspaceProvider.getAtomicMoveWorkspace(uniqueId);
-            return loadImmutableWorkspaceIfExists(work, workspace)
-                .orElseGet(() -> executeInTemporaryWorkspace(work, context, workspace));
-        }
+        ImmutableWorkspace workspace = workspaceProvider.getWorkspace(uniqueId);
+        // Gets a workspace result or compute it,
+        // but only computes it once across all threads that try to compute it at the same time
+        return workspace.getIfRunningOrCompute(concurrentResultMapper(work, workspace), () -> {
```
asodja: 💭 This is the core of the change

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/impl/CacheBasedImmutableWorkspaceProvider.java
```diff
@@ -17,173 +17,139 @@
...
+        }
 
-    @Override
-    public void close() {
-        keyCaches.forEach((id, cache) -> cache.close());
-        cache.close();
-    }
+        @Override
+        public <T> T getIfRunningOrCompute(Function<T, T> concurrentResultMapper, Supplier<T> action) {
+            CompletableFuture<T> thisOperationFuture = new CompletableFuture<>();
```
asodja: 💭 This logic makes sure only one thread does the work for the same key concurrently.

platforms/core-execution/execution/src/test/groovy/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStepConcurrencyTest.groovy
```diff
@@ -16,174 +16,144 @@
...
     def work2Started = new CountDownLatch(1)
 
-    def "atomic move strategy handles race condition by returning earlier execution as up-to-date and discarding temporary workspace of the later one"() {
-        def step = new AssignImmutableWorkspaceStep(deleter, fileSystemAccess, immutableWorkspaceMetadataStore, outputSnapshotter, delegate, LockingStrategy.ATOMIC_MOVE)
-        Map<Thread, Throwable> exceptions = [:]
-        def exceptionHandler = { thread, exception ->
-            exceptions.put(thread, exception)
-            exception.printStackTrace()
-        } as Thread.UncaughtExceptionHandler
+    def "only one thread executes work concurrently"() {
```
asodja, line 92: 💭 This changes a test so we test "only one thread executes work concurrently"

platforms/core-execution/execution/src/test/groovy/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStepWithAtomicMoveTest.groovy
```diff
@@ -1,305 +0,0 @@
...
-import java.nio.file.Files
-import java.nio.file.Paths
-import java.time.Duration
-
-import static org.gradle.internal.execution.Execution.ExecutionOutcome.UP_TO_DATE
-import static org.gradle.internal.execution.steps.AssignImmutableWorkspaceStep.LockingStrategy
-import static org.gradle.internal.execution.workspace.ImmutableWorkspaceProvider.AtomicMoveImmutableWorkspace
-import static org.gradle.internal.execution.workspace.ImmutableWorkspaceProvider.AtomicMoveImmutableWorkspace.TemporaryWorkspaceAction
-
-class AssignImmutableWorkspaceStepWithAtomicMoveTest extends StepSpec<IdentityContext> implements TestSnapshotFixture {
```
asodja, line 50: 💭 This file was deleted, since we don't have atomic move semantics anymore

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformCachingIntegrationTest.groovy
```diff
@@ -900,50 +898,6 @@ class ArtifactTransformCachingIntegrationTest extends AbstractHttpDependencyReso
         output.count("Transformed") == 0
     }
 
-    @Requires(UnitTestPreconditions.NotWindows)
-    def "lock-free transform is run again and old output is removed after it failed in previous build"() {
```
asodja, line 904: 💭 This test was deleted since we don't have atomic move semantics anymore

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformCachingIntegrationTest.groovy
```diff
@@ -1362,8 +1316,50 @@ class ArtifactTransformCachingIntegrationTest extends AbstractHttpDependencyReso
         succeeds ":util:resolve"
     }
 
+    @Issue("https://github.com/gradle/gradle/issues/28974")
+    def "non-incremental transform can recover if metadata is #description"() {
```
asodja, line 1320: 💭 Additional test to check that if metadata is removed or broken we can recover

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformCachingIntegrationTest.groovy
```diff
@@ -2062,28 +2059,46 @@ resultsFile:
...
         // start as new process so journal is not restored from in-memory cache
         executer.withTasks("help").start().waitForFinish()
 
         then:
-        outputDir1.assertDoesNotExist()
-        outputDir2.assertExists()
-        gcFile.lastModified() >= SECONDS.toMillis(beforeCleanup)
+        workspace1.assertExists()
+        workspace2.assertExists()
+        isSoftDeleted(workspace1.name)
```
asodja, line 2083: 💭 We have update this test, since we now do a soft-deletion first

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ConcurrentBuildsArtifactTransformIntegrationTest.groovy
```diff
@@ -155,71 +153,6 @@ tasks.block2.mustRunAfter tasks.blueThings
...
 
-    /**
-     * With current cache implementation we don't use locking for cache entries (Gradle 8.6+)
-     * we run transform in temporary directory and we then use atomic move to move it to the final location.
-     * Due to that we can have multiple builds running the same transform at the same time, but we will always have just one result in the final cache location.
-     *
-     * Additional note: Windows and Unix have different locking logic from 9.1.0.
-     */
-    @Requires(UnitTestPreconditions.Unix)
-    def "concurrent builds can transform the same file at the same time"() {
```
asodja, line 166: 💭 Removed this test, since there is no more atomic move.

`file is transformed once only by concurrent builds` tests behaviour for all platforms now

subprojects/core/src/integTest/groovy/org/gradle/api/CrossBuildScriptCachingIntegrationSpec.groovy
```diff
@@ -55,7 +54,6 @@ class CrossBuildScriptCachingIntegrationSpec extends AbstractIntegrationSpec {
         remappedCachesDir = new File(versionCaches, 'scripts-remapped')
     }
 
-    @ToBeFixedForIsolatedProjects(because = "Expects sequential script compilation")
```
asodja, line 58: 💭 Fixed for isolated projects, since we now allow only one thread to do the same work concurrently

platforms/core-configuration/kotlin-dsl/src/integTest/groovy/org/gradle/kotlin/dsl/KotlinScriptCacheCleanupIntegrationTest.groovy
```diff
@@ -16,21 +16,29 @@
 
 package org.gradle.kotlin.dsl
 
+
```
alllex: 💅  extra line?

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/ImmutableWorkspaceProvider.java
```diff
@@ -31,6 +31,32 @@ interface ImmutableWorkspace {
          *     and for {@link AtomicMoveImmutableWorkspace} this will normally be $GRADLE_USER_HOME/caches/transforms/[gradle-version]/[hash]/
          */
         File getImmutableLocation();
```
alllex, line 31: ❌ The javadoc for this method is no longer accurate, and we should also remove the interfaces for the different types of immutable workspaces, since they are no longer used.

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/ImmutableWorkspaceProvider.java
```diff
@@ -31,6 +31,32 @@ interface ImmutableWorkspace {
...
+        /**
+         * Gets a result from the workspace if it's already running or computes it otherwise.
+         *
+         * This method makes sure only one thread is executing the given action for Workspace at a time.
+         *
+         * <p>
+         * If a result from another thread already exists then {@code concurrentResultMapper} is called, otherwise {@code action} is executed.
+         * </p>
+         */
+        <T> T getIfRunningOrCompute(Function<T, T> concurrentResultMapper, Supplier<T> action);
```
alllex: 🤔 The method name and javadoc make it unclear what happens when the result is already there

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/history/impl/DefaultImmutableWorkspaceMetadataStore.java
```diff
@@ -84,6 +84,8 @@ public void storeWorkspaceMetadata(File workspace, ImmutableWorkspaceMetadata me
                     hashCodeSerializer.write(encoder, hash);
                 }
             }
+            // Write controlling "is complete" byte
+            encoder.writeBoolean(true);
```
alllex, line 88: ❔ How is an extra byte helping here relative to successfully reading the known number of bytes that represents the actual state? For which scenarios are we adding the additional protection?

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
```diff
@@ -66,197 +60,167 @@
...
+                // We need to invalidate snapshots in case another process populated the workspace just before us
+                Runnable invalidateSnapshots = () -> fileSystemAccess.invalidate(ImmutableList.of(workspace.getImmutableLocation().getAbsolutePath()));
+                // First, try to load it again if another process populated it just before us (double-checked locking)
+                WorkspaceLoad load = loadImmutableWorkspaceIfComplete(work, workspace, invalidateSnapshots);
+
+                WorkspaceResult result = load.isSuccess()
+                    ? load.getResult()
+                    : executeInWorkspace(work, context, workspace);
+
+                if (initialLoad.isSoftDeleted()) {
```
alllex: 💅 It would be cleaner to ask this from the second `load` we do from under the lock

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
alllex: 🤔 This API is a sort of Continuation Passing Style code, and feels more complicated here than it needs to be. Also makes the contract of the `getIfRunningOrCompute` method complicated.

I think it would be simpler if the method returned data (as opposed to using control flow) to indicate whether the result was obtained from a competing thread. This also opens more avenues for extension, such as including the time this thread had to wait (and block other work) on the result being produced by another thread. Right now, we are marking the duration as zero, which is not entirely true.

I propose the method to return a wrapper result that indicates what I described above. Then the method can be named simpler, like `getOrCompute()` and only take one argument.

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
```diff
@@ -66,197 +60,167 @@
...
-            AtomicMoveImmutableWorkspace workspace = workspaceProvider.getAtomicMoveWorkspace(uniqueId);
-            return loadImmutableWorkspaceIfExists(work, workspace)
-                .orElseGet(() -> executeInTemporaryWorkspace(work, context, workspace));
-        }
+        ImmutableWorkspace workspace = workspaceProvider.getWorkspace(uniqueId);
+        // Gets a workspace result or compute it,
+        // but only computes it once across all threads that try to compute it at the same time
+        return workspace.getIfRunningOrCompute(concurrentResultMapper(work, workspace), () -> {
+            // If a workspace already exists, just return the result
+            WorkspaceLoad initialLoad = loadImmutableWorkspaceIfExists(work, workspace);
```
alllex: ❔We short circuit here. Can you remind me how do we update the state on disk such that the cleanup strategy would know that the workspace doesn't need to be soft deleted anymore, since it was read/used (without any locks)

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
```diff
@@ -66,197 +60,167 @@
...
+ * <li>We then check if the workspace is soft-deleted</li>
+ * <li>If an immutable workspace already exists and it's not soft-deleted, it is checked for consistency, and is returned
  * if found correct.</li>
- * <li>If the workspace is inconsistent (the output hashes stored in {code metadata.bin} do not match
- * the hashes taken by snapshotting the current outputs), the workspace is moved to a temporary
- * location and we fall back to re-executing the work.</li>
- * <li>If we end up executing the work (either because there is no existing immutable workspace, or it is
- * inconsistent), then a unique temporary workspace directory is provided to the work to create its outputs
- * in, and the work is executed.</li>
+ * <li>If the workspace doesn't exist or is soft-deleted or or {@code metadata.bin} doesn't exist or is inconsistent (the output hashes stored in {@code metadata.bin} do not match
```
alllex:
```suggestion
 * <li>If the workspace doesn't exist or is soft-deleted or {@code metadata.bin} doesn't exist or is inconsistent (the output hashes stored in {@code metadata.bin} do not match
```

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/impl/CacheBasedImmutableWorkspaceProvider.java
```diff
@@ -17,173 +17,139 @@
...
+            return workspace;
+        }
 
-                return cacheContainer.withFileLock(supplier);
-            }
-        };
-    }
+        @Override
+        public <T> T withFileLock(Supplier<T> action) {
+            return cache.withFileLock(entryKey, action);
```
alllex: 💭 Since we are not using the thread-guarding feature of the fine-grained cache for the actual workspace handling, would it be better (as a potential follow-up) to simply remove it from the implementation? `withFileLock => useCache`

We do use `useCache` for clean-up, but do we need to have the thread-guarding there? We know it's at most one cleaner per process anyway.

This would simplify the cache implementation (albeit not much), but will make the contract clear -- callers must manage threads by themselves, if they want to guard against competing threads.

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/impl/CacheBasedImmutableWorkspaceProvider.java
alllex: Why is it beneficial to have it?

It makes the implementation more complicated. However, if the same work is requested multiple times in the build (not concurrently), then the repeated attempts go through the file system anyway. So why not simplify?

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
asodja: When we get ImmutableWorkspace, we mark it as accessed to fileAccessTracker:
https://github.com/gradle/gradle/blob/37d56ea06e97c025789eb285a50b4c960d5191d2/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/impl/CacheBasedImmutableWorkspaceProvider.java#L80-L84

That way cleanup will know that an entry was accessed and won't try to soft delete it

subprojects/core/src/integTest/groovy/org/gradle/api/CrossBuildScriptCachingIntegrationSpec.groovy
alllex: Wouldn't these tests also pass with IP if we removed the protection for competing threads?

I presume, previously we might have kicked off two full compilations, because they would be populating independent one-off workspaces to-be-moved-later-atomically. But now, even if two threads are competing, one would inevitably wait of the file lock, and will not rerun the compilation.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformCachingIntegrationTest.groovy
```diff
@@ -1362,8 +1316,50 @@ class ArtifactTransformCachingIntegrationTest extends AbstractHttpDependencyReso
         succeeds ":util:resolve"
     }
 
+    @Issue("https://github.com/gradle/gradle/issues/28974")
+    def "non-incremental transform can recover if metadata is #description"() {
+        given:
+        buildFile << declareAttributes() << multiProjectWithJarSizeTransform() << withJarTasks() << withFileLibDependency("lib3.jar") << withExternalLibDependency("lib4")
```
alllex, line 1322: 😄I haven't seen leftShift chaining before. I hope it is left-heavy and the order of appending is left-to-right!

subprojects/core/src/integTest/groovy/org/gradle/api/CrossBuildScriptCachingIntegrationSpec.groovy
asodja: Unfortunately not, since file lock is per process, not per thread. File lock protects only against process concurrency, but if you want to take care of thread concurrency you have to handle it inside your process

alllex (approved):
LGTM!

I left a number of questions and comments. Those are mostly discussion, and the implementation looks sound!

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/impl/CacheBasedImmutableWorkspaceProvider.java
asodja: So we definitely need to do the "write operation" in `withFileLock` logic under the thread lock, since file locks protect only against process concurrency, see https://github.com/gradle/gradle/pull/36227#discussion_r2707902998.

The "read operation" thread locking is not that obvious, but it's used because we read snapshots here: https://github.com/gradle/gradle/blob/37d56ea06e97c025789eb285a50b4c960d5191d2/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java#L149-L151, and since snapshots are also cached it can happen that a thread snapshots a workspace while the other thread is writing to a workspace, and then you get incosistency and failures. 

I thought I could fix this by calling `fileSystemAccess.invalidate` just after write to workspace is complete, somewhere here https://github.com/gradle/gradle/blob/97317526c9b4047f23b438d185c160cef84aad7c/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/BroadcastChangingOutputsStep.java#L60-L65
That would remove the thread lock for the "read only case" and would probably save few milliseconds (although not that much from what I tested). But I still got a build failure in one of AGP tests. So for now we definitely need thread lock also for "read operation case".

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/impl/CacheBasedImmutableWorkspaceProvider.java
```diff
@@ -17,173 +17,139 @@
...
-            @Override
-            public File getImmutableLocation() {
-                return workspace;
-            }
+    public static class CachedBasedImmutableWorkspace implements ImmutableWorkspace {
+        private final String entryKey;
+        private final File workspace;
+        private final FineGrainedPersistentCache cache;
+        private final FineGrainedCacheEntrySoftDeleter softDeleter;
+        private final Map<String, CompletableFuture<?>> workspaceResults;
```
alllex: 💭 It feels strange that this static class carries a concurrently mutable container, whose lifetime is larger than that of this class. We also lose the `ConcurrentMap` aspect in the field type.

It's more of a style concern, but I think it would be cleaner if either this class was non-static and the completable future handling logic was in the enclosing class, or if this was modeled more explicitly with a dedicated interface.

platforms/core-configuration/kotlin-dsl/src/integTest/groovy/org/gradle/kotlin/dsl/KotlinScriptCacheCleanupIntegrationTest.groovy
asodja: Fixed

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/ImmutableWorkspaceProvider.java
asodja: Good catch, fixed

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformCachingIntegrationTest.groovy
asodja: I just used the same setup as one of previous test. I agree it's kinda weird, but since it's used all over this test I think there is not a lot of value to change it everywhere

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/impl/CacheBasedImmutableWorkspaceProvider.java
asodja: I agree it's not ideal. This CachedBasedImmutableWorkspace is actually visible just for testing, it was annoymous class before but I's a lot of work to mock CacheBasedImmutableWorkspaceProvider

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
asodja: Ok, I simplified this and introduced ConcurrentResult

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/impl/CacheBasedImmutableWorkspaceProvider.java
asodja: I will just leave, it doesn't seem too complicated. Also it matches `PersistentCache (or ExclusiveCacheAccessCoordinator)` that has `useCache` and `withFileLock` methods

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/impl/CacheBasedImmutableWorkspaceProvider.java
asodja: I changed that and now we reuse `CacheBasedImmutableWorkspaceProvider` for tests. Implementation of `ImmutableWorkspace` is hidden

asodja:
@bot-gradle test this APT

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/ImmutableWorkspaceProvider.java
asodja: Renamed the method and cleanup the JavaDoc

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
asodja: Added a comment to make it obvious that only initial load checks for soft deletion. We don't check again under lock, we leave for the next read to handle it

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/ImmutableWorkspaceProvider.java
```diff
@@ -20,45 +20,62 @@
...
+        private final boolean isProducedByCurrentThread;
+
+        private ConcurrentResult(T value, boolean isProducedByCurrentThread) {
+            this.value = value;
+            this.isProducedByCurrentThread = isProducedByCurrentThread;
+        }
+
+        public T get() {
+            return value;
+        }
```
alllex, line 68: 💅 missing empty line

subprojects/core/src/integTest/groovy/org/gradle/api/CrossBuildScriptCachingIntegrationSpec.groovy
alllex: I see. I was missing that piece of the context. Now the implementation choices make more sense 👍

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
alllex: Thanks! I think the code is more self-documenting now 👍

PR #36230, opened 2026-01-09
Allow multiple project features to use the same name with different targets

ghale:
Previously, feature names were mutually exclusive and two features could not be registered with the same name even if they bound to different target receivers.  

This PR allows multiple features to bind to the same name as long as they have unambiguously different target receivers.  If the targets are ambiguous in any way (e.g. they have the same target or one binds to the super class of the other or one binds to the build model associated with the definition of the other, etc) then we throw an error.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/declarative-dsl-tooling-builders/src/crossVersionTest/groovy/org/gradle/declarative/dsl/tooling/builders/r814/AndroidEcosystemPrototypeCrossVersionTest.groovy
+++ b/platforms/core-configuration/declarative-dsl-tooling-builders/src/crossVersionTest/groovy/org/gradle/declarative/dsl/tooling/builders/r814/AndroidEcosystemPrototypeCrossVersionTest.groovy
@@ -24,7 +24,9 @@ import org.gradle.internal.declarativedsl.evaluator.main.SimpleAnalysisEvaluator
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.UnitTestPreconditions
 import org.gradle.tooling.events.ProgressListener
+import spock.lang.Ignore
 
+@Ignore("Until a version of the Android Ecosystem plugin that is compatible with Gradle 9.3+ is released.")
 @TargetGradleVersion(">=8.14 <9.3")
 @ToolingApiVersion('>=8.14')
 class AndroidEcosystemPrototypeCrossVersionTest extends AbstractDeclarativeDslToolingModelsCrossVersionTest {


ghale:
@bot-gradle test this

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/software/internal/DefaultProjectFeatureImplementation.java
```diff
@@ -63,6 +64,7 @@ public DefaultProjectFeatureImplementation(
         ProjectFeatureApplyAction<T, V, ?> bindingTransform
     ) {
         this.featureName = featureName;
+        this.uniqueId = featureName + targetDefinitionType.getId();
```
eskatos: ❔ Shouldn't we use a separator between the feature name and the type id to prevent collisions, or downright a hasher?

platforms/core-configuration/project-features-api/src/main/java/org/gradle/api/internal/plugins/TargetTypeInformation.java
```diff
@@ -16,14 +16,24 @@
...
     class DefinitionTargetTypeInformation<T> implements TargetTypeInformation<T> {
         public final Class<T> definitionType;
 
         public DefinitionTargetTypeInformation(Class<T> definitionType) {
             this.definitionType = definitionType;
         }
 
+        @Override
+        public String getId() {
+            return Hashing.hashString(definitionType.getName()).toCompactString();
```
eskatos: ❔ Shouldn't we cache the computed hash?

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorFragments.kt
```diff
@@ -79,7 +79,7 @@ fun fragmentsFor(accessor: Accessor): Fragments = when (accessor) {
 }
 
 private fun fragmentsForProjectType(accessor: Accessor.ForProjectType): Fragments = accessor.run {
-    val className = "${accessor.spec.projectFeatureName.original.uppercaseFirstChar()}ContainerElementFactoriesKt"
+    val className = internalNameForAccessorClassOf(accessor.spec)
```
eskatos, line 82: 👍

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/interpreter/defaults/DeclarativeModelDefaultsHandler.kt
```diff
@@ -66,10 +66,14 @@ abstract class DeclarativeModelDefaultsHandler @Inject constructor(
     val modelDefaultsRepository = projectFeatureRegistryBasedModelDefaultsRepository(projectFeatureDeclarations)
 
     override fun apply(target: Any, definition: Any, classLoaderContext: ClassLoaderContext, projectFeatureName: String, plugin: Plugin<*>) {
+        // TODO - this works because there should only be one implementation for each project type.  We'll need to revisit this
+        // when we support defaults for project features which can have multiple implementations.
+        val projectFeature = projectFeatureDeclarations.projectFeatureImplementations[projectFeatureName]?.first()
```
eskatos: 💅 Refine nullability handling
```suggestion
        val projectFeature = projectFeatureDeclarations.projectFeatureImplementations.getValue(projectFeatureName).first()
```

ghale:
@bot-gradle test this

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/interpreter/defaults/DeclarativeModelDefaultsHandler.kt
eskatos: Note that with this suggestion applied, the `!!` below is now unnecessary

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/software/internal/DefaultProjectFeatureImplementation.java
ghale: That's probably worth doing, although the probability of collision here is pretty small (but not zero).

Done.

platforms/core-configuration/project-features-api/src/main/java/org/gradle/api/internal/plugins/TargetTypeInformation.java
ghale: I reckon.  Done.

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/interpreter/defaults/DeclarativeModelDefaultsHandler.kt
ghale: Ah, I should have looked at this closer.  I just accepted your suggestion without thinking too much about why you were suggesting this.  Fixed.

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/software/schemaFromProjectFeatures.kt
```diff
@@ -190,7 +190,7 @@ data class ProjectFeatureInfo<T : Definition<V>, V : BuildModel>(
             emptyList(),
             isDirectAccessOnly = true,
             semantics = FunctionSemanticsInternal.DefaultAccessAndConfigure(
-                accessor = ConfigureAccessorInternal.DefaultCustom(definitionType, customAccessorId),
+                accessor = ConfigureAccessorInternal.DefaultProjectFeature(definitionType, customAccessorId, delegate.featureName),
```
h0tk3y: For some more strictness, I suggest making `RuntimeModelTypeAccessors.getObjectFromCustomAccessor` also check for the accessor being `ProjectFeature` and not just any `Custom` accessor.

subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultSharedModelDefaults.java
```diff
@@ -65,14 +66,20 @@ public ProjectLayout getLayout() {
     @Override
     public <T> void add(String name, Class<T> publicType, Action<? super T> configureAction) {
         if (projectFeatureDeclarations.getProjectFeatureImplementations().containsKey(name)) {
-            ProjectFeatureImplementation<?, ?> projectFeature = projectFeatureDeclarations.getProjectFeatureImplementations().get(name);
+            Set<ProjectFeatureImplementation<?, ?>> implementations = projectFeatureDeclarations.getProjectFeatureImplementations().get(name);
+            if (implementations.isEmpty()) {
+                throw new IllegalArgumentException(String.format("Cannot add default for project type '%s' because it has no implementations.", name));
+            }
+            // TODO - this works for now because we only have one implementation per project type, but we need to revisit this when we support defaults
+            // for features where we could have multiple implementations binding to different target types
```
h0tk3y, line 74: Would it make sense to throw an exception here when there is more than one implementation? That would make it easier to change this later (e.g. implement defaults for project features).

subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultSharedModelDefaults.java
```diff
@@ -92,12 +99,17 @@ public boolean hasMethod(String name, Object... arguments) {
...
         public DynamicInvokeResult tryInvokeMethod(String name, Object... arguments) {
             if (hasMethod(name, arguments)) {
-                ProjectFeatureImplementation<?, ?> projectFeature = projectFeatureDeclarations.getProjectFeatureImplementations().get(name);
-                add(name, projectFeature.getDefinitionPublicType(), Cast.uncheckedNonnullCast(toAction(arguments[0])));
+                Set<ProjectFeatureImplementation<?, ?>> implementations = projectFeatureDeclarations.getProjectFeatureImplementations().get(name);
+                if (implementations.isEmpty()) {
+                    throw new IllegalArgumentException(String.format("Cannot resolve default for project type '%s' because it has no implementations.", name));
+                }
+
+                ProjectFeatureImplementation<?, ?> implementation = implementations.iterator().next();
```
h0tk3y, line 113: It might make sense to throw an exception when there's more than one here as well.

platforms/core-configuration/declarative-dsl-evaluator/src/main/kotlin/org/gradle/internal/declarativedsl/evaluator/main/AnalysisDocumentUtils.kt
```diff
@@ -32,6 +32,8 @@ import org.gradle.internal.declarativedsl.evaluator.runner.stepResultOrPartialRe
 
 object AnalysisDocumentUtils {
     fun documentWithModelDefaults(modelDefaultsSequenceResult: AnalysisSequenceResult, mainSequenceResult: AnalysisSequenceResult): DocumentOverlayResult? {
+        // TODO - This works because there can't be collisions on project types.  This will have to be more specific when we
+        // support model defaults for other feature types.
```
h0tk3y, line 39: By the way, this will likely give an incorrect result in the following case:
```kotlin
defaults {
    foo {
        foo = "foo"
    }
    bar {
        bar = "bar"
    }
}
```
Here we have defaults for two project types, `foo` and `bar`.
Now if we also have a project feature for some nested model named `bar`, we can erroneusly include the defaults for `bar` when in fact the project type is `foo`, for example:

```kotlin
foo {
    someNestedModel {
        bar { } // uses an unrelated feature named 'bar'
    }
}
```

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/software/internal/DefaultProjectFeatureDeclarations.java
```diff
@@ -91,39 +91,27 @@ public void addDeclaration(@Nullable String pluginId, Class<? extends Plugin<Pro
         pluginClasses.computeIfAbsent(pluginKey, k -> new LinkedHashSet<>()).add(pluginClass);
     }
 
-    private Map<String, ProjectFeatureImplementation<?, ?>> discoverProjectFeatureImplementations() {
-        final ImmutableMap.Builder<String, ProjectFeatureImplementation<?, ?>> projectFeatureImplementationsBuilder = ImmutableMap.builder();
+    private Map<String, Set<ProjectFeatureImplementation<?, ?>>> discoverProjectFeatureImplementations() {
+        Map<String, Set<ProjectFeatureImplementation<?, ?>>> projectFeatureDeclarations = new HashMap<>();
```
h0tk3y: I'd suggest using a `Map` implementation that keeps the key insertion order in iteration.

platforms/core-configuration/project-features-api/src/main/java/org/gradle/api/internal/plugins/TargetTypeInformation.java
```diff
@@ -16,12 +16,24 @@
...
+
+    String getId();
+
     class DefinitionTargetTypeInformation<T> implements TargetTypeInformation<T> {
         public final Class<T> definitionType;
+        private final String id;
 
         public DefinitionTargetTypeInformation(Class<T> definitionType) {
             this.definitionType = definitionType;
+            this.id = Hashing.hashString(definitionType.getName()).toCompactString();
```
h0tk3y: I'd prefer a more human-readable ID for the purpose of debugging or troubleshooting of the resulting schemas. Although the original definition target type information is kept in some form as well, having it also in the ID is a bit more straightforward.

ghale: I think I was imagining growing this hash in the future to include additional information specific to the target.  For instance, just hashing the classname, you could have two different classes in two different classloaders that have the same fully qualified name that hash to the same value.  We only have the single classloader now, so this isn't an issue, but we've talked about potentially having a more nuanced classloader structure.  So I could see us eventually wanting to include the actual loaded class into the hash in some way.  That's why I made this some sort of hash.

Having said that, it's probably not necessary now, and relatively easy to change in the future, so using something more readable makes sense if it eases troubleshooting.  How would you suggest representing this - as just the unhashed fully qualified class name?

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/software/schemaFromProjectFeatures.kt
ghale: Makes sense.  Done.

subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultSharedModelDefaults.java
ghale: Sounds good.  Done.

subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultSharedModelDefaults.java
ghale: Done.

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/software/internal/DefaultProjectFeatureDeclarations.java
ghale: Done.

platforms/core-configuration/declarative-dsl-evaluator/src/main/kotlin/org/gradle/internal/declarativedsl/evaluator/main/AnalysisDocumentUtils.kt
ghale: Good catch.  For now, I've changed this so that we only extract the names of project types when determining which defaults need to be overlayed.  

We'll need to address this, though, when we support defaults for other features.  The main problem, I think, is that the document nodes don't contain enough information to disambiguate features with the same name.

platforms/core-configuration/project-features-api/src/main/java/org/gradle/api/internal/plugins/TargetTypeInformation.java
ghale: I've just made this the fully-qualified class name for now.  We can change it if we find something better.

ghale: In other words, a feature named "foo" that binds to "org.gradle.test.Bar" will have an id of "foo-org.gradle.test.Bar".

platforms/core-configuration/declarative-dsl-tooling-models/src/main/kotlin/org/gradle/declarative/dsl/schema/CustomAccessorIdentifier.kt
```diff
@@ -0,0 +1,51 @@
...
+    val name: String
+
+    interface ExtensionAccessorIdentifier : CustomAccessorIdentifier, Serializable
+
+    interface ContainerAccessorIdentifier : CustomAccessorIdentifier, Serializable {
+        val elementTypeClassName: String
+    }
+
+    interface ProjectFeatureIdentifier : CustomAccessorIdentifier, Serializable {
+        val targetTypeClassName: String
```
h0tk3y, line 38: Given that some project features target definition types and some target any definition with a matching build model, should we include that distinction here?

I'm not sure that the corner case is realistic:
```kotlin
class BothDefinitionAndBuildModel : Definition<SomeClass>, BuildModel
```
, with two features, one targeting a definition of `BothDefinitionAndBuildModel`, the other a build model type `BothDefinitionAndBuildModel`.

However, we might still want to track this information, since the tooling might be interested in it.

ghale: I've added some information to the schema to track this.

ghale:
@bot-gradle test this

blindpirate:
This is failing: https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_QuickFeedbackCrossVersion_5_bucket8/107811333

PR #36231, opened 2026-01-09
Remove binary resolving logic from OperatingSystem

asodja:
This moves binary resolving logic out of OperatingSystem in a OperatingSystemFileResolver.

I did that since I moved OperatingSystem out of base-services, but due to that I had to duplicate [FileUtils.withExtension](https://github.com/gradle/gradle/blob/da50de4d9ea1e11dacd782186ce2d93a1515773f/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/FileUtils.java#L163) method that is used in binary resolving logic methods. It seems this logic doesn't belong to OperatingSystem anyway, so I decided to move it to a separate class.


asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/platform/PlatformBinaryResolver.java
```diff
@@ -0,0 +1,146 @@
...
+        } else {
+            return "lib" + libraryName + suffix;
+        }
+    }
+
+    public static PlatformBinaryResolver forOs(OperatingSystem os) {
+        if (os == OperatingSystem.current()) {
+            return forCurrentOs();
+        }
+        return new PlatformBinaryResolver(os);
```
jbartok: It feels inconsistent to have a static/singleton instance for the current OS but not for other ones. Wouldn't it make more sense to have a static `Map<OperatingSystem, PlatformBinaryResolver>` in which to lazily store all the instances that have been actually requested?

I understand that (thread)safe publication is an issue... Make it `ThreadLocal`?

Just wondering...

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/platform/PlatformBinaryResolver.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        int pos = libraryName.lastIndexOf('/');
+        if (pos >= 0) {
+            return libraryName.substring(0, pos + 1) + "lib" + libraryName.substring(pos + 1) + suffix;
+        } else {
+            return "lib" + libraryName + suffix;
+        }
+    }
+
+    public static PlatformBinaryResolver forOs(OperatingSystem os) {
```
jbartok: Don't like these names, `forOs(...)` and `forCurrentOs()`... Wouldn't `instance(...)` and `instance()` do the job instead? And be more familiar with the API users?

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/platform/PlatformBinaryResolver.java
```diff
@@ -0,0 +1,146 @@
...
+package org.gradle.internal.platform;
+
+import org.gradle.internal.os.OperatingSystem;
+import org.jspecify.annotations.Nullable;
+
+import java.io.File;
+
+import static org.gradle.internal.FileUtils.withExtension;
+
+public class PlatformBinaryResolver {
```
jbartok: The name `PlatformBinaryResolver` is not the best... If you see it next to `OperatingSystem`, then yes, the meaning is mostly clear, but if you see it in isolation, it's much harder to figure out what it wants to be... 

Now do I have a better idea.... Maybe `OSSpecificBinaryResolver`? Or `OSSpecificFileTypeResolver`? These aren't great either...

platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/jvm/JvmTest.groovy
```diff
@@ -31,15 +32,16 @@ class JvmTest extends Specification {
...
     SetSystemProperties sysProp = new SetSystemProperties()
-    OperatingSystem os = Mock() {
+    OperatingSystem os = Mock()
+    PlatformBinaryResolver binaryResolver = Mock() {
         getExecutableName(_) >> { String name ->
             return "${name}.exe"
         }
     }
-    OperatingSystem theOs = OperatingSystem.current()
+    PlatformBinaryResolver currentOsBinaryResolver = PlatformBinaryResolver.forCurrentOs()
```
jbartok: Is this correct? Previously it was tied to a mock OS, now it's tied to current os.

platforms/native/platform-native/src/main/java/org/gradle/nativeplatform/platform/internal/OperatingSystemInternal.java
```diff
@@ -17,11 +17,15 @@
...
 import org.gradle.api.tasks.Internal;
+import org.gradle.internal.platform.PlatformBinaryResolver;
 import org.gradle.nativeplatform.platform.OperatingSystem;
 
 public interface OperatingSystemInternal extends OperatingSystem {
     @Internal
     org.gradle.internal.os.OperatingSystem getInternalOs();
 
+    @Internal
+    PlatformBinaryResolver getBinaryResolver();
```
jbartok: This feels wrong even here... It's the opposite of what we are aiming for, which is to make OSs not know about the binary resolving functionality...

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/platform/PlatformBinaryResolver.java
asodja: I didn't want to create a map, since we can instantiate new OperatingSystems in tests. But I see that for most production code we use current Os resolver, and that is why I think it makes sense to reuse the instance

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/platform/PlatformBinaryResolver.java
asodja: I don't like name either, I had several iterations of it. What do you think about OperatingSystemBinaryResolver or OperatingSystemFileResolver?

platforms/native/platform-native/src/main/java/org/gradle/nativeplatform/platform/internal/OperatingSystemInternal.java
asodja: I think we could do this a bit differently, I will take a look

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/platform/PlatformBinaryResolver.java
alllex: I'd also prefer to rename this to something else. OperatingSystemFileResolver is my favorite from previously suggested.

However, I think it would be nice to choose something that works harmoniously with the static methods.

Suggestion: `OperatingSystemFileConventions.current()` and `.of()`. That would be my personal preference. A bit of a mouthful, but muscle memory friendly for auto-complete and unambiguous.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/platform/PlatformBinaryResolver.java
alllex: I'd prefer `current()` and `of()`

platforms/native/platform-native/src/main/java/org/gradle/nativeplatform/platform/internal/OperatingSystemInternal.java
alllex: I think, on the contrary, it's pretty natural to have the getter here. Ideally, we'd have it on the type that is returned by the `getInternalOs()`. But it's tricky, due to the compile dependency on apache-commons

platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/platform/PlatformBinaryResolverTest.groovy
```diff
@@ -0,0 +1,169 @@
...
+ */
+
+package org.gradle.internal.platform;
+
+import org.gradle.internal.os.OperatingSystem
+import org.gradle.test.fixtures.file.TestNameTestDirectoryProvider
+import org.junit.Rule;
+import spock.lang.Specification;
+
+class PlatformBinaryResolverTest extends Specification {
```
alllex: It's good to have this as a separate test.

Ideally, we should run these tests against all platforms we support. Currently, it's implied that all the other platforms not tested here are covered by UNIX. Though, this might change in the future.

It would be cool to have something like `OperatingSystem.getAll()`, and then have a test here that validates the list does not change, and has a comment that if it changes, then these tests need to be adjusted.

In terms of coverage, I think it'd be better if we convert the UNIX tests to `where:` tests, which run against all platforms and capture the expected behavior on each of them.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/platform/PlatformBinaryResolver.java
asodja: OperatingSystemFileConventions would work well with static methods, but here we have an instance.
So maybe something without ending `s` would work better 🤔

platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/platform/PlatformBinaryResolverTest.groovy
asodja: This test is just copied from OperatingSystem, I didn't actually change it. But sure I can add more coverage

platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/jvm/JvmTest.groovy
asodja: I think it's ok, tests pass. I believe before it was  
```
OperatingSystem theOs = OperatingSystem.current()
```
so also not a mock

platforms/native/platform-native/src/main/java/org/gradle/nativeplatform/platform/internal/OperatingSystemInternal.java
asodja: I removed this getter. I now institate OperatingSystemFileResolver in the implementation

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/platform/PlatformBinaryResolver.java
asodja: I went with `current()` and `of()` to mimic `OperatingSystem`

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/platform/PlatformBinaryResolver.java
asodja: I went with `OperatingSystemFileResolver`, it's seems close enough to `OperatingSystem`

asodja:
> I'm out of the loop... what's the goal of moving things out of OperatingSystem? does it let us detangle some dependencies?

So I had to move OperatingSystem out of base-services, since I wanted to use it as a dependency of build cache library.
But since OperatingSystem had a dependency on FileUtils.withExtension that is also in base-services I had to duplicate some logic to remove the dependency.

This duplicated logic was used for methods that resolve `dll/exe/sh` and similar files for OperatingSystem. After a short discussion inside the cortex team we figured that these could be moved out, since it feels they are mostly used just for native and not in general.

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/platform/PlatformBinaryResolverTest.groovy
asodja: I added more coverage, but I didn't touch logic in OperatingSystem, or have a list for it, since I don't want touch other logic in the OperatingSystem class. I will leave that for future.

asodja:
@bot-gradle test this

PR #36241, opened 2026-01-12
Fix indicating task-running builds for build scans

alllex:
Fixes a small bug introduced in
- https://github.com/gradle/gradle/pull/35902

where the computation of a `BuildType` of an invocation was unintentionally changed such that it does not convey tasks being executed for TAPI requests that both request models _and_ run tasks.

Historically, Build Scans publishing has depended on the build indicating that it runs tasks, which was affected by the changes noted above.

---

This PR drops the internal `BuildType` in favor of getting the same information directly from an existing `BuildActionModelRequirements` service.

### DIFF
--- a/build-logic/profiling/src/main/kotlin/gradlebuild.buildscan.gradle.kts
+++ b/build-logic/profiling/src/main/kotlin/gradlebuild.buildscan.gradle.kts
@@ -27,8 +27,6 @@ import gradlebuild.basics.kotlindsl.execAndGetStdoutIgnoringError
 import gradlebuild.basics.logicalBranch
 import gradlebuild.basics.predictiveTestSelectionEnabled
 import gradlebuild.basics.testDistributionEnabled
-import org.gradle.api.internal.BuildType
-import org.gradle.api.internal.GradleInternal
 import org.gradle.internal.operations.BuildOperationDescriptor
 import org.gradle.internal.operations.BuildOperationListener
 import org.gradle.internal.operations.BuildOperationListenerManager
@@ -78,10 +76,6 @@ buildScan {
             tag("PRE_TESTED_COMMIT")
         }
     }
-}
-
-if ((project.gradle as GradleInternal).services.get(BuildType::class.java) != BuildType.TASKS) {
-    buildScan?.tag("SYNC")
 }
 
 fun DevelocityConfiguration.extractCiData() {


alllex:
@bot-gradle test this

asodja (approved):
LGTM overall

Should we also cover this with a test, so we don't break it in the future?

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/legacy/LegacyGradleEnterprisePluginCheckInService.java
```diff
@@ -164,7 +164,7 @@ public boolean isRootProjectHasVcsMappings() {
         @Override
         public boolean isTaskExecutingBuild() {
             boolean forceTaskExecutingBuild = System.getProperty("org.gradle.internal.ide.scan") != null;
-            return forceTaskExecutingBuild || buildType == BuildType.TASKS;
+            return forceTaskExecutingBuild || taskExecutingBuild;
```
mlopatkin: 💅 Can we add a comment describing why `org.gradle.internal.ide.scan` makes every build "task-executing build", while we're here? It really isn't obvious for the reader - now I got it from the PR description, but I'll forget it tomorrow.

alllex: This class is a sort of stub that is only used for legacy _unsupported_ versions of the build scan plugin to allow for a graceful unsupported reason reporting. This is already described in the class javadoc. I only touched it because I was removing the `BuildType`.

If you feel strongly about it, I'd rather remove the property reading logic entirely. I strongly suspect it's not used anywhere anymore. The DV repo does not have occurrences of this property in the codebase.

mlopatkin: > If you feel strongly about it, I'd rather remove the property reading logic entirely. I strongly suspect it's not used anywhere anymore. The DV repo does not have occurrences of this property in the codebase.

Given that this is a legacy class on its way out, I don't insist too strongly.

alllex: I've confirmed with the Build Scans team that the property is not needed in the legacy check-in service, so I removed it entirely to avoid potential confusion in the future

alllex:
@bot-gradle  test this

alllex:
@mlopatkin, @asodja, I've addressed your feedback, PTAL!

alllex:
@bot-gradle test this

build-logic/profiling/src/main/kotlin/gradlebuild.buildscan.gradle.kts
```diff
@@ -80,10 +78,6 @@ buildScan {
     }
 }
 
-if ((project.gradle as GradleInternal).services.get(BuildType::class.java) != BuildType.TASKS) {
-    buildScan?.tag("SYNC")
-}
```
wolfs, line 85: 🤔 Are you tagging sync builds differently now? I think we should still tag them.

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DefaultGradleEnterprisePluginConfig.java
```diff
@@ -26,9 +26,13 @@ public class DefaultGradleEnterprisePluginConfig implements GradleEnterprisePlug
...
 
-    public DefaultGradleEnterprisePluginConfig(StartParameter startParameter, BuildType buildType, GradleEnterprisePluginAutoAppliedStatus autoAppliedStatus) {
+    public DefaultGradleEnterprisePluginConfig(
+        BuildActionModelRequirements requirements,
+        StartParameter startParameter,
+        GradleEnterprisePluginAutoAppliedStatus autoAppliedStatus
+    ) {
         this.buildScanRequest = buildScanRequest(startParameter);
-        this.taskExecutingBuild = buildType == BuildType.TASKS;
+        this.taskExecutingBuild = requirements.isRunsTasks();
```
wolfs, line 35: 🤔 Without diving deeper into the code, is the `isRunsTasks()` true exactly when the buildType was `TASKS`? Or is there a slight change?

build-logic/profiling/src/main/kotlin/gradlebuild.buildscan.gradle.kts
alllex: We have been tagging them differently for a while now with `IDE Sync`
- https://ge.gradle.org/scans/trends?search.query=tag:%22IDE%20Sync%22&search.relativeStartTime=P90D&search.timeZoneId=Europe%2FBerlin

I explained it in the commit message:
> Remove unused SYNC tag
This tag would not be published for a while, since Gradle sync build always runs a Kotlin DSL-related lifecycle task.

> Instead, the build scan conventions plugin publishes the `IDE Sync` tag:
https://github.com/gradle/gradle-org-conventions-plugin/blob/0f2bda42986dfa9e6235746aa7080d338ffb6a38/src/main/java/io/github/gradle/conventions/customvalueprovider/CIBuildCustomValueProvider.java#L69

---

platforms/enterprise/enterprise/src/main/java/org/gradle/internal/enterprise/impl/DefaultGradleEnterprisePluginConfig.java
alllex: It was exactly that:

https://github.com/gradle/gradle/pull/35902/files#diff-3464df057cb240f5a949564d56748b794f61a0972490c39bfb04692fed6fd8b7L109

wolfs (approved):

Issue #30990, opened 2024-10-22
Eliminate potential deadlocks when writing CC fingerprint


mlopatkin:
Writing some CC fingerprint data may trigger another fingerprint event. This includes:
- processing a ValueSource (we store input parameters that can be providers)
- fingerprinting a FileCollection (may contain other file collections or providers), as part of task inputs or when observed by the user code

ValueSources are mostly mitigated now after we've made the fingerprint listener parallel. FileCollection deadlocks are hard to reproduce, but remain theoretically possible and may become an issue when we fingerprint more of them.

mlopatkin:
It is still possible to have a deadlock because of the fingerprint writer lock:
```
Thread A: ValueSource obtained -> acquires writer lock -> starts writing -> triggers env var read -> tries to acquire broadcast lock
Thread B: reads env var -> acquires broadcast lock -> tries to acquire the writer lock
```

bamboo:
> It is still possible to have a deadlock because of the fingerprint writer lock:
Can't we get rid of the writer lock completely via a queue?

mlopatkin:
>Can't we get rid of the writer lock completely via a queue?

We may have to run user code when serializing (providers and serializables), so it isn't easy to offload (I suppose you're suggesting using a writer thread to handle the queue).

abstratt:
I tried to imagine a test case with carefully placed sync points but my imagination is failing me. I was assuming project-specific serialization build operations (with parallel CC), with some sort of mutual dependency, but I guess that is not it. 

@mlopatkin  What are the threads mentioned here? https://github.com/gradle/gradle/issues/30990#issuecomment-2492159107

  ```
  Thread A: ValueSource obtained -> acquires writer lock -> starts writing -> triggers env var read -> tries to acquire broadcast lock
  Thread B: reads env var -> acquires broadcast lock -> tries to acquire the writer lock
  ```

EDIT: Ah, it is the fingerprinter listener that is parallel, and this may have nothing to do with parallel CC:

> ValueSources are mostly mitigated now after we've made the fingerprint listener parallel. FileCollection deadlocks are hard to reproduce, but remain theoretically possible and may become an issue when we fingerprint more of them.

mlopatkin:
> I tried to imagine a test case with carefully placed sync points

This is annoying to test, as there is no control in interesting points. The thread B in the example must wait with the broadcast lock held before acquiring the writer lock, and there is no place to put user-supplied waiting code.

>What are the threads mentioned here?

[...]

PR #36244, opened 2026-01-12
Remove ListenerManager from CC fingerprinting

mlopatkin:
This PR replaces current ListenerManager-based delivery of various events used in CC fingerprinting with injecting the listener directly. This way, the events are no longer subject to ListenerManager restrictions and are generally faster.

The problems with ListenerManager-generated events are:
- They are not reentrant, so a listener cannot emit an event while processing the event of the same type. This may happen during fingerprinting when we serialize `ValueSource` parameters - this can trigger another `ValueSource` to be obtained.
- Broadcasting an event requires acquiring a lock. This can lead to deadlocks if listeners acquire other locks or broadcast other events. The fingerprint writer lock is especially problematic in that regard, see #30990.

The key observation is that the overwhelming majority of events are only consumed by the configuration cache fingerprinting, and the events are limited in the same or "shorter" scopes compared to the fingerprint writing. The handling of the events during fingerprinting is also fully reentrant. Some listeners need to come and go as we're done with configuration work, which is managed by the `FingerprintController`.

To support the last bit, we need to add a small indirection. Now `FingerpringEventHandler` receives events and simply forwards them to the active `FingerprintWriter`. The controller is still responsible for managing the active writer. Now, instead of going through the broadcaster, the clients receive the `FingerpringEventHandler` directly (as an appropriate listener interface), and invoke methods on it.

The only two listeners that still require subscriptions do so because they live in the "longer" `Global` scope. However, they were already somewhat ad hoc with their anonymous broadcasts, so I went ahead and removed `ListenerManager` there too.

Some scopes, like daemon toolchain resolver, did not have CC classes there, so we have to supply empty listener implementations there.

Besides solving the restriction issue, this also removes dispatch overhead (8-25x slower), however it is unlikely to matter much in the grand scheme of things, as there is only one listener most and we're talking about several microseconds here.

There are few downsides worth mentioning:
- This scheme only support single-consumer listeners. If we ever need something special, we'll have to do another ad hoc listener registration class, like we already do for `WorkInputs`.
- Not using ListenerManager is unusual to our codebase, so it adds certain mental overhead.
- There is certain level of contention on volatile reads of the writer reference, which isn't present in ListenerManager - there each listener type has its own reference to read.

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CoupledProjectsListener.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CoupledProjectsListener.kt
@@ -20,7 +20,10 @@ import org.gradle.api.internal.project.ProjectState
 import org.gradle.internal.service.scopes.Scope
 import org.gradle.internal.service.scopes.ServiceScope
 
-
+/**
+ * Receives events about cross project dependencies while building tooling models.
+ * These events are not sent through {@code ListenerManager}.
+ */
 @ServiceScope(Scope.Build::class)
 interface CoupledProjectsListener {
     /**
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/UndeclaredBuildInputListener.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/UndeclaredBuildInputListener.kt
@@ -21,6 +21,10 @@ import org.gradle.internal.service.scopes.ServiceScope
 import java.io.File
 
 
+/**
+ * Receives events of various "environmental things" being accessed by the build logic, like reading files or environment variables.
+ * These events are not sent through {@code ListenerManager}.
+ */
 @ServiceScope(Scope.BuildTree::class)
 interface UndeclaredBuildInputListener {
     /**
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/services/Environment.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/services/Environment.kt
@@ -34,6 +34,10 @@ class ConfigurationCacheEnvironment(
     private val listener: Listener
 ) : DefaultEnvironment() {
 
+    /**
+     * Receives events about reading environmental variables by Gradle runtime.
+     * These events are not sent through {@code ListenerManager}.
+     */
     @ServiceScope(Scope.BuildTree::class)
     interface Listener {
         fun systemPropertiesPrefixedBy(prefix: String, snapshot: Map<String, String?>)
--- a/platforms/core-configuration/file-collections/src/main/java/org/gradle/api/internal/file/collections/FileCollectionObservationListener.java
+++ b/platforms/core-configuration/file-collections/src/main/java/org/gradle/api/internal/file/collections/FileCollectionObservationListener.java
@@ -20,6 +20,10 @@ import org.gradle.api.internal.file.FileCollectionInternal;
 import org.gradle.internal.service.scopes.Scope;
 import org.gradle.internal.service.scopes.ServiceScope;
 
+/**
+ * Receives events of file collections being observed.
+ * These events are not sent through {@code ListenerManager}.
+ */
 @ServiceScope(Scope.BuildTree.class)
 public interface FileCollectionObservationListener {
     /**
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/ValueSourceProviderFactory.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/ValueSourceProviderFactory.java
@@ -50,6 +50,7 @@ public interface ValueSourceProviderFactory {
     /**
      * The listener that is notified when the value of the {@code ValueSource} is computed. There is no ordering guarantees with the
      * {@link ValueListener#valueObtained(ValueListener.ObtainedValue, ValueSource)}.
+     * These events are not sent through {@code ListenerManager}.
      */
     @ServiceScope(Scope.Build.class)
     interface ComputationListener {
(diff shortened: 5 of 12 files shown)


mlopatkin:
@bot-gradle test ACC, PT

mlopatkin:
@bot-gradle test PRF without PTS

mlopatkin:
@bot-gradle test PRF without PTS

mlopatkin:
@bot-gradle test APT

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredBuildInputsIntegrationTest.groovy
```diff
@@ -954,6 +954,34 @@ internal property sun.java.command changes.
...
+            tasks.register("run") {
+                System.properties["foo"] = new CustomSerializable()
+                doLast {
+                    println("foo in doLast = \${System.properties["foo"]}")
+                }
+            }
+        """
+
+        expect:
+        configurationCacheRun("run")
```
bamboo, line 982: 🧪 Wouldn't it make sense to assert the expected messages are printed?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/fingerprint/ConfigurationCacheFingerprintEventHandler.kt
```diff
@@ -0,0 +1,247 @@
...
+    ValueSourceProviderFactory.ValueListener,
+
+    // These listeners have to be registered separately:
+    ScriptFileResolvedListener, // This class is the single listener, but lives in a shorter scope, so cannot be injected there.
+    WorkInputListener,
+
+    // Interfaces not involved with event dispatch.
+    Closeable {
+
+    // IMPORTANT: Why isn't this class use ListenerManager as the transport?
```
bamboo:
```suggestion
    // IMPORTANT: Why doesn't this class use ListenerManager as the transport?
```

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/fingerprint/ConfigurationCacheFingerprintWriter.kt
```diff
@@ -70,28 +62,26 @@ import org.gradle.internal.execution.InputVisitor
...
 import java.util.EnumSet
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicReference
 
-
+/**
+ * This class handles various events during the build and records build configuration inputs accordingly.
+ *
+ * It doesn't listen to events directly, instead it delegates to [ConfigurationCacheFingerprintEventHandler].
+ * If you want to add another event, add it to that class.
```
bamboo, line 82: 👍

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/fingerprint/ConfigurationCacheFingerprintEventHandler.kt
```diff
@@ -0,0 +1,247 @@
...
+    // Broadcasting an event (calling a method on a listener) through LM involves holding a lock.
+    // Not taking contention into account, that lock may cause deadlocks with other locks involved in processing events.
+    // ListenerManager also forbids reentrancy (emitting an event while processing an event of the same type).
+    // Some CC fingerprinting events are inherently reentrant, e.g. handling ValueSource.obtain() call may trigger another ValueSource to be obtained.
+    // And last but not least, the versatility of the ListenerManager has a cost of 8-10 times more expensive broadcast compared to direct method call.
+    // As for most of the events this is the only implementation, the price of the versatility isn't justified.
+    @Volatile
+    var delegate: ConfigurationCacheFingerprintWriter? = null
```
bamboo: ❓ Introduce a setter that disallows setting it to anything other than `null` when it is not `null`?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/DefaultDependencyManagementServices.java
```diff
@@ -487,12 +487,29 @@ DependencyLockingHandler createDependencyLockingHandler(Instantiator instantiato
         }
 
         @Provides
-        DependencyLockingProvider createDependencyLockingProvider(FileResolver fileResolver, StartParameter startParameter, DomainObjectContext context, GlobalDependencyResolutionRules globalDependencyResolutionRules, ListenerManager listenerManager, PropertyFactory propertyFactory, FilePropertyFactory filePropertyFactory) {
+        DependencyLockingProvider createDependencyLockingProvider(
```
bamboo, line 490: 👍

subprojects/core/src/main/java/org/gradle/internal/service/scopes/DefaultWorkInputListeners.java
```diff
@@ -26,10 +25,10 @@
 import java.util.EnumSet;
 
 public class DefaultWorkInputListeners implements WorkInputListeners {
-    private final AnonymousListenerBroadcast<WorkInputListener> broadcaster;
+    private final ListenerBroadcast<WorkInputListener> broadcaster;
 
-    public DefaultWorkInputListeners(ListenerManager listenerManager) {
-        broadcaster = listenerManager.createAnonymousBroadcaster(WorkInputListener.class);
+    public DefaultWorkInputListeners() {
+        broadcaster = new ListenerBroadcast<>(WorkInputListener.class);
```
bamboo, line 31: I see, dispatching via listener manager is no more 👍

bamboo (approved):
Looks good overall.

I left a few suggestions

mlopatkin:
@bot-gradle test this

mlopatkin:
@bot-gradle cancel

PR #36259, opened 2026-01-13
Remove handling of derived names for Named definition objects

ghale:
This changes the instantiation of definition objects so that we no longer potentially derive a name from the parent object.  If a feature author wants a `Named` definition object, they need to handle deriving/setting the name themselves (via an unsafe abstract class).

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

eskatos (approved):
Nice cleanup! 👍

PR #36273, opened 2026-01-14
Fix ip gradleception test brittleness

abstratt:
Asserting on a specific number of total problems is too brittle for Gradleception tests. This PR allows to weaken that assertion so at least the number of unique problems are expected. 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
+++ b/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
@@ -426,14 +426,14 @@ class HasConfigurationCacheProblemsSpec {
     Integer problemsWithStackTraceCount
 
     /**
-     * Whether total problem count should not be strictly verified, only a number greater or equal to
+     * Whether total problem count should be strictly verified, otherwise only a number greater or equal to
      * the number of unique problems should be expected
-     * (useful when there are a large, frequently changing number of duplicates - such as in smoke/Gradleception tests
+     * (disabling this is useful when there is a large, frequently changing number of duplicates - such as in smoke/Gradleception tests
      * - but only unique problems actually matter).
      *
-     * When `true`, `totalProblemsCount` should not be specified.
+     * When `false`, `totalProblemsCount` should not be specified.
      */
-    boolean ignoreDuplicateProblemCount = false
+    boolean enforceTotalProblemCount = true
 
     /**
      * Whether to check for problem messages in the report.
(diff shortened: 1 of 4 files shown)


testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -413,18 +413,28 @@ class HasConfigurationCacheProblemsSpec {
 
     /**
      * An expectation for the total number of reported problems (including non-unique instances).
-     * {@code null} means that no expectation is defined.
+     * {@code null} means it should match the number of unique problems expected
```
abstratt, line 416: This has been incorrect. We default the expected number of total problems to the number of unique problems expected

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsSmokeTest.groovy
```diff
@@ -46,7 +46,6 @@ class GradleBuildIsolatedProjectsSmokeTest extends AbstractGradleBuildIsolatedPr
         def report = fixture.htmlReport(result.output)
 
         report.assertContents {
-            totalProblemsCount = 1
```
abstratt, line 49: this was redundant - only needed if the number of total problems is different than the number of unique problems

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/configurationcache/ConfigurationCacheProblemsFixture.groovy
```diff
@@ -413,18 +413,28 @@ class HasConfigurationCacheProblemsSpec {
...
 
+    /**
+     * Whether total problem count should not be strictly verified, only a number greater or equal to
+     * the number of unique problems should be expected
+     * (useful when there are a large, frequently changing number of duplicates - such as in smoke/Gradleception tests
+     * - but only unique problems actually matter).
+     *
+     * When `true`, `totalProblemsCount` should not be specified.
+     */
+    boolean ignoreDuplicateProblemCount = false
```
6hundreds: 🤔 Not gonna lie, I've made several attempts to accept that "Duplicate" part in the name, but I've failed. Do you think we certainly need it? Just naively, if I'd look for such a method to ignore total problem count, that "Duplicate" would make me to read java doc firstly. But may be I miss some piece of the context

abstratt: Would this be better? @6hundreds 

```
    /**
     * Whether total problem count should be strictly verified, otherwise only a number greater or equal to
     * the number of unique problems should be expected
     * (disabling this is useful when there are a large, frequently changing number of duplicates - such as in smoke/Gradleception tests
     * - but only unique problems actually matter).
     *
     * When `false`, `totalProblemsCount` should not be specified.
     */
    boolean enforceTotalProblemCount = true
```

6hundreds: Yes, that's good to me @abstratt. Thanks

PR #36278, opened 2026-01-15
Substitute unversioned capability selectors with attributes and capabilities

jvandort:
Previously we would silently ignore requests to attach attributes and capability selectors to these component selectors, leading to APIs which should work in theory to silently do something you don't expect

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/dependencysubstitution/DefaultDependencySubstitutions.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/dependencysubstitution/DefaultDependencySubstitutions.java
@@ -265,14 +265,15 @@ public class DefaultDependencySubstitutions implements DependencySubstitutionsIn
                 boolean projectInvolved = substituted instanceof ProjectComponentSelector || notation instanceof ProjectComponentSelector;
 
                 if (substituted instanceof UnversionedModuleComponentSelector) {
-                    final ModuleIdentifier moduleId = ((UnversionedModuleComponentSelector) substituted).getModuleIdentifier();
+                    UnversionedModuleComponentSelector unversioned = (UnversionedModuleComponentSelector) substituted;
+                    final ModuleIdentifier moduleId = unversioned.getModuleIdentifier();
                     if (notation instanceof ModuleComponentSelector) {
                         if (((ModuleComponentSelector) notation).getModuleIdentifier().equals(moduleId)) {
                             // This substitution is effectively a force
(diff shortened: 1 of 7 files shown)


jvandort:
@bot-gradle test this

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/dependencysubstitution/DefaultDependencySubstitutions.java
```diff
@@ -556,6 +550,10 @@ public void attributes(Action<? super AttributeContainer> configurationAction) {
                 selector = DefaultProjectComponentSelector.withAttributes((ProjectComponentSelector) selector, container.asImmutable());
             } else if (selector instanceof ModuleComponentSelector) {
                 selector = DefaultModuleComponentSelector.withAttributes((ModuleComponentSelector) selector, container.asImmutable());
+            } else if (selector instanceof UnversionedModuleComponentSelector) {
+                selector = ((UnversionedModuleComponentSelector) selector).withAttributes(container.asImmutable());
+            } else {
+                throw new IllegalArgumentException("Unknown selector type: " + selector.getClass());
```
autonomousapps, line 556: 🎉

subprojects/core-api/src/main/java/org/gradle/api/artifacts/DependencySubstitutions.java
```diff
@@ -62,7 +62,8 @@ public interface DependencySubstitutions {
     DependencySubstitutions all(Action<? super DependencySubstitution> rule);
 
     /**
-     * Create a ModuleComponentSelector from the provided input string. Strings must be in the format "{group}:{module}:{version}".
+     * Create a ModuleComponentSelector from the provided input string.
```
autonomousapps, line 65: I don't think a `UnversionedModuleComponentSelector` is a `ModuleComponentSelector`?

autonomousapps:
Thank you for this!

subprojects/core-api/src/main/java/org/gradle/api/artifacts/DependencySubstitutions.java
jvandort: indeed, we should reword this to something like `Create a component selector`, and not be specific here

autonomousapps:
I see this is in the 9.5.0 milestone. That would be great! How confident are you that this will be in Gradle 9.5.0?

PR #36280, opened 2026-01-15
Refactor predictiveTestSelectionEnabled accessor

blindpirate:
Refactor `predictiveTestSelectionEnabled` accessor to make it more readable.

### DIFF
--- a/build-logic-commons/basics/src/main/kotlin/gradlebuild/basics/BuildParams.kt
+++ b/build-logic-commons/basics/src/main/kotlin/gradlebuild/basics/BuildParams.kt
@@ -371,15 +371,19 @@ val Project.testSplitOnlyTestGradleVersion: String
 
 
 val Project.predictiveTestSelectionEnabled: Provider<Boolean>
-    get() = systemProperty(PREDICTIVE_TEST_SELECTION_ENABLED)
-        .map { it.toBoolean() }
-        .orElse(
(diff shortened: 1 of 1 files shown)


build-logic-commons/basics/src/main/kotlin/gradlebuild/basics/BuildParams.kt
```diff
@@ -371,15 +371,21 @@ val Project.testSplitOnlyTestGradleVersion: String
 
 
 val Project.predictiveTestSelectionEnabled: Provider<Boolean>
```
blindpirate, line 373: This got a bit refactored because it's already confusing enough with `orElse` and nested `zip`. Changed to plain if/else if/else blocks.

cobexer: ❓ Why do we add a heuristic to detect the promotion build instead of passing `-DenablePredictiveTestSelection=false` from promotion directly and removing all the guesswork?

blindpirate: Ah, stupid me! See https://github.com/gradle/gradle-promote/pull/315

I still want to keep the refactor as I like it to be more readable.

cobexer (approved):

PR #36292, opened 2026-01-15
Ensure all ResolvedDependencyResult instances have a target variant

jvandort:
By definition to be a resolved dependency means that an edge in the graph has successfully found a target variant to point to. Edges that have failed to select a target variant have some reason for that failure and are therefore represented as an UnresolvedDependencyResult.

Due to a number of historical bugs in the resolution engine, it was possible that an edge was forgotten or otherwise failed to be attached to variants. This left the graph in an incomplete or failed state where some edges were both not failing, but also not attached to anything. We have since resolved all known bugs that can cause this state, allowing us to remove the leniency which permits null target variants in a ResolvedDependencyResult. Now, bugs causing invalid states in the resolution engine will fail with an exception instead of silently presenting as an invalid state in a ResolutionResult.

---

The first commit in this PR resolves the last remaining bug in the resolution engine that can cause the invalid state detailed above:

In some cases an edge that would be attached to a target node is deferred, we may defer selection for the module that the edge targets. We defer selection when we know there is another component providing a constraint targeting the same module. We do this because it is possible that the constraint will change the version we target, and therefore the selection we do now may be a waste.

If an edge's attachment is deferred, we make the subsequent constraint edge responsible for attaching it when the constraint is attached. However, we had a flaw where the logic responsible for this would not attempt to attach unattached edges if there were already other attached edges to the target component.
This is possible when after we defer selection once, another edge targeting the same module 'consumes' the deferred selection and successfully attaches an edge to the target module. This is in effect a flaw with selection deferral, in that selection can only be deferred once and once a second edge comes along we perform selection immediately instead of waiting for that constraint edge that triggered the first selection deferral.

Now, we always attempt to attach unattached edges if they exist when attaching a constraint edge. This ensures we never miss a deferred edge when the constraint that caused the deferral becomes attached.

Fixes https://github.com/gradle/gradle/issues/19788

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/ResolvedGraphDependency.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/ResolvedGraphDependency.java
@@ -36,11 +36,11 @@ public interface ResolvedGraphDependency {
     ModuleVersionResolveException getFailure();
 
     /**
-     * Returns the simple id of the selected component, as per {@link ResolvedGraphComponent#getResultId()}.
+     * Returns the simple ID of the selected component, as per {@link ResolvedGraphComponent#getResultId()}.
      *
-     * @return Non-null if {@link #getFailure()} is null.
+     * @throws IllegalStateException if {@link #getFailure()} is not null.
      */
-    @Nullable Long getTargetComponentId();
+    long getTargetComponentId();
 
     /**
      * Not null only when failure is not null.
@@ -51,10 +51,10 @@ public interface ResolvedGraphDependency {
     boolean isConstraint();
 
     /**
-     * Returns the simple id of the selected variant, as per {@link ResolvedGraphVariant#getNodeId()}.
+     * Returns the simple ID of the selected variant, as per {@link ResolvedGraphVariant#getNodeId()}.
      *
-     * @return Non-null if {@link #getFailure()} is null.
+     * @throws IllegalStateException if {@link #getFailure()} is not null.
      */
-    @Nullable Long getTargetVariantId();
+    long getTargetVariantId();
 
 }
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/result/DependencyResultSerializer.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/result/DependencyResultSerializer.java
@@ -58,15 +58,9 @@ public class DependencyResultSerializer {
         componentSelectorSerializer.write(encoder, value.getRequested());
         encoder.writeBoolean(value.isConstraint());
         if (value.getFailure() == null) {
-            Long targetVariantId = value.getTargetVariantId();
-            Long targetComponentId = value.getTargetComponentId();
-            if (targetComponentId != null && targetVariantId != null) {
-                encoder.writeBoolean(true);
-                encoder.writeSmallLong(targetComponentId);
-                encoder.writeSmallLong(targetVariantId);
-            } else {
-                throw new IllegalStateException("Expected target component and variant for edge " + value);
-            }
+            encoder.writeBoolean(true);
+            encoder.writeSmallLong(value.getTargetComponentId());
+            encoder.writeSmallLong(value.getTargetVariantId());
         } else {
             encoder.writeBoolean(false);
             componentSelectionReasonSerializer.write(encoder, value.getReason());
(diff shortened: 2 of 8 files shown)


jvandort:
@bot-gradle test this

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/result/DependencyResultSerializer.java
```diff
@@ -28,7 +28,6 @@
 
 public class DependencyResultSerializer {
     private final static byte SUCCESSFUL = 0;
-    private final static byte SUCCESSFUL_NOTHING_SELECTED = 1;
     private final static byte FAILED = 2;
```
octylFractal, line 32: 💭 For the future, can we convert this to a `boolean` now? And if so, optimize by merging with the `isConstraint()` boolean write as a single `byte`?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/result/DetachedResolvedGraphDependency.java
```diff
@@ -27,25 +28,25 @@
...
+        ComponentSelector requested,
+        @Nullable Long targetComponent,
+        @Nullable ComponentSelectionReasonInternal reason,
+        @Nullable ModuleVersionResolveException failure,
+        boolean constraint,
+        @Nullable Long targetVariant
     ) {
-        assert requested != null;
-        assert failure != null || selected != null;
+        assert (failure != null && reason != null) || (targetComponent != null && targetVariant != null);
```
octylFractal: ❓ Is there a good reason to leave these as `assert` instead of checking them at runtime as well?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/result/DetachedResolvedGraphDependency.java
```diff
@@ -27,25 +28,25 @@
...
-                                           ModuleVersionResolveException failure,
-                                           boolean constraint,
-                                           Long targetVariant
+    public DetachedResolvedGraphDependency(
+        ComponentSelector requested,
+        @Nullable Long targetComponent,
+        @Nullable ComponentSelectionReasonInternal reason,
+        @Nullable ModuleVersionResolveException failure,
+        boolean constraint,
+        @Nullable Long targetVariant
```
octylFractal: ❓ If there are specific ways this should be called, perhaps there should be overloads or `static` factory methods that allow passing the relevant values without marking them as nullable. e.g. one constructor would be `DetachedResolvedGraphDependency(ComponentSelector requested, Long targetComponent, boolean constraint, Long targetVariant)`, and the other would be `DetachedResolvedGraphDependency(ComponentSelector requested, ComponentSelectionReasonInternal reason, ModuleVersionResolveException failure, boolean constraint)`, without any `@Nullable` annotations.

Or more likely, this should just be two separate implementations, but that's more involved.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/result/DetachedResolvedGraphDependency.java
jvandort: This is really just a crappy alternative for multiple constructors. The proper way to do this would be to have two constructors and no asserts/runtime checks at all. These checks are are really to ensure the developer is calling the constructor the "right way"

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/result/DetachedResolvedGraphDependency.java
jvandort: I will add two static factories instead of this mess here

jvandort:
I created two separate implementations of the interface as suggested and switched to serializing a boolean instead of an integer. I would rather avoid the complexity of bit packing the boolean fields for now for the sake of simplicity and maintainability.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/ResolvedGraphDependency.java
```diff
@@ -37,9 +37,10 @@ public interface ResolvedGraphDependency {
 
     /**
      * Returns the simple id of the selected component, as per {@link ResolvedGraphComponent#getResultId()}.
+     *
+     * @return Non-null if {@link #getFailure()} is null.
      */
-    @Nullable
-    Long getSelected();
+    @Nullable Long getTargetComponentId();
```
big-guy: Not now, but do we really need this or the selected variant below to be boxed?

jvandort: We could probably throw an exception if you call this on a failed dependency instead of returning null to avoid boxing in the successful case

PR #36283, opened 2026-01-15
Update smoke and performance tests using NowInAndroid to use AGP 9

eskatos:
This PR 
* follows-up https://github.com/gradle/gradle/pull/36254 which introduced AGP 9 RC
* fixes https://github.com/gradle/gradle/issues/35423
* updates the referenced NowInAndroid commit to use a version that works with AGP 9, but not with AGP 8 anymore
* changes the NiA-based smoke tests to only test AGP 9
* changes the NiA-based performance tests to work with AGP 9

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #36293, opened 2026-01-15
Update tested AGP versions, bringing 9.0.0 GA and dropping 8.x

eskatos:
AGP 8.x is officially supported only with Gradle 8.x
AGP 9.x is officially supported only with Gradle 9.x

Now that AGP 9.0.0 GA is out we can drop best-effort support and test coverage for AGP 8.x

This PR
* follows up https://github.com/gradle/gradle/pull/36283
* cleans up now useless tests, expectations, fixtures, etc.
* upgrade all documentation snippets and samples to AGP 9

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic/build-update-utils/src/test/groovy/gradlebuild/buildutils/tasks/UpdateAgpVersionsTest.groovy
+++ b/build-logic/build-update-utils/src/test/groovy/gradlebuild/buildutils/tasks/UpdateAgpVersionsTest.groovy
@@ -26,9 +26,9 @@ class UpdateAgpVersionsTest extends Specification {
         given:
         def gradleVersion = GradleVersion.version("9.2")
         def allVersions = [
+            "9.0.0-alpha01", "9.0.0-beta01", "9.0.0-rc01",
             "8.8.0", "8.9.0",
-            "9.0.0-alpha01", "9.0.0-beta01", "9.0.0-rc01"
-        ].shuffled()
+        ]
 
         when:
         def selected = UpdateAgpVersions.selectVersionsFrom(gradleVersion, null, allVersions)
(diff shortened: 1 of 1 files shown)


eskatos:
@bot-gradle test AFT AST APT

eskatos:
@bot-gradle test AFT AST APT

eskatos:
@bot-gradle test AFT

eskatos:
@bot-gradle test AST

eskatos:
@bot-gradle test AST

jvandort:
Would it make sense to keep one 8.x release under test? It may be nice to remain compatible with that older releases for users who have trouble upgrading to 9. Plus it acts as a test to verify we dont break existing behavior.

Were we doing anything overly annoying to maintain 8.x compatibility? In practice it does seem like there is _some_ compatibility between android 8.x and gradle 9.x otherwise these tests woudlnt have been passing all along.

eskatos:
@jvandort 

> In practice it does seem like there is some compatibility between android 8.x and gradle 9.x otherwise these tests woudlnt have been passing all along.

We were running just a few small smoke tests against AGP 9 RCs with special handling for the incompatibilities.

> Were we doing anything overly annoying to maintain 8.x compatibility?

For one, we are refraining from refactoring our internals that AGP 8.x wrongly used (e.g. https://github.com/gradle/gradle/pull/34975) just to keep compatibility until now (Gradle 9.3). If we keep AGP 8 compatibility moving forward, we can't touch these internals.

> It may be nice to remain compatible with that older releases for users who have trouble upgrading to 9.

We've been doing that since 9.0. What about we wait for Gradle 9.4 branching before merging this PR and the incompatibility boundary would be Gradle 9.5?

jvandort:
I dont have strong feelings about this and was more curious about the motivation.

My general idea is that if 8.x runs on current gradle, and since our APIs and behaviors are stable within the same major versions, i would expect the "supported" agp 8.x versions to remain functioning for the remainder of 9.x.

The fact that 8.x relies on internal APIs makes it more clear why we would end up breaking these old agp versions in a minor

jvandort:
I dont have strong feelings about this and was more curious about the motivation.

My general idea is that if 8.x runs on current gradle, and since our APIs and behaviors are stable within the same major versions, i would expect the "supported" agp 8.x versions to remain functioning for the remainder of 9.x.

The fact that 8.x relies on internal APIs makes it more clear why we would end up breaking these old agp versions in a minor

jvandort:
If its not too much trouble i think waiting for 9.5 might help out the community in terms of migration pains

eskatos:
* Superseded by https://github.com/gradle/gradle/pull/37198

Issue #35184, opened 2025-10-01
[Backport] Different fingerprints in the compile classpath for the same dependency


asodja:
See issue: https://github.com/gradle/gradle/issues/32464. We should also backport a reproducer added with https://github.com/gradle/gradle/pull/34150.

It was fixed by using Kotlin snapshotter, but we could also fix it for 8.x by appending fallback zip hasher configuration to hasher if not null here:https://github.com/gradle/gradle/blob/bc186c09d72c9cf0b93ea623c2ba0c5b8bf442a4/platforms/jvm/normalization-java/src/main/java/org/gradle/api/internal/changedetection/state/ZipHasher.java#L83-L87 

From: https://gradle.slack.com/archives/C0WTFHWT1/p1751885846837139?thread_ts=1750438715.527669&cid=C0WTFHWT1
> Anže Sodja
> Bisect shows that 5312efecafcec0fb0e8573b494d88ab9dd98c70b is the first commit that fixes this.
> This is from this PR: https://github.com/gradle/gradle/pull/26645/

> Anže Sodja
> It looks to me that
> ClasspathFingerprintingStrategy.compileClasspathFallbackToRuntimeClasspath
> [https://github.com/gradle/gradle/blob/cb0c615fb8e3690971bb7f89ad80f58943360624/pla[…]fingerprint/classpath/impl/ClasspathFingerprintingStrategy.java](https://github.com/gradle/gradle/blob/cb0c615fb8e3690971bb7f89ad80f58943360624/platforms/jvm/normalization-java/src/main/java/org/gradle/internal/fingerprint/classpath/impl/ClasspathFingerprintingStrategy.java#L122-L132)
could poison the cache with incorrect hash.

> Anže Sodja
> Yes, that seems to be the reason. The problem is that for ZipHasher we don't take into account fallback zip hasher when calculating configuration hash here:
> [https://github.com/gradle/gradle/blob/bc186c09d72c9cf0b93ea623c2ba0c5b8bf442a4/pla[…]va/org/gradle/api/internal/changedetection/state/ZipHasher.java](https://github.com/gradle/gradle/blob/bc186c09d72c9cf0b93ea623c2ba0c5b8bf442a4/platforms/jvm/normalization-java/src/main/java/org/gradle/api/internal/changedetection/state/ZipHasher.java#L83-L87)
Luckily in Gradle 9 we don't use this fallback strategy so this is not an issue anymore

[...]

PR #36334, opened 2026-01-19
[Backport] Different fingerprints in the compile classpath for the same dependency

asodja:
Fixes https://github.com/gradle/gradle/issues/35184


asodja:
@bot-gradle test this

PR #36335, opened 2026-01-19
Add build operation around Develocity build finished callback

wolfs:
for making the Develocity build finished callback visible in build operation traces. Note that currently you cannot measure this build operation in Gradle profiler, since the BuildService responsible for measuring it has already been closed when the build operation fires. Though you can use `--build-ops-trace` to capture it.


wolfs:
@bot-gradle merge

PR #36355, opened 2026-01-20
Increase memory for IDE sync test

alllex:
since it's a recent source of flakiness for the `IsolatedProjectsGradleceptionSyncTest`

https://ge.gradle.org/s/4cmna7ss7stn6/tests/task/:smoke-ide-test:smokeIdeTest/details/org.gradle.ide.sync.IsolatedProjectsGradleceptionSyncTest/can%20sync%20gradle%2Fgradle%20build%20without%20problems?focused-execution=1&page=eyJvdXRwdXQiOnsiMCI6M319&top-execution=1#L441

---

Fixes https://github.com/gradle/gradle-private/issues/5028


jvandort:
Shouldnt we theoretically use the same value here as we do in the jvmargs xmx in the root gradle.properties? Is there any reason these should be different?

alllex:
@jvandort, this setting configures how much memory we give to the IDEA that is trying to sync the project, independently from the Gradle daemon that will do Gradle things

https://github.com/gradle/gradle-ide-starter/blob/f5452be3dcffd06cb5f89c05a5fb2f6260d280d6/README.md?plain=1#L34

jvandort:
I see. Disregard then

PR #36356, opened 2026-01-20
Report problem when multiple build file candidates are found.

### LINKED ISSUE #36343: Build scripts should report a warning when multiple candidates are present

hegyibalint:
Fixes #36343

### DIFF
--- a/platforms/core-runtime/build-discovery-reporting/src/main/java/org/gradle/internal/scripts/ScriptResolutionResultReporter.java
+++ b/platforms/core-runtime/build-discovery-reporting/src/main/java/org/gradle/internal/scripts/ScriptResolutionResultReporter.java
@@ -20,7 +20,6 @@ import org.gradle.api.problems.ProblemId;
 import org.gradle.api.problems.ProblemReporter;
 import org.gradle.api.problems.internal.GradleCoreProblemGroup;
 
-import javax.inject.Inject;
 import java.io.File;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


hegyibalint:
@bot-gradle test this

hegyibalint:
@bot-gradle test this

platforms/core-runtime/build-discovery-reporting/src/main/java/org/gradle/internal/scripts/ScriptResolutionResultReporter.java
```diff
@@ -30,11 +29,7 @@ public class ScriptResolutionResultReporter {
     private final ProblemReporter problemReporter;
 
     @Inject
```
eskatos: ❔ Is the `@Inject` annotation required? I only found direct instantiation of this class.

hegyibalint: Not necessarily; it's rather a reflex from my end to mark constructors containing injected services.
Does this cause a problem in your opinion?

eskatos: It causes assumptions. When I read that I assume this is instantiated by something that do injection.
Let's remove it please.

hegyibalint: That is a good point, I will keep this in mind in the future.
I've removed the annotation.

PR #36362, opened 2026-01-20
Restore ModuleVersionSelector conversion, but deprecated

### LINKED ISSUE #36359: ExternalDependency and DependencyConstraint cannot be passed to DependencyResolveDetails#useTarget
### Current Behavior

When running the reproducer, an error occurs:
```
> Could not resolve org.example:bar:1.0.0.
  Required by:
      root project 'simple-testing-project'
   > Cannot convert the provided notation to an object of type ModuleComponentSelector: junit:junit:4.13.2.
     The following types/formats are supported:
       - Instances of ModuleComponentSelector.
       - String or CharSequence values, for example 'org.gradle:gradle-core:1.0'.
       - Maps, for example [group: 'org.gradle', name:'gradle-core', version: '1.0'].
       - Version catalog type-safe accessors.
```

### Expected Behavior

It should function as it did in 9.2.1 and prior, printing the path to the jar:
```
[$GRADLE_USER_HOME/caches/modules-2/files-2.1/junit/junit/4.13.2/8ac9e16d933b6fb43bc7f576336b8f4d7eb5ba12/junit-4.13.2.jar]
```

### Context (optional)

_No response_

In `build.gradle.kts`:
```
plugins {
    java
}

repositories {
    mavenCentral()
}

configurations.all {
    resolutionStrategy {
        eachDependency {
            useTarget(project.dependencies.create("junit:junit:4.13.2"))
        }
    }
}

dependencies {
    implementation("org.example:bar:1.0.0")
}

println(files(configurations["compileClasspath"]).toSet())
```
Run `./gradlew build`.

### Gradle version

9.3.0

9.2.1

### Build scan URL (optional)

_No response_

_No response_

octylFractal:
Fixes #36359

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/dsl/ModuleComponentSelectorParsers.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/dsl/ModuleComponentSelectorParsers.java
@@ -18,13 +18,17 @@ package org.gradle.api.internal.artifacts.dsl;
 
 import org.gradle.api.IllegalDependencyNotation;
 import org.gradle.api.InvalidUserDataException;
+import org.gradle.api.artifacts.ExternalDependency;
 import org.gradle.api.artifacts.MinimalExternalModuleDependency;
+import org.gradle.api.artifacts.ModuleVersionSelector;
 import org.gradle.api.artifacts.VersionConstraint;
 import org.gradle.api.artifacts.component.ModuleComponentSelector;
 import org.gradle.api.internal.artifacts.DefaultModuleIdentifier;
+import org.gradle.api.internal.artifacts.dependencies.DefaultImmutableVersionConstraint;
 import org.gradle.api.provider.Provider;
 import org.gradle.api.provider.ProviderConvertible;
 import org.gradle.internal.component.external.model.DefaultModuleComponentSelector;
+import org.gradle.internal.deprecation.DeprecationLogger;
 import org.gradle.internal.exceptions.DiagnosticsVisitor;
 import org.gradle.internal.typeconversion.MapKey;
 import org.gradle.internal.typeconversion.MapNotationConverter;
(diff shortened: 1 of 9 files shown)


platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/dsl/ModuleComponentSelectorParsers.java
```diff
@@ -162,4 +168,45 @@ private boolean isNotRequiredVersionOnly(VersionConstraint constraint) {
         }
 
     }
+
+    static class ExternalDependencyConverter extends TypedNotationConverter<ExternalDependency, ModuleComponentSelector> {
```
jvandort, line 172: This was an oversight from the prior change, but some of the above error messages include `ModuleVersionSelector` but should instead say `ModuleComponentSelector`

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/dsl/ModuleComponentSelectorParsers.java
```diff
@@ -162,4 +168,45 @@ private boolean isNotRequiredVersionOnly(VersionConstraint constraint) {
...
+        }
+
+        public ExternalDependencyConverter() {
+            super(new TypeInfo<>(ExternalDependency.class));
+        }
+
+        @Override
+        protected ModuleComponentSelector parseType(ExternalDependency notation) {
+            return DefaultModuleComponentSelector.newSelector(
+                notation.getModule(), DefaultImmutableVersionConstraint.of(notation.getVersion())
```
jvandort: You should able to call `getVersionConstraint` on the notation

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/dsl/ModuleComponentSelectorParsers.java
```diff
@@ -162,4 +168,45 @@ private boolean isNotRequiredVersionOnly(VersionConstraint constraint) {
...
+        }
+
+        public ModuleVersionSelectorConverter() {
+            super(new TypeInfo<>(ModuleVersionSelector.class));
+        }
+
+        @Override
+        protected ModuleComponentSelector parseType(ModuleVersionSelector notation) {
+            DeprecationLogger.deprecateAction("Converting an instance of ModuleVersionSelector to ModuleComponentSelector")
+                .withAdvice("If possible, don't create ModuleVersionSelector instances and pass the original notation directly. If not, combine the group, name, and version into the String notation.")
```
jvandort: I think we should adjust the wording to remove the `if possible`. It better be possible that they can do this or else we can't proceed with the deprecation. I think we should recommend the string notation outright if they hit this deprecation

jvandort:
Would like to hear @big-guy 's thoughts on this

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/dsl/ModuleComponentSelectorParsers.java
octylFractal: There are some APIs that still only return `ModuleVersionSelector` right now though (e.g. `DependencyResolveDetails#getTarget()` or `getRequested()`). I don't know if these make any sense to be passed in, but we don't have alternatives right now.

I'll change the wording since these seem a little weird to be using in the first place though.

PR #36367, opened 2026-01-21
Split out compiler workers

jvandort:
Previously, for java, scala, and groovy, both the tasks that run in the daemon and the compiler actions that run in the worker, lived in the same projects. This prevented us from upgrading the version of java that the tasks executed with. 

This was particularly a problem since `langauge-java` depends on `dependency-management`, which prevents us from upgrading the version of java that DM compiles against. 

So, we split out the compiler worker actions for java, scala, and groovy into separate projects:
- `java-compiler-worker`
- `scala-compiler-worker`
- `groovy-compiler-worker`

These contain only classes that the worker loads at runtime and nothing more. Minimal code changes were made as part of splitting up these projects, so these compiler workers might have slightly larger compile classpaths than ideal. Most notably, they depend on `core` and `core-api`, which are still monoliths. This prevents these projects from upgrading to newer java versions, as workers must maintain java 8 compatibility for the foreseeable future. 

The splitting of these projects was performed with the help of a new `SplitProjectTask`, which takes in a list of classes to split out of a project, determines its transitive dependency classes, and moves all of them at once. This makes it much easier to split classes and will be incredibly valuable in the future to help reshape gradle's architecture. 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic/refactoring/src/main/kotlin/gradlebuild/refactoring/DependencyAnalyzer.kt
+++ b/build-logic/refactoring/src/main/kotlin/gradlebuild/refactoring/DependencyAnalyzer.kt
@@ -49,7 +49,7 @@ private data class ClassDependencies(val api: Set<String>, val impl: Set<String>
 }
 
 /**
- * Analyzes Java/Kotlin class files to determine their dependencies.
+ * Analyzes class files to determine their dependencies.
  *
  * This class uses the ASM library to parse bytecode and extract dependency information.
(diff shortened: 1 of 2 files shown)


jvandort:
@bot-gradle test this

PR #36370, opened 2026-01-21
Unify how we invalidate work outputs in execution engine

asodja:
This is a small cleanup after https://github.com/gradle/gradle/pull/36227

Now AssignImmutableWorkspaceStep doesn't do any fileSystem invalidation before work (unless it re-reads an entry). This is anyway done in BuildCacheStep and BroadcastChangingOutputsStep. It saves one `fileSystemAccess.invalidate()` call and it unifies this logic between mutable and immutable work.

### DIFF
--- a/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
+++ b/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
@@ -223,9 +223,9 @@ public class AssignImmutableWorkspaceStep<C extends IdentityContext> implements 
         // There is no previous execution in the immutable case
         PreviousExecutionContext previousExecutionContext = new PreviousExecutionContext(workspaceContext, null);
 
-        // Ensure an empty directory in case of stale files
+        // Ensure an empty directory in case of stale files and execute
+        // Snapshot invalidation is handled later by BuildCacheStep and BroadcastChangingOutputsStep
         ensureEmptyDirectory(workspaceDir);
-        // Execute
         CachingResult delegateResult = delegate.execute(work, previousExecutionContext);
 
         if (delegateResult.getExecution().isSuccessful()) {
--- a/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/OutputVisitor.java
+++ b/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/OutputVisitor.java
@@ -17,20 +17,36 @@
 package org.gradle.internal.execution;
 
 import org.gradle.api.file.FileCollection;
+import org.gradle.api.tasks.Destroys;
+import org.gradle.api.tasks.LocalState;
+import org.gradle.api.tasks.OutputDirectory;
+import org.gradle.api.tasks.OutputFile;
 import org.gradle.internal.file.TreeType;
 
 import java.io.File;
 import java.util.function.Supplier;
 
 public interface OutputVisitor {
+
+    /**
+     * Visits the output property for work, for example for task output properties are annotated with {@link OutputDirectory} or {@link OutputFile} and similar.
+     */
     default void visitOutputProperty(
         String propertyName,
         TreeType type,
         OutputFileValueSupplier value
     ) {}
 
+    /**
+     * Visits the local state location, for example for task local state properties are annotated with {@link LocalState}.
+     * Other work types don't have local state.
+     */
     default void visitLocalState(File localStateRoot) {}
 
+    /**
+     * Visits the destroyable for work, for example for task destroyable properties are annotated with {@link Destroys}.
+     * Other work types don't have destroyable.
+     */
     default void visitDestroyable(File destroyableRoot) {}
 
     abstract class OutputFileValueSupplier implements InputVisitor.FileValueSupplier {


asodja:
@bot-gradle test this

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/BuildCacheStep.java
```diff
@@ -227,11 +224,5 @@ public void visitOutputProperty(String propertyName, TreeType type, OutputFileVa
                 }
             });
         }
-
-        // TODO Make this much more explicit
-        public boolean shouldInvalidateOutputsBeforeLoad() {
-            // We don't need to invalidate outputs for immutable work as it's executed in a unique temporary workspace
```
asodja, line 233: 💭 This is not true anymore, but it works with version on release, since we anyway invalidate cache in AssignImmutableWorkspaceStep

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/UnitOfWork.java
```diff
@@ -151,5 +154,43 @@ default Optional<String> getBuildOperationWorkType() {
...
+    default List<String> getAllOutputLocationsForInvalidation(File workspace) {
+        ImmutableList.Builder<String> builder = ImmutableList.builder();
+        visitOutputs(workspace, new OutputVisitor() {
+            @Override
+            public void visitOutputProperty(String propertyName, TreeType type, OutputFileValueSupplier value) {
+                builder.add(value.getValue().getAbsolutePath());
+            }
+
+            @Override
+            public void visitLocalState(File localStateRoot) {
```
alllex, line 169: 💭 Can we use this opportunity to document the methods on the OutputVisitor? At this point, without prior knowledge it's hard to tell why here we visit everything, while in the "cached" version below we only visit "output properties". What are even output properties?

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/UnitOfWork.java
```diff
@@ -151,5 +154,43 @@ default Optional<String> getBuildOperationWorkType() {
         return Optional.empty();
     }
 
+    /**
+     * Returns all touched output locations that should be invalidated before work.
+     */
+    default List<String> getAllOutputLocationsForInvalidation(File workspace) {
```
alllex, line 160: ❔ Do we only invalidate what we currently see on disk?

Do we cover edge cases when something was deleted on disk, but we still have that entry in caches? In other words, is it possible that broken state on disk can cause us to never invalidate some cache as long as the daemon is alive?

I might be significantly misunderstanding this, so this question is mostly for education.

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
```diff
@@ -225,8 +225,6 @@ private WorkspaceResult executeInWorkspace(UnitOfWork work, C context, Immutable
 
         // Ensure an empty directory in case of stale files
         ensureEmptyDirectory(workspaceDir);
-        // Invalidate snapshots since we cleaned the directory
```
alllex, line 228: 💭 I wonder if we should still document somehow the invalidation strategy. Or do you think it's clear enough that this happens in the `BuildCacheStep`?

I am afraid it might be a bit too far away, despite being an important part of the protocol.

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/UnitOfWork.java
asodja: We invalidate what is work has registered as an output. Example test sets output dir to `build/test/dir` and we will invalidate that path. We don't check if it exists on the disk actually.

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/UnitOfWork.java
asodja: Sure, I will add a comment. For tasks it's basically what is annotated as `@OutputDirectory` it's an `outputProperty`, what is annotated with `@LocalState` it's local state and what is annotated with `@Destroys` it's destroyable. 

Only tasks have local state and destroyables, so other work has only output properties

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/steps/AssignImmutableWorkspaceStep.java
asodja: I added a comment that we invalidate snapshots later

asodja:
@bot-gradle test QFL

platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/UnitOfWork.java
alllex: TIL we have `@LocalState` and `@Destroys`

PR #36377, opened 2026-01-21
Enable IP in the build

alllex:
Enables Isolated Project in the gradle/gradle build itself to allow for frictionless dogfooding on developer machines

Fixes https://github.com/gradle/gradle-private/issues/5057

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,6 +4,7 @@ org.gradle.caching=true
 org.gradle.configuration-cache=true
 org.gradle.configuration-cache.parallel=true
 org.gradle.configuration-cache.entries-per-key=2
+org.gradle.unsafe.isolated-projects=true
 systemProp.org.gradle.classloaderscope.strict=true
 systemProp.gradle.publish.skip.namespace.check=true
 # Kotlin DSL settings


alllex:
@bot-gradle test RFR without pts

alllex:
@bot-gradle cancel

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild/testcleanup/TestFilesCleanupPlugin.kt
```diff
@@ -0,0 +1,42 @@
...
+import java.io.File
+
+class TestFilesCleanupPlugin : Plugin<Project> {
+    override fun apply(project: Project) {
+        val testFilesCleanupExtension = project.extensions.create<TestFileCleanUpExtension>("testFilesCleanup").apply {
+            reportOnly.convention(false)
+        }
+        project.gradle.taskGraph.whenReady {
+            val testFilesCleanupService = project.gradle.sharedServices.registerIfAbsent("testFilesCleanupBuildService-" + project.name, TestFilesCleanupService::class.java) {
+                parameters.rootBuildDir.set(File("/Users/bo/Projects/gradle/build"))
```
blindpirate: @alllex How to access root project's build dir without violating IP?

alllex: In the main build we do it like so, throught a custom extension:
https://github.com/gradle/gradle/blob/8c0f6e0886686eed9522a547af50ceb12804aec4/packaging/distributions-full/build.gradle.kts#L37

Can you use the same approach here?

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild/testcleanup/TestFilesCleanupPlugin.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.gradle.kotlin.dsl.*
+import org.gradle.kotlin.dsl.support.serviceOf
+import java.io.File
+
+class TestFilesCleanupPlugin : Plugin<Project> {
+    override fun apply(project: Project) {
+        val testFilesCleanupExtension = project.extensions.create<TestFileCleanUpExtension>("testFilesCleanup").apply {
+            reportOnly.convention(false)
+        }
+        project.gradle.taskGraph.whenReady {
```
jvandort: taskGraph.whenReady is not IP safe. We need a solution that doesn't rely on this

alllex: IIUC, it's only used here as a point-in-time callback. If we don't touch tasks in the task graph, then there is no IP violation

jvandort: Why do we need to create this service exactly at this point in time?

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.ci-reporting.gradle.kts
```diff
@@ -14,32 +14,17 @@
...
-    rootProject.plugins.apply(TestFilesCleanupRootPlugin::class.java)
-    val globalExtension = rootProject.extensions.getByType<TestFilesCleanupBuildServiceRootExtension>()
-
-    val projectState = objects.newInstance(TestFilesCleanupProjectState::class.java)
-
-    globalExtension.projectStates.put(path, projectState)
-    projectState.projectBuildDir = layout.buildDirectory
-    projectState.projectPath = path
-    projectState.reportOnly = testFilesCleanup.reportOnly
+    plugins.apply(TestFilesCleanupPlugin::class.java)
```
jvandort: Can we just inline the logic here? I'm not sure I see why its necessary to have a separate plugin encapsulate this logic

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild/testcleanup/TestFilesCleanupPlugin.kt
octylFractal: It looks like it's not necessary, I pushed a commit without it.

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.ci-reporting.gradle.kts
```diff
@@ -14,32 +14,31 @@
...
-    val projectState = objects.newInstance(TestFilesCleanupProjectState::class.java)
-
-    globalExtension.projectStates.put(path, projectState)
-    projectState.projectBuildDir = layout.buildDirectory
-    projectState.projectPath = path
-    projectState.reportOnly = testFilesCleanup.reportOnly
+    val testFilesCleanupService = project.gradle.sharedServices.registerIfAbsent("testFilesCleanupBuildService-" + project.name, TestFilesCleanupService::class.java) {
+        parameters.rootBuildDir.set(project.layout.settingsDirectory.dir("build"))
+        parameters.testFilesCleanupExtension.set(testFilesCleanupExtension)
+        parameters.projectPath.set(project.path)
```
jvandort: IIUC this plugin is applied to all jvm projects. I am not sure of the behavior of the configure block. Does it only run for the first project that registered the service? Or does it execute for all projects that call this method?

In any case, calls like `parameters.projectPath.set(project.path)` are very suspicious. Does each project just overwrite the previous project's previous value?

octylFractal: The service is registered with a different name for each project, based on the project name, so they're all different instances.

jvandort: Why create a service at all then?

blindpirate: @jvandort the intention of this plugin/service is running after all tasks finished, collecting the files/reports etc, zipping them to root project build dir so that they can be published as TeamCity artifact, like this:

image

jvandort: But do we need to use a build service in order to register the `OperationCompletionListener`? I understand needing the listener, but does it need to be a `gradle.sharedServices` service if it is no longer shared between projects?

alllex: Yes, because of the CC: https://github.com/gradle/gradle/blob/5173382ca9f9d04249ebff1e745746f0aa0b8e59/subprojects/build-events/src/main/java/org/gradle/build/event/BuildEventsListenerRegistry.java#L44-L46

alllex:
The build logic changes added to this PR has moved, so let's move the discussions and futher comments there:
- https://github.com/gradle/gradle/pull/36550

Now that the CI has guardrails (https://github.com/gradle/gradle/pull/36503) against enabling IP in the gradle.properties, I'll update this PR to only include a single change -- enable IP in the gradle.properties.

alllex:
@bot-gradle test on linux without PTS

alllex:
@bot-gradle test RfR

PR #36401, opened 2026-01-22
cobexer/improve binary compatibility report

cobexer:
- **build-logic: binary-compatibility: remove jQuery dependency**
- **build-logic: binary-compatibility: fix buttons acting on the wrong section**


ov7a:
> build-logic: binary-compatibility: remove jQuery dependency

So the code got switched to vanilla JS, but the original report still uses it, right? I don't see the dependency removal

cobexer:
Yes the plugin still uses jQuery in it's implementation of the report.

We no longer depend on that as our added code is now plain JavaScript.

PR #36406, opened 2026-01-22
Fix progress bar on git bash for windows

### LINKED ISSUE #36319: Progress bar garbled in Git Bash for Windows 2.47 or later, or with experimental pseudo consoles support enabled
### Current Behavior

When using any recent release of Git Bash for Windows, the new Gradle 9.4.0 progress bar and other build outputs appear to be garbled, producing incorrect Unicode characters and overwriting each other. Line feeds and/or carriage returns appear to be wrong.

Image

### Expected Behavior

The Gradle progress bar and other build outputs should always appear normally, either in the old 9.3.0 style or the new 9.4.0 style, even if the user is using a recent version of Git Bash for Windows as his terminal.

The first screenshot is from Gradle 9.3.0 on Git Bash for Windows 2.52, and the second is from Gradle 9.4.0 on Git Bash for Windows 2.46.2. Both appear to be working normally.

Image

Image

### Context (optional)

This issue is reproducible in Git Bash for Windows version 2.47 or later using the default installation options.

In Git Bash for Windows version 2.46.2 or earlier, the issue can only be reproduced if you select the option "Enable experimental support for pseudo consoles" during installation. Using the default installation options, the issue cannot be reproduced.

I have linked a reproducer project here, but it has no special properties: it's just a simple project created with gradle init: https://github.com/dglynch/GradleReproducer

### Gradle version

9.4.0-20260116012355+0000

9.3.0

### Build scan URL (optional)

_No response_

Windows 11
Git Bash from https://github.com/git-for-windows/git/releases/tag/v2.52.0.windows.1

reinsch82:
Fixes #36319

Only use UTF-8 if the code page is 65001

another change is for the unicode progress bar characters, we decided these fit better.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

platforms/core-runtime/native/src/main/java/org/gradle/internal/nativeintegration/console/WindowsConsoleDetector.java
```diff
@@ -28,7 +28,16 @@ public ConsoleMetaData getConsole() {
         // Use Jansi's detection mechanism
         try {
             new WindowsAnsiProcessor(new PrintStream(new ByteArrayOutputStream()), true);
-            return FallbackConsoleMetaData.ATTACHED;
+            boolean disableUnicodeSupportDetection = NativePlatformConsoleDetector.isWindowsUnicodeCodePage(); //see https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers
```
eskatos: ❌ Let's move the documentation link closer to what it is about, on the method that actually check the code page identifier.

platforms/core-runtime/native/src/main/java/org/gradle/internal/nativeintegration/console/NativePlatformConsoleDetector.java
```diff
@@ -57,6 +60,10 @@ public ConsoleMetaData getConsole() {
         }
     }
 
+    public static boolean isWindowsUnicodeCodePage() {
+        return Kernel32.GetConsoleOutputCP() != 65001;
+    }
```
eskatos: ❌ This is quite confusing to read. 65001 is the codepage for `Unicode (UTF-8)` according to the Windows documentation. As I read it, either the name is wrong and should be `isNot...`, or the equality check should be reversed.

I also think it would be better to extract a constant with an explicit name like `WINDOWS_UTF8_CODEPAGE_ID`.

platforms/core-runtime/native/src/main/java/org/gradle/internal/nativeintegration/console/NativePlatformConsoleDetector.java
```diff
@@ -18,12 +18,14 @@
...
 import net.rubygrapefruit.platform.NativeException;
 import net.rubygrapefruit.platform.terminal.Terminals;
+import org.fusesource.jansi.internal.Kernel32;
 import org.gradle.internal.os.OperatingSystem;
 
 import static net.rubygrapefruit.platform.terminal.Terminals.Output.Stderr;
 import static net.rubygrapefruit.platform.terminal.Terminals.Output.Stdout;
 
 public class NativePlatformConsoleDetector implements ConsoleDetector {
+    public static final int WINDOWS_UTF8_CODEPAGE_ID = 65001;
```
eskatos: ❓ Why is this constant public?

platforms/core-runtime/native/src/main/java/org/gradle/internal/nativeintegration/console/NativePlatformConsoleDetector.java
```diff
@@ -57,6 +62,14 @@ public ConsoleMetaData getConsole() {
         }
     }
 
+    public static boolean isWindowsUnicodeCodePage() {
```
eskatos: 💅 Now that this function checks two things, for explicitness. Also, doesn't need to be public.
```suggestion
    static boolean isWindowsWithUnicodeCodePage() {
```

platforms/documentation/docs/src/docs/release/release-notes-assets/gradle-progress-bar-new.gif
eskatos, line 1: 💅 The previous cast shown the progress bar from start to completion, that was much better. I'd rather we don't show an example with 15+ seconds of configuration that doesn't even complete. Also, we can see the mouse cursor in the middle.

eskatos, line 1: ❤️

dglynch:
I have tested the fix in version 9.4.0-20260124022857+0000 and it looks good. I can no longer reproduce the issue.

reinsch82:
@dglynch Thanks for your help :)

Issue #2270, opened 2017-06-09
Default file encoding for daemon should be UTF-8


lptr:
Currently we don't set the `file.encoding` property when starting daemons. Most systems default to UTF-8, but Windows typically uses CP-1252. The difference in default encoding can hurt reusing cached task outputs from a shared cache when used by different OSs.

By defaulting to UTF-8 we are making this discrepancy go away. Users who need to use a different default encoding can make it explicit via `org.gradle.jvmargs`.

### Expected Behavior

Daemon is started with `utf-8` as the default encoding.

### Current Behavior

Daemon is started with the system default encoding. Changing the system default encoding does not change the default encoding for an already running daemon.

### Context

Tasks using `file.encoding` can lead to unintended cache misses when using a shared cache due to the difference in encoding.

comdiv:
Cannot set UTF-8 properties in gradle.properties if Windows is used for compilation - it read them as CP-1251 (in my case)

irgaly:
Any roadmap for this?
I'll feel happy if we don't have to write `org.gradle.jvmargs=-Dfile.encoding=...` in gradle.properties any more.

ov7a:
Related issues:
* https://github.com/gradle/gradle/issues/24498
* https://github.com/gradle/gradle/issues/16763
* https://github.com/gradle/gradle/pull/36407 - might be a breaking change

liutikas:
Is this still targeted for 9.0?

ljacomet:
Unfortunately, no.

We will not achieve this in the time left.

stolsvik:
Do we _still_ need to include this thing `org.gradle.jvmargs=-Dfile.encoding=UTF-8` in `gradle.properties`?

PR #36407, opened 2026-01-22
Source files can have platform-dependent encoding.

### LINKED ISSUE #36399: Non utf-8 c code is no longer buildable
### Current Behavior

If you have a single byte in your c files thats not encoded in utf-8, the build will fail

### Expected Behavior

Encoding is handled a bit nicer...
So either restore the old behavior of 'eh whatever we only need to understand the ascii bits' and ignore errors or make encoding configurable.
If by an offchance this was intended, add this change to the upgrade guide/release notes

### Context (optional)

Ii looks like to me this was caused by 0d44f70852e69ab46eb85c58f856e7a4abfc0942 -> javas 'old' Filereader has 
`              .onMalformedInput(CodingErrorAction.REPLACE)
                .onUnmappableCharacter(CodingErrorAction.REPLACE));`
which is good enough for this usecase but the nio api will fail by default on encoding errors

https://github.com/tg-freigmbh/untitled2

change the encoding of the c file to iso-88591 if git messes with the encoding

### Gradle version

9.3.0

9.2.0

### Build scan URL (optional)

_No response_

Linux jvm 21

ov7a:
Fixes #36399

Partially reverts:
* https://github.com/gradle/gradle/pull/35107

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/native/language-native/src/main/java/org/gradle/language/nativeplatform/internal/incremental/sourceparser/RegexBackedCSourceParser.java
+++ b/platforms/native/language-native/src/main/java/org/gradle/language/nativeplatform/internal/incremental/sourceparser/RegexBackedCSourceParser.java
@@ -28,6 +28,7 @@ import org.jspecify.annotations.Nullable;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileReader;
 import java.io.IOException;
 import java.io.Reader;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


tg-freigmbh:
Well the source code comment isn't always correct, the FileReader also uses UTF-8 (for java >= 21)  it just doesn't throw an exception :)

ov7a:
@tg-freigmbh Can you clarify?

```java
    public FileReader(File file) throws FileNotFoundException {
        super(new FileInputStream(file));
    }
 ```
 ->
 ```java
     / **
     * Creates an InputStreamReader that uses the
     * {@link Charset#defaultCharset() default charset}.
     *
     * @param  in   An InputStream
     *
     * @see Charset#defaultCharset()
     */
    public InputStreamReader(InputStream in) {
        super(in);
        sd = StreamDecoder.forInputStreamReader(in, this,
                Charset.defaultCharset()); // ## check lock object
    }
 ```

tg-freigmbh:
See https://openjdk.org/jeps/400
the 'default charset' is by default UTF-8 from 18 onwards.  It used to be that its the underlying operting systems charset...
so if the daemon is running 17 or less its the os charset, After that its only the os charset if someone uses Dfile.encoding=COMPAT

tg-freigmbh:
ok so system charset in this case= the charset of the jvm which is maybe the os charset... 
so the comment isn't  really wrong as long as you know that the system charsets refers to the default charset of the jvm and not to the encoding of the os.

ov7a:
@tg-freigmbh Ah, I see. I don't think it's an issue per se. In any case, this PR reverts things back to how they were to address the regression. I think the problem you mentioned should be handled in
* https://github.com/gradle/gradle/issues/2270

ov7a:
>  system charsets refers to the default charset 

Thanks, you're correct. I've updated the comment

Issue #36351, opened 2026-01-20
Expose nullable JavaBean properties from definitions in DCL


eskatos:
### Expected Behavior

Nullable JavaBean properties are exposed in DCL

```kotlin
interface SomeType {
  var something: String?
}
```

### Current Behavior (optional)

Such properties are currently ignored by DCL schema building, thus absent from the DSL.

### Context

AGP model as a few of these like `android.namespace`.

This allowance should only be applied to properties and function arguments, not to return types.

bamboo:
Do we really want to allow `null` in Declarative definitions?

In other words, do we expect users to be setting properties to `null` in DCL files?

bamboo:
> Do we really want to allow `null` in Declarative definitions?
> 
> In other words, do we expect users to be setting properties to `null` in DCL files?

I guess the point of the issue is to expose the properties for configuration with proper values.

eskatos:
Yes. DCL has no `null`. 
The goal is to expose nullable JavaBean properties as non-nullable DCL properties.

eskatos:
* Fixed in https://github.com/gradle/gradle/pull/36409 for 9.5
* Backported to 9.4 in https://github.com/gradle/gradle/pull/36440

bamboo:
> Yes. DCL has no `null`.

😮‍💨

PR #36409, opened 2026-01-22
Support nullable properties and parameters in DCL

h0tk3y:
Fixes:
* https://github.com/gradle/gradle/issues/36351

### DIFF
--- a/platforms/core-configuration/declarative-dsl-core/src/test/kotlin/org/gradle/internal/declarativedsl/schemaBuidler/NullableTypesTest.kt
+++ b/platforms/core-configuration/declarative-dsl-core/src/test/kotlin/org/gradle/internal/declarativedsl/schemaBuidler/NullableTypesTest.kt
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.declarativedsl.schemaBuidler
+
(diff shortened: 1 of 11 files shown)


h0tk3y:
@bot-gradle test RFN

h0tk3y:
@bot-gradle test RFN

platforms/core-configuration/declarative-dsl-core/src/test/kotlin/org/gradle/internal/declarativedsl/schemaBuidler/NullableTypesTest.kt
```diff
@@ -0,0 +1,141 @@
...
+        var nestedNullableProp: Nested?
+    }
+
+    interface HasNullableTypeArg {
+        @Suppress("unused")
+        var nullableProp: Box<String?>
+    }
+
+    interface HasNullableConfiguredType {
+        fun configure(action: Nested?.() -> Unit)
```
hfhbd, line 123: what about `fun foo(): Unit`? Does it fail or will it pass now? Same for `fun foo(bar: String): Unit`? And multiple parameters?

h0tk3y: It is currently not reported as an error, but in fact, does not make much sense from the Declarative modelling perspective.

Specifically, such functions are by default imported as value-producing, semantically "pure" ones, which means you cannot invoke them as a statement (just `foo()` on a line will be reported as an error), and you have to assign the returned value to a property or pass it as an argument to another function call. In turn, `Unit`-typed properties or function parameters themselves do not make much sense (although not reported as errors in the current implementation, just because Kotlin allows them).

For now, one can annotate a `Unit`-returning function with `@Adding`, meaning that it has a side effect of adding a new element to the model. 

But the side effect encapsulated in the function's implementation is opaque, and we want to move away from models that the tooling cannot reason about, so we are going to deprecate and remove support for `@Adding` functions, providing instead a more tooling-friendly and predictable way to add items to the model.

hfhbd: I agree with your all of your points, it was more a question if it is tested or will accidentally be supported with these changes.

h0tk3y: I created https://github.com/gradle/gradle/issues/36416 about `Unit` type usages.

PR #36430, opened 2026-01-23
Test CC effect of changing every documented Gradle property

bamboo:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGradlePropertiesFileIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGradlePropertiesFileIntegrationTest.groovy
@@ -16,6 +16,10 @@
 
 package org.gradle.internal.cc.impl
 
+import org.gradle.integtests.fixtures.AvailableJavaHomes
+import org.gradle.test.precondition.Requires
+import org.gradle.test.preconditions.IntegTestPreconditions
+
 class ConfigurationCacheGradlePropertiesFileIntegrationTest extends AbstractConfigurationCacheIntegrationTest {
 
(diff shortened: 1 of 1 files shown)


platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGradlePropertiesFileIntegrationTest.groovy
mlopatkin, line 1: 🎉 Nice harness

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGradlePropertiesFileIntegrationTest.groovy
```diff
@@ -268,4 +272,90 @@ class ConfigurationCacheGradlePropertiesFileIntegrationTest extends AbstractConf
...
+        configurationCacheRun 'help'
+
+        then:
+        if (invalidates) {
+            configurationCache.assertStateStored()
+        } else {
+            configurationCache.assertStateLoaded()
+        }
+
+        where:
```
mlopatkin, line 307: 🤔 We have a few more options that are/were documented, though some may be tested in other places

* `org.gradle.welcome`
* `org.gradle.dependency.verification`
* `org.gradle.configuration-cache.problems`
* `org.gradle.console.unicode` 
* `org.gradle.configuration-cache.integrity-check`
* `org.gradle.configuration-cache.parallel`
* `org.gradle.configuration-cache.read-only`

mlopatkin (approved):
LGTM! I recall a few more that are known to public, so maybe we should test them as well.

PR #36442, opened 2026-01-25
Use OS-specific data directories

### LINKED ISSUE #2805: Be a good citizen on Windows (don't pollute %USERPROFILE%).
On Windows, the default location for GRADLE_USER_HOME is `%USERPROFILE%/.gradle`.  This is not the appropriate location for cache data on Windows for a number of reasons:
1. `%USERPROFILE%` is meant to be an end-user facing directory.  Files should only go here if the user intentionally puts them there via a save dialog box.  It is not meant to be a dumping ground for every application's configuration files.
2. `%USERPROFILE%` roams across the network by default.  After building a single small project, this folder is already ~250MB.  Having this roam across a network on login-logout will kill performance and is really bad.
3. There are very well defined places for cache files on Windows and documentation on how to use them appropriately.

To save you the trouble of having to do the research, the correct location for a cache like `.gradle`  folder is `%LOCALAPPDATA%\gradle\gradle`.  This follows the recommendation of `%LOCALAPPDATA%` for large files that do not need roam across a network and `%LOCALAPPDATA%/<vendor>/<product>` for specific folder.  You could remove the intermediate `gradle` folder if it is believed that the `gradle` vendor will only ever release a single product named `gradle`.

### LINKED ISSUE #8262: Follow the Freedesktop XDG base directory standard (in Linux)
On Linux, Gradle currently creates a `$HOME/.gradle/` (or `$HOME/.gradle-kotlin-dsl/`) folder in the user's home directory. Most applications/programs don't create a folder in the user's home directory anymore, but instead use the `Freedesktop XDG` standard – found here: http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html.

Using this standard:
- `$XDG_CACHE_HOME` says "temporal/non-essential" files go to `$HOME/.cache/`
- `$XDG_CONFIG_HOME` says "configuration files" go to `$HOME/.config/`
- `$XDG_DATA_HOME` says "data files" go to `$HOME/.local/share/`

This is just a request to consider using this standard. It makes a lot of sense; it is easy for the user to understand, makes it easier to backup / locate / restore / manage / clear user's data, and creates less clutter in the user's home directory.

Thanks in advance...and thanks so much for everyone's hard work on this – it's amazing!

JFronny:
Fixes #2805
Fixes #8262

### Context

As mentioned in the attached issues: "%USERPROFILE% is meant to be an end-user facing directory".
This should also improve compatibility with systems for network-shared profiles or backups, though I would have to split out the cache directory to fully achieve this goal, which I deemed to be too big of a change for now.
The code is based on [some existing code in ResolverEventLogger](https://github.com/gradle/gradle/blob/a7b54eb3c74093c24a8ab5b28dfc151d5b7ea79e/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/ResolverEventLogger.kt#L94-L104), so I hope this approach is ok.

The reasoning for the behavior is as follows:
- if a user is already using gradle (indicated by `~/.gradle` existing), they probably don't want to lose their caches. As such, this directory will continue to be used.
- if a user is already using the OS-specific directory, they will probably want to keep using it
- if a user is new (indicated by neither existing), they probably want their data in the OS-appropriate directory. As such, the OS-specific directory is used.
- if both directories exist, I assume that the user has migrated to the OS-specific directory, but ran some older version of gradle which created `~/.gradle`. In this case, the OS-specific directory is used so that the user doesn't lose their cache.

I have used a build with these changes on my Linux machine to verify that they work as expected, which appears to be the case.
Due to my lack of a Windows or MacOS box, I have not been able to do the same for those systems, but since the code is based on an existing class, I am somewhat confident in it regardless.
Before this can be merged, I hope that someone with one of those operating systems can test the code there as well.

Ideally, the GUH resolution code would be shared between the wrapper and Gradle itself instead of being copied to both places, but I was unable to find a good place to move it in order to achieve that.

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/wrapper-shared/src/main/java/org/gradle/wrapper/GradleUserHomeLookup.java
+++ b/platforms/core-runtime/wrapper-shared/src/main/java/org/gradle/wrapper/GradleUserHomeLookup.java
@@ -18,23 +18,27 @@ package org.gradle.wrapper;
 
 import org.gradle.internal.os.OperatingSystem;
 
-import javax.annotation.Nullable;
 import java.io.File;
 
 public class GradleUserHomeLookup {
-    public static final File DEFAULT_GRADLE_USER_HOME = new File(System.getProperty("user.home") + "/.gradle");
+    public static File DEFAULT_GRADLE_USER_HOME = new File(System.getProperty("user.home") + "/.gradle");
     public static final String GRADLE_USER_HOME_PROPERTY_KEY = "gradle.user.home";
     public static final String GRADLE_USER_HOME_ENV_KEY = "GRADLE_USER_HOME";
 
+    private static File osDataCandidate;
     public static File gradleUserHome() {
         String gradleUserHome = System.getProperty(GRADLE_USER_HOME_PROPERTY_KEY);
         if (gradleUserHome == null) {
             gradleUserHome = System.getenv(GRADLE_USER_HOME_ENV_KEY);
         }
         if (gradleUserHome == null) {
-            File osDataDirectory = osDataDirectory();
-            if (osDataDirectory != null) {
-                File osDataCandidate = new File(osDataDirectory, "Gradle");
+            if (osDataCandidate == null) {
+                File osDataDirectory = osDataDirectory();
+                if (osDataDirectory != null) {
+                    osDataCandidate = new File(osDataDirectory, "Gradle");
+                }
+            }
+            if (osDataCandidate != null) {
                 if (osDataCandidate.isDirectory() || !DEFAULT_GRADLE_USER_HOME.exists()) {
                     gradleUserHome = osDataCandidate.getAbsolutePath();
                 }
@@ -46,7 +50,7 @@ public class GradleUserHomeLookup {
         return new File(gradleUserHome);
     }
 
-    public static @Nullable File osDataDirectory() {
+    public static File osDataDirectory() {
         OperatingSystem os = OperatingSystem.current();
         File home = new File(System.getProperty("user.home"));
         if (os.isMacOsX()) {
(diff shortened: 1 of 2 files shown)


ov7a:

***

This might be considered as a breaking change, maybe it would be postponed to Gradle 10 or there should be found a way to opt-in for that (ENV variable maybe? Given that it's an initialization phase, it might be non-trivial).

JFronny:
Regarding the lint fail: I could, of course, rename `DEFAULT_GRADLE_USER_HOME` to match the pattern, but that would cause a public field to change its name which I wanted to avoid, and I'm almost certain that the way I have implemented the test that needs it to be writable isn't ideal. If you have a better approach, I can modify the code to use that instead.
I could also try using reflection to mutate the field, which would make the API cleaner and avoid further divergence with the impl in gradle itself, but would be a bit hacky itself.

mlopatkin:
Thank you for the contribution!

Unfortunately, we're unable to accept it. If we're to change the layout, we'd like to do it in a single pass, and this solution seems to be half-way, compared to what is desired in the linked issues. We'd probably need a full design spec outlining the path forward, potential issues and mitigations for them.

For example, while most of the state in `GRADLE_USER_HOME` is indeed caches, there is at least `gradle.properties` configuration file that may benefit from roaming, and it shouldn't be stored in cache directories. There are also popular third-party plugins that started to use `~/.gradle` as their storage (though this is discouraged), and shuffling things require coordination to avoid breaking them.

I really like the backward compatibility ideas of checking for existing `.gradle` directory, it might be a good way to introduce the new location in a less-breaking fashion.

I'm sorry but we simply don't have enough bandwidth to deal with such an impactful change now, as it doesn't align with the current team priorities.

PR #36445, opened 2026-01-26
Pass CI env env down to subprocesses

blindpirate:
Because we publish many build scans in some tests (like smoke tests, Gradleception, etc), we need to pass down the `CI` env variable, otherwise they will show as local builds.

### DIFF
--- a/build-logic/jvm/src/main/kotlin/gradlebuild/propagated-env-variables.kt
+++ b/build-logic/jvm/src/main/kotlin/gradlebuild/propagated-env-variables.kt
@@ -108,15 +108,15 @@ val credentialsKeywords = listOf(
 )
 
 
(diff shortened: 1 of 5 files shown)


blindpirate:
@bot-gradle test this

build-logic/jvm/src/main/kotlin/gradlebuild/propagated-env-variables.kt
```diff
@@ -16,6 +16,7 @@
 
 package gradlebuild
 
+import com.gradle.scan.agent.serialization.scan.serializer.kryo.it
```
blindpirate:
```suggestion
```

blindpirate:
@bot-gradle squash

blindpirate:
@bot-gradle test this

blindpirate:
@bot-gradle squash and merge

PR #36456, opened 2026-01-26
Fix issue with breaking change to test result file names

### LINKED ISSUE #36379: Gradle 9.3.0 generate JUnit test result files with wrong name
### Current Behavior

When we run our tests we expect the JUnit test reports to start with the `TEST-` prefix. But Gradle 9.3.0 generate files with the prefix `__T-`, `__TE-`, `__TES-` and `__TEST-`.
The package name and class name seems to have impact on the filename. It feels like these are temporary files which are not renamed to their final names. 

I reproduced the issue here:
https://github.com/bdeneuter/gradle-junit-bug

I also committed the test result file with the wrong name:
https://github.com/bdeneuter/gradle-junit-bug/blob/main/build/test-results/test/__TEST-com.rand-AF7CJ0II31764.random_app.random_service.bussiness_logic.MyExternalBusinessEventFilterIntegrationTest.xml

### Expected Behavior

JUnit test result files with a correct name.

### Context (optional)

We noticed on our CI server (Jenkins) that our reporting around number of tests dropped

https://github.com/bdeneuter/gradle-junit-bug

### Gradle version

9.3.0

### Build scan URL (optional)

_No response_

_No response_

ghale:
This fixes two issues:
1) It reverts the change to add `__` to the beginning of truncated file names.
2) It adds logic to ensure that the test prefix (`TEST-`) is both preserved (i.e. not truncated) but also taken into account when truncating file names to a safe length.

Fixes #36379 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/SafeFileLocationUtilsTest.groovy
+++ b/platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/SafeFileLocationUtilsTest.groovy
@@ -149,29 +149,30 @@ class SafeFileLocationUtilsTest extends Specification {
     }
 
     private static final String PREFIX = 'FOO-'
+    private static final int MAX_SAFE_FILE_NAME_LENGTH_WITH_PREFIX_IN_BYTES = MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - Utf8.encodedLength(PREFIX)
 
     def "toSafeFileName hashes overly long paths and preserves a prefix when specified"() {
         expect:
         toSafeFileName(PREFIX, input, false) ==  output
         where:
         input                                                  | output
-        'A' * 256                                              | PREFIX + 'A' * (MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - Utf8.encodedLength(PREFIX)) + '-RH50HNS6NT02C'
-        ('A' * 253) + 'Θ'                                      | PREFIX + 'A' * (MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - Utf8.encodedLength(PREFIX)) + '-CDRIEFT37CF62'
+        'A' * 256                                              | PREFIX + 'A' * MAX_SAFE_FILE_NAME_LENGTH_WITH_PREFIX_IN_BYTES + '-RH50HNS6NT02C'
+        ('A' * 253) + 'Θ'                                      | PREFIX + 'A' * MAX_SAFE_FILE_NAME_LENGTH_WITH_PREFIX_IN_BYTES + '-CDRIEFT37CF62'
         // Hash should preserve extension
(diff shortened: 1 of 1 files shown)


ghale:
@bot-gradle test this

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/SafeFileLocationUtils.java
```diff
@@ -302,13 +322,14 @@ private static String shortenNameAndAddHash(byte[] rawName, String encoded) {
      * but only if the extension(s) can fit within {@code MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES} from the end.
      *
      * @param rawName the original name in UTF-8 bytes
+     * @param lengthOfPrefix the number of bytes to reserve for a fixed prefix before the name
      * @return the index of the first dot indicating the start of extension(s), or {@code -1} if none found
      */
-    private static int findStartOfExtension(byte[] rawName) {
+    private static int findStartOfExtension(byte[] rawName, int lengthOfPrefix) {
         // We look for the first dot that is within (MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - 1) from the end
```
octylFractal:
```suggestion
        // We look for the first dot that is within (MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - lengthOfPrefix - 1) from the end
```

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/SafeFileLocationUtils.java
```diff
@@ -302,13 +322,14 @@ private static String shortenNameAndAddHash(byte[] rawName, String encoded) {
...
         // This ensures that the extension can fit even after adding the hyphen and hash
```
octylFractal:
```suggestion
        // This ensures that the extension can fit even after adding the hyphen, hash, and prefix
```

platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/SafeFileLocationUtilsTest.groovy
```diff
@@ -130,22 +128,71 @@ class SafeFileLocationUtilsTest extends Specification {
...
+    }
+
+    private static final String PREFIX = 'FOO-'
+
+    def "toSafeFileName hashes overly long paths and preserves a prefix when specified"() {
+        expect:
+        toSafeFileName(PREFIX, input, false) ==  output
+        where:
+        input                                                  | output
+        'A' * 256                                              | PREFIX + 'A' * (MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - Utf8.encodedLength(PREFIX)) + '-RH50HNS6NT02C'
```
octylFractal: 💭 Pull out `MAX_SAFE_FILE_NAME_LENGTH_IN_BYTES - Utf8.encodedLength(PREFIX)` to a constant to shorten this?

ghale: Yeah, that's probably better.  Done.

Issue #34568, opened 2025-08-06
Enable `LifecycleAwareProject` wrappers in vintage


6hundreds:

alllex:
Fixed in https://github.com/gradle/gradle/pull/34706

PR #36459, opened 2026-01-26
Note on Project referential equality for 9.3.0

alllex:
When addressing https://github.com/gradle/gradle/issues/34568 , we started wrapping Project instances into delegate-like classes. While there is no public contract that `Project` instances must support referential equality, this behavior was observable practically from the dawn of Gradle.

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -171,6 +171,46 @@ plugins {
 
 === Potential breaking changes
 
+==== Referential equality is not guaranteed for `Project` instances
+
+Instances of `Project` type representing the same logical Gradle project do not provide any guarantee of being equal by reference (`==` in Java, `===` in Kotlin).
+However, historically it was possible to observe that such `Project` instances were exactly the same.
+
+In Gradle 9.3.0, `Project` instances (for the same logical project) **can be different** with respect to the referential equality,
+even if obtained within the same context, e.g., in the same build script.
+This change is necessary to facilitate future performance improvements of Gradle.
+The `Project.equals()` equality behavior remains unchanged.
+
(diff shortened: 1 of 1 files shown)


platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -171,6 +171,45 @@ plugins {
 
 === Potential breaking changes
 
+==== Referential equality is not guaranteed for `Project` instances
+
+Instances of `Project` type generally don't provide guarantees of being equal by reference (`==` in Java, `===` in Kotlin).
```
```suggestion
Instances of `Project` type do not provide any guarantee of being equal by reference (`==` in Java, `===` in Kotlin).
```

jvandort: Not sure I am a fan of the word 'generally' here. To me that implies that in special cases we may provide such guarantee. I'd rather remove 'generally' to assert that there is definitely no guarantee

bamboo:
```suggestion
Instances of `Project` type representing the same logical Gradle project generally don't provide guarantees of being equal by reference (`==` in Java, `===` in Kotlin).
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -171,6 +171,45 @@ plugins {
...
+However, historically it was possible to observe that `Project` instances are often exactly the same.
```
bamboo:
```suggestion
However, historically it was possible to observe that `Project` instances representing the same logical Gradle project were often exactly the same.
```

alllex:
Thanks for the feedback! I applied all suggestions

cobexer:
@bot-gradle cherrypick to release9x

PR #36488, opened 2026-01-27
Fix no-daemon tests spawning a single-use daemon process

mlopatkin:
We missed that because it only happens on Java 24+, and we were running no-daemon tests on Java 21 until recently. Additionally, there were no specific tests covering no-daemon executor not spawning the single-use process.

This commit doesn't clean up all the places that add `--enable-native-access` and equivalents, but it eliminates at least one duplicate.

Having duplicate switches is breaking process compatibility checks but this commit doesn't address that either.

Fixes gradle/gradle-private#5031.


mlopatkin:
@bot-gradle test RfN

mlopatkin:
@bot-gradle test RfR

mlopatkin:
RfR failure seem unrelated and flaky, rerun passed.

PR #31031, opened 2024-10-25
Offload build operation trace writing to a separate thread

jvandort:
Avoid blocking the threads notifying of build operation events. Instad, offload the writing to another thread.

This both ensures the notifying threads don't need to perform IO but also that they do not need to synchronize with each other, as previously we needed to acquire a lock to write to the file

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #36492, opened 2026-01-27
Avoid contention when capturing build operation traces

jvandort:
https://github.com/gradle/gradle/pull/31031 attempted to avoid contention between threads by offloading the work of serializing build operation traces to a single thread. Previously threads were blocked on each other while one thread performed IO. The linked PR did avoid threads blocking on IO, but they still need to block on each other to _submit_ work to the queue.

This commit switches out the async trace writer to use a custom loop and thread park/unpark to avoid as much contention as possible on the producing threads. 

Before:
image

Using LinkedTransferQueue:
image

Before:
image

Using LinkedTransferQueue:
image

Using new Thread + custom loop with ConcurrentLinkedQueue:
image

Using purpose-built Multi-producer single-consumer processor: 
image

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
+++ b/platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
@@ -16,8 +16,6 @@
 
 package org.gradle.internal.concurrent;
 
-
-import org.gradle.internal.UncheckedException;
 import org.jctools.queues.MessagePassingQueue;
 import org.jctools.queues.MpscUnboundedArrayQueue;
 import org.jspecify.annotations.Nullable;
@@ -35,7 +33,7 @@ import java.util.function.Consumer;
  * the values to process.
  * <p>
  * Any failure in the processor will terminate the worker thread. Any subsequent
- * interactions with the processor will rethrow the failure.
+ * interactions with the processor will immediately throw an exception.
  * <p>
  * If the worker thread is interrupted during execution, any remaining unprocessed
  * work is dropped.
@@ -46,7 +44,7 @@ public class MultiProducerSingleConsumerProcessor<T> {
      * The number of items to process in a single iteration of the worker loop,
      * before checking failure status or thread interruption.
      */
-    public static final int BATCH_SIZE = 1024;
+    static final int BATCH_SIZE = 1024;
 
     /**
      * Processes submitted values on a separate thread.
@@ -97,7 +95,7 @@ public class MultiProducerSingleConsumerProcessor<T> {
             }
         };
 
-        this.queue = new MpscUnboundedArrayQueue<>(1024);
+        this.queue = new MpscUnboundedArrayQueue<>(BATCH_SIZE);
         this.worker = new Thread(this::workerLoop, workerThreadName);
         this.worker.setDaemon(true);
     }
@@ -122,7 +120,7 @@ public class MultiProducerSingleConsumerProcessor<T> {
      */
     public void submit(T value) {
         if (failure != null) {
-            throw UncheckedException.throwAsUncheckedException(failure);
+            throw new IllegalStateException("Cannot submit values after processor has failed.", failure);
         }
 
         if (!running) {
@@ -168,7 +166,7 @@ public class MultiProducerSingleConsumerProcessor<T> {
         }
 
         if (failure != null) {
-            throw UncheckedException.throwAsUncheckedException(failure);
+            throw new IllegalStateException("Failure occurred during processor execution.", failure);
         }
     }
 
@@ -180,8 +178,8 @@ public class MultiProducerSingleConsumerProcessor<T> {
                     break;
                 }
 
-                // Invoke the processor with at most 1024 new values.
-                int processed = queue.drain(processor, 1024);
+                // Invoke the processor with at most BATCH_SIZE new values.
+                int processed = queue.drain(processor, BATCH_SIZE);
                 if (failure != null) {
                     break;
                 }
--- a/platforms/core-runtime/concurrent/src/test/groovy/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessorTest.groovy
+++ b/platforms/core-runtime/concurrent/src/test/groovy/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessorTest.groovy
@@ -184,7 +184,8 @@ class MultiProducerSingleConsumerProcessorTest extends ConcurrentSpec {
 
         then:
         def e = waitForFailure { processor.submit(2) }
-        e.is(failureInstance)
+        e.message == "Cannot submit values after processor has failed."
+        e.cause.is(failureInstance)
 
         cleanup:
         stopQuietly(processor)
@@ -230,7 +231,8 @@ class MultiProducerSingleConsumerProcessorTest extends ConcurrentSpec {
 
         then:
         def e = thrown(RuntimeException)
-        e.is(failureInstance)
+        e.message == "Failure occurred during processor execution."
+        e.cause.is(failureInstance)
 
         cleanup:
         stopQuietly(processor)


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

alllex:
@bot-gradle test this

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
```diff
@@ -0,0 +1,182 @@
...
+
+        this.queue = new MpscUnboundedArrayQueue<>(1024);
+        this.worker = new Thread(this::workerLoop, workerThreadName);
+        this.worker.setDaemon(true);
+    }
+
+    /**
+     * Start the handler. This must be called before any values may be submitted.
+     */
+    public void start() {
```
alllex, line 106: 💅 Add a guard for the double-start

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    private void workerLoop() {
+        try {
+            while (running || !queue.isEmpty()) {
+                if (Thread.interrupted()) {
+                    throw new InterruptedException();
+                }
+
+                int processed = queue.drain(processor, 1024);
```
alllex: ❓ What is the failure mode contract that we want to have here?

The javadoc of `drain` and `Consumer.accept` require a contract that the consumer never throws. In the only current usage we have, we break this assumption, since the `TraceWriter.write` rethrows IO exceptions.

Given that we wrap the MPSC queue as an implementation detail, I think we should also not accept the consumer interface directly, and instead accept a generic consumer, which we will wrap internally, ensuring the above-mentioned contract is never broken.

We can either have some generic failure logging here, or we can accept one from the user of this class. But this way we have modeled the contract, instead of assuming it's honored.

On the other hand, if it's guaranteed that the `drain` will rethrow the consumer exception, then our logic here ensures the broken queue is never used anyway. If that's the route we are going, I'd see prefer to have a generic `Consumer` accepted in the constructor, because then we actually *want* the consumer to throw, which contradicts the javadoc on the currently accepted interface.

Also, if we are going with the latter, I think it would be better to be explicit about it in the comments in the code, since we are deliberately going against the contract of the data structure.

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    private void workerLoop() {
+        try {
+            while (running || !queue.isEmpty()) {
+                if (Thread.interrupted()) {
+                    throw new InterruptedException();
+                }
+
```
alllex, line 180: 💅 I'd add a comment here that this call synchronously invokes the consumer on all the drained elements. At least, this wasn't obvious to me

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
```diff
@@ -0,0 +1,182 @@
...
+                if (processed > 0) {
+                    continue;
+                }
+
+                // Signal that we are going to sleep.
+                awake.set(false);
+
+                // It is possible that a new value was submitted after we drained the
+                // queue but before we set the awake flag to false. This would cause
+                // the submitting thread to not attempt to wake up the worker thread.
```
alllex:
```suggestion
                // Double-check: It is possible that a new value was submitted after we drained the
                // queue but before we set the awake flag to false. This would cause
                // the submitting thread to not attempt to wake up the worker thread.
                // By checking queue.isEmpty() again, we catch values added during the race window.
```

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
```diff
@@ -0,0 +1,182 @@
...
+                if (queue.isEmpty() && running) {
+                    LockSupport.park();
+                }
+
+                // Most of the time, this will be set to true by the producer thread before unparking
+                // the worker. However, `park` may return even if unpark was never requested.
+                // Therefore, We must always set the `awake` flag to true unconditionally.
```
alllex:
```suggestion
                // Most of the time, this will be set to true by the producer thread before unparking
                // the worker. However, `park` may return even if unpark was never requested
                // (spurious wakeup). Therefore, we must always set the `awake` flag to true
                // unconditionally to maintain the invariant that awake=true when the worker is active.
```

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
```diff
@@ -0,0 +1,182 @@
...
+     * Submit a value to be processed. This method returns immediately, and
+     * the value will be passed to the processor on a separate thread.
+     */
+    public void submit(T value) {
+        if (failure != null) {
+            throw UncheckedException.throwAsUncheckedException(failure);
+        }
+
+        if (!running) {
+            throw new IllegalStateException("Cannot submit values after processor has been stopped.");
```
alllex: 💅 Check is `worker.state is NEW` to throw a more meaningful exception, when submitting before start. Will help diagnosing the problem faster, if the caller forgets to call start.

bamboo:
Why did you choose `LinkedTransferQueue` and not `ConcurrentLinkedQueue` initially?

jvandort:
> 
Why did you choose LinkedTransferQueue and not ConcurrentLinkedQueue initially?

`ConcurrentLinkedQueue` is not a `BlockingQueue`, only a `Queue`. The `ThreadPoolExecutor` constructor requires a `BlockingQueue`

bamboo:
> `ConcurrentLinkedQueue` is not a `BlockingQueue`, only a `Queue`. The `ThreadPoolExecutor` constructor requires a `BlockingQueue`

I see. So that was before the non-blocking queue design with `MessagePassingQueue`.

And how much better is `MessagePassingQueue` compared to `CLQ`, or is there some functionality missing?

bamboo:
What's the _custom loop_ mentioned in the description?

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
jvandort: The failure mode contract is that if the consumer ever throws an exception, the processor is stopped and cleaned up. Then, any subsequent interaction with the processor re-throws the first thrown exception so that the exception is not lost. 

I updated the code to accept a generic `Consumer` and catch any exceptions it throws, so that the processor we pass to `drain` does not throw any exceptions.

jvandort: This failure contract was documented in the class-level javadoc as 
```

 * <p>
 * Any failure in the processor will terminate the worker thread. Any subsequent
 * calls to {@link #submit(Object)} will throw an exception.
 ```

I've updated the wording to

```
 * Any failure in the processor will terminate the worker thread. Any subsequent
 * interactions with the processor will rethrow the failure.
```

**Comment by jvandort** [comment]
The custom loop is an implementation that uses a `new Thread` and a `ConcurrentLinkedQueue` instead of a `ThreadPoolExecutor` and a `LinkedTransferQueue`. See the second commit for this implementation. 

So the "custom loop" flamegraph in part showcases `ConcurrentLinkedQueue` performance vs MessagePassingQueue (though with a slightly different threading implementation, in particular, the custom loop solution unconditionally calls `unpark` even when not necessary)

File: platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java [hunk 7]
```diff
@@ -0,0 +1,218 @@
...
+ * If the worker thread is interrupted during execution, any remaining unprocessed
+ * work is dropped.
+ */
+public class MultiProducerSingleConsumerProcessor<T> {
+
+    /**
+     * The number of items to process in a single iteration of the worker loop,
+     * before checking failure status or thread interruption.
+     */
+    public static final int BATCH_SIZE = 1024;
```
alllex: ❓ Are we using it in the implementation? Does it have to be public?

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+        this.worker.start();
+    }
+
+    /**
+     * Submit a value to be processed. This method returns immediately, and
+     * the value will be passed to the processor on a separate thread.
+     */
+    public void submit(T value) {
+        if (failure != null) {
+            throw UncheckedException.throwAsUncheckedException(failure);
```
alllex: 💅  This will rethrow an exception that was caught elsewhere without providing context, which can make it harder to understand why `submit` fails. How about wrapping this with a clear message along the lines of "Cannot submit value, because processor was stopped due to a failure".

It does not seem important for us here that we rethrow _exactly_ the exception that we've caught. We already could wrap the underlying failure into an `Unchecked[IO]Exception`.

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * Processes values submitted from many producer threads on a single consumer thread.
+ * This processor is specifically optimized to avoid as much overhead as possible
+ * on the producer-side. This processor is best suited for scenarios like tracing,
+ * where it is beneficial to avoid blocking or contention on the thread producing
+ * the values to process.
+ * <p>
+ * Any failure in the processor will terminate the worker thread. Any subsequent
+ * interactions with the processor will rethrow the failure.
```
alllex:
```suggestion
 * interactions with the processor will immediately throw.
```

The current implementation might now rethrow the exact failure that we received. It's better we don't imply this in the javadoc. I also suggest to always wrap the failure with meaningful context in another comment.

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
jvandort: Made package private (used in tests) and updated the implementation to use it.

PR #36493, opened 2026-01-27
Release locks when waiting for immutable work to complete

jvandort:
Previously, we would block the current thread waiting for immutable work complete, if we were also trying to execute that work while it was already executing. Now, when we block, we drop the worker lease so another thread can leverage the CPU while we wait

Before, notice we have wide "Build accessors classpath" operations with nothing happening in them:
image

After, notice we release the worker lease in these situations:
image

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test apt

PR #36494, opened 2026-01-27
Reproduce and fix #36428

### LINKED ISSUE #36428: Breaking change in 9.3.0 regarding cross-project dependency manipulation
### Current Behavior

- have a build with two projects
- put to the root project
  ```kotlin
  subprojects {
      configurations.configureEach {
          allDependencies.configureEach {
          }
      }
  }
  ```
- run `./gradlew dependencies`

Ignoring that this is of course bad-practice cross-project configuration,
this works up to including 9.2.1 but fails with 9.3.0:
```plain
An exception occurred applying plugin request [id: 'org.gradle.java-library']
> Failed to apply plugin class 'org.gradle.api.plugins.JavaPlugin'.
   > Could not create domain object 'java' (JvmSoftwareComponentInternal) in SoftwareComponent container
      > Could not create domain object 'test' (JvmTestSuite) in TestSuite container
         > Could not create an instance of type org.gradle.api.plugins.jvm.internal.DefaultJvmTestSuite.
            > The value for property 'dependencies.compileOnly' property 'dependencies' is final and cannot be changed any further.
```

If that is an intentional new limitation, it should probably go through the usual deprecation cycle.

### Expected Behavior

Not breaking in a minor update

### Gradle version

9.3.0

9.2.1

big-guy:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/DependencyCollector.java
+++ b/subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/DependencyCollector.java
@@ -35,6 +35,15 @@ import java.util.Set;
  * A {@code DependencyCollector} is used as part of a dependencies block in the DSL. A collector implements
  * a single dependency scope and exposes the declared dependencies on {@link #getDependencies()}.
  *
+ * <p>
+ * To use a {@code DependencyCollector} as a source of dependencies, wire it to the appropriate {@link org.gradle.api.artifacts.Configuration}
+ * with {@link org.gradle.api.artifacts.Configuration#fromDependencyCollector(DependencyCollector)}.
+ * Dependencies will be queried lazily.
+ * </p>
+ * <p>
+ *     Note: No mutations will be allowed after dependencies are read from a {@code DependencyCollector}.
+ * </p>
+ *
  * @apiNote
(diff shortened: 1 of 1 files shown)


platforms/jvm/plugins-jvm-test-suite/src/main/java/org/gradle/api/plugins/jvm/internal/DefaultJvmTestSuite.java
```diff
@@ -101,11 +96,23 @@ public DefaultJvmTestSuite(String name, SourceSetContainer sourceSets, Configura
             toolchain.getParameters().getJupiterVersion().convention(JUnitJupiterTestToolchain.DEFAULT_VERSION);
         }
 
+        // Add test framework dependencies to the dependency collectors BEFORE wiring them to configurations.
```
hfhbd, line 99: Could this explanation also be added to DependencyCollector docs? I had no idea DependencyCollector uses `finalizeValueOnRead` (it is in the sources though)

big-guy: done, thanks for the suggestion

PR #36499, opened 2026-01-28
Clean up perf test split dir

blindpirate:
In performance test, there is a step setting up `performance-test-splits` directory, but it never gets cleaned up. Discussed in https://gradle.slack.com/archives/CBSJ2GUTV/p1769510709157649


cobexer (approved):

cobexer:
@blindpirate PTAL at the test failure

PR #33627, opened 2025-05-27
Version catalog adoption

ljacomet:
For the moment, this covers only production dependencies, not build dependencies.
It is also not complete yet.

PR #36500, opened 2026-01-28
Split version catalog into production, provided, test and build deps

ov7a:
Follow-up to
* https://github.com/gradle/gradle/pull/33627

TODO (will be done in a follow-up PR):
- populate build dependencies

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic-settings/version-catalogs/src/main/kotlin/gradlebuild.version-catalogs.settings.gradle.kts
+++ b/build-logic-settings/version-catalogs/src/main/kotlin/gradlebuild.version-catalogs.settings.gradle.kts
@@ -27,6 +27,9 @@ dependencyResolutionManagement {
         create("libs") {
             from(files(basePath.file("distribution.versions.toml")))
         }
+        create("providedLibs") {
+            from(files(basePath.file("provided.versions.toml")))
+        }
         create("testLibs") {
             from(files(basePath.file("test.versions.toml")))
         }
--- a/gradle/dependency-management/distribution.versions.toml
+++ b/gradle/dependency-management/distribution.versions.toml
@@ -12,7 +12,6 @@ commonsCompress = "1.26.1!!"
 commonsHttpclient = "4.5.14!!"
 commonsIo = "2.15.1!!"
 commonsLang = "3.20.0!!"
-eclipseSisuPlexus = "0.3.5!!"
 fastutil = "8.5.2!!"
 gcs = "v1-rev20220705-1.32.1!!"
 googleApiClient = "1.34.0!!" # our GCS version requires 1.34.0
@@ -39,9 +38,6 @@ jsch = "0.2.16!!"
 jsoup = "1.15.3!!"
 jspecify = "1.0.0!!"
 jsr305 = "3.0.2!!"
-junit = "4.13.2!!" # compileOnly
-junit5 = "5.12.2!!" # compileOnly
-junitPlatform = "1.12.2!!" # compileOnly
 kotlin = "2.3.0!!"
 kotlinxSerialization = "1.9.0!!"
 kryo = "2.24.0!!"
@@ -49,7 +45,6 @@ minlog = "1.2!!"
 nativePlatform = "0.22-milestone-29!!"
 objenesis = "2.6!!" # transitive dependency of kryo
 plexusCipher = "2.0!!"
-plexusClassworlds = "2.7.0!!"
 plexusInterpolation = "1.26!!"
 plexusSecDispatcher = "2.0!!"
 plexusUtils = "3.5.1!!"
@@ -57,9 +52,7 @@ plist = "1.27!!"
 slf4j = "2.0.17!!"
 snakeyaml = "2.0!!"
 sshd = "2.15.0!!"
-testng = "6.3.1!!" # later versions break test cross-version test filtering
 tomlj = "1.0.0!!"
-zinc = "1.12.0!!"
 
 [libraries]
 ant = { group = "org.apache.ant", name = "ant", version.ref = "ant" }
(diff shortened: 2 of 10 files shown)


ov7a:
@bot-gradle test RfN

ov7a:
Build scan comparison with master
https://ge.gradle.org/c/ilrgxjnyu5trm/xdfp2syfortow/dependencies

build-logic-settings/version-catalogs/src/main/kotlin/gradlebuild.version-catalogs.settings.gradle.kts
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+dependencyResolutionManagement {
+    val root = if (rootProject.name == "build-logic") {
+        layout.rootDirectory.dir("..")
+    } else {
+        layout.rootDirectory
+    }
+    val basePath = root.dir("gradle").dir("dependency-management")
```
eskatos, line 25: 💅 Can be simplified
```suggestion
    val basePath = root.dir("gradle/dependency-management")
```

platforms/jvm/testing-jvm-infrastructure/build.gradle.kts
```diff
@@ -71,21 +71,21 @@ dependencies {
         }
         because("We test assertion errors coming from OpenTest4J")
     }
-    testImplementation(libs.junit) {
+    testImplementation(providedLibs.junit) {
```
eskatos, line 74: ❔ `junit` is present in both `providedLibs` and `testLibs`. In this file we reference both. It seems arbitrary. How do we distinguish which one we use when?

Using `providedLibs` for `compileOnly` dependencies kind of make sense. Using `testLibs` for `testImplementation` or `testRuntimeOnly` would also make sense to me.

build-logic-settings/version-catalogs/src/main/kotlin/gradlebuild.version-catalogs.settings.gradle.kts
ov7a: I was not sure if "/" is fully platform independent in this context

platforms/jvm/testing-jvm-infrastructure/build.gradle.kts
ov7a: We have two usages of junit:
1. Our own test infrastructure (most places) - it's testLibs
2. Junit tests integration in java plugins - similarly to testing, it's a provided dependency there

Previously, a single version was used everywhere, but I don't think we should use the same version; thus, I split them.

gradle/dependency-management/build.versions.toml
```diff
@@ -0,0 +1,5 @@
+# some versions are in shared-versions.properties
```
cobexer, line 1: ❓ Will this break dependabot integration?

gradle/dependency-management/distribution.versions.toml
```diff
@@ -1,156 +1,77 @@
+# some versions are in shared-versions.properties
```
cobexer, line 1: ❓ Will this break dependabot integration?

gradle/dependency-management/provided.versions.toml
```diff
@@ -0,0 +1,28 @@
+# some versions are in shared-versions.properties
```
cobexer, line 1: ❓ Will this break dependabot integration?

gradle/dependency-management/shared-versions.properties
```diff
@@ -0,0 +1,21 @@
+#
```
cobexer, line 1: ❌ Turn this file into a proper version catalog for dependabot support

gradle/dependency-management/test.versions.toml
```diff
@@ -0,0 +1,110 @@
+# some versions are in shared-versions.properties
```
cobexer, line 1: ❓ Will this break dependabot integration?

packaging/distributions-dependencies/build.gradle.kts
```diff
@@ -17,191 +17,17 @@ description = "Provides a platform dependency to align all distribution versions
...
-        api(libs.sshdCore)
-        api(libs.sshdOsgi)              { because("It contains sshd-core and sshd-common classes") }
-        api(libs.sshdScp)
-        api(libs.sshdSftp)
-        api(libs.testcontainers)
-        api(libs.testcontainersSpock)
-        api(libs.xerces)
-        api(libs.xmlunit)
+        val distributionDependencies = versionCatalogs.named("libs")
+        distributionDependencies.libraryAliases.forEach { alias ->
```
cobexer, line 24: 🎉

cobexer (changes requested):

gradle/dependency-management/build.versions.toml
ov7a: IDK, but I don't think there's a good solution to this. I considered the following variants:
1. eagerly evaluating primary version catalog - not-cc compatible + uses internals
2. Duplicate versions - easy to miss the full update and have divergent versions. 
3. Provide stubs - it felt as noise and was confusing
4. current approach

And dependabot integration would still have potential issues with:
1. the custom catalog paths
2. dependency verification

So, while I'm not happy with the current solution as well, I don't see a better solution for sharing/aligning versions between different catalogs.

gradle/dependency-management/shared-versions.properties
ov7a: It's hard to achieve. It can be loaded with gradle API (see the comment above - it would break CC and internals should be used). It's hard to be loaded as toml (we need to have extra dependency on toml parser in build logic), and writing a zero-dependency solution to load toml just for that feels an overkill.

PR #36503, opened 2026-01-28
TeamCity: disable IP in CI runs

cobexer:


cobexer:
Verified with an RfR run: https://ge.gradle.org/scans?search.names=gitCommitId&search.startTimeMax=1770245999999&search.startTimeMin=1769986800000&search.tags=CI&search.timeZoneId=Europe%2FVienna&search.values=18a22901e4d60b45c787f157f7a78c2000917bd0

blindpirate (approved):

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractGradleceptionSmokeTest.groovy
```diff
@@ -74,11 +75,15 @@ abstract class AbstractGradleceptionSmokeTest extends AbstractSmokeTest {
         result = runner.buildAndFail()
     }
 
+    SmokeTestGradleRunner runner(String... tasks) {
+        List<String> args = GRADLE_BUILD_TEST_ARGS + (tasks as List<String>);
```
alllex, line 79: ❓ We need to double-check if this has an effect of disabling IP in smoke tests that enabled it in the gradle.properties, since the CLI args take precedence. If that's the case, we are reducing test coverage. I'll take a look shortly.

cobexer: Are there other tests that require IP to be enabled?

I considered this:
https://github.com/gradle/gradle/blob/ddb039842531637d7dd40048605f614e84fc6a70/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractGradleBuildIsolatedProjectsSmokeTest.groovy#L34-L45

alllex: ✅ It's all good. All children of `AbstractGradleceptionSmokeTest` use CLI flags instead of gradle.properties.

We use the properties file for the IDE Sync tests. But there we don't use the Runner infrastructure:
https://github.com/gradle/gradle/blob/d3f1b336289d4c9cb74875caf81925858d6e2c82/testing/smoke-ide-test/src/smokeIdeTest/groovy/org/gradle/ide/sync/IsolatedProjectsGradleceptionSyncTest.groovy#L61

.teamcity/src/main/kotlin/projects/CheckProject.kt
```diff
@@ -26,6 +26,9 @@ class CheckProject(
             StatisticsBasedPerformanceTestBucketProvider(model, File("performance-test-durations.json"), File("performance-tests-ci.json"))
 
         params {
+            if (model.branch.isMaster || model.branch.isExperimental) {
```
alllex, line 29: ❔ Why is it a conditional append of the parameter, instead of unconditional?
Can you summarize, what this condition means in practical terms for the Gradle devs?

cobexer: This means that the switch does not propagate to the release branch and only works on `Master` / `Xperimental` pipelines.

This simply means reduced risk of this affecting release activities.

alllex: Doesn't it mean that once we enable IP in gradle.properties, the release branch won't be safeguarding against disabling IP with the TC parameter?

I think we might need an `else` branch here with an explicit `=false`

cobexer: It will then still inherit https://github.com/gradle/gradle/pull/36503/changes#diff-ad90eed40315b6b5e5567ef3c1e5238ebf5c0e4362939829093db3652bdca9c5R35

https://github.com/gradle/gradle/blob/d181bd60118e14081a15cf856ce1fab0cfcb9ec3/.teamcity/src/main/kotlin/projects/GradleBuildToolRootProject.kt#L35

alllex: I see. Then we should be careful with removing that line, as the TODO there suggests

PR #36520, opened 2026-01-29
Update outgoingVariants reports to hide attrbuteless variants unless `--all` is enabled

tresat:
As these variants can't be selected, and shouldn't be used to resolve, hide them by default.

Add a new indicator to the report when it is run with `--all`.

Clarify the definition of a variant in the docs, ensuring readers understand that attributes should be added and are required for variant-aware selection.

Begun by AI, finalized by humans.

Flag attribute-less configurations with a new indicator on this and resolvableConfigurations reports


lkasso (approved):
Thanks for the good docs.

tresat:
@bot-gradle test this

PR #36526, opened 2026-01-29
Allow project feature apply actions to be declared unsafe

ghale:
This allows project feature bindings to declare that they have an unsafe apply action.  

We remove the `getProject()` and `getProjectLayout()` methods from `ProjectFeatureApplicationContext` and only provide access to an `ObjectFactory`.  With the object factory, you can instantiate an interface that describes the services you will need in the feature apply action.

Unsafe apply actions can access services outside of a limited set of "safe" services.  For instance, they can inject a `ProjectLayout` or even a `Project` object in order to access legacy plugin logic.  

Unsafe apply actions are declared by calling the `DeclaredProjectFeatureBindingBuilder.withUnsafeApplyAction()` method at binding declaration time.  If this method is not called, then the apply action is assumed to be safe.

Safe apply actions can only access the following features:
```
        TaskRegistrar.class
        ProjectFeatureLayout.class
        ConfigurationRegistrar.class
        ObjectFactory.class
        ProviderFactory.class
        DependencyFactory.class
```

We may add others if we find a use case for them.

Note that of the services above, three are new context-specific services:
```
        TaskRegistrar.class
        ProjectFeatureLayout.class
        ConfigurationRegistrar.class
```

These are safe forms of existing services.  For instance, `TaskRegistrar` allows you to register a task, but does not allow you to eagerly create tasks, look up other tasks, or replace tasks.  For now, these services work similar to the existing services (`TaskContainer`, `ProjectLayout` and `ConfigurationRegistrar`).  However, we intend to make them context-specific, such that they can automatically name things in a feature-binding specific context (i.e. to help manage collisions in task naming).  We'll add this capability in future work.  

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/internal/PluginUseServices.java
+++ b/platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/internal/PluginUseServices.java
@@ -17,8 +17,6 @@
 package org.gradle.plugin.internal;
 
 import com.google.common.collect.ImmutableSet;
-import org.gradle.api.artifacts.ConfigurationContainer;
-import org.gradle.api.file.ProjectLayout;
 import org.gradle.api.internal.BuildDefinition;
 import org.gradle.api.internal.artifacts.DependencyManagementServices;
 import org.gradle.api.internal.artifacts.DependencyResolutionServices;
@@ -26,14 +24,11 @@ import org.gradle.api.internal.file.FileCollectionFactory;
 import org.gradle.api.internal.file.FileResolver;
 import org.gradle.api.internal.initialization.ScriptClassPathResolver;
 import org.gradle.api.internal.initialization.StandaloneDomainObjectContext;
-import org.gradle.api.internal.model.ObjectFactoryFactory;
 import org.gradle.api.internal.plugins.PluginInspector;
-import org.gradle.api.internal.plugins.PluginManagerInternal;
 import org.gradle.api.internal.project.ProjectInternal;
 import org.gradle.api.internal.tasks.properties.InspectionScheme;
 import org.gradle.api.internal.tasks.properties.InspectionSchemeFactory;
 import org.gradle.api.problems.internal.InternalProblems;
-import org.gradle.api.tasks.TaskContainer;
 import org.gradle.initialization.ClassLoaderScopeRegistry;
 import org.gradle.internal.Factory;
 import org.gradle.internal.build.BuildIncluder;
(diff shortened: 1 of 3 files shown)


ghale:
@bot-gradle test this

subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/DependencyFactory.java
```diff
@@ -93,6 +94,27 @@ public interface DependencyFactory {
...
+    /**
+     * Create a {@link ProjectDependency} from a {@link Project} path.
+     *
+     * @param projectPath the project
+     * @return the new dependency
+     *
+     * @since 9.5.0
+     */
+    @Incubating
+    ProjectDependency createProjectDependency(String projectPath);
```
eskatos, line 106: ❓ Is this change an intended part of this PR? why?

eskatos: I see, this is to enable creating project dependencies without having access to the `Project` object.

eskatos (approved):

subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/DependencyFactory.java
ghale: Yeah, I added a change to have definitions and build models instantiated with the same safety as the apply action. Otherwise, an apply action could be declared safe, but the build model could, say, inject and expose the `Project` object and so you could do a bunch of unsafe stuff anyways.  But, this `Dependencies` object is our recommended way of exposing `DependencyCollector` functions, and `Dependencies` exposes and uses the `Project` object.

So this is there to push the necessary `Project` object access down to `DependencyFactory` (which is a service) and `Dependencies` can still be used without access to `Project` (i.e. you'll now get an `UnknownServiceException` if you call `Dependencies.getProject()` on a definition or build model object that implements it, but all of the default methods should still just work).

We should probably revisit whether requiring `Dependencies` is still the strategy we want to take.  It might be, but I think it's worth discussing again just to reconfirm.

.idea/gradle.xml
```diff
@@ -110,6 +110,7 @@
             <option value="$PROJECT_DIR$/platforms/core-configuration/declarative-dsl-api" />
             <option value="$PROJECT_DIR$/platforms/core-configuration/declarative-dsl-core" />
             <option value="$PROJECT_DIR$/platforms/core-configuration/declarative-dsl-evaluator" />
+            <option value="$PROJECT_DIR$/platforms/core-configuration/declarative-dsl-internal-utils" />
```
eskatos: ❌ This looks like an unwanted change. Maybe caused by that directory still being present in your worktree?

ghale: Thanks for catching this.  Fixed.

hfhbd:
How will you wire tasks as a feature author? Will it be save to expose a TaskProvider via the BuildModel (eg KotlinBuildModel) and use it in a SafeAction inside of FooFeature?
Or do you require correctly wired FileSystemLocation Providers (that would be good to enforce providers/good practice, IMHO)?

hfhbd:
And could you create a new issue for the upcoming/planned features for the new services, like tasks naming and fixed output locations.

ghale:
> How will you wire tasks as a feature author? Will it be save to expose a TaskProvider via the BuildModel (eg KotlinBuildModel) and use it in a SafeAction inside of FooFeature?

@hfhbd I think there are cases where you might want to expose a `TaskProvider` via your build model.  For instance, to wire a finalizer one way or the other.  However, I think it would be preferable to expose things in the build model more conceptually (e.g. expose a set of "generated sources" as a directory provider backed by a source generator task rather than exposing the task itself).  But I don't think it would be "wrong" to expose a `TaskProvider` in a build model if it made sense to do so.  

So, yes, we would consider that safe because you, as the feature author, have chosen to expose it.  What we primarily want to prevent is _arbitrary_ access to things where other feature authors can reach and fiddle with things you never intended to expose or support.

ghale:
> could you create a new issue for the upcoming/planned features for the new services, like tasks naming and fixed output locations.

@hfhbd I've created [this issue](https://github.com/gradle/gradle/issues/36619) around making these services more context-aware.  Just a warning, though, that this might be tightly coupled to some other work that we've been discussing around declaring inter-feature relationships, so there might be some continued design work and implementation that gets queued up ahead of this (in other words, it might be a little while before we get started on this).

PR #36530, opened 2026-01-30
Offload build operation event handling to separate thread

jvandort:
We handled these serially and during configuration we produce a lot of events, slowing things down. Now, we offload the events to a separate thread and handle them async

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test apt

jvandort:
Now that we have "nested" async actions, where one async action thread can submit to another async action thread, the shutdown process gets complicated. If some parent async thread submits to some child async thread, we cannot shutdown the child until we know the parent has drained their queue, otherwise the parent may attempt to add more values to the child after the child is closed

PR #36531, opened 2026-01-30
Remove unused cache

jvandort:
This cache was causing thread contention when performing configuration in parallel. We determined the cache was never hit by running all tests and throwing an exception if we saw a cache hit. We never observed an exception. This makes sense, as this cache caches the result of project accessor generation, which should only happen once per per project per build.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorsClassPath.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorsClassPath.kt
@@ -91,8 +91,11 @@ class ProjectAccessorsClassPathGenerator @Inject internal constructor(
 ) {
 
     fun projectAccessorsClassPath(scriptTarget: ExtensionAware, classPath: ClassPath): AccessorsClassPath {
-        val classLoaderScope = classLoaderScopeOf(scriptTarget) ?: return AccessorsClassPath.empty
-        val configuredProjectSchemaOf = configuredProjectSchemaOf(scriptTarget, classLoaderScope) ?: return AccessorsClassPath.empty
+        val classLoaderScope = classLoaderScopeOf(scriptTarget)
+            ?: return AccessorsClassPath.empty
+        val configuredProjectSchemaOf = configuredProjectSchemaOf(scriptTarget, classLoaderScope)
+            ?: return AccessorsClassPath.empty
+
         val work = GenerateProjectAccessors(
             scriptTarget,
             configuredProjectSchemaOf,


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

bamboo:
:question: IIRC the cache was added because Kotlin DSL model builders also needed the information. 

I see that [buildScriptModelFor](https://github.com/gradle/gradle/blob/b8a7b229c2ec54ce68851ba844eae32d829e29e5/platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt#L313) and [projectScriptModelBuilder](https://github.com/gradle/gradle/blob/7895fa7f0c18d370e5cb7c8f3f0c8d4dc7003c42/platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt#L248) still call the changed method. Are those no longer used as well or are we simply missing tests?

jvandort:
I see no reference to this model in IJ community, though not sure if their Gradle integration code lives elsewhere: https://github.com/search?q=repo%3AJetBrains%2Fintellij-community%20org.gradle.kotlin.dsl.tooling.models.KotlinBuildScriptModel&type=code

Though it does look like this other model is still queried:
https://github.com/search?q=repo%3AJetBrains%2Fintellij-community+org.gradle.tooling.model.kotlin.dsl.KotlinDslScriptsModel&type=code

jvandort:
In any case, I don't really have the time/motivation to pull this thread any further than provide a simple fix. I think we have two options that don't involve dedicating much more time here:
1. Delete the cache. Rely on the execution engine to handle caching (the underlying work is performed by the execution engine, and perfetto traces already show evidence of multiple projects blocking on the same underlying work performed by the EE)
2. Change the cache to something more suitable for long-running tasks (by wrapping the value in a `FutureTask`)

bamboo:
> 1. Delete the cache. Rely on the execution engine to handle caching

Yes, I think that's the right thing. My comment was more in the direction of removing more code if possible.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorsClassPath.kt
```diff
@@ -91,44 +90,25 @@ class ProjectAccessorsClassPathGenerator @Inject internal constructor(
...
-        val classLoaderScope = classLoaderScopeOf(scriptTarget)
-        if (classLoaderScope == null) {
-            return AccessorsClassPath.empty
-        }
-        return classPathCache.computeIfAbsent(classLoaderScope) {
-            buildAccessorsClassPathFor(classLoaderScope, scriptTarget, classPath)
-                ?: AccessorsClassPath.empty
-        }
+        val classLoaderScope = classLoaderScopeOf(scriptTarget) ?: return AccessorsClassPath.empty
+        val configuredProjectSchemaOf = configuredProjectSchemaOf(scriptTarget, classLoaderScope) ?: return AccessorsClassPath.empty
```
bamboo: 💅 Keep the `?: return ...` in its own line for clarity.

jvandort:
Even if we determined the IDE was not querying some of these models they would still technically need to go through a deprecation cycle.

PR #36545, opened 2026-01-30
Use larger blocking server backlog

jvandort:
Before we would previously only open a limited number of concurrent connections to the blocking server Now, we open 20 concurrent connections, which can overwhelm the backlog and cause connections to be refused on windows. Now, we increase the backlog to avoid failures during testing

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle help

jvandort:
@bot-gradle test this without pts on windows

Issue #22749, opened 2022-11-16
Disallow changes to parameters from BuildServiceRegistration


lacasseio:
As part of [our expert support for Gradle](https://nokee.dev/services/consulting.html), we were faced with a great question: Should you be allowed to modify build service parameters after its registration? We provided a hypothesis to our client and confirmed our suggestion with the Gradle team. Let's suppose we have the following build service registration:

```
interface Parameters extends BuildServiceParameters {
	ListProperty<String> getValue()
}

abstract class MyService implements BuildService<Parameters> {}

def serviceProvider = gradle.sharedServices.registerIfAbsent('service', MyService) {
	parameters {
		value.add('first')
		value.add('second')
	}
}
```

### Expected Behavior
Reaching after the fact to the service registration to modify the [exposed parameters](https://docs.gradle.org/current/javadoc/org/gradle/api/services/BuildServiceRegistration.html#getParameters--) should be disallowed (aka throw an exception upon mutating any parameter's properties), e.g.:

```
gradle.sharedServices.registrations.getByName('service').parameters.value.add('third')
```

### Current Behavior
Changing the parameters from the service registration is always allowed and has unexpected behaviour depending on the state of the service itself (see context).

### Context
There are several reasons why we believe parameters mutation after registration should be disallowed:

- Gradle generally uses the pattern where a getter should access a read-only version of the subject, and an `Action` taking API allows modification of the subject in question. Gradle has some instances where this pattern is somewhat enforced. The details and gotcha of that pattern are not the focus here. Using that pattern as a general guideline, [`BuildServiceRegistration`](https://docs.gradle.org/current/javadoc/org/gradle/api/services/BuildServiceRegistration.html) only provides a getter to the parameters and no `Action` taking API like its counterpart [`BuildServiceSpec`](https://docs.gradle.org/current/javadoc/org/gradle/api/services/BuildServiceSpec.html). This would indicate that users should not reach to a `BuildServiceRegistration` to modify the parameters further.
- When we realize a build service, the parameters go through an isolation process (primarily to enable the configuration cache feature). This isolation rightfully disconnects the `BuildServiceParameters` from the `BuildServiceRegistration` and the `BuildService`. From then on, any changes to the registration parameters are no longer considered by the `BuildService`. This exposes the following confusing behaviour:
```
def serviceRegistration = gradle.sharedServices.registrations.getByName('service')   // <1>

serviceRegistration.parameters.value.add('third')                                    // <2>

println '===> before service realization:'
[...]

Issue #12247, opened 2020-02-16
Support Provider<T> for org.gradle.process.JavaForkOptions#systemProperty value


vlsi:
### Expected Behavior

```kotlin
tasks.test {
    ...
    val libraryDir = developmentBinary.linkFile
        .map { it.asFile.parentFile.absolutePath }
    systemProperty("java.library.path", libraryDir)
}
```

It should lazily compute the property value only in case the task is considered for execution.

### Current Behavior

Currently, Gradle accepts `Object` value, however it calls `#toString()` to convert the value to string.
Of course, there's a workaround to supply an anonymous class that overrides `toString`

Here's a workaround:

```kotlin

/**
 * Gradle does not support [Provider] for [JavaForkOptions.systemProperty], so
 * we pass an object that overrides [Any.toString].
 */
fun Provider<String>.overrideToString() = object {
    override fun toString() = orNull ?: ""
}

...
tasks.test {
    dependsOn(developmentBinary.linkTask)
    val libraryDir = developmentBinary.linkFile
        .map { it.asFile.parentFile.absolutePath }
    systemProperty("java.library.path", libraryDir.overrideToString())
}
```

### Context

I'm building JNI-related project, and I want to pass the location of native library via system property, however it forces me to eagerly compute the location (e.g. even in case the task is not used).

I guess unwrapping `Providers` would not break backward compatibility much, as `Provider#toString` does not typically make much sense.

Gradle 6.1.1 if that matters.

vlsi:
This is still needed. Would you please reopen the issue?

vlsi:
Alternative workaround is to use
```kotlin
jvmArgumentProviders.add(...CommandLineArgumentProvider...)
```

davidburstrom:
I came across the same missing functionality today, on Gradle 7.1.

Vampire:
The work-around with `CommandLineArgumentProvider` is working fine and you can then also declare the input and output properties and implicit task dependencies properly.

But how do you make the `toString` work-around usable?
If I try to do it like mentioned, the build errors out with `Input property 'systemProperties' with value '{foo=bar}' cannot be serialized.`.

vlsi:
>If I try to do it like mentioned, the build errors out with Input property 'systemProperties' with value '{foo=bar}' cannot be serialized..

I guess that depends on Gradle version you have, and the workaround seemed to work in 6.1.1.

[...]

PR #36550, opened 2026-01-31
Make ci-reporting plugin IP compatible

blindpirate:
Previously, the `ci-reporting` plugin registered a global `TestFilesCleanupService` to perform the following tasks:

1. Artifact Collection: It collected report files and directories (e.g., test reports, japicmpReports, performance reports, and TD trace JSONs) from all subprojects, archived them, and moved them to the root project's build directory for TeamCity artifact publication.

2. Cleanup Verification: It verified that all tests successfully cleared their temporary directories. If any uncleared directories were discovered and `reportOnly` was not enabled, the service would throw an exception.

3. Final Cleanup: It performed a final sweep to remove all remaining temporary files.

The original plugin was not IP compatible. This PR refactors to make it compatible: a global service is registered and all projects store data to it.

### DIFF
--- a/build-logic/jvm/src/main/kotlin/gradlebuild.unittest-and-compile.gradle.kts
+++ b/build-logic/jvm/src/main/kotlin/gradlebuild.unittest-and-compile.gradle.kts
@@ -20,7 +20,6 @@ import com.gradle.develocity.agent.gradle.test.DevelocityTestConfiguration
 import gradlebuild.basics.BuildEnvironment
 import gradlebuild.basics.FlakyTestStrategy
 import gradlebuild.basics.accessors.kotlinMainSourceSet
-import gradlebuild.basics.buildRunningOnCi
 import gradlebuild.basics.flakyTestStrategy
 import gradlebuild.basics.maxParallelForks
 import gradlebuild.basics.maxTestDistributionLocalExecutors


build-logic/buildquality/src/main/kotlin/gradlebuild.configure-ci-artifacts.gradle.kts
```diff
@@ -17,31 +17,65 @@ import gradlebuild.basics.BuildEnvironment
...
-    else -> emptyList()
+    project.tasks.withType<Test>().forEach {
+        addReports(it.path, it.traceJson())
+        addReports(it.path, it.genericHtmlReports())
+    }
+
+    testFileCleanUpExtension.taskPathToReports.putAll(taskPathToReports.mapValues { it.value.toSet() })
 }
+
+fun ValidatePlugins.validatePluginsReports() = listOf(outputFile.get().asFile)
```
alllex: 🤔 It is suspicious that we are eagerly evaluating these properties. It means, we won't get the right value, if they are changed during futher project configuration.

Though, this is probably best addressed separately, to not overcomplicate this PR.

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.ci-reporting.gradle.kts
```diff
@@ -14,32 +14,31 @@
...
-    rootProject.plugins.apply(TestFilesCleanupRootPlugin::class.java)
-    val globalExtension = rootProject.extensions.getByType<TestFilesCleanupBuildServiceRootExtension>()
-
-    val projectState = objects.newInstance(TestFilesCleanupProjectState::class.java)
-
-    globalExtension.projectStates.put(path, projectState)
-    projectState.projectBuildDir = layout.buildDirectory
-    projectState.projectPath = path
-    projectState.reportOnly = testFilesCleanup.reportOnly
+    val testFilesCleanupService = project.gradle.sharedServices.registerIfAbsent("testFilesCleanupBuildService-" + project.name, TestFilesCleanupService::class.java) {
```
alllex: ❌ I don't think we should create a service per project. This creates many services, each of which is registered as a listener for all task completions, but practically discards most events.

The previous implementation was better, though it had a flaw of talking to the root project extension, which made it incompatible with IP.

What we should do instead, while keeping the implementation relatively close to what was before, is to have a single build service, to which all projects talk. Instead of updating an extension on the root project, each subproject should call a method on that _shared_ build service, submitting its data, such as the path and the build dir.

The shared build service must protect from potentially concurrent calls, since projects will be configuring in parallel. We have a bug currently, that we don't respect max parallelism with IP (https://github.com/gradle/gradle/issues/29273). In practice, this means the method on the shared service should synchronize on a lock to ensure thread-safety.

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.ci-reporting.gradle.kts
```diff
@@ -14,32 +14,31 @@
...
+        parameters.rootBuildDir.set(project.layout.settingsDirectory.dir("build"))
```
alllex: 💭 Once we switch to a single shared build service, instead of a parameter, we can have a method on it that allows submitting the root project build directory. Since the service should be registered by the root project, it can directly share it's build directory property.

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.ci-reporting.gradle.kts
```diff
@@ -14,32 +14,31 @@
...
+        parameters.projectPath.set(project.path)
+        parameters.projectBuildDir.set(project.layout.buildDirectory)
+        parameters.taskPathToReports.set(testFilesCleanupExtension.taskPathToReports)
+        parameters.testPathToBinaryResultsDirs.set(testFilesCleanupExtension.testPathToBinaryResultsDirs)
+        parameters.reportOnly.set(testFilesCleanupExtension.reportOnly)
+    }
+    project.gradle.serviceOf<BuildEventsListenerRegistry>().onTaskCompletion(testFilesCleanupService)
```
alllex: It should be enough that only the root project registers the service as the task completion listener.

build-logic/buildquality/src/main/kotlin/gradlebuild.arch-test.gradle.kts
```diff
@@ -67,7 +67,10 @@ testing {
                         }
                         testClassesDirs += sharedArchTestClasses.filter { it.isDirectory }
                         classpath += sourceSets["main"].output.classesDirs
-                        systemProperty("package.cycle.exclude.patterns", packageCyclesExtension.excludePatterns.get().joinToString(","))
+                        val excludePatterns = packageCyclesExtension.excludePatterns
+                        doFirst {
+                            systemProperty("package.cycle.exclude.patterns", excludePatterns.get().joinToString(","))
```
alllex, line 73: 💭 Since we are touching this, I'd add a link to https://github.com/gradle/gradle/issues/12247

alllex (changes requested):
Left some comments

There are also `TestFilesCleanupServiceTest` unit tests that probably made assumptions about the service implementation, and some of them no longer hold.

build-logic/buildquality/src/main/kotlin/gradlebuild.arch-test.gradle.kts
blindpirate: I actually expect this change will be merged first, so I can simply fetch and rebase.

build-logic/buildquality/src/main/kotlin/gradlebuild.configure-ci-artifacts.gradle.kts
blindpirate: Now all references are lazy.

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.ci-reporting.gradle.kts
blindpirate: Thanks for the suggestion, now I have changed it to global plugin and service.

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.ci-reporting.gradle.kts
blindpirate: This is done.

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.ci-reporting.gradle.kts
blindpirate: Done.

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.ci-reporting.gradle.kts
blindpirate: Weirdly, if I apply this plugin to root project, I got this error: https://ge.gradle.org/s/mhe43jfimbugc

For now I just keep the old behavior, it's only applied to non-root project.

build-logic/buildquality/src/test/kotlin/gradlebuild/testcleanup/TestFilesCleanupServiceTest.kt
```diff
@@ -185,19 +199,19 @@ class TestFilesCleanupServiceTest {
 
     @Test
     fun `fail build if leftover file found and test passes`() {
-        val result = run(":successful-test-with-leftover:test", "--no-watch-fs").buildAndFail()
+        val result = run(":successful-test-with-leftover:test").buildAndFail()
         assertEquals(TaskOutcome.SUCCESS, result.task(":successful-test-with-leftover:test")!!.outcome)
 
-        assertEquals(1, StringUtils.countMatches(result.output, "Found non-empty test files dir"))
-        assertEquals(1, StringUtils.countMatches(result.output, "Failed to stop service 'testFilesCleanupBuildService'"))
+        assertEquals(2, StringUtils.countMatches(result.output, "Found non-empty test files dir"))
```
blindpirate, line 205: This change was caused by extra `--stacktrace` so the error appears twice.

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.ci-reporting.gradle.kts
```diff
@@ -14,32 +14,57 @@
...
+    val testFilesCleanupService = project.gradle.sharedServices.registerIfAbsent("testFilesCleanupBuildService", TestFilesCleanupService::class.java) {
+        require(project.path == ":") { "Must be applied to root project first, now: ${project.path}" }
+        parameters.taskPathToReports.set(mutableMapOf<String, List<File>>())
+        parameters.testPathToBinaryResultsDirs.set(mutableMapOf<String, File>())
+        parameters.rootBuildDir.set(project.layout.buildDirectory)
+    }
+    if (project.path == ":") {
+        project.gradle.serviceOf<BuildEventsListenerRegistry>().onTaskCompletion(testFilesCleanupService)
+    }
+    testFilesCleanupService.get().parameters.projectStates.put(project.path, TestFilesCleanupProjectState(project.path, project.layout.buildDirectory.get().asFile, testFilesCleanup.reportOnly))
```
alllex: ❌ Build service parameters must not be mutable.

It's unfortunate that this is still allowed, but it's a known issue:
- https://github.com/gradle/gradle/issues/22749

Access to the state that these parameters hold is not synchronized, creating races when projects configuring in parallel execute this code.

As I mentioned in the previous comments, we should move all the state management into the service itself, and expose _methods_ on it. The implementation of these methods can then provide proper synchronization between accesses

As far as I can tell, this should basically have no parameters

build-logic/buildquality/src/main/kotlin/gradlebuild.configure-ci-artifacts.gradle.kts
alllex: I hope that doesn't create more problem for us, given that this PR is already tricky 😓

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.ci-reporting.gradle.kts
blindpirate: Nice catch! I've removed the mutable parameters and made the service thread safe. PTAL.

blindpirate:
@bot-gradle test this

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild/testcleanup/TestFilesCleanupService.kt
```diff
@@ -38,54 +40,71 @@ import java.io.IOException
...
+        testPathToBinaryResultsDir.put(testTaskPath, binaryResultsDir)
+    }
+
+    fun addTaskReports(taskPath: String, reports: Iterable<Any>) {
+        safeAdd(taskPathToReports, taskPath, reports)
+    }
+
+    private fun <K, V> safeAdd(map: ConcurrentHashMap<K, MutableSet<V>>, key: K, values: Iterable<V>) {
+        val valuesInMap = map.computeIfAbsent(key, { k -> Collections.synchronizedSet<V>(HashSet<V>()) })
+        values.forEach { valuesInMap.add(it) }
```
cobexer: ❓ Could this be `valuesInMap.addAll(values)`?

build-logic/buildquality/src/main/kotlin/gradlebuild.arch-test.gradle.kts
cobexer: Please add the link

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild/testcleanup/TestFilesCleanupService.kt
blindpirate: It's actually `Iterable`, but we actually only use List and Set, so we could change it to `Collection`.

blindpirate:
@bot-gradle test this

build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild/testcleanup/TestFilesCleanupService.kt
cobexer: Let's go for Collection

blindpirate:
@bot-gradle test this

cobexer (approved):

alllex:
@bot-gradle test this

PR #36567, opened 2026-02-02
Fix DCL dependency collectors appearances in the schema

h0tk3y:
Fix dependency collectors in DCL appearing as nested models or properties

The DCL schema should not have configuring functions produced from the
`DependencyCollector` getters. They should not be visible as DCL properties
(even read-only ones) either.

<details>
<summary>Before</summary>

image

image

</details>

<details>
<summary>After</summary>

image

image

</details>


h0tk3y:
@bot-gradle test RFN

eskatos (approved):

Issue #17813, opened 2021-07-22
More docs about configuration caching, esp. with build services


mathjeff:
I've found some documentation about configuration caching at https://medium.com/androiddevelopers/configuration-caching-deep-dive-bcb304698070 and https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution which make it easy enough for me to migrate simpler apis to use Gradle's configuration caching.

However, it's hard for me to know how to migrate more complicated APIs to use Gradle's configuration caching, partially because the list of migrations provided doesn't appear to be completely exhaustive and partially because I don't know enough about what configuration caching does to be able to figure out an appropriate migration on my own as easily as I would like.

In particular, the reason I was interested in this was because I was trying to determine the appropriate way to compute some information involving multiple projects and to reuse that information across builds ( https://r.android.com/1771472/3  ). It appears that a BuildService allows saving and reusing some information across builds but I didn't notice this in the documentation.

It would be nice if the configuration caching documentation could either describe more about its interaction with BuildServices or describe more about how configuration caching is implemented, or both.

Thanks!

mathjeff:
My current understanding of the way that configuration caching works is:

1. When configuration caching is enabled, most user-provided configuration logic doesn't execute; instead, the results of that logic get cached by Gradle and the results get reused.
2. Accessing a Project instance at task execution time is disallowed when configuration caching is enabled
3. Build configuration logic can declare that the entire configuration phase must rerun when certain inputs change by declaring those inputs via syntax like `providers.systemProperty('someproperty').forUseAtConfigurationTime()`.
4. Build configuration logic can declare that certain portions of that logic should be rerun by moving that logic into a BuildService, which Gradle will know to recreate when needed. Any parameters given to a BuildService will be serialized by Gradle into the configuration cache and reloaded in subsequent builds (assuming nothing invalidated the cache such as mentioned in #2).
5. If a task makes use of a provider to a BuildService, Gradle will recreate that BuildService whenever a new build begins. If a task makes use of the result of a function call to a BuildService, I think Gradle attempts to cache the result.

bamboo:
This is now part of the Stable Configuration Cache milestone.

mathjeff:
Thanks!

PR #36569, opened 2026-02-02
Fix CC incompatible tests in TaskParametersIntegrationTest

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/TaskParametersIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/TaskParametersIntegrationTest.groovy
@@ -24,15 +24,13 @@ import org.gradle.api.provider.Provider
 import org.gradle.api.provider.SetProperty
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.integtests.fixtures.TestBuildCache
-import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
 import org.gradle.integtests.fixtures.executer.GradleContextualExecuter
 import org.gradle.internal.Actions
 import org.gradle.internal.reflect.validation.ValidationMessageChecker
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.IntegTestPreconditions
+import org.gradle.util.internal.ToBeImplemented
 import spock.lang.Issue
-
-import static org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache.Skip.INVESTIGATE
 
 class TaskParametersIntegrationTest extends AbstractIntegrationSpec implements ValidationMessageChecker {
 
@@ -41,8 +39,9 @@ class TaskParametersIntegrationTest extends AbstractIntegrationSpec implements V
             task foo {
                 inputs.property "a", "hello"
                 inputs.property "b", new Foo()
-                outputs.file "foo.txt"
-                doLast { file("foo.txt") << "" }
+                def outputFile = file("foo.txt")
+                outputs.file(outputFile)
+                doLast { outputFile << "" }
             }
 
             class Foo {
@@ -58,13 +57,13 @@ class TaskParametersIntegrationTest extends AbstractIntegrationSpec implements V
         failure.assertHasCause("Cannot fingerprint input property 'b': value 'xxx' cannot be serialized.")
     }
 
-    @ToBeFixedForConfigurationCache(skip = INVESTIGATE)
     def "deals gracefully with not serializable contents of GStrings"() {
         buildFile << """
             task foo {
                 inputs.property "a", "hello \${new Foo()}"
-                outputs.file "foo.txt"
-                doLast { file("foo.txt") << "" }
+                def outputFile = file("foo.txt")
+                outputs.file(outputFile)
+                doLast { outputFile << "" }
             }
 
             class Foo {
@@ -241,15 +240,15 @@ class TaskParametersIntegrationTest extends AbstractIntegrationSpec implements V
     def "task depends on other task whose outputs are its inputs"() {
         buildFile << """
             task a {
-                outputs.file 'a.txt'
                 def outputFile = file('a.txt')
+                outputs.file(outputFile)
                 doLast {
                     outputFile << "Data"
                 }
             }
 
             task b {
-                inputs.files tasks.a.outputs.files
+                inputs.files(tasks.a.outputs.files)
             }
         """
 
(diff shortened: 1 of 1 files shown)


subprojects/core/src/integTest/groovy/org/gradle/api/tasks/TaskParametersIntegrationTest.groovy
mlopatkin: :x: We're undercounting the accesses when running with configuration cache, because the build service is re-created with the configuration cache. It needs to be a registered empty implementation of `OperationCompletionListener`, as described here: https://docs.gradle.org/current/userguide/build_services.html#operation_listener to circumvent that.

We have https://github.com/gradle/gradle/issues/17813 to document that. Apparently this trick is only word-of-mouth today 😱

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/TaskParametersIntegrationTest.groovy
```diff
@@ -1016,26 +1022,28 @@ task someTask(type: SomeTask) {
         succeeds("printCounts")
         then:
         executedAndNotSkipped(':myTask')
-        outputContains("outputFileCount = 2, inputFileCount = 3, inputValueCount = 1, nestedInputCount = 3, nestedInputValueCount = 1")
+        def expectedInitialCount = GradleContextualExecuter.isConfigCache() ? 2 : 3
```
mlopatkin: ❓ What does `initial` mean in this context?

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/TaskParametersIntegrationTest.groovy
```diff
@@ -24,25 +24,23 @@ import org.gradle.api.provider.Provider
...
 
     def "reports which properties are not serializable"() {
         buildFile << """
             task foo {
                 inputs.property "a", "hello"
                 inputs.property "b", new Foo()
-                outputs.file "foo.txt"
-                doLast { file("foo.txt") << "" }
+                outputs.file("foo.txt")
+                def outputFile = file("foo.txt")
```
mlopatkin: 💅 I'd use the variable for the output:

```suggestion
                def outputFile = file("foo.txt")
                outputs.file(outputFile)

```

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/TaskParametersIntegrationTest.groovy
```diff
@@ -58,13 +56,13 @@ class TaskParametersIntegrationTest extends AbstractIntegrationSpec implements V
...
 
-    @ToBeFixedForConfigurationCache(skip = INVESTIGATE)
     def "deals gracefully with not serializable contents of GStrings"() {
         buildFile << """
             task foo {
                 inputs.property "a", "hello \${new Foo()}"
-                outputs.file "foo.txt"
-                doLast { file("foo.txt") << "" }
+                outputs.file("foo.txt")
+                def outputFile = file("foo.txt")
```
mlopatkin: 💅 
```suggestion
                def outputFile = file("foo.txt")
                outputs.file(outputFile)
```

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/TaskParametersIntegrationTest.groovy
```diff
@@ -627,8 +629,13 @@ task someTask(type: SomeTask) {
         expect:
         fails "test"
         failure.assertHasDescription("A problem was found with the configuration of task ':test' (type 'DefaultTask').")
-        failureDescriptionContains(missingValueMessage { property('output') })
-
+        def expectedMessage
+        if (GradleContextualExecuter.isConfigCache()) {
```
mlopatkin, line 636: 🤔 This discrepancy feels like a minor bug. I'm not sure if the value is actually "configurable" though. Maybe it is worth filing an issue about the diff between CC and vintage and linking it in`@ToBeImplemented` here and in the test above.

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/TaskParametersIntegrationTest.groovy
```diff
@@ -1016,26 +1022,28 @@ task someTask(type: SomeTask) {
...
         executedAndNotSkipped(':myTask')
-        outputContains("outputFileCount = 2, inputFileCount = 3, inputValueCount = 1, nestedInputCount = 3, nestedInputValueCount = 1")
+        outputContains("outputFileCount = 2, inputFileCount = ${expectedInitialCount}, inputValueCount = 1, nestedInputCount = ${expectedNestedInputCount}, nestedInputValueCount = 1")
 
         when:
         succeeds("printCounts")
         then:
         skipped(':myTask')
-        outputContains("outputFileCount = 1, inputFileCount = 2, inputValueCount = 1, nestedInputCount = 3, nestedInputValueCount = 1")
+        outputContains("outputFileCount = 1, inputFileCount = ${expectedInitialCount - 1}, inputValueCount = 1, nestedInputCount = ${expectedNestedInputCount}, nestedInputValueCount = 1")
```
mlopatkin: 🤔 I'm not sure about computing the value here and below - does the formula express the property we're testing or just an accidental match? I'd rather use a number without linking it to `initial`.

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/TaskParametersIntegrationTest.groovy
ov7a: https://github.com/gradle/gradle/issues/36576

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/TaskParametersIntegrationTest.groovy
ov7a: Oh, this causes even more discrepancies in counts between CC and Vintage. Should I create an issue about that?

mlopatkin: > Oh, this causes even more discrepancies in counts between CC and Vintage. Should I create an issue about that?

This is more an optimization opportunity at this point rather than an issue - we don't have a strong guarantee of how many passes we do, as long as it isn't too many. I looked and at least one of the 4 `@Nested` passes with CC looks superfluous.

ov7a: https://github.com/gradle/gradle/issues/36577

PR #36570, opened 2026-02-02
Make WarTaskIntegrationTest CC-compatible

tresat:
Updates test to be CC-compatible.

### DIFF
--- a/platforms/jvm/war/src/integTest/groovy/org/gradle/api/tasks/bundling/WarTaskIntegrationTest.groovy
+++ b/platforms/jvm/war/src/integTest/groovy/org/gradle/api/tasks/bundling/WarTaskIntegrationTest.groovy
@@ -17,7 +17,6 @@
 package org.gradle.api.tasks.bundling
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
-import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
 import org.gradle.integtests.fixtures.archives.TestFileSystemSensitiveArchives
 import org.gradle.test.fixtures.archive.JarTestFixture
 import spock.lang.Issue
@@ -284,7 +283,6 @@ class WarTaskIntegrationTest extends AbstractIntegrationSpec {
     }
 
     @Issue("GRADLE-3522")
-    @ToBeFixedForConfigurationCache(because = "early dependency resolution")
     def "war task doesn't trigger dependency resolution early"() {
         when:
         buildFile << """
@@ -293,8 +291,11 @@ configurations {
 }
 
 task assertUnresolved {
+    def confProvider = provider { configurations.conf.state.toString() }
+    inputs.property("confState", confProvider)
+
     doLast {
-        assert configurations.conf.state == Configuration.State.UNRESOLVED
+        assert confProvider.get() == Configuration.State.UNRESOLVED.toString()
     }
 }
 


platforms/jvm/war/src/integTest/groovy/org/gradle/api/tasks/bundling/WarTaskIntegrationTest.groovy
```diff
@@ -293,8 +291,11 @@ configurations {
 }
 
 task assertUnresolved {
+    def confProvider = provider { configurations.conf.state.toString() }
+    inputs.property("confState", confProvider)
+
     doLast {
-        assert configurations.conf.state == Configuration.State.UNRESOLVED
+        assert confProvider.get() == Configuration.State.UNRESOLVED.toString()
```
big-guy, line 298: I think there's a deeper thing missing here.  Changing the test in this way breaks what it was trying to do. IIUC, configurations.conf could be resolved after `confState` has been observed and this test will still pass.

A few ideas:
1. Could we add an internal flag that prohibits dependency resolution "at the wrong time" and enable that here?
2. Could we add a callback to configurations.conf that asserts that task execution has started?
3. Could we inspect build operations?

mlopatkin: > I think there's a deeper thing missing here. Changing the test in this way breaks what it was trying to do. IIUC, configurations.conf could be resolved after `confState` has been observed and this test will still pass.

Actually, it is vice versa. With CC there is no set guarantee of which task is going to be serialized first, so Gradle can store the `war` task first, resolving the configuration, and then store the provider, obtaining its value (which will likely be `RESOLVED` at this point). Of course, today we store the tasks in (roughly) the execution order, so it doesn't fail.

>1. Could we add an internal flag that prohibits dependency resolution "at the wrong time" and enable that here?

Thinking about it, it makes sense - CC can be allowed to resolve while serializing configurations.

>2. Could we add a callback to configurations.conf that asserts that task execution has started?

This would be CC-incompatible.

>3. Could we inspect build operations?

Perhaps, but today there is no difference between CC-triggered resolution and user-code triggered undeclared resolution, if the latter is something like `provider { conf.files }`

tresat: I've read through this a few times now, and still don't understand.

What is the test trying to do?  Isn't it checking that before `war` runs, `assertUnresolved` has ran and checked the current resolution state of `configurations.conf` during task execution?

tresat: I think I understand a little better now.  Serializing providers to the CC captures their value at serialization time, doesn't it?

Original: guaranteed to assert that conf is unresolved at execution time before war runs.

New: confState is observed earlier (or at serialization time), this may or may not be before the war task has caused resolution to happen during CC serialization, so conf might be resolved by the war task afterward and the test would still pass (false positive).

tresat: This is the bit I don't understand now:

> > Could we add an internal flag that prohibits dependency resolution "at the wrong time" and enable that here?

> 
> Thinking about it, it makes sense - CC can be allowed to resolve while serializing configurations.

mlopatkin: >New: confState is observed earlier (or at serialization time), this may or may not be before the war task has caused resolution to happen during CC serialization, so conf might be resolved by the war task afterward and the test would still pass (false positive).

Without CC it would still be the same, as `provider {}` is lazy and will be computed at the same point in time. With CC we're entering somewhat shaky grounds and implicitly assume that:
1. The `confProvider` is serialized before `War.classpath`
2. Nothing resolves `War.classpath` before serializing it but after serializing `confProvider`

Failing (1) would be a false negative (test will fail), but the probability of us changing the serialization in that way is low. So from the CC standpoint, this is probably good enough.

Failing (2) would be a false positive (test will pass), but there are not so many places where the plugin code may trigger the resolution. It would also likely cause the non-CC executor to fail. I'm deferring to code owners to figure out if this is crucial for the test.

mlopatkin: > This is the bit I don't understand now:
> 
> > > Could we add an internal flag that prohibits dependency resolution "at the wrong time" and enable that here?
> 
> > Thinking about it, it makes sense - CC can be allowed to resolve while serializing configurations.

Roughly speaking, CC resolves dependencies at serialization time, when it encounters `Configuration` and similar classes , like `ArtifactView` and only stores the result of the resolution. If we're to introduce the flag, I would imagine that without CC (in vintage mode), it would prohibit dependency resolution before execution time. With CC we move some work from execution to configuration, so we cannot employ the flag as is - we need to allow this "execution now happening at configuration" work to proceed without failing the build. 

This is quite complicated, we already tried a similar approach when trying to deprecate `Task.getProject` at execution time, but dealing with such calls wrapped in `provider {}` is tricky when trying to align CC and vintage.

big-guy: The bad behavior this test is trying to detect is based on this bug: https://issues.gradle.org/browse/GRADLE-3522.html

We were eagerly resolving the war's classpath during configuration time. 

The war task has a dependency on `assertUnresolved`, so checking the resolved state of `conf` when `assertUnresolved` runs will tell us if `conf` was resolved too early in vintage.

We don't have to check this with a task if we had some other way of enforcing that configurations cannot be resolved "early".

WRT build operations, I was thinking that we would be able to tell the difference between a configuration resolved during configuration time and resolved during CC serialization by looking at the parents/nesting of the build operations. We don't need to detect and enforce this universally, we just need to look at when `conf` is resolved. 

Adding a flag to prohibit this is something we'd like to eventually do, but it sounds like Mike thinks this would be complicated to do.

> Roughly speaking, CC resolves dependencies at serialization time, when it encounters Configuration and similar classes , like ArtifactView and only stores the result of the resolution.

@mlopatkin How does this work if the resolution is something produced by a task or artifact transform? 

Maybe we can split this into three buckets:
1. Resolution during configuration time. This is before/during the time we build the task graph. We want to prevent this and this is what the test was checking for. 
2. Resolution during "serialization time". This is the time after the task graph is built and before execution starts when CC serializes/deserializes things. It's unclear what should be prohibited right now. 
3. Resolution during execution time. We want to prevent _undeclared_ resolution, but we don't care about that for this test.

I don't have a strong feeling about this other than the fix to the test here looks wrong/fragile.

mlopatkin: If we don't care about resolving the configuration at serialization time (in theory, it can be a side effect of serializing some other thing), things get simpler. We have a build operation that covers the CC serialization, for example. But even the current `provider {}`-based solution is good enough for that (or at least as good, as the original test was for non-CC).

>How does this work if the resolution is something produced by a task or artifact transform?

Those remain lazy. CC relies a lot on the knowledge of inner structure of various file collections to take them apart and reconstruct something resembling the original "on the other side". I don't think we even run any transforms while serializing.

PR #36572, opened 2026-02-03
Don't stop searching early for class matches

### LINKED ISSUE #36544: Excluded tests in a suite are no longer being run
### Current Behavior

When using test suites, you can exclude the tests by filtering in the test task to avoid the test being run twice.

With Gradle 9.3.x, the test is no longer run when you exclude it, so you have to choose between running it twice or not running it.

The behaviour is a little different depending on whether you're using JUnit Jupiter or JUnit Vintage:
* With JUnit4, the excluded test is just non being run.
* With JUnit5/6, the test suite fails because it doesn't contain any tests.

### Expected Behavior

I'd expect test suites and exclusion filters to work like in previous versions of Gradle.

### Context (optional)

_No response_

https://github.com/ulrikls/gradle-junit4

### Gradle version

9.3.1

9.2.1

### Build scan URL (optional)

https://scans.gradle.com/s/f7e5tc2ck3tqg

_No response_

octylFractal:
Fixes #36544

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/junit/AbstractJUnitSuitesIntegrationTest.groovy
+++ b/platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/junit/AbstractJUnitSuitesIntegrationTest.groovy
@@ -20,6 +20,7 @@ import org.gradle.api.internal.tasks.testing.report.generic.GenericHtmlTestExecu
 import org.gradle.api.internal.tasks.testing.report.generic.GenericTestExecutionResult
 import org.gradle.api.tasks.testing.TestResult
 import org.gradle.testing.fixture.AbstractTestingMultiVersionIntegrationTest
+import spock.lang.Issue
 
 abstract class AbstractJUnitSuitesIntegrationTest extends AbstractTestingMultiVersionIntegrationTest {
     abstract String getTestFrameworkSuiteDependencies()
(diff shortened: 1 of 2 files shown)


platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/junit/AbstractJUnitSuitesIntegrationTest.groovy
```diff
@@ -85,4 +86,53 @@ abstract class AbstractJUnitSuitesIntegrationTest extends AbstractTestingMultiVe
...
+
+        when:
+        executer.withTasks('test').run()
+
+        then:
+        GenericHtmlTestExecutionResult result = resultsFor()
+        result.assertTestPathsExecuted(
+            ':org.gradle.SomeTestSuite:org.gradle.TestRunBySuite:ok'
+        )
+        result.testPath(":org.gradle.SomeTestSuite:org.gradle.TestRunBySuite:ok").onlyRoot().assertHasResult(TestResult.ResultType.SUCCESS)
```
tresat, line 140: 💭 I know we have other tests for it, but this will probably be clearer to readers if it also asserted that the excluded test was not run outside of the suite.

tresat (approved):

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/junit/AbstractJUnitSuitesIntegrationTest.groovy
octylFractal: I added that.

PR #36588, opened 2026-02-03
Fix CC incompatible tests in CopyTaskIntegrationSpec and CopyTaskEncodingIntegrationSpec

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/CopyTaskIntegrationSpec.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/CopyTaskIntegrationSpec.groovy
@@ -21,6 +21,7 @@ import org.gradle.api.plugins.ExtensionAware
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.integtests.fixtures.TestResources
 import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
+import org.gradle.integtests.fixtures.UnsupportedWithConfigurationCache
 import org.gradle.test.fixtures.file.TestFile
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.UnitTestPreconditions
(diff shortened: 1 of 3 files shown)


subprojects/core/src/testFixtures/groovy/org/gradle/api/tasks/FsOpsFixture.groovy
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.api.tasks
+
+class FsOpsFixture {
```
mlopatkin, line 19: 🎉  Nice

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/CopyTaskIntegrationSpec.groovy
```diff
@@ -1096,25 +1111,30 @@ class CopyTaskIntegrationSpec extends AbstractIntegrationSpec {
...
             'one/sub'
         )
+
+        where:
+        copyMethod << ['copy', 'fsOps.copy']
     }
 
-    @ToBeFixedForConfigurationCache(skip = INVESTIGATE)
+    @UnsupportedWithConfigurationCache(iterationMatchers = [".*#0.*"], because = "legacy Project.copy at execution time")
     def "copy from composite file collection"() {
```
mlopatkin: 💅 `#0` feels a bit fragile, maybe let's add the method to the test name and match on that? Here and in the other places

```suggestion
    @UnsupportedWithConfigurationCache(iterationMatchers = [".* with copy.*"], because = "legacy Project.copy at execution time")
   def "copy from composite file collection with #copyMethod"() {
```

mlopatkin (approved):
Looks fine, though I'd avoid matching on `#0` if possible.

blindpirate (approved):

PR #36591, opened 2026-02-04
Print detailed message when there is performance test failures

blindpirate:
Sometimes we see very confusing TeamCity performance test failure like [this](https://builds.gradle.org/buildConfiguration/Gradle_Release8x_Check_PerformanceTestTestLinux_Trigger/108418604), the report generator fails with "X scenarios failed", but we can't find any in performance report. 

This PR prints more detailed error message, which includes the TC build id of failed scenarios.

### DIFF
--- a/testing/internal-performance-testing/src/main/groovy/org/gradle/performance/results/report/AbstractReportGenerator.java
+++ b/testing/internal-performance-testing/src/main/groovy/org/gradle/performance/results/report/AbstractReportGenerator.java
@@ -115,7 +115,7 @@ public abstract class AbstractReportGenerator<R extends ResultsStore> {
             if (onlyOneTestExecuted) {
                 System.out.println("##teamcity[buildStatus text='" + failedMessage + "']");
             }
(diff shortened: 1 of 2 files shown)


testing/internal-performance-testing/src/main/groovy/org/gradle/performance/results/report/AbstractReportGenerator.java
```diff
@@ -199,26 +199,47 @@ public static class FailureCollector {
...
         private boolean failBuild;
+        private final List<String> failedScenarioSummaries = new ArrayList<>();
 
         public void scenarioFailed() {
             buildFailures++;
             failBuild = true;
         }
 
+        public void scenarioFailed(PerformanceReportScenario scenario) {
+            scenarioFailed();
```
cobexer, line 210: ❓ Do we still need the other signature without a scenario? If not, please merge them instead.

cobexer (approved):

PR #36605, opened 2026-02-05
Fix CC incompatible tests in WellBehavedPluginTest

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/antlr/src/integTest/groovy/org/gradle/api/plugins/antlr/AntlrPluginIntegrationTest.groovy
+++ b/platforms/jvm/antlr/src/integTest/groovy/org/gradle/api/plugins/antlr/AntlrPluginIntegrationTest.groovy
@@ -19,10 +19,15 @@ import org.gradle.integtests.fixtures.WellBehavedPluginTest
 import org.gradle.test.fixtures.archive.JarTestFixture
 import spock.lang.Issue
 
-import static org.gradle.test.fixtures.dsl.GradleDsl.GROOVY
 import static org.gradle.test.fixtures.dsl.GradleDsl.KOTLIN
 
 class AntlrPluginIntegrationTest extends WellBehavedPluginTest {
+    def setup() {
+        buildFile """
+            ${mavenCentralRepository()}
+        """
+    }
+
     @Override
     String getMainTask() {
         return "build"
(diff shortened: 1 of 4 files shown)


platforms/jvm/antlr/src/integTest/groovy/org/gradle/api/plugins/antlr/AntlrPluginIntegrationTest.groovy
```diff
@@ -23,6 +24,12 @@ import static org.gradle.test.fixtures.dsl.GradleDsl.GROOVY
 import static org.gradle.test.fixtures.dsl.GradleDsl.KOTLIN
 
 class AntlrPluginIntegrationTest extends WellBehavedPluginTest {
+    def setup() {
+        buildFile << """
+            ${RepoScriptBlockUtil.mavenCentralRepository()}
```
mlopatkin: 💅  There is `mavenCentralRepository` instance method in `WellBehavedPluginTest`, so calling a helper isn't necessary.

platforms/jvm/antlr/src/integTest/groovy/org/gradle/api/plugins/antlr/AntlrPluginIntegrationTest.groovy
mlopatkin: 💅 As we now have the repo in all build files, we can remove it from the specific tests.

platforms/jvm/antlr/src/integTest/groovy/org/gradle/api/plugins/antlr/AntlrPluginIntegrationTest.groovy
```diff
@@ -23,6 +24,12 @@ import static org.gradle.test.fixtures.dsl.GradleDsl.GROOVY
 import static org.gradle.test.fixtures.dsl.GradleDsl.KOTLIN
 
 class AntlrPluginIntegrationTest extends WellBehavedPluginTest {
+    def setup() {
+        buildFile << """
```
mlopatkin: 💅 Use `buildFile(String)` rather than `<<`
```suggestion
        buildFile """
```

Issue #31141, opened 2024-11-05
Deprecate calling Task.getExtensions at execution time


mlopatkin:
Currently, calling `Task.getExtensions` at execution time:

* emits a configuration cache problem if configuration cache is enabled
* OR emits a deprecation warning, if configuration cache is disabled and `STABLE_CONFIGURATION_CACHE` feature flag is enabled

As part of retiring the feature flag and preparing for CC-only in some not-so-distant major release, we should emit a deprecation warning regardless of the feature flag status.

There is no need to emit a deprecation warning if CC is enabled, because we don't plan to remove the method (unlike e.g. `useLogger`).

thedarkcolour:
What is the alternative way to access a plugin's extension from a task?

mlopatkin:
> What is the alternative way to access a plugin's extension from a task?

A general approach is to capture the necessary state explicitly. The exact way of doing that depends on whether the task/plugin/extension code can be modified and on the type of the state (lazy, like `Property`, or eager, JavaBean-style getter/setter pair). 

For example, a recommended way is to wire a `Property` of a task to the `Property` in extension, like [this](https://github.com/mlopatkin/andlogview/blob/master/gradle/plugins/licenses/src/main/kotlin/name/mlopatkin/gradleplugins/licenses/LicensePlugin.kt). Of course, it is only possible if the interfaces allow it. If the property is a getter/setter pair, it can be "converted" to lazy by wrapping it in `provider {}`. 

It becomes trickier when you don't own the task code, for example, when adding an extra action to an existing task. The solution, however, is similar - capture the necessary state:
```gradle
interface MyTaskExt {
  Property<String> getMessage()
}

tasks.named("fooTask") {
  // defining extension and action in e.g. some plugin
  extensions.add(MyTaskExt, "myExt", objects.newInstance(MyTaskExt))
 
  def extMessage = myExt.message // limit the state to capture
  inputs.property("message", extMessage) // optional, to make task properly cacheable
  doLast {
    println(extMessage.get())
  }
}

// ...
tasks.named("fooTask") {
  myExt.message = "some value" // late value binding in e.g. buildscript
}
```

Doing the same for eager types is similar:
```gradle
class MyTaskExt {
  String message
}

tasks.named("fooTask") {
  // defining extension and action in e.g. some plugin
  extensions.create("myExt", MyTaskExt)
 
  def extMessage = project.provider { myExt.message } // limit the state to capture, wrap it in provider to make read lazy.
[...]

PR #36621, opened 2026-02-05
Fix CC incompatible test in PropertyIntegrationTest

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
+++ b/platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
@@ -17,7 +17,8 @@
 package org.gradle.api.provider
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
-import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
+import org.gradle.integtests.fixtures.UnsupportedWithConfigurationCache
+import org.gradle.integtests.fixtures.executer.GradleContextualExecuter
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.IntegTestPreconditions
 import spock.lang.Issue
@@ -26,7 +27,7 @@ class PropertyIntegrationTest extends AbstractIntegrationSpec {
     def "can use property as task input"() {
         given:
         taskTypeWritesPropertyValueToFile()
-        buildFile << """
+        buildFile """
 
 task thing(type: SomeTask) {
     prop = providers.systemProperty('prop')
@@ -66,7 +67,7 @@ task thing(type: SomeTask) {
 
     def "can define task with abstract Property<#type> getter"() {
         given:
-        buildFile << """
+        buildFile """
             class Param<T> {
                 T display
                 String toString() { display.toString() }
@@ -101,7 +102,7 @@ task thing(type: SomeTask) {
 
     def "can define task with abstract nested property"() {
         given:
-        buildFile << """
+        buildFile """
             interface NestedType {
                 @Input
                 Property<String> getProp()
@@ -137,7 +138,7 @@ task thing(type: SomeTask) {
 
     def "fails when property with no value is queried"() {
         given:
-        buildFile << """
+        buildFile """
             abstract class SomeTask extends DefaultTask {
                 @Internal
                 abstract Property<String> getProp()
@@ -159,13 +160,9 @@ task thing(type: SomeTask) {
         failure.assertHasCause("Cannot query the value of task ':thing' property 'prop' because it has no value available.")
     }
 
-    @Requires(
-        value = IntegTestPreconditions.NotConfigCached,
-        reason = "Config cache does not support extensions during execution, so cause does not include any provenance information"
-    )
     def "fails when property with no value because source property has no value is queried"() {
         given:
-        buildFile << """
+        buildFile """
             interface SomeExtension {
                 Property<String> getSource()
             }
@@ -197,16 +194,22 @@ task thing(type: SomeTask) {
 
         then:
         failure.assertHasDescription("Execution failed for task ':thing'.")
-        failure.assertHasCause("""Cannot query the value of task ':thing' property 'prop' because it has no value available.
+        if (GradleContextualExecuter.isConfigCache()) {
+            failure.assertHasCause("""Cannot query the value of task ':thing' property 'prop' because it has no value available.
+The value of this property is derived from:
+  - Gradle property 'ABC'""")
+        } else {
+            failure.assertHasCause("""Cannot query the value of task ':thing' property 'prop' because it has no value available.
 The value of this property is derived from:
   - extension 'custom2' property 'source'
   - extension 'custom1' property 'source'
   - Gradle property 'ABC'""")
+        }
     }
 
     def "can use property with no value as optional ad hoc task input property"() {
         given:
-        buildFile << """
+        buildFile """
 
 def prop = project.objects.property(String)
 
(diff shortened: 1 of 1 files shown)


platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
```diff
@@ -197,16 +192,22 @@ task thing(type: SomeTask) {
 
         then:
         failure.assertHasDescription("Execution failed for task ':thing'.")
-        failure.assertHasCause("""Cannot query the value of task ':thing' property 'prop' because it has no value available.
+        if (GradleContextualExecuter.isConfigCache()) {
+            failure.assertHasCause("""Cannot query the value of task ':thing' property 'prop' because it has no value available.
```
ov7a, line 198: This looks like a serious problem. Do we have an issue to track this?

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
```diff
@@ -281,9 +281,9 @@ task thing(type: SomeTask) {
         output.count("calculating value") == 0
     }
 
-    @ToBeFixedForConfigurationCache(because = "gradle/configuration-cache#270")
+    @UnsupportedWithConfigurationCache(because = "https://github.com/gradle/configuration-cache/issues/46")
     def "does not calculate task @Input property value when task is skipped due to @SkipWhenEmpty on another property"() {
```
ov7a, line 287: I assume it's expected behaviour

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
```diff
@@ -443,13 +443,12 @@ assert custom.prop.get() == "value 4"
         succeeds()
     }
 
-    @Requires(
-        value = IntegTestPreconditions.NotConfigCached,
-        reason = "Config cache does not support extensions during execution, leading to 'Could not get unknown property 'custom' for task ':wrongValueTypeDsl' of type org.gradle.api.DefaultTask."
+    @UnsupportedWithConfigurationCache(
+        because = "Config cache does not support extensions during execution, leading to 'Could not get unknown property 'custom' for task ':wrongValueTypeDsl' of type org.gradle.api.DefaultTask."
```
ov7a: ❓ Do we have an issue to track this/do we have plans to support this scenario?

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
```diff
@@ -670,15 +668,14 @@ task wrongConventionRuntimeValueType {
     }
 
     @Issue("https://github.com/gradle/gradle/issues/12969")
-    @Requires(IntegTestPreconditions.NotParallelExecutor)
```
ov7a, line 673: The issue is fixed

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
```diff
@@ -613,7 +612,6 @@ task wrongConventionRuntimeValueType {
         'RegularFile'         | 'RegularFile' | 'RegularFileProperty'
     }
 
-    @Requires(IntegTestPreconditions.NotParallelExecutor)
     @Issue("https://github.com/gradle/gradle/issues/12811")
```
ov7a, line 537: The issue is fixed

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
mlopatkin: > ❓ Do we have an issue to track this/do we have plans to support this scenario?

We have https://github.com/gradle/gradle/issues/31141. We don't want to support task extensions at execution time.

mlopatkin: ❌  It doesn't look like the test needs to do the assignment to an extension at execution time. I'd move the property to the task instead.
```gradle
abstract class StringPropTask {
  @Internal
  abstract Property<String> getProp()
}

tasks.register("wrongValueTypeDsl", StringPropTask) { // I don't remember the details of the `task foo` syntax, lol; always register 
    doLast {
       prop = 123
    }
}
```
and so on.

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
mlopatkin: > This looks like a serious problem. Do we have an issue to track this?

I cannot find any, please file a new one. We don't preserve display names for properties today.

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
mlopatkin: > I assume it's expected behaviour

It is.

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
mlopatkin: > The issue is fixed

🤔  I think the test requires `NotParallel` because the build under test itself uses `--parallel`. So there is no point in rerunning it with the parallel executor, it will be the same as forking. So we're just saving some test time. It may not be actually worth the hassle, but I don't have a strong opinion here.

Maybe updating the reason for `@Requires` would help the next reader to avoid the confusion.

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
mlopatkin: > The issue is fixed

Same as above, `--parallel` specified explicitly.

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
ov7a: Can you clarify? Do you mean that 
```
  - extension 'custom2' property 'source'
  - extension 'custom1' property 'source'
```
come from the display name?

mlopatkin: > Do you mean that
> 
> ```
>   - extension 'custom2' property 'source'
>   - extension 'custom1' property 'source'
> ```
> 
> come from the display name?

Yes, the chain is built there: https://github.com/gradle/gradle/blob/e729ac95b88a4591f9e22553a6550e8baba414a7/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/AbstractMinimalProvider.java#L115

As we don't restore display names for the intermediates, the chain has nothing.

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
ov7a: A-ha, got it. Potentially, we can lose some coverage this way (setting a property in an extension), but it doesn't look critical?

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/PropertyIntegrationTest.groovy
ov7a: https://github.com/gradle/gradle/issues/36637

PR #36626, opened 2026-02-05
Move project features classes to more appropriate packages

### LINKED ISSUE #36622: Move packages to more reasonable non-internal locations
The package for many of the project feature related classes don't make sense any more (i.e. org.gradle.api.software or something similar) and in many cases the classes are intended to eventually be part of the public API, but have "internal" in the package name.  

This is technical debt we've been postponing for some time, but we should go ahead an move these into better package names (i.e. "org.gradle.features...") and move the to-be-public classes into packages without "internal" in the name.  This will be a breaking change, but hopefully only something we have to do once.

ghale:
Fixes #36622 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

testing/architecture-test/src/changes/accepted-changes/accepted-public-api-changes.json
```diff
@@ -1,3 +1,12 @@
 {
-    "acceptedApiChanges": []
-}
\ No newline at end of file
+    "acceptedApiChanges": [
+        {
+            "type": "org.gradle.api.tasks.diagnostics.ProjectReportTask",
```
h0tk3y, line 4: `ProjectFeatureDeclarations getProjectFeatureDeclarations()` is already unused in that task type. Can we simply remove the property?

platforms/core-configuration/project-features-api/src/main/java/org/gradle/features/binding/ProjectFeatureApplyAction.java
```diff
@@ -14,15 +14,30 @@
...
+     * Apply configuration from the project feature definition to its build model and execute any necessary build logic.
+     *
+     * @param context the application context
+     * @param definition the project feature definition
+     * @param buildModel the project feature's own build model
+     * @param parentDefinition the parent project feature definition
+     *
+     * @since 9.5.0
+     */
     void transform(ProjectFeatureApplicationContext context, OwnDefinition definition, OwnBuildModel buildModel, ParentDefinition parentDefinition);
```
h0tk3y, line 42: By the way, should we rename this method now? It was originally called `transform` according to the terminology of that time, but we moved away from calling it "transforms".

h0tk3y (changes requested):

testing/architecture-test/src/changes/accepted-changes/accepted-public-api-changes.json
ghale: Yep, I didn't notice it was no longer used.  Removed.

platforms/core-configuration/project-features-api/src/main/java/org/gradle/features/binding/ProjectFeatureApplyAction.java
ghale: I was just thinking about this earlier today while working on #36638.  I reckon we should probably change this to "apply()" or something like that.  I'll do that as part of that work.

platforms/core-configuration/project-features-api/src/main/java/org/gradle/features/annotations/RegistersSoftwareTypes.java
hfhbd: Shouldn't it be `RegistersProjectTypes`? Changing the packages requires recompiling anyway, so the old `SoftwareType` name could be removed with this change too.

ghale: Thanks for bringing this up.  Actually, we should probably just remove this annotation with this PR.  We had always intended to remove it at some point.

PR #36627, opened 2026-02-05
Establish report container ownership hierarchy

tresat:


platforms/core-configuration/model-core/src/main/java/org/gradle/internal/state/OwnerAware.java
```diff
@@ -28,7 +28,7 @@ public interface OwnerAware {
      * Notifies this object that it now has an owner associated with it.
      *
      * @param owner The owner object, if any.
-     * @param displayName The display name for this object.
+     * @param displayName The display name for the owner object.
```
tresat, line 31: "This" here means "the owner object", not `this`, as in the instance executing the `attachOwner` call.

platforms/software/reporting/src/integTest/groovy/org/gradle/api/reporting/internal/ReportContainerIntegTest.groovy
```diff
@@ -37,25 +40,27 @@ class ReportContainerIntegTest extends AbstractIntegrationSpec {
...
-                TestReportContainer(ObjectFactory objectFactory, Describable owner) {
-                    super(DefaultReportContainer.create(objectFactory, Report, factory -> [
-                        factory.instantiateReport(DefaultSingleFileReport, "file1", owner),
-                        factory.instantiateReport(DefaultSingleFileReport, "file2", owner),
-                        factory.instantiateReport(SingleDirectoryReport, "dir1", owner, null),
-                        factory.instantiateReport(SingleDirectoryReport, "dir2", owner, null)
-                    ]))
+                TestReportContainer(ObjectFactory objectFactory) {
+                    super(DefaultReportContainer.create(objectFactory, Report, factory -> {
+                        Describable owner = Describables.of("report container")
```
tresat, line 51: Who owns each report?  The (delegated-to) report container.  This seems like duplication, but is only used to provide a full-er name for each report and doesn't have anything to do with model object ownership.

platforms/software/reporting/src/integTest/groovy/org/gradle/api/reporting/internal/ReportContainerIntegTest.groovy
```diff
@@ -37,25 +40,27 @@ class ReportContainerIntegTest extends AbstractIntegrationSpec {
...
             import ${DefaultSingleFileReport.class.name}
             import ${SingleDirectoryReport.class.name}
             import ${DelegatingReportContainer.class.name}
             import ${DefaultReportContainer.class.name}
 
             import ${Inject.class.name}
 
-            class TestReportContainer extends DelegatingReportContainer<Report> {
-
+            abstract class TestReportContainer extends DelegatingReportContainer<Report> {
```
tresat, line 47: Must be abstract to receive the task as its owner via automatic property instantiation in the Task.  This necessitates removing the `Describable` arg from the `TaskReportContainer` constructor as well, as this is not an injectable service.

platforms/software/reporting/src/integTest/groovy/org/gradle/api/reporting/internal/ReportContainerIntegTest.groovy
```diff
@@ -69,14 +74,8 @@ class ReportContainerIntegTest extends AbstractIntegrationSpec {
             import ${Inject.class.name}
 
             abstract class TestTask extends DefaultTask {
-
                 @Nested
-                TestReportContainer reports
-
-                @Inject
-                public TestTask() {
-                    this.reports = getProject().getObjects().newInstance(TestReportContainer, Describables.quoted("Task", getIdentityPath()))
```
tresat, line 78: Besides being unnecessary when making the property abstract, this is incorrect hierarchy - it establishes that the owner of each report is the task itself, ignoring the 2 levels of containers.

platforms/software/reporting/src/main/java/org/gradle/api/reporting/internal/DefaultReportContainer.java
```diff
@@ -80,7 +85,16 @@ public DefaultReportContainer(
...
         super(type, instantiatorFactory.decorateLenient(servicesToInject), Report.NAMER, callbackActionDecorator);
-        this.addAll(reportGenerator.generateReports(new DefaultReportFactory<>(getInstantiator())));
+
+        @SuppressWarnings("unchecked")
+        DefaultReportFactory<T> factory = (DefaultReportFactory<T>) getInstantiator().newInstance(DefaultReportFactory.class, getInstantiator());
+        Collection<T> reports = reportGenerator.generateReports(factory);
+        this.addAll(reports);
+
+        reports.forEach(r -> {
+            ((OwnerAware) r).attachOwner((ModelObject) this, Describables.of("report container"));
```
tresat: Each report is associated with the container that created it immediately after creation.

platforms/software/reporting/src/main/java/org/gradle/api/reporting/internal/DefaultSingleFileReport.java
```diff
@@ -27,7 +27,12 @@
 import java.io.File;
 
 public abstract class DefaultSingleFileReport extends SimpleReport implements SingleFileReport {
-
+    /**
+     * Creates a single file report.
+     *
+     * @param name The name of the report
+     * @param owner A {@link Describable} that describes the container that contains and owns this report
```
tresat: Clarify that the `owner` param is only used for naming.

platforms/software/reporting/src/main/java/org/gradle/api/reporting/internal/DefaultSingleFileReport.java
```diff
@@ -27,7 +27,13 @@
 import java.io.File;
 
 public abstract class DefaultSingleFileReport extends SimpleReport implements SingleFileReport {
-
+    /**
+     * Creates a single file report.
+     *
+     * @param name The name of the report
+     * @param owner A {@link Describable} that describes the container that contains and owns this report,
+     *   not necessarily the owner itself, only used for naming purposes
```
tresat, line 35: Clarify that the owner param is only used for naming.

platforms/software/reporting/src/main/java/org/gradle/api/reporting/internal/SingleDirectoryReport.java
```diff
@@ -32,9 +31,17 @@ public abstract class SingleDirectoryReport extends SimpleReport implements Dire
     @Nullable
     private final String relativeEntryPath;
 
+    /**
+     * Creates a single directory report.
+     *
+     * @param name The name of the report
+     * @param owner A {@link Describable} that describes the container that contains and owns this report
+     * @param relativeEntryPath The path of the entry point file relative to the report directory, or null if the entry point is the report directory itself,
+     *   not necessarily the owner itself, only used for naming purposes
```
tresat, line 41: Clarify that the owner param is only used for naming.

platforms/software/reporting/src/main/java/org/gradle/api/reporting/internal/DelegatingReportContainer.java
```diff
@@ -32,6 +35,7 @@ public abstract class DelegatingReportContainer<T extends Report> extends Delega
 
     public DelegatingReportContainer(ReportContainer<T> delegate) {
         super(delegate);
+        ((OwnerAware) delegate).attachOwner((ModelObject) this, Describables.of("delegating report container"));
```
tresat, line 38: The delegated-to container should be aware of the delegating container.

Its easy to think that we could just build a hierarchy with only one of these 2 containers.  The problem is that by creating the delegated-to container using DefaultReportContainer.create, the delegated-to container is not automatically wired to have the Task as its owner (only the delegating container has this).  Since we can't pass a reference to `this` when calling `super` when creating `TestReportContainer`, we need to wire both levels of containers.

If we removed the requirement for an immutable `DefaultReportContainer` and instead allowed calling create on reports as needed, we would possibly have more design options in terms of having the report associated directly with the delegating container, and ignore the implementation detail of delegation entirely.

platforms/software/reporting/src/main/java/org/gradle/api/reporting/internal/DefaultReportContainer.java
```diff
@@ -127,12 +131,13 @@ public DefaultReportFactory(Instantiator instantiator) {
...
 
         @Override
-        public <N extends T> N instantiateReport(Class<N> clazz, Object... constructionArgs) {
+        public <N extends R> N instantiateReport(Class<N> clazz, Object... constructionArgs) {
             N report = instantiator.newInstance(clazz, constructionArgs);
             String name = report.getName();
             if (name.equals("enabled")) {
                 throw new InvalidUserDataException("Reports that are part of a ReportContainer cannot be named 'enabled'");
             }
+            ((OwnerAware) report).attachOwner((ModelObject) DefaultReportContainer.this, Describables.of("report container"));
```
tresat, line 140: When we create the reports, attach them to the container that created them and will contain them.

platforms/software/reporting/src/main/java/org/gradle/api/reporting/internal/DefaultReportContainer.java
tresat, line 71: This is really the root problem here.  We can't make this container abstract and have Gradle create it as a property because we have to provide a generator lambda to make the reports and be able to treat the collection as immutable.

platforms/software/reporting/src/main/java/org/gradle/api/reporting/Report.java
```diff
@@ -33,7 +35,8 @@ public interface Report extends Configurable<Report> {
     Namer<Report> NAMER = new Namer<Report>() {
         @Override
         public String determineName(Report report) {
-            return report.getName();
+            //noinspection OptionalOfNullableMisuse
+            return Optional.ofNullable(report.getName()).orElse("unnamed");
```
tresat, line 39: If you run `JavaLibraryInitIntegrationTest` without this change, it will fail with:

```
Caused by: org.gradle.internal.exceptions.LocationAwareException: Could not load the value of field `values` of `org.gradle.api.internal.collections.SortedSetElementSource` bean found in field `collection` of `org.gradle.api.internal.collections.FilteredElementSource` bean found in field `store` of `org.gradle.api.internal.DefaultNamedDomainObjectSet_Decorated` bean found in field `enabled` of `org.gradle.api.reporting.internal.DefaultReportContainer_Decorated` bean found in field `_gr_owner_` of `org.gradle.api.reporting.internal.SingleDirectoryReport_Decorated` bean found in field `html` of `org.gradle.api.internal.tasks.testing.testng.TestNGTestFramework_Decorated` bean found in field `testFramework` of task `:lib:test` of type `org.gradle.api.tasks.testing.Test`.
```

Which ultimately comes from trying to compare the names of deserialized reports to add them to a treeset using `Namer.Comparator` before the (actually non-null) `name` field has been deserialized.  We could also update the Comparator itself to fix this issue.

Issue #29653, opened 2024-06-21
Support Grade working in environments that only support IPv4 or IPv6 networks


big-guy:
### Current Behavior

The system properties that control JVM behavior cannot be supplied by users.  Grade does try to set those system properties in the build process, but the system properties are set too ate.

### Expected Behavior

Either Grade should automatically enable/disable IPv4/IPv6 as necessary or Gradle should allow users to control this.

This was attempted [earlier](https://github.com/gradle/gradle/pull/12741), but this caused a regression with CLI/IDE daemon reuse because the IDE always set java.net.preferIPv4Stack, but the CLI did not.

### Context (optional)

https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F/java/net/doc-files/net-properties.html

java.net.preferIPv4Stack and java.net.preferIPv6Addresses control if IPv4 or IPv6 is supported by the JVM. These properties are read once at startup and cannot be changed later.

You may want to set java.net.preferIPv6Addresses=true on hosts that only support IPv6 networking, but sometimes have to choose between IPv4 and IPv6 resolved IPs. (the opposite is also true). 

### Steps to Reproduce

1. Setup a machine that can only access IPv6 networks
2. Resolve a domain name that has both IPv4 and IPv6 IPs
3. Observe this fails when trying to use IPv4
4. Try to configure Gradle to prefer IPv6

As a workaround, you _can_ set the system properties by setting the environment variable `JAVA_TOOL_OPTIONS`. This propagates to every Java process and produces an annoying message every time Java runs. 

### Gradle version

All

PR #36634, opened 2026-02-06
Allow address resolver properties for Gradle Daemon

snazy:
Currently it is not possible to set the system properties `java.net.preferIPv6Addresses` and `java.net.preferIPv4Stack` for the Gradle Daemon. As for example artifact resolution depends on Java's name resolution, downloads almost always happen via IPv4.

As Java complies with "Hyrum's law", the IP addresses returned by `InetAddress.get*()` yield IPv4 _before_ IPv6 addresses, even if it ignores the system settings and [RFC 6724](https://datatracker.ietf.org/doc/html/rfc6724). To comply with RFC 6724, users have to set `java.net.preferIPv6Addresses=true`. `java.net.preferIPv4Stack` is generally useful when IPv6 issues arise.

Background for this issue: `org.gradle.launcher.daemon.configuration.DaemonParameters#getEffectiveJvmArgs` yields only the _immutable_ system properties for the Gradle daemon JVM arguments, not all JVM arguments configured for example via `org.gradle.jvmargs` in `gradle.properties` or `gradlew`'s `-D` argument.

### Contributor Checklist
- [X] [Review Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md).
- [X] Make sure that all commits are [signed off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff) to indicate that you agree to the terms of [Developer Certificate of Origin](https://developercertificate.org/).
- [X] Make sure all contributed code can be distributed under the terms of the [Apache License 2.0](https://github.com/gradle/gradle/blob/master/LICENSE), e.g. the code was written by yourself or the original code is licensed under [a license compatible to Apache License 2.0](https://apache.org/legal/resolved.html).
- [X] Check ["Allow edit from maintainers" option](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) in pull request so that additional changes can be pushed by Gradle team.
- [X] Provide unit tests (under `<subproject>/src/test`) to verify logic.
- [X] Ensure that tests pass locally: `./gradlew <changed-subproject>:quickTest`.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/process/internal/JvmOptions.java
+++ b/subprojects/core/src/main/java/org/gradle/process/internal/JvmOptions.java
@@ -62,11 +62,22 @@ public class JvmOptions {
     public static final String SSL_TRUSTPASSWORD_KEY = "javax.net.ssl.trustStorePassword";
     public static final String SSL_TRUSTSTORETYPE_KEY = "javax.net.ssl.trustStoreType";
 
+    // The following three properties are used in java.net.InetAddress.initializePlatformLookupPolicy
+    // (static initialization).
+
+    // Useful fallback when using IPv6 causes issues.
+    public static final String JAVA_NET_PREFER_IPV4_STACK_KEY = "java.net.preferIPv4Stack";
+    // Note: InetAddress.get*() yield IPv4 addresses before IPv6 addresses (Hyrum's law).
+    // To comply with https://datatracker.ietf.org/doc/html/rfc6724, users have to set
+    // java.net.preferIPv6Addresses=true.
+    public static final String JAVA_NET_PREFER_IPV6_ADDRESSES_KEY = "java.net.preferIPv6Addresses";
+
     private static final Logger LOGGER = LoggerFactory.getLogger(JvmOptions.class);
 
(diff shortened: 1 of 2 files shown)


snazy:
FYI [this is the code in InetAddress](https://github.com/openjdk/jdk/blob/bfd5bdaf7ffd96cf887fbbfe072414be38ee1b84/src/java.base/share/classes/java/net/InetAddress.java#L397-L423) that determines the behavior.

h0tk3y:
This pull request needs a decision from the team responsible for that area. They have been informed. Response time may vary.

---

@snazy, could you please confirm that the issue linked below is the one this PR is addressing? Also, could you please explain the suggested solution?

Related issue: 
* https://github.com/gradle/gradle/issues/29653

mlopatkin:
In the past, we attempted a similar fix in #12741 but it was reverted because of daemon reuse issues. It looks like modern IDE version no longer add IP stack flags on their own, but it would be great to confirm before we can proceed with PR. IDEA and Android Studio are of most interest.

Issue #31858, opened 2024-12-20
Custom decompression sources do not work with tar trees when configuration cache is used


mlopatkin:
### Current Behavior

Trying to extract BZIP2-compressed TAR fails with:
```
> Unable to expand TAR '/Users/mlopatkin/Projects/gradle/worktrees/gradle/subprojects/core/build/tmp/teŝt files/ArchiveInte.Test/ojv9b/test.ext'
    The tar might be corrupted or it is compressed in an unexpected way.
    By default the tar tree tries to guess the compression based on the file extension.
    If you need to specify the compression explicitly please refer to the DSL reference.
```

### Expected Behavior

Test should complete successfully

### Self-contained Reproducer Project

The [broken test](https://github.com/gradle/gradle/blob/ed8d558d91cf9a6d0091cc7141857b52fb1f3775/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/bundling/ArchiveIntegrationTest.groovy#L325):
```gradle
task copy(type: Copy) {
    from tarTree(resources.bzip2('test.ext'))
    exclude '**/2.txt'
    into 'dest'
}
```

### Gradle version

any

mlopatkin:
FileTreeCodec doesn't take into account the transformation specified in the TarTree, but only the [file path](https://github.com/gradle/gradle/blob/ed8d558d91cf9a6d0091cc7141857b52fb1f3775/platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/FileTreeCodec.kt#L180)

PR #36640, opened 2026-02-06
Fix CC incompatible tests for file tasks

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/bundling/ReuseArchiveIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/bundling/ReuseArchiveIntegrationTest.groovy
@@ -17,12 +17,10 @@
 package org.gradle.api.tasks.bundling
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
-import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
+import org.gradle.integtests.fixtures.executer.GradleContextualExecuter
 import org.gradle.internal.hash.DefaultFileHasher
 import org.gradle.internal.hash.DefaultStreamHasher
 import org.gradle.internal.hash.FileHasher
-
-import static org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache.Skip.INVESTIGATE
 
 class ReuseArchiveIntegrationTest  extends AbstractIntegrationSpec {
 
@@ -31,16 +29,16 @@ class ReuseArchiveIntegrationTest  extends AbstractIntegrationSpec {
      * happens to contain content with the same name as content you're trying to cache?  The content used
      * by the cache no longer agrees with the content coming from the zip itself.
      */
-    @ToBeFixedForConfigurationCache(skip = INVESTIGATE)
     def "pre-existing content in cache dir with same hash is okay"() {
         file("contents/hello.txt") << "hello"
         file("contents").zipTo(file("hello.zip"))
         FileHasher hasher = new DefaultFileHasher(new DefaultStreamHasher())
         def hash = hasher.hash(file("hello.zip"))
-        def cachedFile = file("build/tmp/.cache/expanded/zip_${hash}/hello.txt")
-        def otherFile = file("build/tmp/.cache/expanded/zip_${hash}/other.txt")
+        def cacheDir = GradleContextualExecuter.isConfigCache() ? executer.gradleUserHomeDir.file(".tmp/.cache/expanded") : file("build/tmp/.cache/expanded")
+        def cachedFile = cacheDir.file("zip_${hash}/hello.txt")
+        def otherFile = cacheDir.file("zip_${hash}/other.txt")
 
-        buildFile << """
+        buildFile """
             abstract class CopyAndList extends DefaultTask {
                 @Inject
                 abstract FileSystemOperations getFileSystemOperations()
@@ -62,7 +60,7 @@ class ReuseArchiveIntegrationTest  extends AbstractIntegrationSpec {
                     }
                 }
             }
-            task extract2(type: CopyAndList) {
+            task extract(type: CopyAndList) {
                 unzipped.from(zipTree(file("hello.zip")))
             }
         """


ov7a:
@bot-gradle test this

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/CopyTaskChildSpecIntegrationTest.groovy
```diff
@@ -18,36 +18,47 @@ package org.gradle.api.tasks
...
+            run("copy")
+        }
 
         then:
-        failure.assertHasCause("You cannot add child specs at execution time. Consider configuring this task during configuration time or using a separate task to do the configuration.")
+        if (GradleContextualExecuter.notConfigCache) {
+            failure.assertHasCause("You cannot add child specs at execution time. Consider configuring this task during configuration time or using a separate task to do the configuration.")
+            file("build/output/input.txt").assertDoesNotExist()
+        } else {
+            // See https://github.com/gradle/gradle/issues/32540
```
ov7a, line 57: The fact that this scenario works might surprise a user. However, I don't think it's frequent, and the outcome is favorable (I hope).

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/DeleteTaskIntegrationTest.groovy
```diff
@@ -73,19 +72,21 @@ class DeleteTaskIntegrationTest extends AbstractIntegrationSpec {
...
                         }
                     })
-                    def destroyableFiles = files(destroyablePaths).files
-                    assert destroyableFiles.size() == 3 &&
-                        destroyableFiles.containsAll([
-                            file('foo'),
-                            file('bar'),
-                            file('baz')
-                        ])
+                    def destroyableFiles = objects.fileCollection().from(destroyablePaths).files
```
ov7a, line 75: I tried to change the machinery to use `TaskDestroyablesInternal.getRegisteredFiles`, but this collection is empty at that time.

Overall, I think it's not a good test, because it uses and tests internals. But refactoring it properly may take time and is outside the scope of this PR.

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/bundling/ReuseArchiveIntegrationTest.groovy
```diff
@@ -17,30 +17,28 @@
...
      */
-    @ToBeFixedForConfigurationCache(skip = INVESTIGATE)
     def "pre-existing content in cache dir with same hash is okay"() {
         file("contents/hello.txt") << "hello"
         file("contents").zipTo(file("hello.zip"))
         FileHasher hasher = new DefaultFileHasher(new DefaultStreamHasher())
         def hash = hasher.hash(file("hello.zip"))
-        def cachedFile = file("build/tmp/.cache/expanded/zip_${hash}/hello.txt")
-        def otherFile = file("build/tmp/.cache/expanded/zip_${hash}/other.txt")
+        def cacheDir = GradleContextualExecuter.isConfigCache() ? executer.gradleUserHomeDir.file(".tmp/.cache/expanded") : file("build/tmp/.cache/expanded")
```
ov7a, line 37: This is concerning. Maybe the root cause is the same as for
* https://github.com/gradle/gradle/issues/31858

Is it expected to have a different cache folder with CC?

testing/integ-test/src/integTest/groovy/org/gradle/integtests/SyncTaskIntegrationTest.groovy
```diff
@@ -385,19 +386,22 @@ class SyncTaskIntegrationTest extends AbstractIntegrationSpec implements StableC
...
                         into 'dest'
                     }
                 }
             }
-        '''.stripIndent()
+        """
 
         when:
-        expectTaskGetProjectDeprecations()
+        if (syncMethod == "project.sync") {
```
ov7a, line 402: There is no just `sync(Closure)` (but there's `copy(Closure)`), thus this condition.

I wonder if with `copy` (not `project.copy`) it's a bug in deprecation detection

testing/integ-test/src/integTest/groovy/org/gradle/integtests/SyncTaskIntegrationTest.groovy
```diff
@@ -503,7 +516,6 @@ class SyncTaskIntegrationTest extends AbstractIntegrationSpec implements StableC
         }
     }
 
-    @ToBeFixedForConfigurationCache(skip = ToBeFixedForConfigurationCache.Skip.FLAKY)
```
ov7a, line 506: It's been a long time since this annotation was added. I was unable to reproduce the flakiness and am unsure where it comes from. If nobody knows I suggest to just remove the annotation - it can always be added back with fresh data

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/bundling/ReuseArchiveIntegrationTest.groovy
mlopatkin: >Is it expected to have a different cache folder with CC?

This is suspicious indeed. We need to investigate while we end up in the GUH, perhaps some base directory is not restored properly for the task. Please file an issue.

testing/integ-test/src/integTest/groovy/org/gradle/integtests/SyncTaskIntegrationTest.groovy
mlopatkin: >There is no just sync(Closure) (but there's copy(Closure)), thus this condition.

🤯 Our bytecode magic should handle that.

ov7a:
```
tasks.register("someTask"){
  doLast {
//    copy { // triggers an error with CC
//      from("build/input.txt")
//      into("build/output")
//    }
    project.copy { //CC warning
      from("build/input.txt")
      into("build/output")
    }
//    sync { just sync does not work, because there's no sync(Closure)
//      from("build/input.txt")
//      into("build/output")
//    }
    project.sync { //CC warning
      from("build/input.txt")
      into("build/output")
    }
  }
}
```
I guess this is [expected](https://github.com/gradle/gradle/blob/8fbe0e2275e042f4c4cb4563da520964c8f00b34/subprojects/core-api/src/main/java/org/gradle/api/Project.java#L1820). In any case, I don't think we should add plain `sync` since it's effectively obsolete

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/bundling/ReuseArchiveIntegrationTest.groovy
ov7a: https://github.com/gradle/gradle/issues/36659

PR #36641, opened 2026-02-06
Fix test to run with CC

tresat:


platforms/jvm/language-jvm/src/integTest/groovy/org/gradle/integtests/resolve/consistency/JavaProjectResolutionConsistencyIntegrationTest.groovy
```diff
@@ -238,11 +237,22 @@ class JavaProjectResolutionConsistencyIntegrationTest extends AbstractHttpDepend
...
+                }
+            }
+
             tasks.named('checkDeps') {
+                def myCompileClasspathDeps = configurations.myCompileClasspath
+                inputs.files myCompileClasspathDeps
+
+                // in order to trigger the bug, we need to resolve the configuration
+                // which extends from a resolvable configuration first
+                configurations.myCompileClasspath.resolve()
```
tresat, line 257: This seems to work just fine, and the extending configuration is being resolved first.

platforms/jvm/language-jvm/src/integTest/groovy/org/gradle/integtests/resolve/consistency/JavaProjectResolutionConsistencyIntegrationTest.groovy
```diff
@@ -238,11 +237,22 @@ class JavaProjectResolutionConsistencyIntegrationTest extends AbstractHttpDepend
...
+
                 doFirst {
-                    // in order to trigger the bug, we need to resolve the configuration
-                    // which extends from a resolvable configuration first
-                    configurations.myCompileClasspath.resolve()
+                    println "myCompileClasspath dependencies: " + myCompileClasspathDeps.files
```
tresat, line 260: TaskAction doesn't even seem to be necessary, but I kept it just to verify the task is run.

platforms/jvm/language-jvm/src/integTest/groovy/org/gradle/integtests/resolve/consistency/JavaProjectResolutionConsistencyIntegrationTest.groovy
mlopatkin: > This seems to work just fine, and the extending configuration is being resolved first.

Now we're resolving at configuration time, which is frowned upon. Can we just rely on CC resolving the configuration while serializing? And calling `files()` at execution time should also be enough to trigger resolution in vintage, right?

tresat: We need to ensure the order of configuration resolutions though, to reproduce the bug this is testing.  

When forcing resolution at configuration time, we can rely on `myCompileClasspath` being resolved first:

```
Parallel Configuration Cache is an incubating feature.
Resolving configuration: myCompileClasspath
handling http request: GET /repo/org/foo/1.0/foo-1.0.pom
handling http request: GET /repo/org/foo/1.0/foo-1.0.jar
handling http request: GET /repo/org/bar/1.0/bar-1.0.pom
Resolving configuration: compileClasspath
Resolving configuration: runtimeClasspath

> Task :checkDeps
myCompileClasspath dependencies: [/Users/ttresansky/Projects/gradle-worktrees/master/platforms/jvm/language-jvm/build/tmp/teŝt files/JavaProject.Test/zhihs/user-home/caches/modules-2/files-2.1/org/foo/1.0/db99a9fdbda3eb7fba33b5a4418b9f5a1db63b35/foo-1.0.jar]

```

Without doing this (commenting out the line with `.resolve()`, we'll resolve _all_ the configurations at CC-serialization time, which results in the wrong order to reproduce the bug:

```
Parallel Configuration Cache is an incubating feature.
Resolving configuration: compileClasspath
Resolving configuration: runtimeClasspath
Resolving configuration: myCompileClasspath
 
> Task :checkDeps
myCompileClasspath dependencies: [/Users/ttresansky/Projects/gradle-worktrees/master/platforms/jvm/language-jvm/build/tmp/teŝt files/JavaProject.Test/ibemz/user-home/caches/modules-2/files-2.1/org/foo/1.0/db99a9fdbda3eb7fba33b5a4418b9f5a1db63b35/foo-1.0.jar] 
```

Is there some other way to ensure the ordering checks I added to the test still pass?  Otherwise, maybe I can just mark this test as "DO NOT DUPLICATE THIS BEHAVIOR" or something.

mlopatkin: Ok, I see now. I guess it is good enough to unblock us then. We can revisit the test when/if Gradle starts complaining about resolving stuff at inappropriate time.

PR #36642, opened 2026-02-06
Add sources for published Gradle API JAR

ljacomet:


build-logic/packaging/src/main/kotlin/gradlebuild.public-api-jar.gradle.kts
```diff
@@ -81,17 +82,111 @@ val task = tasks.register<Jar>("jarGradleApi") {
 // and its external dependencies.
 val gradleApiElements = configurations.consumable("gradleApiElements") {
     extendsFrom(externalApi.get())
-    outgoing.artifact(task)
+    outgoing.artifact(apiJarTask)
     configureAsApiElements(objects)
 }
 
+@CacheableTask
```
cobexer: ❓ Could we move the types to https://github.com/gradle/gradle/tree/e09fd46682616eccdea926f93bb24555e0b36beb/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks instead of inlining them here?

build-logic/packaging/src/main/kotlin/gradlebuild.public-api-jar.gradle.kts
```diff
@@ -81,17 +82,111 @@ val task = tasks.register<Jar>("jarGradleApi") {
...
+    }
+}
+
+interface FileParameters : ValueSourceParameters {
+    val inputFile: RegularFileProperty
+}
+
+abstract class FileLinesValueSource : ValueSource<List<String>, FileParameters> {
+    override fun obtain(): List<String> {
+        println("Constructing path prefixes")
```
cobexer: 💅 remove

build-logic/packaging/src/main/kotlin/gradlebuild.public-api-jar.gradle.kts
```diff
@@ -81,17 +82,111 @@ val task = tasks.register<Jar>("jarGradleApi") {
...
+        return parameters.inputFile.map { it.asFile.readLines() }.get()
```
mlopatkin: 💅 Doing get() straight away would probably be marginally more effective (fewer intermediate objects). I'd also put everything into set, so `contains` will be a tad quicker.
```suggestion
        return parameters.inputFile.get().asFile.useLines { it.toSet() }
```

build-logic/packaging/src/main/kotlin/gradlebuild.public-api-jar.gradle.kts
```diff
@@ -81,17 +82,111 @@ val task = tasks.register<Jar>("jarGradleApi") {
...
+        }.files.elements.map { set -> set.map { file -> archiveOperations.zipTree(file.asFile) } }
+    }}) {
+        include { element ->
+            val parent = element.relativePath.parent
+            if (parent == null) false
+            else pathPrefixes.get().contains(parent.pathString)
+        }
+    }
+
+    inputs.files(pathPrefixListTask.flatMap { it.pathPrefixFile }).withPropertyName("pathPrefixList").withPathSensitivity(PathSensitivity.RELATIVE)
```
mlopatkin: 🤔 Why do you need path sensitivity here, it doesn't feel like it matters where the list is located?

build-logic/packaging/src/main/kotlin/gradlebuild.public-api-jar.gradle.kts
```diff
@@ -81,17 +82,111 @@ val task = tasks.register<Jar>("jarGradleApi") {
...
+}
+
+val sourceJarTask = tasks.register<Jar>("sourcesJar") {
+    val pathPrefixes = providers.of(FileLinesValueSource::class.java) {
+        parameters.inputFile.set(pathPrefixListTask.flatMap { it.pathPrefixFile })
+    }
+
+    val archiveOperations = project.serviceOf<ArchiveOperations>()
+    // We use the resolvable configuration, but leverage withVariantReselection to obtain the subset of sources
+    from({ distributionClasspath.map { configuration ->
```
ov7a: 💅 consider extracting `from` argument into a separate var/function(s) for readability

build-logic/packaging/src/main/kotlin/gradlebuild.public-api-jar.gradle.kts
```diff
@@ -81,17 +82,111 @@ val task = tasks.register<Jar>("jarGradleApi") {
...
+            if (parent == null) false
+            else pathPrefixes.get().contains(parent.pathString)
+        }
+    }
+
+    inputs.files(pathPrefixListTask.flatMap { it.pathPrefixFile }).withPropertyName("pathPrefixList").withPathSensitivity(PathSensitivity.RELATIVE)
+    archiveClassifier.set("sources")
+    destinationDirectory = layout.buildDirectory.dir("public-api/gradle-api")
+    // This is needed because of the duplicate package-info.class files
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
```
ov7a, line 119: 💅 Can we narrow the duplicates resolution to handle only `package-info.class`?

build-logic/packaging/src/main/kotlin/gradlebuild.public-api-jar.gradle.kts
ljacomet: Good point, I skipped doing the `toSet` because I already sort and remove duplicates when creating that file.

build-logic/packaging/src/main/kotlin/gradlebuild.public-api-jar.gradle.kts
ljacomet: Indeed, switched to `NONE`

build-logic/packaging/src/main/kotlin/gradlebuild.public-api-jar.gradle.kts
ljacomet: I don't think that's a `CopySpec` feature ...

build-logic/packaging/src/main/kotlin/gradlebuild.public-api-jar.gradle.kts
mlopatkin: >Good point, I skipped doing the toSet because I already sort and remove duplicates when creating that file.

Hmm, it is still a `List` in the latest code. Did you decide to switch back or was it overlooked while moving stuff around?

ljacomet: Sorry, should have been clearer in my answer.

I followed the advice about doing `get` sooner as it made sense. But building a `Set` did not feel needed ... but while typing this I realize this is about `contains` so a `Set` is much better. Will fix as well.

cobexer (approved):

ljacomet:
@bot-gradle test RfN

ljacomet:
@bot-gradle test RfN

Issue #14880, opened 2020-10-15
Direct dependencies should be available in a configuration-cache-friendly way


autonomousapps:
`project.configuration.dependencies` does not seem to be compatible with the configuration cache (and doesn't play very nicely with UP-TO-DATE checks either), but I'm not aware of any other means to get those declared dependencies (vs resolved dependency graph).

### Expected Behavior

Consider this block of pseudo-code
```kotlin
val tuples: List<Tuple> = project.configurations.flatMap { conf ->
  conf.dependencies.toIdentifiers() // takes a DependencySet and converts it to a Set<String>
    .map { identifier ->
      Tuple(conf, identifier)
    }
}
```
I want to be able to create such `Tuple`s through some other mechanism that is serializable and compatible with the configuration cache.

### Current Behavior

I can do the above, but it is one reason my plugin is not compatible with the configuration cache. I've also had issues getting it to be invalidated when a dependency changes (meaning it's considered UP-TO-DATE even when the dependencies change), so I mark it `outputs.upToDateWhen { false }`, which is frustrating.

### Context

The task in question is [here](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/tasks/LocateDependenciesTask.kt), with its implementation delegated to [this class](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/internal/ConfigurationsToDependenciesTransformer.kt#L35).

The reason I need to do this is that my plugin provides very specific advice to users on how they should declare their dependencies. It might advise them to change a dep from `implementation` to `api`, or vice-versa. To do this, it needs to know the configuration on which a dependency has been declared (rather than knowing the dep is on the `compileClasspath`, for example). I have this in one single task to limit the scope of the damage. Several other tasks depend on this information, so only this one task does this particular thing and then writes the result to disk as a json, for consumption by the others.

ljacomet:
One trick you could do is create additional configurations that `extendsFrom` the different resolvable configurations, mark it as `transitive = false`, and use its resolution result as an input.
That should allow you to resolve the UP-TO-DATE aspect.

aSemy:
Something that tripped me up while trying this workaround: I thought it would be a good idea to use a detached configuration (since it shouldn't be used directly), but detached configurations can't extend other configurations https://github.com/gradle/gradle/issues/6881

PR #36643, opened 2026-02-06
Add CC support for FileCollectionDependency

tresat:


platforms/core-configuration/dependency-management-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/dm/FileCollectionDependencyCodec.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.gradle.internal.serialize.graph.WriteContext
+import org.gradle.internal.serialize.graph.readNonNull
+
+/**
+ * Codec for [DefaultFileCollectionDependency].
+ * <p>
+ * Delegates to [FileCollectionCodec] for the serialization of the file collection.
+ */
+class FileCollectionDependencyCodec(
+    private val fileCollectionCodec: FileCollectionCodec
```
tresat, line 34: I'm unsure when in inject another codec in the constructor of a codec, and when to just call `write(value.files)` and let the serialization machinery figure things out.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/DependencyNotationIntegrationSpec.groovy
```diff
@@ -23,7 +23,7 @@ import spock.lang.Issue
 
 class DependencyNotationIntegrationSpec extends AbstractIntegrationSpec {
 
-    @ToBeFixedForConfigurationCache(because = "Task uses the Configuration API")
+    @ToBeFixedForConfigurationCache(because = "DependencySet (conf.incoming.dependencies or conf.dependencies) is not CC compatible")
```
tresat, line 26: This test does NOT do resolution.  It instead tests the deps that are added to a configuration.

It was failing because of `gradleStuff gradleApi()` on line 50, which added a `FileCollectionDependency` - which was not serializable - to a configuration.

Since we're not resolving, you'd think we'd be okay just accessing the `.dependencies` collections to the task action, but this doesn't work since we can't reference `configurations`.

We also can't do:

```
task checkDeps {
    def deps = configurations.conf.incoming.dependencies
    def deps2 = configurations.allowsCollections.dependencies

    doLast {
        assert deps.contains(someDependency)
        // ...
    }
```

Since then we're trying to serialize the unserializable `DependencySet` type.

And we can't do 

task checkDeps {
    def conf1 = configurations.conf
    def conf2 = configurations.allowsCollections

    doLast {
		def deps = conf1.incoming.dependencies
		def deps2 = conf2.dependencies

        assert deps.contains(someDependency)
        // ...
    }
```

Since then we're trying to serialize the actual configuration, not its files, which doesn't work.

We can't do this:

```
task checkDeps {
    def files1 = configurations.conf.files
    def files2 = configurations.allowsCollections.files

    doLast {
		// NO dependencies to be found
    }
```

since we're testing dependencies, not serialization.

I think the best solution might be to just move everything to the task's configuration phase, since we're not actually doing resolution like:

```
task checkDeps {
    def deps = configurations.conf.incoming.dependencies
    def deps2 = configurations.allowsCollections.dependencies

    assert deps.contains(someDependency)
    assert deps.find { it instanceof ExternalDependency && it.group == 'org.mockito' && it.name == 'mockito-core' && it.version == '1.8'  }
    assert deps.find { it instanceof ExternalDependency && it.group == 'org.spockframework' && it.name == 'spock-core' && it.version == '1.0'  }
    assert deps.find { it instanceof ExternalDependency && it.group == 'junit' && it.name == 'junit' && it.version == '4.12' }
    def configuredDep = deps.find { it instanceof ExternalDependency && it.group == 'org.test' && it.name == 'configured' }
    assert configuredDep.version == '1.1'
    assert configuredDep.transitive == false

    assert deps2.size() == 2
    assert deps2.find { it instanceof ExternalDependency && it.group == 'org.mockito' }
    assert deps2.contains(someDependency)
}
```

Which doesn't actually require the new codec added in this PR 😄.

tresat:
Related:
- https://github.com/gradle/gradle/issues/14880

PR #36661, opened 2026-02-09
Fix CC-incompatible tests for plugin devel

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/extensibility/plugin-development/src/integTest/groovy/org/gradle/plugin/devel/impldeps/GradleImplDepsPerformanceIntegrationTest.groovy
+++ b/platforms/extensibility/plugin-development/src/integTest/groovy/org/gradle/plugin/devel/impldeps/GradleImplDepsPerformanceIntegrationTest.groovy
@@ -16,63 +16,45 @@
 
 package org.gradle.plugin.devel.impldeps
 
+
+import org.gradle.integtests.fixtures.BuildOperationsFixture
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.IntegTestPreconditions
-import org.gradle.testing.internal.util.RetryUtil
 
 @Requires(IntegTestPreconditions.NotEmbeddedExecutor) // Gradle API and TestKit JARs are not generated when running embedded
 class GradleImplDepsPerformanceIntegrationTest extends BaseGradleImplDepsIntegrationTest {
 
-    def "Gradle API JAR is generated in an acceptable time frame"() {
-        buildFile << """
+    def operations = new BuildOperationsFixture(executer, temporaryFolder)
+
+    def "#dependency JAR is generated in an acceptable time frame"() {
+        buildFile """
             configurations {
                 deps
             }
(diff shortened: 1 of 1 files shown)


platforms/extensibility/plugin-development/src/integTest/groovy/org/gradle/plugin/devel/impldeps/GradleImplDepsPerformanceIntegrationTest.groovy
```diff
@@ -67,9 +64,10 @@ class GradleImplDepsPerformanceIntegrationTest extends BaseGradleImplDepsIntegra
     static String resolveDependencies(long maxMillis) {
         """
             task resolveDependencies {
+                def depsConf = configurations.named("deps").map { it.resolve() }
                 doLast {
                     def timeStart = new Date()
-                    configurations.deps.resolve()
+                    depsConf.get()
```
mlopatkin: ❓ Doesn't it make the test kind of useless with CC?

mlopatkin: I think the proper solution would be to use build operations. We have one for resolving testkit already:
https://github.com/gradle/gradle/blob/18411b5dfe7d73aab9c1890043d0fe66db16a159/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/runtimeshaded/RuntimeShadedJarFactory.java#L66-L85

It doesn't cover full scope of resolving the configuration, but I think it should be good enough.

ov7a: Done, @mlopatkin PTAL

platforms/extensibility/plugin-development/src/integTest/groovy/org/gradle/plugin/devel/impldeps/GradleImplDepsPerformanceIntegrationTest.groovy
```diff
@@ -16,65 +16,46 @@
...
+                    println("Resolved dependencies: " + depsConf.get().size())
                 }
             }
         """
+
+        when:
+        succeeds("resolveDependencies")
+
+        then:
+        def resolve = operations.all(ResolveConfigurationDependenciesBuildOperationType)
```
mlopatkin: 🤔 Maybe also check that the Jar build operation has happened inside the dependency resolution, so if anything changes (e.g. we stop building the jar at runtime), we'd know that the test becomes useless.

mlopatkin (approved):
Looks good, I only have a small optional suggestion.

PR #36662, opened 2026-02-09
Access-track values() for env map and properties

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -63,8 +63,15 @@ requested plugin cannot be resolved. This behavior now matches that of precompil
 
 Avoid invalid plugin requests in unused precompiled settings plugins.
 
+==== Calling `.values()` on environment variables or system properties now tracked by Configuration Cache
+
+Gradle <<configuration_cache_requirements.adoc#config_cache:requirements:reading_sys_props_and_env_vars, tracks>> accesses to environment variables and system properties during configuration time to record them as inputs to the Configuration Cache, so the cache entry is invalidated when those variables change.
+
+Starting with Gradle 9.5.0, calling `.values()` on the maps returned from `System.getenv()` or `System.getProperties()` records **all** environment variables or system properties respectively as inputs to the Configuration Cache, similarly to how `.forEach()` calls are tracked in previous versions.
+If your build configuration uses `System.getenv().values()` or `System.getProperties().values()`, you may experience more frequent Configuration Cache invalidations.
+
 [[changes_9.4.0]]
-== Upgrading from 9.3.0 and earlier
+== Upgrading from 9.3.1 and earlier
 
 === Potential breaking changes
 


platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java
```diff
@@ -146,6 +146,7 @@ public Set<Object> keySet() {
 
     @Override
     public Collection<Object> values() {
+        reportAggregatingAccess();
```
ov7a, line 149: I wonder if it's all it takes to address the issue? 
If yes, then it is puzzling why it wasn't done initially.
If not, I wonder how methods like `forEach` should be updated?

mlopatkin: > I wonder if it's all it takes to address the issue? If yes, then it is puzzling why it wasn't done initially. If not, I wonder how methods like `forEach` should be updated?

I don't recall the exact reasoning at this point. Perhaps I was reluctant to add too many inputs or something like that and tried to come up with some "clever" way to track values that never happened.

ov7a: Should we just merge it? I mean, it's sorta uniform with what we currently have, and I don't see big issues with it.

mlopatkin: > Should we just merge it? I mean, it's sorta uniform with what we currently have, and I don't see big issues with it.

I'd ask to add an upgrade guide entry for that then. Historically, we sometimes added an opt-out switch for this input, but this might be an overkill here.

ov7a: Added, PTAL

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -36,8 +36,19 @@ The previous step will help you identify potential problems by issuing deprecati
...
+== Upgrading from 9.4.0 and earlier
+
+=== Potential breaking changes
+
+==== Environment variable and system property tracking with Configuration Cache
+
+Gradle <<configuration_cache_requirements.adoc#config_cache:requirements:reading_sys_props_and_env_vars, tracks>> accesses to environment variables and system properties during configuration time via `System.gevEnv()`.
+Some access patterns that potentially enumerate all environment variables or system properties (for example, calling `System.getenv().forEach()`) were already discouraged, and Gradle captures all variables accessed in this way to ensure the Configuration Cache is invalidated when they change.
+Starting with Gradle 9.5.0, `.values()` calls to system properties and environment variables are now also tracked this way, similarly to how `.forEach()` call is tracked in previous versions.
+
```
mlopatkin: 💅 I usually defer to @gradle/bt-docs-reviewers when writing, but here is my take on improving clarity
```suggestion
Gradle <<configuration_cache_requirements.adoc#config_cache:requirements:reading_sys_props_and_env_vars, tracks>> accesses to environment variables and system properties during configuration time to record them as inputs to the Configuration Cache, so the cache entry is invalidated when those variables change. Some access patterns that potentially enumerate all environment variables or system properties (for example, calling `System.getenv().forEach()`) were already discouraged, because Gradle has to record all environment variables to prevent false cache hits.

Starting with Gradle 9.5.0, calling `.values()` on the maps returned from `System.getenv()` or `System.getProperties()` records **all** environment variables or system properties respectively as inputs to the Configuration Cache, similarly to how `.forEach()` calls are tracked in previous versions.
```

mlopatkin (approved):
LGTM, though Claude and I did a pass on trying to improve the upgrade guide a bit.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -36,8 +36,19 @@ The previous step will help you identify potential problems by issuing deprecati
 . Run `gradle wrapper --gradle-version {gradleVersion}` to update the project to {gradleVersion}.
 . Try to run the project and debug any errors using the <<troubleshooting.adoc#troubleshooting, Troubleshooting Guide>>.
 
+[[changes_9.5.0]]
+== Upgrading from 9.4.0 and earlier
+
+=== Potential breaking changes
+
+==== Environment variable and system property tracking with Configuration Cache
```
```suggestion
==== Calling `.values()` on environment variables or system properties now tracked by Configuration Cache
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
lkasso: I had trouble understanding the current explanation, what do you think of:

````
Previously, calling `.values()` on the maps returned by `System.getenv()` or `System.getProperties()` was not tracked by the Configuration Cache. 
Starting with Gradle 9.5.0, these calls now cause all environment variables or system properties to be recorded as Configuration Cache inputs, similar to how `.forEach()` calls are already tracked.

If your build configuration uses `System.getenv().values()` or `System.getProperties().values()`, you may experience more frequent Configuration Cache invalidations. 

Any change to any environment variable or system property will now invalidate your cache, even if your build doesn't actually use that variable.

```kotlin
// This will now track ALL environment variables
val allValues = System.getenv().values()

// This will now track ALL system properties  
val allProps = System.getProperties().values()
```
````

lkasso (commented):
I was having trouble understanding the upgrade guide entry. I made a large suggestion. WDYT?

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
ov7a: I like the title from Laura, and I think the suggestions from Mike are great in terms of explaining technical details.

After reflecting a bit, I decided to remove the second sentence (because we have these details in the guide), preserve technical details from Mike, and add a brief explanation of the direct consequences from Laura. I think that this way it has enough details and pointers.

lkasso (approved):

PR #36665, opened 2026-02-09
Clarify default behavior of new tooling-parallel option

alllex:


platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -416,6 +416,9 @@ org.gradle.tooling.parallel=true
 org.gradle.parallel=false
 ```
 
+When `org.gradle.tooling.parallel` is not specified, it defaults to the value of `org.gradle.parallel`, preserving the existing performance expectations.
```
lkasso:
```suggestion
When `org.gradle.tooling.parallel` is not specified, it defaults to the value of `org.gradle.parallel`, preserving existing behavior and performance characteristics.
```

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -416,6 +416,9 @@ org.gradle.tooling.parallel=true
...
+See [userguide](userguide/performance.html#sec:configure_tooling_api_actions_parallelism) for additional details.
```
lkasso:
```suggestion
For more information, see the [Tooling API parallelism configuration](userguide/performance.html#sec:configure_tooling_api_actions_parallelism) section of the user guide.
```

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -403,7 +403,7 @@ Gradle now provides granular control over how [Tooling API](userguide/tooling_ap
 Previously, parallelism for Tooling API actions was tied directly to the `org.gradle.parallel` property.
 This meant that if you wanted to enable parallel task execution, you were forced to also enable parallel IDE actions, and vice versa.
 
-The new property decouples these two behaviors.
+The new property allows decoupling these two aspects.
```
lkasso:
```suggestion
The new property decouples these two behaviors.
```

alllex: I intentionally changed that to highlight that the property by itself doesn't decouple as in "remove connection between" these two things, but rather it gives the user the opportunity to do so. If the original wording already implied that it's okay, but for a non-native speaker like myself, this reads not precisely enough.

PR #36669, opened 2026-02-09
Update Documentation: update example for precompiled settings script note in upgrade guide

lkasso:


platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -184,14 +184,14 @@ plugins {
 [[deprecate_version_in_precompiled_settings_script_plugins]]
 ==== Deprecation of `version` in precompiled Settings script plugins
 
-The use of `version` within <<implementing_gradle_plugins_precompiled.adoc#implementing_precompiled_plugins, precompiled script plugins>> is now deprecated and will become an error in Gradle 10.0.0.
+The use of `version` within <<implementing_gradle_plugins_precompiled.adoc#settings_pre_compiled_plugins, precompiled Settings script plugins>> is now deprecated and will become an error in Gradle 10.0.0.
 
 In precompiled Settings scripts, `version` has no effect.
```
octylFractal: It's likely that some people would read this as "[Only] in precompiled Settings scripts" which isn't true. WDYT?
```suggestion
In precompiled scripts, `version` has no effect.
```

Issue #35318, opened 2025-10-15
ArtifactCollection can't be deserialised from Configuration Cache when it has failures with typed attributes


antohaby:
### Current Behavior

This error occurs when state is loaded from configuration cache.
On **cold** Gradle Daemon.

```
Caused by: org.gradle.api.GradleException: Could not load the value of field `requested` of `org.gradle.internal.component.resolution.failure.ResolutionCandidateAssessor$AssessedAttribute` bean found in field `onlyOnRequest` of `org.gradle.internal.component.resolution.failure.ResolutionCandidateAssessor$AssessedCandidate` bean found in field `candidates` of `org.gradle.internal.component.resolution.failure.type.AmbiguousArtifactsFailure` bean found in field `failure` of `org.gradle.internal.component.resolution.failure.exception.ArtifactSelectionException` bean found in field `artifactCollection` of `org.jetbrains.kotlin.gradle.utils.LazyResolvedConfigurationWithArtifacts` bean found in field `resolvedDependenciesConfiguration` of `org.jetbrains.kotlin.gradle.plugin.mpp.SourceSetVisibilityProvider$PlatformCompilationData` bean found in field `dependingPlatformCompilations` of `org.jetbrains.kotlin.gradle.plugin.mpp.GranularMetadataTransformation$Params` bean found in field `transformationParameters` of task `:composeApp:transformCommonMainDependenciesMetadata` of type `org.jetbrains.kotlin.gradle.plugin.mpp.MetadataDependencyTransformationTask`.
        at org.gradle.internal.serialize.graph.BeanPropertyExtensionsKt.readPropertyValue(BeanPropertyExtensions.kt:70)
        at org.gradle.internal.serialize.beans.services.BeanPropertyReader.readFieldOf(BeanPropertyReader.kt:93)
        at org.gradle.internal.serialize.beans.services.BeanPropertyReader.readFieldsOf(BeanPropertyReader.kt:73)
        at org.gradle.internal.serialize.beans.services.BeanPropertyReader.readStateOf(BeanPropertyReader.kt:66)
        at org.gradle.internal.serialize.graph.codecs.BeanCodec.readBeanOf(BeanCodec.kt:61)
        at org.gradle.internal.serialize.graph.codecs.BeanCodec.decode(BeanCodec.kt:43)
        at org.gradle.internal.serialize.graph.CombinatorsKt$reentrant$1.decode(Combinators.kt:80)
[...]

PR #36680, opened 2026-02-10
Store Assessed attribute values as Strings to avoid potential for ClassNotFoundExceptions on CC deserialization

tresat:
There is no need to store Attribute values at strongly typed instances when a failure has occurred - we can just get their String value.  This is all that is needed by the failure handling machinery, and it avoids a potential failure case where the value's class is not available for CC deserialization.

Fixes:
- https://github.com/gradle/gradle/issues/35318


tresat:
@bot-gradle test this

Issue #32591, opened 2025-03-03
Directory, file properties lose updates at execution time with CC


alllex:
Currently, Gradle does not preserve the identity of `DirectoryProperty` and `RegularFileProperty` when encoding with CC.

This means that if these properties are added to some file collections but then modified at execution time, the file collections won’t see the update, since after decoding those would be independent objects.
This is a discrepancy compared to how vintage behaves: it trivially keeps the connections, because those stay the same objects.

There is already a to-be-fixed test for `DirectoryProperty` that reproduces the problem:
https://github.com/gradle/gradle/blob/02702425e427dd1093cdec78145dc53194f4b08a/platforms/core-configuration/file-collections/src/integTest/groovy/org/gradle/api/file/RelativePathFilesIntegrationTest.groovy#L100

mlopatkin:
The issue is in the backlog of the relevant team and is prioritized by them.

***

We have several options to solve it:
- disallow modifications for things loaded from CC
- preserve identities for properties

The latter won't solve cross-task sharing, because we want isolation for parallel execution.

PR #36689, opened 2026-02-11
Mark (some) CC incompatible tests

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/CollectionPropertyIntegrationTest.groovy
+++ b/platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/CollectionPropertyIntegrationTest.groovy
@@ -17,8 +17,7 @@
 package org.gradle.api.provider
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
-import org.gradle.test.precondition.Requires
-import org.gradle.test.preconditions.IntegTestPreconditions
+import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
 
 class CollectionPropertyIntegrationTest extends AbstractIntegrationSpec {
     def setup() {
@@ -205,7 +204,7 @@ afterEvaluate {
         outputContains("value: value 1, value 2, value 3")
     }
 
-    @Requires(value = IntegTestPreconditions.NotConfigCached, reason = "https://github.com/gradle/gradle/issues/25516")
+    @ToBeFixedForConfigurationCache(because = "https://github.com/gradle/gradle/issues/36664")
     def "task ad hoc input property is implicitly finalized when task starts execution"() {
         given:
         buildFile """
@@ -269,7 +268,6 @@ task thing {
         "providers.provider { [ 'a', 'b', 'c' ] }" | _
     }
 
-    @Requires(value = IntegTestPreconditions.NotConfigCached, reason = "https://github.com/gradle/gradle/issues/27528")
     def "can set value for string list property using GString values"() {
         buildFile """
             def str = "aBc"
@@ -303,7 +301,6 @@ task thing {
         succeeds("verify")
     }
 
-    @Requires(value = IntegTestPreconditions.NotConfigCached, reason = "https://github.com/gradle/gradle/issues/27528")
     def "can add element to string list property using GString value"() {
         buildFile << """
             def str = "aBc"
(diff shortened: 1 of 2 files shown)


platforms/core-configuration/file-collections/src/integTest/groovy/org/gradle/api/file/FilePropertyLifecycleIntegrationTest.groovy
```diff
@@ -19,6 +19,7 @@ package org.gradle.api.file
 import org.gradle.api.tasks.TasksWithInputsAndOutputs
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
```
ov7a, line 21: I'll work more on this file later

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/CollectionPropertyIntegrationTest.groovy
```diff
@@ -269,7 +268,6 @@ task thing {
         "providers.provider { [ 'a', 'b', 'c' ] }" | _
     }
 
-    @Requires(value = IntegTestPreconditions.NotConfigCached, reason = "https://github.com/gradle/gradle/issues/27528")
```
ov7a, line 272: This and following tests pass with CC

testing/internal-integ-testing/src/integTest/groovy/org/gradle/integtests/fixtures/executer/InProcessGradleExecuterIntegrationTest.groovy
```diff
@@ -34,7 +33,6 @@ class InProcessGradleExecuterIntegrationTest extends Specification {
     def distribution = new UnderDevelopmentGradleDistribution(IntegrationTestBuildContext.INSTANCE)
     def executer = new GradleContextualExecuter(distribution, temporaryFolder, IntegrationTestBuildContext.INSTANCE)
 
-    @ToBeFixedForConfigurationCache
```
ov7a, line 37: The whole class is to test EmbeddedExecutor itself and has class-level `@Requires`

mlopatkin:
I don't think we should mark broken stuff we intend to fix as `UnsupportedWithCC`. This should be reserved for tests that we don't intend to fix due to their inherent incompatibilities with CC.

ov7a:
@mlopatkin Got it. I've updated the annotations. Note that for smoke test plugins - it's not our responsibility, so I think `UnsupportedWithCC` is fine there

mlopatkin (approved):

platforms/core-configuration/file-collections/src/integTest/groovy/org/gradle/api/file/FilePropertyLifecycleIntegrationTest.groovy
```diff
@@ -631,8 +636,12 @@ task thing {
         expect:
         succeeds("after")
         outputContains("get failed: Cannot query the value of extension 'thing' property 'prop' because configuration of root project 'broken' has not completed yet.")
-        outputContains("get from task failed: Failed to calculate the value of extension 'thing' property 'prop'.")
-        outputContains("get from task failed cause: Cannot query the value of task ':producer' property 'output' because task ':producer' has not completed yet.")
+        if (GradleContextualExecuter.notConfigCache) {
+            outputContains("get from task failed: Failed to calculate the value of extension 'thing' property 'prop'.")
+            outputContains("get from task failed cause: Cannot query the value of task ':producer' property 'output' because task ':producer' has not completed yet.")
+        } else {
+            outputContains("prop is accessible in 'before' task")
```
ov7a: Assuming that's expected given the lazy evaluation

mlopatkin: > Assuming that's expected given the lazy evaluation

:x: No, this feels wrong. The issue has been closed but we only delivered part of it, and not the part that is covered by the test.

I've filed https://github.com/gradle/gradle/issues/36710 for this. I suspect that many other tests in the file are also failing because of it rather than #32591.

mlopatkin (changes requested):
We're probably facing another root cause here.

platforms/core-configuration/file-collections/src/integTest/groovy/org/gradle/api/file/FilePropertyLifecycleIntegrationTest.groovy
ov7a: Ok, I'll revert this change and will mark it as blocked by that issue

PR #36692, opened 2026-02-11
Add project() method to DependencyHandler for consistency with Dependencies

### LINKED ISSUE #36668: Inconsistency between `Dependencies` and `DependencyHandler`
### Current Behavior

Both `Dependencies` and `DependencyHandler` expose similar interfaces for declaring dependencies. However, `Dependencies` exposes both `project(String)` and `project()` for creating dependencies on other projects and on the current project, respectively, while `DependencyHandler` has only `project(Map)`.  

This means that each behave in the following ways when declaring a project dependency:

| DSL | Dependencies | DependencyHandler |
| --- | -------------- | --------------------- |
| implementation(project) | does not work | works  |
implementation(project()) | works| does not work |
implementation(project(project.path)) | may not work in 9.5 if compiled with `-Werror` | works (delegates to `Project.getProject()`) |
implementation(project(':foo')) | works | works |
implementation(path: ':foo')) | does not work | works |

In 9.5, if the build script is compiled with `-Werror` then there is no way to specify a dependency on the current project consistently.  In other words, in `Dependencies` you must use `project()` and cannot use `project(project.path)`, while for `DependencyHandler` you must use `project` or `project(project.path)` but not `project()`.

Prior to 9.5, you could use `project(project.path)` in either place, but in the case of `Dependencies` the `project` in `project.path` actually comes from `Dependencies.getProject()` which is an injected getter intended for internal use and, in 9.5, has been marked with `@Deprecated`.

If `-Werror` is _not_ used, this will still work when compiling build scripts, precompiled scripts, or plugins, similar to how it worked in previous versions.

In 9.5, you _can_ capture the project path outside of the `Dependencies` block and then reference it from inside:
```
val projectPath = project.path // uses `Project.getProject()`
dependencies {
    implementation(project(projectPath))
}
```
    

### Expected Behavior

There should be at least one consistent and idiomatic way to declare a dependency on the current project in both `Dependencies` and `DependencyHandler` so that users who set `-Werror` are not required to do something different depending on which block they're configuring.  Supporting `project()` in both places seems like a logical choice.

### Context (optional)

_No response_

```
plugins {
    id("java")
}

dependencies {
    implementation(project()) // does not work
    // implementation(project(project.path)) // works
}

testing {
    suites { 
        val test by getting(JvmTestSuite::class) { 
            useJUnitJupiter("5.7.1") 
        }

        val integrationTest by registering(JvmTestSuite::class) { 
            dependencies {
                implementation(project()) // works
                // implementation(project(project.path)) // does not work with -Werror
            }
        }
    }
}
```

### Gradle version

9.5.0-rc-1

### Build scan URL (optional)

_No response_

_No response_

ghale:
The `project()` method allows creating dependencies on the current project.  

This PR also includes some verbiage in the upgrade guide about the deprecation of  `Dependencies.getProject()` from #36526.  Adding `project()` provides a consistent way to specify a dependency on the current project that will work with both `DependencyHandler` and `Dependencies`. 

Fixes #36668

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -243,7 +243,6 @@ testing {
 }
 ----
 
-The `project()` method can be used to specify a dependency on the current project in both test suites and in the top-level `dependencies {}` block.
 
 [[changes_9.3.0]]
 == Upgrading from 9.2.0 and earlier


ghale:
@bot-gradle test this

subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/DependencyHandler.java
```diff
@@ -330,6 +331,16 @@ public interface DependencyHandler extends ExtensionAware {
...
 
+    /**
+     * Creates a dependency on the current project.
+     *
+     * @return The dependency.
+     *
+     * @since 9.5.0
+     */
+    @Incubating
+    Dependency project();
```
jvandort, line 342: I wonder why this doesn't return a `ProjectDependency` (and why the previous `project(Map<String, ?> notation);` does not)

jvandort, line 342: Now I understand I can declare a dependency on the current project using
```
dependencies {
    implementation(project()) // New
    implementation(project) // Existing
}
```

Since there was an existing way to do this without deprecation, do we plan to eventually deprecate that? Or are we okay have two ways of doing this

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -229,6 +202,46 @@ plugins {
 }
 ----
 
+==== Deprecation of `Dependencies.getProject()` method
+
+The `getProject()` method on `Dependencies` has been deprecated and will be removed in Gradle 10.0.0.
+`Dependencies` is used to configure dependencies in test suites, and `getProject()` was being used by internal Gradle code to retrieve a reference to the current project.
```
```suggestion

`Dependencies` is used to configure dependencies in test suites.
Previously, `getProject()` was used by Gradle internally to reference the current project.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -229,6 +202,46 @@ plugins {
...
+This method was not intended for public use, but it's possible that it could be used unintentionally.
```
```suggestion
While never intended for public use, it could be used unintentionally in build scripts.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -229,6 +202,46 @@ plugins {
...
+
+For example, a configuration like the following may accidentally access this method:
```
```suggestion
For example, the following configuration accidentally invokes the deprecated method because `project.path` resolves against the `Dependencies` object rather than the top-level `Project` object:
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -229,6 +202,46 @@ plugins {
...
+        val integrationTest by registering(JvmTestSuite::class) {
+            dependencies {
+                implementation(project(project.path)) // `project.path` accesses `Dependencies.getProject()`
+            }
+        }
+    }
+}
+----
+
+The configuration above may now fail to compile with an error in 9.5.0 if `-Werror` is set.
```
lkasso, line 228:
```suggestion
NOTE: If you use the `-Werror` flag, this configuration will fail to compile in Gradle 9.5+.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -229,6 +202,46 @@ plugins {
...
+
+Instead, you can specify a dependency on the current project using the `project()` method with no arguments:
```
```suggestion
To depend on the current project, use the `project()` method without arguments. 
This syntax is the idiomatic way to reference the project in both test suites and standard `dependencies {}` blocks:
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -229,6 +202,46 @@ plugins {
...
+        val integrationTest by registering(JvmTestSuite::class) {
+            dependencies {
+                implementation(project())
+            }
+        }
+    }
+}
+----
+
+The `project()` method can be used to specify a dependency on the current project in both test suites and in the top-level `dependencies {}` block.
```
```suggestion
```

lkasso (commented):
I left some suggestions because the upgrade guide was a little hard to understand. I'm not sure if what I wrote is 100% accurate, but hopefully you can use it to clarify.

subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/DependencyHandler.java
ghale: I don't know.  None of the dependency-producing methods on this API return anything other than `Dependency`.  It's possible that the subtypes were not public at some point in the past (but that's pure speculation).  I considered making this `ProjectDependency`, but it felt wrong to do something inconsistent with the other methods here.

ghale: I think deprecating using `Project` here is the right thing to do.  I've created #36801 to handle this.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
ghale: I'm not sure about this change.  Prefacing it with "NOTE" makes it seem like this is some marginally important detail, when this is, in fact, the crucial reason why any of this would be important to a user.  For the vast majority of users, they won't notice this change at all (because the method is still there, it's just deprecated, and `-Werror` is not the default).  However, if you're compiling your scripts or plugins with `-Werror` you'll now get a compilation failure because the method has been marked `@Deprecated`.  The most common case why you would somehow use the `getProject()` method here is to create a dependency on the current project, although this may not be the only reason someone might be accidentally using it.

What we're trying to create is a bread crumb so that if users start getting a compilation failure here, they have something that would explain why.  And, in the most common case of creating a dependency on the current project, we want to describe how to resolve it in a way that works consistently with both test suites and the top-level dependencies block.

But the only reason it would matter to anyone right now is because they're using `-Werror`.

PR #36698, opened 2026-02-11
Break up project type/feature fixtures for manageability

ghale:
The project type/feature fixture classes were getting large and unwieldy due to the number of nested classes they contained.  This breaks up these classes such that the builders are pulled out into their own separate files and the fixtures only contain the methods that combine them into coherent sets of classes.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

PR #36708, opened 2026-02-12
Fix #36601: Set proper file permissions for results-generic.bin on POSIX systems

### LINKED ISSUE #36601: Gradle 9.3.x generates results-generic.bin with wrong permssions
### Current Behavior

We noticed when upgrading from 9.2.1 to 9.3.0 that gradle test generates `build/test-results/test/binary/results-generic.bin` with only user read+write permissions.  The other files generated have proper permissions.  The only file that does not have read "other" perms is `results-generic.bin`.  Its as if umask it not respected.

To reproduce:

```
./gradew test
```

```
❯ ls -l build/test-results/test/binary
total 6
-rw-rw-r--+ 1 nobody nobody   0 Feb  4 12:50 output-events.bin
-rw-------+ 1 nobody nobody 329 Feb  4 12:50 results-generic.bin
```

```
❯ find build -type f ! -perm -o=r
build/test-results/test/binary/results-generic.bin
```

This remains the same in gradle 9.3.1 as well, but first time we saw it was when jumping from 9.2.1 -> 9.3.0

### Expected Behavior

`results-generic.bin` should have similar permissions to the rest of the generated files.

### Context (optional)

`results-generic.bin` should have similar permissions as the rest of the generated files.  IE it should honor umask.

Presumably any gradle project will have this problem.  Here is a barebones spring-boot app that demonstrates the problem from https://start.spring.io

[demo.zip](https://github.com/user-attachments/files/25079832/demo.zip)

### Gradle version

9.3.1

### Build scan URL (optional)

_No response_

Linux

chanani:
Fixes #36601

Since Gradle 9.3.0, `results-generic.bin` is created with owner-only permissions (`-rw-------`) on POSIX systems, while other files in the same directory like `output-events.bin` correctly respect the system umask.

The cause was `SerializableTestResultStore.Writer` using `Files.createTempFile()` + `Files.move()`. `createTempFile()` defaults to 600 permissions on POSIX, and `move()` preserves them.

This PR replaces that pattern with a direct `Files.newOutputStream()` write, which naturally follows the system umask.

An integration test was added in `AbstractTestTaskIntegrationTest` to verify that all binary test result files have group-read and others-read permissions on POSIX systems.

### DIFF
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
@@ -113,7 +113,7 @@ public final class SerializableTestResultStore {
         private Writer(Path serializedResultsFile, Path outputEventsFile, int diskSkipLevels) throws IOException {
             this.diskSkipLevels = diskSkipLevels;
             this.serializedResultsFile = serializedResultsFile;
-            this.temporaryResultsFile = serializedResultsFile.getParent().resolve("in-progress-results-generic.bin");
+            this.temporaryResultsFile = serializedResultsFile.resolveSibling("in-progress-results-generic.bin");
             // Use constants to avoid allocating empty collections if flattening is not enabled
             flatteningIds = isDiskSkipEnabled() ? new HashSet<>() : Collections.emptySet();
             extraFlattenedDescriptors = isDiskSkipEnabled() ? new ArrayList<>() : Collections.emptyList();


chanani:
@octylFractal I've implemented the fix based on your suggestion in #36601 — replaced the `Files.createTempFile()` + `Files.move()` pattern with a direct `Files.newOutputStream()` write, and added an integration test in `AbstractTestTaskIntegrationTest` as you recommended.

Could you take a look when you get a chance? Thanks!

ov7a:

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/AbstractTestTaskIntegrationTest.groovy
```diff
@@ -409,4 +411,22 @@ abstract class AbstractTestTaskIntegrationTest extends AbstractTestingMultiVersi
     private static JavaVersion classFormat(TestFile path) {
         JavaVersion.forClassVersion(path.bytes[7] & 0xFF)
     }
+
+    @Requires(UnitTestPreconditions.FilePermissions)
+    def "binary test result files have correct permissions"() {
```
octylFractal: :x: Can we move this above the `private` methods at the bottom for consistency?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/AbstractTestTaskIntegrationTest.groovy
```diff
@@ -409,4 +411,22 @@ abstract class AbstractTestTaskIntegrationTest extends AbstractTestingMultiVersi
...
+        given:
+        file('src/test/java/MyTest.java') << standaloneTestClass
+
+        when:
+        succeeds 'test'
+
+        then:
+        def binaryDir = file("build/test-results/test/binary")
+        binaryDir.listFiles().each { File f ->
```
octylFractal: :x: I think we should validate this by walking the entire directory, not just the top-level children.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/AbstractTestTaskIntegrationTest.groovy
```diff
@@ -27,6 +27,8 @@ import org.gradle.testing.fixture.AbstractTestingMultiVersionIntegrationTest
 import spock.lang.Issue
 
 import java.time.Duration
+import java.nio.file.Files as NioFiles
```
octylFractal: :x: I think we prefer to leave in the fully-qualified name, as it's easier to read and search when outside the IDE.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/AbstractTestTaskIntegrationTest.groovy
```diff
@@ -409,4 +411,22 @@ abstract class AbstractTestTaskIntegrationTest extends AbstractTestingMultiVersi
...
+        file('src/test/java/MyTest.java') << standaloneTestClass
+
+        when:
+        succeeds 'test'
+
+        then:
+        def binaryDir = file("build/test-results/test/binary")
+        binaryDir.listFiles().each { File f ->
+            assert f.canRead()
+            def perms = NioFiles.getPosixFilePermissions(f.toPath())
```
octylFractal: ❓ Will this work on Windows? We may need to limit this test to only run on non-Windows OSes.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
```diff
@@ -111,15 +108,13 @@ private static int depth(TestDescriptorInternal descriptor) {
...
-            this.serializedResultsFile = serializedResultsFile;
             this.diskSkipLevels = diskSkipLevels;
             // Use constants to avoid allocating empty collections if flattening is not enabled
             flatteningIds = isDiskSkipEnabled() ? new HashSet<>() : Collections.emptySet();
             extraFlattenedDescriptors = isDiskSkipEnabled() ? new ArrayList<>() : Collections.emptyList();
             extraFlattenedResults = isDiskSkipEnabled() ? new ArrayList<>() : Collections.emptyList();
             Files.createDirectories(serializedResultsFile.getParent());
-            temporaryResultsFile = Files.createTempFile(serializedResultsFile.getParent(), "in-progress-results-generic", ".bin");
-            resultsEncoder = new KryoBackedEncoder(Files.newOutputStream(temporaryResultsFile));
+            resultsEncoder = new KryoBackedEncoder(Files.newOutputStream(serializedResultsFile));
```
octylFractal: :x: We should still use a separate file for the in-progress results so we don't leave a file that might be invalid in the results if we fail to close properly. This code should just make a `in-progress-results-generic.bin` file instead of using a temporary file. Then still move it to the final location in the `close()` call.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/AbstractTestTaskIntegrationTest.groovy
chanani: `UnitTestPreconditions.FilePermissions` already excludes Windows. Its `isSatisfied()` returns `true` only when `MacOs || Linux`:
```groovy
// UnitTestPreconditions.groovy
static final class FilePermissions implements TestPrecondition {
    @Override
    boolean isSatisfied() {
        return TestPrecondition.satisfied(MacOs) || TestPrecondition.satisfied(Linux)
    }
}
```

So `getPosixFilePermissions()` will never be called on Windows. No additional guard is needed.

chanani:
Thanks for the review @octylFractal ! All feedback addressed

- [x] Moved permission test above `private` methods.
- [x] Replaced `listFiles()` with `Files.walk()` for full directory traversal.
- [x] Removed `NioFiles` alias, using fully-qualified `java.nio.file.Files`.
- [ ] `UnitTestPreconditions.FilePermissions` already excludes Windows (`MacOs || Linux` only), so no additional guard needed.
- [x] Replaced `createTempFile()` with fixed-name `in-progress-results-generic.bin` via `newOutputStream()` — umask respected, temp file + move pattern preserved.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
```diff
@@ -111,14 +111,14 @@ private static int depth(TestDescriptorInternal descriptor) {
         private final Multimap<TestDescriptorInternal, TestMetadataEvent> metadatas = LinkedHashMultimap.create();
 
         private Writer(Path serializedResultsFile, Path outputEventsFile, int diskSkipLevels) throws IOException {
-            this.serializedResultsFile = serializedResultsFile;
             this.diskSkipLevels = diskSkipLevels;
+            this.serializedResultsFile = serializedResultsFile;
+            this.temporaryResultsFile = serializedResultsFile.getParent().resolve("in-progress-results-generic.bin");
```
octylFractal: 💭 Can we use the method `Path` provides for this?
```suggestion
            this.temporaryResultsFile = serializedResultsFile.resolveSibling("in-progress-results-generic.bin");
```

octylFractal (approved):

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/results/serializable/SerializableTestResultStore.java
chanani: Thank you for the suggestion! I've updated the code to use Path.resolveSibling() as you recommended. 
Please let me know if there are any further requests or if there is anything else I should address.

octylFractal:

chanani:
Hi @octylFractal, it looks like this PR was removed from the merge queue due to some failed TeamCity checks earlier. However, all 17 checks are passing now and there are no conflicts with the base branch.

Could you please re-add this to the merge queue when you have a chance? I believe the previous failure was just a transient CI issue rather than something caused by the changes in this PR. Thanks!

PR #36713, opened 2026-02-12
Fix CC-incompatible tests for build operations & refactor related fixtures

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/integTest/groovy/org/gradle/initialization/CalculateTaskGraphBuildOperationIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/initialization/CalculateTaskGraphBuildOperationIntegrationTest.groovy
@@ -35,20 +35,22 @@ class CalculateTaskGraphBuildOperationIntegrationTest extends AbstractIntegratio
     final operationNotificationsFixture = new BuildOperationNotificationsFixture(executer, temporaryFolder)
 
     def "requested and filtered tasks are exposed"() {
-        createDirs("a", "b", "a/c")
-        settingsFile << """
+        def subprojects = ["a", "b", "a/c"]
+        createDirs(*subprojects)
+        settingsFile """
             include "a"
             include "b"
             include "a:c"
         """
 
-        buildFile << """
-            allprojects {
+        (subprojects + ".").each { dir ->
+            buildFile "$dir/build.gradle", """
                 task otherTask
                 task someTask
                 someTask.dependsOn otherTask
-            }
-        """
+            """
+        }
+
         when:
         succeeds('help')
 
@@ -86,20 +88,22 @@ class CalculateTaskGraphBuildOperationIntegrationTest extends AbstractIntegratio
     }
 
     def "task plan is exposed"() {
-        createDirs("a", "b", "a/c")
-        settingsFile << """
+        def subprojects = ["a", "b", "a/c"]
+        createDirs(*subprojects)
+        settingsFile """
             include "a"
             include "b"
             include "a:c"
         """
 
-        buildFile << """
-            allprojects {
+        (subprojects + ".").each { dir ->
+            buildFile "$dir/build.gradle", """
                 task otherTask
                 task someTask
                 someTask.dependsOn otherTask
-            }
-        """
+            """
+        }
+
         when:
         succeeds('someTask')
 


subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/AbstractCompositeBuildIntegrationTest.groovy
```diff
@@ -221,4 +224,45 @@ public class ${className} implements Plugin<Project> {
...
+        List<String> extraBuildPathsForCC = null
+    ) {
+        treeTaskGraphOps.every { it.displayName == "Calculate build tree task graph" }
+        for (int i = 0; i < expectedParents.size(); i++) {
+            assert treeTaskGraphOps[i].parentId == expectedParents[i]
+        }
+        if (GradleContextualExecuter.notConfigCache) {
+            assert treeTaskGraphOps.size() == expectedParents.size()
+        } else {
+            assert treeTaskGraphOps.size() == expectedParents.size() + 1
```
ov7a: I assume it's expected for now to have an extra loading of the task tree with CC?

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/AbstractCompositeBuildIntegrationTest.groovy
```diff
@@ -221,4 +224,45 @@ public class ${className} implements Plugin<Project> {
         def output = result.output
         assert output.contains(string.trim())
     }
+
+    static void verifyBuildPathOperations(String prefix, List<BuildOperationRecord> operations, List<String> buildPaths, List<Long> parentIds) {
```
mlopatkin: 💅 You can use a Map to pass `buildPath: parentId` pairs rather than parallel lists. IMO, it will be a bit more readable.

Then you'll have something like:
```groovy
verifyBuildPathOperations(
    "Configure build",
    configureOps,
    [
      (":"): root.id,
      (":buildB"): configureOps[0].id,
      (":buildB:buildSrc"): buildSrcOps[0].id],
    ]
)
```

But I'm not sure that this method adds readability in general.

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildBuildSrcBuildOperationsIntegrationTest.groovy
```diff
@@ -69,18 +65,11 @@ class CompositeBuildBuildSrcBuildOperationsIntegrationTest extends AbstractCompo
...
-        loadOps[1].parentId == loadOps[0].id
-
-        loadOps[2].displayName == "Load build (:buildB:buildSrc)"
-        loadOps[2].details.buildPath == ":buildB:buildSrc"
-        loadOps[2].parentId == buildSrcOps[0].id
+        verifyBuildPathOperations(
+            "Load build",
+            loadOps, [":", ":buildB", ":buildB:buildSrc"],
+            [root.id, loadOps[0].id, buildSrcOps[0].id]
+        )
```
mlopatkin: :x: This is somewhat subjective, but I strongly feel that the tests are now much less readable.

`verifyBuildPathOperations` is very generic and conveys no meaning of what is being verified. I'd rather either keep those assertions as is, or add helper functions like:
```groovy
def verifyBuildOp(def op, def displayName, def buildPath, def parent) { /*... */ }
``` 
and rewrite the assertion into
```groovy
verifyBuildOp(loadOps[0], "Load build", ":",root)
verifyBuildOp(loadOps[1], "Load build  (:buildB)",":buildB",loadOps[0])
// etc.
```
Named parameters may give even more clarity to the callsite, but Groovy's handling of those isn't great. I see the greatest value of method vs. assert chain we had before in enforcing certain mandatory verifications, and with named parameters it can only be enforced at runtime. It may be good enough, though.

I'd add conditional verification in-place as well, rather than trying to pass these extra projects, it is easier to read than trying to keep the contract of the `verifyTaskGraphOps` in mind.

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildOperationsIntegrationTest.groovy
```diff
@@ -392,10 +333,8 @@ class CompositeBuildOperationsIntegrationTest extends AbstractCompositeBuildInte
         operations.only(FinishRootBuildTreeBuildOperationType)
     }
 
+    @UnsupportedWithConfigurationCache(because = "buildFinished", iterationMatchers = ".*with buildFinished.*")
```
mlopatkin, line 377: 🎉

mlopatkin (changes requested):
I'm sorry, but I feel the tests became too DRY to be readable.

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildBuildSrcBuildOperationsIntegrationTest.groovy
ov7a: My problem with the assert chain is that it's easy to make a copy-paste error, and it's easy to miss a size check. I agree that Groovy is not great with named parameters and that an (ordered) map is better than parallel arrays. I was not sure about if I should keep the CC-related condition inside or outside the method. I'll update the PR to iterate on what I definitely agree, but let's discuss it further after that.

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/AbstractCompositeBuildIntegrationTest.groovy
```diff
@@ -221,4 +224,41 @@ public class ${className} implements Plugin<Project> {
...
+        assert operations.size() == buildPathsToParentIds.size()
+        buildPathsToParentIds.eachWithIndex { bp, i ->
+            def (buildPath, parentId) = bp
+            assert operations[i].displayName == "$prefix${buildPath == ":" ? "" : " ($buildPath)"}"
+            assert operations[i].details.buildPath == buildPath
+            assert operations[i].parentId == parentId
+        }
+    }
+
+    void verifyTaskGraphOps(
```
ov7a: Very similar checks are performed in
https://github.com/gradle/gradle/blob/8e2ce465c6e34c507ab7027720aca8d064aa68ca/subprojects/core/src/integTest/groovy/org/gradle/initialization/CalculateTaskGraphBuildOperationIntegrationTest.groovy#L330

and

https://github.com/gradle/gradle/blob/f8e539729c3f62acd03ab2831e24bd60e5e2cd0c/subprojects/core/src/integTest/groovy/org/gradle/internal/operations/notify/BuildOperationNotificationIntegrationTest.groovy#L213

and

https://github.com/gradle/gradle/blob/7899bb831b4b8eb2bd5d029cc4a69bb91d5ad5d6/subprojects/core/src/integTest/groovy/org/gradle/initialization/buildsrc/BuildSrcBuildOperationsIntegrationTest.groovy#L79

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/BuildOperationTreeQueries.groovy
```diff
@@ -289,3 +289,63 @@ abstract class BuildOperationTreeQueries {
...
+    TestableBuildOperationRecord(BuildOperationRecord original) {
+        this.id = original.id
+        this.parentId = original.parentId
+        this.displayName = original.displayName
+        this.details = original.details
+    }
+
+    TestableBuildOperationRecord(
+        String displayName,
+        Long parentId,
```
mlopatkin, line 381: 💅 Maybe pass a  `BuildOperationRecord`/`TestableBuildOperationRecord` here? Then the callsite will be cleaner:
```groovy
buildOp(
	displayName: "Configure build (:buildSrc)", 
	parent: buildSrcOps[1], 
	details: ["buildPath": ":buildSrc"]
)
```

cobexer:
@ov7a Tests in `org.gradle.initialization.CalculateTaskGraphBuildOperationIntegrationTest` have errors, see https://ge.gradle.org/s/b3pqtgpvbleae/tests/overview?outcome=FAILED

ov7a:
@cobexer I wonder why it didn't show up in previous failures, see https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_GitHubMergeQueueCheckPass/110342558

cobexer:
It did show up: https://ge.gradle.org/scans/tests?search.names=gitCommitId,PartOf&search.tags=CI&search.timeZoneId=Europe%2FVienna&search.values=5ba77556003c9be1c9b24045c7f821cbfcc66d38,ReadyforNightly

but TeamCity did not show it because the build where it happened was not finished when it got cancelled ("luckily" it had already published the build scan).

ov7a:
@bot-gradle test AST

PR #36714, opened 2026-02-12
Expose failOnNoMatchingTests (or equivalent) on AbstractTestTask

ljacomet:
Add top-level `Property<Boolean>` on `AbstractTestTask` that
shadows `filter.failOnNoMatchingTests`. It is annotated with `@Option` to
expose `--fail-on-no-matching-tests` and `--no-fail-on-no-matching-tests`
CLI flags.

Add integration tests verifying the property can be set via the build
script DSL and via the CLI option. Tests also encode that the task
property has precedence of the filter property.

### DIFF
--- a/platforms/documentation/docs/src/docs/release/notes.md
+++ b/platforms/documentation/docs/src/docs/release/notes.md
@@ -83,6 +83,17 @@ This is useful for running Gradle in automated environments such as CI pipelines
 
 See the [Non-interactive mode](userguide/command_line_interface.html#sec:non_interactive) section in the Gradle User Manual for more information.
 
+### More control over test filtering on the command line
+
+When running `--tests` filters on a task like `test` that spans multiple subprojects, Gradle previously failed the build if the task found no matching tests.
+You can now pass `--allow-no-matches` alongside `--tests` to allow the build to succeed when a filter matches no tests on a given `Test` task:
+
+`./gradlew test --tests "*FooTest*" --allow-no-matches`
+
+This is especially useful in multi-project builds where a filter is only expected to match tests in some subprojects.
+
+The default behavior (fail when no tests match) is unchanged.
+
 ### Build authoring improvements
 Gradle provides [rich APIs](userguide/getting_started_dev.html) for build engineers and plugin authors, enabling the creation of custom, reusable build logic and better maintainability.
 
--- a/platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
@@ -123,6 +123,15 @@ For more details and examples of declaring filters in the build script, please s
 
 The command-line option is especially useful to execute a single test method. When you use `--tests`, be aware that the inclusions declared in the build script are still honored. It is also possible to supply multiple `--tests` options, all of whose patterns will take effect. The following sections have several examples of using the command-line option.
 
+[WARNING]
+====
+When filtering tests, Gradle will fail the build if the filter does not match any test by default.
+
+When running multiple `Test` tasks with a top-level task selector like `test` in a multi-project build, there is a high chance of causing the build to fail if you supply a `--tests` filter that does not match at least one test in every `Test` task.
+
+To change this behavior, pass the additional command-line option `--allow-no-matches` which will allow the build to succeed even if the filter does not match any test on any `Test` task.
+====
+
 NOTE: Not all test frameworks play well with filtering. Some advanced, synthetic tests may not be fully compatible.
 However, the vast majority of tests and use cases work perfectly well with Gradle's filtering mechanism.
 
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
@@ -65,6 +65,7 @@ import org.gradle.api.tasks.Optional;
 import org.gradle.api.tasks.OutputDirectory;
 import org.gradle.api.tasks.TaskAction;
 import org.gradle.api.tasks.VerificationTask;
+import org.gradle.api.tasks.options.NoDisable;
 import org.gradle.api.tasks.options.Option;
 import org.gradle.api.tasks.testing.logging.TestLogging;
 import org.gradle.api.tasks.testing.logging.TestLoggingContainer;
@@ -692,7 +693,7 @@ public abstract class AbstractTestTask extends ConventionTask implements Verific
     }
 
     private boolean shouldFailOnNoMatchingTests() {
-        return patternFiltersSpecified() && getFailOnNoMatchingTests().getOrElse(filter.isFailOnNoMatchingTests());
+        return patternFiltersSpecified() && getAllowNoMatchingTests().map(allow -> !allow).getOrElse(filter.isFailOnNoMatchingTests());
     }
 
     boolean testsAreNotFiltered() {
@@ -795,23 +796,27 @@ public abstract class AbstractTestTask extends ConventionTask implements Verific
     }
 
     /**
-     * Whether the task should fail when a filter was configured but no test matched the filter.
-     * Defaults to true.
+     * Whether to allow the task to succeed when a filter was configured but no tests matched.
+     * When not set, falls back to {@link TestFilter#isFailOnNoMatchingTests()} (default: fail).
      *
      * <p>
      * If configured this takes precedence over {@link TestFilter#isFailOnNoMatchingTests()}.
      * <p>
-     * The main use case for this property is the CLI and being able to invoke a generic test task with a filter that may or may not match tests depending on the project,
-     * and being able to not fail when it doesn't match any tests.
+     * The main use case for this property is the CLI: when running a generic test task with a
+     * filter (e.g. {@code --tests "*FooTest*"}) across a multi-project build, only some subprojects
+     * may have matching tests. Pass {@code --allow-no-matches} to avoid failing in those that don't.
+     * <p>
+     * The exposed CLI option is {@code --allow-no-matches}.
      *
      * @since 9.6.0
      * @see TestFilter#isFailOnNoMatchingTests()
      */
     @Incubating
-    @Option(option = "fail-on-no-matching-tests", description = "Fail the task if a filter was specified but no tests matched.")
+    @Option(option = "allow-no-matches", description = "Allow the task to succeed when a filter was specified but no tests matched.")
+    @NoDisable
     @Optional
     @Input
-    public abstract Property<Boolean> getFailOnNoMatchingTests();
+    public abstract Property<Boolean> getAllowNoMatchingTests();
 
     /**
      * Whether the task should fail if test sources are present, but no tests are discovered during test execution.  Defaults to true.
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
@@ -61,6 +61,7 @@ import org.gradle.api.reporting.Reporting;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.Nested;
+import org.gradle.api.tasks.Optional;
 import org.gradle.api.tasks.OutputDirectory;
 import org.gradle.api.tasks.TaskAction;
 import org.gradle.api.tasks.VerificationTask;
@@ -691,7 +692,7 @@ public abstract class AbstractTestTask extends ConventionTask implements Verific
     }
 
     private boolean shouldFailOnNoMatchingTests() {
-        return patternFiltersSpecified() && filter.isFailOnNoMatchingTests();
+        return patternFiltersSpecified() && getFailOnNoMatchingTests().getOrElse(filter.isFailOnNoMatchingTests());
     }
 
     boolean testsAreNotFiltered() {
@@ -792,6 +793,25 @@ public abstract class AbstractTestTask extends ConventionTask implements Verific
     public TestFilter getFilter() {
         return filter;
     }
+
+    /**
+     * Whether the task should fail when a filter was configured but no test matched the filter.
+     * Defaults to true.
+     *
+     * <p>
+     * If configured this takes precedence over {@link TestFilter#isFailOnNoMatchingTests()}.
+     * <p>
+     * The main use case for this property is the CLI and being able to invoke a generic test task with a filter that may or may not match tests depending on the project,
+     * and being able to not fail when it doesn't match any tests.
+     *
+     * @since 9.6.0
+     * @see TestFilter#isFailOnNoMatchingTests()
+     */
+    @Incubating
+    @Option(option = "fail-on-no-matching-tests", description = "Fail the task if a filter was specified but no tests matched.")
+    @Optional
+    @Input
+    public abstract Property<Boolean> getFailOnNoMatchingTests();
 
     /**
      * Whether the task should fail if test sources are present, but no tests are discovered during test execution.  Defaults to true.
--- a/platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/AbstractTestFrameworkIntegrationTest.groovy
+++ b/platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/AbstractTestFrameworkIntegrationTest.groovy
@@ -243,12 +243,12 @@ abstract class AbstractTestFrameworkIntegrationTest extends AbstractIntegrationS
         failure.assertHasCause("No tests found for given includes: [${testSuite('SomeTest')}.missingMethod](filter.includeTestsMatching)")
     }
 
-    def "can disable fail on no matching tests via task property"() {
+    def "can allow no matching tests via task property"() {
         given:
         createPassingFailingTest()
         buildFile << """
             tasks.withType(AbstractTestTask) {
-                failOnNoMatchingTests = false
+                allowNoMatchingTests = true
             }
         """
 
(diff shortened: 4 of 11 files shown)


platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
```diff
@@ -788,6 +788,24 @@ public TestFilter getFilter() {
...
+     * <p>
+     * The main use case for this property is the CLI and being able to invoke a generic test task with a filter that may or may not match tests depending on the project,
+     * and being able to not fail when it doesn't match any tests.
+     *
+     * @since 9.5.0
+     * @see TestFilter#isFailOnNoMatchingTests()
+     */
+    @Incubating
+    @Option(option = "fail-on-no-matching-tests", description = "Fail the task if a filter was specified but no tests matched.")
+    @Internal // Value already tracked as @Input via @Nested getFilter()
```
octylFractal: :x: This isn't part of the filter, right? The value is consumed directly from the task action, without touching the filter.
```suggestion
    @Input
```

For that matter, why doesn't this delegate to configuring the filter? I feel like it would be better to only have one source for the value. It's what's in the PR description too.

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/AbstractTestFrameworkIntegrationTest.groovy
octylFractal, line 1: ❓ If you keep the current behavior of having two different "fail on no matching" properties, can you add a couple tests showing that the CLI / task property always takes precedence over the `TestFilter` one?

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
ljacomet: You are correct, I tried making it a passthrough but our arch rule blew on it. And so I decided to go with a different approach. But then description and metadata needs to change indeed.

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/AbstractTestFrameworkIntegrationTest.groovy
```diff
@@ -243,6 +243,69 @@ abstract class AbstractTestFrameworkIntegrationTest extends AbstractIntegrationS
...
+        "task=false, filter=true"                        | "failOnNoMatchingTests = false"       | "filter.failOnNoMatchingTests = true"    | false
+        "task=true, filter=false"                        | "failOnNoMatchingTests = true"        | "filter.failOnNoMatchingTests = false"   | true
+        "task=false, filter=false"                       | "failOnNoMatchingTests = false"       | "filter.failOnNoMatchingTests = false"   | false
+        "task=true, filter=true"                         | "failOnNoMatchingTests = true"        | "filter.failOnNoMatchingTests = true"    | true
+        "task=false, filter unset (defaults to true)"    | "failOnNoMatchingTests = false"       | "// filter not configured"               | false
+        "task=true, filter unset (defaults to true)"     | "failOnNoMatchingTests = true"        | "// filter not configured"               | true
+        "task unset, filter=false (task defaults to true)" | "// task property not configured"   | "filter.failOnNoMatchingTests = false"   | false
+        "task unset, filter=true (task defaults to true)"  | "// task property not configured"   | "filter.failOnNoMatchingTests = true"    | true
```
octylFractal: 💭 What does "task defaults to true" mean here? Isn't it just "unset" and it passes through to the filter?

octylFractal (approved):

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/AbstractTestFrameworkIntegrationTest.groovy
ljacomet: Good catch, that text does not bring anything. And I noticed we are missing the case where neither is set.

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/AbstractTestFrameworkIntegrationTest.groovy
```diff
@@ -243,6 +243,70 @@ abstract class AbstractTestFrameworkIntegrationTest extends AbstractIntegrationS
...
+        createPassingFailingTest()
+
+        when:
+        run(testTaskName, "--tests", "${testSuite('SomeTest')}.missingMethod", "--no-fail-on-no-matching-tests")
+
+        then:
+        noExceptionThrown()
+    }
+
+    def "task property takes precedence over filter property for failOnNoMatchingTests: #scenario"() {
```
big-guy: Could you split this into two tests (one that should fail and one that does not) since they have different expectations?

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
```diff
@@ -793,6 +794,25 @@ public TestFilter getFilter() {
...
+     * and being able to not fail when it doesn't match any tests.
+     *
+     * @since 9.5.0
+     * @see TestFilter#isFailOnNoMatchingTests()
+     */
+    @Incubating
+    @Option(option = "fail-on-no-matching-tests", description = "Fail the task if a filter was specified but no tests matched.")
+    @Optional
+    @Input
+    public abstract Property<Boolean> getFailOnNoMatchingTests();
```
big-guy: Can you add this to platforms/documentation/docs/src/docs/dsl/org.gradle.api.tasks.testing.AbstractTestTask.xml?

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
```diff
@@ -793,6 +794,25 @@ public TestFilter getFilter() {
         return filter;
     }
 
+    /**
```
big-guy, line 798: Can you add some docs to java_testing.adoc when this is useful?

Can you also add a release note (it could be a placeholder)?

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
```diff
@@ -793,6 +794,25 @@ public TestFilter getFilter() {
...
+     * If configured this takes precedence over {@link TestFilter#isFailOnNoMatchingTests()}.
+     * <p>
+     * The main use case for this property is the CLI and being able to invoke a generic test task with a filter that may or may not match tests depending on the project,
+     * and being able to not fail when it doesn't match any tests.
+     *
+     * @since 9.5.0
+     * @see TestFilter#isFailOnNoMatchingTests()
+     */
+    @Incubating
+    @Option(option = "fail-on-no-matching-tests", description = "Fail the task if a filter was specified but no tests matched.")
```
big-guy: This is a really awkward option name. There are three words almost in a row that are just O or N.


```suggestion
    @Option(option = "matching-tests", description = "Require at least one test to match filter.")
```

So if someone ran `gradle test --tests "*Foo*" --matching-tests` this would mean "tests must match", which is the way it works now. 

But if someone ran `gradle test --tests "*Foo*" --no-matching-tests` this would mean "no matching tests are OK"

big-guy: I think the name of the property is consistent with this, so we don't have to call it `getMatchingTests` or something awkward like that.

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
big-guy: We don't have to do this now, but should `TestExecutionBuildConfigurationAction` now set this property?

big-guy (changes requested):

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
ljacomet: I don't think it can outside of a major, where we would remove the property on the `TestFilter` level. Because once it is set, that property is ignored.

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
ljacomet: I was going to say that the name is aligned with the `TestFilter` name. But since this is only about CLI, then yes.

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/AbstractTestFrameworkIntegrationTest.groovy
ljacomet: Sure

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
ljacomet: I should have looked into code before. I see that we set the property on `TestFilter` from there. Then we can change it, or not. I don't think it matters much. Unless I am missing something.

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/AbstractTestFrameworkIntegrationTest.groovy
```diff
@@ -243,6 +243,79 @@ abstract class AbstractTestFrameworkIntegrationTest extends AbstractIntegrationS
...
+        then:
+        failure.assertHasCause("No tests found for given includes:")
+
+        where:
+        scenario                    | taskPropertyConfig                | filterPropertyConfig                     | shouldFail
+        "task=true, filter=false"   | "failOnNoMatchingTests = true"    | "filter.failOnNoMatchingTests = false"   | true
+        "task=true, filter=true"    | "failOnNoMatchingTests = true"    | "filter.failOnNoMatchingTests = true"    | true
+        "task=true, filter unset"   | "failOnNoMatchingTests = true"    | "// filter not configured"               | true
+        "task unset, filter=true"   | "// task property not configured" | "filter.failOnNoMatchingTests = true"    | true
+        "task unset, filter unset"  | "// task property not configured" | "// filter not configured"               | true
```
big-guy: We don't need shouldFail now

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/testing/AbstractTestFrameworkIntegrationTest.groovy
```diff
@@ -243,6 +243,79 @@ abstract class AbstractTestFrameworkIntegrationTest extends AbstractIntegrationS
...
+
+        expect:
+        succeeds(testTaskName, "--tests", "${testSuite('SomeTest')}.missingMethod")
+
+        where:
+        scenario                    | taskPropertyConfig                | filterPropertyConfig                     | shouldFail
+        "task=false, filter=true"   | "failOnNoMatchingTests = false"   | "filter.failOnNoMatchingTests = true"    | false
+        "task=false, filter=false"  | "failOnNoMatchingTests = false"   | "filter.failOnNoMatchingTests = false"   | false
+        "task=false, filter unset"  | "failOnNoMatchingTests = false"   | "// filter not configured"               | false
+        "task unset, filter=false"  | "// task property not configured" | "filter.failOnNoMatchingTests = false"   | false
```
big-guy: same as below

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
big-guy: I don't think we need to change it as part of this PR, I was thinking that it was more consistent to change this property vs TestFilter because the TestLauncher is like the programmatic layer on top of the CLI.

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
```diff
@@ -123,6 +123,14 @@ For more details and examples of declaring filters in the build script, please s
 
 The command-line option is especially useful to execute a single test method. When you use `--tests`, be aware that the inclusions declared in the build script are still honored. It is also possible to supply multiple `--tests` options, all of whose patterns will take effect. The following sections have several examples of using the command-line option.
 
+[WARNING]
+====
+When filtering tests, the default in Gradle is to fail if the filter does not match any test.
+That means that using the top level task selector, like `test`, in a multi project has a high chance of failing if your filter matches tests in a single project only.
+
+To change this behavior, you can pass the additional command line option `--no-matching-tests` which will cause the build to succeed even if the filter does not match any test.
```
big-guy:
```suggestion
When filtering tests, Gradle will fail the build if the filter does not match any test by default.

When running multiple `Test` tasks with a top-level task selector like `test` in a multi-project build, there is a high chance of causing the build to fail if you supply a `--tests` filter that does not match at least one test in every `Test` task.

To change this behavior, pass the additional command-line option `--no-matching-tests` which will allow the build to succeed even if the filter does not match any test on any `Test` task.
```

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
```diff
@@ -793,6 +794,31 @@ public TestFilter getFilter() {
...
+     * <p>
+     * If configured this takes precedence over {@link TestFilter#isFailOnNoMatchingTests()}.
+     * <p>
+     * The main use case for this property is the CLI and being able to invoke a generic test task with a filter that may or may not match tests depending on the project,
+     * and being able to not fail when it doesn't match any tests.
+     * <p>
+     * The exposed task options are:
+     * <ul>
+     *     <li>{@code --matching-tests} option for specifying that the task should fail if no tests match the filter. This is the default.</li>
+     *     <li>{@code --no-matching-tests} option for specifying that the task should not fail it no tests match the filter.</li>
```
big-guy: I know I suggested these names, but please check if I'm overthinking this now.

`gradle test --tests "*Foo*" --matching-tests`

1. Run all tests that match "*Foo*" and filter matching tests (implies that zero matches are OK).
2. Run all tests that match "*Foo*" and expect _some_ matching tests. (This PR's meaning)

`gradle test --tests "*Foo*" --no-matching-tests`

1. Run all tests that match "*Foo*" and fail when there are no matching tests.
2. Run all tests that match "*Foo*" and allow no matching tests. (This PR's meaning)

I'm still onboard with (2)'s meaning, but I could see someone else reading these like (1).

big-guy (approved):
Suggestion on wording and test cleanup

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -166,6 +166,10 @@ Gradle Wrapper scripts and application start scripts include links to the GitHub
 Previously, these links always pointed to the latest template versions, which could drift out of sync with the actual generated scripts as templates evolved. 
 This has now been fixed: scripts always link to the exact template version they were generated from.
 
+### More control over test selection behavior on the command line
+
+With the addition of the `--no-matching-tests`, you can now use a `--tests` filter without causing the build to fail when no tests match the filter.
```
lkasso:
````suggestion
When running `--tests` filters on a task like `test` that spans multiple subprojects, Gradle previously failed the build if the task found no matching tests.
You can now pass `--no-matching-tests` alongside `--tests` to allow the build to succeed when a filter matches no tests on a given `Test` task:

`./gradlew test --tests "*FooTest*" --no-matching-tests`

This is especially useful in multi-project builds where a filter is only expected to match tests in some subprojects.

The default behavior (fail when no tests match) is unchanged. 
````

File: platforms/documentation/docs/src/docs/release/notes.md [hunk 12]
```diff
@@ -166,6 +166,10 @@ Gradle Wrapper scripts and application start scripts include links to the GitHub
 Previously, these links always pointed to the latest template versions, which could drift out of sync with the actual generated scripts as templates evolved. 
 This has now been fixed: scripts always link to the exact template version they were generated from.
 
+### More control over test selection behavior on the command line
```
lkasso:
```suggestion
#### More control over test filtering on the command line
```

lkasso (commented):
Some suggestions on the wording.

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java
ljacomet: I see what you are saying and I agree. That name has too much ambiguity. It is even too close to `--tests` in the sense that it could be interpreted as taking a value for those matches ...

I don't like attempting to have a positive that means OK to have no match. It would make the parallel between `AbstractTestTask` and `TestFilter` harder.

What about the following?
* `--require-matching-tests`
* `--no-require-matching-tests`

Same true/false meaning, but simpler and no double negation.

big-guy: What about:

- `--zero-matches` allow 0 matches
- `--no-zero-matches` fail if there are 0 matches

Or change `@Option` to disable the automatic negative sometimes, so we'd only add one option.

And then we could consider `--allow-no-matches` without having to worry about what `--no-allow-no-matches` means. 

We might want to be a bit sensitive on how much typing this is, so we might want an even shorter option: `--none` (meaning "none is OK") or `--allow-none`.

ljacomet:
@big-guy Following our conversation, I pushed another set of changes.

* Rename the option to `--allow-no-matches`
* Disable opposite option creation through a dedicated annotation

If you think that version makes sense, let's push it. Otherwise, as we discussed, let's close and think about a potential other solution.

testing/architecture-test/src/changes/archunit-store/public-api-nullability.txt
```diff
@@ -126,8 +126,6 @@ Class <org.gradle.api.tasks.compile.GroovyForkOptions> is not annotated (directl
 Class <org.gradle.api.tasks.compile.JavaCompile> is not annotated (directly or via an enclosing element) with @org.jspecify.annotations.NullMarked in (JavaCompile.java:0)
 Class <org.gradle.api.tasks.compile.ProviderAwareCompilerDaemonForkOptions> is not annotated (directly or via an enclosing element) with @org.jspecify.annotations.NullMarked in (ProviderAwareCompilerDaemonForkOptions.java:0)
 Class <org.gradle.api.tasks.diagnostics.artifact.transforms.ArtifactTransformReports> is not annotated (directly or via an enclosing element) with @org.jspecify.annotations.NullMarked in (ArtifactTransformReports.java:0)
-Class <org.gradle.api.tasks.options.Option> is not annotated (directly or via an enclosing element) with @org.jspecify.annotations.NullMarked in (Option.java:0)
-Class <org.gradle.api.tasks.options.OptionValues> is not annotated (directly or via an enclosing element) with @org.jspecify.annotations.NullMarked in (OptionValues.java:0)
 Class <org.gradle.api.tasks.scala.IncrementalCompileOptions> is not annotated (directly or via an enclosing element) with @org.jspecify.annotations.NullMarked in (IncrementalCompileOptions.java:0)
```
mlopatkin, line 131: 🤔 Isn't this going to be a potentially source-incompatible for Kotlin clients that inspect the annotations?

I suppose this is a very uncommon use case...

mlopatkin (commented):
I only looked at the AC-owned file, and I have some concerns about source compatibility implied by the change.

ljacomet:
Given that this solution does not give all the expected benefits, closing for now.

Issue #36719, opened 2026-02-12
Unexpected Problem id required for deprecation logger


donat:
Reproducer:
```
org.gradle.internal.deprecation.DeprecationLogger
    .deprecate("1")
//    .withProblemId("my-problem-id")
    .willBeRemovedInGradle10()
    .undocumented()
    .nagUser()
```

Fails, but removing the command makes the build pass.

Also, replacing `.deprecate("1")` with `.deprecateTask("1")` also makes the build pass.

PR #36720, opened 2026-02-12
Don't require setting explicit problem id for deprecation logging

donat:
Fixes https://github.com/gradle/gradle/issues/36719

### Context

When introducing Problems API integration for DeprecationLogger, we missed adding default values for the deprecation entries not providing `WithReplacement`. This did not cause any problems, until we [introduced stricter validation for problem ids and groups](https://github.com/gradle/gradle/pull/35483). This PR adds the missing default value to the message builder.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


donat:
@bot-gradle test this

PR #36721, opened 2026-02-12
Platform-independent dependency lockfiles

ljacomet:
Lockfile generation used `Files.write(Path, Iterable, Charset)` which
internally uses `System.lineSeparator()`, producing CRLF on Windows and
LF on Unix. This caused noisy VCS diffs when developers on different
platforms regenerated lockfiles. Write lockfiles with explicit LF (`\n`)
line endings regardless of platform.

Added upgrade guide entry

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -70,6 +70,13 @@ Gradle <<configuration_cache_requirements.adoc#config_cache:requirements:reading
 Starting with Gradle 9.5.0, calling `.values()` on the maps returned from `System.getenv()` or `System.getProperties()` records **all** environment variables or system properties respectively as inputs to the Configuration Cache, similarly to how `.forEach()` calls are tracked in previous versions.
 If your build configuration uses `System.getenv().values()` or `System.getProperties().values()`, you may experience more frequent Configuration Cache invalidations.
 
+==== Dependency lockfiles are generated with a platform-independent line ending
+
+Dependency lockfiles now use a platform independent line ending (`\n`) instead of the system default.
+Existing lockfiles written with Windows line endings can still be read, but new lockfiles generated will only use platform independent line endings.
+
+During the next lock update using Gradle 9.5.0 or above, if the lockfiles were previously generated with Windows line endings, all line endings will be changed, regardless of whether dependencies have changed.
+
 [[changes_9.4.0]]
 == Upgrading from 9.3.1 and earlier
 


platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -70,6 +70,13 @@ Gradle <<configuration_cache_requirements.adoc#config_cache:requirements:reading
 Starting with Gradle 9.5.0, calling `.values()` on the maps returned from `System.getenv()` or `System.getProperties()` records **all** environment variables or system properties respectively as inputs to the Configuration Cache, similarly to how `.forEach()` calls are tracked in previous versions.
 If your build configuration uses `System.getenv().values()` or `System.getProperties().values()`, you may experience more frequent Configuration Cache invalidations.
 
+==== Dependency lockfiles are generated with a platform-independent line ending
+
+Dependency lockfiles now use a platform independent line ending (`\n`) instead of the system default.
+Lockfiles using Windows line endings will continue to work, but no longer can be generated.
```
ghale:
```suggestion
Existing lockfiles written with Windows line endings can still be read, but new lockfiles generated will only use platform independent line endings.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -70,6 +70,13 @@ Gradle <<configuration_cache_requirements.adoc#config_cache:requirements:reading
...
+
+If you have generated and committed lockfiles on Windows, on the next lock update using Gradle 9.5.0 or above, all line endings will change, regardless of dependency changes.
```
ghale: During the next lock update using Gradle 9.5.0 or above, if the lockfiles were previously generated with Windows line endings, all line endings will be changed, regardless of whether dependencies have changed.

platforms/software/dependency-management/src/test/groovy/org/gradle/internal/locking/DefaultDependencyLockingProviderTest.groovy
ghale, line 1: ❔ Do we have a test that checks that a lock file with windows line endings is still readable?  If not, should we have one to guard against regressions?  I'm imagining someone with `autocrlf true` set on windows or a specific git attribute for lock files that sets it to CRLF, so that even though we generate with LF, it still gets set to CRLF on checkout.

ljacomet: Let me add a test for this. Since only writing of files was changed, it should not be a problem, but better safe than sorry!

PR #36723, opened 2026-02-12
SourceDistributionProvider now searches parent directories for gradle-wrapper.properties

### LINKED ISSUE #36726: SourceDistributionProvider doesn't derive url from distributionUrl for included builds
### Current Behavior

As of 9.4.0-RC-1, `SourceDistributionProvider` derives the source distribution download URL from the `distributionUrl` property in `gradle-wrapper.properties`, see #27863. This behavior is critical for corporate environments with restricted internet access.

However, included builds and buildSrc projects typically lack their own `gradle-wrapper.properties` file as the wrapper configuration exists only in the root project. In these cases, `SourceDistributionProvider` currently defaults to https://services.gradle.org/distributions/gradle-9.4.0-rc-1-src.zip.

### Expected Behavior

The url used to download the src-zip is derived from the `distributionUrl` of the root project, if the project itself doesn't contain a `gradle-wrapper.properties`.

### Context (optional)

The issue leads to import / sync issues in IntelliJ IDEA: https://youtrack.jetbrains.com/issue/IDEA-385844/Syncing-Gradle-buildSrc-project-doesnt-respect-root-projects-gradle-wrapper.properties

1. Configure your network to block https://services.gradle.org
2. Create a new Gradle project in IntelliJ IDEA.
3. Use the following gradle/wrapper/gradle-wrapper.properties file, whereas the distributionUrl is set to your internal Gradle distributions mirror:
```#Mon Feb 09 07:35:01 CET 2026
distributionBase=GRADLE_USER_HOME
distributionPath=wrapper/dists
distributionUrl=https\://nexus.example.com/repository/gradle-distributions/gradle-9.4.0-rc-1-bin.zip
zipStoreBase=GRADLE_USER_HOME
zipStorePath=wrapper/dists
```
4. Since Gradle 9.4.0-RC-1 this works fine for a simple project, as Gradle is now able to derive the url to the corresponding src-zip from the distributionUrl.
5. But as soon as a `buildSrc` project is added, syncing or importing the Gradle project will fail: Gradle tries to download the src-zip for the buildSrc-project from https://services.gradle.org/distributions/gradle-9.4.0-rc-1-src.zip (which might not be available in a corporate environment).

### Gradle version

9.4.0 RC-1

### Build scan URL (optional)

_No response_

OS version: macOS 15.7.3
IDE version: IntelliJ IDEA 2025.3.1

Gradle is configured to load everything from an internal Nexus repository, which means:

```kts
// settings.gradle.kts
pluginManagement {
    repositories { maven(url = "https://nexus.example.com/repository/gradle-plugins/") }
}

dependencyResolutionManagement {
    repositories { maven(url = "https://nexus.example.com/repository/maven-public/") }
}
```

```properties
# gradle/wrapper/gradle-wrapper.properties
distributionBase=GRADLE_USER_HOME
distributionPath=wrapper/dists
distributionUrl=https\://nexus.example.com/repository/gradle-distributions/gradle-9.4.0-rc-1-bin.zip
zipStoreBase=GRADLE_USER_HOME
zipStorePath=wrapper/dists
```

```bash
# Environment variable for Gradle libs (groovy dependencies)
export GRADLE_LIBS_REPO_OVERRIDE=https://nexus.example.com/repository/gradle-libs/
```

kelemen:
When the project is an included build containing build logic, then `gradle-wrapper.properties` is not in the `settingsDir`. This PR works around this problem by trying to locate `gradle-wrapper.properties` in parent directories as well. This should work in most sensible cases, and even if it doesn't, it offers a workaround for people to put a dummy `gradle-wrapper.properties` to a parent directory to set the location of the repository.

If there is another RC for 9.4, then it might be worthwhile to add this change to 9.4 as well, because then the fix of the referenced issue will not break on the rather common case of having a build logic.

* Fixes #35963 (that is the part where the project contains a build logic project)
* Fixes #36726

### Context
See <https://github.com/gradle/gradle/issues/27863#issuecomment-3889996748> for the original report.

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/internal/GradleDistRepoDescriptorLocator.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/internal/GradleDistRepoDescriptorLocator.kt
@@ -64,10 +64,24 @@ class GradleDistRepoDescriptorLocator(
         WrapperCredentials.findCredentials(baseUrl) { System.getProperty(it) }
 
     private
-    fun findStandardWrapperUri(): URI? {
-        val wrapperProperties = WrapperExecutor.wrapperPropertiesForProjectDirectory(rootProjectDir)
-        if (wrapperProperties.exists()) {
+    fun findWrapperProperties(): File? {
+        var rootCandidate: File? = rootProjectDir
+        while (rootCandidate != null) {
+            val resultCandidate = WrapperExecutor.wrapperPropertiesForProjectDirectory(rootCandidate)
+            if (resultCandidate.exists()) {
+                return resultCandidate.takeIf {
+                    resultCandidate.isFile && resultCandidate.canRead()
+                }
+            }
+            rootCandidate = rootCandidate.parentFile
+        }
+        return null
+    }
 
+    private
+    fun findStandardWrapperUri(): URI? {
+        val wrapperProperties = findWrapperProperties()
+        if (wrapperProperties != null) {
             val currentWrapperUri = Properties()
                 .apply { wrapperProperties.inputStream().use { load(it) } }
                 .getProperty(WrapperExecutor.DISTRIBUTION_URL_PROPERTY)
--- a/platforms/core-configuration/kotlin-dsl/src/integTest/kotlin/org/gradle/kotlin/dsl/resolver/SourceDistributionResolverIntegrationTest.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/integTest/kotlin/org/gradle/kotlin/dsl/resolver/SourceDistributionResolverIntegrationTest.kt
@@ -4,22 +4,22 @@ import org.gradle.integtests.fixtures.RepoScriptBlockUtil.mavenCentralRepository
 import org.gradle.kotlin.dsl.fixtures.AbstractKotlinIntegrationTest
 import org.gradle.kotlin.dsl.resolver.internal.GradleDistRepoDescriptorLocator
 import org.gradle.test.fixtures.dsl.GradleDsl
-import org.gradle.test.fixtures.server.http.BlockingHttpServer
-import org.junit.After
-import org.junit.Rule
 
 import org.junit.Test
 
 class SourceDistributionResolverIntegrationTest : AbstractKotlinIntegrationTest() {
 
-    private fun withCustomGradleProperties(distributionUrl: String) {
-        withFile("fake-root/gradle/wrapper/gradle-wrapper.properties", """
+    private fun withCustomGradleProperties(distributionUrl: String, pathPrefix: String = "fake-root/") {
+        withFile("${pathPrefix}gradle/wrapper/gradle-wrapper.properties", """
         distributionUrl=${distributionUrl}
         """)
     }
 
     private fun queryGradleDistRepository() =
         "${GradleDistRepoDescriptorLocator::class.qualifiedName}(project, explicitRootProjectDir = file(\"fake-root\")).gradleDistRepository"
+
+    private fun queryGradleDistRepositoryWithDefaultRoot() =
+        "${GradleDistRepoDescriptorLocator::class.qualifiedName}(project).gradleDistRepository"
 
     private fun verifyPasswordCredentials(expectedUser: String, expectedPassword: String): String {
         return $$"""
@@ -107,6 +107,28 @@ class SourceDistributionResolverIntegrationTest : AbstractKotlinIntegrationTest(
         )
 
         build()
+    }
+
+    @Test
+    fun `test parent dir wrapper properties`() {
+        withCustomGradleProperties("https://my-host.org/my-path/distributions/gradle-9.4.0-bin.zip", "")
+
+        val actualProjectRoot = projectRoot.resolve("child").resolve("grand-child")
+        withProjectRoot(actualProjectRoot) {
+            withSettings("rootProject.name = \"test-project\"")
+            withBuildScript(
+                $$"""
+                val gradleDistRepository = $${queryGradleDistRepositoryWithDefaultRoot()}
+                require(gradleDistRepository.repoBaseUrl == uri("https://my-host.org/my-path/distributions")) {
+                    "Unexpected repoBaseUrl in: ${gradleDistRepository}"
+                }
+                require(gradleDistRepository.artifactPattern == "[module]-[revision](-[classifier])(.[ext])") {
+                    "Unexpected artifactPattern in: ${gradleDistRepository}"
+                }
+                """.trimIndent()
+            )
+        }
+        build(actualProjectRoot)
     }
 
     @Test


chkpnt:
I've tried to test your branch with my local setup. For this, I've built `:distributions-full:binDistributionZip` and `:distributions-full:srcDistributionZip` and put the generated `gradle-9.5.0-bin.zip` and `gradle-9.5.0-src.zip` on my internal Nexus at https://nexus.example.com/repository/gradle-distributions-test/

But with `distributionUrl=https\://nexus.example.com/repository/gradle-distributions-test/gradle-9.5.0-bin.zip`, I get the following error:


```
> Task :prepareKotlinBuildScriptModel UP-TO-DATE
Unexpected exception while resolving Gradle distribution sources: Could not resolve all files for configuration 'detachedConfiguration1'.
org.gradle.api.internal.artifacts.ivyservice.TypedResolveException: Could not resolve all files for configuration 'detachedConfiguration1'.
	at org.gradle.api.internal.artifacts.ResolveExceptionMapper.mapFailure(ResolveExceptionMapper.java:73)
	at org.gradle.api.internal.artifacts.ResolveExceptionMapper.mapFailures(ResolveExceptionMapper.java:65)
	...
	at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:64)
	at org.gradle.internal.concurrent.AbstractManagedExecutor$1.run(AbstractManagedExecutor.java:47)
	at java.base@21.0.7/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base@21.0.7/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base@21.0.7/java.lang.Thread.run(Thread.java:1583)
Caused by: org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find any matches for gradle:gradle:(9.4, 9.5.0-20260211230000+0000] as no versions of gradle:gradle are available.
Searched in the following locations:
Caused by: org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find any matches for gradle:gradle:(9.4, 9.5.0-20260211230000+0000] as no versions of gradle:gradle are available.

  - https://nexus.example.com/repository/gradle-distributions-test/
Required by:
    detached context for root project 'my-gradle-plugin'
```

So where to specify the versions available?

kelemen:
If you want to test this from a local build, then you generally can do two things:

- Set the `GRADLE_VERSION_OVERRIDE` environment variable to `9.5.0` (or whatever version you set in the filename).
- Adjust the `GradleDistRepoDescriptorLocator` to set your version in the filename you are sharing your dist from `project.gradle.gradleVersion`. I generally do this when testing, because setting environment variable is a hassle (I mean, you have to restart the IDE for it).

kelemen:
A warning though: Be careful with setting the version to `9.5.0` though, because then Gradle might store cached dists in places, and then you might get suprised when the real `9.5.0` is out. I would recommend using version number like `9.3.999` (because it is less than the next version `9.4.0` and surely there won't be an official version like this).

chkpnt:
🎉 I can confirm: your fix works fine. 👍

(I've set `GRADLE_VERSION_OVERRIDE` as I didn't understand your second approach.)

h0tk3y:

As [commented](https://github.com/gradle/gradle/issues/36726#issuecomment-3907446570) on the referenced issue, someone will follow up.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/internal/GradleDistRepoDescriptorLocator.kt
```diff
@@ -64,10 +64,22 @@ class GradleDistRepoDescriptorLocator(
...
+        var rootCandidate: File? = rootProjectDir
+        while (rootCandidate != null) {
+            val resultCandidate = WrapperExecutor.wrapperPropertiesForProjectDirectory(rootCandidate)
+            if (resultCandidate.exists()) {
+                return resultCandidate
+            }
+            rootCandidate = rootCandidate.parentFile
+        }
+        return null
+    }
```
donat, line 79: We could probably replace this with 

```
val layout = BuildLayoutFactory().getLayoutFor(rootProjectDir, true)
val wrapperProperties = WrapperExecutor.wrapperPropertiesForProjectDirectory(layout.getRootDirectory())
```

kelemen: I don't think that would work, because this fix is specifically for included build (including buildSrc), and given that included builds contain their own `settings.gradle(.kts)`, it would just stop there, and wouldn't find the `gradle-wrapper.properties`. In fact, for `buildSrc`, the 2nd argument is effectively ignored (and treated the same as `false`), so for `buildSrc` this doesn't seem to do anything.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/internal/GradleDistRepoDescriptorLocator.kt
```diff
@@ -64,10 +64,22 @@ class GradleDistRepoDescriptorLocator(
...
 
     private
-    fun findStandardWrapperUri(): URI? {
-        val wrapperProperties = WrapperExecutor.wrapperPropertiesForProjectDirectory(rootProjectDir)
-        if (wrapperProperties.exists()) {
+    fun findWrapperProperties(): File? {
+        var rootCandidate: File? = rootProjectDir
+        while (rootCandidate != null) {
+            val resultCandidate = WrapperExecutor.wrapperPropertiesForProjectDirectory(rootCandidate)
+            if (resultCandidate.exists()) {
```
eskatos, line 71: 💭 It might be better to use `.isFile() && .canRead()` here to be sure we can actually use the candidate file. It might also be good to just stop traversing up if it can't be read.

kelemen: Makes sense. I have adjusted to code to check if `canRead()` is true (which implies `isFile`). Though I still stop at `exists()`, because if the user has `gradle-wrapper.properties` as a directory, then something is fishy, and probably searching upwards is not intended.

kelemen: Actually, `canRead()` doesn't imply `isFile`, I was wrong (maybe should have tested it before spreading fake news). So, I have added the `isFile` check as well.

donat:
@bot-gradle test RFN

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/internal/GradleDistRepoDescriptorLocator.kt
donat: That's right. At the same time, the code is running in the context of a Tooling model builder, which has access to the current `Project` reference. From that, we can get the root build's root folder, which makes it unnecessary to traverse the file hierarchy to find `gradle-wrapper.properties`.

kelemen: How can I get the main project from the project reference I already have (Is it `(project as ProjectInternal).gradle.defaultProject.rootProject`)? Anyway - though that would be a better starting point than the current - I would still search upwards, because it is possible for someone to directly load the included build (in which case Gradle wouldn't know that this project is just an build logic of something else).

donat:
Superseded by https://github.com/gradle/gradle/pull/36778

PR #36729, opened 2026-02-13
Fix DynamicRevisionRemoteResolveWithMetadataSupplierIntegrationTest for Configuration Cache

tresat:


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/suppliers/DynamicRevisionRemoteResolveWithMetadataSupplierIntegrationTest.groovy
```diff
@@ -263,7 +264,7 @@ abstract class DynamicRevisionRemoteResolveWithMetadataSupplierIntegrationTest e
         checkResolve "group:projectA:1.+": ["group:projectA:1.2", "didn't match version 2.0"], "group:projectB:latest.release": "group:projectB:2.3"
     }
 
-    @ToBeFixedForConfigurationCache
+    @UnsupportedWithConfigurationCache(because = "This test is structured to expect repeated resolutions")
     def "can use --offline to use cached result after remote failure"() {
```
mlopatkin, line 266: ❓ Looking just at the name of the test, it should be feasible to run it with CC.

From what I see, it fails because there is one successful resolution in line 297, and we need it to prime the dependency cache. But it feels like set-up step, rather than the essential part. And the essential part (failure/offline fallback) should still work with CC.

Can we just drop CC state after the priming run to force re-resolve with failure instead? For example, we can use `-D${ConfigurationCacheRecreateOption.PROPERTY_NAME}=true` flag.

tresat: I was unaware of this option.  This seems like exactly what is needed, thanks.

mlopatkin (approved):

Issue #36596, opened 2026-02-04
[Backport] Update develocity plugin to latest version


ljacomet:
Let's update the Develocity plugin version for Gradle 8.14.5

PR #36731, opened 2026-02-13
[Backport] Update AutoAppliedDevelocityPlugin to use 4.3.2

wolfs:
So we ship Gradle 8.14.5 with the latest Develocity plugin.

Fixes https://github.com/gradle/gradle/issues/36596

### DIFF
--- a/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/BuildScanPluginSmokeTest.groovy
+++ b/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/BuildScanPluginSmokeTest.groovy
@@ -33,10 +33,6 @@ import java.util.concurrent.ConcurrentHashMap
 class BuildScanPluginSmokeTest extends AbstractSmokeTest {
 
     enum CI {
-        TEAM_CITY(
-            AbstractSmokeTest.TestedVersions.teamCityGradlePluginRef,
-            "https://raw.githubusercontent.com/etiennestuder/teamcity-build-scan-plugin/%s/agent/src/main/resources/build-scan-init.gradle"
-        ),
         JENKINS(
             AbstractSmokeTest.TestedVersions.jenkinsGradlePluginRef,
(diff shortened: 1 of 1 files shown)


wolfs:
@bot-gradle test AST

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/BuildScanPluginSmokeTest.groovy
```diff
@@ -33,10 +33,6 @@ import java.util.concurrent.ConcurrentHashMap
 class BuildScanPluginSmokeTest extends AbstractSmokeTest {
 
     enum CI {
-        TEAM_CITY(
```
wolfs, line 36: 💭 We are not testing the Teamcity plugin on `master` anymore, so let's not test it here. It fails with the Develocity plugin, since we are using an old version and trying to set the Develocity plugin version that doesn't have an Gradle Enterprise equivalent any more.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/PluginDependenciesSpecExtensions.kt
```diff
@@ -41,5 +41,5 @@ val PluginDependenciesSpec.`gradle-enterprise`: PluginDependencySpec
             .willBeRemovedInGradle9()
             .withUpgradeGuideSection(8, "gradle_enterprise_extension_deprecated")
             .nagUser()
-        return this.id(AutoAppliedDevelocityPlugin.GRADLE_ENTERPRISE_PLUGIN_ID.id).version(AutoAppliedDevelocityPlugin.VERSION)
+        return this.id(AutoAppliedDevelocityPlugin.GRADLE_ENTERPRISE_PLUGIN_ID.id).version(AutoAppliedDevelocityPlugin.GRADLE_ENTERPRISE_PLUGIN_LATEST_VERSION)
```
wolfs, line 44: 💭 With this, `gradle-enterprise` still refers to the Gradle Enterprise plugin, so we don't break people's builds.

PR #36735, opened 2026-02-13
Fix CC-incompatible test in NestedInputIntegrationTest

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/NestedInputIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/NestedInputIntegrationTest.groovy
@@ -20,13 +20,11 @@ import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.integtests.fixtures.DirectoryBuildCacheFixture
-import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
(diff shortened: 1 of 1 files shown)


subprojects/core/src/integTest/groovy/org/gradle/api/tasks/NestedInputIntegrationTest.groovy
```diff
@@ -417,7 +415,7 @@ class NestedInputIntegrationTest extends AbstractIntegrationSpec implements Dire
             def secondString = providers.gradleProperty('secondInput').orNull
             def secondBean = new OtherNestedBean(secondInput: secondString, secondOutputFile: file("${secondOutputFile}"), secondInputFile: file("${secondInputFile}"))
 
-            task taskWithNestedProperty(type: TaskWithNestedProperty) {
+            def taskWithNestedProperty = tasks.create("taskWithNestedProperty", TaskWithNestedProperty) {
```
mlopatkin, line 418: 💅 
```suggestion
            def taskWithNestedProperty = tasks.register("taskWithNestedProperty", TaskWithNestedProperty) {
```

ov7a: It’s compicated, other tests expect eager task instance. I’ve changed it for two reasons:
1. More readable
2. Clearer error with cc (with the current way of declaraction, it fails with “failed to find a property named <taskName>”

PR #36737, opened 2026-02-13
In DCL allow declaring any type of dependencies, with or without extra configuration action

donat:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/dependencyConfigurationSchema.kt
+++ b/platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/dependencyConfigurationSchema.kt
@@ -97,13 +97,13 @@ class DependencyCollectorsComponent : AnalysisSchemaComponent, ObjectConversionC
             FixedTypeDiscovery(
                 PlatformDependencyModifiers::class,
                 listOf(
-                    TypeDiscovery.DiscoveredClass(ProjectDependency::class, listOf(TypeDiscovery.DiscoveredClass.DiscoveryTag.Special("needed for dependencies DSL")))
+                    TypeDiscovery.DiscoveredClass(ProjectDependency::class, TypeDiscovery.DiscoveredClass.DiscoveryTag.Special("needed for dependencies DSL"))
                 )
             ),
             FixedTypeDiscovery(
                 DependencyCollector::class,
                 listOf(
-                    TypeDiscovery.DiscoveredClass(ExternalModuleDependency::class, listOf(TypeDiscovery.DiscoveredClass.DiscoveryTag.Special("needed for dependencies DSL"))),
+                    TypeDiscovery.DiscoveredClass(ExternalModuleDependency::class, TypeDiscovery.DiscoveredClass.DiscoveryTag.Special("needed for dependencies DSL")),
                 )
             ),
         )


donat:
@bot-gradle test this

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/dependencyConfigurationSchema.kt
```diff
@@ -71,9 +97,15 @@ class DependencyCollectorsComponent : AnalysisSchemaComponent, ObjectConversionC
             FixedTypeDiscovery(
                 PlatformDependencyModifiers::class,
                 listOf(
-                    TypeDiscovery.DiscoveredClass(ExternalDependency::class, listOf(TypeDiscovery.DiscoveredClass.DiscoveryTag.Special("needed for dependencies DSL")))
+                    TypeDiscovery.DiscoveredClass(ProjectDependency::class, listOf(TypeDiscovery.DiscoveredClass.DiscoveryTag.Special("needed for dependencies DSL")))
                 )
-            )
+            ),
+            FixedTypeDiscovery(
+                DependencyCollector::class,
```
donat: Not sure if it's the correct solution, but without it, DependencyCollectorsComponentTest fails saying `ExternalModuleDependency` is not part of the schema.

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/DependencyCollectorFunctionExtractorAndRuntimeResolver.kt
```diff
@@ -169,18 +175,44 @@ class DependencyCollectorFunctionExtractorAndRuntimeResolver(
...
+    ): SchemaResult<DataMemberFunction> =
+        host.withTag(dependencyCollectorTag(kClass, name)) {
+            DefaultDataMemberFunction(
+                host.containerTypeRef(kClass)
+                    .orFailWith { return it },
+                name,
+                listOf(dependencyParam),
+                false,
+                FunctionSemanticsInternal.DefaultAddAndConfigure(
+                    @OptIn(LossySchemaBuildingOperation::class) // referencing a predefined type is safe
```
h0tk3y: ❓ Is this opt-in a leftover from some edits?

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/DependencyCollectorFunctionExtractorAndRuntimeResolver.kt
```diff
@@ -169,18 +175,44 @@ class DependencyCollectorFunctionExtractorAndRuntimeResolver(
...
+        }
+
+
     private fun dependencyCollectorTag(kClass: KClass<*>, name: String) =
         SchemaBuildingContextElement.TagContextElement("dependency collector '$name' in $kClass")
 
     private
-    fun buildDeclarativeRuntimeFunction(collectorAccessor: SupportedCallable) = object : DeclarativeRuntimeFunction {
+    fun buildDeclarativeRuntimeFunction(collectorAccessor: SupportedCallable, lambdaReceiverType: DataTypeRef?) = object : DeclarativeRuntimeFunction {
         override fun callBy(receiver: Any?, binding: Map<DataParameter, Any?>, hasLambda: Boolean): DeclarativeRuntimeFunction.InvocationResult {
```
h0tk3y, line 220: 💄 Just to prevent misuse in potential future edits, I'd suggest also validating that `lambdaReceiverType != null` if and only if `hasLambda` is true. It'd be enough to throw an exception if that's not the case, because normally the "DCL runtime" relies on the earlier stages of resolution and should only get correct data.

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/DependencyCollectorFunctionExtractorAndRuntimeResolver.kt
```diff
@@ -169,18 +175,44 @@ class DependencyCollectorFunctionExtractorAndRuntimeResolver(
...
             val dependencyCollector = collectorAccessor.kCallable.call(checkNotNull(receiver)) as DependencyCollector
-            when (val dependencyNotation = binding.values.single()) {
-                is ProjectDependency -> dependencyCollector.add(dependencyNotation)
-                else -> dependencyCollector.add(dependencyNotation.toString())
+            lateinit var target : Dependency
+            when (val dependency = binding.values.single()) {
+                is ProjectDependency -> dependencyCollector.add(dependency, Action { target = it })
+                is Dependency -> dependencyCollector.add(dependency, Action { target = it })
+                else -> dependencyCollector.add(dependency.toString(), Action { target = it })
```
h0tk3y: 💭 Accepting anything and using its `toString()` here sounds error-prone if we edit this in the future. Is only `String` really accepted in the `else` branch? Would it make sense to do `is String -> ...` and `else -> error("...")`?

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/dependencyConfigurationSchema.kt
h0tk3y: You can use `FixedTypeDiscovery` with a `keyClass = null`. That will bring the classes into the schema unconditionally, as opposed to only including them if `keyClass` appears in the schema.

subprojects/core-api/src/main/java/org/gradle/api/artifacts/ModuleDependency.java
```diff
@@ -79,13 +79,15 @@ public interface ModuleDependency extends Dependency, HasConfigurableAttributes<
      * @param excludeProperties the properties to define the exclude rule.
      * @return this
      */
+    @Adding
     ModuleDependency exclude(Map<String, String> excludeProperties);
```
eskatos, line 84: ❓ This map of properties taking function should not be exposed. It makes the definitions hard to reason about for tooling. For one, it is impossible to know what are the available properties.

```kotlin
implementation("commons-beanutils:commons-beanutils:1.9.4") {
    exclude(mapOf("group" to "commons-collections"))
}
```

I'd rather we have definitions look more like this:

```kotlin
implementation("commons-beanutils:commons-beanutils:1.9.4") {
    excludes += listOf(
        module("some.group", "some-name"),
        moduleGroup("commons-collections"),
        moduleName("slf4j-simple"),
    )
}
```

The above is just a proposal to be discussed.

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/DependencyCollectorFunctionExtractorAndRuntimeResolver.kt
donat: Yes. We simply made a copy of the method above and slightly adjusted the content.

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/dependencyConfigurationSchema.kt
donat: Got it. With that in mind, I don't think we should unconditionally include `ExternalModuleDependency` in the schema. Is `DependencyCollector` a good condition here?

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/DependencyCollectorFunctionExtractorAndRuntimeResolver.kt
donat: After cleaning up the code a bit it turned out that lambdaReceiverType is not even needed here. Ptal at the latest state.

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/DependencyCollectorFunctionExtractorAndRuntimeResolver.kt
donat: Right. It's just how it was before. I've added your proposal.

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/dependencyConfigurationSchema.kt
donat: Thanks for the explanation. I don't think we should unconditionally include ExternalModuleDependency in the schema. Is DependencyCollector a good condition here?

h0tk3y: It's OK either way, and `null` is also fine here because if we include the `DepencencyCollectorsComponent` schema building customization, we actually mean to include the dependencies DSL and all types that come with it.

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/dependencycollectors/DependencyCollectorFunctionExtractorAndRuntimeResolver.kt
h0tk3y: Ok, this makes sense.

subprojects/core-api/src/main/java/org/gradle/api/artifacts/ModuleDependency.java
donat: During a separate discussion we agreed that we'll merge the solution as-is and then follow-up with a stricter DSL where the map notation will be disabled and we'll introduce a new TBD design for supporting map-backed configuration.

blindpirate:
The failure was on a test agent. Re-added to queue.

donat:
@blindpirate thank for looking after my changes!

PR #36740, opened 2026-02-13
Mark CC-incompatible test in MissingTaskDependenciesIntegrationTest

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/MissingTaskDependenciesIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/MissingTaskDependenciesIntegrationTest.groovy
@@ -18,6 +18,7 @@ package org.gradle.api.tasks
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
+import org.gradle.integtests.fixtures.UnsupportedWithConfigurationCache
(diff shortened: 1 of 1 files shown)


subprojects/core/src/integTest/groovy/org/gradle/api/tasks/MissingTaskDependenciesIntegrationTest.groovy
```diff
@@ -436,7 +439,7 @@ class MissingTaskDependenciesIntegrationTest extends AbstractIntegrationSpec imp
     }
 
     @Issue("https://github.com/gradle/gradle/issues/20391")
-    @ToBeFixedForConfigurationCache(skip = INVESTIGATE)
+    @UnsupportedWithConfigurationCache(because = "hard to reliably reproduce this corner case due to different evaluation order with CC")
```
ov7a, line 442: I tried a couple of approaches:
- property
- build service
- reordering properties

But failed to reproduce it with CC. This requires a deeper understanding of the CC-engine. But I wonder if this scenario is worth investigation in the first place?

tresat: I think marking this unsupported is the right move.

PR #36742, opened 2026-02-13
Fix CC-incompatible tests in PluginDetectionIntegrationTest

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/integTest/groovy/org/gradle/api/PluginDetectionIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/api/PluginDetectionIntegrationTest.groovy
@@ -17,7 +17,6 @@ package org.gradle.api
 
 import org.gradle.api.plugins.AppliedPlugin
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
-import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
 import org.gradle.test.fixtures.plugin.PluginBuilder
 import spock.lang.Issue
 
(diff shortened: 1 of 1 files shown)


subprojects/core/src/integTest/groovy/org/gradle/api/PluginDetectionIntegrationTest.groovy
```diff
@@ -100,8 +98,6 @@ class PluginDetectionIntegrationTest extends AbstractIntegrationSpec {
             operations << "applied"
 
             task verify { doLast { assert operations == ['applying', 'withType', 'withId', 'withPlugin', 'applied'] } }
-
-            gradle.buildFinished { loader.close() }
```
ov7a, line 104: 1. URlClassloader is autocloseable, 2. we don't care about cleaning resources in tests

mlopatkin: > 1. URlClassloader is autocloseable, 2. we don't care about cleaning resources in tests

As test failures suggest, we're leaving locked files behind.

ov7a: Hmm, I wonder why it's not handled by the test infra. I'll investigate

mlopatkin: >Hmm, I wonder why it's not handled by the test infra. I'll investigate

We don't shut down daemons between tests, AFAIK. So the `loader` remains open there and holds the `plugin.jar`. I recall JVM isn't really eager to unload classes, and these may even be stuck in some cached classloader in the daemon. An explicit `close` is probably the best solution. Actually, we don't have to wait until `buildFinished` to close the `loader`, its job is mostly done by this line anyway.

ov7a:
@bot-gradle test this

PR #36744, opened 2026-02-13
Reduce the amount of memory used by ValueState

big-guy:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/ValueState.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/ValueState.java
@@ -174,6 +174,22 @@ public abstract class ValueState<S> {
 
     public abstract void warnOnUpgradedPropertyValueChanges();
 
+    /**
+     * We create one ValueState for every Property in the build. To lower the overall cost of
+     * each Property, this implementation uses bitset flags instead of separate booleans.
+     *
+     * This also splits out ValueStates with copiers into {@link NonFinalizedValueWithCopier}.
+     * This saves memory by making the class smaller and removing padding for alignment.
+     *
+     * NOTE: Care must be taken to not increase the size of this structure.
+     *
+     * In a build like Gradle's, we have greater than 1 million instances of ValueState.
+     * Adding a single reference can greatly effect the size of this structure and increase
+     * memory consumption in a hard to see way.
+     *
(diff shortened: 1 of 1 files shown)


platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/ValueState.java
```diff
@@ -175,24 +175,24 @@ public void disallowChangesAndFinalizeOnNextGet() {
     public abstract void warnOnUpgradedPropertyValueChanges();
 
     private static class NonFinalizedValue<S> extends ValueState<S> {
+        private static final byte EXPLICIT_VALUE = 0x01;
```
bamboo: 💅 A nice pattern for bitwise constants is to use the shift operator as to make clear the relevant bit positions, e.g., `(1 << 0)`, `(1 << 1)`, etc.

big-guy: 👍

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/ValueState.java
```diff
@@ -175,24 +175,24 @@ public void disallowChangesAndFinalizeOnNextGet() {
     public abstract void warnOnUpgradedPropertyValueChanges();
 
     private static class NonFinalizedValue<S> extends ValueState<S> {
+        private static final byte EXPLICIT_VALUE = 1;
+        private static final byte FINALIZE_ON_NEXT_GET = 1 << 1;
+        private static final byte DISALLOW_CHANGES = 1 << 2;
+        private static final byte DISALLOW_UNSAFE_READ = 1 << 3;
+        private static final byte IS_UPGRADED_PROPERTY_VALUE = 1 << 4;
+        private static final byte WARN_ON_UPGRADED_PROPERTY_CHANGES = 1 << 5;
```
hegyibalint, line 199: 🤔 It is not that usual for us to use bit-fields. I would give a little heads-up above this to explain that we are doing it for performance reasons.

hegyibalint (approved):
Think of the savings! I had some non-blocking questions.

I like the implementation. Getting rid of `copier` if not necessary is also a nice catch.

Reminds me of the story where people said that Java can be as performant as C; however, your code will stop looking like Java at that point 😂

PR #36750, opened 2026-02-14
Unroll SettingsLoader implementations into a plain method

alllex:
Simplifies the code structure by getting rid of a number of small `SettingsLoader` classes and inlining their implementations into the single `SettingsPreparer`.

Delegate nesting used for arranging SettingsLoader steps was making it hard to see the entirety of what happens during the settings evaluation phase. At the same time, the loader "steps" exist for a single reason in the production code -- to allow defining a reduced set of steps used to process settings of included builds.

Simplifying this code is especially helpful because much of the logic of each "step" relies on executing side-effects, instead of returning results. Most side effects are about modifying the state of some common services, like `Gradle`.

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/initialization/DefaultSettingsPreparer.java
+++ b/subprojects/core/src/main/java/org/gradle/initialization/DefaultSettingsPreparer.java
@@ -172,7 +172,7 @@ public class DefaultSettingsPreparer implements SettingsPreparer {
             }
         }
 
-        // Add included builds defined in settings
+        // Load included builds defined in settings and those added above
         gradle.setIncludedBuilds(loadIncludedBuildsRecursively(settings));
 
         if (gradle.isRootBuild()) {
@@ -260,7 +260,6 @@ public class DefaultSettingsPreparer implements SettingsPreparer {
         // Allow a built-in command to run in a directory not contained in the settings file (but don't use the settings from that file)
         for (BuiltInCommand command : builtInCommands) {
             if (command.wasInvoked(startParameter)) {
-                // Allow built-in command to run in a directory not contained in the settings file (but don't use the settings from that file)
                 return true;
             }
         }


alllex:

subprojects/core/src/main/java/org/gradle/initialization/DefaultSettingsPreparer.java
```diff
@@ -16,18 +16,321 @@
...
+                        @Override
+                        public String getIncludedBy() {
+                            return fromBuild == null ? null : fromBuild.getBuildPath().toString();
+                        }
+                    });
+            }
+        });
+    }
+
+    private void loadBuild(GradleInternal gradle) {
```
alllex, line 151: ℹ️  The main change is here -- the `SettingsLoader` matrioshka is fully inlined into a single method.

The different between handling root and non-root builds is represented by `.isRootBuild()` checks.

subprojects/core/src/test/groovy/org/gradle/initialization/DefaultSettingsPreparerTest.groovy
```diff
@@ -0,0 +1,154 @@
...
+import org.gradle.internal.operations.BuildOperationProgressEventEmitter
+import org.gradle.internal.operations.TestBuildOperationRunner
+import org.gradle.internal.scripts.ScriptFileResolver
+import org.gradle.test.fixtures.file.TestNameTestDirectoryProvider
+import org.gradle.util.Path
+import org.gradle.util.TestUtil
+import org.junit.Rule
+import spock.lang.Specification
+
+class DefaultSettingsPreparerTest extends Specification {
```
alllex, line 43: ℹ️  This has moved from `DefaultSettingsLoaderTest` without much changes, except the required mocking.

subprojects/core/src/main/java/org/gradle/initialization/DefaultSettingsPreparer.java
```diff
@@ -16,18 +16,321 @@
...
+        projectRegistry.registerProjects(gradle.getOwner(), settings.getProjectRegistry());
+
+        if (gradle.isRootBuild()) {
+            // Add all included builds from the command-line
+            for (File includedBuildRootDir : gradle.getStartParameter().getIncludedBuilds()) {
+                settings.includeBuild(includedBuildRootDir);
+            }
+        }
+
+        // Add included builds defined in settings
```
mlopatkin: :nail_care: This "Add" has a different meaning compared to one above. 


```suggestion
        // Load included builds defined in settings and those added above
```

subprojects/core/src/main/java/org/gradle/initialization/DefaultSettingsPreparer.java
```diff
@@ -16,18 +16,321 @@
...
+    private boolean useEmptySettings(ProjectSpec spec, SettingsInternal loadedSettings, StartParameter startParameter) {
+        // Use the loaded settings if it includes the target project (based on build file, project dir or current dir)
+        if (spec.containsProject(loadedSettings.getProjectRegistry())) {
+            return false;
+        }
+
+        // Allow a built-in command to run in a directory not contained in the settings file (but don't use the settings from that file)
+        for (BuiltInCommand command : builtInCommands) {
+            if (command.wasInvoked(startParameter)) {
+                // Allow built-in command to run in a directory not contained in the settings file (but don't use the settings from that file)
```
mlopatkin: 💭Why do we have duplicate comments?

subprojects/core/src/main/java/org/gradle/initialization/DefaultSettingsPreparer.java
```diff
@@ -16,18 +16,321 @@
...
+        } else {
+            logger.debug("Loading build definition for build: '{}'", gradle.getIdentityPath());
+            state = findSettingsAndLoadIfAppropriate(gradle, startParameter, buildLayout, gradle.getClassLoaderScope());
+            SettingsInternal settings = state.getSettings();
+            spec = ProjectSpecs.forStartParameter(startParameter, settings);
+            if (useEmptySettings(spec, settings, startParameter)) {
+                // Discard the loaded settings and replace with an empty one
+                logger.debug("Discarding loaded settings and replacing with empty settings for build: '{}'", gradle.getIdentityPath());
+                state.close();
+                state = createEmptySettings(gradle, startParameter, settings.getClassLoaderScope());
```
mlopatkin, line 229: 💭This use of `settings` instead of `gradle` is suspicious, TBH; but it was reverted to the current state, so there should be some meaning to it.

mlopatkin (approved):
LGTM with a small suggestion (and some random thoughts about the pre-existing code that I couldn't hold up to myself)

subprojects/core/src/main/java/org/gradle/initialization/DefaultSettingsPreparer.java
alllex: Copying typo

subprojects/core/src/main/java/org/gradle/initialization/DefaultSettingsPreparer.java
alllex: I also find this extremely suspicious. I have been looking at this complexity from time to time, and my hands have been itching to find a way of getting rid of it. But I didn't have time to understand all the implications yet.

The good thing, I think, is that some complexities of the settings phase, including this one, are more apparent after this change.

PR #34712, opened 2025-08-18
Reuse CC upon unrelated changes to `gradle.properties`

bamboo:
By no longer tracking the content of `gradle.properties` and relying instead on the precise tracking of each property access introduced for #34562.

Fixes #20969

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #36764, opened 2026-02-16
Update release notes to cover less CC invalidations for properties

### LINKED ISSUE #20969: Project properties not used at configuration time can still invalidate configuration cache
### Expected Behavior
It would be helpful if changes to unused Gradle properties didn't invalidate the configuration cache

### Current Behavior
In AndroidX we're finding that changes to Gradle properties that aren't used at configuration time can still invalidate the configuration cache.

Also interesting, we're finding that Gradle properties that aren't used at all can still invalidate the configuration cache ( repro at https://github.com/mathjeff/gradle-samples/tree/master/configuration-cache-unused-property )

### Context

We have a feature ( https://github.com/gradle/gradle/issues/13753 ) that changes some properties of some Gradle tasks, and we'd like it to be possible to avoid reconfiguring all projects and instead to just reread the new property value during task execution.

This issue looks closely related to https://github.com/gradle/gradle/issues/20873 ; if there were a convenient way to exempt a specific Gradle property from invalidating the configuration cache, that would work too.

In the meanwhile we might end up switching from a gradle property to an environment variable, which appears to not invalidate the configuration cache, see https://android-review.googlesource.com/c/platform/frameworks/support/+/2120272

mlopatkin:
Fixes #20969.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs/src/docs/release/notes.md
+++ b/platforms/documentation/docs/src/docs/release/notes.md
@@ -326,7 +326,7 @@ even if no [project properties](userguide/build_environment.html#sec:project_pro
 Consider the following Kotlin DSL example:
 
 ```kotlin
-tasks.register("print") {
+tasks.register("printValue") {
     val value = providers.gradleProperty("value").orElse("N/A")
     doLast {
         println("value: ${value.get()}")
@@ -334,25 +334,27 @@ tasks.register("print") {
 }
 ```
 
-When running the `print` task with the Configuration Cache enabled, Gradle caches the work graph:
-```
-$ ./gradlew --configuration-cache print
-Calculating task graph as no cached configuration is available for tasks: print
+When running the `printValue` task with the Configuration Cache enabled, Gradle caches the work graph:
+```shell
+$ ./gradlew --configuration-cache printValue
 
-> Task :print
+Calculating task graph as no cached configuration is available for tasks: printValue
+
+> Task :printValue
 value: N/A
 
 ...
 Configuration cache entry stored.
 ```
 
-Previous versions of Gradle were unable to reuse this cache entry when re-executing the `print` task after changing anything in `gradle.properties`:
-```
+Previous versions of Gradle were unable to reuse this cache entry when re-executing the `printValue` task after changing anything in `gradle.properties`:
+```shell
 $ echo "value=1" >> gradle.properties
-$ ./gradlew --configuration-cache print
+$ ./gradlew --configuration-cache printValue
+
 Calculating task graph as configuration cache cannot be reused because file 'gradle.properties' has changed.
 
-> Task :print
+> Task :printValue
 value: 1
 
 ...
@@ -362,12 +364,13 @@ Configuration cache entry stored.
 In this release, Gradle now detects that only the `value` property was changed, and that this property was never used during the configuration phase.
 This allows Gradle to reuse the configuration cache entry and start executing tasks faster
 
-```
+```shell
 $ echo "value=1" >> gradle.properties
-$ ./gradlew --configuration-cache print
+$ ./gradlew --configuration-cache printValue
+
 Reusing configuration cache.
 
-> Task :print
+> Task :printValue
 value: 1
 
 ...


mlopatkin:
@bot-gradle test BD

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -318,6 +318,69 @@ configurations {
 
 Gradle provides a [Configuration Cache](userguide/configuration_cache.html) that improves build time by caching the result of the configuration phase and reusing it for subsequent builds.
 
+#### Improved hit rate for changes of project properties
```
alllex:
```suggestion
#### Improved hit rates for changes of project properties
```

💅

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -318,6 +318,69 @@ configurations {
...
+
+Previously, changing any [project property](userguide/build_environment.html#sec:project_properties) defined in `gradle.properties` files or in other places except `-P` command-line arguments resulted in invalidating the [Configuration Cache](userguide/configuration_cache.html), even if the changed property wasn't used during the configuration phase.
```
alllex: > in other places except `-P`

❓   What are the other places?

IIRC, this was all about `gradle.properties` files, albeit different ways of reading values specified there (e.g. system properties)
- https://github.com/gradle/gradle/pull/34712

Unless there are actually other places, I'd vote to update the section title to include "`gradle.properties` files" explicitly

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -318,6 +318,69 @@ configurations {
...
+With previous versions of Gradle, executing the `echo` task after changing the value of `value` in `gradle.properties` was unable to reuse the Configuration Cache:
+```
+$ ./gradlew --configuration-cache echo
+Calculating task graph as no cached configuration is available for tasks: echo
+
+> Task :echo
+value: N/A
+
+BUILD SUCCESSFUL in 319ms
+1 actionable task: 1 executed
```
alllex:
```suggestion
...
```

💅  Let's draw attention to the relevant line

platforms/documentation/docs/src/docs/release/notes.md
mlopatkin: > What are the other places?

You're right, we still invalidate when env vars/system props change.

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -318,6 +318,65 @@ configurations {
 
 Gradle provides a [Configuration Cache](userguide/configuration_cache.html) that improves build time by caching the result of the configuration phase and reusing it for subsequent builds.
 
+#### Improved hit rates for changes of project properties in `gradle.properties` files
```
alllex:
```suggestion
#### Improved hit rates for changes in `gradle.properties` files
```

I think we can drop "project properties", because we are effectively having better hit rates for cases when system-properties-specified-in-gradle-properties-files change. And maybe we also get hits when you just add a comment in the gradle.properties?

mlopatkin: > I think we can drop "project properties", because we are effectively having better hit rates for cases when system-properties-specified-in-gradle-properties-files change. And maybe we also get hits when you just add a comment in the gradle.properties?

Makes sense, I mentioned non-property changes too.

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -318,6 +318,67 @@ configurations {
 
 Gradle provides a [Configuration Cache](userguide/configuration_cache.html) that improves build time by caching the result of the configuration phase and reusing it for subsequent builds.
 
+#### Improved hit rates for changes in `gradle.properties` files
+
+Previously, changing any `gradle.properties` file resulted in invalidating the [Configuration Cache](userguide/configuration_cache.html),
+even if the [project properties](userguide/build_environment.html#sec:project_properties) weren't changed or if the changed property wasn't used during the configuration phase.
```
tresat:
```suggestion
even if no [project properties](userguide/build_environment.html#sec:project_properties) were changed, or any changed properties weren't used during the configuration phase.
```

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -318,6 +318,67 @@ configurations {
...
+value: 1
+
+...
+Configuration cache entry stored.
+```
+
+By detecting that the `value` property is never realized during the configuration phase, this release can reuse the configuration cache and make more scenarios run faster.
+```
+$ ./gradlew --configuration-cache echo
+Calculating task graph as no cached configuration is available for tasks: echo
```
tresat: It's a little odd that the "after" example begins with the "cannot be reused" message being the first thing, and that we're never shown changing the value.  It's also a little hard to see the echo output on the same line as the Gradle invocation, plus using a task named echo and the CLI `echo` tool together.

How about renaming the task to `print` and structuring this example like:

---

When running the `print` task with the configuration cache enabled, Gradle caches the work graph:

```
$ ./gradlew --configuration-cache print
Calculating task graph as no cached configuration is available for tasks: print

> Task :print
value: N/A

...
Configuration cache entry stored.
```

Previous versions of Gradle were unable to reuse this cache entry when re-executing the `print` task after changing anything in `gradle.properties`:

```
$ echo "value=1" >> gradle.properties
$ ./gradlew --configuration-cache print
Calculating task graph as configuration cache cannot be reused because file 'gradle.properties' has changed.

> Task :print
value: 1

...
Configuration cache entry stored.
```

In this release, Gradle now detects that only the `value` property was changed, and that this property was never used during the configuration phase.  This allows Gradle to reuse the configuration cache entry and start executing tasks faster:

```
$ echo "value=1" >> gradle.properties
$ ./gradlew --configuration-cache print
Reusing configuration cache.

> Task :print
value: 1

...
Configuration cache entry reused.
```

mlopatkin: I like the proposed structure more, thanks for the suggestion! Updated.

mlopatkin:
@bot-gradle test BD

PR #36765, opened 2026-02-16
De-incubate networkTimeout and test unspecified behavior

### LINKED ISSUE #27084: networkTimeout is added by the `wrapper` task but it is an incubating feature
### Current Behavior

Gradle wrapper adds `networkTimeout=10000` into `gradle-wrapper.properties` even though it is an incubating feature introduced by https://github.com/gradle/gradle/issues/17575. But this property is already optional. If not present, it would default to the previously hard-coded setting.

### Expected Behavior

Incubating features should not be forced on us. We should be able to opt in to using it. Otherwise if the feature implementation changes, all projects upgrading to Gradle 7.6+ will have to update their projects to fix the incompatibility. For example, you might rename this option to `networkTimeoutInMs` to indicate that the value is in milliseconds. (Which is not a bad thing at all, not knowing the unit is a common issue in software engineering due to how we name our variables/properties.)

### Context (optional)

_No response_

### Steps to Reproduce

* First make sure `gradle-wrapper.properties` does not have a `networkTimeout` property (remove it if it was already there).
* Then run the `wrapper` task.

### Gradle version

7.6.3

### Build scan URL (optional)

_No response_

_No response_

kamalkansal27:
The networkTimeout property is no longer written to gradle-wrapper.properties when set to the default value (10000ms). Also removes the `@Incubating` annotation from getNetworkTimeout() as the API has been stable since 7.6.

Fixes #27084

### DIFF
--- a/platforms/documentation/docs/src/docs/release/notes.md
+++ b/platforms/documentation/docs/src/docs/release/notes.md
@@ -105,6 +105,8 @@ See the User Manual section on the "[Feature Lifecycle](userguide/feature_lifecy
 
 The following are the features that have been promoted in this Gradle release.
 
+* [`getNetworkTimeout()`](javadoc/org/gradle/api/tasks/wrapper/Wrapper.html#getNetworkTimeout()) in `Wrapper`
+
 <!--
 ### Example promoted
 -->


kamalkansal27:
I ran the sanity checks as well after removing the @Incubating annotation but did not got any error as stated in the issue. Am I missing something here?

Screenshot 2026-02-16 at 7 55 52 PM

ljacomet:

ljacomet:
@kamalkansal27 I enabled the actions build, it should fail if the addition to the API changes is required.

h0tk3y:
@kamalkansal27, thanks for the proposed contribution!

We think that if the `networkTimeout` value is explicitly specified, it still makes sense to write it to the properties file even if it matches the current default. The default network timeout might change in a future Gradle version, and if the user specified it, it means they want the specific timeout value they specified, so silently changing that to the new default in a future Gradle version is undesirable. 

Could you please clarify why it would be preferable to write the value in that case anyway?

On the other hand, de-incubating the property and adding one more test to check that the network timeout is not written if unspecified look good. Would you be OK to keep just those changes?

kamalkansal27:
@h0tk3y  Thanks for the review and the detailed explanation! That makes total sense — if a user explicitly specifies the timeout, they want that exact value preserved regardless of future default changes. 

I've updated the PR to keep only:                                                                                                                       
                                                            
- De-incubation of the networkTimeout property (removed @Incubating)                                                                                                            - A new test verifying that networkTimeout is not written to the properties file when unspecified                                                                               
                                                                                                                                                                                  
The logic in WrapperGenerator is reverted so that any explicitly set value is always written, even if it matches the current default.

h0tk3y:
@bot-gradle test this

h0tk3y:
I've removed this PR from the merge queue because it de-incubates some API and for all promoted APIs we need to have a release notes entry. We need to add it there.

PR #36773, opened 2026-02-17
Use Lazy.locking() in ResolveConfigurationResolutionBuildOperationResult

wolfs:
The value might be read from two threads concurrently, causing NPEs in case we don't use locking.

We saw this happening when consuming these build operations in multiple Develocity plugins (something we do internally).

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/ResolveConfigurationResolutionBuildOperationResult.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/ResolveConfigurationResolutionBuildOperationResult.java
@@ -50,7 +50,7 @@ class ResolveConfigurationResolutionBuildOperationResult implements ResolveConfi
         AttributesFactory attributesFactory
     ) {
         this.graphSource = graphSource;
-        this.lazyDesugaredAttributes = Lazy.unsafe().of(() -> desugarAttributes(attributesFactory, requestedAttributes));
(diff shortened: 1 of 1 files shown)


platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/ResolveConfigurationResolutionBuildOperationResult.java
```diff
@@ -50,7 +50,7 @@ public ResolveConfigurationResolutionBuildOperationResult(
         AttributesFactory attributesFactory
     ) {
         this.graphSource = graphSource;
-        this.lazyDesugaredAttributes = Lazy.unsafe().of(() -> desugarAttributes(attributesFactory, requestedAttributes));
+        this.lazyDesugaredAttributes = Lazy.locking().of(() -> desugarAttributes(attributesFactory, requestedAttributes));
```
facewindu, line 53: question: is this causing a small performance regression ?
I'm asking since the problem is only happening when using the DV Gradle plugin multi application internal features, and as such, is not really a production issue or we could argue it's not a Gradle bug in itself (although fixing it at the other end, the DV Gradle plugin, is probably less easy)

wolfs: This is using double checked locking. The overhead should be minimal.

PR #36777, opened 2026-02-17
Fix non-ASCII characters in exported ascii armored keyring UIDs

### LINKED ISSUE #21612: Character encoding issue in `gradle/verification-keyring.keys`
Gradle 7.5.1

The owner of this key 👇  has a `ö` in their name
https://keyserver.ubuntu.com/pks/lookup?search=F067A2FD751AE3E4&fingerprint=on&op=index

When gradle does `--export-keys` it changes the `ö` to an `?` in the `gradle/verification-keyring.keys` file.
My guess is there is a character encoding issue.

To reproduce, add this dependency, and then do `--write-verification-metadata pgp,sha512 --export-keys --refresh-dependencies`
```
io.github.davidburstrom.contester:contester-breakpoint:0.2.0
```

ljacomet:
Use UTF-8 instead of US_ASCII when writing user IDs in the ascii armored keyring file, so that characters like ö are preserved instead of being replaced with ?.

Fixes #21612

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/verification/DependencyVerificationSignatureWriteIntegTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/verification/DependencyVerificationSignatureWriteIntegTest.groovy
@@ -539,6 +539,42 @@ class DependencyVerificationSignatureWriteIntegTest extends AbstractSignatureVer
         outputContains("Exported 1 keys to")
     }
 
+    def "exported ascii armored keyring preserves non-ascii characters in user ids"() {
+        def userId = "Björk Guðmundsdóttir <bjork@example.com>"
+        def keyring = newKeyRingWithUserId(userId)
+        keyServerFixture.registerPublicKey(keyring.publicKey)
+        createMetadataFile {
+            keyServer(keyServerFixture.uri)
+        }
+
+        given:
+        javaLibrary()
+        uncheckedModule("org", "foo", "1.0") {
+            withSignature {
+                keyring.sign(it)
+            }
+        }
+        buildFile << """
+            dependencies {
+                implementation "org:foo:1.0"
+            }
+        """
+
+        when:
+        writeVerificationMetadata()
+        succeeds ":help", "--export-keys"
+
+        then:
+        def exportedKeyRingAscii = file("gradle/verification-keyring.keys")
+        exportedKeyRingAscii.exists()
+        exportedKeyRingAscii.getText("UTF-8").contains("uid    $userId")
+
+        then:
+        // We can read the armored keyring back, the non-ASCII character is not an issue
+        succeeds "compileJava"
+
+    }
+
     @Issue("https://github.com/gradle/gradle/issues/18567")
     def "--export-keys can export keys even with without --write-verification-metadata"() {
         def keyring = newKeyRing()
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/verification/writer/WriteDependencyVerificationFile.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/verification/writer/WriteDependencyVerificationFile.java
@@ -603,7 +603,9 @@ public class WriteDependencyVerificationFile implements DependencyVerificationOv
                     List<String> userIDs = PGPUtils.getUserIDs(pk);
                     for(String uid : userIDs) {
                         hasUid = true;
-                        out.write(("uid    " + uid + "\n").getBytes(StandardCharsets.US_ASCII));
+                        // We can store UTF-8 text despite being an ASCII-armored file,
+                        // as ArmoredInputStream only cares about finding armor headers and does not otherwise care about encoding
+                        out.write(("uid    " + uid + "\n").getBytes(StandardCharsets.UTF_8));
                     }
                     if (hasUid) {
                         out.write('\n');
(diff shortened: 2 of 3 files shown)


platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/verification/writer/WriteDependencyVerificationFile.java
```diff
@@ -603,7 +603,8 @@ private void writeAsciiArmoredKeyRingFile(File ascii, Collection<PGPPublicKeyRin
                     List<String> userIDs = PGPUtils.getUserIDs(pk);
                     for(String uid : userIDs) {
                         hasUid = true;
-                        out.write(("uid    " + uid + "\n").getBytes(StandardCharsets.US_ASCII));
+                        // Even though ArmoredOutputStream is ASCII-armoring the output, we want to be able to support non-ASCII user ids
```
octylFractal: 💭 I think it might be more appropriate to say something like:
```suggestion
                        // We can store UTF-8 text despite being an ASCII-armored file, as ArmoredInputStream only cares about finding armor headers and does not otherwise care about encoding
```
to be clear that the reason this is OK is because it will be ignored when reading the file later.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/verification/DependencyVerificationSignatureWriteIntegTest.groovy
```diff
@@ -539,6 +539,37 @@ class DependencyVerificationSignatureWriteIntegTest extends AbstractSignatureVer
         outputContains("Exported 1 keys to")
     }
 
+    def "exported ascii armored keyring preserves non-ascii characters in user ids"() {
```
ov7a, line 542: ❓ This test is about writing UTF-8.

What about reading? Can a keyring file, containing a UTF-8 record, be correctly updated?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/verification/DependencyVerificationSignatureWriteIntegTest.groovy
```diff
@@ -539,6 +539,37 @@ class DependencyVerificationSignatureWriteIntegTest extends AbstractSignatureVer
...
+                implementation "org:foo:1.0"
+            }
+        """
+
+        when:
+        writeVerificationMetadata()
+        succeeds ":help", "--export-keys"
+
+        then:
+        def exportedKeyRingAscii = file("gradle/verification-keyring.keys")
```
ov7a, line 568: 💭 [RFC 4880](https://www.rfc-editor.org/rfc/rfc4880.html) does not specify anything directly about the text encoding of a keyring file. From what I found quickly, most implementations don't care, but there's a tiny chance that some implementations would require ASCII. I think it's an acceptable risk.

octylFractal: UTF-8 is specified at https://www.rfc-editor.org/rfc/rfc4880.html#section-3.4.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/verification/DependencyVerificationSignatureWriteIntegTest.groovy
ljacomet: Added a step to the test confirming that the keyring can be used.

ov7a: It's better, but I think it would be nice to test all chain: read -> write -> read to ensure nothing is lost

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/verification/writer/WriteDependencyVerificationFile.java
```diff
@@ -603,7 +603,9 @@ private void writeAsciiArmoredKeyRingFile(File ascii, Collection<PGPPublicKeyRin
                     List<String> userIDs = PGPUtils.getUserIDs(pk);
                     for(String uid : userIDs) {
                         hasUid = true;
-                        out.write(("uid    " + uid + "\n").getBytes(StandardCharsets.US_ASCII));
+                        // We can store UTF-8 text despite being an ASCII-armored file,
+                        // as ArmoredInputStream only cares about finding armor headers and does not otherwise care about encoding
+                        out.write(("uid    " + uid + "\n").getBytes(StandardCharsets.UTF_8));
```
big-guy, line 608: Do we ever read this anywhere?

ljacomet: Gradle reads the file when performing verification, to loads keys.

I have added a step to the test to make sure those characters are not an issue.

big-guy: It looks like org.gradle.security.internal.PGPUtils#getUserIDs and PGPPublicKey always reads them as UTF8.

I was imagining a scenario where the reader had a non-UTF8 charset, so we'd write them as UTF8 and then read them back in with windows-1252 or something.

PR #36781, opened 2026-02-17
Cache parsed private keys in a build service

mlopatkin:
Today, in default configuration the signing plugin creates an instance of `PgpSignatory` for every artifact it signs. Prior to this PR, every signatory was eagerly reading a private key from keystore. This is an expensive operation, mostly because of deriving the decryption key from the passphrase to obtain the private key. Reading the keystore also had a non-trivial cost.

Reducing the number of `PgpSignatory` is not easy: the signing plugin relies on project properties to look up the key, which can be changed at any time. The plugin is also very flexible and exposes broad public API, so the interaction between moving parts is very complicated. Reusing signatories can therefore easily break existing builds.

This PR introduces a build-level cache for secret keys in a build service. The cache mostly respects the timeline of reconfiguring the keys: we snapshot the values of relevant properties at the same time as before. Then we use these values as cache keys to reuse already loaded private keys. 

We also reuse the same infrastructure to cache private keys parsed from in-memory representations. Before we only had a project-level cache for those, so multi-module projects with many artifacts might have still suffered longer configuration/execution times. One downside is that the armored string representation of the key sits in memory for longer than before, even after the key has been loaded, but the string interning service mitigates most of the overhead.

In order to use the caching build service correctly, we had to add some plumbing and introduce `PgpSignatoryInternal`.

With Configuration Cache, we only parse the keys at execution time, unless someone examines `PgpSignatory` at configuration time for some reason. The string interning service is discarded at the end of configuration phase, but the armored keys are still deduplicated by CC upon load, like any strings.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/internal/pgp/PgpSignatoryService.java
+++ b/platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/internal/pgp/PgpSignatoryService.java
@@ -40,11 +40,19 @@ public abstract class PgpSignatoryService implements BuildService<BuildServicePa
     private final ConcurrentMap<PgpKeyHandle, ParsedPgpKey> cachedKeys = new ConcurrentHashMap<>();
 
     public ParsedPgpKey readSecretKeyFromKeyRingFile(String keyId, File file, String password) {
-        return cachedKeys.computeIfAbsent(new PgpKeyInKeyringFile(keyId, file, password), handle -> new ParsedPgpKey(PgpSignatoryUtil.readSecretKey(keyId, file), password));
+        return cachedKeys.computeIfAbsent(
+            new PgpKeyInKeyringFile(keyId, file, password),
+            handle ->
+                new ParsedPgpKey(Lazy.locking().of(() -> PgpSignatoryUtil.readSecretKey(keyId, file)), password)
+        );
     }
 
     public ParsedPgpKey readSecretKeyFromArmoredString(@Nullable String keyId, String keyData, String password) {
-        return cachedKeys.computeIfAbsent(new PgpKeyData(keyId, keyData, password), handle -> new ParsedPgpKey(PgpSignatoryUtil.parseSecretKey(keyId, keyData), password));
+        return cachedKeys.computeIfAbsent(
+            new PgpKeyData(keyId, keyData, password),
+            handle ->
+                new ParsedPgpKey(Lazy.locking().of(() -> PgpSignatoryUtil.parseSecretKey(keyId, keyData)), password)
+        );
     }
 
     private interface PgpKeyHandle {}
@@ -106,16 +114,16 @@ public abstract class PgpSignatoryService implements BuildService<BuildServicePa
     }
 
     public static class ParsedPgpKey {
-        private final PGPSecretKey secretKey;
+        private final Lazy<PGPSecretKey> secretKey;
         private final Lazy<PGPPrivateKey> privateKey;
 
-        private ParsedPgpKey(PGPSecretKey secretKey, String password) {
+        private ParsedPgpKey(Lazy<PGPSecretKey> secretKey, String password) {
             this.secretKey = secretKey;
-            privateKey = Lazy.locking().of(() -> PgpSignatoryUtil.extractPrivateKey(secretKey, password));
+            this.privateKey = Lazy.locking().of(() -> PgpSignatoryUtil.extractPrivateKey(secretKey.get(), password));
         }
 
         public PGPSecretKey getSecretKey() {
-            return secretKey;
+            return secretKey.get();
         }
 
         public PGPPrivateKey getPrivateKey() {


mlopatkin:

mlopatkin:
@bot-gradle test PRF, PT without PTS

platforms/software/signing/src/test/groovy/org/gradle/plugins/signing/SignatoriesConfigurationSpec.groovy
```diff
@@ -69,13 +69,14 @@ class SignatoriesConfigurationSpec extends SigningProjectSpec {
...
+        given:
         project.ext["signing.keyId"] = "aaaaaaaa"
         project.ext["signing.secretKeyRingFile"] = "i/dont/exist"
         project.ext["signing.password"] = "anything"
 
-        and:
-        signing.signatory
+        when:
+        // Just getting the signatory doesn't always load the keys
+        signing.signatory.keyId
```
mlopatkin, line 79: 💭 This is a bit of a behavior change, the failure now happens later.

.idea/misc.xml
```diff
@@ -1,13 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
abstratt, line 1: ❓ 

Is this change intended?

mlopatkin: > ❓
> 
> Is this change intended?

Yes, it suppresses "unused method" warning for stuff annotated with `@ServiceReference`. IDEA also decided to sort things alphabetically here.

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/internal/pgp/PgpSignatoryUtil.java
```diff
@@ -0,0 +1,184 @@
...
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.gradle.internal.Cast.uncheckedCast;
+import static org.gradle.internal.IoActions.uncheckedClose;
+
+/**
+ * The internal implementation of PGP signatory handling.
+ * This class exists to share implementation between {@link PgpSignatoryService} and {@link PgpSignatoryFactory}.
+ */
+public class PgpSignatoryUtil {
```
abstratt, line 54: ✏️ 

Methods below (or most of them) were moved from `ProviderSignatoryFactory`

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/internal/pgp/PgpSignatoryUtil.java
```diff
@@ -0,0 +1,184 @@
...
+     * Parses the secret key in ASCII-armored format.
+     * The {@code keyData} may contain one or more master keys with subkeys.
+     * If there is more than one master key or a subkey is needed, the {@code keyId} must be provided.
+     * If omitted, the keyData is treated as a single master key (with subkeys) and the master secret key is returned.
+     *
+     * @param keyId the id of the key, can be null to return the only master secret key
+     * @param keyData the ASCII-armored representation of the key
+     * @return the parsed secret key
+     */
+    public static PGPSecretKey parseSecretKey(@Nullable String keyId, String keyData) {
```
abstratt, line 168: ✏️ 

Moved/renamed from https://github.com/gradle/gradle/pull/36781/changes#diff-f2cd21431c3e9fec103b97e84cf98153b996995ed664f590526be57126f51783L76

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/internal/pgp/PgpSignatoryUtil.java
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * Extracts private key from a secret key using provided password.
+     *
+     * @param secretKey the secret key that contains the private key
+     * @param password the password
+     * @return the decrypted private key
+     * @throws RuntimeException if the key cannot be decrypted
+     */
+    public static PGPPrivateKey extractPrivateKey(PGPSecretKey secretKey, String password) {
```
abstratt, line 149: From: https://github.com/gradle/gradle/pull/36781/changes#diff-923ef42b6461b08575994b8935de1df7a2881c1ccebdbd68a5bf2836604226afL109

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/pgp/PgpSignatory.java
```diff
@@ -39,13 +40,17 @@
...
-    public PgpSignatory(String name, PGPSecretKey secretKey, String password) {
+    PgpSignatory(String name, Provider<PGPSecretKey> secretKey, Provider<PGPPrivateKey> privateKey) {
         this.name = name;
         this.secretKey = secretKey;
-        this.privateKey = createPrivateKey(secretKey, password);
+        this.privateKey = privateKey;
+    }
+
+    public PgpSignatory(String name, PGPSecretKey secretKey, String password) {
+        this(name, Providers.of(secretKey), Providers.of(PgpSignatoryUtil.extractPrivateKey(secretKey, password)));
```
abstratt, line 64: ❓ 

No benefit in extracting private key lazily?

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/pgp/PgpSignatory.java
```diff
@@ -75,7 +80,7 @@ public void sign(InputStream toSign, OutputStream signatureDestination) {
 
     @Override
     public String getKeyId() {
-        PgpKeyId id = new PgpKeyId(secretKey.getKeyID());
+        PgpKeyId id = new PgpKeyId(secretKey.get().getKeyID());
```
abstratt, line 94: 💭 

Maybe extract accessors for getting the resolved secret and private keys

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/internal/pgp/PgpSignatoryService.java
```diff
@@ -0,0 +1,125 @@
...
+    private static final String SERVICE_NAME = PgpSignatoryService.class.getName();
+
+    public static Provider<PgpSignatoryService> obtain(Project project) {
+        return project.getGradle().getSharedServices().registerIfAbsent(SERVICE_NAME, PgpSignatoryService.class);
+    }
+
+    private final ConcurrentMap<PgpKeyHandle, ParsedPgpKey> cachedKeys = new ConcurrentHashMap<>();
+
+    public ParsedPgpKey readSecretKeyFromKeyRingFile(String keyId, File file, String password) {
+        return cachedKeys.computeIfAbsent(new PgpKeyInKeyringFile(keyId, file, password), handle -> new ParsedPgpKey(PgpSignatoryUtil.readSecretKey(keyId, file), password));
```
octylFractal: :x: This does I/O under a lock on the table, which can bottleneck whenever two keys end up in the same bucket (more likely the smaller the map is). This should be done outside of a `computeIfAbsent`.

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/pgp/PgpSignatoryFactory.java
```diff
@@ -73,14 +76,94 @@ public PgpSignatory createSignatory(String name, PGPSecretKey secretKey, String
...
+     * @param name the name for the signatory
+     * @param keyId the key id to look it up in the keyring
+     * @param keyRing the keyring file
+     * @param password the password for the private key
+     * @return the signatory instance
+     *
+     * @since 9.5.0
+     */
+    @Incubating
+    public PgpSignatory createSignatory(Project project, String name, String keyId, File keyRing, String password) {
```
abstratt, line 94: ❓ 

Why do we need to expose these two methods, Mike?

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/pgp/PgpSignatoryFactory.java
```diff
@@ -73,14 +76,94 @@ public PgpSignatory createSignatory(String name, PGPSecretKey secretKey, String
...
+    @Incubating
+    public PgpSignatory createSignatory(Project project, String name, @Nullable String keyId, String key, String password) {
+        return createLazySignatory(project, name, keyId, key, password);
+    }
+
     protected PgpSignatory readProperties(Project project, String prefix, String name) {
         return readProperties(project, prefix, name, false);
     }
 
+    public PGPSecretKey readSecretKey(final String keyId, final File file) {
```
abstratt, line 121: 💭 

I'd avoid adding any more public/protected elements unless required. I guess they may be, and I am just not seeing it

abstratt: Ah, they seem to have been moved from below...

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/pgp/PgpSignatoryFactory.java
mlopatkin: > ❓
> 
> Why do we need to expose these two methods, Mike?

I need to access them in `BaseInMemoryPgpSignatoryProvider` and `BasePgpSignatoryProvider` respectively, which reside in other package. There is an annoying entanglement of this class, `PgpSignatoryInternal`, and `PgpSignatory`; we have to expose some internal bits one way or another. In this case I preferr to hide the provider-accepting constructor of `PgpSignatory`, it would be worse to expose that one, as it is tricky to get right and tempting to use.

I feel that this factory is a lost cause anyway.

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/pgp/PgpSignatory.java
mlopatkin: > ❓
> 
> No benefit in extracting private key lazily?

We don't use this constructor anymore. May be helpful for folks trying to create these things manually.

abstratt: This is to replace the existing eager case... I guess we want to do extraction right away so failure happens now and not only when the private key is needed

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/internal/pgp/PgpSignatoryService.java
mlopatkin: > ❌ This does I/O under a lock on the table, which can bottleneck whenever two keys end up in the same bucket (more likely the smaller the map is). This should be done outside of a `computeIfAbsent`.

Nice catch! I moved it into a lazy wrapper.

platforms/software/signing/src/main/java/org/gradle/plugins/signing/signatory/pgp/PgpSignatoryFactory.java
```diff
@@ -73,14 +76,94 @@ public PgpSignatory createSignatory(String name, PGPSecretKey secretKey, String
...
+    }
+
+    private PgpSignatory createLazySignatory(Project project, String name, String keyId, String keyData, String password) {
+        // keyData may be several KB long, and each project may try to register one, though with the same contents.
+        // Other strings aren't as long/critical and interning them will likely hurt performance more than help.
+        String internedKeyData = KeyDataInternService.obtain(project).get().internKeyData(keyData);
+        return createLazySignatory(project, name, service -> service.readSecretKeyFromArmoredString(keyId, internedKeyData, password));
+    }
+
+    private PgpSignatory createLazySignatory(Project project, String name, SerializableTransformer<PgpSignatoryService.ParsedPgpKey, PgpSignatoryService> keyBuilder) {
```
abstratt, line 150: ❓ 

Couldn't this method be static in some internal class/package (same for the methods that delegate to it)?

Then `BaseInMemoryPgpSignatoryProvider` and `BasePgpSignatoryProvider` could use the internal implementation directly, no?

mlopatkin: > ❓
> 
> Couldn't this method be static in some internal class/package (same for the methods that delegate to it)?
> 
> Then `BaseInMemoryPgpSignatoryProvider` and `BasePgpSignatoryProvider` could use the internal implementation directly, no?

Not unless we either make `PgpSignatoryInternal` or provider-accepting constructor of `PgpSignatory` public. I don't want to do that.

abstratt (approved):

PR #36782, opened 2026-02-17
Compile internal-integ-testing and test fixtures to Java 17

jvandort:
By default, test suites now depend on internal-distribution-testing instead of internal-integ-testing.
Now, only the integ test suite depends on internal-integ-testing by default.
This, in combination with compiling test fixtures to Java 17 by default, allows us to compile internal-integ-testing to java 17.

The TAPI tests still compile against a small subset of test fixtures: build-process-services and problems-api.

Some changes had to be made to some TAPI cross version tests. Specifically, we had to inline two test fixtures from core.
There isn't really other good places for these sorts of test fixtures at the moment, but given that there is not a lot of
test fixtures like this, I felt it was okay to do so. These fixtures were mostly responsible for writing build scripts and
did not encapsulate some other higher level utility

The reasoning behind this change is that internal-integ-testing has compile-time dependencies on most of the gradle distribution,
as it is responsible for implementing the embedded test executor, as well as contains many test fixtures for different projects.
This is a major step towards getting more production projects to compile to Java 17

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test rfr

jvandort:
@bot-gradle test ast

jvandort:
@bot-gradle test rfr

jvandort:
@bot-gradle test rfr

build-logic/integration-testing/src/main/kotlin/gradlebuild.test-fixtures.gradle.kts
```diff
@@ -34,10 +34,7 @@ plugins {
 }
 
 jvmCompile {
-    addCompilationFrom(sourceSets.named("testFixtures")) {
-        // By default, test fixtures compile to the same JVM version as the production code.
-        targetJvmVersion = compilations.named("main").flatMap { it.targetJvmVersion }
-    }
+    addCompilationFrom(sourceSets.named("testFixtures"))
```
jvandort, line 37: Now these test fixtures use the default java version, which is 17

jvandort:
@bot-gradle test acv

jvandort:
@bot-gradle test act

PR #36783, opened 2026-02-17
Include requested capabilities when attributes fail to match

jvandort:
It can be very confusing when you get an attribute matching failure but the message does not mention that the original request was for a _non default_ capability.

For example, previously if you were asking for test fixtures but got a java version attribute matching failure, test fixtures would not appear in the failure at all. Now it does

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesLocalComponentIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesLocalComponentIntegrationTest.groovy
@@ -157,7 +157,7 @@ class CapabilitiesLocalComponentIntegrationTest extends AbstractIntegrationSpec 
         fails("resolve")
 
         then:
-        failure.assertHasCause("Unable to find a variant of 'project :other' with the requested capabilities: [coordinates 'org:capability', feature 'foo']")
+        failure.assertHasCause("Unable to find a variant of 'project :other' with the requested capabilities: ['org:capability', feature 'foo']")
     }
 
     @Issue("https://github.com/gradle/gradle/issues/26377")
@@ -309,8 +309,8 @@ class CapabilitiesLocalComponentIntegrationTest extends AbstractIntegrationSpec 
 
         where:
         requested               | description
-        ["org:foo"]             | "with capability coordinates 'org:foo'"
-        ["org:foo", "org:bar"]  | "with capabilities [coordinates 'org:bar', coordinates 'org:foo']"
+        ["org:foo"]             | "with capability 'org:foo'"
+        ["org:foo", "org:bar"]  | "with capabilities ['org:bar', 'org:foo']"
     }
 
 }
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/VariantsDependencySubstitutionRulesIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/VariantsDependencySubstitutionRulesIntegrationTest.groovy
@@ -176,7 +176,7 @@ class VariantsDependencySubstitutionRulesIntegrationTest extends AbstractIntegra
         fails ':checkDeps'
 
         then:
-        failure.assertHasCause "Unable to find a variant of 'org:lib:1.0' with the requested capability: coordinates 'org:lib-test-fixtures':"
+        failure.assertHasCause "Unable to find a variant of 'org:lib:1.0' with the requested capability: 'org:lib-test-fixtures':"
     }
 
     def "can substitute a project dependency without capabilities with a dependency with capabilities"() {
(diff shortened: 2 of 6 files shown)


jvandort:
@bot-gradle test this

platforms/software/dependency-management/src/main/java/org/gradle/internal/component/resolution/failure/describer/NoCompatibleVariantsFailureDescriber.java
```diff
@@ -60,13 +63,28 @@ public VariantSelectionByAttributesException describeFailure(NoCompatibleVariant
...
+                    .map(it -> style(StyledTextOutput.Style.Info, it.getDisplayName()))
+                    .sorted()
+                    .collect(Collectors.joining(", "));
+                requestTarget += " with capabilities [" + styledSelectors + "]";
+            }
+            formatter.node("No matching variant of " + requestTarget + " was found. The consumer was configured to find " + describer.describeAttributeSet(failure.getRequestedAttributes().asMap()) + " but:");
         }
 
+        // TODO: We should split the candidate variants into those that match the requested capabilities and those that do not.
+        //       The use likely cares about only variants that match the requested capabilities.
```
tresat, line 86: ➕

tresat: Though it's probably unlikely that there are a large number of non-matching variants with different non-matching capabilities.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesLocalComponentIntegrationTest.groovy
```diff
@@ -264,4 +264,53 @@ class CapabilitiesLocalComponentIntegrationTest extends AbstractIntegrationSpec
...
+        fails("resolve")
+
+        then:
+        failure.assertHasCause("No matching variant of project :other $description was found")
+        failure.assertHasErrorOutput("Incompatible because this component declares attribute 'org.gradle.category' with value 'foo' and the consumer needed attribute 'org.gradle.category' with value 'bar'")
+
+        where:
+        requested               | description
+        ["org:foo"]             | "with capability coordinates 'org:foo'"
+        ["org:foo", "org:bar"]  | "with capabilities [coordinates 'org:bar', coordinates 'org:foo']"
```
tresat: 🤔 I don't like the word "coordinates" here, it seems like just noise.

The multiple capabilities message might be misconstrued to mean "any of these capabilities", maybe it should say "with all capabilities"?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesLocalComponentIntegrationTest.groovy
```diff
@@ -264,4 +264,53 @@ class CapabilitiesLocalComponentIntegrationTest extends AbstractIntegrationSpec
         expect:
         succeeds("resolve")
     }
+
+    def "useful error message when target component has matching capability but incorrect attributes"() {
```
tresat, line 268: ❔ Is there any need to also test this on an external component?  It seems like it would work the same.

tresat:
One final thought - there should probably be a test/tests for matching attributes and non-matching capabilities.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesLocalComponentIntegrationTest.groovy
jvandort: It should work the same, this was just the most convenient place to put this test

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesLocalComponentIntegrationTest.groovy
jvandort: Louis also expressed he did not really like this wording. I've updated `DefaultSpecificCapabilitySelector` display name to remove `coordinates`

jvandort:
@bot-gradle test this

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesLocalComponentIntegrationTest.groovy
tresat: I like the new wording.

PR #36789, opened 2026-02-18
Fix issue with manifest content charset being lost with CC enabled

### LINKED ISSUE #31838: Setting JAR manifest encoding doesn't work with configuration cache
### Current Behavior

The specified manifest encoding on the `Jar` task  is ignored when configuration cache is enabled (up to 9.0) or causes Gradle to gracefully degrade to vintage (since 9.0). 

### Expected Behavior

The same specified encoding is used regardless of CC state

### Context (optional)

_No response_

See [this test](https://github.com/gradle/gradle/blob/0043c0c7983822dfa9d0e015e7ef09ef09880e3f/platforms/jvm/language-jvm/src/integTest/groovy/org/gradle/api/tasks/bundling/JarEncodingIntegrationTest.groovy#L136):
```gradle
task jar(type: Jar) {
    from file('test')
    destinationDirectory = file('dest')
    archiveFileName = 'test.jar'
    manifestContentCharset = 'ISO-8859-15'
    manifest {
        attributes 'moji': 'bak€'
    }
}
```

### Gradle version

* up to 8.x (custom charset is ignored)
* 9.0 (Gradle falls back to vintage) 

### Build scan URL (optional)

_No response_

_No response_

ghale:
This simplifies the manifest charset management in light of the fact that a custom manifest can be set on the `Jar` task.  It does the following:

- Changes the underlying storage of manifest content charset to be backed by a property rather than a string.  Note that this does not take the extra step of exposing manifest content charset as a property in the public API, it only changes the underlying implementation.  Exposing it in the public API is orthogonal and can be done as follow up work.
- Changes the implementations of `ManifestInternal` to accept a `Provider<String>` rather than maintaining their own copies of the string.  
- Removes the mutability of the charset on `ManifestInternal` (i.e. the only way to set the value of the manifest charset is by changing the value on `Jar`)
- Moves the implementation of actually writing the manifest into a dedicated class in `ManifestInternal` that can be used by both `DefaultManifest` and `CustomManifestInternalWrapper`

Fixes #31838 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

platforms/jvm/platform-jvm/src/main/java/org/gradle/api/java/archives/internal/DefaultManifest.java
```diff
@@ -256,36 +202,15 @@ public List<ManifestMergeSpec> getMergeSpecs() {
         return manifestMergeSpecs;
     }
 
-    public boolean isEqualsTo(Object o) {
```
ghale, line 259: This method was not being used anywhere

ghale:
@bot-gradle test this

ghale:
> Do we need to add any new smoke tests?

Per our conversation, I don't think so.  It was the smoke tests that we already have that caught the issue with changing the `DefaultManifest` constructor.

Issue #36776, opened 2026-02-17
It should be possible to request `DeclarativeSchemaModel` with a correct settings file and failures in project schema building


h0tk3y:
### Expected Behavior

When there is an unsupported member in the definitions included into the project schema, the IDE should be able to get the `DeclarativeSchemaModel` tooling model without a failure, and the model should contain at least the settings schema, but 
also maybe a partial project schema with the failures listed in it.

### Current Behavior (optional)

Building the `DeclarativeSchemaModel` fails because the unsupported definition is also employed in the `defaults` schema for the settings. The tooling model builder triggers evaluation of settings, which fails, and it does not get to building the model.

### Context

(none)

h0tk3y:
@bot-gradle test RFN

PR #36794, opened 2026-02-18
Report schema building issues as problems

h0tk3y:
* Expose the schema building failure details in the model (needs https://github.com/gradle/gradle/issues/36776 to work resiliently)
* Report schema building issues with Problem API

### DIFF
--- a/platforms/core-configuration/declarative-dsl-core/src/main/kotlin/org/gradle/internal/declarativedsl/schemaBuilder/DataSchemaBuilder.kt
+++ b/platforms/core-configuration/declarative-dsl-core/src/main/kotlin/org/gradle/internal/declarativedsl/schemaBuilder/DataSchemaBuilder.kt
@@ -196,7 +196,7 @@ class DefaultSchemaBuildingHost(override val topLevelReceiverClass: KClass<*>) :
     }
 
     override fun membersWithFailures(kClass: KClass<*>): Map<SupportedCallable, Iterable<SchemaResult.Failure>> =
-        failedMembers[kClass] ?: emptyMap()
+        failedMembers[kClass]?.toSortedMap(compareBy { it.toString() }) ?: emptyMap()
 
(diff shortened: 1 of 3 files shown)


h0tk3y:
@bot-gradle test RFN

h0tk3y:
@bot-gradle test RFN

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/provider/DeclarativeDslScriptPluginFactory.kt
```diff
@@ -43,10 +46,34 @@ class DeclarativeDslScriptPluginFactory @Inject constructor(
...
+
+            else -> emptyList() // TODO: report all other DCL failures as problems
+        }
+    }
+    problems.reporter.report(failureProblems)
+    /**
+     * Instead of [org.gradle.api.problems.ProblemReporter.throwing], we just throw
+     * this single exception here to avoid duplicating the full message of the
+     * exception on each of the problems.
+     */
```
donat, line 76: So there's a custom console rendering for dcl-related issues and there's the rendering for problems, hence the duplication. Is that correct? Would it make sense to simply remove the console rendering?

h0tk3y:
@bot-gradle test RFN

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/provider/DeclarativeDslScriptPluginFactory.kt
h0tk3y: Yes, the duplication would be caused by the exception having the full list of schema building issues attached to each of the problems and thus rendered with details along each of the problems.

Do you suggest throwing an exception here without any details about schema building issues? Is that considered a good practice with Problem API?

PR #36800, opened 2026-02-19
packaging: GenerateClasspathModuleProperties: skip generating aliases for distribution modules that refer to projects for now

cobexer:
* gradle/gradle#36169

Introduced it and in CI builds the `alias.version` of every property file related to any Gradle distribution module contains a timestamp in addition to the base version.

As that timestamp differs between builds we experience cache misses and as a result spend more time executing tests that could be avoided.

### DIFF
--- a/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/GenerateClasspathModuleProperties.kt
+++ b/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/GenerateClasspathModuleProperties.kt
@@ -35,6 +35,7 @@ import org.gradle.api.tasks.CacheableTask
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Nested
+import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.OutputDirectory
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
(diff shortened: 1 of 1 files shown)


cobexer:
@bot-gradle test RfR

cobexer:
Considering #36169 got merged to 9.4.0 I rebased the commit in this PR on top of `release` to enable us merging that into `release` instead of `master` and thus never ship a version that had that version property in the first place.

@jvandort PLMK if that would be a benefit for us

jvandort:
This version will come in handy when we need to load the gradle abi jar from the distribution (once we need ip packaging it in the distribution)

I think we should instead write the base version into the distribution, without the timestamp

cobexer:
Do we already know when that will happen?
Do we know if the `alias.version` property will be the exact thing we need then?
Do we already have a spec for that?

If not, let's delete the version now and introduce it again - in a cache compatible / friendly way - if and when we need it.

jvandort:
It is something we will need for plugin development for the provider api migration, as this is important for compiling plugins against multiple gradle versions. 
There is no spec, but there is a prototype, which depends on the alias (and version) https://github.com/gradle/gradle/pull/36205

The idea is to allow certain dependencies to be loaded from the distribution using their public coordinates. This includes external dependencies too. 

The point is, the alias is useless without the version. If we want to remove the entire alias at the moment, and introduce it later, fine. But, removing just the version is either too little, or not enough.

jvandort:
I am fine removing the entire alias functionality and introducing it later when we need it. But, we should not remove just the version

PR #36805, opened 2026-02-19
Fix IvyJvmLibraryArtifactResolutionIntegrationTest test to run with Configuration Cache

tresat:


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ivy/IvyJvmLibraryArtifactResolutionIntegrationTest.groovy
```diff
@@ -264,8 +263,10 @@ Searched in the following locations:
   - ${module.ivy.uri}""")
 
         when:
-        server.resetExpectations()
-        module.ivy.expectGetMissing()
+        if (!GradleContextualExecuter.isConfigCache()) {
```
mlopatkin, line 266: 💅 Might be worth it to add a comment, why we don't need to reset.
```suggestion
        if (!GradleContextualExecuter.isConfigCache()) {
            // With CC the build caches resolution failure, so rerunning the task makes no new resolution attempt.
```

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ivy/IvyJvmLibraryArtifactResolutionIntegrationTest.groovy
```diff
@@ -264,8 +263,10 @@ Searched in the following locations:
   - ${module.ivy.uri}""")
 
         when:
```
mlopatkin, line 265: 🤔 I wonder why we need to run the resolution twice. With CC it isn't clear if we're examining what we intend to - the exception is simply cached and replayed.

tresat: I wasn't sure either.  It's a very old test, and I figured I'd just add the minimal change necessary for CC for now.  Many of the tests in this class follow (unnecessarily?) the same pattern.

PR #36806, opened 2026-02-19
Fix ComponentSelectionRulesDependencyResolveIntegTest test to work with Configuration Cache

tresat:


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ivy/ComponentSelectionRulesDependencyResolveIntegTest.groovy
```diff
@@ -398,11 +395,14 @@ Required by:
...
 
-            tasks.checkDeps.doLast {
-                def artifacts = configurations.conf.incoming.artifacts.artifacts
-                assert artifacts.size() == 2
-                assert artifacts[0].id.componentIdentifier.version == '${chosen}'
-                assert candidates == ${candidates}
+            tasks.checkDeps {
+                def avArtifacts = configurations.conf.incoming.artifacts
+                doLast {
+                    def artifacts = avArtifacts.artifacts
```
octylFractal: 💅 This isn't an `ArtifactView`, so I think just calling it `artifactCollection` is warranted.

```suggestion
                def artifactCollection = configurations.conf.incoming.artifacts
                doLast {
                    def artifacts = artifactCollection.artifacts
```

tresat: Copy-paste error, good catch, will update.

PR #36815, opened 2026-02-20
Merge filtering strategies for class-based and non-class-based tests

### LINKED ISSUE #36790: Build failing due to "Please remove path-based exclude patterns when running only class-based tests" after 9.4.0-rc-1 update
### Current Behavior
Gradle 9.4.0-rc-1 fails a range of Elasticsearch YAML REST test tasks when muted tests are applied. Example failure:

"Exclude pattern 'org.elasticsearch.smoketest.MlWithSecurityIT.test {yaml=ml/sparse_vector_search/Test sparse_vector search with query vector and pruning config}' is path-based, but no non-class-based tests were found. Please remove path-based exclude patterns when running only class-based tests."

These tasks are class-based (JUnit/Randomized Runner) but include parameterized YAML test display names like `test {yaml=...}`. Gradle 9.4 classifies these as path-based filters and now errors out if the task does not configure non-class-based tests.

### Expected Behavior

Class-based YAML REST tests that use parameterized display names should continue to work as they did in 9.3.x. Excluding `test {yaml=...}` should not be rejected when the task is class-based. The build should not fail due to validation of `excludeTestsMatching` on these display names.

### Context (optional)

Elasticsearch runs YAML REST tests via a class-based runner in `test/yaml-rest-runner` (see `test/yaml-rest-runner/src/main/java/org/elasticsearch/test/rest/yaml/ESClientYamlSuiteTestCase.java`). Individual YAML test cases are represented as parameterized tests with display names like `test {yaml=path/to/test}`. We also centralize muted tests in `muted-tests.yml`, applied to all `Test` tasks via a build service. Gradle 9.4's new non-class-based test support appears to misclassify these display names as path-based filters, breaking existing behavior across multiple QA tasks.

Reproducer repository: https://github.com/breskeby/elasticsearch/

Steps:
1. `git clone https://github.com/breskeby/elasticsearch/`
2. `cd elasticsearch`
3. `git fetch origin update-gradle-wrapper-940`
4. `git checkout update-gradle-wrapper-940`
5. `./gradlew :x-pack:qa:multi-node:javaRestTest --stacktrace`

### Gradle version

9.4.0-rc-1

9.3.1

### Build scan URL (optional)

_No response_

- macOS 26.2 (aarch64)
- JDK 21.0.7+6-LTS (Temurin)
- Command example: `./gradlew :x-pack:qa:multi-node:javaRestTest --stacktrace`

big-guy:
Instead of assuming that any filter that contains a / to represent a filter for a non-class based test, Gradle converts resource-based test sources into class name like ids and applies the class filtering to that.

Fixes #36790 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
@@ -987,8 +987,9 @@ The goal of following the same selection rules as class-based tests is to make i
 
 The conversion from file path to fully qualified name will be ambiguous for some directory structures.
 Directory names that contain dots or file names that contain multiple dots can make it impossible to select an individual test file.
-Some directory or file names may be awkward to select (e.g., paths with spaces).
-Our recommendation is to follow link:https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html[Java package naming conventions] or use a single unambiguous naming convention for directories and files.
+Some directory or file names may be awkward to select (e.g., paths with spaces, `$` or other special characters).
+Non-class-based test definitions should use a single unambiguous naming convention for directories and files.
+Our recommendation is to follow link:https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html[Java package naming conventions] for directories and use simple names for the test files.
 
 All of the following file paths appear to be the same test and cannot be filtered individually.
 [cols="1,1", options="header"]
--- a/platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcher.java
+++ b/platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcher.java
@@ -27,12 +27,12 @@ import java.util.Collections;
  * This class has three public APIs:
  *
  * <ul>
- * <li>Judge whether a test class might be included.
- * <li>Judge whether a test class or class+method is definitely included.
- * <li>Judge whether a test file is definitely included for resource-based tests.
+ * <li>Judge whether a test class might be matched.
+ * <li>Judge whether a test class or class+method is definitely matched.
+ * <li>Judge whether a test file is definitely matched for resource-based tests.
  * </ul>
  *
- * For example, class 'org.gradle.Test' can't be included by pattern 'org.apache.Test', so
+ * For example, class 'org.gradle.Test' can't be matched by pattern 'org.apache.Test', so
  * {@link #mayIncludeClass(String)} will return false when given 'org.gradle.Test'.
  *
  * In all cases, if the pattern starts with an uppercase letter, matching is performed on the
--- a/platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcherTest.groovy
+++ b/platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcherTest.groovy
@@ -64,7 +64,7 @@ class FileTestSelectionMatcherTest extends Specification {
         matcher.matchesFile(included)
         !matcher.matchesFile(excluded)
         where:
-        include << [ "*i*", "included", "*included*", "included*", "*included", "includ*" ]
+        include << [ "*i*", "included", "*included*", "included*", "*included", "includ*", "included.*" ]
     }
 
     def "matchesFile with simple include pattern"() {
@@ -266,8 +266,10 @@ class FileTestSelectionMatcherTest extends Specification {
         matcherWildcard.matchesFile(plainFile)
     }
 
-    def "matchesFile with default packages"() {
+    def "matchesFile within root directory"() {
         expect:
+        // This follows the behavior of ClassTestSelectionMatcher
+        //
         // This captures current behavior, but not desired behavior.
         // We cannot select files in the default package without
         // including other files because *.included requires included


big-guy:
@bot-gradle test this

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
```diff
@@ -956,22 +956,16 @@ When `./gradlew test --info` is run, the output contains lines that confirm that
 
 === Filtering
 
-Non-class-based tests can be filtered in a similar manner as class-based tests using the link:{javadocPath}/org/gradle/api/tasks/testing/TestFilter.html[`TestFilter`] API or `--tests` command line option.
+Non-class-based tests can be filtered in the same way as class-based tests using the link:{javadocPath}/org/gradle/api/tasks/testing/TestFilter.html[`TestFilter`] API or `--tests` command line option.
 
-Non-class-based test filtering is performed using regex matching against the relative path from the project directory to the test definitions file.
+Non-class-based test filtering is performed by converting the path to the test definition into a package like name by converting file path separators into package separators.
```
tresat: ❓ The path from where?  Still relative to the root of the project containing this test task?

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
```diff
@@ -956,22 +956,16 @@ When `./gradlew test --info` is run, the output contains lines that confirm that
...
+A file with the path `src/test/definitions/sub/MoreTests.xml` would be converted into a class name like `sub.MoreTests.xml` by calculating a path relative to the root of the test definitions.
```
tresat: 🤔 We shouldn't call it a "class name".  Let's say package-like name throughout.

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
```diff
@@ -956,22 +956,16 @@ When `./gradlew test --info` is run, the output contains lines that confirm that
...
 
-To apply a filter to non-class-based tests, that filter *must* include at least one `/` character to indicate that it is a file path filter rather than a class name filter.
-This forward slash can appear anywhere in the filter string, even at the start.
-If a filter does start with a `/`, it is still treated as a relative path from the project directory.
+Filtering is only supported at a file level granularity for non-class-based tests. There is no support for filtering individual tests within a test definition file.
 
-NOTE: As the same filter properties are used for both class-based and non-class-based tests, care should be taken to ensure that the filter in the proper format for the given type of test definitions and only matches the intended tests.
+If the previous example was run with the filter: `./gradlew test --info --tests "*ExampleTests.xml"` then only the tests in `ExampleTests.xml` would be executed.
```
tresat: Directory names can start with `*` = more potential for ambiguity.

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
```diff
@@ -956,22 +956,16 @@ When `./gradlew test --info` is run, the output contains lines that confirm that
...
 
-If the previous example was run with the filter: `./gradlew test --info --tests "src/test/definitions/ExampleTests.xml"` then only the tests in `ExampleTests.xml` would be executed.
-
-If it were run with the filter `/.gradlew test --info --tests ".*/sub/.*"` then only the tests in `MoreTests.xml` would be executed.
-
-Filtering is only supported at a file level granularity for non-class-based tests.
-There is no support for filtering individual tests within a test definition file.
+If it were run with the filter `./gradlew test --info --tests "\*sub.*"` then only the tests in `MoreTests.xml` would be executed.
```
tresat: ❓  What is the leading slash doing here?  Isn't it unnecessary?

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
```diff
@@ -956,22 +956,16 @@ When `./gradlew test --info` is run, the output contains lines that confirm that
...
 
-Non-class-based test filtering is performed using regex matching against the relative path from the project directory to the test definitions file.
+Non-class-based test filtering is performed by converting the path to the test definition into a package like name by converting file path separators into package separators.
+A file with the path `src/test/definitions/sub/MoreTests.xml` would be converted into a class name like `sub.MoreTests.xml` by calculating a path relative to the root of the test definitions.
 
-To apply a filter to non-class-based tests, that filter *must* include at least one `/` character to indicate that it is a file path filter rather than a class name filter.
-This forward slash can appear anywhere in the filter string, even at the start.
-If a filter does start with a `/`, it is still treated as a relative path from the project directory.
+Filtering is only supported at a file level granularity for non-class-based tests. There is no support for filtering individual tests within a test definition file.
```
tresat: 🤔 I think this should go below the examples, as it breaks the flow of the explanation here.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResourceBasedTestingCrossVersionTest.groovy
```diff
@@ -88,18 +88,15 @@ class  ResourceBasedTestingCrossVersionTest extends AbstractResourceBasedTesting
...
+        """
         writeTestDefinitions()
 
         when:
         withConnection {
             it.newTestLauncher()
                 .addProgressListener(events, OperationType.TASK, OperationType.TEST)
                 .withTestsFor { tests ->
-                    tests.forTaskPath(':test').includePattern("$DEFAULT_DEFINITIONS_LOCATION/subSomeOtherTestSpec.rbt")
+                    tests.forTaskPath(':test').includePattern("*subSomeOtherTestSpec.rbt")
```
tresat: 🤔 I don't think the leading `*` is necessary here.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/FilePathFilter.java
```diff
@@ -25,21 +25,16 @@
...
 import org.junit.platform.launcher.PostDiscoveryFilter;
 
-import java.io.File;
-import java.nio.file.Path;
-
 /**
  * A JUnit Platform {@link PostDiscoveryFilter} filter that includes or excludes
  * file or directory based tests based on their relative paths to the root directory
  * of the project that contains them.
  */
```
tresat, line 39: 💅 I don't think the javadoc is exactly accurate here anymore.  It should probably say something about "similarly to how `ClassMethodNameFilter` works".  Maybe even this type should be renamed to "ResourceFilter" or something less "exact" sounding.

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcherTest.groovy
```diff
@@ -0,0 +1,75 @@
...
+        def included = root.file("included.test").touch()
+        def excluded = root.file("excluded.test").touch()
+
+        expect:
+        def matcher = createMatcher([include], [], root)
+        matcher.matchesFile(included)
+        !matcher.matchesFile(excluded)
+        where:
+        include << [ "*i*.test", "included.test", "*included.test*", "included.test*", "*included.test", "included.*" ]
+    }
```
tresat, line 68: ❓ These tests don't demonstrate the full behavior well.  What happens with:
- multiple roots
- also containing `sub/included` and `sub/excluded`
- combinations like `sub/included/excluded` with both include and exclude filters supplied.
- Odd file names like we talked about: starting + ending with `*`, `.`, `$`, etc.
- What does just `.` match?  Just `*`?  Just `\`?

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcherTest.groovy
```diff
@@ -0,0 +1,40 @@
...
+
+class ClassTestSelectionMatcherTest extends Specification {
+    def "matchesTest with default packages"() {
+        expect:
+        // This captures current behavior, but not desired behavior.
+        // We cannot select classes in the default package without
+        // including other classes because *.SomeTest requires SomeTest
+        // to be in a package
+        def matcher = new ClassTestSelectionMatcher(["*.SomeTest*"], [], [])
+        !matcher.matchesTest("SomeTest", null) // This isn't desired
```
tresat, line 29: ❓ Does this problem occur with resource-based tests as well?  In the `FileTestSelectionMatcherTest`, what does just `included`/`excluded` match?  How about just `test`?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcher.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Collection;
+
+/**
+ * This class has one responsibility.
+ *
+ * It converts a given file path into something that looks like a class name using a given set of search roots
```
tresat, line 30: 🤔 "into something that looks like a class name" sounds underspecified, and it's easy to get paths that convert into things that look like invalid class FQNs.

In `PatternFilterable` we call them "Ant-style include and exclude patterns or specs".

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcher.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public boolean matchesFile(File file) {
+        try {
+            Path path = file.toPath().toRealPath();
+            for (Path root : roots) {
+                if (path.startsWith(root)) {
+                    String relativePath = TextUtil.normaliseFileSeparators(root.relativize(path).toString());
+                    String packagified = relativePath.replaceAll("/", ".");
+                    return classTestSelectionMatcher.matchesTest(packagified, "");
```
tresat, line 63: 🤔 We probably need more testing on the class matcher now.  It should handle (or at least not explode) for quasi-paths containing things that would be invalid the the class-based testing world.  Leading digits in segments, double periods, $, other chars not allowed at the start or end of FQN segments that are valid directory or file names.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcher.java
```diff
@@ -0,0 +1,67 @@
...
+                }
+            }
+            return false;
+        } catch (IOException e) {
+            return true;
```
tresat, line 68: ❗ Why should an IO exception indicate a match?  Doesn't this likely mean some other process has altered the file (it has to exist to get a path)?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcher.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * It converts a given file path into something that looks like a class name using a given set of search roots
+ * and then uses a regular {@link ClassTestSelectionMatcher} to match against the quasi-class name.
+ *
+ * Examples:
+ * src/test/definitions/foo.test becomes foo.test
+ * src/test/definitions/sub/foo.test becomes sub.foo.test
+ *
+ * Limitations:
+ * This means it's impossible to pick one file or the other if multiple roots have the same structure and file names.
```
tresat, line 41: 🤔 I think this limitation should be put on the public API somewhere, perhaps on the `Test` task itself.  It seems much more likely to occur with resource-based tests, especially if we're splitting into multiple test def roots to enable parallelism.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcher.java
```diff
@@ -46,331 +37,49 @@
  * otherwise, it will be used to match the fully qualified class name.
```
tresat: ❗ Needs updating, there are 3 APIs now.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -0,0 +1,298 @@
...
+ * be included by pattern 'org.apache.Test'
+ * <li>Judge whether a test method is matched exactly.
+ * </ul>
+ *
+ * In both cases, if the pattern starts with an upper-case letter, it will be used to match
+ * the simple class name;
+ * otherwise, it will be used to match the fully qualified class name.
+ */
+@NullMarked
+class ClassTestSelectionMatcher {
```
tresat, line 45: 🎉 Extracting and separating this from `TestSelectionMatcher` and making it more easily unit-testable is a good refactor.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -0,0 +1,298 @@
...
+
+    private static List<ClassTestPattern> prepareClassBasedPatternList(Collection<String> includedTests) {
+        return preparePatternList(includedTests, ClassTestPattern::new);
+    }
+
+    private static <T> List<T> preparePatternList(Collection<String> includedTests, Function<String, T> patternCreator) {
+        return includedTests.stream()
+            .map(patternCreator)
+            .collect(Collectors.toList());
+    }
```
tresat, line 60: 💅 This could be collapsed to a single private method now.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -0,0 +1,298 @@
...
+ * This class has two responsibilities:
+ *
+ * <ul>
+ * <li>Judge whether a test class might be included. For example, class 'org.gradle.Test' can't
+ * be included by pattern 'org.apache.Test'
+ * <li>Judge whether a test method is matched exactly.
+ * </ul>
+ *
+ * In both cases, if the pattern starts with an upper-case letter, it will be used to match
+ * the simple class name;
```
tresat, line 41: ❓ How does this behavior interact with file-based test filtering?  Seems like another potential source of confusion, as some file systems ignore case entirely.

In fact, just the distinction of a quasi-path string with 2+ segments with the same string using different casing seems like it would present issues.  This seems like another impedance mismatch between class names and paths that we (somehow) haven't yet discussed.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
```diff
@@ -400,7 +353,7 @@ class NonClassBasedTestingIntegrationTest extends AbstractNonClassBasedTestingIn
                         testDefinitionDirs.from("$DEFAULT_DEFINITIONS_LOCATION")
 
                         filter {
-                            excludeTestsMatching "$filterPattern"
+                            excludeTestsMatching "*AdditionalDir.*"
```
tresat, line 356: 💭 I've said it before, but I'll say it one more time here: I find this an extremely unintuitive syntax.  It's not a regex.  It's not an Ant-style pattern, it's not how most things typically specify paths or parts of paths...what is it saying?

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
```diff
@@ -766,40 +763,6 @@ public void executeTests() {
         }
     }
 
-    private void verifyAppropriateFilterConfiguration() {
```
tresat, line 769: 💭 Here's where we'd probably add the incubating warning when filtering resource-based tests.  After reading through this full PR, I think we absolutely need that warning if we make this change.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestDefinitionProcessor.java
```diff
@@ -213,20 +221,17 @@ private LauncherDiscoveryRequest createLauncherDiscoveryRequest() {
...
-                    delegatingFilter.addDelegate(FileSource.class, fileFilter);
-                    delegatingFilter.addDelegate(DirectorySource.class, fileFilter);
-                }
+                ClassMethodNameFilter classFilter = new ClassMethodNameFilter(matcher);
+                delegatingFilter.addDelegate(ClassSource.class, classFilter);
+                delegatingFilter.addDelegate(MethodSource.class, classFilter);
+
+                FilePathFilter fileFilter = new FilePathFilter(matcher);
+                delegatingFilter.addDelegate(FileSource.class, fileFilter);
+                delegatingFilter.addDelegate(DirectorySource.class, fileFilter);
```
tresat: 🤔 At this point, we should know which type of tests we have from the selectors we've gathered.  There's only a need to create the ClassMethodName filter if there are class-based tests, and vice-versa.  We can guard each of these with a `if` and avoid a little unnecessary work.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java
```diff
@@ -148,16 +148,16 @@ private static boolean isEngineNode(TestIdentifier testIdentifier) {
     private final TestResultProcessor resultProcessor;
     private final Clock clock;
     private final IdGenerator<?> idGenerator;
-    private final File baseDefinitionsDir;
+    private final Collection<Path> roots;
```
tresat: 💅 Roots isn't very descriptive.  `testDefinitionRoots`?

tresat (changes requested):
I still don't think this is the right approach.  But if we're going to do it, I think we need much more test coverage to demonstrate the behavior we're permitting, I've left a bunch of comments with suggestions.

Other things we should understand:
- How do file names with "$" play - does the conversion to quasi-class regexes still work?  How about ^ and other meaningful in a regex but not allowed in a FQN and allowed in a path characters?  These should be tested.

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
lkasso: Can we say something like: 

```suggestion
Since Gradle’s filtering logic is designed around class and package structures, non-class-based tests are mapped to a "pseudo-package" name. 
The system converts file path separators into dots (`.`), relative to the root of your test definitions.
```

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
lkasso:
```suggestion
[cols="1,1", options="header"]
|===
| Original File Path | Pseudo-Package Name
| src/test/definitions/ExampleTests.xml
| ExampleTests.xml

| src/test/definitions/sub/MoreTests.xml
| sub.MoreTests.xml
|===
```

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
lkasso: Or we can turn it into a note:

```suggestion
Note: Filtering for these tests is currently supported only at the file level. Unlike standard JUnit tests, you cannot filter for individual test cases contained within a specific definition file.
```

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
lkasso:
````suggestion
To run only the tests located in ExampleTests.xml:

```bash
./gradlew test --info --tests "*ExampleTests.xml"
```
````

File: platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc (hunk 4 above)
Comment (lkasso) [inline platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc]:
````suggestion
To run all tests within the sub directory (mapped as a sub-package):

```bash
./gradlew test --info --tests "*sub.*"
```
````

File: platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcherTest.groovy (hunk 9 above)
Comment (big-guy) [reply platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcherTest.groovy:29]: Yes, a similar problem exists with resource-based tests too

File: platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java (hunk 17 above)
Comment (big-guy) [reply platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java:41]: Yeah, there's something odd about this.

For classes, it means that `--tests "Included"` matches `Included` (in the default package) and `sub.Included` because they both have the simple name of "Included".

The similar thing here with resources would be:
```
    def "matchesFile with simple include pattern"() {
        def root = temp.createDir("root")
        def included = root.file("Included").touch()
        def otherIncluded = root.file("other.Included").touch()
        def subIncluded = root.file("sub/Included").touch()

        expect:
        def matcher = createMatcher(["Included"], [], root)
        matcher.matchesFile(included)
        matcher.matchesFile(subIncluded)
        matcher.matchesFile(otherIncluded)
    }
```

The one that looks wrong is `other.Included`, but it makes sense why there's confusion.

File: platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestDefinitionProcessor.java (hunk 20 above)
Comment (big-guy) [reply platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestDefinitionProcessor.java]: Ah, that's a good idea.

File: platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy (hunk 18 above)
Comment (big-guy) [reply platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy:356]: The third option: this is the pattern matching we use for selecting tests that matches packages, class names and method with fuzziness.

File: platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java (hunk 19 above)
Comment (big-guy) [reply platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java:769]: done

File: platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcherTest.groovy (hunk 8 above)
Comment (big-guy) [reply platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcherTest.groovy:68]: I think I've covered the interesting parts that were missing.

File: platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcher.java (hunk 12 above)
Comment (big-guy) [reply platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcher.java:68]: My thinking was that 1) this is an error path that seems unlikely and 2) when it happens, catching this when the test executes (and hopefully fails) is better than failing during discovery or silently ignoring it. 

I could see an argument that we should instead fail during discovery. In scenarios where the test should be included, we'll silently mask a temporary failure here. 

I think to trigger this, we'd need something to fail at the filesystem level or a perfectly timed removal of this file between test discovery and test execution. This seems really similar to a file being deleted *just* before test execution after we've already decided to include it here.

File: platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcher.java (hunk 11 above)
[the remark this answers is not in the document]
Comment (big-guy) [reply platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcher.java:63]: This isn't correct since passing null triggers different selection logic in ClassTestSelectionMatcher

**Comment by big-guy** [comment]
@bot-gradle test this

**Comment by big-guy** [comment]
@bot-gradle test this

File: platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc [hunk 22]
```diff
@@ -956,22 +956,55 @@ When `./gradlew test --info` is run, the output contains lines that confirm that
...
 
-NOTE: As the same filter properties are used for both class-based and non-class-based tests, care should be taken to ensure that the filter in the proper format for the given type of test definitions and only matches the intended tests.
+[cols="1,1", options="header"]
+|===
+| File path | Fully qualified name
+| src/test/definitions/ExampleTests.xml
+| ExampleTests
+
+| src/test/definitions/sub/MoreTests.xml
+| sub.MoreTests
```
tresat, line 973: ❓ Why are these dropping the file extension?  Shouldn't different types of extensions be filterable?  Seems likely that test engines (like Cucumber) might establish a convention of files with the same name but different extensions:
- feature.yaml <- BDD stuff
- feature.js <- verification methods called by the BDD stuff

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
```diff
@@ -956,22 +956,55 @@ When `./gradlew test --info` is run, the output contains lines that confirm that
...
+|===
```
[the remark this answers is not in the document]
tresat: 🤔 CoPilot makes a good point here.  The examples in the table have should have shorter paths starting from the segment after `/definitions/`

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
```diff
@@ -956,22 +956,55 @@ When `./gradlew test --info` is run, the output contains lines that confirm that
...
+
+To run only the tests in `ExampleTests.xml`:
+
+    ./gradlew test --tests "ExampleTests*"
+
+To run all tests within the `sub` directory:
+
+    ./gradlew test --tests "sub.*"
```
tresat, line 982: 🤔 If I have a file named `/something/something/sub.xml`, I'd run it with `sub*` and `sub.*` would fail, which is somewhat unexpected - should we note this here?

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
```diff
@@ -956,22 +956,55 @@ When `./gradlew test --info` is run, the output contains lines that confirm that
...
 
-If the previous example was run with the filter: `./gradlew test --info --tests "src/test/definitions/ExampleTests.xml"` then only the tests in `ExampleTests.xml` would be executed.
+==== Ambiguity with test selection
 
-If it were run with the filter `/.gradlew test --info --tests ".*/sub/.*"` then only the tests in `MoreTests.xml` would be executed.
+The goal of following the same selection rules as class-based tests is to make it easier for users to run all tests within a logical grouping without knowing if they are class-based or non-class-based.
+
+The conversion from file path to fully qualified name will be ambiguous for some directory structures.
+Directory names that contain dots or file names that contain multiple dots can make it impossible to select an individual test file.
```
tresat, line 989: 🤔 I think other special chars like `$` will also confound the conversion to class-based filter patterns.  Should this be noted explicitly in a separate bullet, or as part of this one?

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
```diff
@@ -956,22 +956,55 @@ When `./gradlew test --info` is run, the output contains lines that confirm that
...
+Some directory or file names may be awkward to select (e.g., paths with spaces).
+Our recommendation is to follow link:https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html[Java package naming conventions] or use a single unambiguous naming convention for directories and files.
```
tresat: 🤔 If we're being pedantic, the package naming convention **can't** be used for most files, if those file names have an extension and use a `.`.

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
tresat: This comment still applies.

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
tresat: This comment still applies.  "converted into the package like name `sub.MoreTests.xml`"

Ugh, the problem with package-like name is it still sounds like this is the file `xml` in the dir `sub/MoreTests/`.

Maybe instead of saying "class-like" or "package-like" we should just say something like "the filter specification" everywhere?

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestDefinitionProcessor.java
```diff
@@ -213,17 +222,16 @@ private LauncherDiscoveryRequest createLauncherDiscoveryRequest() {
...
-                    delegatingFilter.addDelegate(ClassSource.class, classFilter);
-                    delegatingFilter.addDelegate(MethodSource.class, classFilter);
-                }
-                if (matcher.hasPathBasedFilters()) {
-                    FilePathFilter fileFilter = new FilePathFilter(matcher, spec.getBaseDefinitionsDir());
+                ClassMethodNameFilter classFilter = new ClassMethodNameFilter(matcher);
+                delegatingFilter.addDelegate(ClassSource.class, classFilter);
+                delegatingFilter.addDelegate(MethodSource.class, classFilter);
+
+                if (hasDirectorySelectors()) {
```
tresat, line 233: We always add the Class-based Filters?  Why?  We can have a test with no classes and detect that in a similar way?

platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/TestFilter.java
```diff
@@ -21,7 +21,16 @@
...
 /**
- * Allows filtering tests for execution. Some examples:
+ * Allows filtering tests for execution.
+ *
+ * Class-based and non-class-based tests are filtered using the same selection logic.
+ *
+ * Classes can be selected by their simple, fully qualified names or partial matches on class name or package.
+ * Methods can be selected by their full names, containing class or partial matches.
+ *
+ * Non-class-based tests are converted to fully qualified names based on their relative path.
```
tresat:
```suggestion
 * Non-class-based tests are converted to package-like names using the relative path from their test definition directory.
```

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/AbstractTestFilteringIntegrationTest.groovy
```diff
@@ -94,6 +94,8 @@ abstract class AbstractTestFilteringIntegrationTest extends AbstractTestingMulti
         succeedsWithTestTaskArguments("test", "--tests=Ok2*")
 
         then:
+        outputDoesNotContain("Filtering non-class-based tests is an incubating feature.")
```
tresat, line 97: 🤔 Seeing it now, it does seem kind of likely that this could be misinterpreted to imply _only filtering_ is incubating, _not_ all of non-class-based testing.  Why not just print this _every time_ that `testDefinitionsDir` is non-empty?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/NonClassBasedTestingIntegrationTest.groovy
tresat: I know, I know, we've discussed this.  But it really seems to tack against "the build system should produce dependable results and users can trust the build outcome is correct (whether it’s success or failure)."  It's hard to trust these patterns if they're anything short of complete paths to files/FQNs to classes.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcher.java
tresat: Deleted or permissions changed at exactly the wrong time seems the most likely way to trigger this.  

I'd argue in favor of failing fast and visibly at this point, since we can't know how the test engine is going to behave if it expects a file to be present/absent and it isn't.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
tresat: What do you think about the larger misalignment between case being always meaningful in Java FQNs but only _maybe_ being meaningful in file system names though?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcher.java
```diff
@@ -16,361 +16,78 @@
...
- * This class has two public APIs:
+ * This class has three public APIs:
  *
  * <ul>
- * <li>Judge whether a test class might be included. For example, class 'org.gradle.Test' can't
- * be included by pattern 'org.apache.Test'
- * <li>Judge whether a test method is matched exactly.
+ * <li>Judge whether a test class might be included.
+ * <li>Judge whether a test class or class+method is definitely included.
+ * <li>Judge whether a test file is definitely included for resource-based tests.
```
tresat: 💅 This should really say "matches", not "included" as we could be matching to see if a test class/file in an _exclude_ filter.  This class' concerns are limited to just matching the pattern, so saying "include" when it's used in the larger context of filtering is slightly ambiguous.

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcherTest.groovy
```diff
@@ -0,0 +1,299 @@
...
+        and:
+        def matcherWildcard = createMatcher(["test*"], [], root)
+        matcherWildcard.matchesFile(bracketFile)
+        matcherWildcard.matchesFile(plainFile)
+    }
+
+    def "matchesFile with default packages"() {
+        expect:
+        // This captures current behavior, but not desired behavior.
+        // We cannot select files in the default package without
```
tresat, line 274: 💅 We should say "root directory" instead of "default package" - files aren't in packages.

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/FileTestSelectionMatcherTest.groovy
tresat: I think the additions are all good and very helpful.  

I still think there is missing coverage:
- I think the `..` in a filter path is also an interesting case.  Not that people _should_ do it, but it seems like something that _might_ work, since we're dealing with files, and we should document (via test) how it works or doesn't.
- What does just "." match?
- What does just ".*" match?
- What does just "*" match? (We should probably document these 3 for class-based filtering as well)
- Parens and quotes in file names?
- Case-sensitive file systems where multiple files/dirs have different names, but the same lowercased name.
- Trailing '$' and leading '^' in a file name, and their potential to mess up the generated pattern.

tresat: Also, the integration testing should probably demo at least some more situations with multiple definition dirs where there are overlapping and semi-overlapping full paths to test files across multiple definition dirs.

tresat: Something else I didn't think about during the initial filtering implementation that might be interesting: symlinks to test definition dirs - which path is actually used by the filtering logic?

eskatos (approved):
The only changes related to TIDE are in the TAPI cross-version test.
Approval from the TIDE team.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestDefinitionProcessor.java
big-guy: I suppose we could base this on the other types of selectors. e.g., if we have class selectors, we need class post-discovery filters. Otherwise, we don't

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
big-guy: We only need to worry about files that are reported as a test source, so in your example, I would assume that only one (yaml) or the other file (js) would be considered the test source by the test engine. 

Maybe a test engine could report all of the files as some kind of combined test source, but in that case, we wouldn't know how to filter it at all because it wouldn't be a FileSource/DirectorySource. If the test engine reports the directory as the test source, individual files would not be something to filter. 

Like the ambiguity between class-based and non-class-based tests in the same Test task, you could have multiple test engines and separate test definitions:
```
org.gradle.Foo - JUnit Jupiter test engine
org/gradle/Foo.bdd - BDD test engine
org/gradle/Foo.toml - Other test engine
```
And in this case, you wouldn't be able to select any one individually with `--tests`.

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
big-guy: I'm not sure if `/something/something` is the path from the project directory or within the source directory. Do you mean `$projectDir/something/something` or `$projectDir/src/test/definitions/something/something`?

In the first case, `sub*` and `sub.*` would both work.  The trailing `.*` is similar to a class selection where `Foo.*` matches all test methods within `Foo`. 
In the second case, neither would match because `sub.xml` fully qualified name is `something.something.sub`

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/AbstractTestFilteringIntegrationTest.groovy
big-guy: I initially worded it that way too, but I think it's more nuanced. I decided this particular behavior is the thing we're worried about.

- Non-class-based testing is a feature of JUnit Platform that we didn't support. Any problems with this going forward are just bugs in Gradle. 
- Gradle APIs (like testDefinitionDirs) are marked incubating.
- Test filtering is an existing API that has new runtime behavior. There's no way to mark the API as incubating, but we can mark the behavior incubating.

What swayed me the most is that this is a JUnit Platform feature that we're catching up to. It would be like marking Java 26 compilation support incubating. We need to support it.

platforms/documentation/docs/src/docs/userguide/reference/platforms/jvm/java_testing.adoc
big-guy: Yeah, I was mostly thinking about directory names. Let me try rewording

PR #36823, opened 2026-02-20
Fix IvyDynamicRevisionRemoteResolveIntegrationTest test for CC and document CC behavior

tresat:


platforms/documentation/docs/src/docs/userguide/optimizing-builds/performance.adoc
```diff
@@ -620,6 +620,15 @@ If a build file or initialization script lowers these values, Gradle queries rep
 When you don't need the absolute latest release of a dependency every time you build, consider
 removing the custom values for these settings.
 
+Note that when cached dynamic versions expire, this will invalidate Configuration Cache entries including these confgurations, as Gradle will need to re-run the configuration phase and re-resolve the dependencies.
```
lkasso, line 623:
```suggestion
Note that when cached dynamic versions expire, this will invalidate Configuration Cache entries including these confgurations.
Gradle will need to re-run the configuration phase and re-resolve the dependencies.
```

platforms/documentation/docs/src/docs/userguide/optimizing-builds/performance.adoc
```diff
@@ -620,6 +620,15 @@ If a build file or initialization script lowers these values, Gradle queries rep
...
+
+When this happens it is indicated with a message in the build output:
```
lkasso, line 625:
```suggestion
When this happens, it is indicated with a message in the build output:
```

platforms/documentation/docs/src/docs/userguide/optimizing-builds/performance.adoc
```diff
@@ -620,6 +620,16 @@ If a build file or initialization script lowers these values, Gradle queries rep
...
+Note that when cached dynamic versions expire, this will invalidate Configuration Cache entries including these confgurations.
+Gradle will need to re-run the configuration phase and re-resolve the dependencies.
+
+When this happens, it is indicated with a message in the build output:
+
+[source,text]
+----
+Calculating task graph as configuration cache cannot be reused because cached version information for org.test:projectA:1.+ has expired.
+----
+
```
mlopatkin: 🎉 Nice!

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ivy/IvyDynamicRevisionRemoteResolveIntegrationTest.groovy
```diff
@@ -549,7 +548,64 @@ task resolveStaleThenFresh {
         succeeds "resolveStaleThenFresh"
 
         and:
-        outputContains("stale:[projectA-1.2.jar],fresh:[projectA-1.3.jar]")
+        if (GradleContextualExecuter.isConfigCache()) { // CC causes even stale conf to be re-resolved, which finds the latest available version
```
mlopatkin: :x: This isn't actually a CC behavior but a side effect of variable serialization order. `fresh` gets serialized before `stale` and updates the available cached version.

If we don't want to rely on CC serialization order (which isn't really specified), we have to resolve at configuration time explicitly.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ivy/IvyDynamicRevisionRemoteResolveIntegrationTest.groovy
```diff
@@ -549,7 +548,64 @@ task resolveStaleThenFresh {
...
+        } else {
+            outputContains("stale:[projectA-1.2.jar],fresh:[projectA-1.3.jar]")
+        }
+    }
+
+    /**
+     * This test ensures that even if the conf is a part of the task fingerprint, if it isn't actually resolved, it doesn't force re-resolution
+     * Compare this with {@link #"should honour dynamic version cache expiry for subsequent resolutions"()} above.
+     */
+    def "expired dynamic version on a conf doesn't invalidate configuration cache if that cache is not resolved"() {
```
mlopatkin: 💅 
```suggestion
    def "expired dynamic version on a conf doesn't invalidate configuration cache if the conf is not resolved"() {
```

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ivy/IvyDynamicRevisionRemoteResolveIntegrationTest.groovy
```diff
@@ -552,6 +551,113 @@ task resolveStaleThenFresh {
         outputContains("stale:[projectA-1.2.jar],fresh:[projectA-1.3.jar]")
     }
 
+    def "dynamic version cache expiry will cause CC invalidation for subsequent resolutions"() {
```
mlopatkin, line 555: 🤔 This one fails with non-CC executors. 

TBH, the behavior we're seeing feels error-prone (with both CC and vintage). As a user I'd be surprised that seemingly the same dependency resolves to different versions depending on some internal timings. 

Not sure if this is something that the consistent resolution addresses (among other things).

tresat: Agreed.  Consistent resolution between `fresh` and `stale` should fix this, yes.  It's a very contrived setup for testing.

PR #36825, opened 2026-02-20
Make TransformingClassLoader thread-safe

### LINKED ISSUE #36824: TransformedClassLoader is not thread-safe
### Current Behavior

`org.gradle.internal.classloader.TransformingClassLoader` is not thread-safe, because of TOCTOU:
https://github.com/gradle/gradle/blob/6afc9edc8051aa9ae820c1f80076b46ba2808bc8/platforms/core-runtime/classloaders/src/main/java/org/gradle/internal/classloader/TransformingClassLoader.java#L79-L83

If two callsites find the package being absent, they both proceed to define, and the last one will get `IllegalArgumentException`.

### Expected Behavior

`TransformingClassLoader` should be thread-safe, because it is registered as parallel capable.

### Context (optional)

With parallel CC load, some less-used Gradle classes can be loaded concurrently.

N/A, reported internally

### Gradle version

8.14

### Build scan URL (optional)

_No response_

_No response_

mlopatkin:
Fixes #36824

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/.teamcity/subprojects.json
+++ b/.teamcity/subprojects.json
@@ -226,7 +226,7 @@
   {
     "name": "classloaders",
     "path": "platforms/core-runtime/classloaders",
-    "unitTests": false,
+    "unitTests": true,
     "functionalTests": false,
     "crossVersionTests": false
   },
--- a/platforms/core-runtime/classloaders/src/test/groovy/org/gradle/internal/classloader/TransformingClassLoaderConcurrencyTest.groovy
+++ b/platforms/core-runtime/classloaders/src/test/groovy/org/gradle/internal/classloader/TransformingClassLoaderConcurrencyTest.groovy
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.classloader
+
+import groovy.transform.CompileStatic
+import org.gradle.internal.classpath.ClassPath
+import org.gradle.internal.classpath.DefaultClassPath
+import org.gradle.test.fixtures.concurrent.ConcurrentSpec
+import spock.lang.Issue
+
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.CyclicBarrier
+import java.util.concurrent.TimeUnit
+
+class TransformingClassLoaderConcurrencyTest extends ConcurrentSpec {
+
+    @Issue("https://github.com/gradle/gradle/issues/36824")
+    def "handles concurrent class loading in same package without race condition"() {
+        given:
+        def classPath = DefaultClassPath.of(ClasspathUtil.getClasspathForClass(TransformingClassLoaderTestClasses))
+        def numClasses = TransformingClassLoaderTestClasses.NUM_CLASSES
+        def classNames = (1..numClasses).collect {
+            "${TransformingClassLoaderTestClasses.name}\$Class${String.format('%02d', it)}"
+        }
+        when:
+        async {
+            try (def classLoader = new TestTransformingClassLoader(classPath)) {
+                def loadingCompleteLatch = new CountDownLatch(numClasses)
+                def loadingBarrier = new CyclicBarrier(numClasses)
+
+                classNames.each { className ->
+                    start {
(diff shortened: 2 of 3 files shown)


mlopatkin:

platforms/core-runtime/classloaders/src/main/java/org/gradle/internal/classloader/TransformingClassLoader.java
```diff
@@ -76,14 +76,29 @@ protected Class<?> findClass(String name) throws ClassNotFoundException {
...
+                // Another thread has beaten us while trying to register this package.
+                // All do it with the same parameters so we can just keep going.
+                // Let's still check that the package is there, just in case.
+                if (getPackage(packageName) == null) {
+                    // This should never happen.
+                    throw new AssertionError("Package '" + packageName + "' cannot be defined but is not registered either in " + this, iae);
+                }
+            }
         }
         return defineClass(name, bytes, 0, bytes.length, codeSource);
```
big-guy, line 93: Is there a similar danger for defineClass too?

Either defineClass blowing up because you create the same class twice or two threads defining the same class with different instances?

big-guy: And could you try to trigger this in some kind of ConcurrentSpec test?

platforms/core-runtime/classloaders/src/main/java/org/gradle/internal/classloader/TransformingClassLoader.java
```diff
@@ -76,14 +76,29 @@ protected Class<?> findClass(String name) throws ClassNotFoundException {
         }
 
         String packageName = StringUtils.substringBeforeLast(name, ".");
-        @SuppressWarnings("deprecation")
         Package p = getPackage(packageName);
         if (p == null) {
-            definePackage(packageName, null, null, null, null, null, null, null);
+            try {
+                definePackage(packageName, null, null, null, null, null, null, null);
+            } catch (IllegalArgumentException iae) {
```
smplio, line 83: Wouldn't it be computationaly cheaper to use synchronized block here, then relying on rewinding the stack to capture stacktrace?

platforms/core-runtime/classloaders/src/main/java/org/gradle/internal/classloader/TransformingClassLoader.java
mlopatkin: > Is there a similar danger for defineClass too?

No, that one is protected by a [class-level lock in the caller](https://github.com/openjdk/jdk17u/blob/e47690176cade6a6f8cf97059cfe914f5f67cd87/src/java.base/share/classes/java/lang/ClassLoader.java#L572), even for parallel classloaders.

platforms/core-runtime/classloaders/src/main/java/org/gradle/internal/classloader/TransformingClassLoader.java
mlopatkin: > Wouldn't it be computationaly cheaper to use synchronized block here, then relying on rewinding the stack to capture stacktrace?

That's a good question, but it the lack of reports suggests that the contention here is relatively uncommon. As a data point, JDK does the same try-catch for its URLClassLoader.

platforms/core-runtime/classloaders/src/main/java/org/gradle/internal/classloader/TransformingClassLoader.java
mlopatkin: > And could you try to trigger this in some kind of ConcurrentSpec test?

I'm not a big fan of non-deterministic tests, but this one seems to be pretty reliable.

platforms/core-runtime/classloaders/src/main/java/org/gradle/internal/classloader/TransformingClassLoader.java
```diff
@@ -76,14 +76,29 @@ protected Class<?> findClass(String name) throws ClassNotFoundException {
...
+                    // This should never happen.
+                    throw new AssertionError("Package '" + packageName + "' cannot be defined but is not registered either in " + this, iae);
+                }
+            }
         }
         return defineClass(name, bytes, 0, bytes.length, codeSource);
     }
 
+    @Override
+    @SuppressWarnings("deprecation") // We still need to support Java 8 where non-deprecated version is not available.
```
Toffikk, line 97: im not sure if im understanding this comment correctly but why do we care about Java 8, if gradle is now 17+

mlopatkin: > im not sure if im understanding this comment correctly but why do we care about Java 8, if gradle is now 17+

Gradle still needs to run some of its code in Java 8+ processes, in particular, in compilation or test workers. For example, when you use toolchains to run your tests on Java 8 vm, there is a separate process spawned, but it contains some bits of Gradle runtime used to communicate test results back to the Gradle daemon.

PR #36827, opened 2026-02-21
Compile daemon-only projects to Java 17

jvandort:
This PR adds an API for certain projects to configure "required" target runtimes, where a target runtime is either "client", "daemon" or "worker". Projects containing tooling API models, or acting as an entry-point to a worker action (test workers, compiler workers) may explicitly declare "required" target runtimes, as they _know_ they must run in a certain runtime. Some projects, like those which are libraries and are consumed by other projects (":concurrent", ":time"), do not care what their target runtimes are, and thus do not need to declare any required runtimes. 

The new `:checkTargetRuntimes` task will read all required runtimes of all projects, and determine the complete set of "computed" runtimes for every project, ensuring that if one project requires a runtime, all projects it depends on will also inherit that requirement. 

We set "required" runtimes for all entry-point projects that we know must run on workers or clients. After computing the final set of runtimes, 77 projects out of 197 total are only accessible on a daemon classpath, and therefore now compile to Java 17.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/testing/distributions-integ-tests/src/integTest/groovy/org/gradle/DocsDistributionIntegrationSpec.groovy
+++ b/testing/distributions-integ-tests/src/integTest/groovy/org/gradle/DocsDistributionIntegrationSpec.groovy
@@ -34,7 +34,7 @@ class DocsDistributionIntegrationSpec extends DistributionIntegrationSpec {
 
     @Override
     int getDistributionSizeMiB() {
-        return 88
+        return 89
     }
 
     @Override


platforms/core-configuration/api-metadata/build.gradle.kts
```diff
@@ -5,6 +5,20 @@ plugins {
...
 
+gradleModule {
+    requiredRuntimes {
+        client = true
+        daemon = true
+        worker = true
+    }
+    computedRuntimes {
+        // Auto-generated by `:checkTargetRuntimes --fix`
+        client = true
```
hfhbd, line 16: What about an Enum: DaemonOnly, ClientOnly, WorkerOnly, ClientAndDaemon, All, to limit the allowed combinations? Or are all combinations really allowed?

jvandort: What would be the benefit of limiting the combinations?

Your list is missing WorkerAndDaemon, which is also a valid and used combination

hfhbd: Just to get rid of Boolean logic. I would expect 1 project should be responsible for 1 type, and exposing 3 booleans allows 9 combinations, vs 6 according to the list.

jvandort: This os the list of runtimes the classes in a project must be able to run in.

If we have a project that contains classes related to communication between the client and daemon, or worker and daemon, the project must be able to compile down to a set if requirements that satisfies both runtimes

hfhbd: BTW couldn't it implemented with a capability instead of using a test task? 🤔

jvandort: You mean an attribute?

Originally i did implement this as an attribute but we still need most of the logic here ti automatically determine which values to set for those projects which do not care about their target runtimes.

And also the error messages from attributes are much worse than what you can write in a task.

Attribute matching is still working behind the scenes verifying the underlying computed java versions

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test rfr

jvandort:
@bot-gradle test act

eskatos (approved):

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

tresat:
This PR looks huge.  Perhaps it would be best to split it into 2:
- The new tasks and machinery required to add the target runtimes (small, but important to check carefully)
- Adding the target runtimes to Gradle modules (large, but cookie-cutter)

jvandort:
I agree this PR is massive, and this is something I usually try to avoid. Though the majority of changes across the files is updating the buildscripts. The build logic changes are more minimal. This PR is already split into distinct commits that follow a logical progression.

jvandort:
@bot-gradle test rfr

platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/ConnectionScopeServices.java
```diff
@@ -136,13 +129,10 @@ ProviderConnection createProviderConnection(
 
         return new ProviderConnection(
             serviceRegistry,
-            buildLayoutFactory,
+            new BuildLayoutFactory(),
```
big-guy: Where was this coming from before?

jvandort:
@bot-gradle test act

jvandort:
@blindpirate we discovered this PR has a memory leak and the CI failures were not flakiness

jvandort:
Strangely, however, the OOMs only appeared when merging the PR and not when running a branch build

jvandort:
@bot-gradle test rfr

PR #36840, opened 2026-02-23
Fix ResolveConfigurationDependenciesBuildOperationIntegrationTest for Configuration Cache

tresat:


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ResolveConfigurationDependenciesBuildOperationIntegrationTest.groovy
```diff
@@ -502,7 +501,7 @@ class ResolveConfigurationDependenciesBuildOperationIntegrationTest extends Abst
             }
 
             task resolve(type: Copy) {
-                from configurations.runtimeClasspath
+                from configurations.runtimeClasspath.files
```
mlopatkin: :x: This doesn't feel right. Isn't the actual issue that `child` project has no repositories defined?

mlopatkin (changes requested):

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ResolveConfigurationDependenciesBuildOperationIntegrationTest.groovy
tresat: No, I think this is fine.

It is a little weird that `child` has no repos, yet has deps.  But `child` is never actually resolved, only the root project (which does have repos) is resolved, which properly gets `child` and its transitive deps using the root repos.  This is verified by the `verifyExpectedOperation` block.  If you tried to actually build or resolve `child` itself, that would fail.

The CC issue was only that the configuration couldn't be serialized.

But I'll move the repos to the settings file anyways, to make it look less weird.

mlopatkin: >The CC issue was only that the configuration couldn't be serialized.

I saw a common issue with the `compileClasspath` of [`child:compileJava`]( https://ge.gradle.org/s/rlpgllrkjedw4/tests/task/:dependency-management:configCacheIntegTest/details/org.gradle.integtests.resolve.ResolveConfigurationDependenciesBuildOperationIntegrationTest/resolved%20components%20contain%20their%20source%20repository%20name%2C%20even%20when%20taken%20from%20the%20cache?focused-exception-line=0-25&top-execution=1). Non-CC execution skips resolving it because there are no sources there. You see that "Configuration cannot be serialized" but this is a red herring, the problem is _why_ it cannot be.

I'd argue that resolving `runtimeClasspath` at configuration time is wrong, as it has `child` components that hasn't been built. It is just `Copy` being very lenient on what it has as inputs.

I'd approach this as follows:
- fix the repos for `child` (you already did)
- remove `.files`
- filter asserted build operations, something like:
```groovy
def ops = operations.all(ResolveConfigurationDependenciesBuildOperationType) {
    it.details.configurationName == "runtimeClasspath"
}
```

tresat: Done, thanks!

PR #36846, opened 2026-02-23
Fix MavenScopesAndProjectDependencySubstitutionIntegrationTest tests for Configuration Cache

tresat:


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/maven/MavenScopesAndProjectDependencySubstitutionIntegrationTest.groovy
```diff
@@ -99,13 +99,15 @@ class MavenScopesAndProjectDependencySubstitutionIntegrationTest extends Abstrac
...
-    @ToBeFixedForConfigurationCache(because = "broken file collection")
     def "when no target configuration is specified then a dependency on maven module includes the runtime dependencies of target project that is using the Java plugin"() {
         mavenRepo.module("org.test", "m1", "1.0").publish()
         mavenRepo.module("org.test", "m2", "1.0").publish()
         mavenRepo.module("org.test", "maven", "1.0")
             .dependsOn("org.test", "replaced", "1.0")
             .publish()
+        if (GradleContextualExecuter.isConfigCache()) {
+            mavenRepo.module("org.test", "ignore-me", "1.0").publish() // Will never get resolved, but needs to exist in the repo since the confs will be resolved + serialized with CC
+        }
```
mlopatkin: ❓  Do we really need to make it conditional?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/maven/MavenScopesAndProjectDependencySubstitutionIntegrationTest.groovy
```diff
@@ -195,7 +197,6 @@ class MavenScopesAndProjectDependencySubstitutionIntegrationTest extends Abstrac
         }
     }
 
-    @ToBeFixedForConfigurationCache(because = "broken file collection")
```
mlopatkin, line 198: 🤔 This annotation ensures that the test fails with CC. Why does it pass now without any changes?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/maven/MavenScopesAndProjectDependencySubstitutionIntegrationTest.groovy
tresat: We need it in the repo for CC.  For non-CC it will never get resolved, so it doesn't matter.  I was trying to make the smallest change, and not affect the non-CC tests.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/maven/MavenScopesAndProjectDependencySubstitutionIntegrationTest.groovy
tresat: I really don't know how this passed on CI.  It failed when I ran it locally, it needed another change.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/maven/MavenScopesAndProjectDependencySubstitutionIntegrationTest.groovy
mlopatkin: > We need it in the repo for CC. For non-CC it will never get resolved, so it doesn't matter. I was trying to make the smallest change, and not affect the non-CC tests.

IMO, extra conditions make it harder to navigate the test logic, so I'd avoid them. It isn't the point of this test to verify that `ignore-me` isn't resolved. A specific test that targets exactly that "NO_SOURCE" resolution skip logic should take care of this.

tresat:
> I'm pretty sure we shouldn't add these accidental conditionals when they can be avoided.

OK, that's fine by me.  I'll remove them and we'll rely on other tests to verify the NO_SOURCE resolution logic.

PR #36852, opened 2026-02-24
Make ConditionalExecutionQueue blocking work notification generic

octylFractal:
This is based on the idea behind ForkJoinPool.managedBlock, though currently without the potentially useful interface of ManagedBlocker.

Currently hooked into DefaultConditionalExecutionQueue only. Adopting this in DefaultPlanExecutor and BuildOperationExecutionContext would be the next steps to allow it to be more effective.

This PR also ensures that DefaultConditionalExecutionQueue handles the blocking notification in a way that works regardless of when items are submitted, and that there is a limit on spawned runners.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultConditionalExecutionQueue.java
+++ b/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultConditionalExecutionQueue.java
@@ -26,7 +26,6 @@ import java.util.Deque;
 import java.util.LinkedList;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -49,10 +48,11 @@ public class DefaultConditionalExecutionQueue<T> implements WorkerThreadPool, Co
     private final Deque<ConditionalExecution<T>> queue = new LinkedList<>();
     private final ReentrantLock lock = new ReentrantLock();
     private final Condition workAvailable = lock.newCondition();
-    private final AtomicInteger blockedWorkerCount = new AtomicInteger();
     private QueueState queueState = QueueState.Working;
     @GuardedBy("lock")
     private int workerCount;
+    @GuardedBy("lock")
+    private int blockedWorkerCount;
 
     public DefaultConditionalExecutionQueue(String displayName, WorkerLimits workerLimits, ExecutorFactory executorFactory, WorkerLeaseService workerLeaseService) {
         this.workerLimits = workerLimits;
@@ -88,8 +88,9 @@ public class DefaultConditionalExecutionQueue<T> implements WorkerThreadPool, Co
         workerCount++;
     }
 
+    @GuardedBy("lock")
     private int getCurrentMaxWorkerCount() {
-        int maxWorkersWithCompensation = workerLimits.getMaxWorkerCount() + blockedWorkerCount.get();
+        int maxWorkersWithCompensation = workerLimits.getMaxWorkerCount() + blockedWorkerCount;
         return Math.min(maxWorkersWithCompensation, workerLimits.getMaxUnconstrainedWorkerCount());
     }
 
@@ -97,7 +98,7 @@ public class DefaultConditionalExecutionQueue<T> implements WorkerThreadPool, Co
     public void notifyBlockingWorkStarting() {
         lock.lock();
         try {
-            blockedWorkerCount.incrementAndGet();
+            blockedWorkerCount++;
             // May not be above max workers even after adding 1 if we're at the max unconstrained worker count
             if (!queue.isEmpty() && workerCount < getCurrentMaxWorkerCount()) {
                 // Start runner to compensate immediately when we already have work to do.
@@ -110,7 +111,14 @@ public class DefaultConditionalExecutionQueue<T> implements WorkerThreadPool, Co
 
     @Override
     public void notifyBlockingWorkFinished() {
-        blockedWorkerCount.decrementAndGet();
+        lock.lock();
+        try {
+            blockedWorkerCount--;
+            // Trigger workers to see if one should exit.
+            workAvailable.signalAll();
+        } finally {
+            lock.unlock();
+        }
     }
 
     @Override


octylFractal:
@bot-gradle test PT

octylFractal:
@bot-gradle test PT

subprojects/core/src/main/java/org/gradle/internal/work/DefaultAsyncWorkTracker.java
```diff
@@ -123,23 +124,25 @@ public boolean hasUncompletedWork(BuildOperationRef operation) {
...
                     }
-                    failures.add(t);
                 }
-            }
 
-            if (failures.size() > 0) {
-                throw new DefaultMultiCauseException("There were failures while executing asynchronous work:", failures);
-            }
-        });
+                if (failures.size() > 0) {
```
reinsch82: 💅 
```suggestion
                if (!failures.isEmpty()) {
```

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultConditionalExecutionQueue.java
```diff
@@ -79,35 +82,37 @@ public void submit(ConditionalExecution<T> execution) {
...
+    public void notifyBlockingWorkStarting() {
         lock.lock();
         try {
-            // Only expand the thread pool if there is work in the queue or we know that work is about to be submitted (i.e. force == true)
-            if (force || !queue.isEmpty()) {
-                Future<?> ignored = executor.submit(new ExecutionRunner());
-                workerCount++;
+            blockedWorkerCount.incrementAndGet();
+            // May not be above max workers even after adding 1 if we're at the max unconstrained worker count
+            if (!queue.isEmpty() && workerCount < getCurrentMaxWorkerCount()) {
```
reinsch82: ❓I think the counter could run into a race condition with `notifyBlockingWorkFinished` since it is also read in `getCurrentMaxWorkerCount`
I guess we could fix this with reading it only ones.

```suggestion
            int count = blockedWorkerCount.incrementAndGet();
            // May not be above max workers even after adding 1 if we're at the max unconstrained worker count
            if (!queue.isEmpty() && workerCount < getCurrentMaxWorkerCount(count)) {
```

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultConditionalExecutionQueue.java
```diff
@@ -79,35 +82,37 @@ public void submit(ConditionalExecution<T> execution) {
...
+    @GuardedBy("lock")
+    private void startRunner() {
+        Future<?> ignored = executor.submit(new ExecutionRunner());
+        workerCount++;
     }
 
-    private int getMaxWorkerCount() {
-        return workerLimits.getMaxWorkerCount();
+    private int getCurrentMaxWorkerCount() {
+        int maxWorkersWithCompensation = workerLimits.getMaxWorkerCount() + blockedWorkerCount.get();
```
reinsch82: ❓ same as below


```suggestion
    private int getCurrentMaxWorkerCount(int counter) {
    int maxWorkersWithCompensation = workerLimits.getMaxWorkerCount() + counter;
```

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultConditionalExecutionQueue.java
```diff
@@ -68,8 +71,8 @@ public void submit(ConditionalExecution<T> execution) {
         lock.lock();
         try {
             // expand the thread pool until we hit max workers
-            if (workerCount < getMaxWorkerCount()) {
-                expand(true);
+            if (workerCount < getCurrentMaxWorkerCount()) {
```
reinsch82, line 74: ❓ same as below

```suggestion
            if (workerCount < getCurrentMaxWorkerCount(blockedWorkerCount.get())) {
```

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultConditionalExecutionQueue.java
```diff
@@ -34,7 +35,7 @@
  */
 // TODO This class, DefaultBuildOperationQueue and ExecutionPlan have many of the same
 // behavior and concerns - we should look for a way to generalize this pattern.
-public class DefaultConditionalExecutionQueue<T> implements ConditionalExecutionQueue<T> {
+public class DefaultConditionalExecutionQueue<T> implements WorkerThreadPool, ConditionalExecutionQueue<T> {
```
reinsch82, line 37: 💅 I want to point out since a tool found it: 
nothing seems to justify the `Conditional` part of the class names anymore.
There for also the comment above "... certain conditions are reached." is not relevant anymore.

Could we correct the names?

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultConditionalExecutionQueue.java
octylFractal: This is a benign race here because if the count goes down below `workerCount`, we just won't start a worker, and I think that's more correct.

But for performance reasons, it's probably best to not incur the potential extra cost of reading the count again, so I'll modify it regardless.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultConditionalExecutionQueue.java
octylFractal: This is not correct, the "conditional" part refers to the fact that a worker lease should be held. This is still true so I don't think changing it is warranted.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultConditionalExecutionQueue.java
octylFractal: This is now a plain field, since it was generally guarded by `lock` before and I realized I need to signal `workAvailable`.

subprojects/core/src/main/java/org/gradle/internal/work/DefaultAsyncWorkTracker.java
octylFractal: Done

reinsch82 (approved):

PR #36858, opened 2026-02-24
Fix MavenRemoteDependencyWithGradleMetadataResolutionIntegrationTest for Configuration Cache

tresat:


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/maven/MavenRemoteDependencyWithGradleMetadataResolutionIntegrationTest.groovy
```diff
@@ -220,10 +218,12 @@ dependencies {
     release 'test:a:1.2'
 }
 task checkDebug {
-    doLast { assert configurations.debug.files*.name == ['a-1.2-debug.jar', 'b-2.0.jar'] }
+    inputs.files(configurations.debug)
+    doLast { assert inputs.files*.name == ['a-1.2-debug.jar', 'b-2.0.jar'] }
```
ov7a, line 222: 💅 While we're here, can these assertions be moved from the build script to the test itself?

https://github.com/gradle/gradle/blob/master/contributing/Testing.md#dont-use-assertions-in-gradle-build-scripts-under-test

tresat: There are many tests like this.  I'd rather make the minimal change in each of these PRs to get them working for CC, then go back and update how they all verify the files in a single PR.

ov7a (approved):

PR #36867, opened 2026-02-25
[Investigation] Prototype performance optimization for Gradle 9.4.0-rc1

asodja:


asodja:
@bot-gradle test APT

asodja:
@bot-gradle test APT

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -1491,4 +1483,75 @@ private static PersistentMap<Class<?>, PersistentArray<ServiceProvider>> collect
             return providersByType;
         }
     }
+
+    private static class ThreadSafeArray<T> {
```
bamboo, line 1487: This is exactly how `PersistentArray` behaves up to 32 elements. Did you measure with PersistentArray before introducing ThreadSafeArray?

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
```diff
@@ -1423,23 +1417,22 @@ private boolean locateAnnotation(Class<? extends Annotation> annotation) {
      * rarely (once per lifecycle handler).
      */
     private static class ServicesSnapshot {
+        private final PersistentList<SingletonService> services;
+        private final PersistentArray<AnnotatedServiceLifecycleHandler> lifecycleHandlers;
+        private final ConcurrentMap<Class<?>, ThreadSafeArray<ServiceProvider>> providersByType;
```
bamboo, line 1422: ❌ Deviates from consistent snapshot semantics.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
asodja: I think the difference is, that this class doesn't copy array on every add (so it's not really ThreadSafeArray, I should rename that). It reuses the same array if possible, since it expects elements indexes before are never changed.

I tried with PersistentArray but I still saw allocation hotspot so I tried to optimized that part. But I dont have profile anywhere so I need to measure it again to get some numbers

bamboo: Ah, I see, it pre-allocates.

platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
asodja: Do you suggest to rename this class or do you think this generally this logic is incorrect now?

What I figured is, that this line: https://github.com/gradle/gradle/blob/a466d0db523830f50208d956aef0116bcccdf211/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java#L1480
creates a lot of allocations. Because every change to a node makes a copy of a node and additionally it makes a copy of a map. And it seems we have a lot of service types (and we also collect all super types of a service type). And we also have a lot of ServiceRegistries that do such discovery.

asodja:
We decided not to merge that

PR #36869, opened 2026-02-25
Do not use kotlin-stub

jvandort:
Dependency resolution sees this version before it sees the platform, so it tries to resolve it. It does this every time we execute the build logic and wastes about 500ms on each cc miss. We instead do not include a version at all. Unlike the stub version, dependency resolution will not attempt to resolve a non-present version.

Before:
image

After:
image

Notice all the downloads in the before trace were from fake kotlin-stub dependencies:
image

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic-commons/build-platform/build.gradle.kts
+++ b/build-logic-commons/build-platform/build.gradle.kts
@@ -15,7 +15,7 @@ dependencies {
         val distributionDependencies = versionCatalogs.named("buildLibs")
         distributionDependencies.libraryAliases.forEach { alias ->
             api(distributionDependencies.findLibrary(alias).get().map { module ->
-                if (module.version == "kotlin-stub") {
+                if (module.group == "org.jetbrains.kotlin") {
                     module.copy().apply {
                         version {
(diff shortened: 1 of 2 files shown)


build-logic-commons/build-platform/build.gradle.kts
```diff
@@ -15,7 +15,7 @@ dependencies {
         val distributionDependencies = versionCatalogs.named("buildLibs")
         distributionDependencies.libraryAliases.forEach { alias ->
             api(distributionDependencies.findLibrary(alias).get().map { module ->
-                if (module.version == "kotlin-stub") {
+                if (module.group == "org.jetbrains.kotlin") {
```
ov7a, line 18: 🤔 I wonder if there's a better way to detect these deps (e.g. verify that version is empty or prefix these deps in a special way), but I couldn't find a solution without downsides. So I guess it's ok to leave it as is.

PR #36874, opened 2026-02-25
Update test to verify using output, not in-task assertions, per testing guidelines

tresat:
Update tests in `:dependency-management` to adhere to https://bt-internal-docs.grdev.net/main-gradle-repo/contributing/Testing/?h=#dont-use-assertions-in-gradle-build-scripts-under-test.

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ArtifactDependenciesIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ArtifactDependenciesIntegrationTest.groovy
@@ -327,12 +327,13 @@ Searched in the following locations:
             task test {
                 def files = configurations.compile
                 doLast {
-                    assert files.collect { it.name } == ['middle2-1.0.jar', 'middle1-1.0.jar', 'leaf3-1.0.jar', 'leaf4-1.0.jar', 'leaf1-1.0.jar', 'leaf2-1.0.jar']
+                    println files.collect { it.name }
                 }
             }
         """
 
         succeeds("test")
+        outputContains("[middle2-1.0.jar, middle1-1.0.jar, leaf3-1.0.jar, leaf4-1.0.jar, leaf1-1.0.jar, leaf2-1.0.jar]")
     }
 
     void exposesMetaDataAboutResolvedArtifactsInAFixedOrder() {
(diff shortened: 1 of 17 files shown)


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/derived/JvmDerivedVariantsIntegrationTest.groovy
```diff
@@ -89,36 +89,35 @@ task resolve {
...
-        assert artifacts.size() == 1
+        println "artifactCount: " + artifacts.size()
         artifacts[0].with {
-            def attributes = variant.attributes
+            def attrs = variant.attributes
             def attributesAsMap = [:]
-            attributes.keySet().each {
-                attributesAsMap[it.name] = String.valueOf(attributes.getAttribute(it))
-            }
-            assert attributesAsMap.size() == 7
```
ov7a, line 99: ❌ I think this assertion should be preserved: it should be printed as verified in the test. As well as "standard" attributes

Applies to similar cases below

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/derived/MultiProjectVariantResolutionIntegrationTest.groovy
ov7a: 💅 this should be removed

ov7a: 💅 These expectations are no longer needed, right?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/gmm/GradleModuleMetadataResolveIntegrationTest.groovy
```diff
@@ -101,13 +101,14 @@ class GradleModuleMetadataResolveIntegrationTest extends AbstractIntegrationSpec
             task resolve {
                 def files = configurations.runtimeClasspath
                 doLast {
-                    assert files*.name == ["foo-1.0.jar", "${expectedFile}"]
+                    println files*.name
```
ov7a: 💅 Would be nice to prefix with something like "Resolved:"

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ArtifactDependenciesIntegrationTest.groovy
```diff
@@ -396,6 +384,23 @@ tasks.register("test", CheckArtifacts) {
...
+        outputContains("artifact1.extension=jar")
+        outputContains("artifact1.classifier=classifier")
+        outputContains("artifact2.name=lib")
+        outputContains("artifact2.type=zip")
+        outputContains("artifact2.extension=zip")
+        outputContains("artifact2.classifier=null")
+        outputContains("artifact3.name=dist")
+        outputContains("artifact3.type=zip")
+        outputContains("artifact3.extension=zip")
+        outputContains("artifact3.classifier=null")
```
ov7a, line 404: 💅 I think here the check should be a bit smarter, since `'null'` will also pass it

ov7a (changes requested):

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/derived/JvmDerivedVariantsIntegrationTest.groovy
tresat: Good catch, I didn't notice this had dropped verification of these attributes and of the count.  Restored.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/derived/MultiProjectVariantResolutionIntegrationTest.groovy
tresat: Removing this, and the task configuration blocks that set it, seemed like a larger change.  But looking at the test file again, I suppose there's no need for them to remain.  Removed.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ArtifactDependenciesIntegrationTest.groovy
tresat: I think this is a common problem, and it's unlikely that something would set the classifier to the string "null", but I'll add a little more smarts here to avoid that.

ov7a (approved):

ghale (approved):

big-guy (approved):

Issue #36548, opened 2026-01-30
DCL: Add a feature to an element (Definition) that is part of a NamedDomainObjectContainer


hfhbd:
### Expected Behavior

```kotlin
@BindsProjectFeature(FooJvmTestSuiteFeature::class)
abstract class FooJvmTestSuiteFeature : ProjectFeatureBinding {
  override fun bind(builder: ProjectFeatureBindingBuilder) {
    builder.bindProjectFeature("foo") { _: FooDefinition, _: JvmDclTestSuite ->
    }
  }
}

@BindsProjectFeature(JvmTestSuiteFeature::class)
abstract class JvmTestSuiteFeature : ProjectFeatureBinding {
  override fun bind(builder: ProjectFeatureBindingBuilder) {
    builder.bindProjectFeature("testing") { definition: DclTestingExtension, _: AdventOfCodeDefinition ->
    }
  }
}

interface DclTestingExtension : Definition<BuildModel.None> {
    @Nested
    fun getSuites(): NamedDomainObjectContainer<JvmDclTestSuite>
}

interface JvmDclTestSuite : Definition<BuildModel.None>, Named { }
```

Full implementation: https://github.com/hfhbd/adventOfCode/tree/extending-feature-element

### Current Behavior (optional)

```
e: Incorrect lifecycle state for definition 'JvmDclTestSuite_Decorated@7a25196d'. Expected it to have the context and build model attached already, got none. Check that the feature's apply action registers the build model for this definition.
```

I guess this is done due to the handling of NDOC via ObjectFactory and not using the declarative schema builder.

### Context

I want to add a feature to a JvmTestSuite (or to an upcoming JVM feature). 

Use-case: I want to generate test code for the underlying sourceSet, that can/should be either the `test` or an `integrationTest` sourceSet.

ljacomet:
This issue needs a decision from the team responsible for that area. They have been informed. Response time may vary.

h0tk3y:
Features can expose nested `Definitions`, and doing that in `NamedDomainObjectContainer` elements is also supported. The current limitation is that the feature's apply action must register the build models for those definitions.

See this example: https://github.com/gradle/gradle/blob/63ff3339843553773004654062b1ea0a44802844/platforms/core-configuration/project-features-demos/src/main/kotlin/org/gradle/api/plugins/java/plugin/JavaProjectTypePlugin.kt#L63

@hfhbd, could you please check if this works for you?

[...]

PR #36876, opened 2026-02-25
Finalize feature definitions before applying features

### LINKED ISSUE #36673: Support for accessing property values set in build definition when apply actions are executed
### Expected Behavior

Property values from build definition should be accessible in apply actions passed to `ProjectTypeBindingBuilder.bindProjectType()` (and similar methods). 

Accessing properties result in the following error `Cannot query the value of property 'extension' because it has no value available`. Seemingly, the property configuration happens after the `bind` methods. 

There's a workaround to execute the same code in an `afterEvaluate` hook. It would be better to have a dedicated hook for when the build definition is fully evaluated (ie a new `afterBuildDefinition` hook).

### Context

I tried to define a `NamedDomainObjectContainer` exposed on the build definition with type `Foo` with a `Property<String> getBar()`member. Then, I tried to use the value of the `bar` Property to define a Gradle task. I was only able to do it in an `afterEvaluate` hook.

ghale:
Changes feature application so that the definition of a feature is configured and finalized before the feature is applied.

There are several behavior changes involved in this:
- We auto-discover nested definitions and expose them as binding points.  This means that it is no longer necessary for user code to call `registerBuildModel(...)`.  Instead, the build model is discovered and registered when the definition is instantiated.  Authors can call `getBuildModel()` to retrieve the build model for a nested definition object.
- There is a new `withNestedBuildImplementationType(Class, Class)` method on `DeclaredProjectFeatureBindingBuilder`.  This allows the feature author to declare which implementation types should be used when instantiating nested build models that are discovered.
- Project feature application now occurs after the feature definition is configured.  In practice, this occurs after conversion (for DCL scripts) or after the top-level project type is configured (for non-DCL scripts).  This does mean that for non-DCL scripts, further modification of the definition is not possible elsewhere in the script (i.e. after the project type configure block).  In other words, all configuration of feature definitions must occur within the configuration of the project type.
- Project feature defaults are not processed until the settings object is fully configured.  This allows Settings plugins to add defaults without causing the list of project features to be finalized before another Settings plugin is applied and tries to register another feature declaration.  In other words, we add all of the project features and defaults first, and only then do we finalize the list of features and process/validate the defaults.
- Project feature definitions are now finalized and cannot be modified by apply actions.  Note that this means finalizing any `Property` objects and disallowing further changes to named domain object collections nested in the definition.

Fixes #36673 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/core-configuration/project-features-demos/src/integTest/groovy/org/gradle/api/plugins/demo/quality/DemoCodeQualityIntegrationTest.groovy
```diff
@@ -56,41 +56,6 @@ class DemoCodeQualityIntegrationTest extends AbstractIntegrationSpec {
         outputDoesNotContain("checkTestDemoSourceQuality")
     }
 
-    def "can apply the demoBytecodeQuality feature to Java"() {
```
ghale, line 59: This test was a duplicate of the previous test.

hfhbd:
Will the automatic build model registration also work for NDOC? https://github.com/gradle/gradle/issues/36548#issuecomment-3854377643

Edit: yes: 26fce80b332d869c7f47ed6cfa6331ebd93f9216

So this pr will also fix #36548

platforms/core-configuration/project-features-demos/src/main/kotlin/org/gradle/api/plugins/java/plugin/GroovyProjectTypePlugin.kt
```diff
@@ -74,6 +74,8 @@ class GroovyProjectTypePlugin : Plugin<Project> {
                 registerJar(model.classes.named("main"), model, services.taskRegistrar)
             }
             .withUnsafeDefinitionImplementationType(DefaultGroovyProjectType::class.java)
+            .withNestedBuildModelImplementationType(GroovyClasses::class.java, GroovyClasses.DefaultGroovyClasses::class.java)
```
hfhbd, line 74: Is this call safe/unsafe? Do I also need to call this when using a safe BuildModel without a Default class? Is there also a test?

ghale: Neither - build models themselves are neither safe or unsafe.  They follow the safety of the apply action in terms of what services are available to them, but there is nothing about the build model itself that is inherently safe or unsafe.  There might be some case in the future where we want to introduce a concept of safety, but for now, they are neither.

You do not need this call if you don't have an implementation class for the build model (i.e. if you just want Gradle to implement the class specfied as the parameter of `Definition`).  This is only when you want to expose a public build model type that other features see, but you want Gradle to actually instantiate a more specific implementation type (that only your feature sees).

There will be tests.  This PR was just put out there as a draft for internal discussion in its current state - it's still unfinished.

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/features/ProjectFeatureApplicationReflectionToConversionResultHandler.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import org.gradle.api.internal.DynamicObjectAware
+import org.gradle.features.internal.binding.ProjectFeatureSupportInternal
+import org.gradle.internal.declarativedsl.evaluator.conversion.ReflectionToConversionResultHandler
+import org.gradle.internal.declarativedsl.mappingToJvm.ReflectionToConversionResult
+
+/**
+ * A {@link ReflectionToConversionResultHandler} that walks through the graph of features and applies them.
+ */
+class ProjectFeatureApplicationReflectionToConversionResultHandler : ReflectionToConversionResultHandler {
```
h0tk3y: 💭 Just a thought that came to my mind while I was looking through this: maybe an alternative that might give a simpler implementation would be to record the applied features (and their definition instances) somewhere in the [`ProjectFeatureComponent` runtime customisation](https://github.com/gradle/gradle/blob/37b521871ef99e2e22435bbe0cff72cf92df03d2/platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/software/schemaFromProjectFeatures.kt#L276) or even in the `ProjectFeatureApplicator` and simply go over a flat list of recorded feature applications and invoke the apply actions one by one rather than inspecting the resulting object graph. I'll read on to see if this is the case, though...

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/features/ProjectFeatureApplicationReflectionToConversionResultHandler.kt
```diff
@@ -0,0 +1,36 @@
...
+
+package org.gradle.internal.declarativedsl.features
+
+import org.gradle.api.internal.DynamicObjectAware
+import org.gradle.features.internal.binding.ProjectFeatureSupportInternal
+import org.gradle.internal.declarativedsl.evaluator.conversion.ReflectionToConversionResultHandler
+import org.gradle.internal.declarativedsl.mappingToJvm.ReflectionToConversionResult
+
+/**
+ * A {@link ReflectionToConversionResultHandler} that walks through the graph of features and applies them.
```
h0tk3y: 💭 Just a thought that came to my mind while I was looking through this: maybe an alternative that might give a simpler implementation would be to collect a full list of the applied features (and their definition instances) somewhere in the [`ProjectFeatureComponent` runtime customisation](https://github.com/gradle/gradle/blob/37b521871ef99e2e22435bbe0cff72cf92df03d2/platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/software/schemaFromProjectFeatures.kt#L276) or even in the `ProjectFeatureApplicator` (by linking them to some conversion feature so they know where to pass the collected feature applications) and simply go over a flat list of recorded feature applications instead of inspecting the resulting object graph. I'll read on to see if this is the case, though...

platforms/core-configuration/project-features-demos/src/integTest/groovy/org/gradle/api/plugins/demo/quality/DemoCodeQualityIntegrationTest.groovy
h0tk3y: Not exactly: the previous test is for the feature bound to the definition, and this one is for another feature bound to the build model. Why not test both?

h0tk3y (commented):
I did the first pass, but I think I need to take a closer look at some of the details, so I'll get back to it.

platforms/core-configuration/project-features-demos/src/integTest/groovy/org/gradle/api/plugins/demo/quality/DemoCodeQualityIntegrationTest.groovy
ghale: Bah, you're right.  When I was looking at it, I just didn't notice that the difference was `demoBytecodeQuality` vs `demoSourceQuality`.  The tests looked the same to me.  I thought this might have just been a cut/paste error.

I'll put it back.

ghale: Reverted.

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/features/ProjectFeatureApplicationReflectionToConversionResultHandler.kt
ghale: Yeah, that's sort of how I handled it on my first attempt.  But it felt very contrived to do things as a side effect of conversion.

We already had the pattern for the analysis step where we pass something up out of the step that a handler could then post-process, and so that's mostly why I landed on this similar implementation.  I'm happy to reconsider, though, if you think this isn't a good approach. 

Is there something about using this pattern that you think would be problematic?  Or something about a different pattern that you feel would be better?

ghale: I've pulled this out so that `ProjectFeatureApplicator` tracks the features that are registered and then we can apply them all at once after conversion.  We do this through the `RuntimeCustomAccessor` that we register for project features. 

In other words, we don't have to walk the object graph to apply features, we just register them as we encounter them during evaluation and then once everything has been configured, then we call the feature applicator to apply them all.

Doing it through the feature applicator is best because it means we can easily capture feature references in defaults and don't have to coordinate handing those off between defaults application and script evaluation.  The feature applicator is project-scoped so it's a good place to gather the application instances.

PR #36877, opened 2026-02-25
Remove two semi-constant fields from collection properties

mlopatkin:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/AbstractCollectionProperty.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/AbstractCollectionProperty.java
@@ -83,10 +83,10 @@ public abstract class AbstractCollectionProperty<T, C extends Collection<T>> ext
     private final ValueCollector<T> valueCollector;
     private CollectionSupplier<T, C> defaultValue;
 
-    AbstractCollectionProperty(PropertyHost host, Class<? extends Collection> collectionType, Class<T> elementType) {
+    AbstractCollectionProperty(PropertyHost host, ValueCollector<T> valueCollector, Class<T> elementType) {
         super(host);
         this.elementType = elementType;
-        valueCollector = new ValidatingValueCollector<>(collectionType, elementType, ValueSanitizers.forType(elementType));
+        this.valueCollector = valueCollector;
         init();
     }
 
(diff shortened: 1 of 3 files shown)


mlopatkin:
@bot-gradle test APT

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/DefaultListProperty.java
```diff
@@ -25,14 +25,25 @@
...
+import java.util.function.Supplier;
 
 import static org.gradle.internal.Cast.uncheckedNonnullCast;
 
 public class DefaultListProperty<T> extends AbstractCollectionProperty<T, List<T>> implements ListProperty<T> {
     private static final SerializableSupplier<ImmutableCollection.Builder<Object>> FACTORY = ImmutableList::builder;
 
     public DefaultListProperty(PropertyHost host, Class<T> elementType) {
-        super(host, List.class, elementType, Cast.uncheckedNonnullCast(FACTORY));
+        super(host, List.class, elementType);
```
alllex: 💅  We are still passing the collection type to super here, even though it's available as a method. Can we make the super constructor call the new `getCollectionType()` method instead? Alternatively, we can lift the `ValueCollector` from the `AbstractCollectionProperty`, then we wouldn't need to pass the collection type inside at all.

alllex (approved):

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/AbstractCollectionProperty.java
```diff
@@ -79,18 +79,14 @@
 public abstract class AbstractCollectionProperty<T, C extends Collection<T>> extends AbstractProperty<C, CollectionSupplier<T, C>>
     implements CollectionPropertyInternal<T, C> {
 
-    private final Class<? extends Collection> collectionType;
     private final Class<T> elementType;
```
alllex, line 82: 💭  If it can make a further difference, we can get rid of this field as well, because it's always available in the `ValueCollector`.

mlopatkin: > 💭 If it can make a further difference, we can get rid of this field as well, because it's always available in the `ValueCollector`.

That's a good point, but for now it is "free" because of the alignment for the most common layout (64-bit with Compressed OOPS and Compressed Class Pointers).

Issue #35921, opened 2025-12-03
Improve detection of non-interactive build invocations


joshfriend:
### Expected Behavior

Gradle should use other environment markers to determine if a build invocation should be interactive or not. [Here are a few examples that could be used](https://docs.firebender.com/get-started/agent#interactive-commands).

### Current Behavior (optional)

Gradle's current detection of interactivity is to check if stdin is available, which is true in most cases where an agent is running builds, but most agents are not able to respond to an interactive prompt in a tooling call that they run.

### Context

Builds executed by LLM agents sometimes run into a case where there is an interactive prompt presented and then the agent waits forever for the command to finish while the command waits forever for the agent to respond to the prompt. I find that this happens with the develocity plugin asking users to accept the TOS when publishing build scans to scans.gradle.com.

eskatos:
This feature request is in the backlog of the relevant team and is prioritized by them.
***
The current detection might not be enough. Thanks for the report.

eskatos:
If you are interested in contributing to Gradle, this issue is actionable and ready for contribution but might be challenging for first-time contributors.

See [CONTRIBUTING.md](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md) for more information.

eskatos:
After some thought, we prefer following the path of tools like `apt-get`, `brew` to use the presence of a non empty conventional `NONINTERACTIVE` environment variable and also add a first class `--non-interactive` cli option.

See 
* https://github.com/gradle/gradle/pull/36880

eskatos:
The `NONINTERACTIVE` env var support was not included in the PR that closed this issue.
See this issue about it:
* https://github.com/gradle/gradle/issues/38155

Issue #35855, opened 2025-11-28
Add console output mode that hides successful tasks and focuses on failures and task output


vlsi:
### Expected Behavior

Tell us how it should work.

I’d like a console output mode (for example `--console=summary`, exact name up to you) that:

* **Does not print a line per successful task** (e.g. `> Task :foo:bar UP-TO-DATE`, `NO-SOURCE`, `SKIPPED`, `FROM-CACHE`) when the task completes without errors and does not emit any output.
* **Still prints all information for failing tasks**:

  * The task header line (`> Task :foo:test FAILED`)
  * The failure details and stacktrace, as in the current console modes.
* **Still prints any explicit output emitted by a task**, regardless of whether it eventually succeeds or fails (e.g. `logger.lifecycle`, `println`, test output, codegen messages, etc).
* **Provides a brief summary at the end of the build**, similar to the current “X actionable tasks: Y executed, Z up-to-date, …”, so I can still see how much work was done.

In other words, for this console mode:

* Successful, “silent” tasks are not listed individually.
* Tasks that fail, or tasks that produce log output, remain visible.
* This should be **opt-in** and should not change the default console behavior. On the other hand, I would probably prefer making the new behavior a default one, so Gradle skips silent successful tasks by default.

### Current Behavior (optional)

In the current console modes, Gradle prints a lifecycle line for each task, for example:

```text
> Task :build-logic-commons:gradle-plugin:pluginDescriptors UP-TO-DATE
> Task :build-logic-commons:gradle-plugin:processResources UP-TO-DATE
> Task :build-logic-commons:gradle-plugin:classes UP-TO-DATE
> Task :build-logic:basics:checkKotlinGradlePluginConfigurationErrors SKIPPED
> Task :build-logic:build-parameters:jar UP-TO-DATE
...
```

For large multi-project builds, this can easily be hundreds of lines where the vast majority of tasks are `UP-TO-DATE`, `NO-SOURCE`, or `SKIPPED`.
Even when nothing fails, the console is filled with “noise” about successful tasks.

Flags like `--quiet` and `--console=plain` help somewhat, but:

* `--quiet` hides *all* lifecycle progress, including helpful information about which task is currently running.
* `--console=plain` removes the rich UI, but it still prints one line per task outcome.

There is currently no console option that keeps failures and task output visible, but automatically hides “silent” successful tasks.

### Context

**Use case: big multi-project builds.**
We have large builds where:

[...]

PR #36880, opened 2026-02-25
Introduce `--non-interactive` command line option

eskatos:
Introduce `--non-interactive` command line option that instructs the build to make no prompting on the console even if an interactive console is available.

This is useful for running Gradle in automated environments such as CI pipelines, scripts, and AI agents where no user input is available.

Init uses the defaults for all options when `--non-interactive` is used.
Build scans plugin skips publishing when ToS have not been accepted.

* Fixes https://github.com/gradle/gradle/issues/35921

First two commits are preparation commits.
Better reviewed commit by commit.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
+++ b/platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
@@ -98,8 +98,6 @@ public abstract class InitBuild extends DefaultTask {
     @Internal
     private ProjectLayoutSetupRegistry projectLayoutRegistry;
 
-    private final boolean isNonInteractive = getProject().getGradle().getStartParameter().isNonInteractive();
-
     /**
      * Should default values automatically be accepted for options that are not configured explicitly?
      * <p>
@@ -429,7 +427,7 @@ public abstract class InitBuild extends DefaultTask {
     }
 
     private UserInputHandler getEffectiveInputHandler() {
-        if (isNonInteractive || getUseDefaults().get()) {
+        if (getUseDefaults().get()) {
             return new NonInteractiveUserInputHandler();
         }
 


eskatos:
@bot-gradle test RFN

eskatos:
@bot-gradle test this

eskatos:
@bot-gradle test this

eskatos:
@bot-gradle test this

eskatos:
@bot-gradle test this

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
```diff
@@ -427,7 +430,8 @@ private static String getRelativePath(File baseDir, File targetFile) {
     }
 
     private UserInputHandler getEffectiveInputHandler() {
-        if (getUseDefaults().get()) {
+        if (isNonInteractive || getUseDefaults().get()) {
+            System.out.println("NON INTERACTIVE INIT");
```
jbartok: Is this intentional or a debug leftover? I guess intentional, just want to confirm

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
```diff
@@ -98,6 +99,8 @@ public abstract class InitBuild extends DefaultTask {
     @Internal
     private ProjectLayoutSetupRegistry projectLayoutRegistry;
 
+    private final Boolean isNonInteractive = getServices().get(StartParameterInternal.class).isNonInteractive();
```
jbartok: Primitive `boolean` would do. Also, maybe just `@Inject` start parameters service?

platforms/core-runtime/start-parameter/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
```diff
@@ -1003,4 +1004,30 @@ public void applyTo(boolean value, StartParameterInternal settings, @Nullable Or
...
+    public static class NonInteractiveBuildOption extends EnabledOnlyBooleanBuildOption<StartParameterInternal> {
+        private static final String ENV_VAR_NAME = "NONINTERACTIVE";
+
+        public NonInteractiveBuildOption() {
+            super(null, CommandLineOptionConfiguration.create("non-interactive", "Do not do interactive prompting."));
+        }
+
+        @Override
+        protected OptionCategory getCategory() {
+            return OptionCategory.LOGGING;
```
jbartok: Doesn't feel like the right category. Maybe `EXECUTION`, or `OTHER`?

platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/action/BuildActionSerializer.java
```diff
@@ -174,6 +174,7 @@ public void write(Encoder encoder, StartParameterInternal startParameter) throws
             valueSerializer.write(encoder, startParameter.getParallelToolingModelBuilding());
             encoder.writeNullableString(startParameter.getDevelocityUrl());
             encoder.writeNullableString(startParameter.getDevelocityPluginVersion());
+            encoder.writeBoolean(startParameter.isNonInteractive());
```
jbartok, line 177: Won't this cause issues with mismatched daemon and TAPI client versions? Doesn't it need some kind of adaptation or something?

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
alllex: +1, let's prefer declarative inject instead of a runtime look-up

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
alllex: If it's not a debug leftover, I am curious about the user value of this

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/initialization/NonInteractiveActivation.groovy
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.integtests.fixtures.initialization
+
+import org.gradle.integtests.fixtures.executer.GradleExecuter
+
+enum NonInteractiveActivation {
+    ENVIRONMENT(["NONINTERACTIVE": "1"]),
```
alllex: ❓ Do we have a test that shows the edge-case of the env var being set to an empty value?

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/command_line_interface.adoc
```diff
@@ -823,6 +823,19 @@ A <<build_environment.adoc#sec:gradle_configuration_properties,Gradle property>>
 `-Dorg.gradle.java.home`::
 A <<build_environment.adoc#sec:gradle_configuration_properties,Gradle property>> that sets the JDK home dir.
 
+[[sec:non_interactive]]
```
alllex, line 826: ❔ Most boolean Gradle CLI options come in two variants for enablement and disablement. Have you given thought of whether it would make sense here as well?

platforms/core-runtime/native/src/main/java/org/gradle/internal/nativeintegration/console/TestOverrideConsoleDetector.java
```diff
@@ -39,10 +39,10 @@ public ConsoleMetaData getConsole() {
     }
 
     @Override
-    public boolean isConsoleInput() {
-        if (Boolean.getBoolean(TestOverrideConsoleDetector.INTERACTIVE_TOGGLE)) {
+    public boolean isInteractiveConsole() {
+        if (Boolean.getBoolean(TestOverrideConsoleDetector.INTERACTIVE_CONSOLE_TOGGLE)) {
```
alllex, line 43: 💅 Since we are touching this code, can we use `InternalOptions` instead of system properties to read the, well, internal options?

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
alllex: 🤔  Maybe we should get rid of `getServices()` on the AbstractTask. It's "internal" API, but it does not matter much, when every person writing a task gets it in the auto-complete.

platforms/core-runtime/daemon-services/src/main/java/org/gradle/api/internal/tasks/userinput/DefaultUserInputHandler.java
```diff
@@ -60,6 +60,7 @@ public boolean interrupted() {
     private class InteractiveUserQuestions implements UserInteraction {
         private boolean hasPrompted;
 
+        @Nullable
```
tresat, line 63: 🤔 This annotation seems odd.

Why are even calling a method to ask a Yes/No question if we are in non-interactive mode?  Shouldn't we keep this method as non-nullable and only call it if we are interactive?

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -76,6 +76,13 @@ Gradle provides a [set of features and abstractions](userguide/java_testing.html
 ### CLI, logging, and problem reporting
 Gradle provides an intuitive [command-line interface](userguide/command_line_interface.html), detailed [logs](userguide/logging.html), and a structured [problems report](userguide/reporting_problems.html#sec:generated_html_report) that helps developers quickly identify and resolve build issues.
 
+#### Non-interactive mode
+
+Gradle now supports a `--non-interactive` command-line option and a `NONINTERACTIVE` environment variable to disable all interactive console prompting.
+This is useful for running Gradle in automated environments such as CI pipelines, scripts, and AI agents where no user input is available.
```
tresat: ❓ Why is using an environment variable to enable this mode also necessary when there is a CLI flag to enable this mode?

The environment variable seems like a potential footgun - if someone sets this on a machine then another dev on that machine might be confused why they suddenly stop seeing any prompts.

It would be good to print something like "Non-Interactive mode set via environment" at build start up if we add this, to make it clear to build runners (humans and agents) why prompts aren't used.

platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/BuildInitInteractiveIntegrationTest.groovy
```diff
@@ -474,4 +475,62 @@ class BuildInitInteractiveIntegrationTest extends AbstractInteractiveInitIntegra
...
+        )
+
+        then:
+        ScriptDslFixture.of(BuildInitDsl.KOTLIN, targetDir, null).assertGradleFilesGenerated()
+
+        where:
+        activation << NonInteractiveActivation.values()
+    }
+
+    def "use defaults when run from an interactive session with --non-interactive and not enough options"() {
```
tresat, line 488: 🤔 The name of this test (and the next) "an interactive session with --non-interactive" is confusing.  I would call that a non-interactive session.

Also, this test uses `#activation`, so it sometimes _is_ an interactive session.

I think these 3 tests need better names, but I don't have a suggestion in mind.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/initialization/NonInteractiveActivation.groovy
tresat: Also, we should clearly show the behavior when setting this var to `false`.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/AbstractGradleExecuter.java
```diff
@@ -199,7 +199,7 @@ public static void doNotPropagateSystemProperty(String name) {
 
     private String profiler = System.getProperty(PROFILE_SYSPROP, "");
 
-    protected boolean interactive;
+    protected boolean interactiveConsole;
```
tresat: 🤔 This change renames `interactive` to `interactiveConsole` throughout the code.

I think a better name for these might be `interactiveSession`.

To me, `interactiveConsole` suggests "You have run Gradle from a headed console, a tool where it is possible for the user to interact with Gradle".  In contract `interactiveSession` suggests "You have run Gradle with the instruction it shouldn't try to interact with the user."  This is a subtle difference, but perhaps worth refactoring the name.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/AbstractGradleExecuter.java
```diff
@@ -440,7 +440,7 @@ public GradleExecuter copyTo(GradleExecuter executer) {
 
         executer
             .withProfiler(profiler)
-            .withForceInteractive(interactive);
+            .withForceInteractiveConsole(interactiveConsole);
```
tresat: 🤔 Similarly, this method on the executor could perhaps be named `withInteractiveSession`.

The word "force" here suggests to me there is some sort of conflict we're overriding - we're not, we're just setting an ordinary flag to control behavior.

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
tresat: I think there is a lot of value in confirming that we're running non-interactive, in case someone forgot about setting the environment var (see my other comment).  I think this is not a good message though:
- It will probably be obvious from the behavior of Gradle that we're not running non-interactively, when Gradle fails to ask interactive questions
- It doesn't say _why_ we're running non-interactively.  Was the env var set?  Or was the flag provided?  That's the important thing to know if you as a user want to change this behavior and make the next Gradle invocation actually be interactive.

platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/action/BuildActionSerializer.java
eskatos: Nope, this is the client/daemon protocol. There is no cross-version handling to take care of.

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
eskatos: Constructor injection would both leak internals and be a potential breaking change. Protected getter injection would leak internals to the task API.

I guess we have to pick a tradeoff here.

platforms/documentation/docs/src/docs/release/notes.md
eskatos: This is trying to follow a convention among command line tools. Many tools (`apt-get`, `brew` etc...) use the presence of a non-empty `NONINTERACTIVE` environment variable as a trigger. But after some more research I see that some other tools actually evaluates the env var value and consider `false` as disabling the non-interactivity, especially in the JS ecosystem ... but for them `yes` == true and `no` == true ... Looks like there are no good conventions.

The more I contemplate this, the more I think we should just add the CLI option, like a lot of tools out there do.

platforms/core-runtime/daemon-services/src/main/java/org/gradle/api/internal/tasks/userinput/DefaultUserInputHandler.java
eskatos: This is unrelated to the changes in this PR. A null return is not about the current session being interactive nor about the new cli option. This method may return null if the prompt is interrupted. I added this annotation because this is the current state of this API.

platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/BuildInitInteractiveIntegrationTest.groovy
eskatos: The "interactive session" part is about the test running with an interactive console, with support for interactivity. When you read all the tests in this class, it is pretty clear.

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
alllex: I doubt the InitBuild task was designed to be extended, and nobody should extend it anyway. So, I wouldn't be too concerned with internal leakage. The protected injection should not be a breaking change, since any child class should also be abstract.

The reason we need to know about non-interactive in this context at all is to print a message. Do we really need to print that message? (I guess we need to have a clear answer in https://github.com/gradle/gradle/pull/36880#discussion_r2964397217)

If we want to indeed print the message, we also need to answer the question of whether user-defined tasks can benefit from having the same knowledge. If the answer is yes, and InitBuild is just one of those tasks, they what we actually need is a public API so that the 3rd-party tasks can also react to whether or not it is running in non-interactive mode.

If we don't want to make this info part of the public API, then we are hitting an old problem that we are still trying to implement the init command as if it's anyone's task. That's why there are questions about internal API leakage in the first place. I personally don't think it's the right design, but it's the one with which we have to live for now.

Either way, if we don't allow 3rd-party tasks reading non-interactive status, then we should treat it as a signal that we should either lock down the ability to extend InitBuild task and say we can use whatever internals we want, or seriously consider whether it has to be re-implemented as an internal thing and not a task.

---

If we are concerned that there are people extending this task, whom we might break, then I'm okay with the `getServices()` trade off. But it's a smell

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/AbstractGradleExecuter.java
eskatos: In that case, `AbstractGradleExecuter.interactiveConsole` actually means "run the test telling Gradle it is run from a headed console". This is used by tests that want to assert on the behaviour of Gradle run from a headed console. I think the name is good.

platforms/documentation/docs/src/docs/release/notes.md
alllex: Supporting the `NONINTERACTIVE` variable would have probably helped with a common use case of running things on CI and never wanting "interactive" there. That is, dev-ops could set this variable for all CI jobs and happily forget about it. In that case, they'd probably want the safest path among possible conventions and it's okay to expected to them do `NONINTERACTIVE=true` and nothing else.

If we don't got with `NONINTERACTIVE`, maybe we should think about the happy path of how dev-ops should configure CI to get that effect? Are there conventions followed by most CI vendors?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/AbstractGradleExecuter.java
eskatos: We're setting an internal property that tells Gradle to discard the detection of a headed console and forcibly consider that it was run with a headed console, for testing purpose.

platforms/core-runtime/native/src/main/java/org/gradle/internal/nativeintegration/console/TestOverrideConsoleDetector.java
eskatos: Tried this. It requires threading the `InternalOptions` down into `:native` and `NativeServices`. This would require too much changes to be considered for these internal **test-only** properties, I gave up.

platforms/documentation/docs/src/docs/release/notes.md
eskatos: I removed the support for the `NONINTERACTIVE` environment variable from the changes.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/AbstractGradleExecuter.java
tresat: I understand, my point is that we're not forcing/changing _the type of console_ being used.  We're changing the behavior of this Gradle invocation, the "session" to act in a certain way.  

Since the type of console doesn't change, it seemed less appropriate to me to have the name suggest we are forcing Gradle to use a certain type of console, rather than telling Gradle to behave in a certain way.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/initialization/NonInteractiveActivation.groovy
eskatos: With the env var support removed, I resolve this conversation as outdated.

platforms/core-runtime/daemon-services/src/main/java/org/gradle/api/internal/tasks/userinput/DefaultUserInputHandler.java
tresat: A quick javadoc explaining this could be helpful.

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
eskatos: This was a debug leftover. I removed it for now.

With the env var support removed and the need for an explicit `--non-interactive, I don't think we need some extra output to denote an invocation with `--non-interactive`. It is better to have consistent output with no surprise.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/command_line_interface.adoc
eskatos: Conventionally the option is named `--non-interactive` and has no positive counterpart. Using our convention for positive/negative options it would have to be named `--no-interactive` or `--no-interactivity` or some other name.

More importantly, forcing `--interactive` on an invocation with no headed console would produce a non useful behaviour when the build is asking for input. For that reason I think we should not have a positive counterpart.

platforms/core-runtime/start-parameter/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
eskatos: Agreed. I did put it in `LOGGING` alongside `-console` and `--console-unicode`. We don't have any options in `OTHER`. `EXECUTION` doesn't feel right. Given the ongoing discussions about `--console` conflating too many aspects and that it should be split (https://github.com/gradle/gradle/issues/35855#issuecomment-3777507976), I think adding a new `CONSOLE` section would make sense.

eskatos: Moved to a new `CONSOLE` category alongside the other related options.

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
eskatos: Bah, found an obvious simpler way to do this without service lookup: start parameter is accessible without any injection nor lookup. Resolving.

eskatos: FTR, this isn't about printing the message but about wiring the right input handler. Without interactivity there is no user input.

platforms/core-runtime/daemon-services/src/main/java/org/gradle/api/internal/tasks/userinput/DefaultUserInputHandler.java
eskatos: Done

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/executer/AbstractGradleExecuter.java
eskatos: I see now, thanks for clarifying. I guess it makes sense. Changes done.

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/LoggingConfigurationBuildOptions.java
```diff
@@ -46,9 +47,10 @@ public class LoggingConfigurationBuildOptions extends BuildOptionSet<LoggingConf
     private final List<? extends BuildOption<? super LoggingConfiguration>> options = Arrays.asList(
         new LogLevelOption(),
         new StacktraceOption(),
+        new WarningsOption(),
         new ConsoleOption(),
         new ConsoleUnicodeOption(),
-        new WarningsOption()
+        new NonInteractiveBuildOption()
```
alllex: 💅  Other things are called just `Option` here, while the new one is the `BuildOption`. Would it be better to keep uniformity or do you think `BuildOption` is more explicit?

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
alllex: >FTR, this isn't about printing the message but about wiring the right input handler. Without interactivity there is no user input.

I missed that part 👍 

---

However, now we are making the info on non-interactivity part of the public API via non-internal `StartParameter`, while it wasn't such in the initial version of the PR.

Is there any Plugin Author or Build Engineer facing docs that we want to add? Or is it a sneaky public API that we don't really want anybody to use programmatically?

platforms/core-runtime/native/src/main/java/org/gradle/internal/nativeintegration/console/TestOverrideConsoleDetector.java
alllex: 😔

platforms/software/build-init/src/main/java/org/gradle/buildinit/tasks/InitBuild.java
eskatos: I made it the same as `--console`. It has Javadoc and is discoverable that way. Currently there is no public API to gather user input from the daemon though, `UserInputHandler` is internal API. I would keep it as it is for now.

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/LoggingConfigurationBuildOptions.java
eskatos: +1 for uniformity. I fixed that.

PR #33971, opened 2025-06-23
TAPI coverage: add test coverage for Gradle versions that don't support the JVM used to run the TAPI client

cobexer:

PR #36884, opened 2026-02-26
Improve Cross Version Test Setup to Prepare for Java Version Differences in Test Worker JVM and Target Gradle Distribution JVM

cobexer:
- **crossVersionTest: only create crossVersionTests and related configuration if the tests are enabled for a project**
- **crossVersionTest: add a separate source set (targeting Java 8) for classes that are sent across the TAPI to different JVM versions so our Test infrastructure is not forced to Java 8 for the next few years**
- **crossVersionTest: compile classes we send across the TAPI boundary with java**
- **crossVersionTest: move Java classes that cross the TAPI to the crossVersionTestFixtures source set**
- **tooling-api: consistently name test classes to end in `Spec` to improve consistency**
- **Let ToolingApiConnector extends GradleConnector for better Kotlin interop**
- **smoke-test: GradleBuildDocumentationConfigurationCacheSmokeTest: `:docs` does not have cross version tests**

Pre-requisite of:
* https://github.com/gradle/gradle/pull/33971

A few stats for the difference before/after applying the changes in this PR:
https://ge.gradle.org/c/sicxe7tlvrwrg/imx2w7qzlrdbc/custom-values (https://github.com/gradle/gradle/commit/6b114ba6e078b63aa03b53f82266b0164cf189a2 vs https://github.com/gradle/gradle/commit/e96851101385c62bb2e8c5ac33027d968b005cfe) 

| Diff | What |
| --- | --- |
| -737 | Configurations (if I read that right that is -30k dependency edges) |
| -500MiB | G1 Old Gen (in that one build invocation) |
| -13k | Tasks |
| -0.8MiB | Cache entry size |
| +931 | Transforms |

### DIFF
--- a/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildDocumentationConfigurationCacheSmokeTest.groovy
+++ b/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildDocumentationConfigurationCacheSmokeTest.groovy
@@ -90,6 +90,7 @@ class GradleBuildDocumentationConfigurationCacheSmokeTest extends AbstractGradle
         result.task("':docs:docsTest'").outcome == TaskOutcome.SUCCESS
     }
 
+    @Ignore("With https://github.com/gradle/gradle/pull/36884 :docs:embeddedCrossVersionTest does not exist any more")
     def "can resolve classpath for :docs:embeddedCrossVersionTest with configuration cache enabled"() {
         given:
         def tasks = [":docs:embeddedCrossVersionTest", "--dry-run"]


build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/shared-configuration.kt
```diff
@@ -232,10 +232,7 @@ fun DistributionTest.setSystemPropertiesOfTestJVM(defaultVersions: String) {
 
 
 internal
-fun Project.configureIde(testType: TestType) {
-    val prefix = testType.prefix
-    val sourceSet = the<SourceSetContainer>().getByName("${prefix}Test")
-
+fun Project.configureIde(sourceSet: SourceSet) {
```
cobexer, line 235: 💅 Could use a better name

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
```diff
@@ -28,27 +32,54 @@ plugins {
     id("gradlebuild.jvm-compile")
 }
 
-val sourceSet = sourceSets.create("${TestType.CROSSVERSION.prefix}Test")
-jvmCompile {
-    addCompilationFrom(sourceSet)
+project.layout.settingsDirectory.file(".teamcity/subprojects.json").asFile.apply {
```
cobexer: 💅 Alternatively we may want to only apply this plugin to projects that have cross version tests instead of selectively making it not have any effect (that would be my preference)

Also, it may be the time to ask if we still need `crossVersionTests` in that JSON file if we do the above.

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
```diff
@@ -28,27 +32,54 @@ plugins {
...
+    if (exists()) {
+        val json = GsonBuilder().create().fromJson(readText(StandardCharsets.UTF_8), JsonArray::class.java)
+        val subprojectInfo = json.firstOrNull { jsonElement ->
+            jsonElement.asJsonObject.get("name").asString.equals(project.name)
+        }
+
+        if (subprojectInfo?.asJsonObject?.get("crossVersionTests")?.asBoolean == true) {
+            val fixturesSourceSet = sourceSets.create("${TestType.CROSSVERSION.prefix}TestFixtures")
```
cobexer: 💅 `...Fixtures` may be an overloaded name here, better name suggestions welcome

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
```diff
@@ -28,27 +32,54 @@ plugins {
...
+                // the production code for the in-development Gradle version
+                addCompilationFrom(fixturesSourceSet) {
+                    targetJvmVersion = 8
+                }
+            }
+
+            project.dependencies {
+                project.configurations[testSourceSet.implementationConfigurationName](fixturesSourceSet.output)
+                project.configurations[fixturesSourceSet.compileOnlyConfigurationName](project(":tooling-api", "shadedRuntimeElements"))
+                project.configurations[fixturesSourceSet.compileOnlyConfigurationName](libs.slf4jApi + ":2.0.17") // TODO: use version from version catalog
```
cobexer: 💅 Is there a way to pick this up automatically from `:tooling-api`? It ends up in the published POM after all: https://repo.gradle.org/ui/repos/tree/PomView/libs-releases/org/gradle/gradle-tooling-api/9.3.1/gradle-tooling-api-9.3.1.pom
```xml
  <dependencies>
    <dependency>
      <groupId>org.slf4j</groupId>
      <artifactId>slf4j-api</artifactId>
      <version>2.0.17</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
```

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiClassLoaderProvider.groovy
```diff
@@ -40,9 +40,13 @@ class ToolingApiClassLoaderProvider {
     private static final GradleDistribution CURRENT_GRADLE = new UnderDevelopmentGradleDistribution(IntegrationTestBuildContext.INSTANCE)
 
     static ClassLoader getToolingApiClassLoader(ToolingApiDistribution toolingApi, Class<?> target) {
-        def testClassPath = [ToolingApiSpecification, target]
+        def testClassPath = [ToolingApiSpecification, target] //FIXME: this is where we forget our additional source set / class directories
```
cobexer: 💅 We should eventually understand why the implementation does this and figure out a better way, let's not block merging this PR on that though.

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiClassLoaderProvider.groovy
```diff
@@ -40,9 +40,13 @@ class ToolingApiClassLoaderProvider {
...
             .collect { ClasspathUtil.getClasspathForClass(it) }
 
+        // TODO: Why do we break classloading in the first place?
```
cobexer: 💅 Should disappear together with the FIXME above

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
```diff
@@ -28,27 +32,54 @@ plugins {
     id("gradlebuild.jvm-compile")
 }
 
-val sourceSet = sourceSets.create("${TestType.CROSSVERSION.prefix}Test")
-jvmCompile {
-    addCompilationFrom(sourceSet)
+project.layout.settingsDirectory.file(".teamcity/subprojects.json").asFile.apply {
+    if (exists()) {
```
blindpirate: Why we need to check the existence of this file? Shouldn't it always exist?

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
```diff
@@ -28,27 +32,54 @@ plugins {
...
+        val json = GsonBuilder().create().fromJson(readText(StandardCharsets.UTF_8), JsonArray::class.java)
+        val subprojectInfo = json.firstOrNull { jsonElement ->
+            jsonElement.asJsonObject.get("name").asString.equals(project.name)
+        }
+
+        if (subprojectInfo?.asJsonObject?.get("crossVersionTests")?.asBoolean == true) {
```
blindpirate: I don't think we apply `cross-version-tests.gradle.kts` for all projects - why do we need to check global `subproject.json` to determine if a project includes cross version tests?

The ideal way is probably conditionally apply this plugin (only on the subproject with crossVersionTests=true) from a higher level.

platforms/ide/tooling-api/src/crossVersionTestModels/java/org/gradle/integtests/tooling/fixture/SideModel.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright 2024 the original author or authors.
```
blindpirate, line 2: Nit: probably update the template to 2026.

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
cobexer: IIRC it's the same reason we have https://github.com/gradle/gradle/blob/60410e28ea7d89c39fc9a030745a51ae6c233354/build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts#L75

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiConnector.groovy
```diff
@@ -34,29 +34,49 @@ class ToolingApiConnector {
...
+        connector.disconnect()
+    }
+
     ToolingApiConnector searchUpwards(boolean searchUpwards) {
         connector.searchUpwards(searchUpwards)
         this
     }
 
-    def methodMissing(String name, args) {
-        connector."$name"(*args)
```
blindpirate, line 43: This was hacky...

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
cobexer: We apply it in https://github.com/gradle/gradle/blob/04a560a948ca7612bd3334b548d288ed0388aa10/build-logic/uber-plugins/src/main/kotlin/gradlebuild.jvm-library.gradle.kts#L35

aka to 221 projects https://ge.gradle.org/s/fnjn2pxrk5d5k/plugins?toggled=WyJncmFkbGVidWlsZC5jcm9zcy12ZXJzaW9uLXRlc3RzLTkxIl0#gradlebuild.cross-version-tests-91

> The ideal way is probably conditionally apply this plugin ...

Exactly (see my comment above)

platforms/ide/tooling-api/src/crossVersionTestModels/java/org/gradle/integtests/tooling/fixture/SideModel.java
cobexer: The code moved from the `BaseModel.groovy` file to separate Java files, thus IMO the year should stay the same https://github.com/gradle/gradle/blob/2f38c9ecb489f43fd0ec3e738fce42f0930658e5/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/fixture/BaseModel.groovy#L2

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiConnector.groovy
cobexer: Yes! The new implementation has other tradeoff though: not overriding a new method from the base class may do something that has no effect... but there is a good chance a new method would come with new tests that would fail if missed here

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
blindpirate: It's a bad name, because they're not fixtures. How about simply `crossVersionTestJava8`?

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiClassLoaderProvider.groovy
```diff
@@ -40,9 +40,13 @@ class ToolingApiClassLoaderProvider {
...
+        File testClassClassFolder = ClasspathUtil.getClasspathForClass(target)
+        testClassPath.add(testClassClassFolder.parentFile.parentFile.toPath().resolve("java/" + testClassClassFolder.name + "Fixtures").toFile())
```
blindpirate: This is a bit hacky and hard to understand. Can we do this:

1. In the `crossVersionTestFixtures` directory, create a marker class `ToolingApiJava8AdditionalClasspath` (similar to `ToolingApiAdditionalClasspath`).
2. We get the directory by `ClasspathUtil.getClasspathForClass(ToolingApiJava8AdditionalClasspath.class)`.

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
blindpirate: I was considering add some extension property, similar to https://github.com/gradle/gradle/blob/81436cfaba36bb171778627e471032c4b679461a/build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/extension/IntegrationTestExtension.kt#L22

So that we explicitly configure it in each subproject's build script:

```
crossVersionTests.enabled = true
```

In this way we also avoid the ugly `if (project.name != "gradle-kotlin-dsl-accessors")` check. There are only 19 subprojects with cross version tests, so it might be an option.

cobexer: I updated the implementation to only apply the `gradlebuild-cross-version-tests` plugin to the 19 projects that have any and thus need it, I could thus remove most of the ugly conditionals / the reading of the `.teamcity/subprojects.json`

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
cobexer: `crossVersionTestJava8` may be an option, but this may become Java 11 for Gradle 10 and will change again every so often thus a name without a java version is better. Any ideas?

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
cobexer: Done

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiClassLoaderProvider.groovy
cobexer: Marker classes are a hack as well, but also don't work here: we are talking about classes from 19 different projects - always relative to the `target` class' location

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r72/JavaVersionCrossVersionSpec.groovy
```diff
@@ -30,7 +30,7 @@ import spock.util.Exceptions
 
 @Issue('https://github.com/gradle/gradle/issues/9339')
 @TargetGradleVersion(">=7.2")
-class JavaVersionCrossVersionTest extends ToolingApiSpecification {
+class JavaVersionCrossVersionSpec extends ToolingApiSpecification {
```
jvandort, line 33: How come we renamed all these tests? Are these renames required for this change, or did you just feel like this was a good time to improve consistency?

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiClassLoaderProvider.groovy
jvandort: I agree this seems really hacky. I don't think we should be making any assumptions about the structure of the build directory here. This is the sort of logic that belongs in build logic.

What exactly are we trying to do here? It seems like we are trying to put the entire set of test classes for the current project and the entire set of cross version test fixtures for the current project and put them on the tooling api classpath. At the very least, I'd rather we compute these paths in the build logic and pass them into the test process as system properties.

But in general, I don't think this pattern is really that sustainable. What happens if my cross version test fixtures have dependencies? This logic doesn't pick that up. What happens if my cross version tests use the cross version test fixtures from another project? This logic doesn't pick that up. 

This also seems to go against the pattern we have right below, in `createTestClassLoader`. 

In practice, the test classes and the cross version test fixtures should already be on the current classpath. `createTestClassLoader` is responsible for defining which classes on the current classpath we want to let through to the tooling api execution. I don't think we should be 'side-loading' these classes which are already on the test classpath the way we are doing here. 

Could you explain the motivation behind these lines here? I am not sure I really understand the reasoning behind these changes.

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
```diff
@@ -28,27 +29,38 @@ plugins {
...
-        }
+createQuickFeedbackTasks(testSourceSet, fixturesSourceSet, releasedVersions)
+createAggregateTasks(testSourceSet, fixturesSourceSet, releasedVersions)
+configureIde(fixturesSourceSet)
+configureIde(testSourceSet)
+configureDependenciesForCrossVersionTests()
+
+fun configureDependenciesForCrossVersionTests() {
+    dependencies {
+        project.configurations[fixturesSourceSet.compileOnlyConfigurationName](project(":tooling-api", "shadedRuntimeElements"))
```
jvandort: We should be declaring the proper attributes here to resolve the shaded runtime elements instead of selecting the variant by name

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
```diff
@@ -28,27 +29,38 @@ plugins {
...
+        project.configurations[fixturesSourceSet.compileOnlyConfigurationName](libs.slf4jApi + ":2.0.17") // TODO: use version from version catalog
+
+        project.configurations[testSourceSet.implementationConfigurationName](fixturesSourceSet.output)
```
jvandort: Its suspicious that we're using a file collection dependency here instead of declaring a dependency on a variant. The way this is wired, no dependencies of test fixtures will make it onto the test classpath. This also means I can't depend on another project's cross version test fixtures. 

We should instead make sure we have proper variants for the test fixtures outputs so they can be included in a dependency graph along with their dependencies, instead of through a file collection dependency

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
```diff
@@ -62,15 +74,38 @@ fun createQuickFeedbackTasks() {
             // We should always be using JUnitPlatform at this point, so don't call useJUnitPlatform(), else this will
             // discard existing options configuration and add a deprecation warning.  Just set the existing options.
             (this.testFramework.options as JUnitPlatformOptions).includeEngines("cross-version-test-engine")
+            this.testClassesDirs += fixturesSourceSet.output.classesDirs
```
jvandort: This does not feel right. The testClassesDirs are the collection of directories that contain tests to execute. We should not be declaring tests in the test fixtures source set.

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
```diff
@@ -62,15 +74,38 @@ fun createQuickFeedbackTasks() {
...
+            this.classpath += fixturesSourceSet.runtimeClasspath
```
jvandort: If we properly declared a dependency on the test fixtures instead of what we're doing above, would not need this. This should be wired automatically.

The above line in question:
`project.configurations[testSourceSet.implementationConfigurationName](fixturesSourceSet.output)`

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
```diff
@@ -91,6 +125,8 @@ fun createAggregateTasks(sourceSet: SourceSet) {
             this.useJUnitPlatform {
                 includeEngines("cross-version-test-engine")
             }
+            this.testClassesDirs += fixturesSourceSet.output.classesDirs
+            this.classpath += fixturesSourceSet.runtimeClasspath
```
jvandort: Same as above, these declarations do not seem right to me.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r72/JavaVersionCrossVersionSpec.groovy
cobexer: Since I already went over all the classes there I took it as an opportunity to consistently name the tests with `...Spec`. I (now) separated that change out into a distinct commit so it's easier to review.

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiClassLoaderProvider.groovy
cobexer: > In practice, the test classes and the cross version test fixtures should already be on the current classpath.

Yes, but: for cross version tests with the tooling API we run two executions: 
* TAPI current -> old Gradle version: this one is easy and works without additional weirdness as the classpath is correct from the get go
* old TAPI -> Gradle current: this is where complexity comes in as Java resolves additional types from the same classloader or upwards - but our TAPI tests have a custom classloader at the very bottom with the old TAPI and that forces us to load the test class, TAPI and models in the same bottom classloader so all test dependencies are from the same classloader with the old TAPI implementation

I'll do some digging to figure out where the separate test task went that existed here: https://github.com/gradle/gradle/commit/6baa62d74ba7d057f8aa25c5f40f14000e59aa79#diff-e1378cecf4e9f6b95b3ca04b2665b269b4dbedb3f5ac96cfc644e6f72631c96eR62

cobexer: > At the very least, I'd rather we compute these paths in the build logic and pass them into the test process as system properties.

I'll push changes with that approach.

> What happens if my cross version test fixtures have dependencies?

That should work with the updated implementation. 

> In practice, the test classes and the cross version test fixtures should already be on the current classpath.

They are! and for `TAPI current -> any Gradle version` we use the setup as defined by the build. But as documented for example in `5.4.3.1. Class and Interface Resolution` https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-5.html the defining loader is used to resolve dependencies and that forces us to add any custom TAPI models, the test classes, fixtures, ... to the new thread context classloader so the correct versions of the TAPI classes are used when the TAPI we desire to use is an older version that has to shadow the current one that is also already on the classpath. That is why we must side load classes here.

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
cobexer: Very true, fixed

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
cobexer: Using the `:distributions-dependencies` platform resolves this

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiClassLoaderProvider.groovy
cobexer: Not needed any longer

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiClassLoaderProvider.groovy
cobexer: This should be fixed fully

platforms/ide/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiClassLoaderProvider.groovy
cobexer: Obsolete

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
cobexer: Done

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
cobexer: Done

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
cobexer: Done

cobexer:
@bot-gradle test RfR

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildDocumentationConfigurationCacheSmokeTest.groovy
```diff
@@ -90,6 +90,7 @@ class GradleBuildDocumentationConfigurationCacheSmokeTest extends AbstractGradle
         result.task("':docs:docsTest'").outcome == TaskOutcome.SUCCESS
     }
 
+    @Ignore("With https://github.com/gradle/gradle/pull/36884 :docs:embeddedCrossVersionTest does not exist any more")
     def "can resolve classpath for :docs:embeddedCrossVersionTest with configuration cache enabled"() {
```
cobexer, line 94: ❓ @ljacomet @jvandort @bamboo What could we do as an alternative for this test? 

It was added in https://github.com/gradle/gradle/pull/29993

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
```diff
@@ -62,15 +113,48 @@ fun createQuickFeedbackTasks() {
...
+        dependsOn("forkingCrossVersionTest")
+    }
+
+    tasks.named("quickFeedbackCrossVersionTest") {
+        dependsOn("quickFeedbackCrossVersionTests")
+    }
+}
+
+fun IntegrationTest.addClasspathSystemPropertyArgumentProvider() {
+    val integTest = project.sourceSets.named("integTest").map { it.runtimeClasspath }
```
cobexer, line 146: Note to self: this line needs to be explained

cobexer: Getting this to be perfect needs to be done separately to avoid spiralling this PR out of control, see https://github.com/gradle/gradle/issues/37033

build-logic/integration-testing/src/main/kotlin/gradlebuild.cross-version-tests.gradle.kts
cobexer: I and @donat came up with `crossVersionTestModels`, I'll rename stuff accordingly before merging

cobexer:
@bot-gradle test RfR

PR #36887, opened 2026-02-26
Internal doc on adding a TAPI operation parameter

eskatos:
Capturing the result of a chat in the internal Javadoc for adding a TAPI operations parameter and the backwards compability aspects.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/connection/ProviderOperationParameters.java
+++ b/platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/connection/ProviderOperationParameters.java
@@ -39,7 +39,7 @@ import java.util.concurrent.TimeUnit;
  *
  * For backwards compatibility, the provider need to provide a default value.
  * Let the getter take a parameter of the same type as its return type.
- * When the provider queries the parameter, it passes the default value that will be used if none available.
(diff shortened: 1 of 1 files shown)


platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/connection/ProviderOperationParameters.java
```diff
@@ -32,6 +32,14 @@
...
  *
  * This is used as an adapter over the {@link org.gradle.tooling.internal.consumer.parameters.ConsumerOperationParameters} instance provided by the consumer.
+ *
+ * Adding a new parameter and exposing it to TAPI clients requires declaring a getter here,
+ * adding it to {@code ConsumerOperationParameters},
+ * and updating the public {@link org.gradle.tooling.LongRunningOperation}.
+ *
+ * For backwards compatibility, the provider need to provide a default value.
+ * Let the getter take a parameter of the same type as its return type.
+ * When the provider queries the parameter, it passes the default value that will be used if none available.
```
donat:
```suggestion
 * When the getter is called and {@code ConsumerOperationParameters} does not have a corresponding getter (without a parameter), the default value from the parameter is returned immediately.
```

PR #36890, opened 2026-02-26
Use consistent wording in help output

donat:
The updated output is below. The diff is available [here](https://github.com/gradle/gradle/pull/36890/changes#diff-e9ebc85b79b4b9daec03cbf86f4990daa9d5d817359c33998525ffdb871b8505).

```
-?, -h, --help                     Shows this help message.
-a, --no-rebuild                   Does not rebuild project dependencies.
--build-cache                      Enables the Gradle build cache. Gradle will try to reuse outputs from previous builds.
--configuration-cache              Enables the configuration cache. Gradle will try to reuse the build configuration from previous builds.
--configuration-cache-problems     Configures how the configuration cache handles problems (fail or warn). Supported values are 'warn', or 'fail' (default).
--configure-on-demand              Configures necessary projects only. Gradle will attempt to reduce configuration time for large multi-project builds. [incubating]
--no-configure-on-demand           Disables the use of configuration on demand. [incubating]
--console                          Specifies which type of console output to generate. Supported values are 'plain', 'colored', 'auto' (default), 'rich', or 'verbose'.
--console-unicode                  Specifies which character types are allowed in the console output. Supported values are 'auto' (default), 'disable', or 'enable'.
--continue                         Continues task execution after a task failure.
--no-continue                      Stops task execution after a task failure.
-D, --system-prop                  Sets a JVM system property (fo example, -Dmyprop=myvalue).
-d, --debug                        Sets log level to debug. Includes the normal stacktrace.
--daemon                           Uses the Gradle daemon to run the build. Starts the daemon if it is not running.
--no-daemon                        Does not use the Gradle daemon to run the build. Useful occasionally if you have configured Gradle to always run with the daemon by default.
-F, --dependency-verification      Configures the dependency verification mode. Supported values are 'strict', 'lenient', or 'off'.
-g, --gradle-user-home             Specifies the Gradle user home directory. Default is ~/.gradle.
-I, --init-script                  Specifies an initialization script.
-i, --info                         Sets the log level to info.
--include-build                    Includes the specified build in the composite.
-M, --write-verification-metadata  Generates checksums for dependencies used in the project. Accepts a comma-separated list.
-m, --dry-run                      Runs the build with all task actions disabled.
--max-workers                      Configures the maximum number of concurrent workers Gradle is allowed to use.
--offline                          Invokes the build without accessing network resources.
-P, --project-prop                 Sets a project property for the build script (for example, -Pmyprop=myvalue).
-p, --project-dir                  Specifies the start directory for Gradle. Default is the current directory.
--parallel                         Builds projects in parallel. Gradle will attempt to determine the optimal number of executor threads to use.
--no-parallel                      Disables parallel project execution.
--priority                         Specifies the scheduling priority for the Gradle daemon and all processes launched by it. Supported values are 'normal' (default) or 'low'.
--problems-report                  Enables the HTML problems report. [incubating]
--no-problems-report               Disables the HTML problems report. [incubating]
--profile                          Profiles build execution time. Generates a report in the <build_dir>/reports/profile directory.
--project-cache-dir                Specifies the project-specific cache directory. Default is .gradle in the root project directory.
--property-upgrade-report          Runs the build with the experimental property upgrade report. [incubating]
-q, --quiet                        Logs errors only.
--refresh-keys                     Refreshes the public keys used for dependency verification.
--rerun-tasks                      Ignores previously cached task results.
-S, --full-stacktrace              Prints the full (very verbose) stacktrace for all exceptions.
-s, --stacktrace                   Prints the stacktrace for all exceptions.
--scan                             Generates a Build Scan (powered by Develocity).
                                   Build Scan and Develocity are registered trademarks of Gradle, Inc.
                                   For more information, please visit https://gradle.com/develocity/product/build-scan/.
--status                           Shows the status of running and recently stopped Gradle daemons.
-t, --continuous                   Enables continuous build. Gradle does not exit and will re-execute tasks when task file inputs change.
--task-graph                       Prints the task graph instead of executing tasks.
-U, --refresh-dependencies         Refreshes the state of dependencies.
--update-locks                     Performs a partial update of the dependency lock. Allows passed-in module notations to change version. [incubating]
-V, --show-version                 Prints version information and continues.
-v, --version                      Prints version information and exits.
-w, --warn                         Sets the log level to warn.
--warning-mode                     Specifies which mode of warnings to generate. Supported values are 'all', 'fail', 'summary' (default), or 'none'.
--watch-fs                         Enables file system watching. Reuses file system data for subsequent builds.
--no-watch-fs                      Disables file system watching.
--write-locks                      Persists dependency resolution for locked configurations. Ignores existing locking information if it exists.
-x, --exclude-task                 Specifies a task to exclude from execution.
--                                 Signals the end of built-in options. Parses subsequent parameters as tasks or task options only.
```

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/cli/src/main/java/org/gradle/cli/SystemPropertiesCommandLineConverter.java
+++ b/platforms/core-runtime/cli/src/main/java/org/gradle/cli/SystemPropertiesCommandLineConverter.java
@@ -29,6 +29,6 @@ public class SystemPropertiesCommandLineConverter extends AbstractPropertiesComm
 
     @Override
     protected String getPropertyOptionDescription() {
-        return "Set system property of the JVM (e.g. -Dmyprop=myvalue).";
+        return "Sets a JVM system property (for example, -Dmyprop=myvalue).";
     }
 }
--- a/platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
+++ b/platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
@@ -28,67 +28,67 @@ To see a list of available tasks, run gradle tasks
 USAGE: gradle [option...] [task...]
 
 -?, -h, --help                     Shows this help message.
--a, --no-rebuild                   Do not rebuild project dependencies.
+-a, --no-rebuild                   Disables rebuilding of project dependencies.
 --build-cache                      Enables the Gradle build cache. Gradle will try to reuse outputs from previous builds.
 --no-build-cache                   Disables the Gradle build cache.
 --configuration-cache              Enables the configuration cache. Gradle will try to reuse the build configuration from previous builds.
 --no-configuration-cache           Disables the configuration cache.
---configuration-cache-problems     Configures how the configuration cache handles problems (fail or warn). Defaults to fail.
---configure-on-demand              Configure necessary projects only. Gradle will attempt to reduce configuration time for large multi-project builds. [incubating]
+--configuration-cache-problems     Configures how the configuration cache handles problems (fail or warn). Supported values are 'warn', or 'fail' (default).
+--configure-on-demand              Configures necessary projects only. Gradle will attempt to reduce configuration time for large multi-project builds. [incubating]
 --no-configure-on-demand           Disables the use of configuration on demand. [incubating]
---console                          Specifies which type of console output to generate. Values are 'plain', 'colored', 'auto' (default), 'rich' or 'verbose'.
---console-unicode                  Specifies which character types are allowed in console output to generate. Values are 'auto' (default), 'disable' or 'enable'.
---continue                         Continue task execution after a task failure.
---no-continue                      Stop task execution after a task failure.
--D, --system-prop                  Set system property of the JVM (e.g. -Dmyprop=myvalue).
--d, --debug                        Log in debug mode (includes normal stacktrace).
---daemon                           Uses the Gradle daemon to run the build. Starts the daemon if not running.
---no-daemon                        Do not use the Gradle daemon to run the build. Useful occasionally if you have configured Gradle to always run with the daemon by default.
+--console                          Specifies which type of console output to generate. Supported values are 'plain', 'colored', 'auto' (default), 'rich', or 'verbose'.
+--console-unicode                  Specifies which character types are allowed in the console output. Supported values are 'auto' (default), 'disable', or 'enable'.
+--continue                         Continues task execution after a task failure.
+--no-continue                      Stops task execution after a task failure.
+-D, --system-prop                  Sets a JVM system property (for example, -Dmyprop=myvalue).
+-d, --debug                        Sets log level to debug. Includes the normal stacktrace.
+--daemon                           Uses the Gradle daemon to run the build. Starts the daemon if it is not running.
+--no-daemon                        Runs the build without the Gradle daemon. Useful occasionally if you have configured Gradle to always run with the daemon by default.
 --export-keys                      Exports the public keys used for dependency verification.
--F, --dependency-verification      Configures the dependency verification mode. Values are 'strict', 'lenient' or 'off'.
+-F, --dependency-verification      Configures the dependency verification mode. Supported values are 'strict', 'lenient', or 'off'.
 --foreground                       Starts the Gradle daemon in the foreground.
--g, --gradle-user-home             Specifies the Gradle user home directory. Defaults to ~/.gradle
--I, --init-script                  Specify an initialization script.
--i, --info                         Set log level to info.
---include-build                    Include the specified build in the composite.
--M, --write-verification-metadata  Generates checksums for dependencies used in the project (comma-separated list)
--m, --dry-run                      Run the builds with all task actions disabled.
---max-workers                      Configure the number of concurrent workers Gradle is allowed to use.
---offline                          Execute the build without accessing network resources.
--P, --project-prop                 Set project property for the build script (e.g. -Pmyprop=myvalue).
--p, --project-dir                  Specifies the start directory for Gradle. Defaults to current directory.
---parallel                         Build projects in parallel. Gradle will attempt to determine the optimal number of executor threads to use.
---no-parallel                      Disables parallel execution to build projects.
---priority                         Specifies the scheduling priority for the Gradle daemon and all processes launched by it. Values are 'normal' (default) or 'low'
---problems-report                  (Experimental) enables HTML problems report
---no-problems-report               (Experimental) disables HTML problems report
---profile                          Profile build execution time and generates a report in the <build_dir>/reports/profile directory.
---project-cache-dir                Specify the project-specific cache directory. Defaults to .gradle in the root project directory.
---property-upgrade-report          (Experimental) Runs build with experimental property upgrade report.
--q, --quiet                        Log errors only.
---refresh-keys                     Refresh the public keys used for dependency verification.
---rerun-tasks                      Ignore previously cached task results.
--S, --full-stacktrace              Print out the full (very verbose) stacktrace for all exceptions.
--s, --stacktrace                   Print out the stacktrace for all exceptions.
---scan                             Generate a Build Scan (powered by Develocity).
+-g, --gradle-user-home             Specifies the Gradle user home directory. Default is ~/.gradle.
+-I, --init-script                  Specifies an initialization script.
+-i, --info                         Sets the log level to info.
+--include-build                    Includes the specified build in the composite.
+-M, --write-verification-metadata  Generates checksums for dependencies used in the project. Accepts a comma-separated list.
+-m, --dry-run                      Runs the build with all task actions disabled.
+--max-workers                      Configures the maximum number of concurrent workers Gradle is allowed to use.
+--offline                          Runs the build without accessing network resources.
+-P, --project-prop                 Sets a project property for the build script (for example, -Pmyprop=myvalue).
+-p, --project-dir                  Specifies the start directory for Gradle. Default is the current directory.
+--parallel                         Builds projects in parallel. Gradle will attempt to determine the optimal number of executor threads to use.
+--no-parallel                      Disables parallel project execution.
+--priority                         Specifies the scheduling priority for the Gradle daemon and all processes launched by it. Supported values are 'normal' (default) or 'low'.
+--problems-report                  Enables the HTML problems report. [incubating]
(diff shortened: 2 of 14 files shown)


platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
```diff
@@ -28,67 +28,67 @@ To see a list of available tasks, run gradle tasks
 USAGE: gradle [option...] [task...]
 
 -?, -h, --help                     Shows this help message.
--a, --no-rebuild                   Do not rebuild project dependencies.
+-a, --no-rebuild                   Does not rebuild project dependencies.
```
alllex: 🤔  I'm not a native speaker, but this doesn't seem right

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
```diff
@@ -28,67 +28,67 @@ To see a list of available tasks, run gradle tasks
...
---no-continue                      Stop task execution after a task failure.
--D, --system-prop                  Set system property of the JVM (e.g. -Dmyprop=myvalue).
--d, --debug                        Log in debug mode (includes normal stacktrace).
---daemon                           Uses the Gradle daemon to run the build. Starts the daemon if not running.
---no-daemon                        Do not use the Gradle daemon to run the build. Useful occasionally if you have configured Gradle to always run with the daemon by default.
+--console                          Specifies which type of console output to generate. Supported values are 'plain', 'colored', 'auto' (default), 'rich', or 'verbose'.
+--console-unicode                  Specifies which character types are allowed in the console output. Supported values are 'auto' (default), 'disable', or 'enable'.
+--continue                         Continues task execution after a task failure.
+--no-continue                      Stops task execution after a task failure.
+-D, --system-prop                  Sets a JVM system property (fo example, -Dmyprop=myvalue).
```
alllex:
```suggestion
-D, --system-prop                  Sets a JVM system property (for example, -Dmyprop=myvalue).
```

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
```diff
@@ -28,67 +28,67 @@ To see a list of available tasks, run gradle tasks
...
+-d, --debug                        Sets log level to debug. Includes the normal stacktrace.
+--daemon                           Uses the Gradle daemon to run the build. Starts the daemon if it is not running.
+--no-daemon                        Does not use the Gradle daemon to run the build. Useful occasionally if you have configured Gradle to always run with the daemon by default.
```
alllex: 🤔  Here also seems fishy, same as `--no-rebuild`

platforms/core-runtime/cli/src/main/java/org/gradle/cli/SystemPropertiesCommandLineConverter.java
```diff
@@ -29,6 +29,6 @@ protected String getPropertyOptionDetailed() {
 
     @Override
     protected String getPropertyOptionDescription() {
-        return "Set system property of the JVM (e.g. -Dmyprop=myvalue).";
+        return "Sets a JVM system property (fo example, -Dmyprop=myvalue).";
```
eskatos: ❌ Typo

```suggestion
        return "Sets a JVM system property (for example, -Dmyprop=myvalue).";
```

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
eskatos: ❌ Same typo

```suggestion
-D, --system-prop                  Sets a JVM system property (for example, -Dmyprop=myvalue).
```

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
```diff
@@ -28,67 +28,67 @@ To see a list of available tasks, run gradle tasks
...
---continue                         Continue task execution after a task failure.
---no-continue                      Stop task execution after a task failure.
--D, --system-prop                  Set system property of the JVM (e.g. -Dmyprop=myvalue).
--d, --debug                        Log in debug mode (includes normal stacktrace).
---daemon                           Uses the Gradle daemon to run the build. Starts the daemon if not running.
---no-daemon                        Do not use the Gradle daemon to run the build. Useful occasionally if you have configured Gradle to always run with the daemon by default.
+--console                          Specifies which type of console output to generate. Supported values are 'plain', 'colored', 'auto' (default), 'rich', or 'verbose'.
+--console-unicode                  Specifies which character types are allowed in the console output. Supported values are 'auto' (default), 'disable', or 'enable'.
+--continue                         Continues task execution after a task failure.
```
lkasso, line 41: Can we add something about the default behavior without --continue, something like:

```suggestion
--continue                         Continues task execution after a task failure instead of failing the build.
```

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
```diff
@@ -28,67 +28,67 @@ To see a list of available tasks, run gradle tasks
...
+--no-continue                      Stops task execution after a task failure.
```
lkasso, line 42: Similar to the feedback above, can we mention something about the default, for example:

```suggestion
--no-continue                      Stops task execution after a task failure (default behavior).
```

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
lkasso: It is ok actually. 
If you want to make an improvement and match it a little more to the rest of the descriptions, consider:

```suggestion
-a, --no-rebuild                   Disables rebuilding of project dependencies.
```

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
lkasso: The english is ok but we can consider the alternative:

```suggestion
--no-daemon                        Disables the Gradle daemon for this build, even if the daemon is enabled by default.
```

Also: `Runs the build without the Gradle daemon. Useful occasionally if you have configured Gradle to always run with the daemon by default.`

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
```diff
@@ -28,67 +28,67 @@ To see a list of available tasks, run gradle tasks
...
--s, --stacktrace                   Print out the stacktrace for all exceptions.
---scan                             Generate a Build Scan (powered by Develocity).
+-g, --gradle-user-home             Specifies the Gradle user home directory. Default is ~/.gradle.
+-I, --init-script                  Specifies an initialization script.
+-i, --info                         Sets the log level to info.
+--include-build                    Includes the specified build in the composite.
+-M, --write-verification-metadata  Generates checksums for dependencies used in the project. Accepts a comma-separated list.
+-m, --dry-run                      Runs the build with all task actions disabled.
+--max-workers                      Configures the maximum number of concurrent workers Gradle is allowed to use.
+--offline                          Invokes the build without accessing network resources.
```
lkasso: We should try to keep the wording similar to the rest, wdyt of:

```suggestion
--offline                          Runs the build without accessing network resources.
```

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
```diff
@@ -28,67 +28,67 @@ To see a list of available tasks, run gradle tasks
...
+--continue                         Continues task execution after a task failure.
+--no-continue                      Stops task execution after a task failure.
+-D, --system-prop                  Sets a JVM system property (fo example, -Dmyprop=myvalue).
+-d, --debug                        Sets log level to debug. Includes the normal stacktrace.
+--daemon                           Uses the Gradle daemon to run the build. Starts the daemon if it is not running.
+--no-daemon                        Does not use the Gradle daemon to run the build. Useful occasionally if you have configured Gradle to always run with the daemon by default.
 --export-keys                      Exports the public keys used for dependency verification.
--F, --dependency-verification      Configures the dependency verification mode. Values are 'strict', 'lenient' or 'off'.
+-F, --dependency-verification      Configures the dependency verification mode. Supported values are 'strict', 'lenient', or 'off'.
 --foreground                       Starts the Gradle daemon in the foreground.
```
lkasso, line 49: Can we be more specific here? Does it start a new daemon in the foreground, or run the build itself in the foreground? What if there is already a daemon in the foreground?

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
donat: > instead of failing the build.

Problem is, the build will fail, just later. With the suggestion one might assume that the build will pass if `--continue` is specified.

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
donat: Makes sense, but I fear this would lead to another discussion about why we don't mark default options for all build options.

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
donat: It's good that you highlighted this option. I tried to use it and it seems broken. I'll investigate later. For now, I'd just leave the label as-is for now.

PR #36891, opened 2026-02-26
Provide mechanism to disallow further changes to domain object collections

### LINKED ISSUE #36881: Investigate how we might "finalize" NamedDomainObjectContainers
We want to be able to finalize definitions before running a feature's apply action.  Property objects on the definition can already be finalized, but NamedDomainObjectContainers can still be modified.

We should look into how we could prevent further mutation of NDOCs when we finalize the definition and see if there's something reasonably cheap we could do that wouldn't involve a lot of time/complexity.

ghale:
This adds a new `disallowChanges()` method to `DomainObjectCollection` that allows a container to marked as immutable and further changes (e.g. add, remove, etc) are no longer allowed.

This is needed for declarative gradle where we want to make collections in definition objects immutable before calling the feature's apply action.

Fixes #36881 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core-api/src/main/java/org/gradle/api/DomainObjectCollection.java
+++ b/subprojects/core-api/src/main/java/org/gradle/api/DomainObjectCollection.java
@@ -220,10 +220,21 @@ public interface DomainObjectCollection<T> extends Collection<T> {
      *
      * @throws IllegalStateException if called from within a lazy configuration action
      *         (e.g., inside a {@link #configureEach} callback)
+     *
      * @since 9.5.0
      */
     @Incubating
     default void disallowChanges() {
         throw new UnsupportedOperationException("disallowChanges() is not supported by this collection");
+    }
+
+    /**
+     * Returns {@code false} if {@link DomainObjectCollection#disallowChanges()} has been called on this collection.
+     *
+     * @since 9.5.0
+     */
+    @Incubating
+    default boolean areChangesAllowed() {
+        return true;
     }
 }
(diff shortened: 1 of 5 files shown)


ghale:
@bot-gradle test this

subprojects/core-api/src/main/java/org/gradle/api/DomainObjectCollection.java
```diff
@@ -203,4 +203,27 @@ public interface DomainObjectCollection<T> extends Collection<T> {
...
+     * prevents new elements from being added or existing elements from being removed.</p>
+     *
+     * <p>Calling this method more than once has no effect.</p>
+     *
+     * @throws IllegalStateException if called from within a lazy configuration action
+     *         (e.g., inside a {@link #configureEach} callback)
+     * @since 9.5.0
+     */
+    @Incubating
+    default void disallowChanges() {
```
smplio, line 227: If users can disallowChanges from public api, shouldn't they also be able to check, if modifications are allowed on a given DOC?

eskatos: +1 to also expose a way to read the state as part of this PR. This boils down to moving `DomainObjectCollectionInternal.areChangesDisallowed()` to the public `DomainObjectCollection`.

ghale: No objection to this.  Done.

PR #36897, opened 2026-02-27
Implement WorkerThreadPool for DefaultBuildOperationQueue

octylFractal:
This is not expected to have significant impact now, but if we can start using `blocking` for build cache I/O work then it will allow it to perform more concurrent work here.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/WorkerThreadPoolHelper.java
+++ b/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/WorkerThreadPoolHelper.java
@@ -20,6 +20,7 @@ import org.jspecify.annotations.Nullable;
 
 import java.util.ArrayDeque;
 import java.util.Deque;
+import java.util.function.Consumer;
 
 /**
  * Common code for handling work for {@link WorkerThreadPool} implementations.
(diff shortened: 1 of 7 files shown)


octylFractal:
@bot-gradle test PT

octylFractal:
@bot-gradle test PT

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/WorkerThreadPoolHelper.java
```diff
@@ -0,0 +1,98 @@
...
+ * Common code for handling work for {@link WorkerThreadPool} implementations.
+ *
+ * <p>
+ * This class is not thread-safe and must be externally synchronized.
+ * </p>
+ */
+public final class WorkerThreadPoolHelper<T> {
+    private final WorkerLimits workerLimits;
+    private final Runnable spawnWorker;
+    private final Deque<T> queue = new ArrayDeque<>();
```
reinsch82, line 56: 🎉 just wanted to suggest to use `ArrayDeque` instead :)

PR #36903, opened 2026-02-27
Add clickable links to files in console output upon failure

tresat:
Enhance failure output with robust clickable links.

IDEs often fails to recognize file paths containing special characters (e.g., spaces, unicode, or symbols) as clickable links in build failure output. This makes navigating to the source of an error more cumbersome.

This update enhances error reporting by:
- Modifying the exception handling to explicitly store the absolute file path where an error occurs.
- Introducing a mechanism to detect if a file path needs URI encoding to be clickable by IDEs.
- Automatically appending a properly encoded `file:` URI with the line number to the console output when necessary.

This ensures that all failure locations, even those in complex paths, provide a direct and clickable link in IDEs, significantly improving the developer experience.

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/internal/buildevents/BuildExceptionReporter.java
+++ b/subprojects/core/src/main/java/org/gradle/internal/buildevents/BuildExceptionReporter.java
@@ -30,6 +30,7 @@ import org.gradle.internal.enterprise.core.GradleEnterprisePluginManager;
 import org.gradle.internal.exceptions.CompilationFailedIndicator;
 import org.gradle.internal.exceptions.ContextAwareException;
 import org.gradle.internal.exceptions.FailureResolutionAware;
+import org.gradle.internal.exceptions.LocationAwareException;
 import org.gradle.internal.exceptions.NonGradleCause;
 import org.gradle.internal.exceptions.NonGradleCauseExceptionsHolder;
 import org.gradle.internal.exceptions.ResolutionProvider;
@@ -46,6 +47,7 @@ import org.gradle.util.internal.GUtil;
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
+import java.io.File;
 import java.io.StringWriter;
 import java.util.ArrayDeque;
 import java.util.Collection;
@@ -255,8 +257,8 @@ public class BuildExceptionReporter implements Action<Throwable> {
         }
 
         @Override
-        protected void visitLocation(String location) {
-            failureDetails.location.text(location);
+        protected void visitLocationException(LocationAwareException exception) {
+            failureDetails.location.text(exception.describeClickableLocation());
         }
 
         @Override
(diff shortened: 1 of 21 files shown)


subprojects/core/src/main/java/org/gradle/internal/buildevents/BuildExceptionReporter.java
```diff
@@ -545,6 +547,31 @@ static void renderStyledError(Failure failure, StyledTextOutput details) {
...
+     * Creates a properly formatted clickable link that IDEs can recognize in console output.
+     * The file path is converted to a {@code file:} URI with proper percent-encoding.
+     * When a line number is provided, it is appended using the {@code :lineNumber} format.
+     *
+     * @param filePath the absolute file path to format as a clickable link
+     * @param lineNumber the optional line number to append to the URI
+     * @return the formatted clickable link text
+     * @throws IllegalArgumentException if the file path cannot be converted to a URI
+     */
+    public static String formatClickableLink(String filePath, @Nullable Integer lineNumber) {
```
eskatos: We have `ConsoleRenderer.asClickableFileUrl()` already. Let's make it better instead of adding yet another different implementation. Moreover, having all the logic related to this colocated would be better than having the `needsUriEncoding` far away from this.

tresat: I somehow didn't notice that method while doing this work.  I'll remove this method and route through that method (adding an overload with lineNumber).

subprojects/core/src/main/java/org/gradle/internal/exceptions/LocationAwareException.java
```diff
@@ -104,4 +144,90 @@ public void appendResolutions(Context context) {
...
+     * @param location the base location string to potentially enhance
+     * @return the location string, with a clickable URI appended if the source path needs encoding
+     * @see #needsUriEncoding(String)
+     * @see ConsoleRenderer#asClickableFileUrl(File, Integer)
+     */
+    private String maybeAddClickableLocation(String location) {
+        if (needsUriEncoding(sourcePath)) {
+            try {
+                String url = new ConsoleRenderer().asClickableFileUrl(new File(sourcePath), lineNumber);
+                return location + "\n  (Clickable link: " + url + ")";
```
eskatos, line 168: ❓ I'm not excited about the parenthesis nor by the "Clickable link:" text. This also adds one more line and clickable link to our end of build output that is already pretty much all over the place. I understand the value of this one, it is probably more important than any other link we output at the end of a build. 

This looks like a lot to me:
```
...
* Where:
Build file '${buildFile.canonicalPath}' line: 1
  (Clickable link: file:///Users/paul/src/gradle-related/gradle/platforms/core-configuration/kotlin-dsl-tooling-models/build.gradle.kts:1)

* What went wrong:
Script compilation errors:
...
```

What about something leaner like:
```
...
* Where:
Build file '${buildFile.canonicalPath}' line: 1
file:///Users/paul/src/gradle-related/gradle/platforms/core-configuration/kotlin-dsl-tooling-models/build.gradle.kts:1

* What went wrong:
Script compilation errors:
...
```

tresat: Yes, Sterling and I feel the same way.  The "Clickable link: " prefix was just to make it clear what I was presenting.  We'll come up with something else.

big-guy:
Even without "Clickable link", I think this is too ugly. 

I was noticing something weird when playing around with the output. Sometimes the file paths were clickable and sometimes they were not clickable when they used to be.

There appears to be Gradle-specific code that looks for specific messages and linkifies them: https://github.com/JetBrains/intellij-community/blob/2f85fa0ae913385e09c6328ef83a4c6500bdb802/plugins/gradle/src/org/jetbrains/plugins/gradle/execution/GradleConsoleFilter.java#L42

But I think this is only triggered when IntelliJ knows Gradle just ran. Maybe we can ask JB about OSC8 support in the IntelliJ terminal and when available, use that consistently.

eskatos:
Leveraging OSC8 support would be so much better 👍

tresat:
> Sometimes the file paths were clickable and sometimes they were not clickable when they used to be.

Which paths?  The Clickable links paths?  I didn't ever notice them being non-clickable while testing this, do you have any ideas how to reproduce?  

I would like to experiment with OSC8 support here next and see if we can just do something to the current links to make them useful.

Overall though, I think the output from a typical build is already very wordy and kind of "ugly" - there are so many different ways of presenting links to output files, what's one more?  I'd rather have usable links that are visually redundant than more compact output with pretty links that don't work.

PR #36904, opened 2026-02-27
Apply optimizations to DirectorySnapshotter and PatternSetSnapshottingFilter

asodja:
...to reduce allocations

This saves about 2% (1.4GB) of allocations for "JavaIncrementalExecutionPerformanceTest.assemble for non-abi change"
Screenshot 2026-02-27 at 18 06 53

### DIFF
--- a/platforms/core-execution/snapshots/src/main/java/org/gradle/internal/snapshot/impl/DirectorySnapshotter.java
+++ b/platforms/core-execution/snapshots/src/main/java/org/gradle/internal/snapshot/impl/DirectorySnapshotter.java
@@ -513,6 +513,8 @@ public class DirectorySnapshotter {
         }
 
         private String getInternedFileName(Path path) {
+            // Path also has getFileName() but it creates additional allocations,
+            // and since this is on a hot path we optimized it
             String absolutePath = path.toString();
             int lastSep = absolutePath.lastIndexOf(File.separatorChar);
             return intern(lastSep < 0 ? absolutePath : absolutePath.substring(lastSep + 1));


asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test APT

platforms/core-execution/snapshots/src/main/java/org/gradle/internal/snapshot/impl/DirectorySnapshotter.java
```diff
@@ -513,9 +512,10 @@ private boolean shouldVisit(Path path, String internedName, boolean isDirectory)
...
         }
 
-        private String getInternedFileName(Path dir) {
-            Path fileName = dir.getFileName();
-            return fileName == null ? "" : intern(fileName.toString());
+        private String getInternedFileName(Path path) {
+            String absolutePath = path.toString();
+            int lastSep = absolutePath.lastIndexOf(File.separatorChar);
+            return intern(lastSep < 0 ? absolutePath : absolutePath.substring(lastSep + 1));
         }
```
mlopatkin, line 521: ❓ Why do we need a hand-written path parser here?

asodja: `dir.getFileName()` does more allocations. It parses a string (what we do here) and then it creates a new `Path`.  This is generally not a problem, but here in this case it actually creates a lot allocations. Also `path.toString()` is cached with Java 17 or 21. 

Note: this optimization was actually fabricated by our AI overlords

alllex: Then I'd add an explicit comment that it's important in this method to avoid extra allocations, to justify more complex code.

Please also consider moving this out into a shared utility class, if this can be useful elsewhere for the same reasons.

asodja: Added a comment. I will avoid adding it to a special utility class for now. Unless it's on a hot path I think we should just call what Java intended. If we find another such case, then we can move it to utility

PR #36907, opened 2026-02-27
Extract Ant-related classes from core and move to the software platform

ghale:
This extracts the ant-related classes from core and core-api and moves them to the following projects under the "software" platform:
- ant-api - public API classes related to Ant 
- ant - internal API classes that may or may not be used by modules in other platforms
- ant-impl - classes internal to ant that should not be used outside of the platform

It also moves ant-worker from the JVM platform to the software platform to logically group this with the other ant projects.  This makes sense because although only JVM stuff uses ant-worker, running an ant action in an isolated worker is a general capability that is not specific to JVM.  This change is not necessary to the overall goal of this PR and could be removed if it seems wrong.

There is one clunky issue that this does not solve:
- `WorkerDaemonServer` in core-execution/daemon-worker-server creates a service registry containing a `DefaultIsolatedAntBuilder` that lives in ant-impl.  This relationship is an architectural violation because the core platform should not depend on the software platform, nor should any external project depend on ant-impl (i.e. things should depend on `IsolatedAntBuilder` in `ant` instead).  This is, of course, an existing violation that was masked by having the classes in core and is largely related to how services are instantiated in the worker daemon at runtime.  We should find a way to inject this service at runtime so that it can be used in a worker daemon without creating this tight coupling that crosses architectural boundaries.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/groovydoc-worker/build.gradle.kts
+++ b/platforms/jvm/groovydoc-worker/build.gradle.kts
@@ -10,8 +10,6 @@ dependencies {
     api(projects.baseServices)
     api(projects.coreApi)
 
-    runtimeOnly(projects.core)
-
     implementation(projects.daemonServerWorker)
     implementation(libs.groovy)
     implementation(libs.guava)
--- a/platforms/jvm/jacoco-workers/build.gradle.kts
+++ b/platforms/jvm/jacoco-workers/build.gradle.kts
@@ -10,7 +10,6 @@ dependencies {
     api(projects.baseServices)
     api(projects.coreApi)
 
-    runtimeOnly(projects.core)
     api(projects.internalInstrumentationApi)
     api(libs.jspecify)
 


platforms/core-configuration/file-collections/src/main/java/org/gradle/api/internal/file/AntFileCollectionBuilder.java
```diff
@@ -40,7 +38,7 @@ public Object addToAntBuilder(Object node, String childNodeName) {
             @SuppressWarnings("unused") // Magic Groovy method
             public Object doCall(Object ignore) {
                 for (File file : files) {
-                    dynamicObject.invokeMethod("file", Collections.singletonMap("file", AntUtil.maskFilename(file.getAbsolutePath())));
+                    dynamicObject.invokeMethod("file", Collections.singletonMap("file", file.getAbsolutePath().replaceAll("\\$", "\\$\\$")));
```
ghale, line 41: Calling `AntUtil.maskFilename` here creates a tangling issue.  We _could_ move this to a separate class in some common module, but it's a simple String manipulation that only happens in a few places, so I opted to inline it instead.

subprojects/core/src/test/groovy/org/gradle/api/internal/project/DefaultProjectTest.groovy
```diff
@@ -187,7 +186,7 @@ class DefaultProjectTest extends Specification {
         rootDir = new File("/path/root").absoluteFile
         buildFile = new File(rootDir, TEST_BUILD_FILE_NAME)
 
-        testAntBuilder = new DefaultAntBuilder(null, antLoggingAdapter)
+        testAntBuilder = new BasicAntBuilder()
```
ghale, line 189: This test only needs an ant builder of some type, and doesn't specifically need a `DefaultAntBuilder`.  Using `DefaultAntBuilder` creates an unwanted dependency on the platform-internal `ant-impl` project, so we use a `BasicAntBuilder` here instead.

subprojects/core/src/testFixtures/groovy/org/gradle/api/tasks/AntBuilderAwareUtil.groovy
```diff
@@ -38,7 +38,7 @@ class AntBuilderAwareUtil {
     }
 
     static def assertSetContains(FileCollection set, Set<String> filenames, Iterable<FileCollection.AntType> types, boolean generic = true) {
-        AntBuilder ant = new DefaultAntBuilder(null, null)
+        AntBuilder ant = new BasicAntBuilder()
```
ghale, line 41: Same here - `BasicAntBuilder` is better if we can get away with using it.

ghale:
@bot-gradle test this

platforms/jvm/groovydoc-worker/build.gradle.kts
```diff
@@ -5,11 +5,13 @@ plugins {
 description = "Implements generation of groovydoc"
 
 dependencies {
+    api(projects.ant)
     api(projects.antWorker)
     api(projects.baseServices)
-    api(projects.core)
     api(projects.coreApi)
 
+    runtimeOnly(projects.core)
```
big-guy: I'm not sure if we need to include runtimeOnly deps.  Do our tests fail without it?

platforms/software/ant-impl/src/main/java/org/gradle/api/internal/project/antbuilder/DefaultIsolatedAntBuilder.java
```diff
@@ -86,7 +86,9 @@ public DefaultIsolatedAntBuilder(ClassPathRegistry classPathRegistry, ClassLoade
         this.baseAntLoader = new CachingClassLoader(new MultiParentClassLoader(antLoader, loggingLoader));
 
         // Need gradle core to pick up ant logging adapter, AntBuilder and such
-        ClassPath gradleCoreUrls = moduleRegistry.getModule("gradle-core-api").getImplementationClasspath();
+        ClassPath gradleCoreUrls = moduleRegistry.getModule("gradle-ant-api").getImplementationClasspath();
```
big-guy, line 89: If this requested gradle-ant-impl, would it get everything?

big-guy: nevermind, this is _inside_ ant-impl

platforms/jvm/groovydoc-worker/build.gradle.kts
ghale: Weirdly, it's not required in some of the places where it was added and it is required in others.  

I removed the runtimeOnly deps wherever it didn't break things.

PR #36914, opened 2026-03-01
Fix typo in documentation anchor

04cb:
Fixes typo in documentation anchor from 'implemention_plugins' to 'implementing_plugins'.


ljacomet:
We consider anchors to be a form of documentation API. So while fixing a typo sounds good, the potential impact on links out there does not make it worthwhile.

The [DCO](https://probot.github.io/apps/dco/) check is failing.

So given both of these issues, closing as we will not accept this change.

PR #36923, opened 2026-03-02
Use CAS to try to avoid unpark

jvandort:
By using CAS, we can potentially avoid calling LockSupport.unpark from multiple threads. Only one thread needs to call unpark, which is an expensive operation involving a syscall.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

PR #36927, opened 2026-03-02
Add progress display name for building task graph

octylFractal:
This can take a lot of time in builds and it's not clear to users what's happening sometimes, as the command-line will just say "100% CONFIGURING" and "Resolving [...]" during task graph dependency resolution, or just the "CONFIGURING" if it's doing CPU-bound work.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/internal/build/BuildOperationFiringBuildWorkPreparer.java
+++ b/subprojects/core/src/main/java/org/gradle/internal/build/BuildOperationFiringBuildWorkPreparer.java
@@ -121,9 +121,9 @@ public class BuildOperationFiringBuildWorkPreparer implements BuildWorkPreparer 
 
         private static String buildingTaskGraphDisplayName(GradleInternal gradle) {
             if (gradle.isRootBuild()) {
-                return "Building root build task graph";
+                return "Building task graph of root build";
             } else {
-                return "Building build '" + gradle.getIdentityPath().asString() + "' task graph";
+                return "Building task graph of build '" + gradle.getIdentityPath().asString() + "'";
             }
         }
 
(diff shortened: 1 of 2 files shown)


subprojects/core/src/main/java/org/gradle/internal/build/BuildOperationFiringBuildWorkPreparer.java
```diff
@@ -118,6 +119,14 @@ public String getBuildPath() {
                 });
         }
 
+        private static String buildingTaskGraphDisplayName(GradleInternal gradle) {
+            if (gradle.isRootBuild()) {
+                return "Building root build task graph";
```
alllex: 💅  Use a predictable prefix in both cases and move the variable part to the end.

In the absolute majority of cases, the users will see only the message for the root task graph and won't know or care about composite builds. In that case, the "root build" information is secondary and having the "task graph" words closer to the beginning make it easier to understand at a glance what's happening.

Having a common prefix also makes it easier to find these operations in tools like Perfetto traces.

My suggestion modulo prepositions:
```
Building task graph of/for root build

Building task graph of/for ':build-logic' included build
```

subprojects/core/src/main/java/org/gradle/internal/build/BuildOperationFiringBuildWorkPreparer.java
```diff
@@ -118,6 +119,14 @@ public String getBuildPath() {
...
+            } else {
+                return "Building build '" + gradle.getIdentityPath().asString() + "' task graph";
```
alllex: ❓  Should it say "task graph" or "work graph"? What do we currently do in other user-facing places?

octylFractal: We say "task graph", see https://docs.gradle.org/current/userguide/build_lifecycle.html#task_graph.

Issue #24864, opened 2023-04-21
Version chooser is broken on release notes page


ov7a:
### Issue type

Missing information

### Problem description

There is no option to choose a version on release notes page:
<img width="513" alt="image" src="https://user-images.githubusercontent.com/4946730/233656516-cf255a34-e2d6-4e28-9891-827e0beb98fe.png">
You can still find other release notes with a direct link like this https://docs.gradle.org/8.1/release-notes.html

And version chooser works appropriately on other pages, e.g. https://docs.gradle.org/current/userguide/what_is_gradle.html
<img width="633" alt="image" src="https://user-images.githubusercontent.com/4946730/233656893-b21681cf-e19b-4f1b-b261-cb11483adb50.png">

### Context (optional)

Firefox 112.0.1 on MacOS

### Page with the problem

https://docs.gradle.org/current/release-notes.html

pnguyen9:
Seems like the version selector has a tendency to close up when you try to move your mouse down to select a version, making it look like a prank selector.

![image](https://github.com/gradle/gradle/assets/22443655/68ad43b6-9426-4149-baea-6e99acad4831)

lkasso:
Additional request: I
This can be much improved by making two sections in the dropdown: recently released and recently visited.
Then the released-section will be stable and predictable, and the other section can be volatile, but for a good reason

PR #36935, opened 2026-03-04
Use 100% instead of 40px to position submenus

octylFractal:
This prevents the menu from closing when moving from the main menu to the submenu, as seen in https://github.com/gradle/gradle/issues/24864#issuecomment-1841126425

I didn't test this locally exactly, because `serveDocs` doesn't have the version selector. I manually edited the CSS on the live site to match.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs/src/docs/css/base.css
+++ b/platforms/documentation/docs/src/docs/css/base.css
@@ -541,7 +541,7 @@ h3.releaseinfo {
     display: none;
     width: 170px;
     background-color: white;
-    top: 40px;
+    top: 100%;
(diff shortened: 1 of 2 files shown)


octylFractal:
Comparison screenshots (left is old, right is new):
Screenshot 2026-03-03 at 7 21 32 PM Screenshot 2026-03-03 at 7 21 58 PM
Screenshot 2026-03-03 at 7 22 45 PM Screenshot 2026-03-03 at 7 22 59 PM

PR #36945, opened 2026-03-04
Deprecate passing Project as a dependency notation

### LINKED ISSUE #36801: Deprecate the use of org.gradle.api.Project as a dependency notation in DependencyHandler
Today, you can declare a `Project` object as a dependency in DependencyHandler.  For example, something like `implementation(project(':foo'))` resolves to `Project.project(String)` and returns a `Project` object which the notation parser in `DependencyHandler` then resolves to a `ProjectDependency`.  However, all we really need in order to resolve to a `ProjectDependency` is the project path (not the project itself) and reaching around and doing things with project objects is not something we want to endorse.

The `Dependencies` object (used by test suites, for example) instead has a `project(String)` method that takes the path and directly resolves to a `ProjectDependency`.  We should do something similar for `DependencyHandler` and deprecate passing a `Project` object to the notation parser.

ghale:
This deprecates passing a `Project` object as a dependency notation.  We add a `project(String)` method to `DependencyHandler` so anywhere someone is calling `Project.project(String)` from inside a `dependencies {}` block, they should transparently get the new method now.  Anyone passing a `Project` object in some other way will get a deprecation warning.

This also introduces a `project(String)` method to the native `ComponentDependencies` interface.  This allows native projects to continue to use the same method signature without needing to resort to awkward calls to `DependencyFactory` to create project dependencies in build scripts.

Fixes #36801 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic/integration-testing/src/main/kotlin/gradlebuild.integration-tests.gradle.kts
+++ b/build-logic/integration-testing/src/main/kotlin/gradlebuild.integration-tests.gradle.kts
@@ -53,7 +53,7 @@ configurations.named("integTestImplementation") {
 
     dependencies.apply {
         // Use a provider so we don't lookup the project when we generate accessors
-        addLater(provider { dependencyFactory.create(project.project(":internal-integ-testing")) })
+        addLater(provider { dependencyFactory.createProjectDependency(":internal-integ-testing") })
     }
 }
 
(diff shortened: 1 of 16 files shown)


ghale:
@bot-gradle test this

ghale:
Superseded by #37227

PR #36948, opened 2026-03-04
Extract credentials/authentication classes out to software platform

ghale:
This extracts the `org.gradle.api.credentials` and `org.gradle.internal.authentication` packages out of core/core-api and places them in three new projects under the "software" platform:

credentials-api - public API classes related to credentials
credentials - internal API classes that may or may not be used by modules in other platforms
credentials-impl - classes internal to the credentials module that should not be used outside of the platform

There is one questionable relationship between dependency-management and credentials-impl which is related to `DefaultBaseRepositoryFactory` directly instantiating `DefaultAuthenticationContainer`.  If we replaced that with an `AuthenticationContainerFactory` service (or some such) we could change this to a dependency on credentials instead.  In any event, dependency-management and credentials-impl are both in the software platform, so this relationship does not seem overly problematic.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/credentials/build.gradle.kts
+++ b/platforms/software/credentials/build.gradle.kts
@@ -5,7 +5,7 @@ plugins {
 description = "Internal API for Gradle authentication schemes"
 
 dependencies {
-    api(projects.baseServices)              // @NonExtensible (org.gradle.api.NonExtensible)
-    api(projects.credentialsApi)            // Credentials + Authentication interfaces
-    api(projects.stdlibJavaExtensions)      // @ServiceScope, Scope
+    api(projects.baseServices)
+    api(projects.credentialsApi)
+    api(projects.stdlibJavaExtensions)
 }
--- a/platforms/software/credentials-api/build.gradle.kts
+++ b/platforms/software/credentials-api/build.gradle.kts
@@ -5,9 +5,10 @@ plugins {
 description = "Public API interfaces for Gradle credentials"
 
 dependencies {
-    api(projects.baseServices)  // @NonExtensible (org.gradle.api.NonExtensible)
+    api(projects.baseServices)
     api(libs.jspecify)
 
-    // @ToBeReplacedByLazyProperty — follow the same pattern as core-api for this annotation
-    compileOnly(projects.internalInstrumentationApi)
+    compileOnly(projects.internalInstrumentationApi) {
+        because("Provides @ToBeReplacedByLazyProperty annotation, following the same pattern as core-api")
+    }
 }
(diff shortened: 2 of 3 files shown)


ghale:
@bot-gradle test this

platforms/software/credentials/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
+plugins {
+    id("gradlebuild.distribution.implementation-java")
+}
+
+description = "Internal API for Gradle authentication schemes"
```
tresat, line 5: 🤔 Would it be clearer to call this `credentials-internal-api`?  

I know that's a break from current convention, but just `credentials` seems less explanatory than ideal - it doesn't really suggest what's in here.

platforms/software/credentials/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
...
+
+dependencies {
+    api(projects.baseServices)              // @NonExtensible (org.gradle.api.NonExtensible)
```
tresat: 💅 Could use `because` instead of comments for these.  Applies elsewhere.

platforms/software/resources-http/build.gradle.kts
```diff
@@ -7,8 +7,8 @@ description = "Implementation for interacting with repositories over HTTP"
 dependencies {
     api(projects.stdlibJavaExtensions)
     api(projects.serviceProvider)
-    api(projects.coreApi)
-    api(projects.core)
+    api(projects.credentials)
+    api(projects.credentialsApi)
```
tresat, line 11: 💅 I know that this is just the `dependency-analysis` plugin requiring both, but if `credentials` here is strictly the _internal_ API types, than it seems like that should imply the _public_ API types and it's a little odd to see both.

Just an observation.

ghale: Not sure we could remove it.  It's a transitive dependency, of course, but `:projectHealth` requires us to declare these direct dependencies rather than relying on transitives.

tresat: Yeah, just a consequence of that plugin.  Though if the `credentials` project is strictly internal API, we could have a global exclusion for needing the public API to be declared if internal is present.  Probably not worth doing, just thinking out loud here.

platforms/software/credentials/build.gradle.kts
ghale: Yeah, good point.  Some of these are not very useful anyways.  I've removed the ones that aren't all that helpful rather than adding more lines/noise to the file.  I've left the ones that are interesting and changed them to `because()` statements.

platforms/software/credentials/build.gradle.kts
ghale: I'm not hostile to the idea of adding something here, but I find `internal-api` to be too wordy.  I don't have a better idea, but I'll think on it.  I'm not going to hold this PR up, though, since we're just following the existing pattern.

PR #36950, opened 2026-03-04
Add note so DV team is warned when file access journal changes

abstratt:
Fixes: https://github.com/gradle/gradle-private/issues/5037

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


subprojects/core/src/main/java/org/gradle/api/internal/changedetection/state/DefaultFileAccessTimeJournal.java
```diff
@@ -32,9 +33,14 @@
 import static org.gradle.internal.serialize.BaseSerializerFactory.FILE_SERIALIZER;
 import static org.gradle.internal.serialize.BaseSerializerFactory.LONG_SERIALIZER;
 
+// Changes to file system layout or file format are allowed, but should be communicated to the DV team
+// See: https://github.com/gradle/dv/issues/59928
```
asodja: ❓ Is it ok we link dv project in our code?

abstratt: Replaced with a `gradle-private` link

PR #36952, opened 2026-03-04
Parallelize Gradle API runtime jar generation

jvandort:
Whenever a distribution is used for the first time on a machine, Gradle needs to generate the Gradle API jar. In future changes, we may consider computing this jar at build-time, however this requires extensive changes. Currently however, this generation process takes ~3.5 seconds.

We parallelize this process, reducing the generation time to ~1.5-2 seconds. In addition, we remove a regex from a very often called method in ImplementationDependencyRelocator, replacing it with an implementation that performs minimal allocation, reducing the GC pressure impact we add due to parallelizing this process.

For a simple build with one kotlin build file and one settings script file, this reduces first use (cold daemon, empty GUH) from ~12.5 seconds to ~10.5 seconds, a ~15% improvement.
image

Before:
image

After:
image

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/runtimeshaded/RuntimeShadedJarCreator.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/runtimeshaded/RuntimeShadedJarCreator.java
@@ -57,8 +57,8 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.PriorityQueue;
 import java.util.Set;
-import java.util.TreeSet;
 import java.util.function.Consumer;
 
 import static java.util.Arrays.asList;
@@ -135,7 +135,7 @@ class RuntimeShadedJarCreator {
                 }
             });
         } finally {
-            writer.stop(Duration.ofMinutes(1));
+            writer.stop(Duration.ofMinutes(5));
         }
 
         writeServiceFiles(builder, services);
@@ -149,7 +149,7 @@ class RuntimeShadedJarCreator {
      * The processed and remapped contents of a file that is to be included
      * in the relocated jar.
      */
-    private static class InputFile implements Comparable<InputFile> {
+    private static final class InputFile implements Comparable<InputFile> {
 
         private final int index;
         private final File file;
@@ -213,7 +213,7 @@ class RuntimeShadedJarCreator {
     ) {
         return new MultiProducerSingleConsumerProcessor<>("shaded jar writer", new Consumer<InputFile>() {
             private int index = 0;
-            private final TreeSet<InputFile> allProcessedFiles = new TreeSet<>();
+            private final PriorityQueue<InputFile> allProcessedFiles = new PriorityQueue<>();
             private final Set<String> seenPaths = new HashSet<>();
 
             @Override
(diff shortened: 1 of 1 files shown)


jvandort:
@bot-gradle test this

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/runtimeshaded/RuntimeShadedJarCreator.java
```diff
@@ -105,6 +145,104 @@ private void processFiles(ClasspathBuilder.EntryBuilder builder, Iterable<? exte
         progressLogger.progress(progressFormatter.incrementAndGetProgress());
     }
 
+    /**
+     * The processed and remapped contents of a file that is to be included
+     * in the relocated jar.
+     */
+    private static class InputFile implements Comparable<InputFile> {
```
octylFractal: 💅 
```suggestion
    private static final class InputFile implements Comparable<InputFile> {
```

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/runtimeshaded/RuntimeShadedJarCreator.java
```diff
@@ -83,19 +104,38 @@ public void create(RuntimeShadedJarType type, final File outputJar, final Iterab
...
+
+                        @Override
+                        public BuildOperationDescriptor.Builder description() {
+                            return BuildOperationDescriptor.displayName("Visiting " + file.getName());
+                        }
+                    });
+                }
+            });
+        } finally {
+            writer.stop(Duration.ofMinutes(1));
```
octylFractal: ❓ Is a minute enough time? On some slower machines I could see this taking a bit, maybe something like 10 minutes would be better?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/runtimeshaded/RuntimeShadedJarCreator.java
```diff
@@ -105,6 +145,104 @@ private void processFiles(ClasspathBuilder.EntryBuilder builder, Iterable<? exte
...
+
+    private static MultiProducerSingleConsumerProcessor<InputFile> createShadedJarWriter(
+        ClasspathBuilder.EntryBuilder builder,
+        ProgressLogger progressLogger,
+        PercentageProgressFormatter progressFormatter,
+        Map<String, List<String>> services
+    ) {
+        return new MultiProducerSingleConsumerProcessor<>("shaded jar writer", new Consumer<InputFile>() {
+            private int index = 0;
+            private final TreeSet<InputFile> allProcessedFiles = new TreeSet<>();
```
octylFractal: :x: I think this would be more understandable (and probably faster?) if implemented as a `PriorityQueue` instead.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/runtimeshaded/RuntimeShadedJarCreator.java
jvandort: I sure hope its enough time. This takes ~2 seconds on an M1 mac. For it to take 30x at long, youd need a really slow computer. We have integ tests which pop once this gets longer than 10 seconds (and they started popping on our CI windows machines, which triggered this PR). 

One minute _should_ be way long enough, but in any case I've increased this to 5.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/runtimeshaded/RuntimeShadedJarCreator.java
jvandort: Replaced. I doubt we get much performance out of this, but PQ is the right tool for the job

octylFractal (approved):

Issue #36961, opened 2026-03-05
9.4.0: Unable to find a variant with the requested capability: coordinates 'org.jetbrains.kotlin:kotlin-gradle-plugin-api-gradle813'


martinbonnin:
### Current Behavior

The following build script:

```kotlin
buildscript {
  repositories {
    mavenCentral()
  }

  dependencies {
    classpath("org.jetbrains.kotlin:kotlin-serialization:2.2.0")
    classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:2.3.0")
  }
}
```

 does not resolve:

```
$ ./gradlew build
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'test-gradle-9.4.0'.
> Could not resolve all artifacts for configuration 'classpath'.
   > Could not resolve org.jetbrains.kotlin:kotlin-gradle-plugin-api:2.2.0.
     Required by:
         buildscript of root project 'test-gradle-9.4.0' > org.jetbrains.kotlin:kotlin-gradle-plugin:2.3.0 > org.jetbrains.kotlin:kotlin-gradle-plugins-bom:2.3.0
      > Unable to find a variant with the requested capability: coordinates 'org.jetbrains.kotlin:kotlin-gradle-plugin-api-gradle813':
           - Variant 'apiElementsWithFixedAttribute' provides 'org.jetbrains.kotlin:kotlin-gradle-plugin-api:2.3.0'
           - Variant 'gradle80ApiElements' provides 'org.jetbrains.kotlin:kotlin-gradle-plugin-api:2.3.0'
           - Variant 'gradle80JavadocElements' provides 'org.jetbrains.kotlin:kotlin-gradle-plugin-api:2.3.0'
           - Variant 'gradle80RuntimeElements' provides 'org.jetbrains.kotlin:kotlin-gradle-plugin-api:2.3.0'
           - Variant 'gradle80SourcesElements' provides 'org.jetbrains.kotlin:kotlin-gradle-plugin-api:2.3.0'
           - Variant 'gradle811ApiElements' provides 'org.jetbrains.kotlin:kotlin-gradle-plugin-api:2.3.0'
           - ...
```

This is a behaviour change between `20260116012355+0000` (good) and `20260117005955+0000` (bad), looks related to https://github.com/gradle/gradle/pull/36292.

### Expected Behavior

Resolution happens as expected

### Self-contained Reproducer Project

```kotlin
buildscript {
  repositories {
[...]

PR #36962, opened 2026-03-05
Do not inherit stale edge failure

jvandort:
https://github.com/gradle/gradle/pull/36292 introduced a bug where if there was a stale edge yet-to-be-removed from the graph that has failed for some reason, and a new edge has been deferred due to a constraint, the deferring constraint might erroneously inherit the failure from the stale failed edge instead of attaching the edge that it deferred and adopting its target node.

We resolve this by attaching all unattached edges targeting the given module, without short circuiting, and only inheriting failures if no target nodes exist.

We also improve an error message that might be misleading, as it did not include the actual selected component name. 

Fixes https://github.com/gradle/gradle/issues/36961

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/constraints/DependencyConstraintsBugsIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/constraints/DependencyConstraintsBugsIntegrationTest.groovy
@@ -22,6 +22,8 @@ import spock.lang.Issue
 
 class DependencyConstraintsBugsIntegrationTest extends AbstractHttpDependencyResolutionTest {
 
+    def resolve = new ResolveTestFixture(testDirectory)
+
     // Ideally this should be a reproducer using generated dependencies but I wasn't able
     // to figure out a reproducer
     @Issue("https://github.com/gradle/gradle/issues/13960")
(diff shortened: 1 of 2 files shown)


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test rfr

jvandort:
@bot-gradle test rfr

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesLocalComponentIntegrationTest.groovy
```diff
@@ -157,7 +157,7 @@ class CapabilitiesLocalComponentIntegrationTest extends AbstractIntegrationSpec
         fails("resolve")
 
         then:
-        failure.assertHasCause("Unable to find a variant with the requested capabilities: [coordinates 'org:capability', feature 'foo']")
+        failure.assertHasCause("Unable to find a variant of 'project :other' with the requested capabilities: [coordinates 'org:capability', feature 'foo']")
```
ljacomet, line 160: 💅 While unrelated to the changes here, I wonder about the benefit of having `coordinates 'some:coord'`. I understand the benefit of `feature 'someFeature'` but not the other one and it adds more text to read.

jvandort: Tom has also brought up the wording here. This is something we can change separately

PR #36976, opened 2026-03-06
Remove unnecessary areChangesAllowed() method 

ghale:
This method was added by #36891.  

After some internal discussion, we've decided that:
1) Querying state is not an existing general capability of lazy things (e.g. `Property` has a `disallowChanges()` method but no way to query that state).
2) Exposing this could lead to some anti-patterns that we would rather not enable (e.g. working around the fact that a container has been locked against further changes).

This is not necessary for anything internal to Gradle, so we're reverting this for now.  We can add it back if sound use cases emerge that require it.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs/src/docs/release/notes.md
+++ b/platforms/documentation/docs/src/docs/release/notes.md
@@ -94,8 +94,7 @@ plugins {
 
 ### Domain Object Collections can be made immutable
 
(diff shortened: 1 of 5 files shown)


ghale:
@bot-gradle test this

PR #36979, opened 2026-03-07
Update Documentation Infrastructure: Improve the documentation README for AI

lkasso:
### Details

### Context
- Create an AI ingestible README for the User Manual of Gradle

### DIFF
--- a/platforms/documentation/docs/README.md
+++ b/platforms/documentation/docs/README.md
@@ -68,6 +68,19 @@ The `t` and `-PquickDocs` flags can be used by the `serveDocs` task as well.
 - [Asciidoctor User Manual](https://asciidoctor.org/docs/user-manual/)
 - [Asciidoctor Gradle Plugin Reference](https://asciidoctor.org/docs/asciidoctor-gradle-plugin/)
 
+### Cross-References and Linking
+
+Good cross-references help readers navigate, but too many become disruptive. 
+
+Follow these guidelines:
+
+- **Be selective.** Every link adds cognitive load and risks pulling readers away from the page. Only include links that are genuinely useful.
+- **Provide context on the page.** When a term, concept, or brief set of steps can be explained in a few sentences, do so rather than linking out.
+- **Avoid duplicate links.** Link to the same destination only once per page, unless sections are far apart or serve different entry points (e.g., a procedure and a troubleshooting section).
+- **Link to the most relevant destination.** Target the most specific relevant page or heading. Don't provide multiple links that serve the same purpose.
+
+**For third-party content**, avoid links unless absolutely necessary. If a brief explanation covers what readers need, provide it on the page. Reserve **third-party links** for cases where the full external resource is genuinely required.
+
 ### Linking to DSL and API References
 
 Whenever you reference a Gradle API class, method, or annotation in prose, link it to the relevant reference documentation. Three path attributes are available:


platforms/documentation/docs/README.md
```diff
@@ -1,181 +1,491 @@
...
+### Build Commands
 
-    ### h3 Some feature (i)
+| Goal                                        | Command                                   |
+|---------------------------------------------|-------------------------------------------|
+| Full preview (recommended)                  | `./gradlew stageDocs`                     |
+| Full preview with continuous rebuild        | `./gradlew stageDocs -t`                  |
+| Fast iteration (no DSL ref, no single-page) | `./gradlew stageDocs -PquickDocs`         |
+| Fast iteration with continuous rebuild      | `./gradlew stageDocs -PquickDocs -t`      |
+| Live reload at `http://localhost:8000`      | `./gradlew serveDocs -PquickDocs`         |
```
tresat: ❗ Duplicate of 2 lines above.

platforms/documentation/docs/README.md
```diff
@@ -1,181 +1,491 @@
...
+```
 
-To generate the user manual for the final preview and see all changes, you normally want to run:
+Always wrap link text in backticks for any code identifier — classes, methods, properties, and annotations alike.
 
-    ./gradlew stageDocs
+### Images
 
-That will generate all the docs in the `build/docs` directory.
+Images live in `docs/src/userguide/img/`. Accepted formats are GIF, GRAPHML, SVG, PNG, and JPEG. Keep file sizes small — avoid large uncompressed images.
```
tresat: 🤔 If this is a guide for AI, should we mention not to generate images?  Or, better, link to the public AI policy doc once it's available?

platforms/documentation/docs/README.md
```diff
@@ -1,181 +1,491 @@
...
-This will automatically enter continuous build and rebuild the documentation when you make changes. It takes 30-40 seconds to rebuild the documentation in "quick mode".
+Every heading should have an anchor declared on the line immediately above it. This enables direct linking from other pages. Anchor IDs should use `snake_case`.
 
-If you really want to generate just the user manual, you can run:
+| Heading level           | Anchor required? |
+|-------------------------|------------------|
+| `=` (page title)        | Required         |
+| `==` (section)          | Required         |
+| `===` (subsection)      | Recommended      |
+| `====` (sub-subsection) | Recommended      |
```
tresat:
```suggestion
| `===` (subsection)      | As Needed        |
| `====` (sub-subsection) | As Needed        |
```

platforms/documentation/docs/README.md
```diff
@@ -1,181 +1,491 @@
...
+
+### Checking for Broken Links
+
+Always run the following after making changes to ensure no internal links are broken:
+
+```bash
+./gradlew :docs:checkDeadInternalLinks
+```
+
+## Code Snippets
```
tresat, line 183: 🤔 Code snippets could perhaps be a separate page, to keep context small.

platforms/documentation/docs/README.md
```diff
@@ -1,181 +1,491 @@
...
+        └── buildServices.sample.conf # Exemplar config (covers both DSLs)
+```
+
+**What belongs in each directory:**
+
+| Directory  | Contains                                                                    |
+|------------|-----------------------------------------------------------------------------|
+| `groovy/`  | Groovy DSL source files (`.gradle`)                                         |
+| `kotlin/`  | Kotlin DSL source files (`.gradle.kts`)                                     |
+| `common/`  | Files shared between both DSLs (e.g. `gradle.properties`, version catalogs) |
```
tresat: 🤔 Also, Java source and tests.

platforms/documentation/docs/README.md
```diff
@@ -1,181 +1,491 @@
...
+| `tests/`   | Testing instructions and expected outputs (e.g. `*.out`, `*.sample.conf`)   |
+
+**What must NOT be in these directories:**
+- Gradle wrapper files (`gradlew`, `gradlew.bat`, `gradle/wrapper/`)
+- Build output directories (`build/`)
+- Any other Gradle infrastructure files
```
tresat: 💅 What files are you trying to exclude here?  I'm not sure what this means.

platforms/documentation/docs/README.md
```diff
@@ -206,85 +516,209 @@ This graph is generated by [asciiflow.com](https://asciiflow.com/). You can copy
...
+
+### Testing Snippets with Exemplar
+
+Gradle uses [Exemplar](https://github.com/gradle/exemplar) to automatically test all snippets in `src/snippets/`. Exemplar discovers, executes, and verifies the output of snippet projects by looking for `*.sample.conf` files inside each snippet's `tests/` directory.
+
+When a snippet is tested, Exemplar copies the snippet to a temporary directory, runs the commands declared in the `.sample.conf` file, and compares the actual output against the corresponding `.out` file. If the output doesn't match, the test fails.
+
+By default, all snippets automatically go through a basic sanity check equivalent to running `./gradlew check` — this happens as part of CI without any manual intervention. You do not need to trigger this yourself (see [`BaseSamplesTest`](https://github.com/gradle/gradle/blob/a503d4a36c53e43a8857da3115fa744612c6ad36/subprojects/docs/src/docsTest/java/org/gradle/docs/samples/BaseSamplesTest.java#L66)).
```
tresat: ❗ I don't think the sanity check here actually tries to run the tests.  Isn't that left to the actual test-and-compare output step?  I think it just compiles them.  Running `check` in a typical JVM project _would_ execute the tests though.

platforms/documentation/docs/README.md
```diff
@@ -206,85 +516,209 @@ This graph is generated by [asciiflow.com](https://asciiflow.com/). You can copy
...
+A `.out` file is a plain text snapshot of the expected console output for a snippet command. Exemplar compares the actual output of running the snippet against this file to verify correctness.
+
+A typical `.out` file looks like this:
+
 ```
-./gradlew :docs:docsTest --tests "*.snippet-java-toolchain-task_kotlin_*"
+> Task :myTask
+Some result
+
+BUILD SUCCESSFUL in 0s
```
tresat: 🤔 I think we should note that including a time here "0s", for all but the most trivial examples, introduces something that's a great source of false negatives if the time doesn't exactly match.  I usually leave it out.

platforms/documentation/docs/README.md
```diff
@@ -206,85 +516,209 @@ This graph is generated by [asciiflow.com](https://asciiflow.com/). You can copy
...
+
+The snippet folder path maps to a test filter string by replacing `/` separators with `-` and dropping the `snippets/` prefix — e.g. `snippets/buildlifecycle/flowAction/` becomes `snippet-buildlifecycle-flow-action_*`.
+
+### Testing with Configuration Cache
+
+```bash
 ./gradlew :docs:docsTest --tests "*.snippet-java-toolchain-task_*" -PenableConfigurationCacheForDocsTests=true
 ```
 
+You can also set `enableConfigurationCacheForDocsTests=true` in `gradle.properties`.
```
tresat: 🤔 Should clarify which `gradle.properties` you mean here - not the one in the test example itself.

platforms/documentation/docs/README.md
```diff
@@ -206,85 +516,209 @@ This graph is generated by [asciiflow.com](https://asciiflow.com/). You can copy
...
+
+### Snippets Used in Integration Tests
+
+Some snippets in `src/snippets/` are also used as fixtures in Gradle's integration tests via the `@UsesSample` annotation in the `internal-integ-testing` module. This means that renaming, moving, or modifying a snippet can break integration tests that you may not be aware of.
+
+Before renaming or restructuring a snippet, search the codebase for any references to it via `@UsesSample`:
+
+```java
+@UsesSample("java/application")
```
tresat, line 648: TIL - I somehow hadn't encountered this before.

platforms/documentation/docs/README.md
```diff
@@ -1,181 +1,491 @@
...
+[source, kotlin]
+----
+pluginManagement {
+    repositories {
+        gradlePluginPortal()
+        maven {
+            url = uri("<path-to>/convention-plugins/build/repo")
+        }
+    }
+}
```
tresat: ❗ I don't think this guide should show example code here.  This makes it look like inlining Gradle DSL in docs files is okay, as long as the DSL is wrapped in a `====` block, which we want to avoid.

platforms/documentation/docs/README.md
lkasso: I will link to the public AI policy at the top of the README doc once it's available. 
I don't want to accept images in PRs (AI-generated or not) anyway, so I added it. Thanks for this.

platforms/documentation/docs/README.md
lkasso: I prefer a single doc if possible. It's the only place to get all the info for our contributors AND the AI.

platforms/documentation/docs/README.md
lkasso: I thought I could find more file types to exclude. I did not, so I deleted this section.

platforms/documentation/docs/README.md
lkasso: Good point. Removed.

platforms/documentation/docs/README.md
lkasso: I encounter it all the time! I am going to get help from dev-prod to handle this differently. I don't think these should live in docs tbh.

platforms/documentation/docs/README.md
lkasso: LMK if you are ok with the change I made.

platforms/documentation/docs/README.md
lkasso: Updated to a shorter example as per the 3-line rule.

platforms/documentation/docs/README.md
lkasso: Yup, you are right, I thought it ran a `check` by default, but I actually went through the doc test code, and only the `.sample.conf` tests are run. I updated the description accordingly.

platforms/documentation/docs/README.md
tresat: I don't see a related change?

The point I'm trying to clarify is that readers could interpret this to mean either set the `gradle.properties` at the root of the `gradle/gradle` build, or add a `gradle.properties` to the example and set the property in that.

platforms/documentation/docs/README.md
tresat: I see the argument that most substantial changes to docs will include changes to snippets, so keep them together.  I think both AI and human readers wouldn't have an issue following a link.  I suppose it's not a big deal either way.

platforms/documentation/docs/README.md
lkasso: I changed it to: 

> You can also set `enableConfigurationCacheForDocsTests=true` in the `gradle.properties` file in the root of the `gradle/gradle` repository.

PR #36981, opened 2026-03-07
Improve errors for unknown services in repository metadata rules

alllex:
The error messages become less generic and more actionable.

This also removes the last usage of `ServiceRegistry.EMPTY`, which allows to drop it entirely, making further simplifications to the service registries easier.

### DIFF
--- a/platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/repositories/AbstractArtifactRepositoryTest.groovy
+++ b/platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/repositories/AbstractArtifactRepositoryTest.groovy
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.api.internal.artifacts.repositories
+
+import org.gradle.api.artifacts.repositories.RepositoryResourceAccessor
+import org.gradle.api.model.ObjectFactory
+import org.gradle.internal.service.ServiceLookup
+import org.gradle.internal.service.UnknownServiceException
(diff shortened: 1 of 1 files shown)


alllex:
@bot-gradle test this on linux without PTS

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/AbstractArtifactRepository.java
```diff
@@ -231,12 +230,15 @@ public Object get(Type serviceType) throws UnknownServiceException, ServiceLooku
             if (service != null) {
                 return service;
             }
-            return ServiceRegistry.EMPTY.get(serviceType);
+            String availableServices = repositoryResourceAccessor != null
+                ? "ObjectFactory, RepositoryResourceAccessor"
+                : "ObjectFactory";
```
tresat: 🤔 Seems bad that these are hardcoded strings instead of a generated list.  They should at least use `ObjectFactory.class.getSimpleName()` to handle refactoring.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/AbstractArtifactRepository.java
```diff
@@ -231,12 +230,15 @@ public Object get(Type serviceType) throws UnknownServiceException, ServiceLooku
...
+            throw new UnknownServiceException(serviceType, "Service of type " + serviceType + " is not available for repository metadata rules. Available services: " + availableServices + ".");
         }
 
         @Override
         public Object get(Type serviceType, Class<? extends Annotation> annotatedWith) throws UnknownServiceException, ServiceLookupException {
-            return ServiceRegistry.EMPTY.get(serviceType, annotatedWith);
+            throw new UnknownServiceException(serviceType, "Service of type " + serviceType + " annotated with @" + annotatedWith.getSimpleName() + " is not available for repository metadata rules.");
```
tresat: 🤔 Why does this not list all available services as well?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/AbstractArtifactRepository.java
alllex: I went for string literals for simplicity, but I don't mind using the class literals. Updated

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/AbstractArtifactRepository.java
alllex: Made it print the services as well.

alllex:
@tresat, I've addressed your feedback. PTAL!

PR #36982, opened 2026-03-07
Replace ContainsServices with AbstractServiceRegistry

alllex:
for ease of testing

This allows to remove a number of mocking classes without reducing the test coverage.


alllex:
@bot-gradle test this

PR #36985, opened 2026-03-08
Extract process and worker process infrastructure from core into core-runtime and core-execution platforms

ghale:
This extracts the classes related to process handling and worker processes from core and core-api into three projects:
- process-services-api - (new in core-runtime) these are public api classes related to process handling
- process-services - (existing in core-runtime) internal api classes related to process handling
- worker-process-services - (new in core-execution) internal api classes related to starting and configuring worker processes

There are five classes in `org.gradle.process` that are left in core-api because they're deeply tangled with the task annotations and providers packages.  We're leaving these her for now until we can resolve the tangling either by extracting some of these packages or through breaking changes in a major.

These changes also extract a few interfaces into process-services related to Java modularity.  This is done to avoid platform access violations (i.e. process-services in core-runtime needing to depend on jvm-services in the jvm platform).  This is not ideal as these classes are not directly related to process management.  In a follow up PR, I'll pull these out, either into a better existing module (if it makes sense) or into their own module.  

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/core-runtime/client-services/src/main/java/org/gradle/launcher/daemon/client/DefaultDaemonStarter.java
```diff
@@ -234,7 +236,11 @@ private DaemonStartupInfo startProcess(List<String> args, File workingDir, File
             DaemonOutputConsumer outputConsumer = new DaemonOutputConsumer();
 
             // This factory should be injected but leaves non-daemon threads running when used from the tooling API client
-            RootClientExecHandleBuilderFactory execActionFactory = RootClientExecHandleBuilderFactory.of(gradleUserHome);
```
ghale, line 237: `gradleUserHome` wasn't being used by the `of()` method.  I think it was refactored out at some point, but never removed.

ghale:
@bot-gradle test this

blindpirate:
@bot-gradle test this

PR #36986, opened 2026-03-08
Remove `ServiceLookup.get` method with annotation parameter

alllex:
Since it has not been meaningfully implemented and was unused


alllex:
@bot-gradle test without PTS

PR #36987, opened 2026-03-09
Write performance DB profiling output to file instead of stdout

blindpirate:
Closes https://github.com/gradle/gradle-private/issues/5105

In the past we wrote performance DB profiling log output to stdout directly, which messed up the test output. This PR changes it to write to a profiling file instead of stdout.

Verified in https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_PerformanceTestTestLinuxbucket7/109795024

### DIFF
--- a/build-logic/performance-testing/src/main/groovy/gradlebuild/performance/reporter/PerformanceReporter.groovy
+++ b/build-logic/performance-testing/src/main/groovy/gradlebuild/performance/reporter/PerformanceReporter.groovy
@@ -41,32 +41,6 @@ class PerformanceReporter {
         this.fileOperations = fileOperations
     }
 
-    static class PerformanceReportExecOutputStream extends OutputStream {
-        private File profilingOutput;
-        private StringBuilder buffer = new StringBuilder()
-
-        PerformanceReportExecOutputStream(File profilingOutput) {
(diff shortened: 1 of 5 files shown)


testing/internal-performance-testing/src/main/groovy/org/gradle/performance/fixture/CrossVersionPerformanceTestRunner.groovy
```diff
@@ -217,8 +217,8 @@ class CrossVersionPerformanceTestRunner extends PerformanceTestSpec {
         def builder = GradleBuildExperimentSpec.builder()
             .projectName(testProject)
             .displayName(displayName)
-            .warmUpCount(warmUpRuns)
-            .invocationCount(runs)
+            .warmUpCount(1)
```
cobexer: ❓ Why?

blindpirate: This is for easier test, reverted.

blindpirate:
@bot-gradle squash

cobexer (approved):

PR #36989, opened 2026-03-09
Detach AS app if it's already attached

blindpirate:
Sometimes, if a build didn't clean up AS mounting properly (e.g. build is cancelled), the next build will fail with error:

```
org.gradle.tooling.internal.consumer.DefaultFailure: The directory /Volumes/AndroidStudioForGradle/Android Studio.app already exists. Please unmount it via `hdiutil detach /Volumes/AndroidStudioForGradle`.
```

Let's run `detach` automatically if it's attached.


cobexer:
@blindpirate please rebase it onto `release` as we need this on all active branches!

PR #37006, opened 2026-03-09
Add structured AI-friendly test failure output

big-guy:
## Summary

- Add a structured `failure-index.json` file written alongside HTML test reports, containing machine-parseable test failure data (exception details, filtered stack traces, stdout/stderr, source locations, re-run filters)
- Include exception messages in SHORT format console test failure output (e.g., `java.io.IOException: oops at File.java:11` instead of just `java.io.IOException at File.java:11`)
- Use `TestDescriptor.getSource()` to include source file paths, line numbers, and method/class info in the failure index

## Test plan

- [x] Unit tests for `TestFailureIndexWriter` covering JSON output, all source types (FileSource, MethodSource, ClassSource, NoSource), truncation, stacktrace filtering, JSON escaping, stdout/stderr capture, framework failures, null filter edge case, FileSource without position
- [x] Updated `ShortExceptionFormatterTest` for new message format
- [x] `sanityCheck` passes
- [x] `:testing-base:quickTest` passes

This is non-scientific, but I asked Claude to run a build with a failing test with 9.4.0 and this change and by looking at _only_ the output, guess what was broken.

9.4.0:
> The build fails because of a single test failure:
> 
>   AppTest > appHasAGreeting() FAILED — AssertionFailedError at AppTest.java:12
> 
> The test appHasAGreeting() is asserting something about the app's greeting and the assertion doesn't match. The app likely returns a value that differs from what the test expects (e.g., null or a different string).

This PR:
> Same test failure, but this version gives more detail in the error message:
> 
>   AppTest > appHasAGreeting() FAILED — "app should have a greeting ==> expected: not <null>"
> 
> The App.getGreeting() method is returning null. The test at AppTest.java:12 expects it to return a non-null value (likely using assertNotNull).

And if I allow it to look at the failure-index:
> The problem is clear from both the console output and the failure index:
> 
>   AppTest.appHasAGreeting() calls Assertions.assertNotNull() with the message "app should have a greeting", and App.getGreeting() is returning null.
> 
> The fix: App.getGreeting() needs to return an actual string instead of null.

### DIFF
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/logging/TestFailureIndexWriter.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/logging/TestFailureIndexWriter.java
@@ -163,10 +163,10 @@ public class TestFailureIndexWriter implements TestListenerInternal {
         writeFailureBlock(writer, details, descriptor.getClassName());
 
         // Output
-        writeOutputBlock(writer, entry.outputBuffer);
+        String filter = buildTestFilter(descriptor);
+        writeOutputBlock(writer, entry.outputBuffer, filter != null);
 
         // Filter for re-running
-        String filter = buildTestFilter(descriptor);
         if (filter != null) {
             writeJsonField(writer, "filter", filter, false);
         }
@@ -183,8 +183,8 @@ public class TestFailureIndexWriter implements TestListenerInternal {
             FileSource fileSource = (FileSource) source;
             writer.write("{\n");
             writeJsonField(writer, "type", "file", true, 8);
-            writeJsonField(writer, "file", fileSource.getFile().getPath(), true, 8);
             FilePosition position = fileSource.getPosition();
+            writeJsonField(writer, "file", fileSource.getFile().getPath(), position != null, 8);
             if (position != null) {
                 writer.write("        \"line\": " + position.getLine());
                 Integer column = position.getColumn();
@@ -261,7 +261,7 @@ public class TestFailureIndexWriter implements TestListenerInternal {
         writer.write("\n      },\n");
     }
 
-    private void writeOutputBlock(BufferedWriter writer, @Nullable OutputBuffer buffer) throws IOException {
+    private void writeOutputBlock(BufferedWriter writer, @Nullable OutputBuffer buffer, boolean hasMore) throws IOException {
         writer.write("      \"output\": {\n");
 
         if (buffer != null) {
@@ -289,7 +289,11 @@ public class TestFailureIndexWriter implements TestListenerInternal {
             writer.write("        \"stderrComplete\": true");
         }
 
-        writer.write("\n      },\n");
+        writer.write("\n      }");
+        if (hasMore) {
+            writer.write(",");
+        }
+        writer.write("\n");
     }
 
     static String filterStackTrace(String stacktrace, @Nullable String testClassName) {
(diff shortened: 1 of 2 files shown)


big-guy:
@eskatos @tresat don't look at the actual changes, but here's a couple of suggestions from Claude on improving the test output.

1. Including the assertion message in the default test output
2. Generating a "failure index" that contains just information about failed tests (e.g., the stacktrace and test output) so you don't need to run with --info or parse XML files

WDYT?

tresat:
> @eskatos @tresat don't look at the actual changes, but here's a couple of suggestions from Claude on improving the test output.
> 
> 1. Including the assertion message in the default test output
> 2. Generating a "failure index" that contains just information about failed tests (e.g., the stacktrace and test output) so you don't need to run with --info or parse XML files
> 
> WDYT?

My first thought is: What's the problem with having Claude get this information from the XML reports?  I have an instruction in my `agents.md` section on testing to always look there for test failures.

big-guy:
> My first thought is: What's the problem with having Claude get this information from the XML reports? I have an instruction in my `agents.md` section on testing to always look there for test failures.

The idea was to get them actionable information sooner. With the XML files, they need to...
- Determine which tests failed based on console output so it knows what to look for later
- Locate the XML file. maybe they can assume a certain location, but worse case is they need to `find` it.
- Parse the XML. Piotr talked about this being a problem when the files were too large and he had scripts to extract just the important parts.

This is a similar idea to Leonard's GAIN: https://github.com/gradle/user-leonard84-gradle-ai-native

We could potentially include extra information that only makes sense when Develocity is available (like remote execution, test history, etc).

Of course, this might be another reason to generate open test reporting vs our own bespoke thing.

tresat:
> The idea was to get them actionable information sooner.

I don't think this will make much of a difference, but we'd need to measure.  It doesn't seem to take very long for AI to read the XML, or understand where failures are occurring once it does.  I haven't noticed the problems Piotr mentioned.  The bigger problem seems to be that AI doesn't understand _that it should_ read the XML unless explicitly instructed.  

I wonder if, like how various tools suggest good `.gitignore` rules for Gradle projects, the goal should be to get various tools to suggest good `agents.md` additions for Gradle projects?  Or if perhaps we should generate them as part of `init` or post a sample file as a Best Practice?

> Add a structured failure-index.json file written alongside HTML test reports, containing machine-parseable test failure data (exception details, filtered stack traces, stdout/stderr, source locations, re-run filters)

I think failure-index.json will be only marginally more useful than the XML, which is already quite machine parseable.  Perhaps in extreme cases the difference in wordiness will matter, but just eyeballing these files you can see that the chars used for the XML "wrapping" is typically already far less than that used for the useful "content" when stack traces are invovled.  I doubt the token savings will pay off unless there are a very large number of test failures.

> Include exception messages in SHORT format console test failure output (e.g., java.io.IOException: oops at File.java:11 instead of just java.io.IOException at File.java:11)

I like the default inclusion of exception messages in console output.  We should consider how long messages look with console wrapping and clearly delimit them.  I suppose we'd need a flag to disable this as well, for anyone relying on the absence of this info (either parsing output somehow or when the messages contain secrets).

> Use TestDescriptor.getSource() to include source file paths, line numbers, and method/class info in the failure index

Isn't this already present in the stack trace?  I suppose you don't have an actual file path, you have a FQN for a class, which the AI then needs to locate in the project, so this is an improvement.  I haven't noticed located classes ever being a particular problem though.

Other Thoughts
- failure-index.json isn't a report, isn't configured like other reports, and shouldn't live as a sibling to the HTML reports.  It should be located somewhere else.
- failure-index isn't an index, it's the complete contents.  Should it be structured as an index file, that only points to individual test result .json files in a directory, so that AI readers can limit context usage by only reading info about tests they care about?
- If we have the info stored in the index about what tests failed, I would love an invocation like `gradlew test --retry` that automatically reruns all the failures and only the failures.
- Machine-readable build scans would be even better.  Adding extra historical info from DV seem like it wouldn't be useful in terms of fixing test failures 95+% of the time.

PR #37008, opened 2026-03-09
Add project(String) method to DependencyHandler

ghale:
This adds a new `project(String)` method to DependencyHandler to allow adding project dependencies without using actual `Project` objects.  This clears the way for us to deprecated passing a `Project` object as a dependency notation in 9.6.

This also simplifies the dependency factory infrastructure by pushing `ProjectFinder` down to the only place it's actually used (`DefaultProjectDependencyFactory`) so that we aren't passing it around through a bunch of intermediate method calls.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependenciesIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependenciesIntegrationTest.groovy
@@ -131,6 +131,26 @@ class ProjectDependenciesIntegrationTest extends AbstractDependencyResolutionTes
         outputContains 'project :sub'
     }
 
+    def "can declare project dependency on root project using project path string"() {
+        settingsFile << """
+            rootProject.name = 'root'
+            include 'sub'
+        """
+        file("sub/build.gradle") << """
+            apply plugin: 'java-library'
+            dependencies {
+                implementation project(':')
+            }
+        """
+        buildFile << """
+            apply plugin: 'java-library'
+        """
+
+        expect:
+        succeeds 'sub:dependencies', '--configuration', 'compileClasspath'
+        outputContains 'root project :'
+    }
+
     def "can add constraint on root project"() {
         given:
         mavenRepo.module("org", "foo").publish()


ghale:
@bot-gradle test this

platforms/jvm/scala/src/integTest/groovy/org/gradle/scala/ScalaPluginIntegrationTest.groovy
```diff
@@ -105,7 +105,7 @@ class ScalaPluginIntegrationTest extends MultiVersionIntegrationSpec {
                 apply plugin: 'scala'
                 dependencies {
                     implementation("${scalaDependency(version.toString())}")
-                    implementation(project(":java").sourceSets.main.output)
+                    implementation(project.project(":java").sourceSets.main.output)
```
tresat, line 108: 💅 `project.project`?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependenciesIntegrationTest.groovy
```diff
@@ -114,6 +114,23 @@ class ProjectDependenciesIntegrationTest extends AbstractDependencyResolutionTes
         failure.assertHasCause("Project with path ':unknown' could not be found.")
     }
 
+    def "can declare project dependency using project path string"() {
```
tresat, line 117: 🤔 It might be good to have a test for `project(":")` to demo the behavior.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/BuildscriptResolutionIntegrationTest.groovy
```diff
@@ -45,7 +45,7 @@ class BuildscriptResolutionIntegrationTest extends AbstractIntegrationSpec {
                 }
 
                 dependencies {
-                    classpath project(":")
+                    classpath project.project(":")
```
tresat, line 48: 💅 `project.project`?

platforms/jvm/scala/src/integTest/groovy/org/gradle/scala/ScalaPluginIntegrationTest.groovy
ghale: This is intentional.  In this case, we're not creating a dependency on a project, we're creating a dependency on a sourceset.  So, using the implicit delegation calls `DependencyHandler.project()` which doesn't get us what we want - we have to explicitly call `Project.project()`.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/BuildscriptResolutionIntegrationTest.groovy
ghale: Also intentional.  In this case, we're testing that declaring project dependencies in a buildscript fail to resolve.  Since this is in the buildscript, you get an error when you call `DependencyHandler.project()` (i.e. you get it immediately during declaration) so we use `Project.project()` to get the dependency declared and then verify that the resolution itself fails.

They're basically edge case tests only there to pin down behavior.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependenciesIntegrationTest.groovy
ghale: Done.

PR #37018, opened 2026-03-10
Update Kotlin embedded compiler docs to reflect supported architectures

### LINKED ISSUE #37017: Update list of supported architectures of the Kotlin embedded compiler
### Issue type

### Problem description

We claim that the Kotlin embedded compiler only works on the x86_64 architectures: https://github.com/gradle/gradle/blob/c21b5da6c5afab06a792763a7eedc83c21851400/platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/kotlin_dsl.adoc?plain=1#L32

### Context (optional)

Page was last edited 8 years ago, probably at that time, only x86_64 was supported.

### Page with the problem

https://docs.gradle.org/current/userguide/kotlin_dsl.html

VedantMadane:
Fixes #37017

Updates the Kotlin DSL documentation to reflect that the embedded Kotlin compiler now supports both x86-64 and aarch64 (ARM64) architectures, not just x86-64 as previously documented.

The documentation was last edited 8 years ago when only x86-64 was supported. The Kotlin compiler embeddable and Gradle now support aarch64 on Linux, macOS (Apple Silicon), and Windows ARM.

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/kotlin_dsl.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/kotlin_dsl.adoc
@@ -29,7 +29,7 @@ TIP: If you are interested in migrating an existing Gradle build to the Kotlin D
 [[kotdsl:prerequisites]]
 == Prerequisites
 
-* The embedded Kotlin compiler works on Linux, macOS, Windows, Cygwin, FreeBSD, and Solaris on x86-64 architectures.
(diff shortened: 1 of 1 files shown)


eskatos:
Thanks for the contribution.
As noted in the issue in https://github.com/gradle/gradle/issues/37017#issuecomment-4030254993, we would prefer not to have the supported platforms listed in our documentation but instead link to some official Kotlin material. That way our documentation will not rot in the future.
Could you please adjust?

ov7a:
Also, the [DCO](https://probot.github.io/apps/dco/) check is failing, please look into that as we can't accept the PR otherwise.

PR #37027, opened 2026-03-10
Migrate DefaultBuildEventsListenerRegistry to MultiProducerSingleConsumerProcessor

jvandort:
Previously we relied on a LinkedTransferQueue and a hand-written event processing thread. MPSCP implements this behind a reusable abstraction.
    
Along the way, we determined that our existing MPSCP is not as performant as LTQ. We add strategic spin waits and yields to the worker loop, and now we beat LTQ's performance. For reference, LTQ performs a single yield in its 'take' implementation. Updating our MPSCP to use a single yield produces slightly better performanec than LTQ, as it can avoid linking and allocation overhead of LTQ. Increasing to 5 yields improves performance by ~35%.
    
In a series of local benchmarks executed against gradle/gradle, different approaches producer threads spent varying amount of times submitting work to the queue:
- existing LTQ: 3.3k samples
- MPSCP without optimization: 5.3k samples
- MPSCP with optimizations: 2.1k samples

MPSCP without optimization:
image

LTQ: 
image

MPSCP with single yield:
image

MPSCP with spin loop & multiple yields:
image

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
+++ b/platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
@@ -131,6 +131,32 @@ public class MultiProducerSingleConsumerProcessor<T> {
             }
         }
 
+        doSubmit(value);
+    }
+
+    /**
+     * Submit a value to be processed, discarding the value if the processor is
+     * shutdown or in a failure state.
+     */
+    public void maybeSubmit(T value) {
+        if (failure != null) {
+            return;
+        }
+
+        if (!running) {
+            if (worker.getState() == Thread.State.NEW) {
+                throw new IllegalStateException("Cannot submit values before processor has been started.");
+            }
+            return;
+        }
+
+        doSubmit(value);
+    }
+
+    /**
+     * Submit the value to the queue, waking up the worker thread if necessary.
+     */
+    private void doSubmit(T value) {
         if (!queue.offer(value)) {
             throw new IllegalStateException("Failed to offer value to queue");
         }
--- a/platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
+++ b/platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
@@ -16,6 +16,7 @@
 
 package org.gradle.internal.concurrent;
 
+import org.gradle.api.JavaVersion;
 import org.jctools.queues.MessagePassingQueue;
 import org.jctools.queues.MpscUnboundedArrayQueue;
 import org.jspecify.annotations.Nullable;
@@ -45,6 +46,23 @@ public class MultiProducerSingleConsumerProcessor<T> {
      * before checking failure status or thread interruption.
      */
     static final int BATCH_SIZE = 1024;
+
+    /**
+     * Allows us to use {@link Thread#onSpinWait()}, while remaining compatible with JDK8.
+     */
+    private static final boolean AT_LEAST_JDK9 = JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_1_9);
+
+    /**
+     * Number of hardware spins before we yield to the OS scheduler, allowing us to
+     * avoid the overhead of a syscall before waiting for more values.
+     */
+    private static final int SPIN_TRIES = 100;
+
+    /**
+     * Number of yields to execute before parking, allowing us to keep the {@link #awake}
+     * flag true for as long as possible, preventing producers from needing to unpark.
+     */
+    private static final int YIELD_TRIES = 5;
 
     /**
      * Processes submitted values on a separate thread.
@@ -197,7 +215,11 @@ public class MultiProducerSingleConsumerProcessor<T> {
         }
     }
 
+    @SuppressWarnings("ThreadPriorityCheck")
     private void workerLoop() {
+        int idleSpins = 0;
+        int idleYields = 0;
+
         try {
             while (running || !queue.isEmpty()) {
                 if (Thread.interrupted()) {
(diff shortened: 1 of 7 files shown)


jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
```diff
@@ -46,6 +47,23 @@ public class MultiProducerSingleConsumerProcessor<T> {
      */
     static final int BATCH_SIZE = 1024;
 
+    /**
+     * Allows us to use {@link Thread#onSpinWait()}, while remaining compatible with JDK8.
+     */
+    private static final boolean AT_LEAST_JDK9 = JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_1_9);
```
reinsch82, line 53: 💅 maybe
```suggestion
    private static final boolean AT_LEAST_JDK9 = JavaVersion.current().isJava9Compatible();
```

subprojects/build-events/src/main/java/org/gradle/internal/build/event/DefaultBuildEventsListenerRegistry.java
```diff
@@ -195,66 +186,33 @@ private UserCodeSource getCurrentContext() {
...
-        private final TransferQueue<Object> events = new LinkedTransferQueue<>();
-        private final AtomicReference<Throwable> failure = new AtomicReference<>();
+        private final MultiProducerSingleConsumerProcessor<T> processor;
 
-        public AbstractListener(@Nullable UserCodeSource registrationPoint, ExecutorFactory executorFactory) {
+        public AbstractListener(@Nullable UserCodeSource registrationPoint) {
             this.registrationPoint = registrationPoint;
-            this.executor = executorFactory.create("build event listener");
-            Future<?> ignored = executor.submit(this::run);
+            this.processor =  new MultiProducerSingleConsumerProcessor<>("build event listener", this::handle);
```
reinsch82, line 196: 💅 
```suggestion
            this.processor = new MultiProducerSingleConsumerProcessor<>("build event listener", this::handle);
```

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
```diff
@@ -184,11 +232,45 @@ private void workerLoop() {
...
+                    }
+                    continue;
+                }
+
+                // Yield the worker thread to the CPU, queueing us up for later execution, allowing
+                // us to keep the `awake` flag enabled. This reduces the number of times producer
+                // threads need to unpark this thread.
+                if (idleYields < YIELD_TRIES) {
+                    idleYields++;
+                    Thread.yield();
```
octylFractal, line 268: ❓ I don't think this makes sense. Have you benchmarked with/without this, on all OSes? At least on Linux, this calls out to `sched_yield`, which is still a context switch into the kernel and surely somewhere as bad as `park`, right? Additionally, it can yield for longer than expected, and might actually take longer than just doing the context switch.

I wouldn't assume that just since `LinkedTransferQueue` uses it, we also should. It was probably written before `onSpinWait` existed.

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
```diff
@@ -184,11 +232,45 @@ private void workerLoop() {
...
+                    if (AT_LEAST_JDK9) {
+                        Thread.onSpinWait();
+                    } else {
+                        Thread.yield();
+                    }
+                    continue;
+                }
+
+                // A quick spin wait buffer, before we proceed with the overhead of yielding.
+                if (idleSpins < SPIN_TRIES) {
```
octylFractal, line 255: ❓ Instead of implementing this ourselves, would it be worth trying to see if the existing `MpscBlockingConsumerArrayQueue` is as good or better? It unfortunately doesn't use `onSpinWait`  at all, probably due to compatibility issues, but it might still be a win? Especially if we don't need to maintain all this logic ourselves and could contribute minor improvements to JCTools instead.

octylFractal (commented):
I am OK with this, but I would prefer to see the MPSC class benchmarked with JMH to allow verifying the performance is as expected on all platforms / architectures.

I would really like to at least see Linux profiles as well to ensure  it's working as expected there.

reinsch82 (commented):
just nitpicking here.

thanks for taking this on.

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
jvandort: The advantage of yield over park is we can delay parking, meaning we can delay setting awake=false. This increases the chances of a producer thread seeing an awake consumer and not needing to call `unpark` on it. 

The point here is to alleviate load on the producer threads, as they are less likely to need to syscall when they are submitting work to this async executor.

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
jvandort: Definitely worth comparing `MpscBlockingConsumerArrayQueue` to this PR

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/MultiProducerSingleConsumerProcessor.java
jvandort: For the record this does show improvements in a number of performance tests on linux, though a JMH might help us optimize the number of spin waits and yields. 

> which is still a context switch into the kernel and surely somewhere as bad as park, right

The point here is we're not quite as worried about syscalls on this consumer thread as we are on the producer threads -- though it would still make sense to optimize the number of times we call yield. 

I did profile this at the time and I recall the number of samples spent inside the consumer spin waiting and yielding was less than the number of samples producers spent unparking the consuming thread. So, at least on my machine, this should reduce total load

Issue #37021, opened 2026-03-10
Memoize pluginEntriesFrom to avoid redundant JAR scanning and excessive byte[] allocations


big-guy:
## Summary

`pluginEntriesFrom` opens and reads JAR files to extract plugin metadata. This method is called multiple times with the same classpath during a single build, resulting in redundant I/O and allocations.

## Profiling Data

Profiling `gradle help` on the gradle/gradle build with JFR shows:

- **~15 GiB total allocations**, of which **~5 GiB are `byte[]`**
- **~10% of all `byte[]` allocations** (~500 MiB) originate from `pluginEntriesFrom` in `GeneratePluginSpecBuilderAccessorsKt`
- The method is called ~3 times with the same classpath, re-opening the same JAR files each time

I think this single call site accounts for ~20% of all allocations when considering the full call tree (JAR reading, properties parsing, etc.).

## Root Cause

There are multiple call sites that invoke `pluginEntriesFrom` / `pluginTreesFrom` on the same classpath with no caching.

## Impact

This would reduce allocations during by ~500 MiB–1 GiB (avoiding ~2 redundant full classpath scans)

spider-yamet:
@big-guy May I resolve this issue? I have some experience on this area, Hoppe this project is open for contribution.
Looking forward to discuss implementation details.

Regards

big-guy:
@spider-yamet Sorry, I opened this after talking with @eskatos - he already had a local fix. 

There are lots of other issues to fix :)

PR #37030, opened 2026-03-10
Cache plugin entries extracted from JAR files

eskatos:
`pluginEntriesFrom` opens and reads JAR files to extract plugin metadata. This method is called multiple times with the same classpath during a single build, resulting in redundant I/O and allocations.

Introduce a `KotlinDslPluginEntryCache` to avoid redundantly scanning JAR files and loading plugin descriptor properties files.

Build invocations re-generating plugin spec builders accessors are doing too much work and allocations. Keeping a cross-build in-memory cache would be holding up memory unnecessarily for all the build invocations that don't regenerate these accessors.

Register `KotlinDslPluginEntryCache` as a build-tree scoped service holding a persistent cache decorated with an in-memory cache.

The in-memory decorator cache keeps a maximum of arbitrarily 1000 entries to avoid redundant processing of the same JARs at the build-tree scope.

In an extreme scenario with 1000 different JARs processed at the build-tree scope, with each 10 plugin descriptors with very long plugin IDs and implementation class names (10000 plugins total), the in-memory cache size could be ~3.5MB (~48B per key, ~3.6KB per value). A more reasonable scenario with 200 different JARs and 200 plugins total would hold ~40KB of heap.

The persistent cross-build cache allows build invocations which trigger generating plugin spec builder accessors to quickly read the plugin entries from the cache without scanning the JARs. The underlying persistent store is cleaned up daily with 7 days LRU strategy.

Thread the cache service through the different call sites: plugin spec builder accessor generation, precompiled script plugin support, and Gradle API build-time extensions generation.

Because of the latter, the `pluginEntriesFrom` function is shared with the gradle/gradle build logic. The build logic using `pluginEntriesFrom` has its own `PluginEntryCache` implementation that only caches in-memory over a short-lived task action using a simple map.

* Fixes https://github.com/gradle/gradle/issues/37021

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic/kotlin-dsl-shared-runtime/src/main/kotlin/org/gradle/kotlin/dsl/internal/sharedruntime/codegen/PluginEntry.kt
+++ b/build-logic/kotlin-dsl-shared-runtime/src/main/kotlin/org/gradle/kotlin/dsl/internal/sharedruntime/codegen/PluginEntry.kt
@@ -23,9 +23,17 @@ import java.util.jar.JarEntry
 import java.util.jar.JarFile
 
 
+interface PluginEntryCache {
+    fun computeIfAbsent(jar: File, producer: (File) -> List<PluginEntry>): List<PluginEntry>
+}
+
 data class PluginEntry(val pluginId: String, val implementationClass: String)
 
 
+fun pluginEntriesFrom(jar: File, pluginEntryCache: PluginEntryCache): List<PluginEntry> =
+    pluginEntryCache.computeIfAbsent(jar, ::pluginEntriesFrom)
+
+private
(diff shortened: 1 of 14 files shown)


eskatos:
@bot-gradle test APT

eskatos:
@bot-gradle test APT

eskatos:
On the `gradle/gradle` build, this cut the related byte[] allocations in half.
The performance tests are all passing with few percent improvements for some scenarios.

This still needs some polish.

PR #37032, opened 2026-03-11
Use Kotlin's Build Tools API for compiling Kotlin scripts

### LINKED ISSUE #34181: Investigate using Kotlin Build Tools API for Kotlin DSL script compilation
Instead of the embedded compiler scripting API.

----

We would need to replace code inside [compileKotlinScriptModuleTo](https://github.com/gradle/gradle/blob/master/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt#L207) with [CompilationService.compileJvm](https://github.com/JetBrains/kotlin/blob/master/compiler/build-tools/kotlin-build-tools-api/src/main/kotlin/org/jetbrains/kotlin/buildtools/api/CompilationService.kt#L82) using [in-process strategy](https://github.com/JetBrains/kotlin/blob/5328f40682ce5eb40b449fa658d0537554ee332b/compiler/build-tools/kotlin-build-tools-api/src/main/kotlin/org/jetbrains/kotlin/buildtools/api/CompilerExecutionStrategyConfiguration.kt#L21) and [classpath based incremental compilation strategy](https://github.com/JetBrains/kotlin/blob/5328f40682ce5eb40b449fa658d0537554ee332b/compiler/build-tools/kotlin-build-tools-api/src/main/kotlin/org/jetbrains/kotlin/buildtools/api/jvm/JvmCompilationConfiguration.kt#L54).

<details>
<summary>Notes for matching embedded API to BTA arguments</summary>

```diff
diff --git a/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt b/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
index be6b96479ab..529ba1aa2f7 100644
--- a/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
@@ -216,11 +216,11 @@ fun compileKotlinScriptModuleTo(
     withRootDisposable {
         withCompilationExceptionHandler(messageCollector) {
             val configuration = compilerConfigurationFor(messageCollector, compilerOptions).apply {
-                put(OUTPUT_DIRECTORY, outputDirectory)
-                setModuleName(moduleName)
-                addScriptingCompilerComponents()
-                add(SamWithReceiverConfigurationKeys.ANNOTATION, HasImplicitReceiver::class.qualifiedName!!)
-                add(AssignmentConfigurationKeys.ANNOTATION, SupportsKotlinAssignmentOverloading::class.qualifiedName!!)
+                put(OUTPUT_DIRECTORY, outputDirectory) // -d
+                setModuleName(moduleName) // -module-name
+                addScriptingCompilerComponents() // -Xplugin pass jars to the required compiler plugins
+                add(SamWithReceiverConfigurationKeys.ANNOTATION, HasImplicitReceiver::class.qualifiedName!!) // -P plugin:org.jetbrains.kotlin.samWithReceiver:annotation=<value>
+                add(AssignmentConfigurationKeys.ANNOTATION, SupportsKotlinAssignmentOverloading::class.qualifiedName!!) // -P plugin:org.jetbrains.kotlin.assignment:annotation=<value>
             }
 
             val environment = kotlinCoreEnvironmentFor(configuration)
@@ -411,15 +411,15 @@ class LoggingOutputStream(val log: (String) -> Unit) : OutputStream() {
 private
 fun compilerConfigurationFor(messageCollector: MessageCollector, compilerOptions: KotlinCompilerOptions): CompilerConfiguration =
     CompilerConfiguration().apply {
-        put(CommonConfigurationKeys.MESSAGE_COLLECTOR_KEY, messageCollector)
-        put(CommonConfigurationKeys.USE_FIR, true) // Enables K2
-        put(RETAIN_OUTPUT_IN_MEMORY, false)
-        put(JVM_TARGET, compilerOptions.jvmTarget.toKotlinJvmTarget())
-        put(JDK_HOME, File(System.getProperty("java.home")))
-        put(SAM_CONVERSIONS, JvmClosureGenerationScheme.CLASS)
-        addJvmSdkRoots(PathUtil.getJdkClassesRootsFromCurrentJre())
-        put(CommonConfigurationKeys.LANGUAGE_VERSION_SETTINGS, gradleKotlinDslLanguageVersionSettingsFor(compilerOptions))
-        put(CommonConfigurationKeys.ALLOW_ANY_SCRIPTS_IN_SOURCE_ROOTS, true)
+        put(CommonConfigurationKeys.MESSAGE_COLLECTOR_KEY, messageCollector) // JvmCompilationConfiguration.useLogger
+        put(CommonConfigurationKeys.USE_FIR, true) // Enables K2 // Should be enabled by LV 2.0+
+        put(RETAIN_OUTPUT_IN_MEMORY, false) // not actual
+        put(JVM_TARGET, compilerOptions.jvmTarget.toKotlinJvmTarget()) // -jvm-target
+        put(JDK_HOME, File(System.getProperty("java.home"))) // -jdk-home
+        put(SAM_CONVERSIONS, JvmClosureGenerationScheme.CLASS) // -Xsam-conversions=class
+        addJvmSdkRoots(PathUtil.getJdkClassesRootsFromCurrentJre()) // Possibly not needed?
+        put(CommonConfigurationKeys.LANGUAGE_VERSION_SETTINGS, gradleKotlinDslLanguageVersionSettingsFor(compilerOptions)) // -language-version
+        put(CommonConfigurationKeys.ALLOW_ANY_SCRIPTS_IN_SOURCE_ROOTS, true) // -Xallow-any-scripts-in-source-roots - possibly not needed?
     }
```

</details>

An interesting side-effect of this could be the following - BTA exposes [finishProjectCompilation](https://github.com/JetBrains/kotlin/blob/5328f40682ce5eb40b449fa658d0537554ee332b/compiler/build-tools/kotlin-build-tools-api/src/main/kotlin/org/jetbrains/kotlin/buildtools/api/CompilationService.kt#L97) method which drops compiler caches for classpath jars. That means build scripts classpath caches should be reused during non-incremental first compilation and it should speedup overall project scripts compilation as build script classpath is mostly the same.

Another possible side effect - we may be able to "easily" swap Kotlin compiler version by changing just BTA-impl version. Kotlin folks are trying to support one major forward and three major versions back for BTA-API <-> BTA-impl compatibility.

### LINKED ISSUE #34952: Resolve `K1Deprecation` opt-ins before Kotlin 2.3.0
### Issue type

Polishing

### Problem description

* We added some `@OptIn(K1Deprecation::class)` in Kotlin DSL script compilation usage of the embedded Kotlin compiler API
* These will need to be resolved before we can upgrade the embedded Kotlin to 2.3.0

### Context (optional)

* The opt-ins were added in https://github.com/gradle/gradle/pull/34902 for upgrading to Kotlin 2.2.20

jbartok:
Fixes:
* https://github.com/gradle/gradle/issues/34181

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


bamboo:
cc @jvandort

jbartok:
@bot-gradle test APT

jbartok:
@bot-gradle test APT

jbartok:
@bot-gradle test APT

jbartok:
@bot-gradle test this APT

jbartok:
@bot-gradle test APT

jbartok:
@bot-gradle test this

jbartok:
@bot-gradle test this APT

jbartok:
Accidental Git fuckery, reopening.

jbartok:
@bot-gradle test this

jbartok:
@bot-gradle test this

jbartok:
@bot-gradle test this APT AFT AST

jbartok:
Overview for review-ers:

The Build Tools API (https://kotlinlang.org/docs/build-tools-api.html) provides a unified API between build systems and the Kotlin compiler ecosystem. This PR makes use of it to replace our legacy system for compiling Kotlin build scripts. 

The changes require an embedded Kotlin version of `2.4.0` or later, which we already have on `master`.

When using BTA, there are multiple ways for setting it up:
* in-process compilation vs. external compiler daemon (like KGP uses)
* sharing a long-lived session between compile operations vs. using a fresh one for each operation
* using an isolated class-loader for loading BTA vs. just relying on caller's (see "Follow-up" section)
* setting the "kotlin.environment.keepalive" property vs. not (can be used to trigger the reuse of long-lived application environment objects in the Kotlin compiler)

When initially switching from our legacy way of compiling build scripts to BTA all combination of the above factors turned out to be slower than what we had... The good thing is that once we are on BTA we can benefit from any improvements the Kotlin folks make simply by upgrading the embedded Kotlin version.

But, then, "why is this PR still open", you might ask, quite reasonably. Well, turns out, that after switching to BTA it becomes possible to turn on Incremental Compilation relatively simply and that yields huge improvements when scripts have ABI changes and need to be recompiled. Something like 60% improvement in our performance tests. IC has the disadvantage of having a costly initial setup, but we have found a strategy mitigating that too: 
* on the first compilation of a build script we don't turn IC on at all (this should shield people not touching their build scripts from the IC overhead cost)
* IF there is a second recompilation (i.e. the build script is being modified), then we turn IC on and pay for its setup cost
* WHEN there are more recompilations (i.e. the build script is being modified more), then we start enjoying the performance benefits IC brings

This strategy can be disputed, of course, but with it, our performance tests almost pass..., meaning that there are some issues with first use ones (varies greatly between local and TC tests), but the ABI change recompilation case is very fast!

Besides this IC strategy, the combination of other factors the PR settled on (based on the best performance results is):
* in-process mode (so no compile daemon - which is not something we probably want anyway, because sharing the KGP one is practically impossible, according to the Kotlin folks, so we would end up always having two)
* no "kotlin.environment.keepalive" flag
* no isolated classloader
* long-lived session

The biggest challenge in making IC work were having the right caches, making them efficient and working out their cleanup strategies. Most important part to review. There are two new caches involved:
* classpath-entry cache (see `KotlinDslClasspathEntrySnapshotStore`)
  * keyed by script compilation classpath entries (jars or class file directories, represented by their content's hashes)
  * part of the values are ABI hashes of the classpath entries, used by compilation avoidance (see `KotlinCompileClasspathFingerprinter`), something we had before this PR too and which remains in place
  * the extra part of the values are BTA snapshots used for IC; BTA compares the classpath's snapshots against those it recorded on the previous compile to find which dependency classes changed, and recompiles only what's affected
  * both types of values in this cache are computed in the same processing step now
* script IC cache (see `KotlinDslIncrementalCompilationCache`)
  * keyed by script identity hash 
  * contains multiple values for each script:
    * IC state - BTA's IC working state (source hashes, who-references-what, dirty tracking)
    * compilation outputs - the stable destination the compiler writes class files into
    * sources - stable script-text files handed to the compiler

IC is best-effort and can never break a build: any IC failure (an unreadable or reclaimed classpath snapshot, or corrupt IC state) is caught and retried as a full compile, and a compilation killed mid-emit is detected on the next build (via a marker) and rebuilt from scratch.

One factor that complicates things in particular is how Kotlin script compilation uses immutable workspaces in the execution engine and needs compilation results to be put in a new workspace specific folder whenever compilation inputs changes. IC is the opposite, it needs a stable, mutable directory, so some gymnastics is required to reconcile the two (see `KotlinCompiler.copyOutputs()`).

Follow-up: 
 * using an isolated class-loader for calling BTA would allow compiling build scripts with a Kotlin version different than the embedded one (as long as compatible), a feature desired by the Kotlin folks, but out-of-scope for this PR

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/cache/KotlinDslIncrementalCompilationCache.kt
```diff
@@ -0,0 +1,167 @@
...
+import org.gradle.internal.service.scopes.Scope
+import org.gradle.internal.service.scopes.ServiceScope
+import java.nio.file.Files
+import java.nio.file.Path
+import kotlin.io.path.createDirectories
+
+
+/**
+ * On-disk cache of per-script incremental-compilation state, backed by a
+ * [FineGrainedPersistentCache] under `<gradleUserHome>/caches/<gradleVersion>/kotlin-dsl-ic/` and
```
eskatos: ❓ I wonder about the paths of the different caches introduced in this PR.

Before this PR we have `<gradleUserHome>/caches/<gradleVersion>/kotlin-dsl/` that contains two subfolders: `accessors` and `scripts`. It makes it very easy to wipe all the Kotlin DSL related caches and doesn't clutter the `<gradleVersion>` directory with many entries.

Can't we put all the new caches under `<gradleUserHome>/caches/<gradleVersion>/kotlin-dsl/`?

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/execution/Interpreter.kt
```diff
@@ -73,7 +77,15 @@ import java.nio.file.Path
  * @see ResidualProgramCompiler
  */
 internal
-class Interpreter(val host: Host, val buildOperationRunner: BuildOperationRunner) {
+class Interpreter(
```
eskatos: ❌ Let's keep the `Interpreter` constructor with a minimal set of fields and move the services to the `Host`.

platforms/core-configuration/kotlin-dsl/build.gradle.kts
```diff
@@ -38,12 +38,14 @@ dependencies {
 
     api(libs.groovy)
     api(libs.guava)
+    api(libs.kotlinBuildToolsApi)
```
eskatos: ❓ Can we limit the exposure of the BTA types to `internal` so we can have this dependency in the `implementation` scope instead?

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/cache/KotlinDslIncrementalCompilationCache.kt
jbartok: Makes sense, done.

platforms/core-configuration/kotlin-dsl/build.gradle.kts
jbartok: Makes sense, culprit(s) found, done.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/execution/Interpreter.kt
jbartok: Done.

jbartok:
@bot-gradle test this APT

jbartok:
@bot-gradle test this

jbartok:
@bot-gradle test APT

jbartok:
@bot-gradle test this APT

jbartok:
@bot-gradle test this APT

jbartok:
@bot-gradle test this AFT AST

jbartok:
Will be wrapped up via https://github.com/gradle/gradle/pull/38593

Issue #8755, opened 2019-03-12
Define behaviour for adding a ConfigurableFileCollection to itself


ljacomet:
Currently, if one adds a `ConfigurableFileCollection` to itself, directly or indirectly, this creates a `StackOverflowError` when resolving the file collection, as can be seen in #8748.

Gradle should detect such situations and either fail with a clear exception indicating the problem or ignore that element.

This probably applies to `FileCollection` as well.

adammurdoch:
The behaviour should also make sense for setting the value of a lazy type using something derived from itself. Some examples:

- adding a `DomainObjectCollection` to itself
- adding a `Property<List<T>>` to itself
- setting the value of a `Property<T>` to a `Provider<T>` derived from property value
- adding a `TaskDependency` to itself.

adammurdoch:
This is fixed for a common case where += is used from the Groovy DSL: https://github.com/gradle/gradle/pull/12842

jjohannes:
We would like to restore this change in the `JavaExec` task: https://github.com/gradle/gradle/commit/22ca8e2e3714209e9000b4825b349c2ba246ad47

For that, we have to make at lease this test pass (on top of what was done in #12842):
https://github.com/gradle/gradle/commit/22ca8e2e3714209e9000b4825b349c2ba246ad47#diff-f4ec7d2e34e668d08bf7369bcf62f9faR177

We should also add test coverage for other ways of composing a file collection with 'cycles'.

mlopatkin:
Here is the summary of the current behaviors of the lazy types.

## ConfigurableFileCollection
Can be added to itself, a derived file collection can be added to itself. It works by introspecting the added value and replacing the reference to the collection with the snapshot before addition (aka shallow copy).

Cannot contain a provider derived from itself (circular evaluation of the provider):
```gradle
def col = files("build.gradle")
def provider = col.elements.map { file(it.name + ".kts") }
col.from(col)
```

## Properties and Providers
These do not support self-references. The only exceptions are providers returned by `NamedDomainObjectContainer`, in particular, `TaskProvider`. These can resolve themselves in their configuration callbacks, though this exposes partially configured Task instance.

## DomainObjectCollection
[...]

PR #37034, opened 2026-03-11
Add reproducer for `SOE` when `cfc1.setFrom(cfc1-cfc2)`

abstratt:
Add reproducer for SOE when cfc1.setFrom(cfc1-cfc2)

Issue:  #8755


abstratt:
@bot-gradle squash and merge

Issue #35808, opened 2025-11-25
Allow users to specify an explicit bind address to be used for client-daemon communication


abstratt:
### Expected Behavior

Most tools and systems out there allow users to explicit specify what network address to bind to. Gradle should do so as well.

Gradle attempts to automatically figure out what local binding address to use for local (IPC) communication. That does not work correctly in some environments/setups. 

Actually, Gradle already sort of supports this: back in the day, [ad hoc support](https://docs.gradle.org/2.2/release-notes.html#compatibility-with-openshift) was added for Openshift environments via detection of `OPENSHIFT(..)_IP` environment variables. Other than undocumented and poorly named, that feature works very well, and non-Openshift users have been using that obscure feature as a workaround. 

This is a request to at a minimum extend the existing feature to use a more suitably named environment variable such as `GRADLE_BIND_ADDRESS` or `GRADLE_DAEMON_BIND_ADDRESS`, and to document that feature appropriately.

### Current Behavior (optional)

Gradle may choose an incorrect address to bind to, preventing IPC between launcher and daemon from working. 

Many users report issues related to that:

https://github.com/gradle/gradle/issues/35711 (which will be marked as a duplicate of the current one) 
https://github.com/gradle/gradle/issues/12731#issuecomment-756529989
https://github.com/gradle/gradle/issues/12232#issuecomment-895087251
https://github.com/gradle/gradle/issues/14528

### Context

Workaround is to use an undocumented feature built for a specific environment most users never heard of.

DevonianTeuchter:
+1  This issue is especially prevalent on z/OS where multiple TCP/IP stacks are available, IPv4 and IPv6 may or may not be configured, bind addresses might resolve to 0.0.0.0 and network adapter enumeration by the JVM changes between Java levels.  Currently, only specifying an OPENSHIFT_*_IP variable provides a reliable way to ensure consistency by hard-coding the listening address.

DevonianTeuchter:
Q: To progress this, would a PR that does either of the following be of interest:
1. Renames the existing variable to GRADLE_BIND_ADDRESS  
2. Adds a variable GRADLE_BIND_ADDRESS that functions as the legacy OPENSHIFT_*_IP variable and deprecate the OPENSHIFT_*_IP variable

Documentation for either option would need to be added.
The alternative might be to look at reworking the logic to be consistent on z/OS though this is core functionality and the change would only be of benefit to a single platform, with limited (current) userbase.

[...]

PR #37043, opened 2026-03-11
Support GRADLE_DAEMON_BIND_ADDRESS env variable to support customizing daemon local binding ip

asodja:
Fixes https://github.com/gradle/gradle/issues/35808

### DIFF
--- a/platforms/core-runtime/messaging/src/main/java/org/gradle/internal/remote/internal/inet/InetAddressFactory.java
+++ b/platforms/core-runtime/messaging/src/main/java/org/gradle/internal/remote/internal/inet/InetAddressFactory.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
 
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.net.SocketException;
 import java.net.UnknownHostException;
 import java.util.Set;
 
@@ -91,9 +92,6 @@ public class InetAddressFactory {
         wildcardBindingAddress = new InetSocketAddress(0).getAddress();
 
         if (!findGradleDaemonBindAddress() && !findOpenshiftAddress()) {
-            if (inetAddresses == null) { // For testing
-                inetAddresses = new InetAddresses();
-            }
             findLocalBindingAddress();
         }
     }
@@ -101,7 +99,10 @@ public class InetAddressFactory {
     /**
      * Prefer first loopback address if available, otherwise use the wildcard address.
      */
-    private void findLocalBindingAddress() {
+    private void findLocalBindingAddress() throws SocketException {
+        if (inetAddresses == null) { // For testing
+            inetAddresses = new InetAddresses();
+        }
         if (inetAddresses.getLoopback().isEmpty()) {
             logger.debug("No loopback address for local binding, using fallback {}", wildcardBindingAddress);
             localBindingAddress = wildcardBindingAddress;
--- a/platforms/core-runtime/messaging/src/test/groovy/org/gradle/internal/remote/internal/inet/InetAddressFactoryTest.groovy
+++ b/platforms/core-runtime/messaging/src/test/groovy/org/gradle/internal/remote/internal/inet/InetAddressFactoryTest.groovy
@@ -109,7 +109,7 @@ class InetAddressFactoryTest extends Specification {
         0 * addresses._
     }
 
-    def "GRADLE_DAEMON_BIND_ADDRESS skips auto-detection regardless of OPENSHIFT address"() {
+    def "GRADLE_DAEMON_BIND_ADDRESS takes precedence over OPENSHIFT address"() {
         when:
         environment["OPENSHIFT_FOO_IP"] = "10.0.0.1"
         environment["GRADLE_DAEMON_BIND_ADDRESS"] = "192.168.1.10"


asodja:
@bot-gradle test this

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -113,6 +113,21 @@ myCollection.add(MyType("other")) // this will fail
 myCollection.remove(main)         // this will fail
 ```
 
+### Explicit bind address for client-daemon and cross-daemon communication
+
+Gradle now supports the `GRADLE_DAEMON_BIND_ADDRESS` environment variable to explicitly specify the network address used for client-daemon and cross-daemon communication.
+
+Previously, Gradle attempted to auto-detect the local bind address, which could fail in environments with specific network configurations (multiple network interfaces, multiple TCP/IP stacks, etc.).
```
abstratt: 💅 

"Previously, Gradle always attempted to auto-detect..."

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -113,6 +113,21 @@ myCollection.add(MyType("other")) // this will fail
...
+A workaround existed via the `OPENSHIFT_*_IP` environment variables, but undocumented.
+__
+Setting `GRADLE_DAEMON_BIND_ADDRESS` to an IP address or hostname will override the auto-detected address:
```
abstratt: ❓ 

Will override or skip auto-detection altogether?

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -113,6 +113,21 @@ myCollection.add(MyType("other")) // this will fail
 myCollection.remove(main)         // this will fail
 ```
 
+### Explicit bind address for client-daemon and cross-daemon communication
+
+Gradle now supports the `GRADLE_DAEMON_BIND_ADDRESS` environment variable to explicitly specify the network address used for client-daemon and cross-daemon communication.
+
+Previously, Gradle attempted to auto-detect the local bind address, which could fail in environments with specific network configurations (multiple network interfaces, multiple TCP/IP stacks, etc.).
+A workaround existed via the `OPENSHIFT_*_IP` environment variables, but undocumented.
```
abstratt: 💭 Maybe we should not even mention the support for via `OPENSHIFT_*_IP`? No one that does not use it yet will ever need it anyways.

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -113,6 +113,21 @@ myCollection.add(MyType("other")) // this will fail
...
+__
+Setting `GRADLE_DAEMON_BIND_ADDRESS` to an IP address or hostname will override the auto-detected address:
+
+```text
+GRADLE_DAEMON_BIND_ADDRESS=192.168.1.10 ./gradlew build
+```
+
+`GRADLE_DAEMON_BIND_ADDRESS` takes precedence over any `OPENSHIFT_*_IP` variable if both are set.
```
abstratt: 💭 

I would not mention this

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/build_environment.adoc
```diff
@@ -535,6 +535,19 @@ Specifies the `GRADLE_USER_HOME` directory for Gradle to store its global config
 +
 Defaults to `USER_HOME/.gradle` if not set.
 
+`GRADLE_DAEMON_BIND_ADDRESS`::
+Specifies the IP address or hostname that Gradle uses to bind the local socket for client-daemon and cross-daemon communication.
++
+By default, Gradle automatically selects the loopback address (or wildcard address as a fallback) for this socket.
+This may not work correctly in environments with specific network configurations, such as multiple network interfaces, multiple TCP/IP stacks etc.
++
+Setting this variable overrides the auto-detected address:
```
abstratt: ❓  Overrides or skips auto-detection?

platforms/documentation/docs/src/docs/release/notes.md
abstratt: Hmm, it doesn't skip, it just overrides indeed. Shouldn't we just skip in case we have either override?

abstratt: If we skipped detection, we could avoid these calls:

* `NetworkInterface.getNetworkInterfaces()`, via `InetAddresses.analyzeNetworkInterfaces()`, because we may not need `InetAddresses` at all
* `NetworkInterface.getInetAddresses`, via `InetAddresses.analyzeNetworkInterface(NetworkInterface)` for the same reason

abstratt: By not instantiating `InetAddresses`, we would avoid a couple of system calls.

platforms/documentation/docs/src/docs/release/notes.md
asodja: Fixed

platforms/documentation/docs/src/docs/release/notes.md
asodja: Fixed

platforms/documentation/docs/src/docs/release/notes.md
asodja: Removed `OPENSHIFT_*_IP`

platforms/documentation/docs/src/docs/release/notes.md
asodja: Removed

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/build_environment.adoc
asodja: Fixed

asodja:
@bot-gradle test this

platforms/core-runtime/messaging/src/main/java/org/gradle/internal/remote/internal/inet/InetAddressFactory.java
```diff
@@ -110,31 +110,26 @@ private void findLocalBindingAddress() {
         }
     }
 
-    private void handleOpenshift() {
```
abstratt, line 103: 💅 

`find` suggests it is a read-only operation, `boolean handleX()` seems better to me, since their main purpose is to mutate state

abstratt: Never mind, @asodja , this is in line with the existing `findLocal...` method

platforms/core-runtime/messaging/src/main/java/org/gradle/internal/remote/internal/inet/InetAddressFactory.java
```diff
@@ -78,14 +88,14 @@ private void init() throws Exception {
...
-        if (inetAddresses == null) { // For testing
-            inetAddresses = new InetAddresses();
-        }
-
         wildcardBindingAddress = new InetSocketAddress(0).getAddress();
 
-        findLocalBindingAddress();
-        handleOpenshift();
+        if (!findGradleDaemonBindAddress() && !findOpenshiftAddress()) {
+            if (inetAddresses == null) { // For testing
```
abstratt: 💅 Maybe move this lazy initialization to `findLocalBindingAddress()`, as it is the only client

platforms/core-runtime/messaging/src/test/groovy/org/gradle/internal/remote/internal/inet/InetAddressFactoryTest.groovy
```diff
@@ -100,28 +100,27 @@ class InetAddressFactoryTest extends Specification {
...
-        defaultAddresses()
         environment["GRADLE_DAEMON_BIND_ADDRESS"] = "192.168.1.10"
 
         then:
         factory.localBindingAddress == InetAddress.getByName("192.168.1.10")
+        0 * addresses._
     }
 
-    def "GRADLE_DAEMON_BIND_ADDRESS takes precedence over OPENSHIFT address"() {
+    def "GRADLE_DAEMON_BIND_ADDRESS skips auto-detection regardless of OPENSHIFT address"() {
```
abstratt: 💅 

It seems the original test case name was still accurate, this is not about auto-detection, but about precedence between the two env vars

abstratt:
Thanks, Anže, I left a couple of comments

PR #37044, opened 2026-03-11
Make ant-impl an api-java project

ghale:
This project was busted out of `core` in #36907.
 
For now, ant-impl needs to be an api-java project because it contains `AntTarget`, which is a public api class with a dependency on core (for `ConventionTask`).  Once `ConventionTask` is pulled out of core, presumably we could pull `AntTarget` higher up, potentially to the ant-api project and make ant-impl an implementation-java project.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

PR #37046, opened 2026-03-11
Update non production dependencies

ljacomet:
This is motivated by older dependencies having CVEs and done for hygiene

* log4j-core to 2.25.3
* aicompressor to 2.0.3
* jetty to 9.4.58.v20250814
* selenium to 2.52.2

Also updated dependency verification metadata.

The test change is needed because Jetty added more validation and is now
complaining when setting the resourceBase while the server is started.
And it is started before every test thanks to the `@Rule`


platforms/core-configuration/kotlin-dsl/build.gradle.kts
```diff
@@ -2,6 +2,14 @@ import com.github.jengelman.gradle.plugins.shadow.ShadowJavaPlugin.Companion.sha
 import gradlebuild.basics.PublicKotlinDslApi
 import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
 
+buildscript {
+    dependencies {
+        constraints {
```
cobexer: ❓ do we need this given we have https://github.com/gradle/gradle/blob/a13d5ee3f8d34a2b21e9378089c0718060353620/build-logic-commons/build-platform/build.gradle.kts#L14-L15?

gradle/verification-metadata.xml
```diff
@@ -778,9 +783,9 @@
             <sha256 value="7de9b7ade957c6a1996e9c3d0e1e4d1e1840c081ebf16bbd3c2a6a5b05b88891" reason="Artifact is not signed"/>
          </artifact>
       </component>
-      <component group="io.airlift" name="aircompressor" version="0.27">
-         <artifact name="aircompressor-0.27.jar">
-            <pgp value="9E93DB1192CEE3D3B581AABB37E9D58EA4598641"/>
+      <component group="io.airlift" name="aircompressor" version="2.0.3">
+         <artifact name="aircompressor-2.0.3.jar">
+            <pgp value="64EE0DF33CBEB345471E1233824B40F0D3D8E241"/>
```
cobexer: 💅 Maybe move the key to the section above into a `trusted-key` to simplify updates

platforms/core-configuration/kotlin-dsl/build.gradle.kts
ljacomet: No we do not, you are correct.

ljacomet:
@bot-gradle test RfR

ljacomet:
I cannot reproduce the smoke test failure locally, let's see if this merges.

PR #37052, opened 2026-03-11
Hash segments of HTML test report

### LINKED ISSUE #36996: Test Report: Nested: File name too long
### Current Behavior

Upgrading the JetBrains/kotlin project to Gradle 9.4 caused build failures when generating test reports.
```
File name too long: '/opt/buildAgent/work/17e640964edd2053/kotlin/native/native.tests/klib-ir-inliner/build/reports/tests/test/org.jetbrains.kotlin.konan.test.blackbox.NativeCodegenBoxWithInlinedFunInKlibTestGenerated/org.jetbrains.kotlin.konan.test.blackbox.NativeCodegenBoxWithInlinedFunInKlibTestGenerated$Box/org.jetbrains.kotlin.konan.test.blackbox.NativeCodegenBoxWithInlinedFunInKlibTestGenerated$Box$Ranges/org.jetbra-8UB7PO1MP775A.kotlin.konan.test.blackbox.NativeCodegenBoxWithInlinedFunInKlibTestGenerated$Box$Ranges$Stepped/o-U8GQRQ01FA18E.kotlin.konan.test.blackbox.NativeCodegenBoxWithInlinedFunInKlibTestGenerated$Box$Ranges$Stepped$Unsigned/org-8GQH5QGA24K4S.test.blackbox.NativeCodegenBoxWithInlinedFunInKlibTestGenerated$Box$Ranges$Stepped$Unsigned$Expression/o-L2QJDBRKVDRF2.blackbox.NativeCodegenBoxWithInlinedFunInKlibTestGenerated$Box$Ranges$Stepped$Unsigned$Expression$DownTo/o-24H6DVU76EQE0.NativeCodegenBoxWithInlinedFunInKlibTestGenerated$Box$Ranges$Stepped$Unsigned$Expression$DownTo$Reversed/testStepThenReversedThenStepThenReversed().html'.
```

The very long pathname can be observed by using `@Nested` tests

```kotlin
class SampleTest {
    @Nested
    inner class ThisIsAVeryLongClassName {
        @Nested
        inner class ThisIsAVeryLongClassName1 {
            @Nested
            inner class ThisIsAVeryLongClassName2 {
                @Nested
                inner class ThisIsAVeryLongClassName3 {

                    @Nested
                    inner class ThisIsAVeryLongClassName4 {
                        @Test
                        fun thisIsAlsoARatherLongMethodName() {
                            println("Hello, world!")
                        }
                    }
                }
            }
        }
    }
}
```

Nested tests will create a folder for each nesting with the FQN of the nested class. This can exceed the maximum path length.

### Expected Behavior

I expect that `@Nested` tests, similar to static nested classes, generated a single filepath within bounds. 

### Context (optional)

_No response_

[reproducer.zip](https://github.com/user-attachments/files/25837831/reproducer.zip)

### Gradle version

9.4.0

### Build scan URL (optional)

https://ge.jetbrains.com/s/6ccrtufthkfi4

_No response_

octylFractal:
This prevents easily going over the max file path length due to whatever name is generated for nodes by the test engine. Fixes #36996

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
@@ -16,6 +16,7 @@
 
 package org.gradle.api.internal.tasks.testing.report.generic;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
@@ -34,7 +35,8 @@ import org.gradle.api.internal.tasks.testing.worker.TestEventSerializer;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
 import org.gradle.api.tasks.testing.TestOutputEvent;
-import org.gradle.internal.SafeFileLocationUtils;
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hashing;
 import org.gradle.internal.UncheckedException;
 import org.gradle.internal.concurrent.CompositeStoppable;
 import org.gradle.internal.file.Deleter;
@@ -59,7 +61,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
+import java.util.Base64;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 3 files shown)


platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
```diff
@@ -71,26 +73,31 @@
...
-        } else if (isLeaf && !Objects.equals(path.getName(), "index")) {
-            // Avoid using a directory for each leaf node unless its name clashes (i.e. "index")
+            return "index.html";
+        } else if (isLeaf) {
+            // Avoid using a directory for each leaf node
             // This reduces VFS overhead from many directories for large test suites
             String prefix = String.join("/", Iterables.transform(
                 path.getParent().segments(),
-                name -> SafeFileLocationUtils.toSafeFileName(name, true)
+                GenericHtmlTestReportGenerator::hashSegment
```
tresat, line 92: 🤔 I think we should be more selective here.  Most of the time the hierarchy won't be deep, the class names will be fine (non-nested, reasonable length), and using these names for naming leaf dirs is more ergonomic than using a hash.

Can't we keep the simple case unchanged?  Check if there is "$" = nested class or the length > some cutoff.  If not, use the name as before?

octylFractal: The problem with that is that we would need to start doing collision detection against the hashes to ensure that for some reason we don't end up with the exact 11 characters that match a class name. As-is the fingerprint hash I'm using is designed to minimize these among the hash output. This would make the change more complicated and I would like to minimize changes for a patch release.

Additionally, the max length we could do before it would start cutting significantly into nesting capability would be something like 128 characters, and because it currently includes a package name, we're already quite close to that with a lot of our own test classes. So it might not end up helping very much.

Perhaps in 9.6.0 we can consider alternatives to this?

tresat: > The problem with that is that we would need to start doing collision detection against the hashes to ensure that for some reason we don't end up with the exact 11 characters that match a class name.

While it's true that this would increase the odds of a hash collision, the percent it increases the odds by is probably so infinitesimal that practically speaking it could just be ignored.  The odds of this happening without someone maliciously reverse-engineering a particular classname are probably not much higher than a hash collision between the path segments.  I should have stated this assumption out loud here.

> Additionally, the max length we could do before it would start cutting significantly into nesting capability would be something like 128 characters, and because it currently includes a package name, we're already quite close to that with a lot of our own test classes. So it might not end up helping very much.

Yeah, including the FQNs could be a problem with only 128 chars.  If we dropped to using just simple names then we'd have more room, but a very real chance of name collisions.

I ultimately dislike how we keep sacrificing the usability of typical use cases in order to support extreme use cases.  The vast majority of projects won't nest classes more than 1 level deep.

To be clear, I think this is a fine solution for that extreme case, but I think it's just not needed most of the time.  Ideally, we could do something like 1) check if long names would be a problem before writing any files, 2) do things the current way if not, and start hashing segments if so.  I understand that persisting the result of the check in 1) would add more complexity, but it might be worth it?

I'm fine shipping this as is and thinking of improvements in 9.6.

octylFractal: I like the idea of doing this as a fallback, but I tried to implement this and since we don't actually know the full file path at certain parts of the code, it gets more involved than I would like. Let's plan to do this in 9.6.0.

PR #37065, opened 2026-03-12
Use blocking notification during remote build cache load

octylFractal:
This isn't exactly right, but it provides some performance improvement.

It seems that DV does very little CPU work in their load function, so this should be a good enough approximation.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/DefaultBuildCacheController.java
+++ b/platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/DefaultBuildCacheController.java
@@ -72,8 +72,6 @@ import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
 
 public class DefaultBuildCacheController implements BuildCacheController {
-    private final BlockingNotifier blockingNotifier;
-
     @VisibleForTesting
     final RemoteBuildCacheServiceHandle remote;
 
@@ -97,9 +95,8 @@ public class DefaultBuildCacheController implements BuildCacheController {
         OriginMetadataFactory originMetadataFactory,
         Interner<String> stringInterner
     ) {
-        this.blockingNotifier = blockingNotifier;
         this.local = toLocalHandle(config.getLocal(), config.isLocalPush(), buildOperationRunner);
-        this.remote = toRemoteHandle(config.getBuildPath(), config.getRemote(), config.isRemotePush(), buildOperationRunner, buildOperationProgressEventEmitter, logStackTraces, disableRemoteOnError);
+        this.remote = toRemoteHandle(config.getBuildPath(), config.getRemote(), config.isRemotePush(), buildOperationRunner, buildOperationProgressEventEmitter, logStackTraces, disableRemoteOnError, blockingNotifier);
         this.tmp = toTempFileStore(config.getLocal(), temporaryFileFactory);
         this.packExecutor = new PackOperationExecutor(
             buildOperationRunner,
@@ -139,7 +136,7 @@ public class DefaultBuildCacheController implements BuildCacheController {
         tmp.withTempFile(((BuildCacheKeyInternal) key).getHashCodeInternal(), file -> {
             Optional<BuildCacheLoadResult> remoteResult;
             try {
-                remoteResult = blockingNotifier.blocking(() -> remote.maybeLoad(key, file, f -> packExecutor.unpack(key, entity, f)));
+                remoteResult = remote.maybeLoad(key, file, f -> packExecutor.unpack(key, entity, f));
             } catch (Exception e) {
                 throw new BuildCacheOperationException("Could not load from remote cache: " + e.getMessage(), e);
             }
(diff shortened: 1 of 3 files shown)


octylFractal:
A run without verbose locking (which affects the performance significantly here) with the `rebuild_bl_abi_change_pip_cold` scenario:
Screenshot 2026-03-11 at 6 12 18 PM

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultWorkerLeaseService.java
```diff
@@ -549,9 +586,28 @@ DefaultWorkerLease newUnmanagedLease() {
...
             super(displayName, coordinationService, owner, parent);
         }
+        public void markDormant() {
+            isDormant = true;
+        }
+
+        @Override
+        protected boolean acquireLock() {
+            if (super.acquireLock()) {
+                return true;
```
reinsch82, line 602: ❓ should `isDormant` also be set to `true` here?

octylFractal: Do you mean `false`? Because I think you're right, but with `false`.

reinsch82: Yes I meant `false`:)

octylFractal:
@bot-gradle test PT

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLeaseRegistry.java
```diff
@@ -90,7 +92,8 @@ public interface ProjectLeaseRegistry {
     /**
      * {@link #blocking(Factory)}, but returns no result.
      */
-    void blocking(Runnable action);
+    @Override
+    void blocking(@NonNull Runnable action);
```
jvandort: Can we null mark this entire class instead of per-parameter non-nulls?

platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/DefaultBuildCacheController.java
```diff
@@ -135,7 +139,7 @@ private Optional<BuildCacheLoadResult> loadRemoteAndStoreResultLocally(BuildCach
         tmp.withTempFile(((BuildCacheKeyInternal) key).getHashCodeInternal(), file -> {
             Optional<BuildCacheLoadResult> remoteResult;
             try {
-                remoteResult = remote.maybeLoad(key, file, f -> packExecutor.unpack(key, entity, f));
+                remoteResult = blockingNotifier.blocking(() -> remote.maybeLoad(key, file, f -> packExecutor.unpack(key, entity, f)));
```
jvandort: Putting the `blocking` here also drops the lock for the unpacking operation. How CPU bound is unpacking vs IO bound? 

Did we consider just wrapping [loadInner](https://github.com/gradle/gradle/blob/cb0c615fb8e3690971bb7f89ad80f58943360624/platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/service/BaseRemoteBuildCacheServiceHandle.java#L101), which should just block for the HTTP request?

octylFractal: My instinct is that gzip is pretty cheap so unpacking should also be heavily I/O bound. I could adjust it to just wrap `loadInner` and the writes to disk by `TarBuildCacheEntryPacker`?

jvandort: I'd prefer to measure IO vs CPU time before we wrap packing as well. I think we're going to get the most value out of this by dropping the locks during network-bound traffic anyway since I expect that to be order(s) of magnitude slower (latency wise) than local disk IO

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLeaseRegistry.java
octylFractal: Yeah, though I'm unsure if the other methods will be perfectly accurate. I suppose people can adjust as needed.

platforms/core-execution/build-cache/src/main/java/org/gradle/caching/internal/controller/DefaultBuildCacheController.java
octylFractal: Fair. I'll move it for now to get this merged in quicker, we can iterate further if we find that unpacking is showing up as a hot spot.

PR #37089, opened 2026-03-12
Make BuildScriptVisibilityIntegrationTest CC compatible

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/integTest/groovy/org/gradle/api/BuildScriptVisibilityIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/api/BuildScriptVisibilityIntegrationTest.groovy
@@ -17,10 +17,11 @@
 package org.gradle.api
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
-import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
+import org.gradle.integtests.fixtures.ToBeFixedForIsolatedProjects
+import org.gradle.integtests.fixtures.executer.GradleContextualExecuter
 
 class BuildScriptVisibilityIntegrationTest extends AbstractIntegrationSpec {
-    @ToBeFixedForConfigurationCache(because = "test expects scripts evaluation")
+    @ToBeFixedForIsolatedProjects(because = "project cannot dynamically look up a method in the parent project")
     def "methods defined in project build script are visible to descendant projects"() {
         createDirs("child1")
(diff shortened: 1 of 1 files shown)


subprojects/core/src/integTest/groovy/org/gradle/api/BuildScriptVisibilityIntegrationTest.groovy
```diff
@@ -49,13 +48,17 @@ println "child: " + doSomethingElse(11)
...
-        outputContains("child: {11}")
-        outputContains("child: [11]")
+        if (GradleContextualExecuter.notConfigCache) {
+            outputContains("root: {10}")
+            outputContains("root: [10]")
+            outputContains("child: {11}")
+            outputContains("child: [11]")
+        } else {
+            outputDoesNotContain("root:")
+            outputDoesNotContain("child:")
```
tresat, line 60: ❗ Doesn't this verify the exact opposite of the method's name?  This is testing that the methods **are not** visible from the child project.

I think if this is meant the case, it would be better to just mark these tests as `@UnsupportedWithConfigurationCache` and leave a note.

ov7a: They're visible for the first time but not for the second time because CC is reused - I think it's a reasonable expectation?

tresat: Doesn't this ensure that there aren't visible even once, though?

ov7a: Sorry, I'm not following. The first time is covered by lines 42-47. What do you mean by "there aren't visible even once"?

tresat: Oh, I missed that there was a 2nd `succeeds()` call right below the `and:` - didn't notice this when I expanded the folded section of code.  I was wondering how the `outputContains` in lines 44-47 could coexist with the `outputDoesNotContain` in lines 57-58.

This makes sense to me now, sorry for the confusion.

blindpirate:
@bot-gradle test this

PR #37092, opened 2026-03-12
Replace per-operation ConcurrentHashMap with more performant structure

big-guy:
ProgressShieldingBuildOperationListener tracked active operations in a ConcurrentHashMap, allocating Node objects on every started/finished call. Replace with a monotonic ID watermark: listeners record the most recent operation ID at registration time and ignore events for older operations. This eliminates ~425MB of ConcurrentHashMap$Node allocations observed in profiling and removes all per-operation map overhead.

There were two other similar maps in other places that were replaced as well.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/logging/build.gradle.kts
+++ b/platforms/core-runtime/logging/build.gradle.kts
@@ -33,7 +33,6 @@ dependencies {
     api(libs.slf4jApi)
 
     implementation(projects.concurrent)
-    implementation(libs.fastutil)
     implementation(projects.io)
     implementation(projects.messaging)
     implementation(projects.serviceRegistryBuilder)
--- a/platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/sink/OutputEventTransformer.java
+++ b/platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/sink/OutputEventTransformer.java
@@ -22,14 +22,13 @@ import org.gradle.internal.logging.events.ProgressCompleteEvent;
 import org.gradle.internal.logging.events.ProgressEvent;
 import org.gradle.internal.logging.events.ProgressStartEvent;
 import org.gradle.internal.logging.events.RenderableOutputEvent;
-import it.unimi.dsi.fastutil.longs.LongOpenHashSet;
-import it.unimi.dsi.fastutil.longs.LongSet;
-import it.unimi.dsi.fastutil.longs.LongSets;
 import org.gradle.internal.operations.BuildOperationCategory;
 import org.gradle.internal.operations.OperationIdentifier;
 import org.gradle.util.internal.GUtil;
 
+import java.util.Collections;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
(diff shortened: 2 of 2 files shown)


big-guy:
Bah, this has a fundamental race problem.
Imagine:
- op1 has id=1
- op2 has id=2
- op2 is started
- listener registered
- op1 started
- op2 progress - Wrong!
- op2 finished
- op1 finished

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/SynchronizedLongHashSet.java
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * A thread-safe primitive long hash set using open addressing with linear probing.
+ *
+ * Uses 0 as the empty sentinel and -1 as the tombstone for deleted entries,
+ * so neither value can be stored as an element.
+ *
+ * Zero per-element object allocation — only allocates a new backing array when rehashing.
+ */
+public class SynchronizedLongHashSet {
```
mlopatkin: ❓ We have `LongOpenHashSet` from `fastutil`, maybe it can work instead of a hand-rolled implementation? Looks like there is a synchronized wrapper in `LongSets` as well, though we may choose manual synchronization to save an extra object.

big-guy: Yep, I thought about that after.  This was an easy replacement.

platforms/core-runtime/build-operations/src/main/java/org/gradle/internal/operations/DefaultBuildOperationListenerManager.java
```diff
@@ -84,34 +85,43 @@ public BuildOperationListener getBroadcaster() {
...
             }
         }
 
         @Override
         public void finished(BuildOperationDescriptor buildOperation, OperationFinishEvent finishEvent) {
-            active.remove(buildOperation.getId());
-            delegate.finished(buildOperation, finishEvent);
+            OperationIdentifier id = buildOperation.getId();
+            if (id != null && activeOps.remove(id.getId())) {
+                delegate.finished(buildOperation, finishEvent);
```
mlopatkin, line 123: 🤔  We forward `started` when the `id` is null; does it make sense to forward finished there too?

This is unlikely to happen, though, CHM is null-hostile.

big-guy: I think the nullability of getId is mostly wrong and this is never null. 

This is the only place that can create a BuildOperationDescriptor with a null id and it's just used for creating a display name:
https://github.com/gradle/gradle/blob/09deb3ed3aeab351aa5189be9bb5306652536b01/platforms/native/platform-native/src/main/java/org/gradle/nativeplatform/toolchain/internal/DefaultCommandLineToolInvocationWorker.java#L58

PR #37093, opened 2026-03-12
Add option to enable/disable differential stack generation

hegyibalint:
This PR adds a `--generate-diffs` (default `true`) option.

The reason to add this is to skip the lengthy and expensive `perl` run generating the differential stacks and SVGs if not needed. 

This can speed up Ad-Hoc evaluations where the raw stacks are the only thing needed.

### DIFF
--- a/.teamcity/src/main/kotlin/util/AdHocPerformanceScenario.kt
+++ b/.teamcity/src/main/kotlin/util/AdHocPerformanceScenario.kt
@@ -48,7 +48,13 @@ abstract class AdHocPerformanceScenario(
             param("checks", "all")
             text("runs", "40", display = ParameterDisplay.PROMPT, allowEmpty = false)
             text("warmups", "10", display = ParameterDisplay.PROMPT, allowEmpty = false)
-            text("generateDiffs", "true", display = ParameterDisplay.PROMPT, allowEmpty = false, description = "Whether to generate differential flame graphs after profiling")
+            text(
+                "generateDiffs",
+                "true",
+                display = ParameterDisplay.PROMPT,
+                allowEmpty = false,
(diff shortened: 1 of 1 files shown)


hegyibalint:
@bot-gradle test this

build-logic/performance-testing/src/main/groovy/gradlebuild/performance/tasks/PerformanceTest.groovy
```diff
@@ -236,6 +239,11 @@ abstract class PerformanceTest extends DistributionTest {
         this.checks = checks
     }
 
+    @Option(option = "generate-diffs", description = "Whether to generate differential flame graphs after profiling")
+    void setGenerateDiffs(boolean generateDiffs) {
```
6hundreds, line 243: Maybe it's more idiomatic to keep it like https://github.com/gradle/gradle/blob/d54ece1fbd38e5288bb7d39fd04e5fc8a68bf25c/build-logic/performance-testing/src/main/groovy/gradlebuild/performance/tasks/PerformanceTest.groovy#L252-L254

hegyibalint: I don't get your improvement. This is about enabling/disabling a heavy post-processing of the stacks, which applies to all ad-hoc runs.

If you have a concrete idea, please suggest a new parameter name!

6hundreds: The idea is to have it as `Provider`-based property, not as a regular setter/getter

PR #37094, opened 2026-03-12
Increate test worker memory, so it can process larger stacks

hegyibalint:
Ran into some problems when trying to run AdHoc performance scenarios on slow performance tests.
Even with reduced `runs` amount (~10), the stacks are simply too large.

This PR bumps up the worker memory to 8GB (completely arbitrary incrase), with the hopes of it making these OOM problems go away.


hegyibalint:
@bot-gradle test this

PR #37096, opened 2026-03-12
Enable zip64 for performance test results zip

hegyibalint:
We have large enough performance measurement scenarios where the resulting ZIP can be more than 4GB large, requiring ZIP64.


hegyibalint:
@bot-gradle test this

hegyibalint:
> I don't think TeamCity will accept such large artifacts; thus, working with huge performance test results might need direct agent access to look into the results.

Even the right-click save-as wouldn't work in your opinion?
I don't care about looking into the files there TBH, if you are doing this, you just want the ZIP.

PR #37097, opened 2026-03-12
Extract normalization packages out to core-execution project

ghale:
Core and core-api are large monoliths that contain api and implementation classes that cross multiple platforms.  Having these classes lumped together encourages tangling and circular relationships between platforms and can lead to relationships that do not comply with our platform rules.  It also leads to extremely coarse dependencies that drag in much more than is necessary and can cause classes to be misplaced solely in order to workaround the limitations of these coarse relationships.  Breaking up core/core-api into smaller projects helps to prevent tangling and allows for narrower, more granular dependency relationships.  It also clears the path for us to start enforcing platform relationships in a way that we can't easily do so today.

This extracts the `org.gradle.normalization` and `org.gradle.normalization.internal` packages from `core` and `core-api` and moves them to two new projects in the core-execution platform:
- normalization-api: The public API classes associated with normalization
- normalization: The internal API and implementation classes associated with normalization

Additionally, we hoist normalization-java up to the core-execution platform.  This is because normalization has dependencies on this project, but normalization-java also needs to be self-contained as it is published and used by downstream projects.  If we left it in the jvm platform, we would have an illegal platform dependency (i.e. core-execution -> jvm).  If we refactor the relationship such that normalization-java instead depends on normalization, we add additional dependencies for downstream consumers.(i.e. normalization and _all its dependencies_ would need to now be published).  If we put normalization in the jvm platform, it just creates a new cross-platform violation.

So, the least disruptive solution is to simply move normalization-java to core-execution, which allows it stay self-contained for consumers, and avoids platform violations.  This makes sense, too, as this project is not really related to _building_ java projects.  It is instead functionality related to snapshotting classpaths, which is more widely applicable than only to java projects.  It probably should have been a part of core-execution in the first place.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

bamboo:
Hi @ghale, could you add the main reason(s) why we're doing this at the very top of the PR description?

ghale:
> Hi @ghale, could you add the main reason(s) why we're doing this at the very top of the PR description?

Done.

platforms/core-execution/normalization-api/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
+import org.gradle.api.publish.internal.component.ConfigurationVariantDetailsInternal
+import org.gradle.kotlin.dsl.assign
```
big-guy: These look unnecessary?

platforms/core-execution/normalization/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
+import org.gradle.api.publish.internal.component.ConfigurationVariantDetailsInternal
+import org.gradle.kotlin.dsl.assign
```
big-guy: This looks unnecessary?

big-guy (changes requested):

platforms/core-execution/normalization-api/build.gradle.kts
ghale: Bah, I totally missed that.  Removed.

platforms/core-execution/normalization/build.gradle.kts
ghale: Removed

big-guy (approved):

PR #37110, opened 2026-03-13
Align ProjectIdentity and Project display names

alllex:
This allows for uniformly presenting the project display name, regardless of the relevant type being used:
- ProjectIdentity
- ProjectState
- Project

This is **not a breaking change** as the display name of the component identifier is not a public contract.

The changeset is large mainly due to the fact that `ProjectComponentIdentifier.getDisplayName()` was producing a string that was not aligned with that of `Project`'s display name, and the component identifier's display name was used in a lot of tests and fixtures for assertions.

### DIFF
--- a/subprojects/core-api/src/main/java/org/gradle/api/internal/project/ProjectIdentity.java
+++ b/subprojects/core-api/src/main/java/org/gradle/api/internal/project/ProjectIdentity.java
@@ -119,11 +119,7 @@ public final class ProjectIdentity implements DisplayName {
         this.projectName = projectName;
 
         this.buildTreePath = computeProjectIdentityPath(buildPath, projectPath);
-
-        // TODO: This is inconsistent with DefaultProject.getDisplayName.
-        // We should change this to match that of DefaultProject.
-        String prefix = Path.ROOT.equals(buildTreePath) ? "root project" : "project";
-        this.displayName = Describables.memoize(Describables.of(prefix, buildTreePath.asString()));
+        this.displayName = computeDisplayName(buildTreePath, projectName);
     }
 
     /**
@@ -156,6 +152,13 @@ public final class ProjectIdentity implements DisplayName {
         return projectName;
     }
 
+    /**
+     * Returns the display name of this project in a human-readable format.
+     * <ul>
+     *     <li>For the root project: {@code root project 'projectName'}</li>
+     *     <li>For subprojects: {@code project ':identity:path:of:project'}</li>
+     * </ul>
+     */
     @Override
     public String getDisplayName() {
         return displayName.getDisplayName();
@@ -188,4 +191,11 @@ public final class ProjectIdentity implements DisplayName {
         return buildTreePath.hashCode();
     }
 
+    private static DisplayName computeDisplayName(Path buildTreePath, String projectName) {
+        return Describables.memoize(
+            Path.ROOT.equals(buildTreePath)
+                ? Describables.withTypeAndName("root project", projectName)
+                : Describables.withTypeAndName("project", buildTreePath.asString())
+        );
+    }
 }
--- a/subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
+++ b/subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
@@ -878,16 +878,10 @@ public abstract class DefaultProject extends AbstractPluginAware implements Proj
 
     /**
      * Returns the display name of this project in a human-readable format.
-     * <p>
-     * Currently:
-     * <ul>
-     *     <li>For the root project: {@code root project 'projectName'}</li>
-     *     <li>For subprojects: {@code project ':identity:path:of:project'}</li>
-     * </ul>
      */
     @Override
     public String getDisplayName() {
-        return owner.getDisplayName().getDisplayName();
+        return getProjectIdentity().getDisplayName();
     }
 
     @Override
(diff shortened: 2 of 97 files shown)


alllex:
@bot-gradle test this without pts

alllex:
@bot-gradle test this without pts

alllex:
@bot-gradle test this without pts

subprojects/core-api/src/main/java/org/gradle/api/internal/project/ProjectIdentity.java
```diff
@@ -118,14 +118,11 @@ private ProjectIdentity(
...
         this.projectName = projectName;
 
-        this.buildTreePath = computeProjectIdentityPath(buildPath, projectPath);
-
-        // TODO: This is inconsistent with DefaultProject.getDisplayName.
-        // We should change this to match that of DefaultProject.
-        String prefix = Path.ROOT.equals(buildTreePath) ? "root project" : "project";
-        this.displayName = Describables.memoize(Describables.of(prefix, buildTreePath.asString()));
+        buildTreePath = computeProjectIdentityPath(buildPath, projectPath);
+        displayName = computeDisplayName(buildTreePath, projectName);
```
alllex: ℹ️  This is the main functional change. The display name is adjusted to match exactly the display name of the `Project` instances, despite its pecularity

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
```diff
@@ -878,16 +878,10 @@ private Project evaluationDependsOn(ProjectInternal projectToEvaluate) {
...
-     * Currently:
-     * <ul>
-     *     <li>For the root project: {@code root project 'projectName'}</li>
-     *     <li>For subprojects: {@code project ':identity:path:of:project'}</li>
-     * </ul>
      */
     @Override
     public String getDisplayName() {
-        return owner.getDisplayName().getDisplayName();
+        return getProjectIdentity().getDisplayName();
```
alllex, line 884: ℹ️  So now the `ProjectIdentity` can be used as a display name for the `Project`

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
```diff
@@ -314,12 +313,7 @@ private class ProjectStateImpl implements ProjectState, Closeable {
 
         @Override
         public DisplayName getDisplayName() {
-            Path identityPath = identity.getBuildTreePath();
-            if (identityPath.equals(Path.ROOT)) {
-                return Describables.withTypeAndName("root project", identity.getProjectName());
-            } else {
-                return Describables.withTypeAndName("project", identityPath.asString());
-            }
+            return identity;
```
alllex, line 316: ℹ️  So now the `ProjectIdentity` can be used as a display name for the `Project`

subprojects/core-api/src/main/java/org/gradle/api/internal/project/ProjectIdentity.java
```diff
@@ -156,6 +153,15 @@ public String getProjectName() {
         return projectName;
     }
 
+    /**
+     * Returns the display name of this project in a human-readable format.
+     * <p>
+     * The display name is not strictly a public contract, but it has always been computed like this:
```
jvandort: It's not strictly a public contract, yes, but it's also not even kinda a public contract. Display names are for display and can be freely changed unless otherwise stated

I don't think it's necessary to call this out at all and doing so makes it seem like this is some special case

You also mention these display names "has always" been computed like written. But, the "root project" format is relatively new within the last major or so. Before that root project was printed like all other projects (project :)

subprojects/core-api/src/main/java/org/gradle/api/internal/project/ProjectIdentity.java
```diff
@@ -118,14 +118,11 @@ private ProjectIdentity(
...
-
-        // TODO: This is inconsistent with DefaultProject.getDisplayName.
-        // We should change this to match that of DefaultProject.
-        String prefix = Path.ROOT.equals(buildTreePath) ? "root project" : "project";
-        this.displayName = Describables.memoize(Describables.of(prefix, buildTreePath.asString()));
+        buildTreePath = computeProjectIdentityPath(buildPath, projectPath);
+        displayName = computeDisplayName(buildTreePath, projectName);
     }
 
+
```
jvandort: Extra newline

subprojects/core-api/src/main/java/org/gradle/api/internal/project/ProjectIdentity.java
```diff
@@ -118,14 +118,11 @@ private ProjectIdentity(
         this.projectPath = projectPath;
         this.projectName = projectName;
 
-        this.buildTreePath = computeProjectIdentityPath(buildPath, projectPath);
-
-        // TODO: This is inconsistent with DefaultProject.getDisplayName.
-        // We should change this to match that of DefaultProject.
-        String prefix = Path.ROOT.equals(buildTreePath) ? "root project" : "project";
-        this.displayName = Describables.memoize(Describables.of(prefix, buildTreePath.asString()));
+        buildTreePath = computeProjectIdentityPath(buildPath, projectPath);
```
jvandort: Nitpick, but can we go back to using `this.` to set fields in the constructor?

jvandort (approved):
Nice

subprojects/core-api/src/main/java/org/gradle/api/internal/project/ProjectIdentity.java
alllex: > You also mention these display names "has always" been computed like written. But, the "root project" format is relatively new within the last major or so. Before that root project was printed like all other projects (project :)

This might be true in some places, but not for the instances of `Project`, where that behavior seems to have been present since about Gradle 3.3:
https://github.com/gradle/gradle/commit/2c009b27b97c1564344f3cc93258ce5a0e18a03f#diff-27893c9a1d7c3c069980827799f450f02ba316abaafb65a5a8e9cd13e62346bdL655-R675

alllex: Having said that, I agree we can drop the line about the public contract from this javadoc

jvandort: Interesting. I stand corrected, I thought this was something Tom added

PR #37121, opened 2026-03-15
Add sync-only Isolated Projects flag

### LINKED ISSUE #37026: Isolated Project Support for Syncs Only
### Expected Behavior

Adding a flag, such as, `org.gradle.unsafe.isolated-projects-sync=true` which enables IP for syncs _only_. 

`org.gradle.unsafe.isolated-projects=true` is currently a build-wide switch. When enabled through `gradle.properties`, it affects both IDE sync (Tooling API model building) and normal command-line builds.

For some builds, this makes adoption difficult. A team may want to use Isolated Projects primarily as an IDE sync optimization, but cannot do so without also changing the behavior of regular builds.

### Context

Isolated Projects provides clear value for IDE sync scenarios, which Gradle explicitly calls out in the documentation. However, enabling it globally means:
	•	IDE sync may succeed and appear healthy
	•	regular builds may still hit Isolated Projects violations or behavioral differences

This creates an awkward adoption path. In practice, a team may want:
	•	faster IDE syncs for developers
	•	no change to normal CLI build behavior yet
	•	time to incrementally fix IP violations before enforcing the feature for all builds

paulhundal530:
## Summary
- Adds `org.gradle.unsafe.isolated-projects.sync` build option that enables Isolated Projects for IDE sync (model building) while leaving regular builds unaffected
- Sync operations (`isCreatesModel=true`) use full IP mode; build operations (`isRunsTasks=true`) use Configuration Cache mode only
- Same CC validation as full IP: Configuration Cache cannot be explicitly disabled when this flag is enabled
- Full IP flag takes precedence when both are set

## Motivation
This provides a migration path toward full Isolated Projects adoption. Users can benefit from IP during IDE sync without requiring their entire build to be IP-compliant.

## Test plan
- [x] Unit tests in `BuildModelParametersProviderTest` covering sync, build, mixed, CC validation, precedence, and isolation verification
- [x] Manual verification: flag detected from `gradle.properties`
- [x] Manual verification: builds use CC mode (no IP)
- [x] Manual verification: sync activates IP mode with "Isolated Projects is an incubating feature." message
- [x] Manual verification: `--no-configuration-cache` correctly fails with CC validation error
- [x] Manual verification: cross-project access violation detected during sync

Fixes https://github.com/gradle/gradle/issues/37026

Signed-off-by: Paul Hundal <paulhundal530@gmail.com>

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
@@ -98,6 +98,7 @@ object BuildModelParametersProvider {
             // TODO:isolated should this also disable IP?
             ccDisabledReason != null -> vintageMode(requirements, startParameter, options, ccDisabledReason)
             startParameter.isolatedProjects.get() -> isolatedProjectsMode(requirements, startParameter, options)
+            startParameter.isolatedProjectsSync.get() -> isolatedProjectsSyncMode(requirements, startParameter, options)
             startParameter.configurationCache.get() ->
                 if (requirements.isCreatesModel) {
                     // CC by itself does not yet support caching models or caching of the work graph that runs before model building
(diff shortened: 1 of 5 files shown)


alllex:
Thank you for the contribution. The changes are largely in the right place and mostly sufficient.
However, we need to find better naming, because Gradle doesn't have a concept of a "sync".

We are also thinking about other (future) aspects of the IP mode, such as parallelism vs incrementality and whether/how they should be enabled via flags. We want the picture to be coherent, since there are already many flags, and the behavior of each newly added flag must be considered alongside others. That is to say, that a decision to "quickly add one more boolean flag" is not necessarily an obvious one.

The use case is valid, and we want to provide a solution. The design might take a bit of time, though.

alllex:
I'm going to close this PR since we won't be using these changes directly. It's more valuable to have the corresponding issue open. The PR can still be used as a reference for the implementation, since it's linked.

PR #37129, opened 2026-03-16
Optimize Property.get() hot path with array-based cycle detection

big-guy:
Replace ReferenceOpenHashSet + ReferenceArrayList in EvaluationContext with a simple identity-based array stack. Linear scan for cycle detection is fast at typical evaluation depths (< 10) and avoids HashSet overhead on every property read.

Benchmarks show 5x throughput improvement for direct-value property reads (~45K → ~229K ops/ms) and ~15% improvement for provider-backed reads.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/functional/src/main/java/org/gradle/internal/evaluation/EvaluationContext.java
+++ b/platforms/core-runtime/functional/src/main/java/org/gradle/internal/evaluation/EvaluationContext.java
@@ -30,9 +30,6 @@ import org.jspecify.annotations.Nullable;
  * This class is thread-safe, but the returned contexts must be closed by the same thread that opens them.
  */
 public final class EvaluationContext {
-
-    private static final int INITIAL_CAPACITY = 8;
-
     private static final EvaluationContext INSTANCE = new EvaluationContext();
 
     private final ThreadLocal<PerThreadContext> threadLocalContext = ThreadLocal.withInitial(() -> new PerThreadContext(null));
@@ -151,6 +148,19 @@ public final class EvaluationContext {
      * deeper stacks.
      */
     private final class PerThreadContext implements EvaluationScopeContext {
+        /**
+         * This was chosen by looking at the typical stack sizes on the
+         * Gradle build. We're trying to balance resizing vs wasted space.
+         *
+         * Most stacks were small (only 1 element). The biggest stack was 19 elements.
+         *
+         * This size represented about 85% of all contexts, so Gradle does
+         * no resizing in those cases.
+         *
+         * The extreme case still requires 2 resizings.
+         */
+        private static final int INITIAL_CAPACITY = 8;
+
         private final ReferenceArrayList<EvaluationOwner> stack;
 
         @Nullable


platforms/core-runtime/functional/src/main/java/org/gradle/internal/evaluation/EvaluationContext.java
```diff
@@ -146,27 +144,38 @@ EvaluationScopeContext nested() {
...
         private void push(EvaluationOwner owner) {
-            if (objectsInScope.add(owner)) {
-                evaluationStack.add(owner);
-            } else {
+            // Identity scan for cycle detection
+            if (isInScope(owner)) {
                 throw prepareException(owner);
             }
+            if (size == stack.length) {
+                stack = Arrays.copyOf(stack, stack.length * 2);
```
asodja: 💅 Note that ArrayList is doing size = size * 1.5 (size = size + (size >> 1) which is a bit is a bit less agressive

platforms/core-runtime/functional/src/main/java/org/gradle/internal/evaluation/EvaluationContext.java
```diff
@@ -146,27 +144,38 @@ EvaluationScopeContext nested() {
...
+     *
+     * Cycle detection scans the array linearly, which is fast for the typical
+     * shallow evaluation depths (< 10). Automatically resizes to accommodate
+     * deeper stacks.
+     */
     private final class PerThreadContext implements EvaluationScopeContext {
-        private final Set<EvaluationOwner> objectsInScope = new ReferenceOpenHashSet<>(EXPECTED_MAX_CONTEXT_SIZE);
-        private final List<EvaluationOwner> evaluationStack = new ReferenceArrayList<>(EXPECTED_MAX_CONTEXT_SIZE);
+        private @Nullable EvaluationOwner[] stack;
+        private int size;
```
asodja: ❓ Can size ever become big enough that stack becomes slow and we should use Set again?

platforms/core-runtime/functional/src/main/java/org/gradle/internal/evaluation/EvaluationContext.java
```diff
@@ -146,27 +144,38 @@ EvaluationScopeContext nested() {
...
+     * Uses a simple identity-based array stack.
+     *
+     * Cycle detection scans the array linearly, which is fast for the typical
+     * shallow evaluation depths (< 10). Automatically resizes to accommodate
+     * deeper stacks.
+     */
     private final class PerThreadContext implements EvaluationScopeContext {
-        private final Set<EvaluationOwner> objectsInScope = new ReferenceOpenHashSet<>(EXPECTED_MAX_CONTEXT_SIZE);
-        private final List<EvaluationOwner> evaluationStack = new ReferenceArrayList<>(EXPECTED_MAX_CONTEXT_SIZE);
+        private @Nullable EvaluationOwner[] stack;
```
mlopatkin: ❓ Is using `ReferenceArrayList` significantly slower?

platforms/core-runtime/functional/src/main/java/org/gradle/internal/evaluation/EvaluationContext.java
```diff
@@ -146,27 +144,38 @@ EvaluationScopeContext nested() {
         return getContext().nested();
     }
 
+    /**
+     * Uses a simple identity-based array stack.
+     *
+     * Cycle detection scans the array linearly, which is fast for the typical
+     * shallow evaluation depths (< 10). Automatically resizes to accommodate
```
mlopatkin, line 147: 🤔 Do we have data to back the "typical shallow" assumption?

mlopatkin (commented):
Fine, and good to see perf improvements here, but is reimplementing the ReferenceArrayList necessary?

platforms/core-runtime/functional/src/main/java/org/gradle/internal/evaluation/EvaluationContext.java
big-guy: No, you're right. This could just be a ReferenceArrayList still.

platforms/core-runtime/functional/src/main/java/org/gradle/internal/evaluation/EvaluationContext.java
big-guy: I took a look at gradle/gradle:

- 144k evaluation contexts
- 45% have a stack size of 1
- 85% of all contexts have a stack size <=8
- 99% of all contexts have a stack size <=12
- largest stack size was 19

~8 seems to be a good size. 4 would be too small and 16 too large.

Issue #31479, opened 2024-12-02
DCL: Parser error when using kotlin.dsl.dcl=true


hfhbd:
### Current Behavior

```stacktrace
java.util.NoSuchElementException: Collection contains no element matching the predicate.
at org.gradle.internal.declarativedsl.parsing.LightTreeUtilKt.expectSingleOfKind(LightTreeUtil.kt:305)
at org.gradle.internal.declarativedsl.parsing.GrammarToTree.toplevelNode(GrammarToTree.kt:722)
at org.gradle.internal.declarativedsl.parsing.GrammarToTree.scriptNodes(GrammarToTree.kt:707)
at org.gradle.internal.declarativedsl.parsing.GrammarToTree.script(GrammarToTree.kt:169)
at org.gradle.internal.declarativedsl.parsing.DefaultLanguageTreeBuilder.build(LanguageTreeBuilder.kt:18)
at org.gradle.internal.declarativedsl.evaluator.runner.AnalysisStepRunner.languageModelFromLightParser(AnalysisStepRunner.kt:124)
at org.gradle.internal.declarativedsl.evaluator.runner.AnalysisStepRunner.parseAndResolve(AnalysisStepRunner.kt:109)
at org.gradle.internal.declarativedsl.evaluator.runner.AbstractAnalysisStepRunner.runInterpretationSequenceStep(AnalysisStepRunner.kt:63)  
at org.gradle.internal.declarativedsl.evaluator.runner.AbstractAnalysisStepRunner.runInterpretationSequenceStep(AnalysisStepRunner.kt:50) 
at org.gradle.internal.declarativedsl.evaluator.conversion.AnalysisAndConversionStepRunner.runInterpretationSequenceStep(AnalysisAndConversionStepRunner.kt:68)
at org.gradle.internal.declarativedsl.evaluator.DefaultDeclarativeKotlinScriptEvaluator.runInterpretationSequence(DeclarativeKotlinScriptEvaluator.kt:100)
at org.gradle.internal.declarativedsl.evaluator.DefaultDeclarativeKotlinScriptEvaluator.evaluate(DeclarativeKotlinScriptEvaluator.kt:89)
```

### Expected Behavior

No error

### Context (optional)

I tried the new `org.gradle.kotlin.dsl.dcl=true` and it often, but not always fails in random projects. Maybe it is caused by a race-condition, because cleaning `project/.gradle/build` helps.

### Self-contained Reproducer Project

[...]

PR #37136, opened 2026-03-16
Various optimisations in the Kotlin DSL

eskatos:
This is the result of scrutiny of the Kotlin DSL subprojects, looking for inefficiencies.
There is no big change, just adjustments, mostly avoiding unnecessary allocations and I/O.
The focus is mainly on accessors generation.

Performance test suite shows no regression and 1 to 3% improvements on some scenarios.

Benchmark for IP (https://github.com/gradle/isolated-projects-testbed) using `resync_st_scr_change_pip_warm` scenario shows ~2% (mean) / ~5% (median) improvement on total sync time. Standard deviation is divided by two. JFR allocation profiling shows ~75MB less allocations in that same scenario with the reduction clearly coming from Kotlin DSL code.
[benchmark.html.zip](https://github.com/user-attachments/files/26019950/benchmark.zip)

One commit per change.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/kotlin-dsl-provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/precompiled/PrecompiledScriptPlugin.kt
+++ b/platforms/core-configuration/kotlin-dsl-provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/precompiled/PrecompiledScriptPlugin.kt
@@ -99,9 +99,8 @@ class PrecompiledScriptPluginFactory {
     }
 }
 
-@ConsistentCopyVisibility
 internal
-data class PrecompiledScriptPlugin internal constructor(
+class PrecompiledScriptPlugin internal constructor(
     internal val scriptFile: File,
     val scriptFileName: String,
     val scriptText: String,


jbartok:
I don't fully get the changes you did related to `KotlinLexer`, but be aware that it's not thread safe and we had issues in the past: https://github.com/gradle/gradle/issues/31479

Otherwise very nice!

eskatos:
I pulled `KotlinLexer` up to reuse it as it is relatively expensive to create and hold a bunch of memory. In some cases we were instantiating a lot of them unnecessarily. It's all in a single threaded context.

eskatos:
@bot-gradle test RFN

eskatos:
@bot-gradle test RFN

platforms/core-configuration/kotlin-dsl-provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/precompiled/PrecompiledScriptPlugin.kt
```diff
@@ -82,49 +97,36 @@ data class PrecompiledScriptPlugin(internal val scriptFile: File) {
...
-        get() = scriptTypeMatch.match.value
-
-    private
-    val scriptTypeMatch by lazy {
-        KotlinScriptTypeMatch.forName(scriptFileName)!!
-    }
-
+@ConsistentCopyVisibility
+internal
+data class PrecompiledScriptPlugin internal constructor(
```
jbartok: This class has a bunch of properties now, which it didn't use to have before. Since it's a data class, all those will be included in the generated `equals()`, `hashCode()` and such methods, so IF it goes into a collection or something, equality checks, for example, will be much more expensive. Maybe making it a regular class instead of a data class would be more appropriate.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/ClassNamesFromTypeStrings.kt
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+package org.gradle.kotlin.dsl.accessors
+
+import org.gradle.kotlin.dsl.internal.sharedruntime.codegen.primitiveKotlinTypeNames
+import org.gradle.kotlin.dsl.internal.sharedruntime.codegen.typeProjectionStrings
+
+
+internal
+class ClassNamesFromTypeStrings {
```
jbartok, line 24: This is unbounded, is that a good idea (considering that it can basically end up holding a cartesian product of class name combinations)? Maybe at least log a warning if the content becomes more than an obscene number? Or make it bounded by size?

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/ClassNamesFromTypeStrings.kt
```diff
@@ -0,0 +1,73 @@
...
+class ClassNamesFromTypeString(
+    val all: List<String>,
+    val leaves: List<String>
+)
+
+internal
+fun classNamesFromTypeString(typeString: String): ClassNamesFromTypeString {
+    val all = mutableListOf<String>()
+    val leafs = mutableListOf<String>()
+    val buffer = StringBuilder()
```
jbartok, line 42: Maybe this buffer could be a `ThreadLocal` instead? Maybe that way its reuse would be more useful?

eskatos: `ClassNamesFromTypeStrings` is only used in single threaded contexts.
The cost of a `ThreadLocal` would defeat the optimization.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/ClassNamesFromTypeStrings.kt
eskatos: This is short-lived, not a service. I would avoid adding complexity here until we see it is a problem.

platforms/core-configuration/kotlin-dsl-provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/precompiled/PrecompiledScriptPlugin.kt
eskatos: Good point! Fixed

PR #37145, opened 2026-03-16
Forbid Project.findProperty/property/hasProperty/getProperties in Iso…

### LINKED ISSUE #36998: Forbid Project.findProperty et al with Isolated Projects
There are a number of methods on `Project` that implicitly touch the mutable state of other projects:

- `project.findProperty`, `project.hasProperty` and `project.property`
- `project.properties`

Using these methods is already very likely to result in an IP-violation. However, today the violation will only be issued when the actual parent project access happens at runtime.

This hinders comprehensibility, because it's hard to reason about when and how the usage of these APIs makes plugins or build logic incompatible with IP.

These APIs are also practically harmless when used in the context of a root project. However, when the same code is copied in a subproject it stops working. Similarly, if these APIs were used in a plugin that was only applied to the root project, it appeared compatible. However, when the same plugin is applied to a subproject it stops being compatible.

One particular example is [an IP-compatibility regression in KGP](https://youtrack.jetbrains.com/issue/KT-81148/Publishing-helpers-in-KGP-are-incompatible-with-Isolated-Projects). Even though KGP is rigorously tested against IP, the majority of the tests are in single-project builds, making incompatibilities hard to detect without rewriting a large corpus of tests.

---

We should cleanly forbid any usage of these APIs. The IP-violations should be issued unconditionally, regardless of whether the methods are invoked on the root project or if the values happen to be found in the containers of the immediate project.

We should also consider marking these APIs in a static way to empower linters or IDEs to act on the information.

---

Related:
- https://github.com/gradle/gradle/issues/28857

reinsch82:
Fixes #36998.

When Isolated Projects is enabled, calling the explicit Project property API
(findProperty, property, hasProperty, getProperties/properties) from a build
script now reports an IP violation with deferred severity, causing the build
to fail after the configuration-cache store phase.

### Implementation

`ProjectScript.initDynamicLookupRoutine` creates a `DelegatingDynamicLookupRoutine`
wrapper around the build-scoped `TrackingDynamicLookupRoutine`. The wrapper intercepts:

- `invokeMethod()` for method names `"property"` and `"findProperty"` (script-level
  method calls like `property('name')`)
- `hasProperty()` and `getProperties()` (always explicit API calls)
- `property()` is **not** intercepted — it is shared with `BasicScript.getProperty()`
  for implicit property access (`version`, `tasks`, etc.)
- `findProperty()` is **not** intercepted — already covered by `invokeMethod()`

The violation reporter is created lazily (via `Suppliers.memoize`) by
`PropertyApiViolationReporterFactory` to avoid Kotlin class loading during
script init, which would interfere with `DynamicCallProblemReporting`.

`DefaultProject.getProperty()` calls `dynamicLookupRoutine.property()` directly
instead of delegating through `property()`, so normal Groovy property access
(`project.version`, `it.buildDir`) does not trigger violations.

### New classes

| Class | Module | Purpose |
|-------|--------|---------|
| `DelegatingDynamicLookupRoutine` | core | Java wrapper preserving varargs pass-through for `invokeMethod` |
| `PropertyApiViolationReporterFactory` | core | Build-scoped factory interface for violation reporters |
| `IsolatedProjectsPropertyApiViolationReporterFactory` | configuration-cache | IP-mode factory returning per-project reporters |
| `DefaultPropertyApiViolationReporterFactory` | configuration-cache | No-op factory for non-IP mode |
| `IsolatedProjectsPropertyApiViolationReporter` | configuration-cache | Kotlin reporter using `ProblemFactory` DSL |

### Tests

- `IsolatedProjectsAccessFromGroovyDslIntegrationTest`: new test for all four
  property API methods; updated parent-project lookup and chain test expectations
- `DynamicObjectIntegrationTest`: use `ext.has()`/`ext.get()` under IP mode

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

PR #37155, opened 2026-03-17
Generate issue list for patch release

ljacomet:
This automates the process of retrieving the issues fixed in a patch release and include them in release notes.

I confirmed this "works" for automating part of the work that went into
* #37148

Want to have it available, and possibly tweak it when using it for the next occasion.

### DIFF
--- a/build-logic/build-update-utils/src/main/kotlin/gradlebuild/buildutils/tasks/UpdateFixedIssuesInReleaseNotes.kt
+++ b/build-logic/build-update-utils/src/main/kotlin/gradlebuild/buildutils/tasks/UpdateFixedIssuesInReleaseNotes.kt
@@ -1,0 +1,140 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


build-logic/build-update-utils/src/main/kotlin/gradlebuild/buildutils/tasks/UpdateFixedIssuesInReleaseNotes.kt
```diff
@@ -0,0 +1,141 @@
...
+        }
+        val version = milestone.get()
+        val parts = version.split(".")
+        require(parts.size == 3 && parts[2].toIntOrNull()?.let { it != 0 } == true) {
+            "Version '$version' is not a patch release. Expected format: x.y.z with z > 0."
+        }
+        val issues = getFixedIssues()
+        if (issues.isEmpty()) {
+            println("No fixed issues found for milestone ${milestone.get()}")
+            return
```
blindpirate: If I read the code correctly, we should not return here, as L75 and L76 needs to use the empty list to clear the old content.

PR #37114, opened 2026-03-13
Make HttpScriptPluginIntegrationSpec compatible with CC and improve the error message for remote resource fetching with CC

ov7a:
### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

Issue #36177, opened 2026-01-02
Configuration Cache fails to restore project from included build if it is included on buildscript classpath


jvandort:
### Current Behavior

A configuration in one project depends on a single project from an included build. An artifact transform is applied to the configuration and used as a task input. The project from the included build is _also_ included on the root project's buildscript classpath.

With CC disabled, the transform is successfully executed. With CC enabled, the following failure is thrown:
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':toolingApiShadedJar'.
> Could not resolve all files for configuration ':runtimeClasspath'.
   > Failed to transform kotlin-dsl-shared-runtime.jar (project :kotlin-dsl-shared-runtime) to match attributes {artifactType=foo, org.gradle.category=library, org.gradle.dependency.bundling=external, org.gradle.jvm.version=17, org.gradle.libraryelements=jar, org.gradle.usage=java-runtime}.
      > project :kotlin-dsl-shared-runtime not found.

* Try:
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights from a Build Scan (powered by Develocity).
> Get more help at https://help.gradle.org.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':toolingApiShadedJar'.
	at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:38)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:210)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:205)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:67)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:60)
[...]

Issue #24273, opened 2023-03-09
Cannot query the contents of a file collection loaded from configuration cache when the collection runs transforms and is not declared as an input


adammurdoch:
Some state required to run transforms in this case is written to the configuration cache, but some state is missing leading to internal exceptions.

### Expected Behavior

Gradle should either support or deprecate this case, or both.

### Current Behavior

Gradle fails with various internal exceptions in this case.

See [this slack thread](https://gradle.slack.com/archives/CH43P3A10/p1678370149586499) for a reproducer.

### Your Environment

Build scan URL:

asodja:
Something similar happens only with this PR https://github.com/gradle/gradle/pull/29481.

When running:
`./gradlew :tooling-api:toolingApiShadedJar` from commit `0480d28ccf00c3636c7b06d2f4e2a4af665f6793` you get:
```
* What went wrong:
Execution failed for task ':tooling-api:toolingApiShadedJar'.
> Could not resolve all files for configuration ':tooling-api:jarsToShade'.
   > Failed to transform gradle-java-api-extractor-8.10.jar (project :build-logic:java-api-extractor) to match attributes {artifactType=buildReceipt, minified=true, org.gradle.category=library, org.gradle.dependency.bundling=external, org.gradle.jvm.version=8, org.gradle.libraryelements=jar, org.gradle.usage=java-runtime}.
      > project :build-logic:java-api-extractor not found.
``` 
Where project :build-logic:java-api-extractor is a project that is shared between build-logic and runtime logic. Build scan: https://ge.gradle.org/s/42wrsr5adeke4.

But when running without cc like:
```
./gradlew :tooling-api:toolingApiShadedJar --no-configuration-cache
```
build succeeds, see https://ge.gradle.org/s/zdtmnwdsrcble.

--- 
Explanation 
> Mikhail Lopatkin
I think we don't add tasks that complete at configuration time to the resulting work graph, and then prune projects that have no tasks in the graph. This looks like a consequence of the pruning.

A workaround was to directly depend to some task in java-api-extractor, see [5e783f135802752046464e64c6c5c2aba92ece26](https://github.com/gradle/gradle/pull/29481/commits/5e783f135802752046464e64c6c5c2aba92ece26).

See also this [Slack discussion](https://gradle.slack.com/archives/CH43P3A10/p1719850868313719).

tresat:
[...]

Issue #16179, opened 2021-02-16
Configuration cache entry potentially becomes broken if files from a file collection vanish


Vampire:
I have a task with a property like
```kt
@get:Classpath
protected val classPath: FileCollection by lazy {
    project.configurations.detachedConfiguration(
        project.dependencies.create("com.sun.xml.ws:jaxws-tools:2.3.3")
    )
}
```

If I execute the task with configuration cache, the configuration is resolved and the list of files stored to the configuration cache.
If I rerun the task, everything is fine and working too.
But if now one of the files vanishes, (for example I deleted cache files to get some more free disk space, or the dependency comes from a file-based repository that is not stored to the Gradle cache and files vanish, or whatever other reason that could exist) and you try to rerun the task, then the file is missing and the task fails to work properly.

One could say that files in the Gradle cache should not be touched manually,
but usually my PoV is, if it is a cache it should be safe to being deleted and recreated when needed and missing,
it is a cache after all.

So maybe the configuration cache entry should also store whether the files in the collection actually exist or similar.
Or there should be a way to have the resolution be done at execution time instead of configuration time when configuration cache is used. The above configuration when used without configuration cache is only resolved at execution time and only when actually needed. With configuration cache it is resolved "eagerly" and the result persisted to the cache.

ov7a:
Sorry for the late reply.

The issue is in the backlog of the relevant team and is prioritized by them.

PR #37159, opened 2026-03-17
Fix remaining dependency management tests for Configuration Cache

tresat:
See commit-by-commit changes which do one class at a time.

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/CachedMissingModulesIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/CachedMissingModulesIntegrationTest.groovy
@@ -541,7 +541,7 @@ Required by:
         when:
         server.resetExpectations()
         if (!GradleContextualExecuter.configCache) {
-            // We should not need to do this.  When vintage (non-CC) mode is deprecated and removed, this if will no longer be necessary
+            // We should not need to do this.  When vintage (non-CC) mode is deprecated and removed, this condition will no longer be necessary
             repo1Module.pom.expectHeadMissing()
         }
 


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformCachingIntegrationTest.groovy
```diff
@@ -1000,7 +1000,7 @@ class ArtifactTransformCachingIntegrationTest extends AbstractHttpDependencyReso
         (outputDirs.parentFile.name as Set).size() == 2
     }
 
-    @ToBeFixedForConfigurationCache(because = "project :lib-project:producer not found.")
+    @UnsupportedWithConfigurationCache(because = "Under CC, project references from included builds are not found when transforms are resolved from the configuration cache entry")
```
ov7a, line 1003: 🤔 Do we provide a good error message in this scenario?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/UndeclaredDependencyResolutionIntegrationTest.groovy
```diff
@@ -17,12 +17,11 @@
 package org.gradle.integtests.resolve.transform
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
-import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
 import org.gradle.integtests.fixtures.UnsupportedWithConfigurationCache
 
 // This tests current behaviour, not desired behaviour
 class UndeclaredDependencyResolutionIntegrationTest extends AbstractIntegrationSpec implements ArtifactTransformTestFixture {
-    @ToBeFixedForConfigurationCache(because = "Transform nodes are not serialized when transform of project artifact is not declared as a dependency of another node")
+    @UnsupportedWithConfigurationCache(because = "under CC, transform nodes for project artifacts are not serialized when the transform is not declared as a dependency of the task, causing 'project not found' errors during cache replay")
```
ov7a: 🤔 Should the error be improved?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/PublishAndResolveIntegrationTest.groovy
```diff
@@ -52,11 +51,16 @@ class PublishAndResolveIntegrationTest extends AbstractDependencyResolutionTest
         """
 
         expect:
+        if (GradleContextualExecuter.configCache) {
+            // Under CC, resolution occurs during cache entry writing before tasks execute.
+            // Pre-publish so the artifact exists when CC serializes the resolve task's inputs.
+            // Can't just rely on the dependsOn to run publish first - waiting until task execution time is too late.
+            succeeds ':publish'
```
ov7a: 🤔 Should we verify that publishing still succeeds afterwards?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/TaskDependencyResolutionEventsIntegrationTest.groovy
```diff
@@ -16,11 +16,11 @@
 package org.gradle.integtests.resolve
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
-import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
+import org.gradle.integtests.fixtures.UnsupportedWithConfigurationCache
 
 class TaskDependencyResolutionEventsIntegrationTest extends AbstractIntegrationSpec {
 
-    @ToBeFixedForConfigurationCache(because = "broken file collection")
+    @UnsupportedWithConfigurationCache(because = "under CC, the configuration is resolved during cache entry writing, which triggers the beforeResolve/afterResolve callbacks that this test expects not to fire during task dependency calculation")
```
ov7a, line 23: 🤔 Do we have a good error explaining that?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/ArtifactCacheUnusedEntryCleanupIntegrationTest.groovy
```diff
@@ -255,7 +255,8 @@ class ArtifactCacheUnusedEntryCleanupIntegrationTest extends AbstractHttpDepende
         files[1].assertExists()
     }
 
-    @ToBeFixedForConfigurationCache(because = "does not re-download missing artifacts")
+
+    @UnsupportedWithConfigurationCache(because = "re-download of deleted artifacts requires dependency resolution, which is skipped when configuration cache entry is reused")
     def "downloads deleted files again when they are referenced"() {
```
mlopatkin, line 263: :x: This should remain ToBeFixedForCC, as the behavior is broken. Let's link it to https://github.com/gradle/gradle/issues/16179. We then mark the tests as blocked in the spreadsheet.

Same applies to all tests in the file.

```suggestion
    @ToBeFixedForConfigurationCache(because = "https://github.com/gradle/gradle/issues/16179")
def "downloads deleted files again when they are referenced"() {
```

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformCachingIntegrationTest.groovy
mlopatkin, line 1003: :x: This looks like a legitimate bug. The https://github.com/gradle/gradle/issues/36177 is very close to what I'm seeing here, but maybe filing a new specific one would be better. But it should remain TBFCC for now.

ov7a: is it
* https://github.com/gradle/gradle/issues/24273 ?

mlopatkin: > is it
> 
> * Cannot query the contents of a file collection loaded from configuration cache when the collection runs transforms and is not declared as an input?

It's close, but there is an `@InputFiles` dependency on the file collection, and adding an explicit input doesn't help either. So this doesn't fit the description fully.

tresat: https://github.com/gradle/gradle/issues/37202 added

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/TaskDependencyResolutionEventsIntegrationTest.groovy
tresat: I don't understand what we'd need an error for in this case.  Nothing is going wrong, it's just the CC serialization triggers the callbacks earlier, which makes this test not useful.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/UndeclaredDependencyResolutionIntegrationTest.groovy
tresat: I think instead of improving the error, we should ban using transforms like this.  The dependency should be declared.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/PublishAndResolveIntegrationTest.groovy
tresat: If the publish task succeeded, isn't that enough?  For non-CC, we were not even checking that the publish task ran, just counting on the task dependency to make this work.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/UndeclaredDependencyResolutionIntegrationTest.groovy
ov7a: Then, I think an issue should be opened and it should be referenced from the annotation

tresat: Done.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/UndeclaredDependencyResolutionIntegrationTest.groovy
```diff
@@ -18,10 +18,12 @@ package org.gradle.integtests.resolve.transform
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.integtests.fixtures.UnsupportedWithConfigurationCache
+import spock.lang.Issue
 
 // This tests current behaviour, not desired behaviour
 class UndeclaredDependencyResolutionIntegrationTest extends AbstractIntegrationSpec implements ArtifactTransformTestFixture {
     @UnsupportedWithConfigurationCache(because = "under CC, transform nodes for project artifacts are not serialized when the transform is not declared as a dependency of the task, causing 'project not found' errors during cache replay")
+    @Issue("https://github.com/gradle/gradle/issues/37219")
```
mlopatkin, line 27: :x: Please keep `@ToBeFixedForConfigurationCache`. It is much more effective than `@Unsupported` because the test starts to fail when we actually fix the thing, so there is more incentive to update the test.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/ArtifactCacheUnusedEntryCleanupIntegrationTest.groovy
mlopatkin: Looks like this comment got lost?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformCachingIntegrationTest.groovy
mlopatkin: ❓ By looking at the diff, the test seems to be fixed by the work done in #36177. Could you please check if the test actually passes with CC and remove the suppression if so?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/ArtifactCacheUnusedEntryCleanupIntegrationTest.groovy
```diff
@@ -293,7 +294,8 @@ class ArtifactCacheUnusedEntryCleanupIntegrationTest extends AbstractHttpDepende
         journal.assertExists()
     }
 
-    @ToBeFixedForConfigurationCache(because = "does not re-download missing artifacts")
+
+    @UnsupportedWithConfigurationCache(because = "re-download of deleted resources requires dependency resolution, which is skipped when configuration cache entry is reused")
```
mlopatkin: ❓ It looks like the cache failure is about mismatch in the server expectation because of an extra `HEAD` request CC issues.

Maybe something like https://github.com/gradle/gradle/pull/37114/changes#diff-94d1de19fc9e75764ccd9f6ba89a8fd1964a8ad99183752107daba848e348a40R183 can work there too?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/ArtifactCacheUnusedEntryCleanupIntegrationTest.groovy
mlopatkin: ❓ Can this test be salvaged by moving the resource out of `doLast`, like:

```suggestion
           task uriText {
               def res = resources.text.fromUri("${server.uri}/$uniqueFileName")
               doLast {
                   print res.asString()
               }
           }
```

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/ArtifactCacheUnusedEntryCleanupIntegrationTest.groovy
```diff
@@ -356,7 +359,8 @@ class ArtifactCacheUnusedEntryCleanupIntegrationTest extends AbstractHttpDepende
         resource.assertExists()
     }
 
-    @ToBeFixedForConfigurationCache(because = "does not re-download missing artifacts")
+
+    @UnsupportedWithConfigurationCache(because = "re-download of deleted artifacts requires dependency resolution, which is skipped when configuration cache entry is reused")
     def "redownloads deleted artifacts for artifact query"() {
```
mlopatkin, line 369: ❌ This should have the same treatment (`TBFCC` linking to #16179) as in https://github.com/gradle/gradle/pull/37159/changes#r2948868423

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/RecoverFromBrokenResolutionIntegrationTest.groovy
mlopatkin, line 1: 🤔 I don't see any changes that can make these tests pass...

tresat: I don't understand why these were passing either.  Re-investigating.

tresat: I think this might have been reverted accidentally in a force push.  I remember fixing these already.  Regardless, they are re-fixed.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/PublishAndResolveIntegrationTest.groovy
```diff
@@ -88,11 +92,13 @@ class PublishAndResolveIntegrationTest extends AbstractDependencyResolutionTest
         """
 
         expect:
+        if (GradleContextualExecuter.configCache) {
```
jbartok: This way of fixing, the explicit call of the publish task doesn't really match the name of the test, specifically the "dependent task" bit. Or how should I phrase it... If task dependency and things it triggers is important, then we can't call tasks explicitly. Maybe it would be a better fix to have a task dependency based test case which just runs in vintage mode and then have another test case, where we call publish explicitly and that use case can run both in vintage & cc mode? Wdyt?

jbartok: Btw, I'm not talking about this specific test method, but all similar ones in this class.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/CachedMissingModulesIntegrationTest.groovy
```diff
@@ -538,15 +540,16 @@ Required by:
 
         when:
         server.resetExpectations()
-        // TODO - should not need to do this
-        repo1Module.pom.expectHeadMissing()
+        if (!GradleContextualExecuter.configCache) {
+            // TODO - should not need to do this
```
jbartok: Maybe just keep it a simple comment (as opposed to a TODO), removing vintage mode completely will automatically get rid of the problem.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/RecoverFromBrokenResolutionIntegrationTest.groovy
```diff
@@ -70,7 +71,11 @@ class RecoverFromBrokenResolutionIntegrationTest extends AbstractHttpDependencyR
         fails 'retrieve'
 
         and:
-        failure.assertHasDescription("Execution failed for task ':retrieve' (registered in build file 'build.gradle').")
+        if (GradleContextualExecuter.isConfigCache()) {
+            failureDescriptionStartsWith("Configuration cache state could not be cached")
```
jbartok: Maybe extend this prefix a bit to show that for example the ":retrieve" task name is also there, doesn't get lost in the CC form of the failure.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/CachedMissingModulesIntegrationTest.groovy
tresat: This comment was pre-existing, but you're right.  I'll update it.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/PublishAndResolveIntegrationTest.groovy
tresat: You're right, just making the task run isn't what this is meant to test.

The problem is that serializing Copy is trying to find the files from Publish, and _can't_ (as they don't exist, since that task hasn't run yet, since we're not even in the Task Execution phase).

I've just reverted changes to these tests and adding an explanation to the `@ToBeFixed`.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/CachedMissingModulesIntegrationTest.groovy
```diff
@@ -541,7 +541,7 @@ Required by:
         when:
         server.resetExpectations()
         if (!GradleContextualExecuter.configCache) {
-            // TODO - should not need to do this
+            // We should not need to do this.  When vintage (non-CC) mode is deprecated and removed, this if will no longer be necessary
```
jbartok: I read the "this if" part 5 times until I realized it's not a mistake. I would say use "condition" instead of "if".

jbartok (approved):
Small nitpic, looks good (sorry for the delay).

PR #37162, opened 2026-03-17
[Investigation] Implement ListProperty/SetProperty += and -=

asodja:


asodja:
@bot-gradle test this

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/AbstractCollectionProperty.java
```diff
@@ -608,4 +610,70 @@ public void replace(Transformer<? extends @Nullable Provider<? extends Iterable<
...
+            }
+            Set<T> toRemove = new HashSet<>(rightValue.getWithoutSideEffect());
+            ImmutableCollection.Builder<T> builder = getCollectionFactory().get();
+            for (T element : leftValue.getWithoutSideEffect()) {
+                if (!toRemove.contains(element)) {
+                    builder.add(element);
+                }
+            }
+            C result = Cast.uncheckedNonnullCast(builder.build());
+            return Value.of(result).withSideEffect(SideEffect.fixedFrom(leftValue));
```
abstratt, line 670: ❓ 

Don't we need to include only the side effects from the elements from `leftValue` that are actually in `result`?

Issue #24948, opened 2023-05-01
Dependency Verifcation adds/removes `oss-parent` key non-deterministically


yogurtearl:
### Expected Behavior

When you rerun gradle with `--refresh-keys --write-verification-metadata pgp,sha512 --export-keys` multiple times in a row it sometimes adds in `oss-parent` and sometimes it doesn't. 

Re-running ` --write-verification-metadata` with the exact same input multiple time should always give the exact same output. 

### Current Behavior

Gradle non-deterministically adds/removes this 👇  from the `gradle/verification-metadata.xml` file when generated multiple times with the same inputs.

It seems to only happen for a very small list of deps that are not used directly, but presumably pulled in by gradle itself? 

```
<trusted-key id="34441e504a937f43eb0daef96a65176a0fb1cd0b" group="org.codehaus.groovy" name="groovy-bom" version="3.0.14"/>
<trusted-key id="44fbdbbc1a00fe414f1c1873586654072ead6677" group="org.sonatype.oss" name="oss-parent" version="9"/>
```

### Context (optional)

I think we started seeing this in Gradle 7.6.1

### Steps to Reproduce

1. use a blank gradle build (with project replicator or from `init` plugin) 
2. add `resolveAndLockall` task [from the docs](https://docs.gradle.org/current/userguide/dependency_locking.html#lock_all_configurations_in_one_build_execution)
3. `rm gradle/verification-metadata.xml` 
4. run this command:
    ```
    ./gradlew :resolveAndLockAll --no-configuration-cache --refresh-keys --write-locks --write-verification-metadata pgp,sha512 --export-keys --stacktrace
    ```
5. run steps 3 and 4 multiple times comparing the output of `gradle/verification-metadata.xml` each time. Notice that the contents of `gradle/verification-metadata.xml` are not always consistent. 

It should always return the same result for the same inputs. 

### Gradle version

8.0.2

### Your Environment (optional)

linux, JDK 17, ephemeral CI

jbartok:
Thank you for providing a valid reproducer.

The issue is in the backlog of the relevant team, but this area of Gradle is currently not a focus one, so it might take a while before a fix is made.

mathjeff:
As a user, this reminds me of https://github.com/gradle/gradle/issues/20194 where we found parent poms were excluded from our dependency-verification.xml too

yogurtearl:
@mathjeff yeah, I think it is related... 

[...]

Issue #20194, opened 2022-03-16
--write-verification-metadata not adding entries for parent poms


mathjeff:
### Expected Behavior
We'd like `./gradlew --write-verification-metadata ${taskThatResolvesArtifacts} -m` to generate entries in `gradle/verification-metadata.xml` for all signed artifacts, even if they're just parent poms

### Current Behavior
We're observing when running `./gradlew --write-verification-metadata` that it isn't adding entries into `gradle/verification-metadata.xml` for the keys used to sign parent poms

For example, https://android-review.googlesource.com/c/platform/frameworks/support/+/2024786/8 was generated by removing some entries from `gradle/verification-metadata.xml` and running https://android-review.googlesource.com/c/platform/frameworks/support/+/2019456/9, but it failed at https://android-build.googleplex.com/builds/pending/P31485959/androidx_device_tests/latest/view/logs/build_error.log with:

```
A problem occurred evaluating script.
> Dependency verification failed for configuration ':buildSrc:plugins:cacheableApi'
  6 artifacts failed verification:
    - apache-15.pom (org.apache:apache:15) from repository maven2
    - apache-18.pom (org.apache:apache:18) from repository maven2
    - apache-9.pom (org.apache:apache:9) from repository maven2
    - mojo-parent-40.pom (org.codehaus.mojo:mojo-parent:40) from repository maven2
    - tomlj-1.0.0.jar (org.tomlj:tomlj:1.0.0) from repository maven2
    - tomlj-1.0.0.pom (org.tomlj:tomlj:1.0.0) from repository maven2
  This can indicate that a dependency has been compromised. Please carefully verify the signatures and checksums. Key servers are disabled, this can indicate that you need to update the local keyring with the missing keys.
```

The .pom file of each of these dependencies appear to specify `<packaging>pom</packaging>` (even tomlj-1.0.0.pom)

### Context
This makes it more difficult for us to remove the 16-character versions of some keys from `gradle/verification-metadata` when Gradle adds the 40-character versions, see https://github.com/gradle/gradle/issues/20193

mathjeff:
Also related, I observed that Gradle didn't add an entry for Gradle Enterprise either

From https://android-build.googleplex.com/builds/pending/P31484702/androidx_device_tests/latest/view/logs/build_error.log :


```
FAILURE: Build failed with an exception.

* What went wrong:
Dependency verification failed for configuration 'classpath':
[...]

PR #37165, opened 2026-03-17
Check for metadata file existence on deserialize  

jvandort:
When we load a binary metadata file from cache, we also provide the location of the original
source file that produced the binary metadata file. It is possible that this original source
file is cleared from the cache while the binary file is not, meaning it is possible to load
the binary metadta from the cache even when the original source file does not exist anymore.

During dependency verification, we need this original source file. Previously, we checked for
the existence of this original source file each time we consumed the metadata from the repository. This can be a massive number of times per build, and we likely consume the same metadata file more than once. Instead, we only check for existence once when we load the binary file from the cache.

This change saves 2.5-3% on `gradle/gradle sanityCheck --dry-run`
image

Before:
image

After:
image

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/modulecache/PersistentModuleMetadataCache.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/modulecache/PersistentModuleMetadataCache.java
@@ -82,6 +82,10 @@ public class PersistentModuleMetadataCache extends AbstractModuleMetadataCache {
             if (entry.isMissing()) {
                 return new DefaultCachedMetadata(entry, null, timeProvider);
             }
+
+            // TODO: `getModuleDescriptor` performs IO and blocks for a non-negligible amount of time.
+            // We should avoid holding the `useCache` lock for this duration, as we prevent others from
+            // using this cache, even for different keys.
             MutableModuleComponentResolveMetadata metadata = moduleMetadataStore.getModuleDescriptor(key);
             if (metadata == null) {
                 // Descriptor file has been deleted - ignore the entry


jvandort:
@bot-gradle test this

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/metadata/DefaultMetadataFileSourceCodec.java
```diff
@@ -72,10 +72,18 @@ private DefaultMetadataFileSource createSource(byte[] sha1, String group, String
         ModuleComponentArtifactIdentifier artifactId = createArtifactId(group, module, version, name);
         HashCode hashCode = HashCode.fromBytes(sha1);
         File metadataFile = fileStore.whereIs(artifactId, hashCode.toString());
+        if (!metadataFile.exists()) {
```
ljacomet, line 75: ❓ What would it take to make sure the file is here, including its hierarchy for POM files?

This would make dependency verification with metadata verification more consistent.

jvandort: We should probably not do it here, but would attach multiple MetadataFileSource objects to this one metadata object. 

From the em engines perspective, we don't even know the parent exists so atm we wouldn't check for it. So we'd also need to include information in the metadata listing all the parents, so we can iterate over them and verify each file.

ljacomet: That makes sense and my question was a kind of scope creep anyway. 

Related:
* #20194
* #24948

PR #37168, opened 2026-03-18
Optimize slow queries

blindpirate:
### Context

Fixes https://github.com/gradle/gradle-private/issues/4986

According to the profiling log, history queries in `CrossVersionResultsStore` and `BaseCrossBuildResultsStore` could be as slow as 70s. SQL explain showed MySQL choosing poor plans for the old SQL:

```
SELECT id, startTime, ...
FROM testExecution
WHERE testClass = ? AND testId = ? AND testProject = ?
  AND startTime >= ?
  AND (channel LIKE ? OR channel LIKE ? OR teamCityBuildId IN (?, ?))
ORDER BY startTime DESC
LIMIT ?;
```

It scans by STARTTIME and filtering 170k+ rows instead of using composite indexes.

### Solution
Replace the single OR-based filter with a `UNION DISTINCT` of two branches: one for channel patterns and one for teamCityBuildId. Each branch can use its index, avoiding the slow full scan. Then we perform an `ORDER BY` on the subquery result.

```
SELECT h.id, h.startTime, ...
FROM (
  SELECT id, startTime, ...
  FROM testExecution
  WHERE testClass = ? AND testId = ? AND testProject = ?
    AND startTime >= ?
    AND (channel LIKE ? OR channel LIKE ?)
  UNION DISTINCT
  SELECT id, startTime, ...
  FROM testExecution
  WHERE testClass = ? AND testId = ? AND testProject = ?
    AND startTime >= ?
    AND teamCityBuildId IN (?, ?)
) AS h
ORDER BY h.startTime DESC
LIMIT ?;
```

Profiling result shows that all queries are <500ms now.

### DIFF
--- a/testing/internal-performance-testing/src/main/groovy/org/gradle/performance/results/AbstractWritableResultsStore.groovy
+++ b/testing/internal-performance-testing/src/main/groovy/org/gradle/performance/results/AbstractWritableResultsStore.groovy
@@ -19,6 +19,8 @@ package org.gradle.performance.results
 import com.google.common.collect.ImmutableMap
 import groovy.transform.CompileStatic
 
+import java.sql.PreparedStatement
+import java.sql.SQLException
 import java.sql.Timestamp
 import java.time.LocalDateTime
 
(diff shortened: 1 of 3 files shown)


testing/internal-performance-testing/src/main/groovy/org/gradle/performance/fixture/CrossVersionPerformanceTestRunner.groovy
```diff
@@ -217,8 +217,8 @@ class CrossVersionPerformanceTestRunner extends PerformanceTestSpec {
         def builder = GradleBuildExperimentSpec.builder()
             .projectName(testProject)
             .displayName(displayName)
-            .warmUpCount(warmUpRuns)
-            .invocationCount(runs)
+            .warmUpCount(1)
```
blindpirate: This is for faster test, will revert before merging.

blindpirate:
@bot-gradle test PT

blindpirate:
@bot-gradle squash

PR #37172, opened 2026-03-18
Mark cc-compatibility for some artifact transforms tests

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
@@ -19,7 +19,6 @@ package org.gradle.integtests.resolve.transform
 import org.gradle.integtests.fixtures.AbstractHttpDependencyResolutionTest
 import org.gradle.integtests.fixtures.BuildOperationsFixture
 import org.gradle.integtests.fixtures.ToBeFixedForConfigurationCache
-import org.gradle.integtests.fixtures.UnsupportedWithConfigurationCache
 import org.gradle.integtests.fixtures.resolve.ResolveTestFixture
 import org.gradle.internal.file.FileType
 import org.gradle.operations.dependencies.transforms.ExecutePlannedTransformStepBuildOperationType
@@ -1706,7 +1705,7 @@ Found the following transformation chains:
         outputContains("files: [b.jar]")
     }
 
-    @UnsupportedWithConfigurationCache(because = "Resolution happens during configuration time, so the transform is not triggered")
+    @ToBeFixedForConfigurationCache(because = "Resolution happens during configuration time, so the transform is not triggered. Also, lenient is not respected https://github.com/gradle/gradle/issues/37420")
     def "user gets a reasonable error message when a transform input cannot be downloaded and proceeds with other inputs"() {
         def m1 = ivyHttpRepo.module("test", "test", "1.3")
             .artifact(type: 'jar', name: 'test-api')
@@ -2118,7 +2117,7 @@ Found the following transformation chains:
         failure.assertHasCause("broken")
     }
 
-    @UnsupportedWithConfigurationCache(because = "Resolution happens during configuration time, so the transform is not triggered")
+    @ToBeFixedForConfigurationCache(because = "Resolution happens during configuration time, so the transform is not triggered, Also, lenient is not respected https://github.com/gradle/gradle/issues/37420")
     def "collects multiple failures"() {
         def m1 = mavenHttpRepo.module("test", "a", "1.3").publish()
         def m2 = mavenHttpRepo.module("test", "broken", "2.0").publish()


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformWithDependenciesIntegrationTest.groovy
```diff
@@ -1357,21 +1359,27 @@ abstract class ClasspathTransform implements TransformAction<TransformParameters
...
-            transformStep1('common.jar'),
-            transformStep1('slf4j-api-1.7.25.jar'),
-            transformStep1('hamcrest-core-1.3.jar'),
-            transformStep1('junit-4.11.jar': ['hamcrest-core-1.3.jar']),
-            transformStep2('common.jar'),
-            transformStep2('slf4j-api-1.7.25.jar'),
-            transformStep2('hamcrest-core-1.3.jar'),
-            transformStep2('junit-4.11.jar': ['hamcrest-core-1.3.jar'])
-        )
+        if (GradleContextualExecuter.notConfigCache) {
```
ov7a, line 1371: 🤔 the test name implies that it's the expected behavior

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
```diff
@@ -1705,7 +1706,7 @@ Found the following transformation chains:
         outputContains("files: [b.jar]")
     }
 
-    @ToBeFixedForConfigurationCache(because = "treating file collection visit failures as a configuration cache problem adds an additional failure to the build summary; exception chain is different when transform input cannot be resolved")
+    @UnsupportedWithConfigurationCache(because = "Resolution happens during configuration time, so the transform is not triggered")
```
tresat: ❗ Doesn't lenient resolution need to be supported with CC?  Shouldn't these tests remain `@ToBeFixed`?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformParallelIntegrationTest.groovy
```diff
@@ -273,7 +273,7 @@ class ArtifactTransformParallelIntegrationTest extends AbstractDependencyResolut
         outputContains("Transforming b.jar to b.jar.txt")
     }
 
-    @ToBeFixedForConfigurationCache(because = "Files are downloaded when cache entry is written")
+    @UnsupportedWithConfigurationCache(because = "Files are downloaded when cache entry is written")
     def "files are transformed as soon as they are downloaded"() {
```
tresat: ❗ This seems like behavior we still want to support and test (that files are transformed as soon as they are downloaded).  We just need a CC-compatible way to test it.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformWithDependenciesIntegrationTest.groovy
tresat: ❓ Why is this not also the case with CC?  The resolution failed, how could the transforms have been executed?

ov7a: Ah, the test is unclear.

Transforms are still executed for the app project, but not executed for lib.jar. I'll add a comment

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformParallelIntegrationTest.groovy
ov7a: 🤔 Is there a solid way to test this?

tresat: Perhaps have the transform write something when it is performed, then check that the write happened prior to task execution (when running with CC).

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
ov7a: The error is somewhat reasonable already
```
Configuration cache state could not be cached: field `elements` of `org.gradle.internal.serialize.codecs.core.ResolutionBackedFileCollectionSpec` bean found in field `mainSpec` of task `:resolve` of type `org.gradle.api.tasks.Copy`: error writing value of type 'java.util.ArrayList'
> Could not download test-impl-1.3.jar (test:test:1.3)
   > Could not get resource 'http://127.0.0.1:56928/repo/test/test/1.3/test-impl-1.3.jar'.
      > Could not GET 'http://127.0.0.1:56928/repo/test/test/1.3/test-impl-1.3.jar'. Received status code 500 from server: broken

```
But it happens during a different time. I'm pretty sure it's covered in another place

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformParallelIntegrationTest.groovy
ov7a: After trying out stuff, I came to the realization that it's not possible:
- https://github.com/gradle/gradle/issues/37371

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformParallelIntegrationTest.groovy
```diff
@@ -81,7 +81,7 @@ class ArtifactTransformParallelIntegrationTest extends AbstractDependencyResolut
 
                         println "[" + currentTime + "] Running transform for " + input.name
 
-                        ${server.callFromBuildUsingExpression("input.name")}
+                        ${server.callFromBuildUsingExpression("\"transform-call-\" + input.name")}
```
ov7a, line 84: It was very hard to read tests without that

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
tresat: The error message is reasonable.  What is not reasonable is that CC stops and doesn't honor the contract of lenient:

```
/**
         * Specify if the view should be resolved in a 'lenient' fashion.
         *
         * When set to <code>true</code>, this view will resolve as many artifacts and/or files as possible
         * collecting any failures.
         *
         * When set to <code>false</code>, any failures will be propagated as exceptions when the view is resolved.
         */
        void setLenient(boolean lenient);
```  

CC should have tried to transform the other jars, even though the overall resolution fails.

ov7a: Ah, I get it now. Thanks!

I've created an issue
* https://github.com/gradle/gradle/issues/37420

PR #37173, opened 2026-03-18
fix: adjust timeout in test

reinsch82:
the timeout length doesn't influence the runtime length of the test, because the task run is intended to be aborted early.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/core-runtime/logging/src/integTest/groovy/org/gradle/internal/logging/console/TaskbarProgressResetFunctionalTest.groovy
```diff
@@ -63,7 +61,7 @@ class TaskbarProgressResetFunctionalTest extends AbstractIntegrationSpec {
         buildFile << """
             task block {
                 doFirst {
-                    Thread.sleep(10_000)
+                    Thread.sleep(60_000)
```
ljacomet, line 64: ❓ Is there a reason you can't leverage the blocking server for this?
It would remove any non-determinism from the test.

PR #37174, opened 2026-03-18
Move some process api classes back to core-api

ghale:
The provider API work creates a dependency on providers in these classes.  Providers are too deeply entangled with core-api to extract right now.  So, moving these classes back until we can disentangle providers from core-api.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

PR #37182, opened 2026-03-18
Make some tests in core CC-compatible

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/integTest/groovy/org/gradle/configuration/project/LifecycleProjectEvaluatorIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/configuration/project/LifecycleProjectEvaluatorIntegrationTest.groovy
@@ -19,6 +19,7 @@ package org.gradle.configuration.project
 import org.gradle.execution.taskgraph.NotifyTaskGraphWhenReadyBuildOperationType
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.integtests.fixtures.BuildOperationsFixture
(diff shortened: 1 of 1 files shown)


ov7a:
@bot-gradle test this

subprojects/core/src/integTest/groovy/org/gradle/internal/operations/logging/LoggingBuildOperationProgressIntegTest.groovy
```diff
@@ -371,10 +370,23 @@ class LoggingBuildOperationProgressIntegTest extends AbstractIntegrationSpec {
...
+            }
+
+            def cleanup = gradle.sharedServices.registerIfAbsent("listenerCleanup", ListenerCleanupService) {}
+            cleanup.get().with {
+                it.manager = manager
+                it.listener = listener
+            }
+
+            tasks.register("t") {
+                usesService(cleanup)
```
tresat, line 388: 🤔 Is this cleanup service even necessary?  Why do we care if the listeners aren't removed?

ov7a: It affects the daemon and other tests. (This was my first approach)

ov7a: It may be fixed with an isolated daemon, but I don't think it's worth it.

PR #37210, opened 2026-03-20
Extract hashing implementation classes from core to new hashing-services module

ghale:
Note that there is an existing `hashing` project, but this is currently a published library.  We can't move these classes in there because they would drag new dependencies that would then need to be published, so we use a new project for this.

There are some service registrations that should ideally move with these, too, but they are still too entangled with other stuff in core.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

PR #37213, opened 2026-03-20
Add `getBoolean` to `InternalOptions` for null-safe boolean option access

alllex:
The existing `getOption().get()` pattern for `InternalFlag` options silently unboxes `Boolean` to `boolean`, risking an NPE with no diagnostic context. The new `getBoolean(InternalFlag)` default method adds an explicit null check with a clear error message and returns a primitive `boolean` directly.

All callsites reading `InternalFlag` values via `.getOption(...).get()` are migrated to use `getBoolean(...)`, including the Kotlin `getInternalFlag` extension function and the operator overload in `BuildModelParametersProvider`.


alllex:
Superseded by https://github.com/gradle/gradle/pull/37222

PR #37221, opened 2026-03-20
Split out :process-services-base

jvandort:
Split :process-services into :process-services-base, pulling out all classes related to `ClientExecHandleBuilderFactory`. This is the most internal infrastructure for starting processes, and doesn't pull in any `:core` or `:core-api` dependencies like `FileCollection`, `ForkOptions`, etc. 

The intention of this project is for use in the CLI and TAPI client, so we can break it off from the daemon dependency chain. Client code should not depend on core API classes, as they have a different lifecycle (restricted by deprecation cycles for the API) and runtime constraints (daemon code can compile to Java 17, client has more restrictions)

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

PR #37222, opened 2026-03-20
Add conveniences and reduce public types of InternalOptions API

alllex:
Make `InternalFlag`, `IntegerInternalOption`, and `StringInternalOption` package-private so that all interaction with internal options goes through the `InternalOptions` interface.

Callers now declare options via factory methods (`InternalOptions.ofBoolean`, `ofInt`, `ofString`, `ofStringOrNull`) and read values via typed accessors (`getBoolean`, `getInt`, `getString`, `getStringOrNull`, `getValue`) instead of constructing concrete subtypes and calling `getOptionValue(...).get()`.

Users of the interface in Java now get the same conveniences as previously only the Kotlin code had.

### DIFF
--- a/platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationTrace.java
+++ b/platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationTrace.java
@@ -186,7 +186,7 @@ public class BuildOperationTrace implements Stoppable {
 
     @Nullable
     private static Path resolveBasePath(InternalOptions internalOptions, File userActionRootDir) {
-        String basePath = internalOptions.getOptionValue(TRACE_OPTION).get();
+        String basePath = internalOptions.getValueOrNull(TRACE_OPTION);
         if (basePath == null || basePath.equals("false")) {
             return null;
         }
@@ -197,7 +197,7 @@ public class BuildOperationTrace implements Stoppable {
 
     @Nullable
     private static Set<String> getFilter(InternalOptions internalOptions) {
-        String filterProperty = internalOptions.getOptionValue(FILTER_OPTION).get();
+        String filterProperty = internalOptions.getValueOrNull(FILTER_OPTION);
         if (filterProperty == null) {
             return null;
         }
--- a/platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/InternalOptions.java
+++ b/platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/InternalOptions.java
@@ -42,6 +42,16 @@ public interface InternalOptions {
      * Resolves non-null value for the given option.
      */
     default <T> T getValue(InternalOption<T> option) {
+        T value = getOptionValue(option).get();
+        // The compiler ensures `InternalOption<@Nullable Type>` can't be passed, but the IDE can't see it and complains
+        assert value != null : "Option '" + option.getPropertyName() + "' is expected to have a non-null value, but was null";
+        return value;
+    }
+
+    /**
+     * Resolves nullable-value for the given option.
+     */
+    default <T extends @Nullable Object> T getValueOrNull(InternalOption<T> option) {
         return getOptionValue(option).get();
     }
 
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/transform/DefaultTransformInvocationFactory.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/transform/DefaultTransformInvocationFactory.java
@@ -147,7 +147,7 @@ public class DefaultTransformInvocationFactory implements TransformInvocationFac
     }
 
     private boolean isCachingDisabledByProperty(Transform transform) {
-        String experimentalProperty = internalOptions.getOptionValue(CACHING_DISABLED_PROPERTY).get();
+        String experimentalProperty = internalOptions.getValueOrNull(CACHING_DISABLED_PROPERTY);
         if (experimentalProperty != null) {
             if (experimentalProperty.isEmpty() || experimentalProperty.equals("true")) {
                 return true;
--- a/platforms/core-runtime/build-option/src/test/groovy/org/gradle/internal/buildoption/DefaultInternalOptionsTest.groovy
+++ b/platforms/core-runtime/build-option/src/test/groovy/org/gradle/internal/buildoption/DefaultInternalOptionsTest.groovy
@@ -197,6 +197,39 @@ class DefaultInternalOptionsTest extends Specification {
 
     //endregion
 
+    //region getValueOrNull
+
+    def "getValueOrNull resolves value for #optionType option"() {
+        sysProps["org.gradle.internal.prop1"] = sysProp
+
+        expect:
+        options.getValueOrNull(option) == result
+
+        where:
+        optionType | sysProp | option                                                        | result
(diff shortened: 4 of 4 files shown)


alllex:

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/ConfigurationCacheStartParameter.kt
ov7a: 💅 Should it be renamed to `internalOptions`?

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationTrace.java
```diff
@@ -188,7 +186,7 @@ public BuildOperationTrace(File userActionRootDir, InternalOptions internalOptio
 
     @Nullable
     private static Path resolveBasePath(InternalOptions internalOptions, File userActionRootDir) {
-        String basePath = internalOptions.getOption(TRACE_OPTION).get();
+        String basePath = internalOptions.getOptionValue(TRACE_OPTION).get();
```
ov7a: 🤔 Shouldn't it be `getString`?

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/DefaultInternalOptions.java
```diff
@@ -28,7 +28,7 @@ public DefaultInternalOptions(Map<String, String> properties) {
     }
 
     @Override
-    public <T extends @Nullable Object> Option.Value<T> getOption(InternalOption<T> option) {
+    public <T extends @Nullable Object> Option.Value<T> getOptionValue(InternalOption<T> option) {
```
ov7a, line 31: 🤔 I think there's a term overload here.
First time I read this, I saw `.getOptionValue(..).get()` and thought it was an optional value.

It feels that `getOption().get()` or  `getOption().getValue()` reads better. Maybe even `getConfiguredOption().getValue()`

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/InternalOptions.java
```diff
@@ -25,8 +25,101 @@
  */
 @ServiceScope({Scope.CrossBuildSession.class, Scope.BuildTree.class})
 public interface InternalOptions {
+
+    /**
+     * Resolves the value of a given option, allowing to determine whether it was set explicitly.
+     */
+    <T extends @Nullable Object> Option.Value<T> getOptionValue(InternalOption<T> option);
```
ov7a, line 32: 🤔 Should this even be exposed, given a bunch of helpers below?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/transform/DefaultTransformInvocationFactory.java
```diff
@@ -147,7 +147,7 @@ private ProjectInternal determineProducerProject(TransformStepSubject subject) {
     }
 
     private boolean isCachingDisabledByProperty(Transform transform) {
-        String experimentalProperty = internalOptions.getOption(CACHING_DISABLED_PROPERTY).get();
+        String experimentalProperty = internalOptions.getOptionValue(CACHING_DISABLED_PROPERTY).get();
```
ov7a: 💅 Should it be `getString`?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/ConfigurationCacheStartParameter.kt
alllex: This seems like a reasonable drive-by improvement 👍

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/InternalOptions.java
alllex: Yes, because we have internal options that represent a more involved abstraction. See `InvocationScenarioParameter`

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/DefaultInternalOptions.java
alllex: The `Option.Value` abstraction is shared between internal options and public build options. Renaming `get()` to `getValue()` there would significantly extend the scope of change in this PR. I'd like to avoid that.

I chose the `getOptionValue()` name because it directly corresponds with the `Option.Value` type that has been around for a while. If anything, I think it's more straightfoward this way.

However, even in the longer term, I don't see a great need to rename `get()` to `getValue()`. In the majority of cases for internal options one of the helpers should be used, that does unwrapping for you. For the public build options we usually don't have generically named accessors, and instead name then after the feature, like `configurationCache` here:
https://github.com/gradle/gradle/blob/a7f7ee579899bd5150d28a79f740d37012a2671e/platforms/core-runtime/start-parameter/src/main/java/org/gradle/api/internal/StartParameterInternal.java#L171

---

@ov7a , If we are not going to rename `get()` to `getValue()`, would you still insist on renaming `getOptionValue()` to something else?

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationTrace.java
alllex: The `InternalOptions.getString()` method is for ad-hoc properties, while `TRACE_OPTION` is already defined as an `InternalOption` field. The types don't match.

However, we also can't use `internalOptions.getValue()` here, because the trace option is nullable.

Your comment made me realize we are actually missing `getValueOrNull` for symmetry 🙏

platforms/core-runtime/build-option/src/main/java/org/gradle/internal/buildoption/DefaultInternalOptions.java
ov7a: Hmm. 

What about just `getValue()`? Since it will be called as `internalOptions.getValue(...)`  most of the time, IIUC?

alllex: The `getValue()` is already a method that unwraps `Option.Value`. Then we'd have to rename it to something else.

I've also just added `getValueOrNull(option): T?` that unwraps nullable values. We'd need a better name for it as well.

Overall, I think we'd mostly use the helper methods, so it's better to have the current `getValue(option): T` method with a short and sweet name, rather than renaming `getOptionValue()` to that name.

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationTrace.java
ov7a: It was confusing at first, but I get it now. The asymmetry between (`getValue(definedOption)` vs `getString("adhoc option name")` ) is annoying. Given the limited number of types (string, boolean, and int), I'd like to either use plain types everywhere or introduce `getValue<Type>("adhoc option")` for symmetry (not sure if the Java type system is powerful enough to handle that, though).

But I also don't think it should be a blocker

PR #37226, opened 2026-03-21
Fix some CC-broken tests of native plugins

mlopatkin:


mlopatkin:
@bot-gradle test ACC

platforms/native/language-native/src/integTest/groovy/org/gradle/language/cpp/CppApplicationPublishingIntegrationTest.groovy
```diff
@@ -59,7 +59,7 @@ class CppApplicationPublishingIntegrationTest extends AbstractCppPublishingInteg
                 targetPlatform.set(binary.map { it.compileTask.get().targetPlatform.get() })
                 toolChain.set(binary.map { it.toolChain })
                 installDirectory = layout.projectDirectory.dir("install")
-                lib(configurations.install.filter { it != configurations.install.files[0] })
+                lib(configurations.install.elements.map { (it as List).drop(1) })
```
ov7a, line 62: ❔  Why does the first element need to be dropped here?
Can it be expressed at a higher level somehow?

mlopatkin: > ❔ Why does the first element need to be dropped here? Can it be expressed at a higher level somehow?

It is an executable that is handled separately below - at least, this is my understanding. Given that this is annotated as "HACK", I wouldn't spend time polishing this further.

PR #37227, opened 2026-03-21
Deprecate using Project objects as dependency notation

ghale:
Add deprecation warnings when a Project object is passed directly as a dependency or dependency constraint notation. Users should use DependencyHandler.project(String) or DependencyFactory.createProjectDependency(String) instead.

Specifically, this PR does the following:
  - Deprecates passing `Project` objects directly as dependency or dependency constraint notation (e.g., `implementation project` or `dependencies.create(project)`)
  - Directs users to use `DependencyHandler.project(String)` or `DependencyFactory.createProjectDependency(String)` instead
  - Marks `DependencyFactory.create(Project)` as `@Deprecated`
  - Updates existing callers to use one of the replacement methods
  - Updates existing tests that relied on the deprecated notation
  - Adds upgrade guide section about the deprecation

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/DependencyFactory.java
+++ b/subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/DependencyFactory.java
@@ -91,7 +91,7 @@ public interface DependencyFactory {
      *
      * @param project the project
      * @return the new dependency
-     * @deprecated Use {@link #createProjectDependency(String)} instead.
+     * @deprecated Use {@link #createProjectDependency(String)} instead. This method will be removed in Gradle 10.0.
      */
     @Deprecated
     ProjectDependency create(Project project);


ghale:
@bot-gradle test this

platforms/jvm/jacoco/src/main/java/org/gradle/testing/jacoco/plugins/JacocoReportAggregationPlugin.java
```diff
@@ -126,7 +130,7 @@ public void apply(Project project) {
         project.getPlugins().withId("jvm-test-suite", plugin -> {
             // Depend on this project for aggregation
             jacocoAggregation.configure(conf -> {
-                conf.getDependencies().add(project.getDependencyFactory().create(project));
+                conf.getDependencies().add(getDependencyFactory().createProjectDependency());
```
jvandort, line 133: Curious why we changed this to injecting a dependency factory instead of calling `project.getDependencyFactory` that was already there. Seems we did this in a number of places -- I'm fine with this just curious why we made the change

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/DetachedConfigurationsIntegrationTest.groovy
```diff
@@ -147,7 +147,7 @@ class DetachedConfigurationsIntegrationTest extends AbstractIntegrationSpec {
 
             def detached = configurations.detachedConfiguration()
             detached.attributes.attribute(Attribute.of("attr", String), "value")
-            detached.dependencies.add(dependencies.create(project(":")))
+            detached.dependencies.add(dependencies.create(dependencies.project(":")))
```
jvandort: This would probably be more clear as 
`detached.dependencies.add(dependencies.project())`

Right now `:` is the current project, so `dependencies.project(":")` returns a `ProjectDependency` on the current project, which we then pass back into `dependencies.create`

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/DetachedConfigurationsIntegrationTest.groovy
```diff
@@ -253,7 +253,7 @@ class DetachedConfigurationsIntegrationTest extends AbstractIntegrationSpec {
             ${mavenTestRepository()}
 
             def copy = configurations.detachedConfiguration(
-                dependencies.create(project(":")),
+                dependencies.create(dependencies.project(":")),
```
jvandort: We shouldn't need to pass the result of `dependencies.project` back into `dependencies.create`, right? Plus, the current project is `:` already, can we just call `dependencies.project()`?

subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/DependencyFactory.java
```diff
@@ -91,7 +91,9 @@ public interface DependencyFactory {
      *
      * @param project the project
      * @return the new dependency
+     * @deprecated Use {@link #createProjectDependency(String)} instead.
```
jvandort:
```suggestion
     * @deprecated Use {@link #createProjectDependency(String)} instead. This method will be removed in Gradle 10.0.
```

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/notations/DependencyConstraintProjectNotationConverter.java
```diff
@@ -44,6 +45,11 @@ public void describe(DiagnosticsVisitor visitor) {
 
     @Override
     public void convert(Project notation, NotationConvertResult<? super DependencyConstraint> result) throws TypeConversionException {
+        DeprecationLogger.deprecateAction("Using a Project object as a dependency constraint notation")
+            .withAdvice("Please use the project(String) method on DependencyHandler or the createProjectDependency(String) method on DependencyFactory instead.")
```
jvandort, line 49: The recommended solutions are to use APIs which return a `Dependency`, but this converter produces constraints.

Just wanna confirm that using any of these recommended APIs and then passing them into Object-accepting methods intended for dependency constraints still works, as we convert from dependency to dependency constraint?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependencyNotationDeprecationIntegrationTest.groovy
```diff
@@ -0,0 +1,247 @@
...
+class ProjectDependencyNotationDeprecationIntegrationTest extends AbstractIntegrationSpec {
+
+    // region Deprecation Expected
+
+    def "emits deprecation when Project object is used directly as dependency notation"() {
+        given:
+        settingsFile << "include 'sub'"
+        file("sub/build.gradle") << """
+            plugins { id 'java-library' }
+        """
```
jvandort: Curious why `sub` is included in this test at all? the root build file adds a dependency on itself, so it doesn't seem like `sub` is involved at all

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependencyNotationDeprecationIntegrationTest.groovy
```diff
@@ -0,0 +1,247 @@
...
+            dependencies {
+                implementation project.getDependencyFactory().createProjectDependency(":sub")
+            }
+        """
+
+        expect:
+        succeeds("dependencies")
+    }
+
+    // endregion
```
jvandort, line 260: Can we add some constraint sanity-checks to this `No Deprecation Expected` section?

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -36,6 +36,54 @@ The previous step will help you identify potential problems by issuing deprecati
...
+}
+----
+=====
+====
+
+Or, when using the `DependencyFactory` directly (e.g., from a plugin):
+
+[source,java]
+----
+project.getDependencyFactory().createProjectDependency(":some-project");
```
jvandort: Can we show what to "do" with this one of these dependencies when using it "directly"?

e.g.
```
Configuration conf = getSomeConfiguration()
ProjectDependency dep = project.getDependencyFactory().createProjectDependency(":some-project")
conf.getDependencies().add(dep)
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -36,6 +36,54 @@ The previous step will help you identify potential problems by issuing deprecati
...
+Instead, use the `project()` method on `DependencyHandler` or the `createProjectDependency()` method on `DependencyFactory`:
+
+====
+[.multi-language-sample]
+=====
+.build.gradle
+[source,groovy]
+----
+dependencies {
+    implementation project(":some-project")
```
```suggestion
    implementation(project(":some-project"))
```

So we use a syntax that is the same in Groovy and Kotlin

platforms/jvm/jacoco/src/main/java/org/gradle/testing/jacoco/plugins/JacocoReportAggregationPlugin.java
ghale: There's not a super strong reason, but there is a reason.  I've just been thinking a lot about plugins lately and what we need to do to understand their requirements, inputs, outputs, etc in the future.  Injecting a service allows us to statically understand that this plugin uses this service in a way that we can't if we only reference it at runtime.  It's admittedly a low-to-no value change, and if there was a strong argument against it, I wouldn't object to changing it back.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/notations/DependencyConstraintProjectNotationConverter.java
ghale: The `DependencyConstraintNotationParser` has a [ProjectDependencyNotationConverter](https://github.com/gradle/gradle/blob/56938d9c5c6dea153325b40f6074511610367b90/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/notations/DependencyConstraintNotationParser.java#L108) that allows it to convert a `ProjectDependency` into a `ProjectDependencyConstraint`.  We have some tests that use `DependencyHandler.project(String)` when declaring constraints and these seem to work fine.  So, I think this is fine.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/DetachedConfigurationsIntegrationTest.groovy
ghale: I wasn't looking very closely at what this test does.  I'm not sure why it was written this way in the first place, I was just mechanically changing the dependency to an equivalent.  Fixed.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/DetachedConfigurationsIntegrationTest.groovy
ghale: Again, being too mechanical while fixing tests and not paying enough attention to what it was actually doing.  Thanks for being thorough.  Fixed.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependencyNotationDeprecationIntegrationTest.groovy
ghale: Bah.  Copy/paste error.  Fixed.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependencyNotationDeprecationIntegrationTest.groovy
ghale: Done.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
ghale: Done.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
ghale: Done.

platforms/jvm/jacoco/src/main/java/org/gradle/testing/jacoco/plugins/JacocoReportAggregationPlugin.java
jvandort: Like I said I'm fine with this notation, was just curious why you preferred the other. Seme like a good reason to me. The less we can rely on a Project god object in plugins, the more we can reason about that that plugin depends on/uses

PR #37229, opened 2026-03-21
Fix client OOM when buffering output during user input prompt

### LINKED ISSUE #37228: Client process can OOM when buffering output during user input prompt
### Current Behavior

When a build prompts for user input via `UserInputHandler` while parallel tasks produce significant log output, the client process (wrapper/launcher) crashes with `OutOfMemoryError`. The daemon detects the disconnection and cancels the build. The user sees only the prompt. Buffered output is never replayed.

Also reproducible with `--debug` logging, which increases event volume enough to exhaust the client heap even without parallel execution.

### Expected Behavior

The build completes successfully. Output produced while the prompt is displayed is buffered without exhausting memory and replayed in full after the user responds.

### Context (optional)

`AbstractUserInputRenderer` buffers all incoming `OutputEvent`s in an unbounded `ArrayList` while waiting for user input. With parallel execution, concurrent tasks can easily produce enough output to exceed the client's 64MB heap (e.g., 4 subprojects x 50,000 log lines = ~80MB of buffered events).

```groovy
// settings.gradle
include 'sub1', 'sub2', 'sub3', 'sub4'
```

```groovy
// build.gradle
def handler = services.get(org.gradle.api.internal.tasks.userinput.UserInputHandler)
tasks.register('askYesNo') {
    def result = handler.askUser { it.askYesNoQuestion("Continue?") }
    doLast {
        println("Answer: " + result.getOrElse("<no answer>"))
    }
}
```

```groovy
// sub1/build.gradle (same for sub2, sub3, sub4)
tasks.register('verboseTask') {
    doLast {
        50_000.times { i ->
            logger.lifecycle("[$project.name] Output line $i: ${'x' * 180}")
        }
    }
}
```

Run with:
```shell
./gradlew askYesNo verboseTask --parallel
```

The build crashes before the user can respond to the prompt.

### Gradle version

9.5 (and all prior versions; the buffering has been unbounded since `AbstractUserInputRenderer` was introduced)

### Build scan URL (optional)

N/A, the client crashes before the build completes.

_No response_

eskatos:
* Fixes #37228

This PR adds disk overflow to the user input event buffer. Events are kept in memory up to a small threshold (~2,500 events, ~1MB), then spilled to a temporary file using Gradle's existing output events serialization. On resume, in-memory events are replayed first, then events are streamed back from disk one at a time. The overflow file is deleted after replay and registered with `deleteOnExit()` as a safety net.

Client-local timing signals are filtered during pause. `UpdateNowEvent`, `FlushOutputEvent`, and `EndOutputEvent` have no serializers and no value on replay. Skipping them is both necessary and correct.

`AbstractUserInputRenderer` implementations, part of the client output pipeline, are invoked serially behind `ThrottlingOutputEventListener`'s lock.

Because of test logging, `ForkedTestClasspathFactory` is touched to add access to `GradleUserHomeLookup` from `:wrapper-shared` to the testing infrastructure implementation.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/console/AbstractUserInputRenderer.java
+++ b/platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/console/AbstractUserInputRenderer.java
@@ -47,6 +47,7 @@ import java.util.Objects;
 public abstract class AbstractUserInputRenderer implements OutputEventListener {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AbstractUserInputRenderer.class);
+    private static final long MAX_OVERFLOW_FILE_SIZE = 1_024L * 1_024L * 1_024L; // 1GB
 
     static final int MEMORY_QUEUE_LIMIT = 2_500;
 
@@ -113,6 +114,10 @@ public abstract class AbstractUserInputRenderer implements OutputEventListener {
         if (overflowFailed) {
             eventQueue.add(event);
             return;
+        }
+        if (overflowFile != null && overflowFile.length() >= MAX_OVERFLOW_FILE_SIZE) {
+            cleanupOverflow();
+            throw new IllegalStateException("User input overflow file exceeded " + MAX_OVERFLOW_FILE_SIZE + " bytes, aborting to prevent filling up the disk");
         }
         try {
             if (overflowEncoder == null) {


eskatos:
@bot-gradle test RFN

eskatos:
@bot-gradle test RFN

eskatos:
@bot-gradle test RFN

eskatos:
@bot-gradle test RFN

platforms/core-runtime/file-temp/src/main/java/org/gradle/api/internal/file/temp/TempFiles.java
```diff
@@ -21,12 +21,20 @@
...
+import java.nio.file.attribute.PosixFilePermissions;
+import java.util.Set;
 
 /**
  * Security safe API's for creating temporary files.
  */
 public final class TempFiles {
 
+    private static final Set<PosixFilePermission> OWNER_ONLY_PERMISSIONS =
+        PosixFilePermissions.fromString("rw-------");
```
octylFractal: 💭 Why use the string which is a bit hard to parse, instead of just `EnumSet.of(PosixFilePermission.OWNER_READ, PosixFilePermission.OWNER_WRITE)`?

platforms/core-runtime/file-temp/src/main/java/org/gradle/api/internal/file/temp/TempFiles.java
```diff
@@ -47,6 +55,11 @@ static File createTempFile(@Nullable String prefix, @Nullable String suffix, Fil
         if(prefix.length() <= 3) {
             prefix = "tmp-" + prefix;
         }
-        return File.createTempFile(prefix, suffix, directory);
+        Path dir = directory.toPath();
+        if (Files.getFileStore(dir).supportsFileAttributeView("posix")) {
+            return Files.createTempFile(dir, prefix, suffix,
+                PosixFilePermissions.asFileAttribute(OWNER_ONLY_PERMISSIONS)).toFile();
```
octylFractal: 💭 Caching the whole attribute could be worthwhile, if you're going to cache the permissions?

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/console/AbstractUserInputRenderer.java
```diff
@@ -60,13 +90,51 @@ public void onOutput(OutputEvent event) {
...
+        try {
+            if (overflowEncoder == null) {
+                overflowFile = temporaryFileProvider.createTemporaryFile("user-input-overflow-", ".bin");
+                overflowFile.deleteOnExit();
+                overflowEncoder = new KryoBackedEncoder(Files.newOutputStream(overflowFile.toPath()));
+            }
+            outputEventSerializer.write(overflowEncoder, event);
+            overflowEventCount++;
+        } catch (Exception e) {
+            LOGGER.warn("Failed to write overflow event to disk, falling back to in-memory buffer", e);
```
octylFractal, line 131: ❓ Is this log safe? Isn't there a chance that this logs, causes more output, and comes back to write more events before `overflowFailed` is set? Highly unlikely, but possible.

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/console/AbstractUserInputRenderer.java
```diff
@@ -60,13 +90,51 @@ public void onOutput(OutputEvent event) {
...
+            eventQueue.add(event);
+            return;
+        }
+        try {
+            if (overflowEncoder == null) {
+                overflowFile = temporaryFileProvider.createTemporaryFile("user-input-overflow-", ".bin");
+                overflowFile.deleteOnExit();
+                overflowEncoder = new KryoBackedEncoder(Files.newOutputStream(overflowFile.toPath()));
+            }
+            outputEventSerializer.write(overflowEncoder, event);
```
octylFractal, line 128: ❓ Can we cap the size of the output file? Or at least check how full the disk is and cap it to some amount below that? I don't think we should leave the potential to fill up the disk if someone leaves a build running at some input point.

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/console/AbstractUserInputRenderer.java
```diff
@@ -102,6 +170,56 @@ private void replayEvents() {
             delegate.onOutput(iterator.next());
             iterator.remove();
         }
```
octylFractal: 💭 Not for now, but I noticed we're using `remove` on an `ArrayList` from the front, so this is going to repeatedly re-copy the items forwards. It would be better to either iterate over the whole thing, then clear it; or switch it to a proper queue implementation like `ArrayDeque`.

platforms/core-runtime/logging/src/test/groovy/org/gradle/internal/logging/console/UserInputConsoleRendererTest.groovy
```diff
@@ -161,6 +170,167 @@ class UserInputConsoleRendererTest extends Specification {
         renderer.eventQueue.empty
     }
 
+    def "overflows to disk and replays all events in order"() {
```
octylFractal, line 173: ❓ This doesn't look like it tests that it actually overflows to disk, right? If for some reason the `if` wasn't present, wouldn't this test still pass?

I would be satisfied with an assertion before the resume that there is a file in the `tmpDir`.

platforms/core-runtime/logging/src/test/groovy/org/gradle/internal/logging/console/UserInputConsoleRendererTest.groovy
```diff
@@ -161,6 +170,167 @@ class UserInputConsoleRendererTest extends Specification {
...
+        renderer.onOutput(new UserInputResumeEvent(123))
+
+        then:
+        1 * console.getBuildProgressArea() >> buildProgressArea
+        1 * console.buildOutputArea >> textArea
+        1 * console.getBuildProgressArea() >> buildProgressArea
+        tmpDir.listFiles().length == 0
+    }
+
+    def "handles multiple pause/resume cycles with overflow"() {
```
octylFractal, line 240: :x: Likewise, this test also isn't verifying that a file gets created at all. It just asserts there aren't any files afterwards, but that doesn't mean one was ever made.

platforms/core-runtime/logging/src/test/groovy/org/gradle/internal/logging/console/UserInputConsoleRendererTest.groovy
```diff
@@ -161,6 +170,167 @@ class UserInputConsoleRendererTest extends Specification {
...
+
+        then:
+        1 * console.getBuildProgressArea() >> buildProgressArea
+        1 * console.buildOutputArea >> textArea
+        1 * console.getBuildProgressArea() >> buildProgressArea
+        replayed.size() == totalEvents
+        tmpDir.listFiles().length == 0
+    }
+
+    def "disk failure recovery across pause/resume cycles"() {
```
octylFractal, line 286: ❌ This test also isn't verifying that a file gets created at all, or doesn't in the first cycle.

platforms/core-runtime/logging/src/test/groovy/org/gradle/internal/logging/console/UserInputStandardOutputRendererTest.groovy
octylFractal, line 1: :x: Similar concerns as to `UserInputConsoleRendererTest`. Also, this is a lot of test case duplication it seems, is it possible to put them in a common (parent) test class?

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/console/AbstractUserInputRenderer.java
eskatos: It should not be an issue if it happens because this is all running serially behind `ThrottlingOutputEventListener`'s lock. New logging events will get fed to the top of the `OutputEventListener`s pipeline and processed in due time.

platforms/core-runtime/logging/src/test/groovy/org/gradle/internal/logging/console/UserInputConsoleRendererTest.groovy
eskatos: Good point! I adjusted all the test cases to assert on the presence/absence of an overflow file. This uncovered a bug on intermittent i/o issues, fixed in the same commit.

platforms/core-runtime/logging/src/test/groovy/org/gradle/internal/logging/console/UserInputStandardOutputRendererTest.groovy
eskatos: Just like many of the other test cases already duplicated between these two test classes, it is hard because of the mocks stubs and assertions. I did an attempt to extract a base class but it felt way less readable. If we ever have to touch the overflow behavior in the future, the failures in the tests will guide the required changes. I would prefer to keep things like they are.

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/console/AbstractUserInputRenderer.java
eskatos: This is a good question. I ruled out dropping events. I think we should simply fail the build in case the cap is hit. However, checking available space might be problematic. It varies over time and might be a slow operation, this could be especially problematic if we want to check regularly. With that in mind, a fixed cap comes to mind but then it is hard to pick a meaningful one.

platforms/core-runtime/file-temp/src/main/java/org/gradle/api/internal/file/temp/TempFiles.java
eskatos: Done

platforms/core-runtime/file-temp/src/main/java/org/gradle/api/internal/file/temp/TempFiles.java
eskatos: Done

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/console/AbstractUserInputRenderer.java
eskatos: Done

platforms/core-runtime/logging/src/test/groovy/org/gradle/internal/logging/console/UserInputConsoleRendererTest.groovy
eskatos: Done

platforms/core-runtime/logging/src/test/groovy/org/gradle/internal/logging/console/UserInputConsoleRendererTest.groovy
eskatos: Done

eskatos:
Thank you for the review Octavia, PTAL

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/console/AbstractUserInputRenderer.java
octylFractal: I think 1 or 5 GB is a decent choice, as a first value. A GB tends to be around when people start asking "what is taking up so much space", at least in my experience.

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/console/AbstractUserInputRenderer.java
octylFractal: OK, makes sense.

platforms/core-runtime/logging/src/test/groovy/org/gradle/internal/logging/console/UserInputStandardOutputRendererTest.groovy
octylFractal: Alright.

platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/console/AbstractUserInputRenderer.java
eskatos: Thanks. I went ahead and added a 1GB cap.

octylFractal (approved):

PR #37232, opened 2026-03-22
Fix toolchain downloads failing when Content-Disposition header lacks filename parameter

### LINKED ISSUE #37104: JVM toolchain downloads don't handle `Content-Disposition: attachment` header
### Current Behavior

If you have JDKs downloaded from
 https://jdk-cache.mycorp.com/jdks-bin/jdk22.32.21-ca-crac-jdk22.0.2-macosx_aarch64.tar.gz

but the have header 
`Content-Disposition: attachment` with no file name, the Gradle give an error like this: 

```
Unable to download toolchain matching the requirements ({languageVersion=22, vendor=any vendor, implementation=vendor-specific, nativeImageCapable=false}) from 'https://jdk-cache.mycorp.com/jdks-bin/jdk22.32.21-ca-crac-jdk22.0.2-macosx_aarch64.tar.gz
', due to: Can't determine filename for resource located at: https://jdk-cache.mycorp.com/jdks-bin/jdk22.32.21-ca-crac-jdk22.0.2-macosx_aarch64.tar.gz
```

### Expected Behavior

If the filename is missing from `Content-Disposition: attachment` HTTP header, then Gradle should just use the  file name from the path as a fallback. 

### Context (optional)

This is preventing us from using Toolchains with our https://jdk-cache.mycorp.com/jdks-bin/ 

Changing the `Content-Disposition: attachment` HTTP header is not easily done.

References on `Content-Disposition`
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition
- https://datatracker.ietf.org/doc/html/rfc2616#section-19.5.1
- https://www.rfc-editor.org/in-notes/rfc2183.html
- https://www.w3.org/Protocols/HTTP/Issues/content-disposition.txt

Also, HTTP headers are case insensitive: 
- https://datatracker.ietf.org/doc/html/rfc2616#section-4.2

> HTTP header fields, which include general-header ([section 4.5](https://datatracker.ietf.org/doc/html/rfc2616#section-4.5)),
   request-header ([section 5.3](https://datatracker.ietf.org/doc/html/rfc2616#section-5.3)), response-header ([section 6.2](https://datatracker.ietf.org/doc/html/rfc2616#section-6.2)), and
   entity-header ([section 7.1](https://datatracker.ietf.org/doc/html/rfc2616#section-7.1)) fields, follow the same generic format as
   that given in [Section 3.1 of RFC 822](https://datatracker.ietf.org/doc/html/rfc822#section-3.1) [[9](https://datatracker.ietf.org/doc/html/rfc2616#ref-9)]. Each header field consists
   of a name followed by a colon (":") and the field value. Field names
   **_are case-insensitive_**.

none, relies on internal jdk repo

### Gradle version

9.3.1

The relevant code is here:
https://github.com/gradle/gradle/blob/b631911858264c0b6e4d6603d677ff5218766cee/platforms/software/resources-http/src/main/java/org/gradle/internal/resource/transport/http/HttpResponseResource.java#L80-L108

rpalcolea:
Fixes #37104

### Context

When downloading JDK toolchains from corporate repositories that return a `Content-Disposition: attachment` header without a `filename` parameter, Gradle fails with:

  > Can't determine filename for resource located at: ...

The root cause is in `HttpResponseResource.getFilename()`: when a `Content-Disposition` header is present but contains no `filename` parameter, the URL-based filename fallback is never reached because it sits inside the `else` branch (only executed when the header is entirely absent).

This affects any organization using an internal artifact repository or proxy that serves JDK archives with a bare `Content-Disposition: attachment` header, which is valid per RFC 2183.

This PR:
  - Restructures `getFilename()` so the URL path fallback always runs when the header provides no usable filename
  - Replaces the fragile `indexOf`: based Content-Disposition parsing with a robust implementation that handles case-insensitive parameter names (RFC 2616), quoted strings with escapes, semicolons inside quotes, and RFC 5987 `filename*` encoding
  - Adds unit tests covering all parsing edge cases and an integration test that reproduces the exact failure scenario from the issue

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/language-java/src/integTest/groovy/org/gradle/jvm/toolchain/JavaToolchainDownloadSpiIntegrationTest.groovy
+++ b/platforms/jvm/language-java/src/integTest/groovy/org/gradle/jvm/toolchain/JavaToolchainDownloadSpiIntegrationTest.groovy
@@ -159,6 +159,44 @@ class JavaToolchainDownloadSpiIntegrationTest extends AbstractIntegrationSpec {
         result.assertNotOutput(testPath)
     }
 
+    @Issue("https://github.com/gradle/gradle/issues/37104")
+    @Requires(IntegTestPreconditions.Java11HomeAvailable)
+    def "toolchain download succeeds when server sends Content-Disposition attachment without filename"() {
+        given:
+        Assume.assumeFalse(JavaVersion.current() == JavaVersion.VERSION_11)
+
+        def jdkRepository = new JdkRepository(JavaVersion.VERSION_11)
+        def uri = jdkRepository.start()
+        jdkRepository.expectHead()
+        jdkRepository.expectGetWithContentDispositionAttachment()
+
+        executer
+            .requireOwnGradleUserHomeDir("needs to not have cached toolchains")
+            .withToolchainDownloadEnabled()
+
+        settingsFile << """
+            ${applyToolchainResolverPlugin("CustomToolchainResolver", singleUrlResolverCode(uri))}
+        """
+
+        buildFile << """
+            apply plugin: "java"
+
+            java {
+                toolchain {
+                    languageVersion = JavaLanguageVersion.of(11)
+                }
+            }
+        """
+
+        file("src/main/java/Foo.java") << "public class Foo {}"
+
+        expect:
+        succeeds("compileJava")
+
+        cleanup:
+        jdkRepository.stop()
+    }
+
     def "custom toolchain registries are consulted in order"() {
         settingsFile << """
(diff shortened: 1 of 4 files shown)


ljacomet:

big-guy:
This looks valid, but I wonder if we should even care about the file name from Content-Disposition. It would be simpler if we just controlled the file name ourselves.

big-guy:
@rpalcolea would it make more sense for us to just generate our own name for this? Do we need to always use the name provided by the server?

PR #37242, opened 2026-03-23
Relax failure assertions to fix flaky tests

tresat:
Use prefix matching for task failure descriptions instead of exact matches. This avoids brittle assertions on task registration provenance, which can vary between environments and caused integration tests to be flaky.


tresat:
@bot-gradle test this

PR #37249, opened 2026-03-23
Fix threadPoolFactoryClass configuration for TestNG >= 7.10

### LINKED ISSUE #28955: TestNG's threadPoolFactoryClass parameter broken for TestNG >= 7.10
### Current Behavior

Using TestNG 10.7.1, the latest version, a `org.gradle.api.InvalidUserDataException: The version of TestNG used does not support setting thread pool factory class` exception is thrown when setting `threadPoolFactoryClass`.

### Expected Behavior

Using TestNG 10.7.1, Gradle should allow configuration of the `threadPoolFactoryClass` option.

### Context (optional)

I think the issue is that Gradle is reflectively checking for the presence of `setExecutorFactoryClass` on `org.testng.testNG`. However, it looks like this changed to `setExecutorServiceFactory` in [this PR](https://github.com/testng-team/testng/pull/3075/files#diff-fd7fcf3b0ee340e791a1bdb372f4eeef4986d4daec08691434053a6079c8d092L846) which went into 7.10.

### Steps to Reproduce

github.com/clayburn/testng-issue

### Gradle version

8.7

### Build scan URL (optional)

https://ge.solutions-team.gradle.com/s/4oz3keudfzdie

_No response_

rpalcolea:
Fixes #28955

### Context

TestNG 7.10.0 replaced `setExecutorFactoryClass(String)` with `setExecutorServiceFactory(IExecutorServiceFactory)`. Gradle's TestNGTestRunner only tried the old method via reflection, so configuring `threadPoolFactoryClass` with TestNG >= 7.10 threw `InvalidUserDataException: The version of TestNG used does not support setting thread pool factory class`.

This PR updates `TestNGTestRunner.setThreadPoolFactoryClass()` to try the new API first (loading and instantiating the user's class as an `IExecutorServiceFactory`), then fall back to the legacy `setExecutorFactoryClass(String)` API for TestNG 7.0–7.9. This mirrors the existing pattern used for `setConfigFailurePolicy()`.

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/testng/TestNGTestRunner.java
+++ b/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/testng/TestNGTestRunner.java
@@ -182,7 +182,28 @@ public class TestNGTestRunner {
         }
     }
 
+    /**
+     * The setter for the thread pool factory has a different signature depending on TestNG version.
+     * This method uses reflection to detect the API and calls the version with the correct signature.
+     *
+     * TestNG >= 7.10 uses {@code setExecutorServiceFactory(IExecutorServiceFactory)}.
+     * TestNG 7.0–7.9 uses {@code setExecutorFactoryClass(String)}.
+     */
     private void setThreadPoolFactoryClass(TestNG testNg, String threadPoolFactoryClass) {
+        try {
+            Class<?> factoryInterface = Class.forName("org.testng.IExecutorServiceFactory", false, applicationClassLoader);
+            Class<?> factoryClass = applicationClassLoader.loadClass(threadPoolFactoryClass);
+            if (!factoryInterface.isAssignableFrom(factoryClass)) {
+                throw new InvalidUserDataException(String.format(
+                    "The thread pool factory class '%s' does not implement org.testng.IExecutorServiceFactory.", threadPoolFactoryClass));
+            }
+            Object factoryInstance = JavaReflectionUtil.newInstance(factoryClass);
+            JavaMethod.of(TestNG.class, Object.class, "setExecutorServiceFactory", factoryInterface).invoke(testNg, factoryInstance);
+            return;
+        } catch (ClassNotFoundException e) {
+            // New API not available, fall through to legacy API
+        }
+
         try {
             JavaMethod.of(TestNG.class, Object.class, "setExecutorFactoryClass", String.class).invoke(testNg, threadPoolFactoryClass);
         } catch (org.gradle.internal.reflect.NoSuchMethodException e) {
--- a/platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/testng/TestNGThreadPoolFactoryIntegrationTest.groovy
+++ b/platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/testng/TestNGThreadPoolFactoryIntegrationTest.groovy
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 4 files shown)


ov7a:

big-guy:
This looks reasonable.  We'll look at incorporating this PR soon.

Thanks again.

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/testng/TestNGTestRunner.java
```diff
@@ -182,7 +182,28 @@ private void setGroupByInstances(TestNG testNg, boolean value) {
...
+            Class<?> factoryClass = applicationClassLoader.loadClass(threadPoolFactoryClass);
+            if (!factoryInterface.isAssignableFrom(factoryClass)) {
+                throw new InvalidUserDataException(String.format(
+                    "The thread pool factory class '%s' does not implement org.testng.IExecutorServiceFactory.", threadPoolFactoryClass));
+            }
+            Object factoryInstance = JavaReflectionUtil.newInstance(factoryClass);
+            JavaMethod.of(TestNG.class, Object.class, "setExecutorServiceFactory", factoryInterface).invoke(testNg, factoryInstance);
+            return;
+        } catch (ClassNotFoundException e) {
+            // New API not available, fall through to legacy API
```
ghale, line 204: ❌  I think there is misleading behavior here.  If the factory interface fails to load, then this falls through and we use the legacy API, which is correct.  However, if the factory interface _does_ load, but `threadPoolFactoryClass` fails to load (e.g. not on classpath) it also falls through to the legacy API.  On 7.10+, the user would then get a misleading error - that "the version of TestNG used does not support setting thread pool factory class" - which is not actually the problem.  

We should wrap a `ClassNotFoundException` try/catch specifically around the load for `IExecutorServiceFactory` and only fall through to the legacy behavior when that fails.  We should throw a separate informative error when the thread pool factory class cannot be loaded.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/testng/TestNGThreadPoolFactoryIntegrationTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+
+package org.gradle.testing.testng
+
+import org.gradle.integtests.fixtures.AbstractIntegrationSpec
+import spock.lang.Issue
+
+@Issue("https://github.com/gradle/gradle/issues/28955")
+class TestNGThreadPoolFactoryIntegrationTest extends AbstractIntegrationSpec {
+
+    def "can configure threadPoolFactoryClass with TestNG 7.10+ (new IExecutorServiceFactory API)"() {
```
ghale, line 25: 💅 Are we sure that the executor service factory class was actually used?  If a future regression caused it not to be set, this test would still pass, right?  It would be better to have the custom class do something we can validate, like writing a file or dropping something into the output.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/testng/TestNGThreadPoolFactoryIntegrationTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+ * limitations under the License.
+ */
+
+package org.gradle.testing.testng
+
+import org.gradle.integtests.fixtures.AbstractIntegrationSpec
+import spock.lang.Issue
+
+@Issue("https://github.com/gradle/gradle/issues/28955")
+class TestNGThreadPoolFactoryIntegrationTest extends AbstractIntegrationSpec {
```
ghale, line 23: 💅 We should have coverage for two cases introduced by this change:
1) the specified class cannot be loaded
2) the specified class can be loaded but does not implement `IExecutorServiceFactory`.

ghale (changes requested):
This looks good.  I left a few minor comments, but we should fix the issue with the misleading class not found exception flow before merging.

ghale:
These changes are merged in #38333

PR #37254, opened 2026-03-23
Report invalid toolchain installation paths as warnings via Problems API instead of failing the build 

### LINKED ISSUE #34554: Report relative/invalid path issues in toolchains without failing the build
### Expected Behavior

The build should not fail if relative or invalid paths are provided in `org.gradle.java.installations.paths`. The user should still be able to find some kind of warning about it, perhaps via Problems API.

If no toolchains are found, these warnings should be directly surfaced in the error.

The build fails (https://github.com/gradle/gradle/issues/31659, https://github.com/gradle/gradle/issues/34548).

### Context

This is potentially an issue for users given https://github.com/gradle/gradle/issues/34548.
It would make sense to allow invalid state here since these paths aren't a requirement to run the build. If we can still find the toolchain elsewhere, it's fine.

rpalcolea:
fixes #34554

### Context
When users provide relative or unresolvable paths in `org.gradle.java.installations.paths` (or via `org.gradle.java.installations.fromEnv`), the build crashes with `UnsupportedOperationException: Cannot convert relative path ... to an absolute file.` This affects cross-platform builds where Unix-style absolute paths like `/opt/jdk-24`  are treated as relative on Windows, and setups where relative paths are committed to `gradle.properties` in version control.

These paths are optional configuration — valid toolchains may exist elsewhere, so an unresolvable entry should not prevent the build from running.

#### Note on `InternalProblems` nullability

My understanding is that `DaemonClientToolchainServices` runs before the build tree is set up (when resolving toolchains to find a compatible daemon), so `InternalProblems` cannot be wired there. Happy to hear/adapt if you folks see a better approach

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/jvm/jvm-services/src/main/java/org/gradle/internal/jvm/inspection/DefaultJavaInstallationRegistry.java
```diff
@@ -170,32 +182,45 @@ private Set<InstallationLocation> collectInstallations(List<InstallationSupplier
...
+    }
+
+    private void reportProblem(InstallationLocation installationLocation, String message) {
+        problemReporter.reportProblemIfNeeded(logger, installationLocation, message);
+        if (problems != null) {
+            ProblemId problemId = ProblemId.create("invalid-jvm-installation", "Invalid JVM installation", TOOLCHAIN_PROBLEM_GROUP);
+            problems.getInternalReporter().report(problemId, spec -> spec
+                .contextualLabel(message)
+                .severity(installationLocation.isAutoDetected() ? Severity.ADVICE : Severity.WARNING)
+                .solution("Ensure that the configured JVM installation path is a valid, absolute path to a JDK or JRE installation")
```
tresat: ❓ Question to Problems API devs - if the severity is merely ADVICE/WARNING, is the solution ever used?

tresat (commented):
It would be good to also have integration test coverage on this.

big-guy:
This looks reasonable, but we'll need to look at this more because we've sometimes accepted relative paths (WRT toolchains) and sometimes not and I think both options causes problems depending on what you think is the "right" behavior.

tresat:
@big-guy The error message here suggests to use a valid _absolute_ path.  I added tests to show the behavior with relative paths, which I think seem reasonable.

PR #37255, opened 2026-03-23
Fix native-image capability detection on Windows by also checking for native-image.cmd

### LINKED ISSUE #36118: JVM toolchain with nativeImageCapable flag on Windows fails
### Current Behavior

JDK zip is downloaded to Gradle cache and unzipped, but then throws an exception when verifying capabilities.
```Toolchain installation 'D:\dev\dep-caches\gradle\jdks\bellsoft-25-amd64-windows.2' does not provide the required capabilities: [NATIVE_IMAGE]```

### Expected Behavior

Successful build.

### Context (optional)

I believe the issue is in `org.gradle.internal.jvm.inspection.JvmInstallationMetadata`.
```
if (getToolByExecutable("native-image").exists()) {
    capabilities.add(JavaInstallationCapability.NATIVE_IMAGE);
}
```

On Windows, `getToolByExecutable` looks for a file with an `.exe` extension. However, in both Graal and Liberica, the actual tool is `bin/native-image.cmd`.

`settings.gradle.kts`
```
plugins {
    id("org.gradle.toolchains.foojay-resolver-convention") version "1.0.0"
}
```

`build.gradle.kts`
```
javaToolchains.launcherFor {
    vendor = JvmVendorSpec.BELLSOFT
    languageVersion = JavaLanguageVersion.of(25)
    nativeImageCapable = true
}.get()
```

### Gradle version

9.2.1

### Build scan URL (optional)

_No response_

_No response_

### PR DESCRIPTION
**Pull request by rpalcolea**
Fixes #36118

### Context

When requesting a JVM toolchain with `nativeImageCapable = true` on Windows, Gradle fails to detect the native-image tool because GraalVM and Liberica distributions ship it as `native-image.cmd` rather than `native-image.exe`. This causes the build to fail with "Toolchain installation does not provide the required capabilities: [NATIVE_IMAGE] even though the JDK was correctly downloaded and extracted.

The fix adds a fallback check for `native-image.cmd` on Windows in both the capability detection (`JvmInstallationMetadata`) and daemon compatibility check (`DaemonCompatibilitySpec`).

The native-image `.cmd` detection can only be exercised on Windows with a GraalVM or Liberica JDK installed. The existing capability detection logic (`gatherCapabilities`) uses `OperatingSystem.current()` internally, which cannot be overridden in unit tests. 

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/daemon-protocol/src/main/java/org/gradle/launcher/daemon/context/DaemonCompatibilitySpec.java
+++ b/platforms/core-runtime/daemon-protocol/src/main/java/org/gradle/launcher/daemon/context/DaemonCompatibilitySpec.java
@@ -71,7 +71,8 @@ public class DaemonCompatibilitySpec implements ExplainingSpec<DaemonContext> {
             if (daemonJvmCriteria.isCompatibleWith(potentialContext.getJavaVersion(), potentialContext.getJavaVendor())) {
                 if (daemonJvmCriteria.isNativeImageCapable()) {
                     // Need to assess that $JAVA_HOME/bin/native-image exists
-                    return new File(new File(potentialContext.getJavaHome(), "bin"), OperatingSystem.current().getExecutableName("native-image")).exists();
+                    // GraalVM and Liberica distributions ship native-image as a .cmd script on Windows
+                    return hasNativeImageTool(potentialContext.getJavaHome());
                 } else {
                     return true;
                 }
@@ -96,6 +97,17 @@ public class DaemonCompatibilitySpec implements ExplainingSpec<DaemonContext> {
             }
         } catch (IOException e) {
             // ignore
+        }
+        return false;
+    }
+
+    private static boolean hasNativeImageTool(File javaHome) {
+        File binDir = new File(javaHome, "bin");
+        if (new File(binDir, OperatingSystem.current().getExecutableName("native-image")).exists()) {
+            return true;
+        }
+        if (OperatingSystem.current().isWindows()) {
+            return new File(binDir, "native-image.cmd").exists();
         }
         return false;
     }
--- a/platforms/core-runtime/daemon-protocol/src/test/groovy/org/gradle/launcher/daemon/context/DaemonCompatibilitySpecSpec.groovy
+++ b/platforms/core-runtime/daemon-protocol/src/test/groovy/org/gradle/launcher/daemon/context/DaemonCompatibilitySpecSpec.groovy
@@ -27,8 +27,10 @@ import org.gradle.test.fixtures.file.TestFile
 import org.gradle.test.fixtures.file.TestNameTestDirectoryProvider
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.FileSystemTestPreconditions
+import org.gradle.test.preconditions.OsTestPreconditions
 
 import org.junit.Rule
+import spock.lang.Issue
 import spock.lang.Specification
 
 class DaemonCompatibilitySpecSpec extends Specification {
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by ov7a** [comment]

This PR has a valid DCO. The relevant team for this area will confirm the implementation choices and will decide if test coverage is needed.

**Comment by big-guy** [comment]
We'll take a look at this for 9.x

File: platforms/core-runtime/daemon-protocol/src/main/java/org/gradle/launcher/daemon/context/DaemonCompatibilitySpec.java [hunk 1]
```diff
@@ -100,6 +101,17 @@ private boolean jvmCompatible(DaemonContext potentialContext) {
         return false;
     }
 
+    private static boolean hasNativeImageTool(File javaHome) {
+        File binDir = new File(javaHome, "bin");
+        if (new File(binDir, OperatingSystem.current().getExecutableName("native-image")).exists()) {
```
octylFractal, line 106: 💭 Probably in the future we should have `getExecutableNames` (and similar for `getToolByExecutable`), and check all of `PATHEXT` on Windows, just to be safe. But this is fine for now.

octylFractal (approved):
This looks good, thank you. I added some tests and rebased it, and verified myself that it works on Windows with Liberica Native Image Kit.

platforms/core-runtime/daemon-protocol/src/test/groovy/org/gradle/launcher/daemon/context/DaemonCompatibilitySpecSpec.groovy
```diff
@@ -188,6 +190,58 @@ class DaemonCompatibilitySpecSpec extends Specification {
         clientStatus << [true, false]
     }
 
+    def "native-image capable criteria is satisfied by the native-image executable"() {
```
tresat, line 193: 💅 Could add a variation where both this file and `native-image.cmd` both exist, as an additional blackbox guarantee of behavior.

platforms/core-runtime/daemon-protocol/src/main/java/org/gradle/launcher/daemon/context/DaemonCompatibilitySpec.java
```diff
@@ -100,6 +101,17 @@ private boolean jvmCompatible(DaemonContext potentialContext) {
         return false;
     }
 
+    private static boolean hasNativeImageTool(File javaHome) {
```
tresat, line 104: ❓ How should this deal with non-lowercase file names?  Should they be allowed?  If so, we should parameterize the new tests to ensure they work.

platforms/core-runtime/launcher/src/integTest/groovy/org/gradle/launcher/daemon/DaemonToolchainIntegrationTest.groovy
```diff
@@ -158,4 +159,19 @@ class DaemonToolchainIntegrationTest extends AbstractIntegrationSpec implements
...
+        // Java 10 is not available
+        def java10 = AvailableJavaHomes.getAvailableJdks(JavaVersion.VERSION_1_10)
+        Assume.assumeTrue(java10.isEmpty())
+        writeJvmCriteria(JavaVersion.VERSION_1_10)
+        writeNativeImageCapableCriteria()
+        captureJavaHome()
+
+        expect:
+        fails("help")
+        failure.assertHasDescription("Cannot find a Java installation on your machine (${OperatingSystem.current()}) matching: {languageVersion=10, vendor=any vendor, implementation=vendor-specific, nativeImageCapable=true}")
```
tresat, line 175: 💅 This error message could be read to imply the _name of your machine_ is "Windows" or something similar.  Can it be rephrased to be unambiguous?

platforms/jvm/jvm-services/src/main/java/org/gradle/internal/jvm/inspection/JvmInstallationMetadata.java
```diff
@@ -261,6 +261,21 @@ private Set<JavaInstallationCapability> gatherCapabilities() {
             return capabilities;
         }
 
+        /**
+         * Checks for the native-image tool, which may be shipped as either an executable
+         * (native-image / native-image.exe) or a command script (native-image.cmd) on Windows.
+         */
+        private boolean hasNativeImageTool() {
```
tresat, line 268: 🤔 This duplicates the code in `DaemonCompatibilitySpec`.  Can these 2 methods be consolidated into a helper?

platforms/jvm/jvm-services/src/test/groovy/org/gradle/internal/jvm/inspection/JvmInstallationMetadataTest.groovy
```diff
@@ -0,0 +1,81 @@
...
+import org.gradle.test.precondition.Requires
+import org.gradle.test.preconditions.OsTestPreconditions
+import org.junit.Rule
+import spock.lang.Issue
+import spock.lang.Specification
+
+/**
+ * Tests capability detection in {@link JvmInstallationMetadata}.
+ */
+class JvmInstallationMetadataTest extends Specification {
```
tresat, line 31: 🤔 These tests are very similar to `DaemonCompatibilitySpecSpec`.  With a helper class, we could perhaps test the `hasNativeImageTool` call directly as well.

tresat (commented):
I think the non-lowercase name behavior should probably be validated.  Other than that, my suggestions are just minor cleanup.

platforms/core-runtime/launcher/src/integTest/groovy/org/gradle/launcher/daemon/DaemonToolchainIntegrationTest.groovy
octylFractal: This is outside the scope of this PR as that's part of `ToolchainProvisioningException`.

platforms/jvm/jvm-services/src/main/java/org/gradle/internal/jvm/inspection/JvmInstallationMetadata.java
octylFractal: I think this is a good follow-up PR but that we shouldn't do this here, just for this code. There is other code that is shared.

platforms/core-runtime/daemon-protocol/src/main/java/org/gradle/launcher/daemon/context/DaemonCompatibilitySpec.java
octylFractal: This is difficult to test as NTFS is case-insensitive by default. So the test can only really validate that it works on most Windows installations already, as we have no case-sensitive NTFS CI.

octylFractal: I added the test, but with a note about this.

platforms/core-runtime/daemon-protocol/src/test/groovy/org/gradle/launcher/daemon/context/DaemonCompatibilitySpecSpec.groovy
octylFractal: Done.

platforms/core-runtime/daemon-protocol/src/main/java/org/gradle/launcher/daemon/context/DaemonCompatibilitySpec.java
tresat: Could we use an in-memory file system like [jimfs](https://github.com/google/jimfs) here?

octylFractal: We would need to migrate relevant portions to use `Path` instead of `File`.

tresat: Perhaps just something else to document for a followup then.

PR #37262, opened 2026-03-24
Remove unnecessary getConventionValue call

big-guy:
We do not have a destination property anymore, so this should always return null

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


tresat:
Can `getProjectLayout()` also be removed?

big-guy:
> Can `getProjectLayout()` also be removed?

Good catch. Removed

PR #37266, opened 2026-03-24
Use hash-based path shrinking in HTML test reports only when needed

### LINKED ISSUE #37098: Use full name hashing only as a fallback in HTML test report
I stashed my work on this at
https://github.com/gradle/gradle/compare/ot/fix/hash-as-fallback-for-html-report

Originally discussed in https://github.com/gradle/gradle/pull/37052#discussion_r2927171156

octylFractal:
Fixes #37098

Add a path-length check that detects when the report output directory combined with test paths would exceed filesystem limits. When it does, all segment names in the report are shrunk to short hashes; otherwise the original (safe) names are kept as-is.

Also switches hash function to farmHashFingerprint64 for better non-adversarial collision resistance and speed.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/SafeFileLocationUtilsTest.groovy
+++ b/platforms/core-runtime/base-services/src/test/groovy/org/gradle/internal/SafeFileLocationUtilsTest.groovy
@@ -19,6 +19,7 @@ package org.gradle.internal
 import com.google.common.base.Utf8
 import org.apache.commons.lang3.RandomStringUtils
 import org.gradle.api.GradleException
+import org.gradle.internal.os.OperatingSystem
 import spock.lang.Specification
 
 import java.nio.charset.StandardCharsets
@@ -223,7 +224,7 @@ class SafeFileLocationUtilsTest extends Specification {
 
     def "PathLimitChecker returns WITHIN_LIMIT for short paths"() {
         expect:
-        new SafeFileLocationUtils.PathLimitChecker('/base').check(buildSegmentsToFile('a', 'b', 'c.html')) == SafeFileLocationUtils.PathLimitCheckResult.WITHIN_LIMIT
+        new SafeFileLocationUtils.PathLimitChecker(absolutePrefix() + 'base').check(buildSegmentsToFile('a', 'b', 'c.html')) == SafeFileLocationUtils.PathLimitCheckResult.WITHIN_LIMIT
     }
 
     def "PathLimitChecker returns EXCEEDS_LIMIT when path exceeds limit but can be shrunk"() {
@@ -412,7 +413,12 @@ class SafeFileLocationUtilsTest extends Specification {
     }
 
     private static String basePathWithRoom(int room) {
-        return '/' + 'B' * (MAX_PATH_LENGTH - 1 - room)
+        def prefix = absolutePrefix()
+        return prefix + 'B' * (MAX_PATH_LENGTH - prefix.length() - room)
+    }
+
+    private static absolutePrefix() {
+        OperatingSystem.current().isWindows() ? 'C:/' : '/'
     }
 
     private static List<SafeFileLocationUtils.Segment> buildSegmentsToFile(String... names) {


platforms/core-runtime/base-services/src/main/java/org/gradle/internal/SafeFileLocationUtils.java
```diff
@@ -367,6 +391,166 @@ private static int getSafeLength(byte[] rawName, int maxBytes) {
...
+        }
+
+        // Path exceeds limit — check if hashing segments can bring it under.
+        int shrunkByteLength = getPathByteLength(segments, SafeFileLocationUtils::minimumSafeName);
+        if (basePathByteLength + shrunkByteLength <= MAX_PATH_LENGTH) {
+            return PathLimitCheckResult.EXCEEDS_LIMIT;
+        }
+
+        // Even with hashing, path exceeds limit
+        return PathLimitCheckResult.UNSHRINKABLE;
```
tresat: 🤔 This bit seems excessively commented and the code could be compressed with `?:`.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
```diff
@@ -73,35 +71,44 @@
 public abstract class GenericHtmlTestReportGenerator implements TestReportGenerator {
 
     private static final Logger LOG = Logging.getLogger(GenericHtmlTestReportGenerator.class);
-    private static final HashFunction HASHER = Hashing.farmHashFingerprint64();
 
-    @VisibleForTesting
-    public static String hashSegment(String name) {
-        byte[] hashBytes = HASHER.hashUnencodedChars(name).asBytes();
-        return Base64.getUrlEncoder().withoutPadding().encodeToString(hashBytes);
+    public static String getFilePath(boolean shrink, org.gradle.util.Path path, boolean isLeaf) {
```
tresat:
```suggestion
    public static String buildFilePath(boolean shrink, org.gradle.util.Path path, boolean isLeaf) {
```

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
```diff
@@ -73,35 +71,44 @@
...
+        for (String dirName : directoryNames) {
+            builder.add(SafeFileLocationUtils.Segment.directory(dirName));
+        }
+        builder.add(SafeFileLocationUtils.Segment.file(fileName));
+        return builder.build();
     }
 
-    private static String getFilePath(TestTreeModel tree) {
-        return getFilePath(tree.getPath(), tree.getChildren().isEmpty());
+    private static String getFilePath(boolean shrink, TestTreeModel tree) {
```
tresat:
```suggestion
    private static String buildFilePathForModel(boolean shrink, TestTreeModel tree) {
```

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
```diff
@@ -163,10 +170,44 @@ private void generateReport(TestTreeModel root, List<TestOutputReader> outputRea
         LOG.info("Finished generating test html results ({}) into: {}", clock.getElapsed(), reportsDirectory);
     }
 
+    private static SafeFileLocationUtils.PathLimitCheckResult checkAllPaths(SafeFileLocationUtils.PathLimitChecker pathLimitChecker, TestTreeModel tree) {
```
tresat: 💅 We should import `PathLimitCheckResult` and not qualify the inner class everywhere.

tresat: 🤔 Can all or most of this file length checking be moved out of `GeneritHTmlTestReportGenerator`?  It doesn't seem like a core responsibility of the class at the same level of abstraction as the rest of what it does.

Could it live directly in the `HtmlReportFileGenerator`?

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
```diff
@@ -163,10 +170,44 @@ private void generateReport(TestTreeModel root, List<TestOutputReader> outputRea
...
     private void generateFiles(TestTreeModel model, final List<TestOutputReader> outputReaders) {
         try {
             HtmlReportRenderer htmlRenderer = new HtmlReportRenderer();
             buildOperationRunner.run(new DeleteOldReportOperation(fileCollectionFactory, deleter, reportsDirectory));
+            final String basePath = reportsDirectory.toAbsolutePath().toString();
+            final SafeFileLocationUtils.PathLimitChecker pathLimitChecker = new SafeFileLocationUtils.PathLimitChecker(basePath);
+            final SafeFileLocationUtils.PathLimitCheckResult allPathsCheckResult = checkAllPaths(pathLimitChecker, model);
+            if (allPathsCheckResult == SafeFileLocationUtils.PathLimitCheckResult.UNSHRINKABLE) {
+                throw new GradleException(
+                    "Cannot generate HTML test report because some test result paths exceed the file system limit and cannot be shortened." +
```
tresat: 🤔 We should print the unshrinkable path in the error message.  We should put the 2nd and 3rd sentence as solutions to try instead - which would mean using a narrow exception type that inplements `ResolutionProvider`.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlTestReportGenerator.java
octylFractal: It's called from `GenericHtmlTestReportGenerator.generateFiles`, so I don't see why it should be (somewhat arbitrarily) moved to the inner class.

tresat: I was thinking that extracting all the path calculation logic to somewhere else would make the `GenericHtmlTestReportGenerator` easier to read and understand.

`GenericHtmlTestReportGenerator`'s single responsibility would then be generating the HTML for the report, not calculating a path for it.  Maybe `HtmlReportFileGenerator` is not the right place for the path calculation, but I don't think it makes to `GenericHtmlTestReportGenerator` very cohesive to keep it there, alongside the actual HTML-writing, considering the complexity of the path calculation.

octylFractal: I can move it to a new upper-level helper, perhaps `HtmlTestReportPath`?

tresat: Sure.  I'm not picky about the where, just doesn't seem like it should stay here.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/HtmlTestReportPath.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.common.collect.ImmutableList;
+import org.gradle.internal.SafeFileLocationUtils;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Utilities related to generating file paths for the HTML test report.
+ */
+public class HtmlTestReportPath {
```
tresat: ❓ Can anything else here be made `private`?  

The name also suggests a value type to me, how about `HtmlTestReportPathShrinker`.

Other than these small issues, this looks good to me.

octylFractal: I used `HtmlTestReportPathBuilder`, since the shrinking isn't inherent, and made the `buildSegments` private.

octylFractal:
@bot-gradle test this

octylFractal:
@bot-gradle test this

PR #37268, opened 2026-03-25
Deprecate executing Gradle clients on JVM versions < 11

jvandort:
Gradle clients must support at least 5 prior Gradle versions. Starting in Gradle 10, that lower limit will be Gradle 5, which had a maximum JVM version of 11. This permits us to raise the minimum Java version of Gradle clients to Java 11 in Gradle 10.
    
To prepare for raising this limit, we emit warnings in the TAPI client and the CLI client. Then, we add appropriate tests to verify these warnings are emitted when running these clients on deprecated JVM versions.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test rfr

jvandort:
@bot-gradle test rfr

big-guy:
I haven't looked at the changes, but we should also consider which versions of the JVM we'll support seeing a good error message. I think it's OK to raise the minimum version, but we should still produce a good "This version is unsupported" message all the way down to 8.

jvandort:
> but we should still produce a good "This version is unsupported" message all the way down to 8.

This is something we decided to remove for Java 6 and 7. Not sure why we made that decision, but likely since the infrastructure we had was not worth maintaining for those versions.

I could be convinced something like this is desirable for java versions > 7. Last I thought about doing this was to have an asm class generator, so we didn't need to run an old versioned compiler to produce low enough bytecode. 

In any case, I don't think such a mechanism is needed in order add the deprecation.

platforms/ide/tooling-api/src/integTest/groovy/org/gradle/integtests/tooling/ToolingApiClientJdkCompatibilityTest.groovy
```diff
@@ -80,6 +80,7 @@ abstract class ToolingApiClientJdkCompatibilityTest extends AbstractIntegrationS
 
             dependencies {
                 implementation 'org.gradle:gradle-tooling-api:${distribution.version.baseVersion.version}'
+                implementation 'org.slf4j:slf4j-simple:2.0.17'
```
big-guy, line 83: What is this about?

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/consumer/DefaultGradleConnector.java
```diff
@@ -45,6 +48,22 @@ public class DefaultGradleConnector extends GradleConnector implements ProjectCo
...
+            LOGGER.warn("Executing the Gradle Tooling API Client on JVM versions {} and lower has been deprecated. " +
+                    "This will fail with an error in Gradle {}. " +
+                    "Use JVM {} or greater to execute the Gradle Tooling API Client. " +
+                    "Consult the upgrading guide for further information: {}",
+                SupportedJavaVersions.FUTURE_MINIMUM_CLIENT_JAVA_VERSION - 1,
+                currentMajorGradleVersion + 1,
+                SupportedJavaVersions.FUTURE_MINIMUM_CLIENT_JAVA_VERSION,
+                new DocumentationRegistry().getDocumentationFor("upgrading_version_" + currentMajorGradleVersion, "minimum_client_jvm_version")
+            );
+        }
```
big-guy: I don't think we can access these things here.

1. JavaVersion is going to be the copy from the tooling API jar and I think we've had problems in the past of this blowing up in newer versions of the JVM.
2. SupportedJavaVersions and DocumentationRegistry are a new dependency. 
3. This also doesn't show up as a real deprecation.

I think this is one of the first deprecations like this, so I don't have a good suggestion about how to do this differently. 

Assuming we had the deprecation in 9.6.  We have four scenarios when running on JVM 8:
1. TAPI 9.5 w/ Gradle 9.6 => Do we expect a TAPI 9.5 on JVM8 to be able to talk to a Gradle 10 on JVM17 when the client side requires JVM11?  If yes, no warning is OK. If no, we need a warning here.
2. TAPI 9.5 w/ Gradle 9.5 => no warning is OK
3. TAPI 9.6 w/ Gradle 9.6 => warning here is required
4. TAPI 9.6 w/ Gradle 9.5 => warning here is required

(1) and (3) could be handled all "provider" side, but it would be nice to warn consumers in (4) that the TAPI library itself is going to be changing in TAPI 10 even if they only run builds <Gradle 10. 

In some sense, there are two deprecations here:
1. Deprecation of the TAPI client moving from 8 -> 11.
2. Deprecation of the Gradle "launcher" moving from 8 -> 11.

big-guy (commented):
We need TIDE to look at this before moving forward

platforms/ide/tooling-api/src/integTest/groovy/org/gradle/integtests/tooling/ToolingApiClientJdkCompatibilityTest.groovy
jvandort: The tooling API bundles the SLF4J API but no implementation. This allows us to actually see SLF4J logs in `ToolingApiClientMinJdkCompatibilityTest`

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/internal/consumer/DefaultGradleConnector.java
jvandort: 1. Updated to `JavaVersionParser.parseMajorVersion(System.getProperty("java.version"))` to avoid JavaVersion
2. `DocumentationRegistry` lives in `logging` which we already depend on. `SupportedJavaVersions` is indeed a new dependency. But, are we not allowed to add new dependencies here?
3. Correct this will only show up as a log statement for users who have an SLF4J implementation on the classpath. However, we also did this for [ProjectBuilder](https://github.com/gradle/gradle/blob/99bc92b36e6c9e0160a131a9b240c46bbbaa964b/platforms/extensibility/unit-test-fixtures/src/main/java/org/gradle/testfixtures/internal/ProjectBuilderImpl.java#L134) when upgrading the daemon to Java 17

jvandort: Assuming Gradle 10 has a "client" version of Java 11, it would compile both its consumer and provider client-related classes to Java 11. 

This means a Gradle <10 client running on JDK 8 would load Java 11 classes from the Gradle 10 target distribution. It would seem the provider should also be emitting a warning. Given this I wonder how we would avoid a double deprecation warning in case 3

big-guy: >  Given this I wonder how we would avoid a double deprecation warning in case 3

I think we don't avoid it.  It's two deprecations:

1. You're using TAPI 9.6 on JVM8 to talk to _any_ Gradle version. This is going away.
2. You're talking to Gradle 9.6+ with TAPI on JVM8. This is going away.

In the future, the way we should have done this is to raise TAPI's minimum JVM version to 11 first, while keeping the launcher at 8, so when the launcher moved to 11, it's less likely that TAPI clients would even expect 8 support.

big-guy: > SupportedJavaVersions is indeed a new dependency. But, are we not allowed to add new dependencies here?

My concern was that this is blurring TAPI consumer (from the tooling api jar) vs TAPI provider (from the distribution) because SupportedJavaVersions is about the distribution and not TAPI.

For instance, SupportedJavaVersions.FUTURE_MINIMUM_CLIENT_JAVA_VERSION = 11 in Gradle 9.6. That's what gets included in TAPI 9.6.

In Gradle 10.1, SupportedJavaVersions.FUTURE_MINIMUM_CLIENT_JAVA_VERSION = 17. That now gets included in TAPI 10.1, but there are two deprecations:

1. TAPI itself needs to be Java 17 compatible
2. Gradle client needs to be Java 17 compatible

I think it could be confusing to look at the distribution's SupportedJavaVersions and see a different value because it's coming from an older tooling API jar. 

That may just mean that SupportedJavaVersions needs a TAPI equivalent that's about TAPI supported versions only.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -564,6 +564,13 @@ If you rely on seeing attributeless variants or configurations in these reports,
 
 === Deprecations
 
+[[minimum_client_jvm_version]]
+==== Running Gradle Clients on older JVMs
+
+Starting in Gradle 10.0.0, Gradle will require JVM 11 or later to run the CLI client and the Tooling API client.
```
lkasso, line 570:
```suggestion
Starting in Gradle 10, Gradle will require JVM 11 or later to run the CLI client and the Tooling API client.
```

big-guy:
I punted this to 9.7 because we need TIDE input.

PR #37271, opened 2026-03-25
Avoid TaskOriginTracker

jvandort:
TaskOriginTracker relies on PluginApplicationTracker, which relies on collection callback action decorator build operations. We want to remove these collection callback operations. Luckily, a recent change included a task's original UserCodeSource as part of the task identity, meaning we can reuse that, with some additional tweaks to the data it holds, to remove our dependency on TaskOriginTracker

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/UserCodeSourceCodec.kt
+++ b/platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/UserCodeSourceCodec.kt
@@ -37,8 +37,10 @@ object UserCodeSourceCodec : Codec<UserCodeSource> {
             is UserCodeSource.Script -> {
                 writeSmallInt(2)
                 writeString(value.displayName.displayName)
-                writeNullableString(value.uri?.toASCIIString())
+                writeNullableString(value.uri?.toString())
             }
+
+            else -> error("Unexpected user code source type: ${value.javaClass.name}")
         }
     }
 
--- a/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeSource.java
+++ b/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeSource.java
@@ -16,7 +16,7 @@
 
 package org.gradle.internal.code;
 
-import org.gradle.internal.DisplayName;
+import org.gradle.api.Describable;
 import org.jspecify.annotations.Nullable;
 
 import java.net.URI;
@@ -29,25 +29,25 @@ public interface UserCodeSource {
     /**
      * Returns the display name of the user code.
      */
-    DisplayName getDisplayName();
+    Describable getDisplayName();
 
     /**
      * User code that is applied as a binary plugin.
      */
     class Binary implements UserCodeSource {
 
-        private final DisplayName displayName;
+        private final Describable displayName;
         private final String className;
         private final @Nullable String pluginId;
 
-        public Binary(DisplayName displayName, String className, @Nullable String pluginId) {
+        public Binary(Describable displayName, String className, @Nullable String pluginId) {
             this.displayName = displayName;
             this.className = className;
             this.pluginId = pluginId;
         }
 
         @Override
-        public DisplayName getDisplayName() {
+        public Describable getDisplayName() {
             return displayName;
         }
 
@@ -72,16 +72,16 @@ public interface UserCodeSource {
      */
     class Script implements UserCodeSource {
 
-        private final DisplayName displayName;
+        private final Describable displayName;
         private final @Nullable URI uri;
 
-        public Script(DisplayName displayName, @Nullable URI uri) {
+        public Script(Describable displayName, @Nullable URI uri) {
             this.displayName = displayName;
             this.uri = uri;
         }
 
         @Override
-        public DisplayName getDisplayName() {
+        public Describable getDisplayName() {
             return displayName;
         }
 
--- a/platforms/core-configuration/configuration-problems-base/src/main/kotlin/org/gradle/internal/configuration/problems/PropertyProblem.kt
+++ b/platforms/core-configuration/configuration-problems-base/src/main/kotlin/org/gradle/internal/configuration/problems/PropertyProblem.kt
@@ -16,7 +16,7 @@
 
 package org.gradle.internal.configuration.problems
 
-import org.gradle.internal.DisplayName
+import org.gradle.api.Describable
 import org.gradle.internal.cc.impl.problems.JsonWriter
 import org.gradle.internal.code.UserCodeSource
 import org.gradle.internal.configuration.problems.StructuredMessage.Fragment.Reference
@@ -179,7 +179,7 @@ sealed class PropertyTrace {
 
     @ConsistentCopyVisibility
     data class BuildLogic private constructor(
-        val source: DisplayName,
+        val source: Describable,
         val lineNumber: Int? = null
     ) : PropertyTrace() {
         constructor(location: Location) : this(location.sourceShortDisplayName, location.lineNumber)
--- a/platforms/software/dependency-management/src/main/java/org/gradle/internal/management/DefaultDependencyResolutionManagement.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/internal/management/DefaultDependencyResolutionManagement.java
@@ -19,6 +19,7 @@ import com.google.common.collect.ImmutableList;
 import groovy.lang.Closure;
 import org.gradle.api.Action;
 import org.gradle.api.ActionConfiguration;
+import org.gradle.api.Describable;
 import org.gradle.api.InvalidUserCodeException;
 import org.gradle.api.NamedDomainObjectList;
 import org.gradle.api.artifacts.ComponentMetadataDetails;
@@ -183,7 +184,7 @@ public class DefaultDependencyResolutionManagement implements DependencyResoluti
 
     private void repoMutationDisallowedOnProject(ArtifactRepository artifactRepository) {
         UserCodeApplicationContext.Application current = context.current();
-        DisplayName displayName = current == null ? null : current.getSource().getDisplayName();
+        Describable displayName = current == null ? null : current.getSource().getDisplayName();
         if (displayName == null) {
             displayName = UNKNOWN_CODE;
         }
@@ -201,7 +202,7 @@ public class DefaultDependencyResolutionManagement implements DependencyResoluti
 
     private void ruleMutationDisallowedOnProject(DisplayName ruleName) {
         UserCodeApplicationContext.Application current = context.current();
-        DisplayName displayName = current == null ? null : current.getSource().getDisplayName();
+        Describable displayName = current == null ? null : current.getSource().getDisplayName();
         if (displayName == null) {
             displayName = UNKNOWN_CODE;
         }


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/UserCodeSourceCodec.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.gradle.internal.serialize.graph.Codec
+import org.gradle.internal.serialize.graph.ReadContext
+import org.gradle.internal.serialize.graph.WriteContext
+import java.net.URI
+
+
+object UserCodeSourceCodec : Codec<UserCodeSource> {
+
+    override suspend fun WriteContext.encode(value: UserCodeSource) {
+        when (value) {
```
tresat, line 30: 🤔 As we don't have sealed classes yet, this should throw a sensible error in the `else` case for writing as well as for reading.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/UserCodeSourceCodec.kt
```diff
@@ -0,0 +1,64 @@
...
+            is UserCodeSource.Binary -> {
+                writeSmallInt(1)
+                writeString(value.displayName.displayName)
```
tresat, line 33: ❗ This seems like a potential problem - `DisplayName` is an interface, and the implementation make contain more necessary fields than just `displayName`.  

I think the `UserCodeSource` should probably store as a field the result of `getCapitalizedDisplayName()`, rather than the complete `DisplayName` instance itself.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/UserCodeSourceCodec.kt
```diff
@@ -0,0 +1,64 @@
...
+                writeString(value.className)
+                writeNullableString(value.pluginId)
+            }
+            is UserCodeSource.Script -> {
+                writeSmallInt(2)
+                writeString(value.displayName.displayName)
+                writeNullableString(value.uri?.toASCIIString())
```
tresat: ❓ Is the ASCII string conversion needed?  Other Codecs do `writeString(url.toString())`.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeSource.java
```diff
@@ -16,46 +16,82 @@
...
+
+    /**
+     * User code that is applied as a script plugin.
+     */
+    class Script implements UserCodeSource {
+
+        private final DisplayName displayName;
+        private final @Nullable URI uri;
+
+        public Script(DisplayName displayName, @Nullable URI uri) {
```
tresat: ❓ Does anywhere other than test code actually use a nullable URL?  If not, can we just update the tests to supply a dummy and make this non-null?

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeSource.java
```diff
@@ -16,46 +16,82 @@
...
+    /**
+     * User code that is applied as a binary plugin.
+     */
+    class Binary implements UserCodeSource {
+
+        private final DisplayName displayName;
+        private final String className;
+        private final @Nullable String pluginId;
+
+        public Binary(DisplayName displayName, String className, @Nullable String pluginId) {
```
tresat: 🤔 We could create 2 sub-types for the case of a binary-by-name and binary-by-id, but that's probably not worth it.  Could also do 2 Constructors to remove the null arg - also not sure if that is worth it.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -128,21 +126,12 @@ class TaskNodeCodec(
...
-                writeByte(1)
-            }
-
-            else -> {
-                writeByte(2)
-                writeString(source.displayName.displayName)
-                writeNullableString(source.pluginId)
-            }
+    suspend fun WriteContext.writeUserCodeSource(source: UserCodeSource?) {
+        if (source == null) {
```
tresat, line 130: 🤔 The downside to not having a separate sentinel type for "By Rule" and just using `null` to track them is you get odd-looking code like this: write this code source, except when this code source is `null`.

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TaskOperationMapper.java
```diff
@@ -97,13 +100,40 @@ public DefaultTaskDescriptor createDescriptor(ExecuteTaskBuildOperationDetails d
...
         Set<InternalOperationDescriptor> dependencies = operationDependenciesResolver.resolveDependencies(details.getTaskNode());
-        InternalPluginIdentifier originPlugin = taskOriginTracker.getOriginPlugin(details.getTask().getTaskIdentity());
+        InternalPluginIdentifier originPlugin = toInternalPluginId(details.getTask().getTaskIdentity().getUserCodeSource());
         DefaultTaskDescriptor descriptor = new DefaultTaskDescriptor(id, taskIdentityPath, taskPath, displayName, parent, dependencies, originPlugin);
         DefaultTaskDescriptor descriptorWithoutDependencies = new DefaultTaskDescriptor(id, taskIdentityPath, taskPath, displayName, parent, emptySet(), originPlugin);
         descriptors.put(details.getTask().getTaskIdentity(), descriptorWithoutDependencies);
         return descriptor;
     }
 
+    private static @Nullable InternalPluginIdentifier toInternalPluginId(@Nullable UserCodeSource source) {
```
tresat, line 110: ❓ Would it be better to add `InternalPluginIdentifier toPluginIdentifier()` on `UserCodeSource` and move this code to the 2 implememntations?  It doesn't really seem to belong in this mapper class.

jvandort: In a way yes that would be nice, but `UserCodeSource` lives in `base-services` (for some tangling reasons that are not worth dealing with here) and (Default)InternalPluginIdentifier lives in `daemon-messaging`/`tooling-api`.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeSource.java
jvandort: I think the null arg is just fine. Definitely don't wanna add more subtypes

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeSource.java
jvandort: On `ResourceLocation`, the source of the script plugin content, both `getFile` and `getURI` can be null. This can happen if you define a `ResourceLocation` as a string. Not sure if it is possible for you to apply a plugin just from a string, but strictly following nullability of the types we are working with, this is possible. 

If it is not possible to create a script plugin with just a String, then `ResourceLocation` is probably the wrong type to use and we should fix that separately

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/UserCodeSourceCodec.kt
jvandort: Changed to toString

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/UserCodeSourceCodec.kt
jvandort: Done

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
jvandort: I'm not sure I see anything wrong with null here. Whether you use null, optional, or some sentinel singleton instance of the class, you are using it to encode some "I dont really know" state. Yes, we need an extra null check here, but if we used optional, we'd have to check if present, and if we used a sentinel we'd need to define the sentinel class, and then add encode/decode paths for that. 

What we're missing IMO is generic `readNullable` and `writeNullable`, where we already have a [readNullableString](https://github.com/gradle/gradle/blob/cb0c615fb8e3690971bb7f89ad80f58943360624/platforms/core-runtime/serialization/src/main/java/org/gradle/internal/serialize/AbstractDecoder.java#L70) and [writeNullableString](https://github.com/gradle/gradle/blob/cb0c615fb8e3690971bb7f89ad80f58943360624/platforms/core-runtime/serialization/src/main/java/org/gradle/internal/serialize/AbstractEncoder.java#L77) that does this exact same thing behind the scenes

jvandort: I think the key point here is I don't want to pollute the actual modeled data because it takes a little more code to serialize it. `UserCodeSource` is instantiated when we apply a plugin. As such, it precisely models the plugin being applied. A "by rule" user code source does not fit into that model, as it does not describe a plugin being applied.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/UserCodeSourceCodec.kt
jvandort: The goal of Describable/DisplayName is to avoid eagerly allocating/rendering strings in case you don't need them in the end. I'd rather not do that here. I've instead replaced `DisplayName` with `Describable`, since nobody used the capitalized display name anyway.

jvandort:
@bot-gradle test this

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/TaskOperationMapper.java
tresat: Got it.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
tresat: We definitely don't want a By Rule code source.  I was thinking more along the lines of an `UnknownCodeSource` to avoid so much `null`.

jvandort: I'm not sure I see the problem with null here

tresat: I admit it's a minor concern.  You don't think it's odd to need to pass a nullable to a method, and the first line of the method is "Was I passed null?  If so, write a sentinel." vs. either not calling the method or passing it a sentinel in the first place, that would give some semantic context to a reader about what situation we're in?

jvandort: To me I don't think the caller really cares. We are trying to serialize a task identity. It may or may not have a user codec. And when I deserialize it, it also may or may not have a codec. I wouldn't really want to pollute the call-site with a null check inline, because the caller doesn't care. If its null when serializing, you'll get null when deserializing. The nullness does not affect the user. 

I really don't want to argue about something so small. This PR is only part of a larger piece of work and I don't want to hold anything up talking about `null`. 

However I still don't think an `Unknown` UserCodeSource fits in the model

tresat: Yeah, not something to get held up on.  Everything else is addressed, so I marked approved.

PR #37276, opened 2026-03-25
Fix scope of `kotlin-script-caching-disabled` internal property

eskatos:
The `kotlin-script-caching-disabled` internal property only disabled caching of .kts compilation unit but not the associated accessor generation.

Accessors generation is in general much faster than .kts compilation itself. The sheer number of units of work for accessors generation incurs a noticeable penalty when checking for the availability of build cache entries, and storing them, compared to their actual execution.

Extract `KotlinDslInternalOptions` to centralize access to the property and use it in all the units of work involved in Kotlin DSL script compilation.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinDslInternalOptions.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinDslInternalOptions.kt
@@ -25,5 +25,5 @@ internal object KotlinDslInternalOptions {
     val CACHING_DISABLED_PROPERTY: InternalOption<Boolean> =
         InternalOptions.ofBoolean("org.gradle.internal.kotlin-script-caching-disabled", false)
     val CACHING_DISABLED_REASON: CachingDisabledReason =
-        CachingDisabledReason(CachingDisabledReasonCategory.NOT_CACHEABLE, "Caching of Kotlin script compilation disabled by property")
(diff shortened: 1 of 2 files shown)


platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinDslInternalOptions.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.gradle.internal.buildoption.InternalOption
+import org.gradle.internal.buildoption.InternalOptions
+import org.gradle.internal.execution.caching.CachingDisabledReason
+import org.gradle.internal.execution.caching.CachingDisabledReasonCategory
+
+internal object KotlinDslInternalOptions {
+    val CACHING_DISABLED_PROPERTY: InternalOption<Boolean> =
+        InternalOptions.ofBoolean("org.gradle.internal.kotlin-script-caching-disabled", false)
+    val CACHING_DISABLED_REASON: CachingDisabledReason =
+        CachingDisabledReason(CachingDisabledReasonCategory.NOT_CACHEABLE, "Caching of Kotlin script compilation disabled by property")
```
h0tk3y: 💄 Given that it's no longer just script compilation, would it make sense to rephrase the reason?

```suggestion
        CachingDisabledReason(CachingDisabledReasonCategory.NOT_CACHEABLE, "Caching of Kotlin script compilation and Kotlin DSL accessors generation disabled by property")
```

PR #37277, opened 2026-03-25
Update PMD test assertions for task provenance

tresat:
Update the expected failure messages in PMD integration tests to reflect changes in task provenance reporting. Tasks are now identified as registered by the PMD plugin rather than simply within the build file.


tresat:
@bot-gradle test this

PR #37283, opened 2026-03-25
Organize service registration per Gradle mode

alllex:
Today, `configuration-cache` module serves as an entry point into the orchestration of a Gradle invocation, even if CC is disabled. This is due to the fact that we need to choose different service implementations _depending on_ whether CC (or IP) is enabled. The `configuration-cache` module directly knows about all the implementations and selects between the needed one. The module also houses the logic that determines the mode of operation itself.

Eventually, we want to separate these concerns and we also want to make IP implementation less dependent on the CC internals. Specifically, we'd want to have another module whose job it is to choose the mode and orchestrate the invocation. While the `configuration-cache` module can be concerned with the _infrastructure_ required for CC operations.

This PR is a step towards that direction. It is intended as a pure refactoring. It reorganizes the service registration, mostly following the existing patterns, but bringing the related services closer and making it clear whether some services are needed in one or more modes.


alllex:

alllex:
@coderabbitai review

PR #37286, opened 2026-03-26
Start daemon with correct classpath

### LINKED ISSUE #30304: Gradle writes classpath arg-file for workers in wrong encoding leading to workers immediately crashing
### Current Behavior

When setting `GRADLE_USER_HOME` to a directory with non-ASCII characters on Windows, any Gradle functionality that launches workers (i.e. test workers when running `gradlew test`) will fail.

### Expected Behavior

Setting `GRADLE_USER_HOME` to a directory with non-ascii characters should work.

### Context (optional)

This issue occurs because the JVM will read `@`-files using the native platform encoding. This ignores any and all system properties trying to change the default encoding to UTF-8.

The issue arises from the options file that is being written here: https://github.com/gradle/gradle/blob/6586eb029f216bf8acfbea7bcfcb7cde8518a689/subprojects/core/src/main/java/org/gradle/process/internal/worker/child/ApplicationClassesInSystemClassLoaderWorkerImplementationFactory.java#L103
(By the way, that suffix should be `.txt`, it's missing the period).

I can see using --info that it is passing such a file to the worker process:
```
C:\jdk21\bin\java.exe  [...] @Z:\Gradlé\.tmp\gradle-worker-classpath12335646236452669598txt
```

When inspecting that file, I see that it was written in UTF-8 encoding, but the JVM will read it using windows-1251, leading to a broken classpath.

Indeed, delving into `ArgsWriter` shows that there's an existing issue (https://github.com/gradle/gradle/issues/29303) that doesn't quite capture the problem: The JVM itself *also* doesn't use its default encoding to read @-files, it uses the unmodified default platform encoding (the one you'd get when you call Windows ANSI APIs).

**Solution**
The @-files for the JVM have to be written using the charset found in the `native.encoding` system property provided since Java 17. Before that, I think there is only the internal/undocumented `sun.jnu.encoding` property that contains the unmodified encoding when a user choses to override the default I/O encoding with `file.encoding`.

### Steps to Reproduce

1) Download the sample project https://docs.gradle.org/current/samples/sample_building_java_libraries.html
2) Set `GRADLE_USER_HOME` to something containing non-ascii characters (i.e. `C:\Lé Gradle`)
3) run `gradlew test`

### Gradle version

8.10

### Build scan URL (optional)

_No response_

Java 21
Windows
Gradle 8.10

windows-1251 default codepage

jvandort:
Previously, we started the daemon with a 'skeleton' classpath, then loaded the proper classpath upon boot. Instead, we load the correct classpath upfront and place that on the java classpath upon startup.

This opens the door to using AOT cache, which requires the classes-to-be-cached to live on the app classpath.

Fixes https://github.com/gradle/gradle/issues/30304

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle help

jvandort:
@bot-gradle test this on windows

jvandort:
@bot-gradle test act on windows

platforms/core-runtime/launcher/src/integTest/groovy/org/gradle/launcher/SystemClassLoaderTest.groovy
```diff
@@ -1,106 +0,0 @@
...
-import org.gradle.integtests.fixtures.AbstractIntegrationSpec
-import org.gradle.integtests.fixtures.executer.GradleContextualExecuter
-import org.gradle.internal.instrumentation.agent.AgentUtils
-import org.gradle.test.precondition.Requires
-import org.gradle.test.preconditions.TestExecutionPreconditions
-
-/**
- * Verifies that Gradle doesn't pollute the system class loader.
- *
- * This is important for plugins that need to use isolated class loaders to avoid conflicts.
```
jvandort, line 27: Plugins which need a "pure" isolated classpath can use `java.lang.ClassLoader.getSystemClassLoader().getParent()` on older Java versions, and `getPlatformClassLoader` on Java 9+

jvandort:
@bot-gradle test apt

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/process/ArgWriter.java
```diff
@@ -16,130 +16,130 @@
...
-            } else if (needsQuoting(str)) {
-                writer.print('\"');
-                writer.print(str);
-                writer.print('\"');
-            } else {
-                writer.print(str);
-            }
+    private void writeArgument(String arg, PrintWriter writer) {
+        if (backslashEscape) {
+            arg = arg.replace("\\", "\\\\").replace("\"", "\\\"");
```
octylFractal: 💭 Can we use `StringUtils.replaceEach` here to avoid double-allocating every arg when we are escaping?

platforms/core-runtime/client-services/src/main/java/org/gradle/launcher/daemon/client/DefaultDaemonStarter.java
```diff
@@ -201,6 +204,26 @@ public DaemonStartupInfo startDaemon(boolean singleUse) {
         );
     }
 
+    private String getOrCreateClasspathArgsFile(List<String> args) {
+        File argsFile = new File(daemonDir.getVersionedDir(), "daemon-classpath.txt");
+        if (!argsFile.exists()) {
+            File tmpFile = new File(daemonDir.getVersionedDir(), "daemon-classpath.tmp." + UUID.randomUUID());
```
octylFractal: :x: Let's not reinvent temp files and just use `Paths.createTempFile`. We can still scope to the correct directory with that.

platforms/core-runtime/client-services/src/main/java/org/gradle/launcher/daemon/client/DefaultDaemonStarter.java
```diff
@@ -201,6 +204,26 @@ public DaemonStartupInfo startDaemon(boolean singleUse) {
...
+            ArgWriter.javaStyle().generateArgsFile(args, tmpFile);
+            try {
+                Files.move(tmpFile.toPath(), argsFile.toPath(), StandardCopyOption.ATOMIC_MOVE);
+            } catch (AtomicMoveNotSupportedException e) {
+                try {
```
octylFractal: :x: I don't think this is the right way to handle this. If we can't guarantee atomicity, we can't share the arg file since it might be viewed partially written. We should just add a marker file that we'll need to check for to handle this case. Unfortunately we'll always need it, since we can't ask the file system if it supports atomic move :/.

Also, `REPLACE_EXISTING` isn't fool-proof: https://bugs.openjdk.org/browse/JDK-8323194

platforms/core-runtime/client-services/src/main/java/org/gradle/launcher/daemon/client/DefaultDaemonStarter.java
```diff
@@ -201,6 +204,26 @@ public DaemonStartupInfo startDaemon(boolean singleUse) {
...
+                    Files.move(tmpFile.toPath(), argsFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
+                } catch (IOException ex) {
+                    throw new UncheckedIOException(ex);
+                }
+            } catch (IOException e) {
+                throw new UncheckedIOException(e);
```
octylFractal: :x: Since not every operation is atomic, it's possible for two daemons to start trying to write this at the same time and then on some OSes, an exception will be thrown on move due to an existing file. We should check for instance of `FileAlreadyExistsException` and retry everything when that occurs (to handle the non-atomic-move case as well).

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/process/ArgWriter.java
```diff
@@ -16,130 +16,130 @@
...
+            return Collections.emptyList();
         }
-        argsFile.getParentFile().mkdirs();
+
+        return Collections.singletonList(doGenerateArgsFile(argsFile, args));
+    }
+
+    private String doGenerateArgsFile(File argsFile, List<String> args) {
         try {
-            // TODO(https://github.com/gradle/gradle/issues/29303)
```
octylFractal, line 91: 💅 Let's link this issue so it's closed if this rewrite also fixes it.

platforms/core-runtime/client-services/src/main/java/org/gradle/launcher/daemon/client/DefaultDaemonStarter.java
jvandort: No longer relevant

platforms/core-runtime/client-services/src/main/java/org/gradle/launcher/daemon/client/DefaultDaemonStarter.java
jvandort: No longer relevant

platforms/core-runtime/client-services/src/main/java/org/gradle/launcher/daemon/client/DefaultDaemonStarter.java
jvandort: After a conversation with sterling I've replaced this with including the runtime classpath as a `Class-Path` manifest attribute, which avoids the need for an args file

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/process/ArgWriter.java
jvandort: This issue was already closed in favor of the issue I linked in the commit. Its already linked in the PR

octylFractal: Ah, got it.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/process/ArgWriter.java
jvandort: Done

jvandort:
@bot-gradle test this

octylFractal (approved):

PR #37282, opened 2026-03-25
Update Kotlin embedded compiler docs to link to official Kotlin targets

VedantMadane:
Fixes #37017

Replaces the inline list of supported architectures with a link to the official [Kotlin/Native target support documentation](https://kotlinlang.org/docs/native-target-support.html), so the Gradle docs stay current without requiring manual updates as Kotlin adds new platform support.

### Changes

- Removed hardcoded platform/architecture list from `kotlin_dsl.adoc` prerequisites
- Added link to official Kotlin documentation for supported targets

### Addresses feedback from #37018

- **@eskatos**: Replaced inline platform list with a link to official Kotlin material, per [#37017 (comment)](https://github.com/gradle/gradle/issues/37017#issuecomment-4030254993)
- **@ov7a**: Added `Signed-off-by` trailer to comply with DCO

PR #37287, opened 2026-03-26
Update Kotlin embedded compiler docs to include ARM64 architecture

### LINKED ISSUE #37017: Update list of supported architectures of the Kotlin embedded compiler

JivinSardine:
The embedded Kotlin compiler now supports ARM64 (aarch64) architectures in addition to x86-64, on Linux, macOS, and Windows.

This updates the documentation which previously only mentioned x86-64 architectures.

Fixes #37017

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/kotlin_dsl.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/reference/dsl-apis/kotlin_dsl.adoc
@@ -29,7 +29,7 @@ TIP: If you are interested in migrating an existing Gradle build to the Kotlin D
 [[kotdsl:prerequisites]]
 == Prerequisites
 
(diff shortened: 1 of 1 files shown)


ov7a:
Thank you for the contribution!

There's already a PR to fix this:
- https://github.com/gradle/gradle/pull/37282

See also some history in
- https://github.com/gradle/gradle/pull/37018

Closing as a duplicate.

PR #36656, opened 2026-02-09
Add initial IP performance tests

jvandort:
We add cross version tests to verify IP does not regress between versions and add cross build tests to verify ip performs better than vintage.

We do not yet test specific scenarios like build logic abi change

### Contributor Checklist
- [ ] [Review Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md).
- [ ] Make sure that all commits are [signed off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff) to indicate that you agree to the terms of [Developer Certificate of Origin](https://developercertificate.org/).
- [ ] Make sure all contributed code can be distributed under the terms of the [Apache License 2.0](https://github.com/gradle/gradle/blob/master/LICENSE), e.g. the code was written by yourself or the original code is licensed under [a license compatible to Apache License 2.0](https://apache.org/legal/resolved.html).
- [ ] Check ["Allow edit from maintainers" option](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) in pull request so that additional changes can be pushed by Gradle team.
- [ ] Provide integration tests (under `<subproject>/src/integTest`) to verify changes from a user perspective.
- [ ] Provide unit tests (under `<subproject>/src/test`) to verify logic.
- [ ] Update User Guide, DSL Reference, and Javadoc for public-facing changes.
- [ ] Ensure that tests pass sanity check: `./gradlew sanityCheck`.
- [ ] Ensure that tests pass locally: `./gradlew <changed-subproject>:quickTest`.

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #37289, opened 2026-03-26
Improvements for Android Sync performance tests

alllex:
Preparation for: https://github.com/gradle/gradle/pull/36656

The **project-under-test is now always checked out into its own directory `project-checkout`** that is located under the working directory of the performance test. This is to ensure that auxiliary files that are required for or generated by Gradle Profiler are not part of the project-under-test. This is especially relevant for IDE Sync scenarios, because live IDE instances track and react to file changes on disk.

Related to the above, the dedicated gradle-user-home is also moved out of the project checkout directory to avoid IDEs and our own file-system watching reacting to a frequently changed directory.

The layout of the working directory for a single Android Studio sync invocation via Gradle Profiler is now:
```
- <working dir>
  - project-checkout
  - gradle-home
  - gradle-user-home
  - studio-sandbox
  - profile.log
```

Additionally, we now **always ensure that the project-under-test has its own git repository**. This is done in particular for IDE Sync scenarios to avoid a situation where an IDE's built-in git integration traverses the file system up the tree. Before these changes, the situation was that the IDE considered the project-under-test to be part of the enclosing gradle/gradle git repository.

Last but not least, we introduce a dedicated `AndroidSyncPerformanceTestFixture` and further clean up the performance test runner setup.

### DIFF
--- a/testing/performance/src/performanceTest/groovy/org/gradle/performance/regression/java/JavaConfigurationCachePerformanceTest.groovy
+++ b/testing/performance/src/performanceTest/groovy/org/gradle/performance/regression/java/JavaConfigurationCachePerformanceTest.groovy
@@ -35,7 +35,7 @@ import static org.junit.Assert.assertTrue
 
 class JavaConfigurationCachePerformanceTest extends AbstractCrossVersionPerformanceTest {
 
-    private static String CC_STATE_DIR = ".gradle/configuration-cache"
+    private static String configurationCacheStateDir = ".gradle/configuration-cache"
 
     def setup() {
         runner.minimumBaseVersion = "6.6"
@@ -53,6 +53,12 @@ class JavaConfigurationCachePerformanceTest extends AbstractCrossVersionPerforma
         runner.tasksToRun = ["assemble"]
         // use the deprecated property so it works with previous versions
         runner.args = ["-D${ConfigurationCacheOption.DEPRECATED_PROPERTY_NAME}=true"]
+
+        runner.addInterceptor { builder ->
+            builder.invocation {
+                buildLog(new File(workingDirectory, "build.log"))
+            }
+        }
 
         and:
         runner.useDaemon = daemon == hot
@@ -84,7 +90,7 @@ class JavaConfigurationCachePerformanceTest extends AbstractCrossVersionPerforma
             @Override
             void beforeBuild(BuildContext context) {
                 if (action == storing) {
-                    new TestFile(invocationSettings.projectDir).file(CC_STATE_DIR).deleteDir()
+                    new TestFile(invocationSettings.projectDir).file(configurationCacheStateDir).deleteDir()
                 }
             }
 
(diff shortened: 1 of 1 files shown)


alllex:
@bot-gradle test apt

6hundreds:
Does this PR contain code from https://github.com/gradle/gradle/pull/36656? Is it intended?

alllex:
@6hundreds, yes it does. It extracts a large part of it into this PR. Once this is merged, we'll update that other PR to only have the new IP tests added.

alllex:
@bot-gradle test apt

testing/internal-performance-testing/src/main/groovy/org/gradle/performance/AbstractCrossBuildPerformanceTest.groovy
```diff
@@ -73,12 +74,24 @@ class AbstractCrossBuildPerformanceTest extends AbstractPerformanceTest {
...
+        runner
+    }
+
     protected void defaultSpec(GradleBuildExperimentSpec.GradleBuilder builder) {
     }
 
     protected void finalizeSpec(GradleBuildExperimentSpec.GradleBuilder builder) {
     }
 
+    protected static BaselineVersion buildBaselineResults(CrossBuildPerformanceResults results, String name) {
```
jvandort: Looks like this method isn't called anywhere in this PR

testing/performance/src/performanceTest/groovy/org/gradle/performance/regression/java/JavaConfigurationCachePerformanceTest.groovy
```diff
@@ -93,7 +93,8 @@ class JavaConfigurationCachePerformanceTest extends AbstractCrossVersionPerforma
                     def tag = action == storing
                         ? "Calculating task graph as no (cached configuration|configuration cache) is available"
                         : "Reusing configuration cache"
-                    File buildLog = new File(invocationSettings.projectDir, "profile.log")
+                    File workingDir = invocationSettings.projectDir.parentFile
```
jvandort: Any way we can easily avoid the `parentFile` call here and instead wire the working dir directly?

testing/internal-performance-testing/src/main/groovy/org/gradle/performance/AbstractCrossBuildPerformanceTest.groovy
alllex: Removed

testing/performance/src/performanceTest/groovy/org/gradle/performance/regression/java/JavaConfigurationCachePerformanceTest.groovy
alllex: I found a way of specifying a `buildLog` file, which will be used to write all stdout/err of the Gradle invocation. The check now reads this file instead of the entire `profile.log`.

Also, apparently, the location of `profile.log` is not customizable today with Gradle Profiler, even with the direct access to the APIs

alllex:
@bot-gradle test apt

PR #37290, opened 2026-03-26
Sort cache keys only when serializing instrumentation analysis

hegyibalint:
## Solution

Sorting is a serialization concern. Introduce `SortedMapSerializer` and `SortedSetSerializer` — standalone serializers that sort via `Arrays.sort` at write time, with wire format identical to `MapSerializer`/`SetSerializer`. Callers switch from `TreeMap`/`TreeSet` to plain `HashMap`/`HashSet`.

### Changed files

- **`SortedMapSerializer`** / **`SortedSetSerializer`** — new, in `serialization` module
- **`DefaultInstrumentationAnalysisSerializer`** — uses sorted serializers for `writeTypesMap`
- **`InstrumentationAnalysisTransform`** — `TreeMap` → `HashMap`, `TreeSet` → `HashSet`, `ImmutableSortedSet` → `Collectors.toSet()`
- **`MergeInstrumentationAnalysisTransform`** — `TreeMap` → `HashMap`, drops `ImmutableSortedSet.copyOf` wrapping

## Results

| | Median | vs Baseline |
|---|---|---|
| Baseline (master) | 35.22 s | — |
| `gradle10/provider-api-migration` (no fix) | 38.22 s | **+7.86%** |
| Migration + this fix | 35.90 s | **+1.94%** |

JMH microbenchmark confirms the sorted serializers are **1.2–1.6x faster** than TreeMap+TreeSet for the write path at realistic sizes (100–1000 keys, 3–10 values per key).

## Notes

- The remaining ~2% gap vs baseline is within run-to-run noise on shared CI infra (stdev ~0.4s on a 35s scenario). Earlier runs of the same approach measured +0.55% and +1.69%.
- The existing integration test `BuildScriptClasspathInstrumentationIntegrationTest.should analyze plugin artifacts` validates serialization ordering end-to-end.
- `SortedMapSerializer`/`SortedSetSerializer` are general-purpose and available for other deterministic serialization needs.

### DIFF
--- a/platforms/core-runtime/serialization/src/main/java/org/gradle/internal/serialize/SortedMapSerializer.java
+++ b/platforms/core-runtime/serialization/src/main/java/org/gradle/internal/serialize/SortedMapSerializer.java
@@ -20,6 +20,8 @@ import com.google.common.collect.Maps;
 import org.jspecify.annotations.NullMarked;
 
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -45,7 +47,8 @@ public class SortedMapSerializer<K extends Comparable<K>, V> extends AbstractSer
     @Override
     public void write(Encoder encoder, Map<K, V> value) throws Exception {
         List<K> sortedKeys = new ArrayList<>(value.keySet());
-        sortedKeys.sort(null);
+        Collections.sort(sortedKeys);
+
         encoder.writeInt(sortedKeys.size());
         for (K key : sortedKeys) {
             keySerializer.write(encoder, key);


hegyibalint:
@bot-gradle test APT

platforms/core-runtime/serialization/src/main/java/org/gradle/internal/serialize/SortedMapSerializer.java
```diff
@@ -0,0 +1,84 @@
...
+        this.valueSerializer = valueSerializer;
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    @Override
+    public void write(Encoder encoder, Map<K, V> value) throws Exception {
+        // We cannot create a K[] directly due to type erasure; Comparable[] is the
+        // closest array type available since K extends Comparable<K>.
+        K[] sortedKeys = (K[]) value.keySet().toArray(new Comparable[0]);
+        Arrays.sort(sortedKeys);
```
octylFractal: 💭 Is there some advantage to this over:
```suggestion
        List<K> sortedKeys = new ArrayList<>(value.keySet());
        sortedKeys.sort();
```
It would remove the need for the cast, comment, and extra (though likely JIT-removed) array creation. I suppose it allocates one extra object, but I don't know if that's really worth anything compared to keeping the code more readable.

platforms/core-runtime/serialization/src/main/java/org/gradle/internal/serialize/SortedSetSerializer.java
```diff
@@ -0,0 +1,78 @@
...
+        this.entrySerializer = entrySerializer;
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    @Override
+    public void write(Encoder encoder, Set<T> value) throws Exception {
+        // We cannot create a T[] directly due to type erasure; Comparable[] is the
+        // closest array type available since T extends Comparable<T>.
+        T[] sorted = (T[]) value.toArray(new Comparable[0]);
+        Arrays.sort(sorted);
```
octylFractal: 💭 Same here, using a `List` reads better.

platforms/core-runtime/serialization/src/main/java/org/gradle/internal/serialize/SortedMapSerializer.java
```diff
@@ -0,0 +1,84 @@
...
+        for (K key : sortedKeys) {
+            keySerializer.write(encoder, key);
+            valueSerializer.write(encoder, value.get(key));
+        }
+    }
+
+    @Override
+    public Map<K, V> read(Decoder decoder) throws Exception {
+        int size = decoder.readInt();
+        Map<K, V> map = new LinkedHashMap<>(size);
```
octylFractal: 💭 I wonder if being consistent here and using `Maps.newLinkedHashMapWithExpectedSize` would be better. It is actually different due to some quirks in the JDK Map implementations, see https://github.com/google/guava/blob/e7a6a200d7171c78774d5805be77351b7b84eab5/guava/src/com/google/common/collect/Maps.java#L287-L308 which is called by the `Maps`/`Sets` creation functions.

platforms/core-runtime/serialization/src/main/java/org/gradle/internal/serialize/SortedMapSerializer.java
hegyibalint: I will see how this performs. I have a JMH bench to test these solutions!

subprojects/core/src/main/java/org/gradle/api/internal/initialization/transform/MergeInstrumentationAnalysisTransform.java
```diff
@@ -115,11 +114,11 @@ private void doMergeAndOutputAnalysis(File input, TransformOutputs outputs) {
         InstrumentationTypeRegistry registry = getInstrumentationTypeRegistry();
 
         InstrumentationDependencyAnalysis data = serializer.readDependencyAnalysis(input);
-        Map<String, Set<String>> dependenciesSuperTypes = new TreeMap<>();
+        Map<String, Set<String>> dependenciesSuperTypes = new HashMap<>();
         for (String className : data.getDependencies().keySet()) {
             Set<String> superTypes = registry.getSuperTypes(className);
             if (!superTypes.isEmpty()) {
-                dependenciesSuperTypes.put(className, new TreeSet<>(superTypes));
+                dependenciesSuperTypes.put(className, superTypes);
```
asodja, line 121: ❓ Is it safe to not do a copy?

subprojects/core/src/main/java/org/gradle/api/internal/initialization/transform/InstrumentationAnalysisTransform.java
```diff
@@ -176,7 +175,7 @@ private InstrumentationArtifactMetadata getArtifactMetadata(File artifact) {
     }
 
     private static Map<String, Set<String>> toMapWithKeys(Set<String> keys) {
-        TreeMap<String, Set<String>> map = new TreeMap<>();
+        HashMap<String, Set<String>> map = new HashMap<>(keys.size());
```
asodja: ❗  Note, that this will always resize a map, since map will resize when it will get to 0.75 of capacity. It's better to use `Maps.newHashMapWithExpectedSize(keys.size())` that calculates the size, if you don't want map resize

subprojects/core/src/main/java/org/gradle/api/internal/initialization/transform/utils/DefaultInstrumentationAnalysisSerializer.java
```diff
@@ -24,6 +24,8 @@
 import org.gradle.internal.serialize.HierarchicalNameSerializer;
 import org.gradle.internal.serialize.MapSerializer;
 import org.gradle.internal.serialize.SetSerializer;
+import org.gradle.internal.serialize.SortedMapSerializer;
```
asodja, line 27: ❓ What has changed on the provider-api branch infra related? It would be great if we can merge to master to have less changes on the Provider API branch (and also get optimization benefits on master)

hegyibalint: Excellent question, I guess not much. I will try to rebase to `master`, and see if we have any blockers

subprojects/core/src/main/java/org/gradle/api/internal/initialization/transform/InstrumentationAnalysisTransform.java
hegyibalint: Yeah, probably you mention the same thing as Octavia: https://github.com/gradle/gradle/pull/37290/changes#r2996438910

platforms/core-runtime/serialization/src/main/java/org/gradle/internal/serialize/SortedMapSerializer.java
hegyibalint: Thanks for the link, now I see how this works!
Eventually I went with what you linked, `Maps.newLinkedHashMapWithExpectedSize`

subprojects/core/src/main/java/org/gradle/api/internal/initialization/transform/InstrumentationAnalysisTransform.java
hegyibalint: See that comment, I've decided to go with the proposed `Maps.newLinkedHashMapWithExpectedSize`

platforms/core-runtime/serialization/src/main/java/org/gradle/internal/serialize/SortedMapSerializer.java
hegyibalint: Made some JMH benchmarks with claude and It seems `ArrayList` indeed does deliver the same performance!

subprojects/core/src/main/java/org/gradle/api/internal/initialization/transform/utils/DefaultInstrumentationAnalysisSerializer.java
hegyibalint: Done!

subprojects/core/src/main/java/org/gradle/api/internal/initialization/transform/MergeInstrumentationAnalysisTransform.java
hegyibalint: At the end of the method, we write and dispose of this map, so I think the data handling _inside_ the method is fine.

However, are you worried that the underlying `registry.getSuperTypes(className)` would change?

hegyibalint: I've made CC look up all backing implementations of registries, and point out any chance where the underlying map would be mutated. It sounds like this is _not_ the case, and we should be getting a stable view of the data,

hegyibalint:
@bot-gradle test this

platforms/core-runtime/serialization/src/main/java/org/gradle/internal/serialize/SortedMapSerializer.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public SortedMapSerializer(Serializer<K> keySerializer, Serializer<V> valueSerializer) {
+        this.keySerializer = keySerializer;
+        this.valueSerializer = valueSerializer;
+    }
+
+    @Override
+    public void write(Encoder encoder, Map<K, V> value) throws Exception {
+        List<K> sortedKeys = new ArrayList<>(value.keySet());
+        sortedKeys.sort(null);
```
octylFractal: 💅 This is minor, but I would prefer to use `Collections.sort(sortedKeys)` over `.sort(null)` for type-safety. They're completely equivalent otherwise (the first just calls the second).

hegyibalint: I understand what you are trying to say; I don't think "type-safety" is the correct word here, it rather the fact that you expect `Collections.sort(array)` to mutate the array, meanwhile `array.sort(null)` is much more ambigous in its behavior

octylFractal: It is actually about type safety -- `array.sort(null)` will not check that the key type is `Comparable`, but `Collections.sort` will.

Issue #35132, opened 2025-09-25
Deprecate project.properties


liutikas:
### Expected Behavior

Deprecate project.properties is marked deprecated so users know not to use it

### Context

project.properties is problematic for various reasons:
- it causes performance issues as it is eager
- it is inherits properties from parent projects
- it falls back to calling project.ext.get()
- given the tracking of properties for CC invalidation this method defeats that optimization entirely

lkasso:
This issue needs a decision from the team responsible for that area. They have been informed. Response time may vary.

abstratt:
This feature request is in the backlog of the relevant team and is prioritized by them.

***

This API is not aligned with CC or Isolated Projects (as properties may reach into parent projects).

PR #37205, opened 2026-03-19
WIP: Deprecate accessing parent properties

jvandort:
## Summary

Deprecate implicit property and method lookups from parent projects in Gradle Vintage mode, targeting removal in Gradle 10. Improve the Isolated Projects violation for the same scenarios, and introduce a `NO_IMPLICIT_PARENT_PROPERTY_LOOKUP` feature preview that opts into the Gradle 10 target behavior early for both Vintage and Isolated Projects.

### Context

Gradle's Groovy-based dynamic property resolution walks up the project hierarchy: when a child project references `foo` (bare identifier, `property("foo")`, `findProperty("foo")`, `hasProperty("foo")`, or `getProperty("foo")`) and `foo` isn't defined locally, Gradle searches the parent project chain until it finds a match. The same applies to methods defined in parent build scripts.

This implicit inheritance:
- Creates hidden (and often unintended) coupling between projects.
- Makes build logic hard to understand — it's unclear where a value comes from.
- Is fundamentally incompatible with Isolated Projects (IP), which reports a violation on any such access.
- Leads to false-positive IP violations: a `findProperty("local")` call that was never meant to reach the parent still triggers a violation because the resolution walks the chain mechanically.

Design spec: [Remove Implicit Parent Project Property Lookup — GBT Spec](https://docs.google.com/document/d/1AQ_IS9GCiC3kFNW_lrh8thUZSOLLgdyQy626cFu1MtI/edit).

### What this PR does

**Deprecation in Vintage mode (default 9.6 behavior):**
- Emits a deprecation warning whenever a property or method lookup on a child project successfully resolves from a parent.
- Distinguishes explicit API calls (`property()`, `findProperty()`, `hasProperty()`, `getProperty()` on `Project`) from Groovy dynamic resolution via a new `CallerContext` type, so messages are accurate to how the user actually triggered the lookup.
- All deprecation messages link to the new upgrade guide section `deprecated_accessing_parent_project_properties`.

**Isolated Projects mode improvements:**
- The redundant deprecation warning is suppressed in IP mode (the violation already covers it).
- The IP violation message now includes the property/method name and the calling API context (e.g., `"via 'property()'"`, `"via 'hasProperty()'"`, `"via 'val ... by project'"`, or `"via 'foo' or getProperty('foo') in build script"`).
- IP violations now link to the same upgrade guide section.
- Kotlin DSL `by project` delegation is tracked with a dedicated caller context.

[...]

PR #37291, opened 2026-03-26
Deprecate getProperties() on Project and scripts

reinsch82:
Fixes https://github.com/gradle/gradle/issues/35132

## Summary
- Deprecate `Project.getProperties()` and script `getProperties()` as preparation for project isolation
- These APIs return a merged view of all properties including inherited parent project properties, which conflicts with the isolation model
- Extracted from #37205 to keep the `getProperties()` deprecation separate from the parent property lookup deprecation

## Changes
- Add deprecation warnings in `DefaultProject.getProperties()` and `BasicScript.getProperties()`
- Add `@Deprecated` annotations to `getProperties()` throughout the `DynamicObject` hierarchy
- Remove `DeprecationLogger.whileDisabled` wrappers in lookup routines (deprecation now emitted at API entry points)
- Add upgrade guide sections (`deprecated_project_get_properties`, `deprecated_script_get_properties`)
- Add integration tests for Kotlin DSL, Groovy build script, settings script, and init script

## Test plan
- [x] `compileAll` passes
- [x] `sanityCheck` passes
- [x] `GetPropertiesDeprecationIntegrationTest` — all 4 tests pass
- [x] CI

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/build_environment.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/build_environment.adoc
@@ -185,6 +185,30 @@ $ export ORG_GRADLE_PROJECT_myProperty='Hi, world'
 +
 This is typically the preferred method for supplying project properties, especially secrets, to unattended builds like those running on CI servers.
 
+=== Accessing a project property
+
+The recommended way to access a project property is through the <<lazy_configuration.adoc#lazy_properties,Provider API>>:
+
+[source,kotlin]
+----
+val myProperty: Provider<String> = providers.gradleProperty("myProperty")
+----
+
+The returned `Provider` is lazy and compatible with the <<configuration_cache_requirements.adoc#config_cache:requirements:reading_sys_props_and_env_vars,Configuration Cache>>.
+The provider resolves the property from the build-level sources listed above: command-line `-P` arguments, `org.gradle.project.pass:[*]` system properties, `ORG_GRADLE_PROJECT_pass:[*]` environment variables, and `gradle.properties` files in the Gradle User Home, build root, and Gradle installation directories.
+
+[NOTE]
+====
+`providers.gradleProperty()` does not include properties from `gradle.properties` files in subproject directories, nor extra properties or other properties set dynamically on individual `Project` instances.
+To access those, use `project.findProperty("name")` or extra properties directly.
+====
+
+For other ways to access properties, see the following methods on the link:{groovyDslPath}/org.gradle.api.Project.html[Project] object:
+
+- `findProperty("name")` — returns the value or `null`, checking dynamically configured containers, like extra properties, and parent projects as well.
+- `property("name")` — returns the value or throws if the property is missing
+- `hasProperty("name")` — checks if the property exists
+
 It is possible to change the behavior of a task based on project properties specified at invocation time.
 Suppose you’d like to ensure release builds are only triggered by CI.
 A simple way to handle this is through an `isCI` project property:
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -93,6 +93,131 @@ Problem severity can no longer be set explicitly when creating a new instance.
 Instead, it is determined by the reporting method: `ProblemReporter.report()` produces warnings and `ProblemReporter.throwing()` produces errors.
 Calling `.severity()` on `ProblemSpec` is now a no-op and will be removed in Gradle 10.0.
 
+[[deprecated_get_properties]]
+==== Deprecation of `getProperties()`
+
+The following usages of `getProperties()` have been deprecated and will be removed in Gradle 10.0.0:
+
+- link:{javadocPath}/org/gradle/api/Project.html#getProperties()[`Project.getProperties()`] — calling `getProperties()` or accessing the `properties` property on a `Project` instance.
+- Calling `getProperties()` (or accessing the `properties` property) on a build script, settings script, or init script. This is a separate API on the script base class, but has the same problems.
+
+This method, which historically existed for convenience, makes build logic harder to understand and navigate due to the untyped nature of the properties and the fact that they are sourced from many different locations and containers.
+Using this method also hurts performance of builds by eagerly resolving properties from the environment, reducing the potential for more frequent Configuration Cache hits.
+The hierarchical look-up of properties through the parent projects chain creates implicit coupling of the mutable state between projects.
+This impedes the potential of scalability in the future model, where projects are <<sharing_build_logic_between_subprojects.adoc#sec:convention_plugins_vs_cross_configuration,isolated from each other>>.
+
+To access <<build_environment.adoc#sec:project_properties,project properties>>, prefer `providers.gradleProperty("name")` which returns a type-safe `Provider` and is compatible with lazy configuration.
+Note that `providers.gradleProperty()` resolves properties at the build level — it does not include properties from `gradle.properties` files in subproject directories, nor extra properties or other properties set dynamically on individual projects.
+If you need to access <<writing_build_scripts_intermediate.adoc#sec:extra_properties,extra properties>>, use `ext` properties directly.
+
+The following table shows common migration scenarios for project properties:
+
+[cols="1a,1a", options="header"]
+|===
+| Before (deprecated) | After (recommended)
+
+| Accessing a single project property:
+[source,kotlin]
+----
+val value = project.properties["myProp"]
+----
+[source,groovy]
+----
+def value = project.properties.myProp
+----
+| Wire the provider as a task property for better Configuration Cache support:
+[source,kotlin]
+----
+val myProp: Provider<String> =
+    providers.gradleProperty("myProp")
+----
+If you need the value immediately:
+[source,groovy]
+----
+def value = providers.gradleProperty("myProp")
+    .orNull
+----
+
+| Accessing a property with a default value (Kotlin, Java, or Groovy):
+[source,kotlin]
+----
+val value = project.properties
+    .getOrDefault("myProp", "default")
+----
+Accessing a property with a default value (Groovy, idiomatic):
+[source,groovy]
+----
+def value = project.properties["myProp"] ?: "default"
+----
+| Use `providers.gradleProperty()` with `getOrElse()`:
+[source,kotlin]
+----
+val value = providers
+    .gradleProperty("myProp")
+    .getOrElse("default")
+----
+[source,groovy]
+----
+def value = providers.gradleProperty("myProp")
+    .getOrElse("default")
+----
+
+| Kotlin property delegation:
+[source,kotlin]
+----
+val myProp: String by project.properties
+----
+| Delegate from `project` directly:
+[source,kotlin]
+----
+val myProp: String by project
+----
+
+| Forwarding properties as system properties:
+[source,kotlin]
+----
+test {
+    systemProperties(
+        project.properties.filter {
+            it.key.startsWith("test.")
+        }
+    )
+}
+----
+| Use `providers.gradlePropertiesPrefixedBy()`:
+[source,kotlin]
+----
+test {
+    systemProperties(
+        providers
+            .gradlePropertiesPrefixedBy("test.")
+            .get()
+    )
+}
+----
+
+| Token replacement in resource files (common in Minecraft mods, Spring Boot, etc.):
+[source,groovy]
+----
+processResources {
+    filesMatching("plugin.yml") {
+        expand project.properties
+    }
+}
+----
+| Build an explicit map of only the properties you need:
+[source,groovy]
+----
+processResources {
+    filesMatching("plugin.yml") {
+        expand(
+            "version": project.version,
+            "name": project.name
+        )
+    }
+}
+|===
+
 [[changes_9.5.0]]
 == Upgrading from 9.4.0 and earlier
 
@@ -186,21 +311,6 @@ This is a result of the link:https://github.com/gradle/gradle/pull/36510[Windows
 If you are using a custom start script template that references the exit environment variable, update your template to remove this dependency.
 There is no replacement API, the exit environment variable concept has been removed entirely.
 
-[[deprecated_project_get_properties]]
-==== Deprecation of `Project.getProperties()`
-
-link:{javadocPath}/org/gradle/api/Project.html#getProperties()[`Project.getProperties()`] has been deprecated and will be removed in Gradle 10.0.0.
-
-This method returns a merged view of all properties including inherited parent project properties, which conflicts with the project isolation model.
-
-To access a specific property, use `project.findProperty("name")` or `providers.gradleProperty("name")` instead.
-
-[[deprecated_script_get_properties]]
-==== Deprecation of calling `getProperties()` on a script
-
-Calling `getProperties()` (or accessing the `properties` property) on a build script, settings script, or init script has been deprecated and will be removed in Gradle 10.0.0.
-
-To access specific properties, use `ext` properties or `providers.gradleProperty("name")` instead.
 
 [[changes_9.4.0]]
 == Upgrading from 9.3.1 and earlier
--- a/subprojects/core-api/src/main/java/org/gradle/api/Project.java
+++ b/subprojects/core-api/src/main/java/org/gradle/api/Project.java
@@ -1477,9 +1477,9 @@ public interface Project extends Comparable<Project>, ExtensionAware, PluginAwar
      * are available for a project.</p>
      *
      * @return A map from property name to value.
-     * @deprecated This method will be removed in Gradle 10.0. Use {@link #findProperty(String)} or
+     * @deprecated This method will be removed in Gradle 10.0.0. Use {@link #findProperty(String)} or
      * {@link org.gradle.api.provider.ProviderFactory#gradleProperty(String)} instead.
-     * See the <a href="https://docs.gradle.org/current/userguide/upgrading_version_9.html#deprecated_project_get_properties">upgrade guide</a> for more details.
+     * See the <a href="https://docs.gradle.org/current/userguide/upgrading_version_9.html#deprecated_get_properties">upgrade guide</a> for more details.
      */
     @Deprecated
     @HiddenInDefinition
--- a/subprojects/core/src/integTest/groovy/org/gradle/api/SettingsDslIntegrationSpec.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/api/SettingsDslIntegrationSpec.groovy
@@ -36,10 +36,13 @@ class SettingsDslIntegrationSpec extends AbstractIntegrationSpec {
     def "can dynamically access properties"() {
 
         given:
-        file("gradle.properties") << "someProjectProperty=true"
+        propertiesFile << "someProjectProperty=true"
         settingsFile << """
-            if (findProperty('someProjectProperty') == 'true') {
-                println('signal')
+            if (getProperty('someProjectProperty') == 'true') {
+                println('getProperty worked')
+            }
+            if (someProjectProperty == 'true') {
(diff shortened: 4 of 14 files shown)


reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

platforms/documentation/docs/src/snippets/bestPractices/avoidGradlePropertiesInSubProjects-avoid/groovy/app/build.gradle
```diff
@@ -1,7 +1,7 @@
 // This file is located in /app
 tasks.register("printProperties") { // <1>
-    def propA = project.properties.get("propertyA") // <2>
-    def propB = project.properties.get("propertyB")
+    def propA = project.getProperty("propertyA") // <2>
```
alllex: ❌  There is no such API on `Project` interface. Is it a Groovy thing?

Also, `property()` is not equivalent, because it will throw if the property is missing. The replacement should be `findProperty()`

There seem to be many changes like this in the PR

subprojects/core/src/integTest/groovy/org/gradle/api/SettingsDslIntegrationSpec.groovy
```diff
@@ -38,7 +38,7 @@ class SettingsDslIntegrationSpec extends AbstractIntegrationSpec {
         given:
         file("gradle.properties") << "someProjectProperty=true"
         settingsFile << """
-            if (properties.someProjectProperty == 'true') {
+            if (getProperty('someProjectProperty') == 'true') {
```
alllex, line 41: 🤔  There is no `getProperty()` method on `Project`. Is this a Groovy thing?

subprojects/core/src/main/java/org/gradle/api/internal/project/DynamicLookupRoutine.java
```diff
@@ -34,6 +34,7 @@ public interface DynamicLookupRoutine {
     @Nullable Object findProperty(DynamicObject receiver, String propertyName);
     void setProperty(DynamicObject receiver, String name, @Nullable Object value);
     boolean hasProperty(DynamicObject receiver, String propertyName);
+    @Deprecated
```
alllex: ❓  Is it useful to deprecate internal APIs?

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
```diff
@@ -1176,8 +1176,13 @@ public boolean hasProperty(String propertyName) {
         return dynamicLookupRoutine.hasProperty(extensibleDynamicObject, propertyName);
     }
 
+    @SuppressWarnings("deprecation")
     @Override
     public Map<String, ? extends @Nullable Object> getProperties() {
```
alllex, line 1181: ❓ It seems we are missing the source-level deprecation on the `Project` interface

It would be nice to link to the userguide upgrade section, so that users interacting with the code in the IDE can learn why it's deprecated and what to use instead, without necessarily running builds.

subprojects/core/src/integTest/groovy/org/gradle/api/internal/project/GetPropertiesDeprecationIntegrationTest.groovy
```diff
@@ -0,0 +1,57 @@
...
+
+        expect:
+        executer.expectDocumentedDeprecationWarning(expectedDeprecation)
+        succeeds("help")
+
+        where:
+        scriptType               | scriptFile            | expectedDeprecation
+        "Kotlin build script"    | "build.gradle.kts"    | PROJECT_DEPRECATION
+        "Groovy build script"    | "build.gradle"        | SCRIPT_DEPRECATION
+        "Groovy settings script" | "settings.gradle"     | SCRIPT_DEPRECATION
```
alllex, line 45: 💅  I'd add a short comment that this API is not accessible in `settings.gradle.kts`

subprojects/core/src/integTest/groovy/org/gradle/api/internal/project/GetPropertiesDeprecationIntegrationTest.groovy
```diff
@@ -0,0 +1,57 @@
...
+
+    def "accessing properties on an init script is deprecated"() {
+        def initScript = file("init.gradle") << "properties"
+
+        when:
+        executer.expectDocumentedDeprecationWarning(SCRIPT_DEPRECATION)
+        fails("help", "-I", initScript.absolutePath)
+
+        then:
+        failure.assertHasCause("The default project is not yet available for build")
```
alllex, line 56: 👍  Interesting discovery

subprojects/core/src/integTest/groovy/org/gradle/api/SettingsDslIntegrationSpec.groovy
reinsch82: I guess it works because it is on `DefaultProject`

subprojects/core/src/main/java/org/gradle/api/internal/project/DynamicLookupRoutine.java
jvandort: I think so. If that internal API relies on or implements behavior that is deprecated, it allows internal developers to know if some code is being removed in the next major.

I usually deprecate all the internal classes/methods related to a feature when I deprecate that feature

platforms/core-configuration/base-diagnostics/src/main/java/org/gradle/api/tasks/diagnostics/PropertyReportTask.java
```diff
@@ -73,9 +74,10 @@ protected PropertyReportModel calculateReportModelFor(Project project) {
         return computePropertyReportModel(project);
     }
 
+    @SuppressWarnings("deprecation")
     private PropertyReportTask.PropertyReportModel computePropertyReportModel(Project project) {
         PropertyReportModel model = new PropertyReportModel();
-        Map<String, ?> projectProperties = project.getProperties();
+        Map<String, ?> projectProperties = DeprecationLogger.whileDisabled(project::getProperties);
```
big-guy, line 80: What's the plan for this once this method is removed?  Would we still keep it as an internal thing or would we still have a getProperties that just returns something else?

subprojects/core/src/integTest/groovy/org/gradle/api/SettingsDslIntegrationSpec.groovy
reinsch82: My revised understanding says: It works because of `getProperty` on `BasicScript`.
So to me this makes sense.

alllex: I see. Then it's okay to test `getProperty` as well, but I'd argue that it's not the primary pattern that test should be exercising. Judging from the test name, it should test the most common ways of accessing a dynamic property. In Groovy, that would be simply `someProjectProperty` mention. We can keep the `getProperty()` variant for completeness, but I think it's secondary.

subprojects/core-api/src/main/java/org/gradle/api/Project.java
```diff
@@ -1477,7 +1477,11 @@ void afterEvaluate(@DelegatesTo(Project.class)
      * are available for a project.</p>
      *
      * @return A map from property name to value.
+     * @deprecated This method will be removed in Gradle 10.0. Use {@link #findProperty(String)} or
```
alllex: 💅 It's `10.0.0`

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -186,6 +186,114 @@ This is a result of the link:https://github.com/gradle/gradle/pull/36510[Windows
...
+link:{javadocPath}/org/gradle/api/Project.html#getProperties()[`Project.getProperties()`] has been deprecated and will be removed in Gradle 10.0.0.
+
+This method returns a merged view of all properties including inherited parent project properties, which conflicts with the project isolation model.
+
+To access a specific property, prefer `providers.gradleProperty("name")` which returns a type-safe `Provider` and is compatible with lazy configuration.
+Note that `providers.gradleProperty()` does not look through the project hierarchy — it only returns properties set directly on the current project.
+If you need hierarchical property lookup (searching parent projects), use `project.findProperty("name")` instead, but prefer `providers.gradleProperty()` when the hierarchical lookup is not needed.
+
+The following table shows common migration scenarios:
+
```
reinsch82, line 114: the table

Image

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -186,6 +186,114 @@ This is a result of the link:https://github.com/gradle/gradle/pull/36510[Windows
 If you are using a custom start script template that references the exit environment variable, update your template to remove this dependency.
 There is no replacement API, the exit environment variable concept has been removed entirely.
 
+[[deprecated_project_get_properties]]
+==== Deprecation of `Project.getProperties()`
+
+link:{javadocPath}/org/gradle/api/Project.html#getProperties()[`Project.getProperties()`] has been deprecated and will be removed in Gradle 10.0.0.
+
+This method returns a merged view of all properties including inherited parent project properties, which conflicts with the project isolation model.
```
alllex: 💭 We should be a bit more thorough with the reasoning on why we are removing the method, given its tenure.

Here is my wording, but it likely can be improved and also enriched with links to the documentation sections:

> This method, which historically existed for convenience, makes build logic harder to understand and navigate, due to the untyped nature of the properties and the fact that they are sources from many different locations and containers. Using this method also hurts performance of builds by eagerly resolving properties from the environment, reducing the potential for more frequent Configuration Cache hits. The hierarchical look-up of properties through the parent projects chain creates implicit coupling of the mutable state between projects. This impedes the potential of scalability in the future model, where projects are isolated from each other.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -186,6 +186,114 @@ This is a result of the link:https://github.com/gradle/gradle/pull/36510[Windows
...
+            it.key.startsWith("test.")
+        }
+    )
+}
+----
+| Use `providers.gradlePropertiesPrefixedBy()`:
+[source,kotlin]
+----
+test {
+    systemProperties(
```
alllex: ❓ Is there such an overload of `systemProperties()` for which this would work?

While it might not be ideal, but until the Provider Migration happens, we'd have to realize the provider map eagerly here with `.get()`

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -186,6 +186,114 @@ This is a result of the link:https://github.com/gradle/gradle/pull/36510[Windows
 If you are using a custom start script template that references the exit environment variable, update your template to remove this dependency.
 There is no replacement API, the exit environment variable concept has been removed entirely.
 
+[[deprecated_project_get_properties]]
+==== Deprecation of `Project.getProperties()`
```
alllex: ❓ Is this the right section on the page?

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -186,6 +186,114 @@ This is a result of the link:https://github.com/gradle/gradle/pull/36510[Windows
...
+| Accessing a single property:
+[source,groovy]
+----
+def value = project.properties["myProp"]
+def value = project.properties.myProp
+----
+| Use `providers.gradleProperty()` or `findProperty()`:
+[source,groovy]
+----
+def value = providers.gradleProperty("myProp").orNull
```
alllex: 💅 Better to not advise immediately resolving the Provider and suggest wiring it as a property, if used by a task to improve CC hit rates.

I think it would also be better to be explicit with the type and use Kotlin:
```kotlin
val valueProvider: Provider<String> = ...
// don't resolve the provider immediately, unless you need its value
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -186,6 +186,114 @@ This is a result of the link:https://github.com/gradle/gradle/pull/36510[Windows
...
+Note that `providers.gradleProperty()` does not look through the project hierarchy — it only returns properties set directly on the current project.
+If you need hierarchical property lookup (searching parent projects), use `project.findProperty("name")` instead, but prefer `providers.gradleProperty()` when the hierarchical lookup is not needed.
+
+The following table shows common migration scenarios:
+
+[cols="1a,1a", options="header"]
+|===
+| Before (deprecated) | After (recommended)
+
+| Accessing a single property:
```
alllex: 💭 I think we should separate cases of accessing Project properties (https://docs.gradle.org/current/userguide/build_environment.html#sec:project_properties) from dynamically set properties, like extra properties. 

I'd also make this separation clearer in the paragraphs above.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -186,6 +186,114 @@ This is a result of the link:https://github.com/gradle/gradle/pull/36510[Windows
...
+----
+val myProp: String by project.properties
+----
+| Delegate from `project` directly:
+[source,kotlin]
+----
+val myProp: String by project
+----
+
+| Token replacement in resource files:
```
alllex: 💭 This is an interesting use case, but seems niche, so I'd put it last and add more context to its description

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -186,6 +186,114 @@ This is a result of the link:https://github.com/gradle/gradle/pull/36510[Windows
...
 
+[[deprecated_project_get_properties]]
+==== Deprecation of `Project.getProperties()`
+
+link:{javadocPath}/org/gradle/api/Project.html#getProperties()[`Project.getProperties()`] has been deprecated and will be removed in Gradle 10.0.0.
+
+This method returns a merged view of all properties including inherited parent project properties, which conflicts with the project isolation model.
+
+To access a specific property, prefer `providers.gradleProperty("name")` which returns a type-safe `Provider` and is compatible with lazy configuration.
+Note that `providers.gradleProperty()` does not look through the project hierarchy — it only returns properties set directly on the current project.
```
alllex: 🤔

> it only returns properties set directly on the current project.

I don't think this is technically accurate. Can we be more precise?

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -186,6 +186,114 @@ This is a result of the link:https://github.com/gradle/gradle/pull/36510[Windows
...
+----
+def value = providers.gradleProperty("myProp").orNull
+def value = findProperty("myProp")
+----
+
+| Accessing a property with a default value:
+[source,kotlin]
+----
+val value = project.properties
+    .getOrDefault("myProp", "default")
```
alllex, line 145: 💭 This is Kotlin specific, because `getOrDefault` is an extension from Kotlin stdlib. It would be nice to add a Groovy variant as well, which would be `properties["foo"] ?: "bar"`

subprojects/core/src/main/java/org/gradle/api/internal/project/DynamicLookupRoutine.java
alllex: @reinsch82, if you agree with Justin, feel free to keep the deprecations for internal code. I was mainly curious

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
reinsch82: but is it though?
`getOrDefault` is a `Map` method.
So it works in both.

`properties["foo"] ?: "bar"` I guess you could consider this more idiomatic though.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/build_environment.adoc
```diff
@@ -185,6 +185,30 @@ $ export ORG_GRADLE_PROJECT_myProperty='Hi, world'
...
+
+The recommended way to access a project property is through the <<lazy_configuration.adoc#lazy_properties,Provider API>>:
+
+[source,kotlin]
+----
+val myProperty: Provider<String> = providers.gradleProperty("myProperty")
+----
+
+The returned `Provider` is lazy and compatible with the <<configuration_cache_requirements.adoc#config_cache:requirements:reading_sys_props_and_env_vars,Configuration Cache>>.
+The provider resolves the property from the build-level sources listed above: command-line `-P` arguments, `org.gradle.project.*` system properties, `ORG_GRADLE_PROJECT_*` environment variables, and `gradle.properties` files in the Gradle User Home, build root, and Gradle installation directories.
```
alllex: 💅 Does this render correctly? The IDE shows me that the `*` is interpreted as a starter for bold text, instead of being part of the backticked string. Maybe the asterisk needs to be escaped?
I didn't look at the rendered version, though

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/build_environment.adoc
```diff
@@ -185,6 +185,30 @@ $ export ORG_GRADLE_PROJECT_myProperty='Hi, world'
...
+
+[NOTE]
+====
+`providers.gradleProperty()` does not include properties from `gradle.properties` files in subproject directories, nor extra properties or other properties set dynamically on individual `Project` instances.
+To access those, use `project.findProperty("name")` or extra properties directly.
+====
+
+For other ways to access properties, see the following methods on the link:{groovyDslPath}/org.gradle.api.Project.html[Project] object:
+
+- `findProperty("name")` — returns the value or `null`, includes the full project property scope (extra properties, parent project properties)
```
alllex:
```suggestion
- `findProperty("name")` — returns the value or `null`, checking dynamically configured containers, like extra properties, and parent projects as well.
```

I'd like to avoid introducing a "full project property scope". The javadoc has more precise description.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -186,6 +312,8 @@ This is a result of the link:https://github.com/gradle/gradle/pull/36510[Windows
 If you are using a custom start script template that references the exit environment variable, update your template to remove this dependency.
 There is no replacement API, the exit environment variable concept has been removed entirely.
 
+|===
```
alllex: ❓ Is this a left-over? At least for the in-IDE preview this makes an enormous table cell 😄

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -93,6 +93,132 @@ Problem severity can no longer be set explicitly when creating a new instance.
 Instead, it is determined by the reporting method: `ProblemReporter.report()` produces warnings and `ProblemReporter.throwing()` produces errors.
 Calling `.severity()` on `ProblemSpec` is now a no-op and will be removed in Gradle 10.0.
 
+[[deprecated_project_get_properties]]
+[[deprecated_script_get_properties]]
```
alllex: 💅 We probably don't need to different links. I think we can have just `[[deprecated_get_properties]]`

PR #37298, opened 2026-03-27
Add sanity-check skill

blindpirate:
With this skill we can fix SanityCheck failures with a single command:

image

image

Test with tens of real-world SanityCheck failures in GBT builds.

### DIFF
--- a/.claude/skills/sanity-check/SKILL.md
+++ b/.claude/skills/sanity-check/SKILL.md
@@ -1,0 +1,75 @@
+---
+name: sanity-check
+description: Run `./gradlew sanityCheck` from the repository root, inspect failures, apply the relevant fix instructions, and retry until the build passes or 5 attempts have been made. Use when the user asks to run the sanity check workflow or iteratively resolve failures reported by `sanityCheck`.
+---
+
+# Sanity Check
+
+Run this workflow from the repository root.
+
+## Goal
+
+Get `./gradlew sanityCheck` to pass.
+
+## Workflow
+
+1. Set `maxAttempts=5`.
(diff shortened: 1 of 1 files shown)


.claude/skills/sanity-check/SKILL.md
```diff
@@ -0,0 +1,75 @@
+---
+name: sanity-check
+description: Run `./gradlew sanityCheck` from the repository root, inspect failures, apply the relevant fix instructions, and retry until the build passes or 5 attempts have been made. Use when the user asks to run the sanity check workflow or iteratively resolve failures reported by `sanityCheck`.
```
6hundreds, line 3: As the main advantage of skills is progressive context disclosure, length of the description is important thing to care about. It's like an interface to the skill. However, the instruction `and retry until the build passes or 5 attempts have been made` looking like implementation detail to me and shouldn't be a part of the description

.claude/skills/sanity-check/SKILL.md
```diff
@@ -0,0 +1,75 @@
...
+Run this workflow from the repository root.
+
+## Goal
+
+Get `./gradlew sanityCheck` to pass.
+
+## Workflow
+
+1. Set `maxAttempts=5`.
+2. Set `attempt=1`.
```
6hundreds, line 17: Interesting, never seen usage like this, `set` to where? How does it work?

blindpirate: It's pseudocode, like explaining an algorithm to a human.

blindpirate:
Superseded by https://github.com/gradle/agents/pull/14

Issue #37256, opened 2026-03-23
Lazy configuration eager resolution Gradle 9.4+


wakingrufus:
### Current Behavior

I believe this was introduced in this commit:
https://github.com/gradle/gradle/commit/0adb5a12525f06d83d6f43e7414b238b698b7501#diff-bf3f17dc83eebd0509d66ef191ecf410e9fe6a3e20b1d95de8de69959d66e658R192
since this change, if I do something like this:
```
project.getConfigurations().named(to.getRuntimeClasspathConfigurationName(), c -> {
            c.getAttributes().addAllLater(
                    project.getConfigurations().getByName(from.getRuntimeClasspathConfigurationName()).getAttributes()
            );
        });
```
to copy attributes from one configuration to another, it breaks with an error like this:
```
Could not determine the dependencies of task ':compileIntegrationTestJava'.
> Could not resolve all dependencies for configuration ':integrationTestCompileClasspath'.
   > This container is immutable and cannot be mutated. Creating a Named value is not supported.
```

### Expected Behavior

The configuration should not be immutable yet

### Self-contained Reproducer Project

it happens in some of the tests for https://github.com/nebula-plugins/nebula-project-plugin/pull/103 when we upgrade the gradle wrapper to 9.4.0 (it was previously working fine)

### Gradle version

9.4.0

ov7a:
This issue needs a decision from the team responsible for that area. They have been informed. Response time may vary.

hfhbd:
The error message was introduced in https://github.com/gradle/gradle/pull/35068

jvandort:
Minimal reproducer:
```
    // FreezableAttributeContainerTest.groovy

    interface Foo extends Named {}

    def "foo"() {
        def attr = Attribute.of("test", Foo)

        def freezable = createContainer()
        freezable.attributeProvider(attr, new DefaultProvider<Foo>(() -> {
            freezable.named(Foo, "value")
        }))

        def foo = factory.mutable()
        foo.addAllLater(freezable)
        freezable.freeze()

        expect:
        foo.getAttribute(attr)
    }
```

Note, a configuration uses a FreezableAttributeContainer behind the scenes, so `Configuration#getAttributes` returns a `FreezableAttributeContainer`. In this test, `createContainer` also returns a `FreezableAttributeContainer`. 

[...]

PR #37299, opened 2026-03-27
addAllLater inherits frozen entries from FreezableAttributeContainer

jvandort:
Previously, addAllLater would retain a mutable reference to the entries of a FreezableAttributeContainer. This caused some providers which relied on the named method of a FreezableAttributeContainer to throw exceptions after the container was frozen but consumers of the frozen container's entries still held a mutable reference to its unfrozen entries.

Fixes: https://github.com/gradle/gradle/issues/37256

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test rfr

PR #37319, opened 2026-03-27
Update task annotation validation errors

ljacomet:
These changes make it more explicit whenever the error is linked to having `@Optional` present, either with `@Internal` or alone.

This mainly introduces the following new error messages:
* First:
```
Execution failed for task ':validatePlugins' (registered by plugin 'org.gradle.java-gradle-plugin').
> Plugin validation failed with 1 problem:
    - Error: Type 'MyTask' property 'badProperty' is missing an input or output annotation.
      
      Reason: @Optional is a modifier annotation and has no effect without an input or output annotation.
      
      Possible solutions:
        1. Add an input or output annotation.
        2. Replace @Optional with @Internal for ignoring this property.
      
      For more information, please refer to https://docs.gradle.org/current/userguide/validation_problems.html#missing_annotation in the Gradle documentation.
  For more on how to annotate task properties, please refer to https://docs.gradle.org/current/userguide/incremental_build.html#sec:task_input_output_annotations in the Gradle documentation.
```
* Second:
```
Execution failed for task ':validatePlugins' (registered by plugin 'org.gradle.java-gradle-plugin').
> Plugin validation failed with 1 problem:
    - Error: Type 'MyTask' property 'badProperty' annotated with @Internal should not be also annotated with @Optional.
      
      Reason: @Internal properties are excluded from up-to-date checks; @Optional is redundant and not allowed here.
      
      Possible solutions:
        1. Remove the @Optional annotation.
        2. Remove the @Internal annotation and add an input or output annotation.
      
      For more information, please refer to https://docs.gradle.org/current/userguide/validation_problems.html#ignored_property_must_not_be_annotated in the Gradle documentation.
  For more on how to annotate task properties, please refer to https://docs.gradle.org/current/userguide/incremental_build.html#sec:task_input_output_annotations in the Gradle documentation.
```

It also has minor modifications on the previous message:

* Now:
```
Execution failed for task ':validatePlugins' (registered by plugin 'org.gradle.java-gradle-plugin').
> Plugin validation failed with 1 problem:
    - Error: Type 'MyTask' property 'tree.nonAnnotated' is missing an input or output annotation.
      
      Reason: Properties must be annotated so that Gradle knows how to handle them during up-to-date checking.
      
      Possible solutions:
        1. Add an input or output annotation.
        2. Mark it as @Internal.
      
      For more information, please refer to https://docs.gradle.org/current/userguide/validation_problems.html#missing_annotation in the Gradle documentation.
  For more on how to annotate task properties, please refer to https://docs.gradle.org/current/userguide/incremental_build.html#sec:task_input_output_annotations in the Gradle documentation.
```
* Before:
```
Execution failed for task ':validatePlugins' (registered by plugin 'org.gradle.java-gradle-plugin').
> Plugin validation failed with 1 problem:
    - Error: Type 'MyTask' property 'tree.nonAnnotated' is missing an input or output annotation.
      
      Reason: A property without annotation isn't considered during up-to-date checking.
      
      Possible solutions:
        1. Add an input or output annotation.
        2. Mark it as @Internal.
      
      For more information, please refer to https://docs.gradle.org/current/userguide/validation_problems.html#missing_annotation in the Gradle documentation.
  For more on how to annotate task properties, please refer to https://docs.gradle.org/current/userguide/incremental_build.html#sec:task_input_output_annotations in the Gradle documentation.
```

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/properties/annotations/MissingPropertyAnnotationHandler.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/properties/annotations/MissingPropertyAnnotationHandler.java
@@ -17,6 +17,7 @@
 package org.gradle.internal.properties.annotations;
 
 import org.gradle.api.problems.internal.GradleCoreProblemGroup;
+import org.gradle.api.tasks.Optional;
 import org.gradle.internal.reflect.annotations.PropertyAnnotationMetadata;
 import org.gradle.internal.reflect.validation.TypeValidationContext;
 import org.gradle.util.internal.TextUtil;
@@ -40,9 +41,17 @@ public interface MissingPropertyAnnotationHandler {
             .forProperty(annotationMetadata.getPropertyName())
             .id(TextUtil.screamingSnakeToKebabCase(missingAnnotation), "Missing annotation", GradleCoreProblemGroup.validation().property())
             .contextualLabel("is missing " + displayName)
-            .documentedAt(userManual("validation_problems", missingAnnotation.toLowerCase(Locale.ROOT)))
-            .details("A property without annotation isn't considered during up-to-date checking")
-            .solution("Add " + displayName)
-            .solution("Mark it as @Internal");
+            .documentedAt(userManual("validation_problems", missingAnnotation.toLowerCase(Locale.ROOT)));
+        if (annotationMetadata.isAnnotationPresent(Optional.class)) {
+            problem
+                .details("@Optional is a modifier annotation and has no effect without an input or output annotation")
+                .solution("Add an input or output annotation")
+                .solution("Replace @Optional with @Internal for ignoring this property");
+        } else {
+            problem
+                .details("Properties must be annotated so that Gradle knows how to handle them during up-to-date checking")
+                .solution("Add " + displayName)
+                .solution("Mark it as @Internal");
+        }
     });
 }
--- a/platforms/extensibility/plugin-development/src/integTest/groovy/org/gradle/plugin/devel/tasks/AbstractPluginValidationIntegrationSpec.groovy
+++ b/platforms/extensibility/plugin-development/src/integTest/groovy/org/gradle/plugin/devel/tasks/AbstractPluginValidationIntegrationSpec.groovy
@@ -136,7 +136,7 @@ abstract class AbstractPluginValidationIntegrationSpec extends AbstractIntegrati
             verifyAll(receivedProblem(0)) {
                 fqid == 'validation:property-validation:missing-annotation'
                 contextualLabel == 'Type \'MyTask\' property \'badTime\' is missing an input or output annotation'
-                details == 'A property without annotation isn\'t considered during up-to-date checking'
+                details == 'Properties must be annotated so that Gradle knows how to handle them during up-to-date checking'
                 solutions == [
                     'Add an input or output annotation',
                     'Mark it as @Internal',
@@ -149,7 +149,7 @@ abstract class AbstractPluginValidationIntegrationSpec extends AbstractIntegrati
             verifyAll(receivedProblem(1)) {
                 fqid == 'validation:property-validation:missing-annotation'
                 contextualLabel == 'Type \'MyTask\' property \'oldThing\' is missing an input or output annotation'
-                details == 'A property without annotation isn\'t considered during up-to-date checking'
+                details == 'Properties must be annotated so that Gradle knows how to handle them during up-to-date checking'
                 solutions == [
                     'Add an input or output annotation',
                     'Mark it as @Internal',
@@ -162,7 +162,7 @@ abstract class AbstractPluginValidationIntegrationSpec extends AbstractIntegrati
             verifyAll(receivedProblem(2)) {
                 fqid == 'validation:property-validation:missing-annotation'
                 contextualLabel == 'Type \'MyTask\' property \'options.badNested\' is missing an input or output annotation'
-                details == 'A property without annotation isn\'t considered during up-to-date checking'
+                details == 'Properties must be annotated so that Gradle knows how to handle them during up-to-date checking'
                 solutions == [
                     'Add an input or output annotation',
                     'Mark it as @Internal',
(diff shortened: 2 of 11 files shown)


pbielicki (approved):
lgtm apart from little things we discussed on Slack

lkasso (approved):

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/properties/annotations/MissingPropertyAnnotationHandler.java
```diff
@@ -40,9 +41,17 @@ public interface MissingPropertyAnnotationHandler {
...
             .contextualLabel("is missing " + displayName)
-            .documentedAt(userManual("validation_problems", missingAnnotation.toLowerCase(Locale.ROOT)))
-            .details("A property without annotation isn't considered during up-to-date checking")
-            .solution("Add " + displayName)
-            .solution("Mark it as @Internal");
+            .documentedAt(userManual("validation_problems", missingAnnotation.toLowerCase(Locale.ROOT)));
+        if (annotationMetadata.isAnnotationPresent(Optional.class)) {
+            problem
+                .details("@Optional is a modifier annotation and has no effect without an input or output annotation")
+                .solution("Add an input or output annotation")
```
alllex, line 48: 💅 The second solution is very actionable, while this first solution is not so much. If we had just one or two input/output annotations, I'd directly mention them here. Alas, we have like 10 of them.

It seems they are all explicitly listed on the `@Optional` javadoc. Can we point users to to that javadoc on this line to make this more actionable?

I know we are linking to the userguide page explaining it at the bottom, but this requires a bit of switching back and forth.

platforms/extensibility/plugin-development/src/integTest/groovy/org/gradle/plugin/devel/tasks/AbstractPluginValidationIntegrationSpec.groovy
```diff
@@ -149,7 +149,7 @@ abstract class AbstractPluginValidationIntegrationSpec extends AbstractIntegrati
             verifyAll(receivedProblem(1)) {
                 fqid == 'validation:property-validation:missing-annotation'
                 contextualLabel == 'Type \'MyTask\' property \'oldThing\' is missing an input or output annotation'
```
alllex, line 151: 💅 I know it's a bit out of scope for this change, but the very first line doesn't read well for me. E.g. it's not very clear whether the "type" represents the owner of the property or the type of the property.

I think it would have been much better if the property name went first and it was clear that the type is the owner

```
Property 'oldThing' on type 'MyTask' is missing ...
```

I know there is also logic to clarify which plugin the type belongs to. It's put at the front as well, but I think it's also secondary information.

platforms/extensibility/plugin-development/src/integTest/groovy/org/gradle/plugin/devel/tasks/AbstractPluginValidationIntegrationSpec.groovy
```diff
@@ -901,6 +901,87 @@ abstract class AbstractPluginValidationIntegrationSpec extends AbstractIntegrati
         }
     }
 
+    def "reports @Optional used without an input or output annotation"() {
```
alllex, line 904: ❓ Is this technically a breaking change or was the error reported in some other way before?

platforms/extensibility/plugin-development/src/integTest/groovy/org/gradle/plugin/devel/tasks/AbstractPluginValidationIntegrationSpec.groovy
```diff
@@ -901,6 +901,87 @@ abstract class AbstractPluginValidationIntegrationSpec extends AbstractIntegrati
...
+        """
+
+        expect:
+        assertValidationFailsWith([
+            error(missingAnnotationWithOptionalConfig { type('MyTask').property('badProperty').missingInputOrOutput() }, 'validation_problems', 'missing_annotation'),
+        ])
+
+        and:
+        if (isProblemsApiCheckEnabled()) {
+            verifyAll(receivedProblem(0)) {
```
alllex, line 928: 💭 It would be nice to have a fixture method that not only returns the first Problems API problem, but also verify that there is only one.

I imagine something like `verifyAll(expectSingleProblem())`

platforms/extensibility/plugin-development/src/integTest/groovy/org/gradle/plugin/devel/tasks/AbstractPluginValidationIntegrationSpec.groovy
ljacomet: The error existed but was not clear and actionable, so not a breaking change.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/properties/annotations/MissingPropertyAnnotationHandler.java
ljacomet: My take is that at this stage the developer already has a basic understanding of this area and has read at least something about task up-to-date checks and the the like. So in this context, I don't think the addition at this particular place is needed. And if they know nothing, then a javadoc link is not what they need, they need the links below.

platforms/extensibility/plugin-development/src/integTest/groovy/org/gradle/plugin/devel/tasks/AbstractPluginValidationIntegrationSpec.groovy
ljacomet: I agree that makes sense, let me look into it. Depending on the size, I may do it in this PR or in a separate one.

ljacomet: This additional change is quite massive, will do a follow up PR for it and not attempt to include it here.

alllex (approved):

PR #37331, opened 2026-03-29
Simplify virtual platform variants

jvandort:
Previously, lenient platforms had two variant types: an "implicit" variant selected when a normal dependency targeted the platform, and a "source-aware" variant selected when a virtual platform edge targeted it. The implicit variant unconditionally created constraints for all participating modules at the platform's own version without checking whether those versions exist. This caused order-dependent behavior: if an explicit dependency on the platform resolved before any virtual edges, all participants would be pinned to a static version that could conflict with the dynamic version computed by the source-aware variant.

This change removes the implicit variant and uses the source-aware variant for all cases, which checks candidate versions for existence before creating constraints.

Additional fixes:
- getReason() now correctly reports "constrained by platform" for constraint edges and "belongs to platform" for ownership edges
- platformId now always refers to the platform's own component ID, not the target module's ID
- The variant is cached on the component instead of recreated per edge
- Removed unused platformNode parameter threaded through multiple layers
- Derived isTransitive() from !constraint instead of a separate field

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

PR #37333, opened 2026-03-29
Add POM files as a consumable variant

big-guy:
## Summary

Exposes Maven POM files as a variant-aware artifact so consumers can resolve a published module's POM (and its full parent chain) via `ArtifactView` with `withVariantReselection()`.

### New public API (incubating)

- `Category.METADATA` — new category for tooling-only metadata variants
- `MetadataFormat` attribute with `MAVEN` constant — disambiguates which metadata format a variant carries

### Resolution

- `JavaEcosystemVariantDerivationStrategy` synthesizes a `pom` variant for Maven modules
  - Works for **all Maven dependencies**, including those that publish Gradle Module Metadata: when GMM is present, the POM variant is added supplementally alongside the explicit GMM variants
  - Includes the full `<parent>` chain as additional artifacts (Option B2)
- `AbstractLazyModuleComponentResolveMetadata` / `AbstractRealisedModuleComponentResolveMetadata` always invoke `deriveSupplementalVariants` to add the metadata variant — no gate on existing variant names
- Parent POM chain is plumbed through `GradlePomModuleDescriptorParser`, propagated through metadata copies, and serialized in the cache (cache layout version bumped)
- Per-artifact component IDs are preserved through serialization so parent POM artifacts resolve from the correct module

### Publishing

- `MavenPublishPlugin` registers a consumable `pomElements` configuration so **project** dependencies can resolve the producing project's POM via the same query
- `pomElements` is **not** included in the published Gradle Module Metadata — it is only a local resolution affordance
- Regression test asserts the GMM produced by `generateMetadataFileForMavenPublication` does not contain `pomElements` or `"category": "metadata"`

### Documentation

- User-manual section on resolving POMs with `withVariantReselection()`
- Worked example: extracting license information from a dependency's POM

## Known follow-up

- The synthesized `pom` variant carries `Category=metadata`. Under the default attribute compatibility rule, a consumer that does not specify `Category` matches *any* producer value, which makes the metadata variant compete with normal-resolution variants in tests that use bare configurations (e.g. `MavenRemoteDependencyWithGradleMetadataResolutionIntegrationTest`). Needs a `Category` compatibility rule that makes producer-`Category=metadata` incompatible with consumer-silent, mirroring how `Category=verification` stays out of normal classpath resolution.

## Test plan

- [x] Resolve POM for direct + transitive Maven deps
- [x] Resolve parent and grandparent POMs via the parent chain
- [x] Resolve POM variant for Maven modules that also publish GMM
- [x] Resolve POM variant for project dependencies via `pomElements`
- [x] Shared parent POM at different versions resolves correctly
- [x] `pomElements` is **not** present in published GMM (regression test)
- [x] Existing `DerivedVariantsResolutionIntegrationTest` passes
- [ ] `Category=metadata` compatibility rule (follow-up — see Known follow-up)
- [ ] `sanityCheck` (binary compat, architecture tests)

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/reference/dependency-management/controlling-dependency-resolution/artifact_views.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/reference/dependency-management/controlling-dependency-resolution/artifact_views.adoc
@@ -82,6 +82,22 @@ include::{snippetsPath}/reference/dependency-management/controlling-dependency-r
 When this API is used, the attributes used for variant reselection are specified solely by the `ArtifactView.getAttributes()` method.
 The graph resolution attributes specified on the configuration are completely ignored during variant reselection.
 
+=== Resolving POM files
+
+Similarly, you can use variant reselection to resolve the POM metadata files for all dependencies.
+For external Maven modules without Gradle Module Metadata, Gradle derives a `metadata` variant containing the module's POM file and its parent POM chain.
+
+NOTE: This example uses incubating APIs.
+
+====
+include::sample[dir="snippets/reference/dependency-management/controlling-dependency-resolution/performingResolution-artifactResolution/kotlin",files="build.gradle.kts[tags=pom-variant-reselection]"]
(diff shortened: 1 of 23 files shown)


big-guy:
@bot-gradle test this

PR #37338, opened 2026-03-30
Remove agent requirements

blindpirate:
Closes https://github.com/gradle/gradle-private/issues/5147

In the past, there are some build types with agent requirements "not running on EC2". This PR removes these requirements and use agent pool compatibility instead.

This PR also groups the smoke tests to be a smoke test project in case we need to re-enable the limitation because TeamCity agent compatibility is project based.

### DIFF
--- a/.teamcity/src/main/kotlin/configurations/Gradleception.kt
+++ b/.teamcity/src/main/kotlin/configurations/Gradleception.kt
@@ -10,7 +10,6 @@ import common.customGradle
 import common.functionalTestParameters
 import common.getBuildScanCustomValueParam
 import common.gradleWrapper
-import common.requiresNotEc2Agent
 import common.requiresNotSharedHost
 import common.skipConditionally
 import jetbrains.buildServer.configs.kotlin.BuildSteps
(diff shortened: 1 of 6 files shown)


.teamcity/src/main/kotlin/configurations/Gradleception.kt
```diff
@@ -64,8 +63,6 @@ class Gradleception(
             "Builds Gradle with the version of Gradle which is currently under development (twice)$descriptionSuffix"
 
         requirements {
-            // Gradleception is a heavy build which runs ~40m on EC2 agents but only ~20m on Hetzner agents
```
blindpirate, line 67: The comment is no longer true.

.teamcity/src/main/kotlin/promotion/BasePromotionBuildType.kt
```diff
@@ -40,7 +39,6 @@ abstract class BasePromotionBuildType(
 
         requirements {
             requiresOs(Os.LINUX)
-            requiresNotEc2Agent()
```
blindpirate, line 43: We will use agent pool to control which agents the promotion builds are running on.

blindpirate:

blindpirate:
@bot-gradle cherrypick to release7x

PR #37348, opened 2026-03-30
Fix backward compatibility issues with `--generate-diffs` parameter

hegyibalint:
Fixes a problem caused by
 - https://github.com/gradle/gradle/pull/37093

Where older branches have no clear way of executing ad-hoc tests anymore.
This is solved by defaulting on no additional option, and rely on the build-in default in the task. 
This allows past and future branches to default on a policy without any problem.

An explicit flag is only added if skipping is selected.

### DIFF
--- a/.teamcity/src/main/kotlin/util/AdHocPerformanceScenario.kt
+++ b/.teamcity/src/main/kotlin/util/AdHocPerformanceScenario.kt
@@ -48,12 +48,16 @@ abstract class AdHocPerformanceScenario(
             param("checks", "all")
             text("runs", "40", display = ParameterDisplay.PROMPT, allowEmpty = false)
             text("warmups", "10", display = ParameterDisplay.PROMPT, allowEmpty = false)
-            text(
+            select(
                 "generateDiffs",
-                "true",
+                "",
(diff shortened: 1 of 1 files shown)


.teamcity/src/main/kotlin/util/AdHocPerformanceScenario.kt
```diff
@@ -48,12 +48,15 @@ abstract class AdHocPerformanceScenario(
...
                 "generateDiffs",
-                "true",
+                "",
                 display = ParameterDisplay.PROMPT,
-                allowEmpty = false,
-                description = "Whether to generate differential flame graphs after profiling",
+                description = "Skips differential flame graph generation to avoid OOM in large adhoc tests.",
+                options = listOf(
+                    "Generate diffs" to "",
+                    "Skip diffs generation" to "--no-generate-diffs",
```
hegyibalint: TIL that Gradle for all boolean fields generates a `--no-<<option_name>>` positional.

cobexer:
Could you please rebase this on top of `release` @hegyibalint

hegyibalint:
@cobexer  Made a mess... never say retargeting branch in here...
Done!

Issue #37124, opened 2026-03-15
DCL: Remove ProjectFeatureApplicationContext.registerBuildModel


hfhbd:
### Expected Behavior

No api to manually register a build model for a nested definition, eg part of NDOC

### Current Behavior (optional)

You can call `ProjectFeatureApplicationContext.registerBuildModel` inside a `NDOC<Definition>`.

### Context

With https://github.com/gradle/gradle/pull/36876, all nested build models are automatically registered:

> We auto-discover nested definitions and expose them as binding points. This means that it is no longer necessary for user code to call registerBuildModel(...). Instead, the build model is discovered and registered when the definition is instantiated. Authors can call getBuildModel() to retrieve the build model for a nested definition object.

If it is no longer necessary, there should be no api to do it.

ov7a:
This issue needs a decision from the team responsible for that area. They have been informed. Response time may vary.

h0tk3y:
We should not remove it just yet. The auto-discovery that we have only works reliably for safe definitions. If a definition has code that runs lambdas against objects not exposed via properties, nested definition discovery won't find those. So, for unsafe definitions, such as in migrating plugins, we still need to provide the plugin author with a way to register the build models for the nested definition instances if they are not discoverable automatically.

eskatos:
The issue is in the backlog of the relevant team and is prioritized by them.
***

We should remove this from the safe `ProjectFeatureApplicationContext` and move it to the unsafe one.

Related
* https://github.com/gradle/gradle/issues/37123

PR #37350, opened 2026-03-30
Move build model registration to internal service only available in unsafe apply context

donat:
Fixes https://github.com/gradle/gradle/issues/37124

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/internal/PluginUseServices.java
+++ b/platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/internal/PluginUseServices.java
@@ -29,11 +29,9 @@ import org.gradle.api.internal.project.ProjectInternal;
 import org.gradle.api.internal.tasks.properties.InspectionScheme;
 import org.gradle.api.internal.tasks.properties.InspectionSchemeFactory;
 import org.gradle.api.problems.internal.InternalProblems;
-import org.gradle.features.internal.binding.DefaultBuildModelRegistrar;
 import org.gradle.features.internal.binding.DefaultModelDefaultsApplicator;
 import org.gradle.features.internal.binding.DefaultProjectFeatureApplicator;
 import org.gradle.features.internal.binding.DefaultProjectFeatureDeclarations;
-import org.gradle.features.internal.binding.BuildModelRegistrarInternal;
 import org.gradle.features.internal.binding.ModelDefaultsApplicator;
 import org.gradle.features.internal.binding.ModelDefaultsHandler;
 import org.gradle.features.internal.binding.ProjectFeatureApplicator;
@@ -208,11 +206,6 @@ public class PluginUseServices extends AbstractGradleModuleServices {
         @Provides
         ModelDefaultsApplicator createModelDefaultsApplicator(List<ModelDefaultsHandler> modelDefaultsHandlers) {
             return new DefaultModelDefaultsApplicator(modelDefaultsHandlers);
-        }
-
-        @Provides
-        BuildModelRegistrarInternal createBuildModelRegistrar(InstantiatorFactory instantiatorFactory, ServiceRegistry services, ProjectInternal project) {
-            return instantiatorFactory.inject(services).newInstance(DefaultBuildModelRegistrar.class, project.getObjects());
         }
     }
 }
--- a/platforms/core-configuration/project-features-api/src/main/java/org/gradle/features/binding/BuildModelRegistrar.java
+++ b/platforms/core-configuration/project-features-api/src/main/java/org/gradle/features/binding/BuildModelRegistrar.java
@@ -17,8 +17,6 @@
 package org.gradle.features.binding;
 
 import org.gradle.api.Incubating;
-import org.gradle.internal.service.scopes.Scope;
-import org.gradle.internal.service.scopes.ServiceScope;
 
 /**
  * Provides explicit build model registration.
@@ -31,7 +29,6 @@ import org.gradle.internal.service.scopes.ServiceScope;
  * @since 9.6.0
  */
 @Incubating
-@ServiceScope(Scope.Project.class)
 public interface BuildModelRegistrar {
     /**
      * Creates, registers, and returns a new build model instance for the given {@code definition} instance.
--- a/platforms/extensibility/plugin-use/src/main/java/org/gradle/features/internal/binding/BuildModelRegistrarInternal.java
+++ b/platforms/extensibility/plugin-use/src/main/java/org/gradle/features/internal/binding/BuildModelRegistrarInternal.java
@@ -20,12 +20,9 @@ import org.gradle.features.binding.BuildModel;
 import org.gradle.features.binding.BuildModelRegistrar;
 import org.gradle.features.binding.Definition;
 import org.gradle.features.binding.ProjectFeatureApplicationContext;
-import org.gradle.internal.service.scopes.Scope;
-import org.gradle.internal.service.scopes.ServiceScope;
 
 import java.util.Map;
 
-@ServiceScope(Scope.Project.class)
 public interface BuildModelRegistrarInternal extends BuildModelRegistrar {
 
     /**
--- a/platforms/extensibility/plugin-use/src/main/java/org/gradle/features/internal/binding/DefaultBuildModelRegistrar.java
+++ b/platforms/extensibility/plugin-use/src/main/java/org/gradle/features/internal/binding/DefaultBuildModelRegistrar.java
@@ -23,22 +23,18 @@ import org.gradle.internal.Cast;
 import org.gradle.internal.inspection.DefaultTypeParameterInspection;
 import org.gradle.internal.inspection.TypeParameterInspection;
 
-import javax.inject.Inject;
 import java.util.Map;
 
-public abstract class DefaultBuildModelRegistrar implements BuildModelRegistrarInternal {
-
-    @Inject
-    protected abstract ProjectFeatureApplicator getProjectFeatureApplicator();
-
-    @Inject
-    protected abstract ProjectFeatureDeclarations getProjectFeatureRegistry();
+public class DefaultBuildModelRegistrar implements BuildModelRegistrarInternal {
 
     private final ObjectFactory objectFactory;
+    private final ProjectFeatureApplicator projectFeatureApplicator;
+    private final ProjectFeatureDeclarations projectFeatureDeclarations;
 
-    @Inject
-    public DefaultBuildModelRegistrar(ObjectFactory objectFactory) {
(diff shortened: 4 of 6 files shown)


donat:
@bot-gradle test QFL

donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test this

platforms/core-configuration/project-features-api/src/main/java/org/gradle/features/registration/BuildModelRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+ * While a feature's primary definition object has its build model registered automatically,
+ * nested definition objects (such as container elements) require explicit registration via this service.
+ * <p>
+ * This service is available for injection in unsafe apply actions.
+ *
+ * @since 9.6.0
+ */
+@Incubating
+@ServiceScope(Scope.Project.class)
+public interface BuildModelRegistry {
```
donat: What is the proper place for this interface?

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/internal/PluginUseServices.java
```diff
@@ -207,5 +209,10 @@ ProjectFeatureApplicator createProjectFeatureApplicator(
         ModelDefaultsApplicator createModelDefaultsApplicator(List<ModelDefaultsHandler> modelDefaultsHandlers) {
             return new DefaultModelDefaultsApplicator(modelDefaultsHandlers);
         }
+
+        @Provides
+        BuildModelRegistry createBuildModelRegistry(InstantiatorFactory instantiatorFactory, ServiceRegistry services) {
+            return instantiatorFactory.inject(services).newInstance(DefaultBuildModelRegistry.class);
+        }
```
donat: What is the correct ObjectFactory here?

```suggestion
        BuildModelRegistry createBuildModelRegistry(InstantiatorFactory instantiatorFactory, ServiceRegistry services, ProjectInternal project) {
            return instantiatorFactory.inject(services).newInstance(DefaultBuildModelRegistry.class, project.getObjects());
        }
```

platforms/core-configuration/project-features-api/src/main/java/org/gradle/features/registration/BuildModelRegistry.java
ghale: Are you asking which project it should be under?  Or are you asking what package it should be in?  If the former, project-features-api is the right place.  If the latter, I think I'd put it in org.gradle.features.binding because it's not like the other classes in registration (which are services for registering build logic).  This service is more about binding the feature than actual build logic.

ghale: 💅 I might call this class `BuildModelRegistrar` instead of "Registry" since it only covers registration and can't be queried.

platforms/core-configuration/project-features-api/src/main/java/org/gradle/features/registration/BuildModelRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+     * This method must only be used on nested definition objects, such as container elements, and not on a feature's primary definition object, which has its
+     * build model registered automatically.
+     * <p>
+     * A build model must be registered for a definition before {@link ProjectFeatureApplicationContext#getBuildModel(Definition)} is used on it.
+     *
+     * @throws IllegalStateException if there is already a build model instance registered for the definition.
+     *
+     * @since 9.6.0
+     */
+    <T extends Definition<V>, V extends BuildModel> V registerBuildModel(T definition, Map<Class<?>, Class<?>> nestedBuildModelTypesToImplementationTypes);
```
ghale: ❌ This should probably be on an internal interface.  I don't see plugin authors needing to use this.  This would only be used by the feature applicator.

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/internal/PluginUseServices.java
ghale: I would use the project object factory.  This is what is used for everything else, and what was inherently used by `ProjectFeatureApplyContext` before.

platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/features/internal/builders/types/ProjectTypePluginClassBuilder.groovy
```diff
@@ -127,6 +128,9 @@ class ProjectTypePluginClassBuilder {
             interface Services {
                 @javax.inject.Inject
                 ${TaskRegistrar.class.name} getTaskRegistrar();
+
+                @javax.inject.Inject
+                ${BuildModelRegistrar.class.name} getBuildModelRegistry();
```
ghale:
```suggestion
                ${BuildModelRegistrar.class.name} getBuildModelRegistrar();
```

platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/features/internal/ProjectFeatureFixture.groovy
```diff
@@ -399,7 +399,7 @@ trait ProjectFeatureFixture extends ProjectTypeFixture {
 
     PluginBuilder withProjectFeaturePluginThatBindsToMultipleTargets() {
         def projectTypeDefinition = new ProjectTypeDefinitionWithMultipleNestedBindingLocations()
-        def projectType = new ProjectTypePluginClassBuilder(projectTypeDefinition)
+        def projectType = new ProjectTypePluginClassBuilder(projectTypeDefinition).withUnsafeApplyAction()
```
ghale: ❓ Looking at this fixture, I think the calls to `registerBuildModel()` in `ProjectTypeDefinitionWithMultipleNestedBindingLocations` are no longer necessary since we implemented automatically discovering and registering build models.  In other words, we can remove the overridden `getBuildModelMapping()` there and then we don't have to mark this unsafe.

However, I think once we do that, we no longer have anything calling `registerBuildModel()` which means we're not testing whatever case it's still needed for.  We should add some sort of coverage for this.

donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test this

platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/features/internal/ProjectFeatureFixture.groovy
ghale: The scenario where we need this is that we have a function like:
```
fun foo(action: Foo.() -> Unit) { 
    val myFoo = objectFactory.newInstance(Foo)
    action(myFoo)
}
```
Which would be configured like:
```
foo {
    bar = 1
}
```
And the author wants to allow features to bind to foo (say a feature baz):
```
foo {
    bar = 1
    baz {
        buzz = 2 
    }
}
```
Since `myFoo` is not exposed with a getter, we can't discover it and register its build model at binding time.  Instead, the author will need to do this inside the function with `registerBuildModel()`.

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/internal/PluginUseServices.java
ghale: I was just thinking about this some more and I don't think we want to make this a generally available service.  Instead, this should be a service that is only available in `UnsafeServicesForApplyAction`.  The reason is that this service will _create_ the build model object if it doesn't exist.  If we don't use the object factory associated with the apply action, it won't have access to the additional context-specific services (like `ProjectFeatureLayout`).  And we don't impose any restrictions for unsafe services right now, but if we did, we would want those to apply to build model instances, too.  

So we really should use the object factory associated with the feature and its `ServiceLookup`.

donat:
@bot-gradle test this

platforms/core-configuration/project-features-api/src/main/java/org/gradle/features/registration/BuildModelRegistry.java
donat: Done

platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/features/internal/ProjectFeatureFixture.groovy
donat: I implemented the test coverage, PTAL!

donat:
@bot-gradle test this

platforms/core-configuration/project-features/src/integTest/groovy/org/gradle/features/ProjectFeatureDeclarationIntegrationTest.groovy
```diff
@@ -710,6 +710,31 @@ class ProjectFeatureDeclarationIntegrationTest extends AbstractIntegrationSpec i
         outputContains("Binding GroupFeatureDefinition")
     }
 
+    def "can register build model for a non-discoverable nested definition"() {
+        given:
+        PluginBuilder pluginBuilder = withProjectTypeUsingNonDiscoverableDefinition()
+        pluginBuilder.addBuildScriptContent pluginBuildScriptForJava
+        pluginBuilder.prepareToExecute()
```
ghale, line 717: ❌ I think we should do one more thing for this test.  The point of registering the build model is so that it's exposed for other features to bind to.  We should also have a feature that binds to the registered build model.  You can probably just reuse [this feature fixture](https://github.com/gradle/gradle/blob/c24961eb07b44caa48f52fc0cc926538a858087b/platforms/core-configuration/declarative-dsl-provider/src/testFixtures/groovy/org/gradle/features/internal/ProjectFeatureFixture.groovy#L300-L303).

And then our dcl becomes something like:
```
testProjectType {
    id = "test"
    foo {
        bar = "baz"
        feature {
            text = "foo"
        }
    }
}
```

donat:
@bot-gradle test this

platforms/core-configuration/project-features/src/integTest/groovy/org/gradle/features/ProjectFeatureDeclarationIntegrationTest.groovy
donat: I've implemented this, please check.

donat:
@bot-gradle test this

Issue #10979, opened 2019-10-08
ExtensionAware delegate provider in Kotlin DSL is very counterintuitive when used for extension properties


netvl:
### Expected Behavior

```kotlin
// in settings.gradle.kts

rootProject.name = "root"

include("project-1", "project-2")

gradle.allprojects {
    var someProperty: String by extra
    someProperty = name
}
```

```kotlin
// in build.gradle.kts

var Project.someProperty: String by extra

allprojects {
    println(someProperty)
}
```

I expect this script to print

```
root
project-1
project-2
```

### Current Behavior

Instead, this script prints

```
root
root
root
```

### Context

In my case, I have two groups of projects in a build written in different JVM languages and for different purposes. Within each group these projects have some common configuration logic. In order to set up this common configuration logic, I mark these projects with a boolean flag inside `settings.gradle.kts`, and then use this flag in the root `build.gradle.kts` to perform common configuration, roughly like this:

```kotlin
val Project.isFoo: Boolean by extra
val Project.isBar: Boolean by extra

allprojects.filter { it.isFoo }.forEach {
    ...
}

allprojects.filter { it.isBar }.forEach {
    ...
}
```

The problem is, the `by extra` delegation here won't work, since `extra` means extra properties of the top-level `Project` (because that's where these extension properties are declared). Instead of `by extra`, I have to use something like

```kotlin
val Project.isFoo: Boolean
    get() = extra.get("isFoo") as Boolean

val Project.isBar: Boolean
    get() = extra.get("isBar") as Boolean
```

The reason why this happens is that `ExtensionAware` delegate provider, quite naturally, uses the respective `ExtensionAware` object to retrieve properties from, ignoring the receiver object completely.

Ideally `extra` or probably some other delegate provider should allow using the target extra properties map. At the very least, if used like above, it should throw an exception, because currently it is very confusing because it compiles but does not work correctly.

### Steps to Reproduce 

See examples above.

### Your Environment

Gradle 5.6.2

netvl:
Side note: I'm not sure what kind of issue this is (a bug or a feature request), but since it is an example of very confusing behavior, I decided it is more like a bug than a potential feature.

bamboo:
Thanks for the report, @netvl.

I'd argue this is a feature request. And if I understand it correctly I think the title would better reflect it rephrased into something like:

> `ExtensionAware.extra` property delegate should honor receiver when used to declare an extension property.

What do you think?

netvl:
[...]

PR #37359, opened 2026-03-30
Add Best Practice to avoid Kotlin property delegates

tresat:
Document the recommendation to use explicit API calls instead of Kotlin DSL property delegates. This prevents implicit name coupling, where renaming a variable unexpectedly changes a task or configuration name, potentially breaking builds or CI scripts. It also avoids known correctness issues with property delegates (#10979).

./gradlew :docs:checkDeadInternalLinks -q :docs:docsTest --tests "*.bestPractices_avoidKotlinPropertyDelegates*"

image

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_general.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_general.adoc
@@ -808,3 +808,72 @@ This structure uses an extension type to hold values, allowing properties to be 
 === Tags
 
 <<tags_reference.adoc#tag:properties,`#properties`>>
+
+[[avoid_kotlin_property_delegates]]
+== Use Explicit API Instead of Kotlin Property Delegates
+
+Use explicit API calls like `tasks.register("name")`, `configurations.create("name")`, and `extra.set("name", value)` instead of the Kotlin DSL property delegate forms `by tasks.registering`, `by configurations.creating`, and `by extra`.
+
+=== Explanation
+
+The Kotlin DSL provides delegate extensions that derive entity names from the Kotlin variable name.
+This affects tasks (`by tasks.registering`, `by tasks.creating`), configurations (`by configurations.creating`), extra properties (`by extra`), and Gradle properties (`by project`, `by settings`).
+While these can appear concise, Kotlin property delegates are somewhat uncommon, even among experienced Kotlin developers, and can not be easily reproduced in other Gradle DSLs.
+
+The main risk is *implicit name coupling*: the registered entity's name is silently tied to the variable name.
+Renaming a variable renames the task or configuration, which can break CLI invocations, CI scripts, and dependent builds without any compiler warning.
+With the explicit API, the entity name is a visible string literal, and the variable can be named independently -- for example, `val verify = tasks.register("verifyCompatibilityWithPublishedArtifacts")` keeps a short local alias while preserving the descriptive task name.
+
+Avoid using any Kotlin property delegate offered by the Gradle Kotlin DSL.
+The explicit API is equally concise and avoids the pitfalls of implicit naming and delegate-specific correctness bugs.
+
+=== Example
+
+==== Don't Do This
+
+[.multi-language-sample]
+=====
+[source, kotlin]
+.build.gradle.kts
+----
+include::{snippetsPath}/bestPractices/avoidKotlinPropertyDelegates-avoid/kotlin/build.gradle.kts[tags=avoid-this]
+----
+=====
+
+<1> Configuration name is implicitly derived from the variable name `myLibs`. Renaming this variable silently renames the configuration.
+<2> `by extra` attaches a property whose name is derived from the variable. When used with a `Project` receiver (e.g. `var Project.prop by extra`), the delegate silently ignores the receiver and always reads from the declaring project's extra properties, returning wrong values in multi-project builds (link:https://github.com/gradle/gradle/issues/10979[#10979]).
+<3> Task name is implicitly derived from the variable name `greet`. Renaming this variable silently renames the task.
+
+==== Do This Instead
+
+++++
+<div style="text-align: right;">
+  <a class="download-project-link"
+     data-base-path="https://github.com/gradle/gradle/tree/master/platforms/documentation/docs/src/snippets/bestPractices/avoidKotlinPropertyDelegates-do/"
+     href="https://download-directory.github.io/?url=https://github.com/gradle/gradle/tree/master/platforms/documentation/docs/src/snippets/bestPractices/avoidKotlinPropertyDelegates-do/kotlin">
+    <img src="https://img.shields.io/badge/Download%20Project-GitHub-blue?logo=github&style=flat" alt="Download"/>
+  </a>
+</div>
+++++
+
+[.multi-language-sample]
+=====
+[source, kotlin]
+.build.gradle.kts
+----
+include::{snippetsPath}/bestPractices/avoidKotlinPropertyDelegates-do/kotlin/build.gradle.kts[tags=do-this]
+----
+=====
+
+<1> Configuration name is an explicit string `"myLibs"`. The variable can be renamed freely.
+<2> `extra.set` explicitly names the property. Use `extra["buildNumber"]` to read it back.
+<3> Task name is an explicit string `"greet"`. The variable can be renamed without affecting the task name.
+
+=== References
+
+- <<kotlin_dsl.adoc#kotdsl:kotlin_dsl,Kotlin DSL Primer>>
+- link:https://kotlinlang.org/docs/delegated-properties.html[Kotlin Delegated Properties]
+
+=== Tags
+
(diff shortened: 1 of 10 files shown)


platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_general.adoc
```diff
@@ -808,3 +808,72 @@ This structure uses an extension type to hold values, allowing properties to be
...
+[.multi-language-sample]
+=====
+[source, kotlin]
+.build.gradle.kts
+----
+include::{snippetsPath}/bestPractices/avoidKotlinPropertyDelegates-avoid/kotlin/build.gradle.kts[tags=avoid-this]
+----
+=====
+
+<1> Configuration name is implicitly derived from the variable name `myLibs`. Renaming this variable silently renames the configuration.
```
adam-enko, line 843: I don't understand how renaming a variable is 'silent'. It would cause a compilation failure, unless the variable was renamed in all places.

platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_general.adoc
```diff
@@ -808,3 +808,72 @@ This structure uses an extension type to hold values, allowing properties to be
...
+Use explicit API calls like `tasks.register("name")`, `configurations.create("name")`, and `extra.set("name", value)` instead of the Kotlin DSL property delegate forms `by tasks.registering`, `by configurations.creating`, and `by extra`.
+
+=== Explanation
+
+The Kotlin DSL provides delegate extensions that derive entity names from the Kotlin variable name.
+This affects tasks (`by tasks.registering`, `by tasks.creating`), configurations (`by configurations.creating`), extra properties (`by extra`), and Gradle properties (`by project`, `by settings`).
+While these can appear concise, Kotlin property delegates are somewhat uncommon, even among experienced Kotlin developers, and can not be easily reproduced in other Gradle DSLs.
+
+The main risk is *implicit name coupling*: the registered entity's name is silently tied to the variable name.
+Renaming a variable renames the task or configuration, which can break CLI invocations, CI scripts, and dependent builds without any compiler warning.
```
adam-enko, line 824: I'm confused by the relevance of this statement in this context. I understand how renaming something could cause runtime failures, but I don't follow how this makes using delegates undesirable. Any renaming could cause problems.

platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_general.adoc
```diff
@@ -808,3 +808,72 @@ This structure uses an extension type to hold values, allowing properties to be
...
+[[avoid_kotlin_property_delegates]]
+== Use Explicit API Instead of Kotlin Property Delegates
+
+Use explicit API calls like `tasks.register("name")`, `configurations.create("name")`, and `extra.set("name", value)` instead of the Kotlin DSL property delegate forms `by tasks.registering`, `by configurations.creating`, and `by extra`.
+
+=== Explanation
+
+The Kotlin DSL provides delegate extensions that derive entity names from the Kotlin variable name.
+This affects tasks (`by tasks.registering`, `by tasks.creating`), configurations (`by configurations.creating`), extra properties (`by extra`), and Gradle properties (`by project`, `by settings`).
+While these can appear concise, Kotlin property delegates are somewhat uncommon, even among experienced Kotlin developers, and can not be easily reproduced in other Gradle DSLs.
```
adam-enko, line 821: > Kotlin property delegates are somewhat uncommon

Is this true?

On GitHub I see 10.4k `.kts` files with ` by tasks.`
https://github.com/search?q=%2F+by+tasks%5C.%2F+path%3A**%2F*.gradle.kts+%28NOT+is%3Afork%29&type=code

but only 1.7k `.kts` files with ` = tasks.create(` or ` = tasks.register(`
https://github.com/search?q=%28%2F+%3D+tasks%5C.create%5C%28%2F+OR+%2F+%3D+tasks%5C.register%5C%28%2F+%29++path%3A**%2F*.gradle.kts+%28NOT+is%3Afork%29&type=code

I also believe most of Gradle's examples use property delegates.

platforms/documentation/docs/src/snippets/bestPractices/avoidKotlinPropertyDelegates-avoid/kotlin/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
+// tag::avoid-this[]
+val myLibs by configurations.creating // <1>
+var buildNumber by extra(42) // <2>
```
adam-enko, line 3: I don't understand what `buildNumber` is supposed to demonstrate in this example, because it's not used. Could the example be updated to demonstrate how it could cause an unexpected result?

platforms/documentation/docs/src/snippets/bestPractices/avoidKotlinPropertyDelegates-avoid/kotlin/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
...
+
+val greet by tasks.registering { // <3>
+    val configName = myLibs.name
+    doLast {
+        println("Configuration: $configName")
+    }
+}
```
adam-enko, line 10: I think this example would better demonstrate the problem by changing `val configName = myLibs.name` to `val configName = configurations.named("myLibs")` - thus demonstrating that renaming `val myLibs` to something else would cause a failure. 

However, I think this would be a weak example for demonstrating this need for this best-practice, because it would just demonstrate the benefits of the delegate property approach, since `val myLibs` can be renamed in an IDE and so it would automatically update.

adam-enko (commented):
I am against this proposal. I hope you make sure discouraging such a common and convenient pattern is really necessary, before it triggers a lot of churn in updating all the affected code.

Creating tasks/configurations in buildscripts using the delegates is really very convenient. It saves duplicating the name. It helps make sure renames are propagated. It encourages using references. It can help prevent mistakenly reusing the same name.

For extra properties, aren't they going to be deprecated with the Provider API migration? Maybe I'm misremembering. In any case, I personally try to avoid using them, so if this best-practice was updated to only apply to extra properties then I'd be okay with that.

For plugin authors: I agree that delegates should be considered very carefully (they can unexpectedly affect the generated DSL accessors, and there's no easy way to run a binary-compatibility check on those).

platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_general.adoc
tresat: It isn't obvious that the variable controls the name of the configuration.  Depending on how the configuration is obtained in other build logic, changing the variable may lead to successful build script compilation, but logic errors at runtime, instead of a preferred fail-fast behavior.

platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_general.adoc
tresat: Renaming other variables will either break compilation, or (if done through some automatic refactoring tool) be completely correct and result in 0 behavior changes.

That renaming a variable could have an additional effect is a surprising behavior to most people not familiar with this delegate mechanism.

platforms/documentation/docs/src/docs/userguide/reference/best-practices/best_practices_general.adoc
tresat: The Gradle docs not fully adhering to our own best practices is, sadly, a very known issue.  We're working on improving them, but there are many other higher priority tasks, even in terms of documentation tasks alone.

`by tasks` is arguably an even larger problem than the other delegates, as now _even valid CLI invocations_ can be made invalid by merely renaming a variable.  This is very poor UX and stomps on the principle of least surprise.

It's not even saving much typing versus `tasks.register()`.

But I take your point about it being fairly common.  I'll reword that.

platforms/documentation/docs/src/snippets/bestPractices/avoidKotlinPropertyDelegates-avoid/kotlin/build.gradle.kts
tresat: Good idea.

platforms/documentation/docs/src/snippets/bestPractices/avoidKotlinPropertyDelegates-avoid/kotlin/build.gradle.kts
tresat: Good update idea, and thanks for all your feedback here.

Indeed, some IDEs would understand the delegate mechanics and automatically update.  But editing Gradle scripts isn't limited to those IDEs, which is part of the motivation for recommending usage patterns that don't rely on IDE assistance to work well.

tresat:
This is probably no longer necessary due to https://github.com/gradle/gradle/pull/37556.

tresat:
> This is probably no longer necessary due to #37556.

@eskatos - Unless you disagree.

PR #37367, opened 2026-03-31
Implement overnight AGP version update pipelines

### LINKED ISSUE #37364: Make the AGP plugin updates more automatable
## Context

Our `updateAgpVersions` task takes care of upgrading a [`agp-versions.properties`](https://github.com/gradle/gradle/blob/57c1cb3c39e0154f09a15ab13e35de21f0a6899f/gradle/dependency-management/agp-versions.properties) file based on the two main streams of development:
 - their release branch, that has semantic versions like `9.2.0-alpha05`
 - their development branch that has a build ID-based version: `15054661`

This means that we have two different cadences with different importances:
 - releases, that we should pick up ASAP
 - nightly builds, that should be kept up with some periodicity

## Problem

Our upgrade process currently is manual: one of us runs `updateAgpVersions`, pushes it, makes a PR, and then merges it.

We should improve this pipeline to be more automated: we already have tests in place to catch if we have issues with the releases or not.

## Proposed solution

The `updateAgpVersions` should be extended to be able to allow or disallow nightly updates (i.e. a `--no-nightly-builds` flag). This would allow us to execute the pipeline below with a daily cadence:
 - Every day: only update _stable_ versions
 - One day in a week (e.g. Sunday): update _all_ versions

As written in context, this allows us to pick up stable versions ASAP, meanwhile reducing the churn of updating the nightly versions.

### Technical implementation

We should rely on CI to do most of the heavy lifting here.
A new pipeline would:
 - checkout a branch `android/agp-update-<<YYYYMMDD>>`
 - run the `updateAgpVersions` task
   - weekdays 1-6: `--no-nightly-builds`
   - weekday 7: no flag (or pass `--nightly-builds`)
 - commit and push `agp-versions.properties`
 - create PR, and start the merge

If we have no problems, the PR will be automatically merged.
If not, the PR will fail, and the TIDE team can have a look.

hegyibalint:
Fixes #37364 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/.teamcity/scripts/update_agp_and_create_pr.sh
+++ b/.teamcity/scripts/update_agp_and_create_pr.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+set -e
+
+# Script to update AGP versions and create a pull request
+#
+# Usage:
+#   ./update_agp_and_create_pr.sh
+#
+# Environment variables:
+#   DEFAULT_BRANCH  - The default branch to create the pull request on (e.g. "master")
+#   GITHUB_TOKEN    - GitHub bot token
+
+post() {
+    local endpoint="$1"
+    local data="$2"
+
+    local response=$(curl -X POST \
+        -H "Authorization: token $GITHUB_TOKEN" \
+        -H "Accept: application/vnd.github.v3+json" \
+        -H "Content-Type: application/json" \
+        "https://api.github.com/repos/gradle/gradle$endpoint" \
+        -d "$data" \
+        -w "\n%{http_code}" \
+        2>/dev/null)
+
+    local http_code=$(echo "$response" | tail -n1)
+    local body=$(echo "$response" | head -n -1)
+
+    if [[ "$http_code" -ge 400 ]]; then
+        printf "Error: HTTP %s - %s\n" "$http_code" "$body" >&2
+        exit 1
+    fi
+
+    echo "$body"
+}
+
+main() {
+    : "${DEFAULT_BRANCH:?DEFAULT_BRANCH environment variable is required}"
+    : "${GITHUB_TOKEN:?GITHUB_TOKEN environment variable is required}"
+
+    ./gradlew updateAgpVersions
+
+    git add gradle/dependency-management/agp-versions.properties
+    git add platforms/documentation/docs/src/docs/userguide/releases/compatibility.adoc
+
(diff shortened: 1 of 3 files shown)


.teamcity/scripts/update_agp_and_create_pr.sh
```diff
@@ -0,0 +1,92 @@
...
+        echo "No changes to commit"
+        exit 0
+    fi
+
+    BRANCH_NAME="tide/update-agp-$(date +%Y%m%d-%H%M%S)"
+    git switch -c $BRANCH_NAME
+    git commit --signoff --author="bot-gradle <bot-gradle@gradle.com>" -m "Update AGP versions"
+    git push https://${GITHUB_TOKEN}@github.com/gradle/gradle.git $BRANCH_NAME
+
+    PR_TITLE="Update AGP versions"
```
eskatos, line 66: 💅 
```suggestion
    PR_TITLE="Update tested AGP versions"
```

.teamcity/scripts/update_agp_and_create_pr.sh
```diff
@@ -0,0 +1,92 @@
...
+    echo "PR_RESPONSE: $PR_RESPONSE"
+
+    PR_NUMBER=$(echo "$PR_RESPONSE" | jq -r '.number' 2>/dev/null)
+
+    post "/issues/$PR_NUMBER/comments" '{
+        "body": "@bot-gradle test and merge"
+    }'
+
+    post "/issues/$PR_NUMBER/labels" '{
+        "labels": ["re:android"]
```
eskatos, line 84: 💅 Using similar labels as past manual PRs:
```suggestion
        "labels": ["a:chore", "in:performance-test", "in:smoke-test", "re:android"]
```

hegyibalint:
@eskatos I've stripped down this change based on our observations:
 - daily change is too much, because if we have problems, PRs would start to pile up
 - we came to the conclusion that instead of throttling PRs, we could keep a weekly cadence: simpler solution, the responsible can manage a weekly new PR

That cut away a lot of the changes here. No `--nightly-build` flag, no differentiation.
One unified update that would happen on Saturday at 22:00. That's all.

PR #37373, opened 2026-03-31
Extract org.gradle.caching into build-cache-api and build-cache-core

ghale:
Move the org.gradle.caching package family out of core and core-api into two new projects in the core-execution platform:

- build-cache-api: public API classes (BuildCacheConfiguration, BuildCache, BuildCacheServiceFactory, etc.) moved from core-api
- build-cache-core: internal implementation classes (controllers, factories, local-cache wiring) and their tests moved from core

BuildCacheServices stays in core due to its dependency on BuildState and RootBuildState.

Update dependency declarations in downstream projects:
- core-api, core, build-cache-http, configuration-cache, kotlin-dsl, and internal-integ-testing now declare direct dependencies on the new projects as needed.
- Dependency classifications (api vs implementation) validated via projectHealth on all affected projects.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

Issue #36267, opened 2026-01-14
DeprecationLogger has not been initialized error when using TestKit to run Gradle 3.4


wolfs:
### Current Behavior

When trying to run Gradle 3.4 builds via test kit we get the following error:
```
java.lang.IllegalStateException: DeprecationLogger has not been initialized. Most probably, it's because you are trying to use it from the launcher/wrapper. It's not available there. Move the deprecation logging to the daemon or use LOGGER.warn() instead. Another reason could be that you are trying to use it from a unit test. In that case, either fix the test to not use deprecated features, or mark it with @ExpectDeprecation. If you hit this error as a user of Gradle, please report it as a bug. The original deprecation message was: The version of Gradle you are using (3.4) is deprecated with TestKit. TestKit will only support the last 5 major versions in future. has been deprecated. This will fail with an error in Gradle 10. For more information, please refer to https://docs.gradle.org/9.2.1/userguide/third_party_integration.html#sec:embedding_compatibility in the Gradle documentation.   
at org.gradle.internal.deprecation.DeprecationLogger.nagUserWith(DeprecationLogger.java:238)
at org.gradle.internal.deprecation.DeprecationMessageBuilder$WithDocumentation.nagUser(DeprecationMessageBuilder.java:197)
at org.gradle.testkit.runner.internal.ToolingApiGradleExecutor.checkDeprecationWarning(ToolingApiGradleExecutor.java:201) |  
at org.gradle.testkit.runner.internal.ToolingApiGradleExecutor.run(ToolingApiGradleExecutor.java:127)
at org.gradle.testkit.runner.internal.DefaultGradleRunner.run(DefaultGradleRunner.java:360)
at org.gradle.testkit.runner.internal.DefaultGradleRunner.build(DefaultGradleRunner.java:273)
```

https://github.com/gradle/gradle/blob/7e720deb862a7e7e38eb2a7ad3322b3772763303/platforms/extensibility/test-kit/src/main/java/org/gradle/testkit/runner/internal/feature/BuildResultOutputFeatureCheck.java#L52

### Expected Behavior

No error, Gradle 9.x still should support running builds with Gradle 3.4 via TestKit.

### Context (optional)

* https://github.com/gradle/gradle/pull/33239

### Self-contained Reproducer Project

Try to run a Build using Gradle 3.4 via TestKit when using Gradle 9.2.

### Gradle version

9.2

h0tk3y:
[...]

PR #37378, opened 2026-03-31
Use JUL logger instead of DeprecationLogger for TestKit unsupported version warnings

h0tk3y:
Fixes:
* https://github.com/gradle/gradle/issues/36267

### DIFF
--- a/platforms/extensibility/test-kit/src/integTest/groovy/org/gradle/testkit/runner/GradleRunnerUnsupportedGradleVersionFailureIntegrationTest.groovy
+++ b/platforms/extensibility/test-kit/src/integTest/groovy/org/gradle/testkit/runner/GradleRunnerUnsupportedGradleVersionFailureIntegrationTest.groovy
@@ -16,6 +16,7 @@
 
 package org.gradle.testkit.runner
 
+import org.gradle.integtests.fixtures.versions.ReleasedVersionDistributions
 import org.gradle.testkit.runner.fixtures.NonCrossVersion
 import org.junit.Assume
 
@@ -28,7 +29,7 @@ import static org.gradle.tooling.internal.consumer.DefaultGradleConnector.MINIMU
 @NonCrossVersion
 class GradleRunnerUnsupportedGradleVersionFailureIntegrationTest extends BaseGradleRunnerIntegrationTest {
 
-    // Gradle 3.5.1 is above RUN_BUILDS.since (2.6) but below MINIMUM_SUPPORTED_GRADLE_VERSION (4.0),
+    // Pick a Gradle version above RUN_BUILDS.since (2.6) but below MINIMUM_SUPPORTED_GRADLE_VERSION,
     // so it triggers the deprecation warning path in ToolingApiGradleExecutor/BuildResultOutputFeatureCheck.
     // Regression test for https://github.com/gradle/gradle/issues/36267
     def "succeeds running build with deprecated Gradle version below minimum supported version"() {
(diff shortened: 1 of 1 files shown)


platforms/extensibility/test-kit/src/main/java/org/gradle/testkit/runner/internal/feature/BuildResultOutputFeatureCheck.java
```diff
@@ -49,11 +52,19 @@ public void verify() {
...
+            LOGGER.warning(messageSupplier.apply(targetGradleVersion.getVersion())
+                + " TestKit will only support the last 5 major versions in future."
+                + " This will fail with an error starting with Gradle 10."
+                + " Consult the upgrading guide for further information: https://docs.gradle.org/current/userguide/tooling_api.html#sec:embedding_compatibility");
         }
+
+        // DeprecationLogger is not normally available in this context, which is outside Gradle runtime.
+        // Keep the call to DeprecationLogger in unreachable code so this deprecation can still be found by usage of DeprecationLogger and is addressed later.
+        @SuppressWarnings("unused")
+        Runnable keepTheDeprecationLoggerCall = () -> DeprecationLogger.deprecate("Running TestKit with unsupported old Gradle versions.")
```
donat, line 64: ❌ This doesn't seem to be used anywhere. Am I missing something?

platforms/extensibility/test-kit/src/integTest/groovy/org/gradle/testkit/runner/GradleRunnerUnsupportedGradleVersionFailureIntegrationTest.groovy
```diff
@@ -17,11 +17,63 @@
...
+
+    // Gradle 3.5.1 is above RUN_BUILDS.since (2.6) but below MINIMUM_SUPPORTED_GRADLE_VERSION (4.0),
+    // so it triggers the deprecation warning path in ToolingApiGradleExecutor/BuildResultOutputFeatureCheck.
+    // Regression test for https://github.com/gradle/gradle/issues/36267
+    def "succeeds running build with deprecated Gradle version below minimum supported version"() {
+        // The old Gradle versions fail to run on the modern JDKs, so this test is embedded-only
+        Assume.assumeTrue(embedded)
+
+        given:
+        def oldGradleVersion = "3.5.1"
```
donat: 🤔 Is there an easy way to test agains the most recent unsupported Gradle version, instead of using a hardcoded version?

platforms/extensibility/test-kit/src/main/java/org/gradle/testkit/runner/internal/feature/BuildResultOutputFeatureCheck.java
h0tk3y: The comment above explains my idea:

> Keep the call to DeprecationLogger in unreachable code so this deprecation can still be found by usage of DeprecationLogger and is addressed later.

Do you think there is a better way to keep this deprecation discoverable?

donat: On a second thought, it makes sense to have this unreachable code. Maybe creating an issue and mentioning in the comment would be useful. It's just an idea; I'm unblocking this PR.

platforms/extensibility/test-kit/src/integTest/groovy/org/gradle/testkit/runner/GradleRunnerUnsupportedGradleVersionFailureIntegrationTest.groovy
h0tk3y: Done, it seems that we have `ReleasedVersionDistributions` for that.

PR #37383, opened 2026-04-01
Remove TestPerformanceTest

blindpirate:
This PR removes `TestPerformanceTest`, which is an offline performance test. It's blocking the AWS migration because it's in ReadyForRelease project but we don't want to assign the whole ReadyForRelease project to bare metal pool. This build is almost never used now, so let's remove it.


blindpirate:

blindpirate:
@bot-gradle cherrypick to release7x

PR #37384, opened 2026-04-01
Fix packageInfoData cache key for relative path of package-info files

blindpirate:
`packageInfoData` fingerprints its file inputs relative to each source root, which can hide moves between projects and reuse stale cached output. Add a derived input with repo-relative package-info paths so the cache key changes when files move while remaining relocatable across machines.

### DIFF
--- a/build-logic-settings/architecture-docs/src/main/kotlin/gradlebuild/GeneratePackageInfoDataTask.kt
+++ b/build-logic-settings/architecture-docs/src/main/kotlin/gradlebuild/GeneratePackageInfoDataTask.kt
@@ -24,6 +24,7 @@ import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.CacheableTask
+import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFiles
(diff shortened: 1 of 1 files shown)


build-logic-settings/architecture-docs/src/main/kotlin/gradlebuild/GeneratePackageInfoDataTask.kt
```diff
@@ -52,6 +53,11 @@ abstract class GeneratePackageInfoDataTask : DefaultTask() {
     @get:PathSensitive(PathSensitivity.RELATIVE)
     abstract val packageInfoFiles: ConfigurableFileCollection
 
+    @get:Input
+    // Preserve the project path in the cache key, since packageInfoFiles are fingerprinted relative to each source root.
+    val packageInfoFilePaths: List<String>
+        get() = packageInfoFiles.files.map { it.relativeTo(baseDir).path.replace(File.separatorChar, '/') }.sorted()
```
blindpirate, line 59: `baseDir` is root project dir (`settingsDirectory`).

cobexer:
@blindpirate please rebase & merge towards `release` to make sure builds on release don't pick up wrong outputs from the cache!

blindpirate:
As we discussed, this doesn't need to be in `release` as `master` and `release` won't share build cache (different Gradle version and build classpath).

PR #37386, opened 2026-04-01
Speed up FileBackedBlockStore in backward compatible way

### LINKED ISSUE #37406: Slow fingerprinting and incremental compilation on some cloud storages
On some cloud storages (like AWS with default EBS disk) where IOPs are limited we see slow build times, see analysis:
https://github.com/gradle/gradle/issues/28236#issuecomment-4067203021

The problem is aggressive padding in BTreeIndexCache, w have a potential fix for it https://github.com/gradle/gradle/pull/28344.
The fix has some problems, like if daemon crashes then cache can get corrupted. Alternative approach is https://github.com/gradle/gradle/pull/37386. Alternatively we could also try different storage, like MVStore or MapDB.

A workaround is, to use faster disk, e.g. SSD local disk on AWS.

Related issue:
https://github.com/gradle/gradle/issues/28236

asodja:


asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

big-guy:
@coderabbitai review

cobexer:
@coderabbitai review

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/btree/FileBackedBlockStore.java
```diff
@@ -75,6 +76,11 @@ private RandomAccessFile randomAccessFile(String mode) throws FileNotFoundExcept
     @Override
     public void close() {
         try {
+            // Trim chunk-growth padding so the next open() starts allocating right after the last block,
+            // and we don't waste disk space. If a daemon is killed before close is called, we still may waste max 64KB.
```
big-guy: Can we refer to FILE_GROWTH_CHUNK_SIZE here instead of 64KB?

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/btree/FileBackedBlockStore.java
```diff
@@ -75,6 +76,11 @@ private RandomAccessFile randomAccessFile(String mode) throws FileNotFoundExcept
...
+            if (nextBlock < currentFileSize) {
+                file.setLength(nextBlock);
+            }
             file.close();
         } catch (IOException e) {
```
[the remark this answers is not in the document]
big-guy: This looks like a possibly real problem, which might make it harder to recover on OSes like Windows.

e.g., assume that `close` would not fail, but `setLength` does.

setLength fails and then we do not close the file. The exception bubbles up into Gradle internals, so it's unclear what we would do, but hopefully this is happening on shutdown/end of build.

For Linux, it doesn't matter much if the file is left open. We might leak resources in a long running daemon, but as long as we aren't still writing to it, it wouldn't affect other builds.

For Windows, if the file is left open, we could have a new problem where one daemon keeps the file open and is in a zombie/bad state and new daemons are unable to reopen this file. 

This wouldn't happen since we only close the file today.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/btree/FileBackedBlockStore.java
```diff
@@ -222,10 +228,13 @@ public void write() throws Exception {
             outputStream.writeInt((int) bytesWritten);
             output.done();
 
-            // Pad
+            // Grow the file in 64KB chunks instead of exact sizes to reduce ftruncate syscalls.
```
big-guy: Same thing here. Refer to FILE_GROWTH_CHUNK_SIZE instead of 64KB.

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/btree/FileBackedBlockStore.java
```diff
@@ -222,10 +228,13 @@ public void write() throws Exception {
...
+            // On cloud block storage (e.g. EBS), ftruncate triggers expensive metadata journal
+            // commits. Growing it in chunks reduces the number of ftruncate syscalls and increases performance on such storages.
             if (currentFileSize < finalSize) {
-                file.setLength(finalSize);
-                currentFileSize = finalSize;
+                long newSize = (finalSize + FILE_GROWTH_CHUNK_SIZE - 1) / FILE_GROWTH_CHUNK_SIZE * FILE_GROWTH_CHUNK_SIZE;
```
big-guy: Could we move `finalSize` down to here and calculate finalSize with chunk alignment so we don't have another similarly named variable?

Can finalSize be calculated with `pos + getSize()` vs `pos + HEADER_SIZE + TAIL_SIZE + payloadSize`?

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/btree/BTreeIndexedCacheTest.java
```diff
@@ -249,13 +250,12 @@ public void canHandleLargeNumberOfEntries() {
 
         checkAddsAndRemoves(Collections.reverseOrder(), values);
 
-        // need to make this better
-        assertTrue(cacheFile.length() < (long)(1.4 * len));
+        // Tolerance accounts for free list fragmentation and chunk-aligned file growth
+        assertTrue(cacheFile.length() < 2 * len);
```
big-guy: Shouldn't we know what the length will be exactly? What do you mean by free list fragmentation here?

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/btree/BTreeIndexedCacheTest.java
```diff
@@ -373,6 +374,44 @@ public void canUseFileAsKey() {
         cache.close();
     }
 
+    @Test
+    public void fileGrowsInChunksToReduceSystemCalls() {
+        createCache();
+        for (int i = 0; i < 100; i++) {
```
big-guy: Can we write enough entries that it causes a resize of the file? Right now, I think we only go up to the first chunk and stay there.

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/btree/BTreeIndexedCacheTest.java
```diff
@@ -373,6 +374,44 @@ public void canUseFileAsKey() {
...
+        cache.reset();
+        for (int i = 0; i < 100; i++) {
+            assertThat(cache.get("key_" + i), equalTo(i));
+        }
+
+        verifyAndCloseCache();
+    }
+
+    @Test
+    public void chunkGrowthPaddingIsTrimmedOnClose() {
```
big-guy, line 402: Can we add a test that takes a good cache file, resizes it to be chunk aligned without writing any new data (with RandomAccessFile directly) and then try to read the data out of it again?

This would simulate:
- Gradle uses the cache, sizes the file to be chunk aligned but does not close the file
- Gradle process is killed
- New Gradle process tries to read the file again

big-guy (changes requested):
This looks very promising.

asodja:
@bot-gradle test this APT

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/btree/BTreeIndexedCacheTest.java
asodja: I think it's a bit tricky to figure our exact size, this was calculated by AI

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/btree/FileBackedBlockStore.java
asodja: Good point, reorganized code a bit. I hope it's more readable now

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/btree/FileBackedBlockStore.java
asodja: Should be fixed

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/btree/FileBackedBlockStore.java
asodja: Should be fixed

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/btree/BTreeIndexedCacheTest.java
asodja: Should be fixed now

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/btree/BTreeIndexedCacheTest.java
asodja: Added additional test

platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/btree/FileBackedBlockStore.java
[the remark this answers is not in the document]
asodja: Should be fixed now

asodja:
@bot-gradle test this APT

platforms/core-execution/persistent-cache/src/test/groovy/org/gradle/cache/internal/btree/BTreeIndexedCacheTest.java
asodja: I now use exact size that was measured. But predicting it is harder

asodja:
@bot-gradle test this APT

big-guy (approved):

PR #37387, opened 2026-04-01
Add initial Android Sync with IP performance tests

alllex:
### New performance tests

- Performance regression tests
  - Setup: IP enabled
  - Scenarios
    - Studio Sync after build logic ABI change: cold daemon, warn daemon
- Performance comparison tests
  - Lines:
    - "Moderne" -- modern setup with Parallel CC and Parallel Execution, realistically used by performance-aware teams
    - Moderne-CoD -- modern setup plus Configure-on-Demand, since it's popular in the Android world
    - IP -- default IP as enabled by the main flag
  - Scenarios
    - Studio Sync after build logic ABI change
  
Test projects:
- `android500Kts` -- beefy Android project using latest AGP (https://github.com/gradle/isolated-projects-testbed/tree/dcf1dbe751bdbf82e766472fe45695a78081af46/android-500-kts)

### Continuous testing

Because the test project is so big and because Studio sync takes long, all the tests are treated as slow performance tests and are made part of the `ReadyForRelease`.

- Each regression test takes ~1h to run
- Comparison test takes [~2h to run](https://builds.gradle.org/buildConfiguration/Gradle_Master_Util_Performance_AdHocPerformanceScenarioLinuxAmd64/111690353)

We can tweak the warm-up and runs counts to reduce the time, but it's best we do it once we get a better idea of how stable the measurements are.

In the future, we plan to run the same tests with smaller projects, which can allow making them part of `ReadyForNightly`.


alllex:
@bot-gradle test apt

alllex:
@bot-gradle test apt

build-logic/performance-testing/src/main/groovy/gradlebuild.performance-templates.gradle
```diff
@@ -422,6 +422,15 @@ performanceTest.registerAndroidTestProject("nowInAndroidBuild", ["-Xms4g", "-Xmx
     }
 }
 
+// When manually opening the project in Android Studio 2025.3, it recommends this max memory based on the project size
+performanceTest.registerAndroidTestProject("android500Kts", ["-Xms8g", "-Xmx8g"], RemoteProject) {
+    setRemoteUriAndRefFromGradleProperty("isolatedProjectsTestbedRef")
+    subdirectory = 'android-500-kts'
+    doLast {
+        setMaxWorkers(outputDirectory, 8)
```
6hundreds, line 430: ❓ `outputDirectory` or `projectDirectory`?

testing/internal-performance-testing/src/main/groovy/org/gradle/performance/fixture/TestProjects.groovy
```diff
@@ -29,7 +29,6 @@ class TestProjects {
         if (!jvmArgs.find { it.startsWith("-Xmx") }) {
             throw new IllegalArgumentException("Test project needs to specify -Xmx in gradle.properties. org.gradle.jvmargs = ${jvmArgs?.join(' ')}")
         }
-        verifyGradlePropertiesSettingSpecified(gradleProperties, "org.gradle.parallel")
```
6hundreds, line 32: 💅 Then the line above has to be adjusted as well
https://github.com/gradle/gradle/blob/1a5e48701b602def428d9c79e3793adff7a4f798/testing/internal-performance-testing/src/main/groovy/org/gradle/performance/fixture/TestProjects.groovy#L23

testing/performance/src/performanceTest/groovy/org/gradle/performance/crossbuild/ip/IsolatedProjectsAndroidSyncPerformanceComparisonTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+                    "-Dorg.gradle.configuration-cache.parallel=true",
+                    "-Dorg.gradle.configureondemand=true",
+                )
+            }
+        }
+
+        runner.buildSpec {
+            displayName("ip")
+            invocation {
+                args(
```
6hundreds, line 80: 💭  We don't care about `-Dorg.gradle.parallel=true` since we're exercising sync, where `--parallel` has no benefit, right?

testing/performance/src/performanceTest/groovy/org/gradle/performance/crossbuild/ip/IsolatedProjectsAndroidSyncPerformanceComparisonTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+    def "sync Studio after included build logic refactoring"() {
+        given:
+        def runner = getRunner() // otherwise, IDEA thinks it's PerformanceTestRunner despite the override
+
+        runner.addBuildMutator { settings ->
+            new ApplyAbiChangeToKotlinSourceFileMutator(new File(settings.projectDir, "build-logic/convention/src/main/kotlin/org/example/awesome/utils.kt"))
+        }
+
+        // 'Moderne' configuration that is used by performance aware teams
+        runner.baseline {
```
6hundreds, line 52: ❓ Is it ok to have 2 baselines here and on the line 65?

testing/performance/src/performanceTest/groovy/org/gradle/performance/crossbuild/ip/IsolatedProjectsAndroidSyncPerformanceComparisonTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+        !moderneWithConfigureOnDemand.significantlyFasterThan(ip)
+    }
+
+    @Override
+    protected void defaultSpec(GradleBuildExperimentSpec.GradleBuilder builder) {
+        builder.warmUpCount(5)
+            .invocationCount(10)
+            .invocation {
+                args(
+                    // realistic defaults
```
6hundreds, line 109: 🤔 If we care about realistic defaults, should we add `--parallel`? Another question is `fs-watching` feature is a realistic default as well and does it have any effect on sync

build-logic/performance-testing/src/main/groovy/gradlebuild.performance-templates.gradle
alllex: It's `RemoteProject.outputDirectory`, the directory where the template is copied.

testing/internal-performance-testing/src/main/groovy/org/gradle/performance/fixture/TestProjects.groovy
alllex: We still require max-workers, so I think it's fine to leave the message as is.

testing/performance/src/performanceTest/groovy/org/gradle/performance/crossbuild/ip/IsolatedProjectsAndroidSyncPerformanceComparisonTest.groovy
alllex: Yes, the `baseline` method is basically an alias for `buildSpec`. It's for readability only

testing/performance/src/performanceTest/groovy/org/gradle/performance/crossbuild/ip/IsolatedProjectsAndroidSyncPerformanceComparisonTest.groovy
alllex: Parallel Execution is automatically enabled when IP is enabled, so configuring it here would have been redundant.

testing/performance/src/performanceTest/groovy/org/gradle/performance/crossbuild/ip/IsolatedProjectsAndroidSyncPerformanceComparisonTest.groovy
alllex: The `--parallel` setting varies between the (base)lines in the method above. It's enabled explicitly in the moderne and moderne-cod, and implicitly for IP.

The defaults here are those that are shared across all the (base)lines. So `fs-watching` is a good candidate. I'd include in a follow-up after checking its effects on sync.

PR #37389, opened 2026-04-01
Remove lambda-taking methods from feature binding builders

### LINKED ISSUE #37125: DCL: Remove apply action via lambda api
### Expected Behavior

There should be 1 way to apply a DCL feature: via a dedicated `ApplyAction` class.

There are 2 ways to apply a DCL feature:
1. via a dedicated `ApplyAction` class.
2. via an ad-hoc lambda.

### Context

The api initially only supported the configuration with a lambda, but now it also supports an ApplyAction.
Having two ways to do the same thing is confusing.

By only keeping the `ApplyAction` class approach, there is also no need for the `ProjectFeatureApplicationContext.getObjectFactory`, it can be injected if needed.

ghale:
Remove lambda-based bindProjectType/bindProjectFeature overloads in favor of class-based ProjectTypeApplyAction/ProjectFeatureApplyAction implementations. 

Also remove the `getObjectFactory()` method from the public API for `ProjectFeatureApplicationContext`.

Fixes #37125 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -69,5 +69,5 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_17" default="true" project-jdk-name="17" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_17" project-jdk-name="17" project-jdk-type="JavaSDK" />
 </project>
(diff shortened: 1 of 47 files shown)


ghale:
@bot-gradle test this

eskatos (approved):

Issue #30073, opened 2024-08-01
Check that all `Property<Boolean>` upgraded methods has Kotlin source compatibility equivalent


asodja:
After we migrate all planned properties to lazy types, we should check that all  `Property<Boolean>` `getX` methods contain Kotlin source compatibility `getIsX` method.

This can be done just one time before we 9.0, so maybe we don't need any automation at all, just a manual sanity check.

asodja:
Also, let's remove deprecations for all of them

asodja:
This was done on Provider API migration branch with https://github.com/gradle/gradle/pull/37399

PR #37399, opened 2026-04-02
[infra] Undeprecate getIsX() property getters and add missing ones

asodja:
We can deprecate them later, but they are disturbance when doing a migration

Fixes https://github.com/gradle/gradle/issues/30073


asodja:
@bot-gradle test this

PR #37402, opened 2026-04-02
Change getContentCharset() signature back to String

### LINKED ISSUE #37400: NoSuchMethodError with Gradle 9.5.0 RC1 and Intellij Idea 2026.1
### Current Behavior

See https://youtrack.jetbrains.com/issue/IDEA-387893/Gradle-9.5-RC1-incompatible-with-IDEA-2026.1-Ultimate.

Intellij is calling a Gradle-internal method which has changed its signature.

### Expected Behavior

We should find a way to keep this working, even though they shouldn't be calling an internal method for this information.  

### Context (optional)

_No response_

This should be reproducible while syncing a build that invokes `WarModelBuilder` in Intellij 2026.1.

### Gradle version

9.5.0-rc-1

### Build scan URL (optional)

_No response_

_No response_

ghale:
Intellij is using `ManifestInternal.getContentCharset()` in a model builder which is causing a NoSuchMethod error now.  We're changing the signature back to keep this working.

Fixes #37400 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/platform-jvm/src/main/java/org/gradle/api/java/archives/internal/CustomManifestInternalWrapper.java
+++ b/platforms/jvm/platform-jvm/src/main/java/org/gradle/api/java/archives/internal/CustomManifestInternalWrapper.java
@@ -76,7 +76,7 @@ public class CustomManifestInternalWrapper implements ManifestInternal {
     }
 
     @Override
-    public @Nullable Provider<String> getContentCharset() {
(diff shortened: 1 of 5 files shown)


ghale:
@bot-gradle test this

PR #37404, opened 2026-04-02
TeamCity: test all cross version tests with the max supported Java version

cobexer:
- Change cross version test coverage to run with `JvmCategory.MAX_VERSION` instead of `MIN_VERSION` on both Linux and Windows
- IDEs embed the Tooling API but don't limit themselves to Java 8/17; running the TAPI client on the max Java version catches cross-version problems ahead of TAPI users


cobexer:
@bot-gradle test RfR

cobexer:
@bot-gradle test RfR

PR #37407, opened 2026-04-02
Extract parent from dynamic object delegates

reinsch82:
Refactor ExtensibleDynamicObject to walk the parent hierarchy explicitly instead of including the parent in the flat delegate array. This makes it possible to detect when a property or method lookup crosses a
project boundary.

Key changes:
- Add HierarchicalDynamicObject interface with getParent
- ExtensibleDynamicObject implements HierarchicalDynamicObject
- Parent removed from CompositeDynamicObject delegates
- Explicit parent-walk loops in hasProperty, tryGetProperty, hasMethod, and doTryInvokeMethod
- Move closure-as-method logic from deleted MixInClosurePropertiesAsMethodsDynamicObject into ExtensibleDynamicObject.tryInvokeMethod
- Make CompositeDynamicObject concrete with constructor
- Update CrossProjectModelAccess to use HierarchicalDynamicObject

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


subprojects/core-api/src/main/java/org/gradle/internal/metaobject/HierarchicalDynamicObject.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.gradle.internal.metaobject;
+
+import org.jspecify.annotations.Nullable;
+
+/**
+ * A dynamic object is a member of a hierarchy of dynamic objects.
+ *
```
alllex, line 22: 💅 Add paragraph separator and reformat javadoc

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/metaobject/CompositeDynamicObject.java
```diff
@@ -85,6 +85,7 @@ public DynamicInvokeResult trySetPropertyWithoutInstrumentation(String name, @Nu
     }
 
     @Override
+    @Deprecated
```
alllex: 🤔 We should either deprecate the method on the base interface or nothing at all. This is not the only override that is deprecated

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectModelAccessTrackingParentDynamicObject.kt
```diff
@@ -105,7 +113,7 @@ class CrossProjectModelAccessTrackingParentDynamicObject(
 
     private
     enum class MemberKind {
-        PROPERTY, METHOD
+        PROPERTY, METHOD, PARENT
```
alllex: ❓Is the new member unused?

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/metaobject/AbstractDynamicObject.java
```diff
@@ -29,6 +29,7 @@
  * An empty {@link DynamicObject}.
  */
 public abstract class AbstractDynamicObject implements DynamicObject {
+    @Override
```
alllex, line 32: ❔ Should we remove this method from here, since it's now on the interface?

subprojects/core-api/src/main/java/org/gradle/internal/metaobject/HierarchicalDynamicObject.java
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.gradle.internal.metaobject;
+
+import org.jspecify.annotations.Nullable;
+
+/**
+ * A dynamic object is a member of a hierarchy of dynamic objects.
```
alllex:
```suggestion
 * A dynamic object that is a member of a hierarchy of dynamic objects.
```

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectModelAccessTrackingParentDynamicObject.kt
```diff
@@ -24,20 +24,28 @@ import org.gradle.internal.configuration.problems.ProblemsListener
...
-    private val delegate: DynamicObject,
+    private val delegate: HierarchicalDynamicObject,
     private val referrerProject: ProjectInternal,
     private val problems: ProblemsListener,
     private val coupledProjectsListener: CoupledProjectsListener,
     private val problemFactory: ProblemFactory,
     private val dynamicCallProblemReporting: DynamicCallProblemReporting
-) : DynamicObject {
+) : HierarchicalDynamicObject {
+    override fun getDisplayName(): String {
```
alllex: 💅 Can we move this method after other overrides, since it's less significant, and also make sure there is an empty line between the class header and it's methods for readability?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectModelAccessTrackingParentDynamicObject.kt
```diff
@@ -135,7 +143,7 @@ class CrossProjectModelAccessTrackingParentDynamicObject(
                         }
 
                         // method lookup is more clear from the stack trace, so keep the minimal trace pointing to the location:
-                        MemberKind.METHOD -> location
+                        else -> location
```
alllex: ❓ If we don't need a new MemberKind, we probably don't need to change this

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -175,19 +171,148 @@ private DynamicObject snapshotInheritable() {
...
+    public DynamicInvokeResult tryInvokeMethod(String name, @Nullable Object... arguments) {
+        DynamicInvokeResult result = doTryInvokeMethod(name, arguments);
+        if (result.isFound()) {
+            return result;
+        }
 
-            @Override
-            public String getDisplayName() {
-                return dynamicDelegate.getDisplayName();
+        DynamicInvokeResult propertyResult = tryGetProperty(name);
```
alllex, line 280: 💭 Please add a high-level comment that explains what we do here and why

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -175,19 +171,148 @@ private DynamicObject snapshotInheritable() {
...
+        if (propertyResult.isFound()) {
```
alllex: 💅 Can avoid extra indent by returning early in the opposite case

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -161,12 +157,12 @@ public void addObject(DynamicObject object, Location location) {
...
      * @return an object containing the inheritable properties and methods of this object.
      */
-    public DynamicObject getInheritable() {
+    public HierarchicalDynamicObject getInheritable() {
         return new InheritedDynamicObject();
     }
 
     private DynamicObject snapshotInheritable() {
-        final List<DynamicObject> delegates = new ArrayList<>(4);
+        final List<DynamicObject> delegates = new ArrayList<>(5);
```
alllex: ❓ Why 5? Can we also drop `final` ?

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -91,21 +98,10 @@ private void updateDelegates() {
...
         System.arraycopy(delegates, 0, objects, 0, idx);
-        setObjects(objects);
-
-        if (addedParent) {
-            idx--;
-            objects = new DynamicObject[idx];
-            System.arraycopy(delegates, 0, objects, 0, idx);
-            setObjectsForUpdate(objects);
-        }
+        this.delegateObjects = new CompositeDynamicObject(objects, this::getDisplayName);
```
alllex: 💭 Now that this is simplified, we can replace `delegates` with an array list, in the same way  `snapshotInheritable()` is implemented

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -38,16 +46,17 @@
...
+public class ExtensibleDynamicObject extends AbstractDynamicObject implements HierarchicalDynamicObject {
 
     public enum Location {
         BeforeConventionNotInherited, BeforeConvention, AfterConvention
     }
 
     private final AbstractDynamicObject dynamicDelegate;
     @Nullable
-    private DynamicObject parent;
+    private HierarchicalDynamicObject parent;
```
alllex: 💅 Can we move the parent field right after `delegateObjects`, since the parent is semantically used after them in most methods?

My personal preference would also be to put the nullable annotation next to the type, instead of the separate line

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -56,8 +65,10 @@ public enum Location {
     private DynamicObject afterConvention;
     private final DynamicObject extraPropertiesDynamicObject;
 
+    private CompositeDynamicObject delegateObjects;
```
alllex, line 68: 💭 Please add a one-line high level comment what this field represents

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
reinsch82: we should define the place of the nullable annotation. I don't mind either way, but to align here would be nice.

alllex: I realized, we actually did, and so far it's to put it on another line. It's documented here: https://bt-internal-docs.grdev.net/gbt/how-to/develop/nullability-guide/?h=nulla#annotate-fields-methods-and-constructors

I personally don't like this, and would prefer to have it on the same line, so that it brings the nullability _closer_ to the type. This is also more similar to Kotlin, where the nullability is directly on the type. This is also similar to Java's upcoming [null-restricted types](https://openjdk.org/jeps/8303099). Having said that, there is no semantic difference, and there was some alignment to put them on the other line.

❌  The annotation definitely must not come before the visibility modifier, if it's on the same line

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectModelAccessTrackingParentDynamicObject.kt
alllex: Did you intentionally decide to keep `else`? If yes, can you share why?

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -38,26 +46,26 @@
...
-    private DynamicObject beforeConvention;
-    @Nullable
-    private DynamicObject afterConvention;
     private final DynamicObject extraPropertiesDynamicObject;
 
+    @Nullable private DynamicObject beforeConventionNotInherited;
+    @Nullable private DynamicObject beforeConvention;
+    @Nullable private DynamicObject afterConvention;
+
+    // The composite of this object's own delegates, excluding the parent hierarchy.
```
alllex: > composite of this object's own delegates

This mostly repeats the type and the field name, not adding useful context. Try adding something that can help the reader to understand what are the delegates without needing to look through all usages of the field.

The exclusion of the parent from there is a useful comment

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -161,12 +150,12 @@ public void addObject(DynamicObject object, Location location) {
...
      * @return an object containing the inheritable properties and methods of this object.
      */
-    public DynamicObject getInheritable() {
+    public HierarchicalDynamicObject getInheritable() {
         return new InheritedDynamicObject();
     }
 
     private DynamicObject snapshotInheritable() {
-        final List<DynamicObject> delegates = new ArrayList<>(4);
+        List<DynamicObject> delegates = new ArrayList<>();
```
alllex: 💭  I'd try to keep the performances/resources decision out of this refactoring PR. I imagine the fixed sizes where there for a reason. Maybe we know we'll create a ton of those objects. The default ArrayList constructor allocates an arrray of 10 elements, I believe. Let's keep the known cap for now.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -175,19 +164,155 @@ private DynamicObject snapshotInheritable() {
...
+            if (result.isFound()) {
+                return result;
+            }
+            parent = parent.getParent();
+        }
+        return DynamicInvokeResult.notFound();
+    }
+
+    /**
+     * Tries to invoke a method by name. If no method is found on the delegates or parents,
```
alllex: 💭  As it stands, this is a javadoc that explains the implementation of the method. It would be better to keep the implementation-related comments next to the implementation code itself. That's why I'd put the relevant part of this comment directly next to the code in the method, so it can help the readers of the implementation.

If you keep it in the javadoc, it's harder to know that there is a good explanation of what the code does that lives in a javadoc.

subprojects/core-api/src/main/java/org/gradle/internal/metaobject/HierarchicalDynamicObject.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2016 the original author or authors.
```
alllex: 💅 Year?

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -238,6 +363,7 @@ public DynamicInvokeResult tryGetProperty(String name) {
         }
 
         @Override
+        @Deprecated
```
alllex: ❓

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/metaobject/CompositeDynamicObject.java
```diff
@@ -15,30 +15,33 @@
...
  * Presents a {@link DynamicObject} view of multiple objects at once.
  *
  * Can be used to provide a dynamic view of an object with enhancements.
  */
-public abstract class CompositeDynamicObject extends AbstractDynamicObject {
+@NullMarked
+public class CompositeDynamicObject extends AbstractDynamicObject {
 
-    private static final DynamicObject[] NONE = new DynamicObject[0];
+    private final List<DynamicObject> objects;
```
alllex, line 34: ❌ Let's keep the performance/memory decisions out of this refactoring PR.

I imagine, we instantiate a ton of these objects and using more memory for each would result in waste. Since we don't have good tests for increased memory consumption, I would be conservative with changes in such an area.

Making it a list also increases GC pressure, because every for-loop below now has to allocate an iterator.

My recommendation would be to keep the field as an array of the size exactly required to keep the known number of elements. I'll leave it up to you to decide whether that should happen before the constructor call or inside the constructor

alllex (changes requested):

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectModelAccessTrackingParentDynamicObject.kt
alllex: I must've look at a stale diff. It's been reverted in the lastest state

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -175,19 +169,159 @@ private DynamicObject snapshotInheritable() {
...
+    @Override
+    public DynamicInvokeResult tryInvokeMethod(String name, @Nullable Object... arguments) {
+        // First, try to find an actual method on delegates or parents
+        DynamicInvokeResult result = doTryInvokeMethod(name, arguments);
+        if (result.isFound()) {
+            return result;
+        }
+
+        // No method found — fall back to looking for a property with the method name
+        // whose value can be invoked (Closure, NamedDomainObjectContainer, or callable)
```
alllex, line 279: This old code is so much more readable now with the good comments. Thanks 👏

reinsch82:
> I still think that we should keep the delegates in the CompositeDynamicObject as an array, rather than a List.

@alllex I think it is plausibly no difference in performance and it makes the code safer, no accidentally to small arrays anymore.

PR #37411, opened 2026-04-02
Ant: Update 1.10.15 to 1.10.16

cobexer:
- Update Apache Ant dependency from 1.10.15 to 1.10.16 https://dist.apache.org/repos/dist/dev/ant/RELEASE-NOTES-1.10.16.html

Changes from Ant 1.10.15 TO Ant 1.10.16
=======================================

Changes that could break older environments:
-------------------------------------------

 * <xslt> now uses the same logic to compare file timestamps when
   determining whether a target file is out-of-date with respect to
   the source file or stylesheet that most other tasks use. This means
   it will assume a default timestamp granularity that depends on the
   current operating system.
   A new granularity attribute allows you to override the assumed
   granularity.
   Under certain edge cases this means xslt will now not process files
   it would have processed before (when the timestamps of source or
   stylesheet are very close or even equal to the timestamp of the
   target). In this case you can set granularity to 0 to get back to
   the behavior of 1.10.15.
   Bugzilla Report 65756

 * the <mail> task as well as MailLogger will now check the server
   identity as specified by RFC 2595 when JavaMail is used in
   combination with TLS (plain TLS or StartTLS).
   The check can be disabled with a new MailLogger property
   MailLogger.tls.checkserveridentity or a new <mail> task attribute
   checkServerIdentity.
   Bugzilla Report 69416

 * Ant now recognizes Windows junctions and treats them just like
   symbolic links in general. With this builds on Windows that use
   followSymlinks="false" and rely on Ant following Windows junctions
   will break. In order to get the old behavior you need to set
   followSymlinks to true and exclude real symbolic links via the
   <symlink> selector.
   Bugzilla Report 66293

Fixed bugs:
-----------

 * <scp> now properly handles IPv6 addresses as hostnames.
   Bugzilla Report 59160

 * javac task has been fixed to generate the "-classpath" option
   only when there are any classpath elements present.
   Bugzilla Report 69586

 * URLResource#getName could strip the first character of a resource
   path even if it was not a file separator.
   Bugzilla Report 69680

 * ant.bat now exits with a non-zero exit code if ANT_HOME is not set.
   Github Pull Request apache/ant#222

 * A regression in ftp task caused some files to not be downloaded.
   That has been fixed now.
   Bugzilla Report 69485

Other changes:
--------------

 * Ant will no longer set a java.lang.SecurityManager at runtime
   if the "java.security.manager" system property is set to
   "disallow".
   Github Pull Request apache/ant#216

 * <mail> and MailLogger can now enforce the use of STARTLS rather
   than silently fall back to unencrypted authentication via a new
   MailLogger property and a new <mail> task attribute.
   Bugzilla Report 69416

 * added a Windows specific <mklink> task that can be used to create
   hard links, symbolic links and junctions.

 * added <delete link="..."> that can be used delete symbolic links or
   Windows junctions. For symbolic links this duplicates what
   <symlink action="delete" ...> does - it has been introduced to
   handle symlinks and junctions via a single API.

 * added <canCreateSymlink> condition that evaluates to true if the
   current Ant process can create symbolic links.

 * added <windowsjunction> file selector which only select directories
   that are Windows junctions.

 * added a new actOnLinkTargets attribute to <setpermissions> to
   control whether the permissions apply to a symbolic link or Windows
   junction or the target of the respective links. The old behavior
   of changing the link's target remains as default.

 * a new combination of attributes allows <sshexec> to mask sensitive
   data specified for the command line to execute.
   Github Pull Request apache/ant#124


cobexer:
@coderabbitai review

blindpirate (approved):

cobexer:
Superseded by #37770

PR #37414, opened 2026-04-02
Check for actual files in SubprojectsInfo

ghale:
hasDescendantDirWithFiles now returns false for directories that exist but contain only empty subdirectories, ensuring subproject metadata accurately reflects whether tests actually exist or not.

Without this, you can have an empty test directory (perhaps because all the tests were moved out) but the json generation marks it as having tests (because it only looks for the directory and not whether it contains anything).  However, because there are no files, git does not produce the directory on CI and the `checkSubprojectsInfo` task fails there because it thinks the subprojects info is out of date.  This can be confusing because running `generateSubprojectsInfo` locally doesn't actually change the json file, but it still fails on CI.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

PR #37415, opened 2026-04-02
Fix BuildOperationListener finished() callback not delivered when started() was missed

### LINKED ISSUE #37403: Callback behavior of `BuildOperationListener` changed in 9.5.0 RC1
### Current Behavior

The Develocity plugin complains about:
> buildScan.value() called after the build has finished.
The buildScan extension cannot be used after the build has finished.
This method call will be ignored.

on builds in gradle/gradle as of the second build on a daemon.

### Expected Behavior

The develocity plugin does not display warnings regardless of the number of builds executed previously on the same daemon.

### Context (optional)

Relevant commit range: https://github.com/gradle/gradle/compare/v9.5.0-M7...v9.5.0-RC1

in gradle/gradle check out https://github.com/gradle/gradle/pull/37351 aka https://github.com/gradle/gradle/commit/7f733954681fc53539d47909414fdfcffda81a4e
run any gradle build
(setup: ./gradlew --stop)
`./gradlew tasks` -> Build Scan published, no warnings
`./gradlew tasks` -> Build Scan published, warnings
`./gradlew --stop` -> daemon stopped
`./gradlew tasks` -> Build Scan published, no warnings

### Gradle version

9.5.0-RC1

### Build scan URL (optional)

_No response_

_No response_

big-guy:
Fixes #37403

ProgressShieldingBuildOperationListener was gating finished() delivery on whether started() had been observed, causing listeners registered after an operation started to never receive the finished() callback. This broke the Develocity plugin on subsequent daemon builds. Now finished() is always forwarded while progress() remains filtered.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/build-operations/src/main/java/org/gradle/internal/operations/DefaultBuildOperationListenerManager.java
+++ b/platforms/core-runtime/build-operations/src/main/java/org/gradle/internal/operations/DefaultBuildOperationListenerManager.java
@@ -89,7 +89,7 @@ public class DefaultBuildOperationListenerManager implements BuildOperationListe
      * for operations whose started notification was not delivered to this listener.
      *
      * Tracks active operations in a primitive long hash set to avoid per-operation
-     * object allocation. Only delivers progress/finished if started was seen.
+     * object allocation. Only delivers progress if started was seen.
      */
     private static class ProgressShieldingBuildOperationListener implements BuildOperationListener {
 


platforms/core-runtime/build-operations/src/main/java/org/gradle/internal/operations/DefaultBuildOperationListenerManager.java
```diff
@@ -119,9 +119,10 @@ public void progress(OperationIdentifier operationIdentifier, OperationProgressE
         @Override
         public void finished(BuildOperationDescriptor buildOperation, OperationFinishEvent finishEvent) {
```
jvandort, line 120: We should update the javadoc on this class to reflect the changed behavior

Issue #37091, opened 2026-03-12
`validatePlugins` fails with `NoClassDefFoundError` for implementation dependencies used only privately in superclass tasks


FinlayRJW:
### Current Behavior

when a task extends a task from another library, `validatePlugins` fails with a `NoClassDefFoundError` if the superclass task uses types from `implementation` dependencies — even when those types are only used in private fields and method bodies (i.e. not part of the public API).
```
> Task :consumer-plugin:validatePlugins FAILED
Execution failed for task ':consumer-plugin:validatePlugins'.
> There was a failure while executing work items
   > A failure occurred while executing org.gradle.plugin.devel.tasks.internal.ValidateAction
      > com/example/internal/InternalType
```

### Expected Behavior

`validatePlugins` should succeed. The `implementation` dependency types are private implementation details of the superclass — they are not part of the task's public API and should not need to be on the consumer's compile classpath for validation to pass.

### Context (optional)

This forces plugin authors to declare internal dependencies as api even when the types are only used privately, leaking `implementation` details to consumers. The workaround is to extract all usage of those types into a separate helper class in the same jar so that the task class bytecode itself never references them — but this is fragile and unintuitive.

The root cause appears to be that `ValidatePlugins` uses the `compileClasspath` ([see here](https://github.com/gradle/gradle/blob/b631911858264c0b6e4d6603d677ff5218766cee/platforms/extensibility/plugin-development/src/main/java/org/gradle/plugin/devel/plugins/JavaGradlePluginPlugin.java#L287)) to load task classes for inspection. Since implementation dependencies are not on the `compileClasspath`, the JVM cannot load the superclass. Using the `runtimeClasspath` instead would include implementation dependencies and resolve this.

### Self-contained Reproducer Project

https://github.com/FinlayRJW/validate-plugins-repro

### Gradle version

Reproduces on both 8.14.4 and 9.4.0

ov7a:
The issue is in the backlog of the relevant team and is prioritized by them.

h0tk3y:
Thanks! `validatePlugins` should indeed have access to the runtime classpath. However, we need to take special care about the compile-only dependencies, which might be needed for plugin validation as well.

PR #37419, opened 2026-04-03
Use runtimeClasspath + compileClasspath to validate plugins

h0tk3y:
Fixes:

* https://github.com/gradle/gradle/issues/37091

Use the `runtimeClasspath` to ensure that privately used 
`implementation` dependencies omitted from the `compileClasspath`
do not cause crashes in plugin validation when it tries to load
the plugin classes.

Keep `compileClasspath` on the plugin validation classpath because:
* It is common for plugins to depend on another plugin with a `compileOnly` dependency.
* It is a backup for corner cases in some builds, which would start failing with the switch to `runtimeClasspath`.


h0tk3y:
@bot-gradle test RFN

h0tk3y:
@bot-gradle test RFN

PR #37435, opened 2026-04-06
Avoid IO when estimating metadata fetching cost

jvandort:
In dependency resolution we try to estimate whether it is expensive to resolve a component's metadata, and if so, we resolve that component's metadata in parallel. Previously, this estimate itself could be expensive, triggering IO or waiting on project locks. This even happened if a component was cached locally on-disk but not yet loaded into memory.

We update implementations of metadata fetching to never touch IO or acquire shared locks, so that all IO and locking is performed in parallel.

We further update the confusing FAST/EXPENSIVE/CHEAP terminology to FAST/EXPENSIVE/MISSING, as CHEAP was easy to confuse with FAST, and in practice only represented scenarios where metadata was missing and known to be missing.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test apt

PR #37439, opened 2026-04-06
Do not decode metadata while holding cache lock

jvandort:
We performed work of encoding and decoding parsed metadata while holding the exclusive cache lock. While we likely cannot avoid holding the lock while reading/writing the backing file, we are able to encode and decode the associated data before and after the lock is released.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test apt

Issue #35795, opened 2025-11-24
Test against the XML with GenericTestExecutionResult


octylFractal:
### Issue type

Other

### Problem description

We lost a lot of XML coverage when moving to `GenericTestExecutionResult`. We should fix that.

big-guy:
We should consider replacing this with Open Test reporting and test against that instead of relying on JUnit XML test fixtures.

PR #37442, opened 2026-04-06
Replace GenericTestExecutionResult.TestFramework with TestPathSelector

octylFractal:
This is a first step towards #35795, by introducing TestPathSelector which allows a future XML result fixture to properly discard non-leaf nodes.

Any differences in test paths are instead resolved by adjusting the selector before passing it into the results fixture, currently only by specifying a different literal based on the test framework.

This also removes the entire concept of a "pre-normalized" path and supports escaping characters in the path.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/software/testing-base/src/testFixtures/groovy/org/gradle/api/internal/tasks/testing/report/generic/TestPathSelector.java
```diff
@@ -0,0 +1,391 @@
...
+ *         selecting tests from HTML and JUnit XML reports with the same code, as JUnit XML reports lack intermediate nodes.
+ *         For example, {@code :com:example:_} will match {@code :com:example}, but not {@code :com:example:MyTest}.
+ *         But it will not fail if an XML report does not contain a node for {@code :com:example}, but only {@code :com} and {@code :com:MyTest}.
+ * </ul>
+ *
+ * @param segments the segments of the selector, in order. Note that the leaf marker segment ({@code _}) is not included in this list, but is instead represented by the {@code furtherLeafNodes} flag
+ * @param furtherLeafNodes whether this selector matches further leaf nodes
+ * @param classMarkerIndex the 0-based index of the segment marked as the class, or {@code null} if no class marker is present
+ */
+public record TestPathSelector(ImmutableList<Segment> segments, boolean furtherLeafNodes, @Nullable Integer classMarkerIndex) {
```
tresat: 🤔 I don't love the idea of yet another regex-like string format control syntax to learn to write tests.

Why not a `TestPathSelector` interface, with multiple implementations (regex, non-leafs, etc) and/or static factory methods to create each?

I don't want to have to look up this syntax everytime I see special chars in the test path, or want to match against an odd test name.  I think it would be better to have the name of a factory method, and the arguments to it, or the type itself just explain what special conditions you're testing and how matching is performed.

tresat (commented):
I do like removing the need to state the framework explicitly after this change, it's much less code.  That said, I don't like calls to `maybeParenthesis` in tests.  Either there are or there aren't, why are we unsure?  `testCaseNodeHasParentheses` is also a slight downgrade, IMHO - we're now talking at a very low level of detail rather than stating what framework we're using.  

Switching most tests to just use `testPath()` and removing the concept of "pre-normalized" is great.  The format strings in `TestPathSelectorTest` are another "testing-specific DSL" that I really don't like (see other comment).

octylFractal:
`maybeParentheses` is an existing method, which delegates based on the test source configuration. I could rename it, but it's mostly orthogonal to the current PR. I can change `testCaseNodeHasParentheses` though, to just be `isJUnitJupiter`.

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/api/internal/tasks/testing/report/generic/TestPathSelector.java
octylFractal: I'm not really sure how that would work in the face of having non-leaves _and_ regex, or the class marker, which is something that already happens in some of the tests I tried to change.

If you really don't want the special syntax, I can make it so these are just built using helper methods, but it's quite wordy I think:
```java
// Before
result.testPath(':org.gradle.ASuite:org.gradle.OkTest:ok').onlyRoot().assertStdout(equalTo(""))
// After
result.testPath(
  literal('org.gradle.ASuite'),
  classMarker(literal('org.gradle.OkTest')),
  literal('ok')
).onlyRoot().assertStdout(equalTo(""))

// === or ===

// Before
result.testPath(':org.gradle.ASuite:org.gradle.OkTest').onlyRoot().assertStdout(equalTo("before OkTest class out\nafter OkTest class out\n"))
// After
result.testPath(
  literal('org.gradle.ASuite'),
  classMarker(literal('org.gradle.OkTest')),
  furtherLeafNodes()
).onlyRoot().assertStdout(equalTo("before OkTest class out\nafter OkTest class out\n"))
```

tresat: I think that would be much clearer, especially since we'll only use `testPath` by itself like 99% of the time.  Then we can easily find instances of tests using the other functionality and they would be very visibly marked when reading the code.

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/api/internal/tasks/testing/report/generic/TestPathSelector.java
```diff
@@ -0,0 +1,352 @@
...
+                seenClassMarker = true;
+            }
+        }
+    }
+
+    public static TestPathSelector of(Segment... segments) {
+        return new TestPathSelector(ImmutableList.copyOf(segments));
+    }
+
+    public sealed interface MatchResult extends Describable {
```
tresat, line 64: `record`s and sealed types, it's like reading code from the future... 😄

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/api/internal/tasks/testing/report/generic/TestPathSelector.java
```diff
@@ -0,0 +1,352 @@
...
+        enum Matched implements MatchResult {
+            INSTANCE;
```
tresat, line 66: ❓ Wouldn't empty `record`s be better here:                                                                     
```
  record Matched() implements MatchResult { ... }
  record NotAbsolute() implements MatchResult { ... }
```

You'd have to do `new MatchResult.Matched()`, or provide a factory method (and maybe keep a static `INSTANCE` field anyway, but since we're sealed the `this == Matched.INSTANCE` can be `this instanceof Matched`, and then we'd be able to do exhaustive matching in switches.

None of which are huge benefits, but I think it's kind of weird that some of the elements are `enum` and some are `record`.

I'm mostly just thinking out loud if this mixed enum + record pattern is a good one, and if there are any downsides to it or a uniform hierarchy (other than a little extra ceremony to make it uniform).  It's the first time I've seen it and it seems a little off to me, but I'm very unsure about that assessment.  Is this a typical pattern?

I much prefer the functionality of these factory methods to the interpreted special chars in strings.

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/api/internal/tasks/testing/report/generic/TestPathSelector.java
```diff
@@ -0,0 +1,352 @@
...
+            @Override
+            public String getDisplayName() {
+                return "path is not absolute";
+            }
+        }
+
+        record SegmentCountMismatch(int expected, int actual) implements MatchResult {
+            @Override
+            public String getDisplayName() {
+                return String.format("segment count mismatch (expected: %d, actual: %d)", expected, actual);
```
tresat, line 86: 🤔 These aren't really names.  Could we add a different method to the `MatchResult` interface and use that instead, instead of `Describable`?

octylFractal: Yeah, I wish `Describable` was really offering a `describe()` method instead. I can remove the interface and change to that.

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/api/internal/tasks/testing/report/generic/TestPathSelector.java
octylFractal: So, this is probably overkill here, but it's worth considering elsewhere -- if we do use an empty record, it still allocates each time, whereas the enum is a one-time cost. Personally I've just gotten used to writing it this way because I have had to deal with using the pattern in a lot of hot spots, so I opt for consistency in design rather than spending effort to wonder if I need to optimize or not. I find the enum is generally not too much harder to read.

re: exhaustive matching in switches, you can still do `instanceof Matched` even if it's an enum so there's no issue there. We unfortunately don't have pattern matching for switch yet (it was preview in 17) so this won't matter until we move to Java 21 base, whenever that is.

tresat: Cool, I didn't realize the exhaustive matching worked could work with enums like this.  Wouldn't the allocations be the same using a `private static final INSTANCE` with a class though?  They would be extra boilerplate code for nebulous benefit.  I'll try and convince myself that mixed-type sealed hierarchies like this are the new normal.

I wonder if we should opt-in to a preview feature like exhaustive matching that we now know (with the benefit of hindsight) will be stabilized in a later version?

tresat (approved):

platforms/software/testing-base/src/testFixtures/groovy/org/gradle/api/internal/tasks/testing/report/generic/TestPathSelector.java
octylFractal: Yes, it's the same as a `class` with an `INSTANCE` field. I'm unsure which is really preferable though. I suppose that the `class` communicates intent better as it's not really an enumeration that the `INSTANCE` represents.

Opting-in would maybe be nice in test code, but I'm unsure how well the preview bytecode version would play with Groovy and Kotlin. Opting-in in production code is right out, as we need to support different versions of Java for the daemon and preview features are not cross-version compatible.

octylFractal:
@bot-gradle test this

PR #37466, opened 2026-04-08
Reject bad Property type arguments upon store with understandable message

### LINKED ISSUE #19122: Configuration cache should not allow caching with a task using Property<Configuration> as input
### Expected Behavior
when applying `gradle-xjc-plugin` in your build, you should receive an error when feeding the configuration cache due to a reference to the Configuration object:
https://github.com/unbroken-dome/gradle-xjc-plugin/blob/82c1d69e0de91d11a61ae6153f89491adafa3a84/src/main/kotlin/org/unbrokendome/gradle/plugins/xjc/XjcGenerate.kt#L114

1. Feed the configuration cache
`./gradlew processResources --configuration-cache`
=> Report an error thant configuration cache can't be fed

### Current Behavior
1. Feed the configuration cache
`./gradlew processResources --configuration-cache`

_Calculating task graph as no configuration cache is available for tasks: processResources
0 problems were found storing the configuration cache.
BUILD SUCCESSFUL_

2. Retrieve configuration from the cache
`./gradlew processResources --configuration-cache`

_Reusing configuration cache.

* What went wrong:
Could not load the value of field `__catalogResolutionClasspath__` of task `:instrumentation:jaxws:jaxws-2.0-common-testing:xjcGenerate` of type `org.unbrokendome.gradle.plugins.xjc.XjcGenerate`.
> Cannot set the value of a property of type org.gradle.api.artifacts.Configuration using a provider of type org.gradle.api.internal.file.DefaultFileCollectionFactory$ResolvingFileCollection._

### Context
Currently trying to setup configuration cache

### Steps to Reproduce 
https://github.com/open-telemetry/opentelemetry-java-instrumentation can be used to demonstrate the issue

### Your Environment
❯ ./gradlew -v

Gradle 7.4-20211122230135+0000

Build time:   2021-11-22 23:01:35 UTC
Revision:     ae3758aa7698196c03e7290116ea4b13388986c0

Kotlin:       1.5.31
Groovy:       3.0.9
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          11.0.12 (Eclipse Foundation 11.0.12+7)
OS:           Mac OS X 11.5 x86_64

Build scan URL:
https://scans.gradle.com/s/ckprgsj5duckm

tresat:
Using types like Configuration or SourceDirectorySet as a value type
for a Property causes confusing type-mismatch errors during cache
loading. This change introduces a check during serialization to fail
fast with a clear diagnostic message and resolution suggestion.

Fixes https://github.com/gradle/gradle/issues/19122.

### DIFF
--- a/platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/UnsupportedPropertyValueTypes.kt
+++ b/platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/UnsupportedPropertyValueTypes.kt
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.serialize.beans.services
+
+import org.gradle.api.artifacts.Configuration
+import org.gradle.api.file.SourceDirectorySet
+import org.gradle.api.provider.MapProperty
+
+/**
+ * Types that cannot be used as the value type `T` of a `Property<T>`
+ * in the configuration cache.
+ *
+ * A `Property<T>` where `T` matches one of these types (or a subtype)
+ * cannot survive a serialization round-trip: the codec that handles
+ * the resolved value produces a different type on deserialization
+ * than `Property<T>` expects, causing a confusing type-mismatch error
+ * at load time.
+ *
+ * Used by property codecs to fail fast at store time with a clear
+ * diagnostic message.
+ */
+object UnsupportedPropertyValueTypes {
+    private val unsupported: Map<Class<*>, String> = mapOf(
+        Configuration::class.java to "Use a @InputFiles ConfigurableFileCollection instead.",
+        SourceDirectorySet::class.java to "Use a @InputFiles ConfigurableFileCollection instead.",
+    )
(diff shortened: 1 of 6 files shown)


tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
Smoke tests for the Shadow plugin are failing.

The Shadow plugin uses the exact pattern we're trying to ban here: https://github.com/GradleUp/shadow/blob/0ad3f42b3e984ee1fadc71aba405ff83e0181483/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt#L140 and is already marked CC-compatible https://plugins.gradle.org/plugin/com.gradleup.shadow.

Looking into how easy it would be to change their usage.

tresat:
@bot-gradle test this

tresat:
One other thought - we could memoize the lookup in `UnsupportedPropertyValueTypes#check` if performance is a concern.  I just didn't want to make this more complex than necessary to begin with.

platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/UnsupportedPropertyValueTypes.kt
```diff
@@ -0,0 +1,72 @@
...
+ * the resolved value produces a different type on deserialization
+ * than `Property<T>` expects, causing a confusing type-mismatch error
+ * at load time.
+ *
+ * Used by property codecs to fail fast at store time with a clear
+ * diagnostic message.
+ */
+object UnsupportedPropertyValueTypes {
+    private val unsupported: Map<Class<*>, String> = mapOf(
+        Configuration::class.java to "Use a @InputFiles ConfigurableFileCollection instead.",
```
hfhbd, line 38: Instead of using a hardcoded short error message, what about adding a link to the user guide that contains more information about unsupported types and possible replacements, aka "don't do this, do this"?

Or does the build failure already contain such a link?

tresat: I don't think there is any section specifically addressing this Configuration -> FileCollection serialization behavior.  Perhaps we should add one.

Issue #37339, opened 2026-03-30
Remove InternalProblemReporter.reportError() and InternalProblemBuilder.internalSeverity()


donat:
### Issue type

Refactoring

### Problem description

`InternalProblemReporter.reportError()` and `InternalProblemBuilder.internalSeverity()` were introduced as temporary escape hatches in the severity refactoring (https://github.com/gradle/gradle/pull/37273) where severity is now determined by the reporting context rather than set explicitly via `ProblemSpec.severity()`. 

These methods exist for call sites where the problem is reported separately from the exception that fails the build. Removing them   requires reworking those call sites so that the exception and the problem report flow through the same throwing() path, which in turn requires updating build failure rendering and complex test fixtures.

PR #37467, opened 2026-04-08
Migrate all version catalog errors to the Problems API 

donat:
- Route every version catalog throw through `ProblemReporter.throwing()`, 
  - Some codepaths still relied on throwing `InvalidUserDataException`.
  - Others made use of the temporal `InternalProblemReporter.reportError()` API 
- Drop the legacy catalog throw helpers and the rendering helpers.
- Drop the helpers that reconstructed rendered failure strings; assert rendered failures as-is in tests.
- Strip trailing periods from contextual labels for consistency with the rest of the codebase.

Partially addresses https://github.com/gradle/gradle/issues/37339

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


donat:
@bot-gradle test QFL

donat:
@bot-gradle test QFL

donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test this

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/catalog/TomlDependenciesExtensionIntegrationTest.groovy
```diff
@@ -705,15 +705,17 @@ my-other-lib = {group = "org.gradle.test", name="lib2", version.ref="rich"}
...
-        })
+        verifyContains(failure.error, """
+> Import of external catalog file failed
+    In version catalog libs, import of external catalog file failed
+      File '${path.absolutePath}' doesn't exist
+      ${versionCatalogDocsFor("catalog_file_does_not_exist")}
+      Possible solution: Make sure that the catalog file '${path.name}' exists before importing it.""")
 
         and:
         verifyAll(receivedProblem) {
```
donat, line 707: Not sure if we both need `verifyContains` (above) and `verifyAll` (problem instance assertion). Should be resolved in the review.

subprojects/core/src/main/java/org/gradle/api/internal/catalog/problems/DefaultCatalogProblemBuilder.java
```diff
@@ -15,77 +15,21 @@
...
-    }
-
-    public static String getProblemString(ProblemInternal problem) {
-        ProblemDefinition definition = problem.getDefinition();
-        String contextualLabel = problem.getContextualLabel();
-        String renderedLabel = contextualLabel == null ? definition.getId().getDisplayName() : contextualLabel;
-        return getProblemString(renderedLabel, problem.getDetails(), problem.getSolutions(), definition.getDocumentationLink());
-    }
-
-    public static String getProblemString(String label, String details, List<String> solutions, DocLink documentationLink) {
```
donat, line 65: The rendering is handled by `DefaultProblemWriter`.

subprojects/core/src/main/java/org/gradle/api/internal/catalog/problems/VersionCatalogProblemId.java
```diff
@@ -25,21 +25,23 @@ public enum VersionCatalogProblemId {
 
     ACCESSOR_NAME_CLASH("Accessor name clash"),
     CATALOG_FILE_DOES_NOT_EXIST("Import of external catalog file failed"),
-    INVALID_ALIAS_NOTATION("The alias notation is invalid"),
```
donat, line 28: Some Id displaynames are renamed for consistency.

subprojects/core/src/testFixtures/groovy/org/gradle/api/internal/catalog/problems/VersionCatalogErrorMessages.groovy
```diff
@@ -38,534 +41,4 @@ trait VersionCatalogErrorMessages {
         String expected = normaliseLineSeparators(rawExpected).replaceAll("(?m)^( +)\n", "\n").trim()
         assert actual.contains(expected)
     }
-
-    String nameClash(@DelegatesTo(value = NameClash, strategy = Closure.DELEGATE_FIRST) Closure<?> spec) {
```
donat, line 42: All string-building from the tests are now gone 🎉

reinsch82: I'm not sure we gain a lot here. We have more places to fix now, when we change the strings.

donat: There are a few reasons for the change
- There's complex logic in the production and in the test code that are practically doing the same thing. Any change on one side requires the other one in sync. This cumbersome to maintain in the long term and this is something that we wanted to get rid of for a long time.
- The new assertions make fixing failing tests very easy. It will be evident what is the expected and the actual string, especially with a help of an AI agent.
- There's a related comment on this PR. We may not even want to do assertions on the rendered strings at all. One could argue that the rendering is tested in a generic way and here the important aspect is to verify whether the content (ie the fields in the problem instance) is correct. I would like to discuss this here, as part of the review process.

reinsch82: > There's complex logic in the production and in the test code that are practically doing the same thing. Any change on one side requires the other one in sync. This cumbersome to maintain in the long term and this is something that we wanted to get rid of for a long time.

I'm all in on not having cumbersome logic twice, I wouldn't want to solve it by inlining one of the duplications though. I'd rather use the same constants and logic in both places, thereby we are testing the exact output and can test the logic separately.

> The new assertions make fixing failing tests very easy. It will be evident what is the expected and the actual string, especially with a help of an AI agent.

This doesn't ring true to me. For me this translates to "We can introduce more duplication, because AI can do repetitive tasks." Assuming that the same standards still apply in an AI dominant environment, we'd still optimize for less duplication.

> There's a related comment on this PR. We may not even want to do assertions on the rendered strings at all. One could argue that the rendering is tested in a generic way and here the important aspect is to verify whether the content (ie the fields in the problem instance) is correct. I would like to discuss this here, as part of the review process.

This would at least reduce the cumbersome part of changing the same string in multiple places when changing it just once in the production code. I can relate to that. It requires good coverage of the string building parts. That shouldn't be a big issue with AI. It's an "incomplete" end-to-end test though in some way.

subprojects/core/src/main/java/org/gradle/api/internal/catalog/problems/VersionCatalogProblemId.java
```diff
@@ -25,21 +25,23 @@ public enum VersionCatalogProblemId {
...
-    INVALID_DEPENDENCY_NOTATION("The dependency notation is not valid"),
-    INVALID_PLUGIN_NOTATION("The plugin notation is not valid"),
-    INVALID_MODULE_NOTATION("The module notation is not valid"),
+    INVALID_ALIAS_NOTATION("Invalid alias notation"),
+    RESERVED_ALIAS_NAME("Reserved alias name"),
+    INVALID_DEPENDENCY_NOTATION("Invalid dependency notation"),
+    INVALID_PLUGIN_NOTATION("Invalid plugin notation"),
+    INVALID_MODULE_NOTATION("Invalid module notation"),
+    INVALID_TOML_DEFINITION("Invalid TOML definition"),
+    INVALID_VERSION_NOTATION("Invalid version notation"),
```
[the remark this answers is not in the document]
donat: This comment is correct. In fact, the links have been broken before. I've pushed a change to add ascii anchors to all sections matching with the corresponding problem ids.

platforms/documentation/docs/src/docs/userguide/how-to/dependency-management/how_to_fix_version_catalog_problems.adoc
```diff
@@ -19,6 +19,7 @@ Version catalogs help manage dependencies efficiently, but sometimes errors occu
 
 This guide provides quick solutions to common version catalog issues in Gradle.
 
+[[accessor_name_clash]]
```
reinsch82, line 22: 🎉

subprojects/core/src/testFixtures/groovy/org/gradle/api/internal/catalog/problems/VersionCatalogErrorMessages.groovy
donat: I tend to agree. This is my personal bias that I like simple-to-understand tests at the expense of verbosity and duplication. But I get your points. I'm leaning towards deleting the rendered strings from test assertions and only validating the content. There's a reasonable amount of test coverage for generic problem rendering, which provides enough justification.

donat: @reinsch82 fixed; ptal

donat:
@bot-gradle test  this

reinsch82 (approved):

PR #37471, opened 2026-04-08
Extract domain-object-collections from core

ghale:
Core and core-api are large monoliths that contain api and implementation classes that cross multiple platforms. Having these classes lumped together encourages tangling and circular relationships between platforms and can lead to relationships that do not comply with our platform rules. It also leads to extremely coarse dependencies that drag in much more than is necessary and can cause classes to be misplaced solely in order to workaround the limitations of these coarse relationships. Breaking up core/core-api into smaller projects helps to prevent tangling and allows for narrower, more granular dependency relationships. It also clears the path for us to start enforcing platform relationships in a way that we can't easily do so today.

Extract the domain object collection infrastructure from subprojects/core into a new platforms/core-configuration/ domain-object-collections module. This includes the DefaultDomainObjectCollection hierarchy, element sources, collection event system, and related service registrations.

Production classes moved:
- org.gradle.api.internal.collections package 
- org.gradle.api.internal domain object collection hierarchy classes
- NameValidator moved to base-services

Service registrations for CollectionCallbackActionDecorator and DomainObjectCollectionFactory moved from core scope services to a new DomainObjectCollectionServices class.

Test classes split between domain-object-collections and core (TaskContainer test classes remain), with abstract integration test bases shared via testFixtures.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

big-guy (approved):

PR #37479, opened 2026-04-09
Extend CC instrumentation to cover more cases of directory content walking

### LINKED ISSUE #33317: The CC instrumentation misses some cases when using Kotlin `walk()`
### Current Behavior

The instrumentation is missing some cases where the buildscripts uses CC input. 

For an example, with the following build script:

```kotlin
rootDir.walk().forEach {
    if (it.name == "doit") {
        tasks.register("doit") {
            doLast {
                println("Doing the thing!")
            }
        }
    }
}
```

The following is happening:

```
$ touch doit

# initial run
$ ./gradlew doit
Calculating task graph as no cached configuration is available for tasks: doit

> Task :doit
Doing the thing!

# changing the filesystem here
$ rm doit 

# CC did not detect it
$ ./gradlew doit
Reusing configuration cache.

> Task :doit
Doing the thing!

# A failure was expected instead
$ ./gradlew doit --no-configuration-cache

[Incubating] Problems report is available at: file:///Users/martinbonnin/git/test-cc/build/reports/problems/problems-report.html

* What went wrong:
Task 'doit' not found in root project 'test-cc'. Some candidates are: 'init'. 
```

I believe this might be because `kotlin-stdlib` is not being instrumented maybe? I haven't been able to reproduce without usage of `walk()` up there. 

### Expected Behavior

The configuration cache gets invalidated when I change the filesystem.

### Context (optional)

_No response_

See above

### Gradle version

8.13

### Build scan URL (optional)

_No response_

_No response_

jbartok:
Fixes:
* https://github.com/gradle/gradle/issues/33317

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredBuildInputsDynamicGroovyBuildScriptIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredBuildInputsDynamicGroovyBuildScriptIntegrationTest.groovy
@@ -16,6 +16,8 @@
 
 package org.gradle.internal.cc.impl.inputs.undeclared
 
+import spock.lang.Issue
+
 class UndeclaredBuildInputsDynamicGroovyBuildScriptIntegrationTest extends AbstractUndeclaredBuildInputsIntegrationTest implements GroovyPluginImplementation {
     @Override
     String getLocation() {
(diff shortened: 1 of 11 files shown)


platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredBuildInputsDynamicGroovyBuildScriptIntegrationTest.groovy
```diff
@@ -26,4 +28,98 @@ class UndeclaredBuildInputsDynamicGroovyBuildScriptIntegrationTest extends Abstr
     void buildLogicApplication(BuildInputRead read) {
         groovyDsl(buildFile, read)
     }
+
+    @Issue("https://github.com/gradle/gradle/issues/33317")
+    def "invalidates configuration cache when a deeply nested file observed via recursive NIO #apiName is removed"() {
```
tresat: ❓ Would it be better to add these methods to the abstract base class, so they are tested in all types of build logic?

At least the java NIO ones, the Kotlin-specific ones like `walk()` are in the proper place.

I see there are some slight syntax differences in `"invalidates configuration cache when a deeply nested file observed via recursive NIO #apiName is removed"` and `"invalidates configuration cache when a new deeply nested file is added under a directory walked via recursive NIO #apiName"`.  But I _think_ they are close enough that with an abstract method to handle the difference in `treeRootPath` declarations, the other syntax differences could be removed by using a large value instead of `MAX_VALUE` and avoiding `use` or `withCloseable` - which aren't really vital to this example anyway.

subprojects/core/src/main/java/org/gradle/internal/classpath/FileUtils.java
```diff
@@ -62,17 +62,25 @@ public static void tryReportFileOpened(Path path, String consumer) {
...
         File file = toFileIfAvailable(path);
         if (file != null) {
-            Instrumented.directoryContentObserved(file, consumer);
+            tryReportDirectoryContentObserved(file, consumer);
         }
     }
 
+    public static void tryReportDirectoryContentObserved(File file, String consumer) {
+        Instrumented.directoryContentObserved(file, consumer);
+    }
```
tresat: 💅 The changes in this file don't seem worthwhile.  Why not just have `recordWalkObservations` call the static methods on `Instrumented` directly?

platforms/core-runtime/classpath/src/main/java/org/gradle/internal/classpath/declarations/KotlinStdlibFileInterceptors.java
```diff
@@ -124,6 +130,59 @@ public static Object intercept_useLines(
...
+        recordWalkObservations(walk, consumer);
+        return walk;
+    }
+
+    /**
+     * Eagerly traverses a {@link FileTreeWalk} to register the filesystem entries it observes
+     * as configuration cache inputs. Returning a new iterator from {@link FileTreeWalk#iterator()}
+     * later still works correctly; the caller pays a one-time traversal cost at configuration time.
+     */
+    private static void recordWalkObservations(FileTreeWalk walk, String consumer) {
```
tresat, line 182: ❗ There is a race condition here, but I don't know how relevant it is.  Consider:
1. We call `intercept_walkBottomUp` 
2. We call `FileTreeWalk walk = FilesKt.walkBottomUp(self);` - do something with the files that exist
3. An external file system change occurs and a file is deleted or added
4. We perform `recordWalkObservations(walk, consumer);` - this doesn't see the change in 3 = tracked file info is incorrect

I think to be correct we'd have to:
1. We call `intercept_walkBottomUp` 
2. We perform `recordWalkObservations(walk, consumer);` and store the resulting files info in a local
3. We call `FileTreeWalk walk = FilesKt.walkBottomUp(self);` - do something with the files that exist
4. An external file system change occurs and a file is deleted or added
5. We perform `recordWalkObservations(walk, consumer);` again
6. We compare the result of 2 and 5 - if they're the same we're good, if they differ then we know something fishy is happening.  

I guess the safest thing at this point would be to not store the cache entry.  But that would fail the build with store-load. I wonder if there's a way we could use the cache entry once for this build and then immediately mark it invalid so it isn't reused.  Or maybe even loading it once isn't safe enough?

subprojects/core/src/main/java/org/gradle/internal/classpath/declarations/NioFileInterceptors.java
```diff
@@ -368,4 +376,222 @@ public static FileChannel intercept_open(
...
+        @Override
+        public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+            tryReportFileSystemEntryObserved(file, consumer);
+            return delegate.visitFile(file, attrs);
+        }
+
+        @Override
+        public FileVisitResult visitFileFailed(Path file, IOException exc) throws IOException {
+            tryReportFileSystemEntryObserved(file, consumer);
+            return delegate.visitFileFailed(file, exc);
```
tresat: ❓ If visiting a file failed, should it still be considered "observed" here?  It can't have been used for configuration purposes by Gradle if this happens.

It's suspicious that the other visitors below throw the exception while this doesn't.

I wonder, since failures are likely to be transient (files deleted/modified during visiting, file permissions, I/O errors) and might succeed immediately on the next try, maybe it would be safest upon hitting a failure to just fail the build?  Or at least invalidate the cache entry immediately after loading it for this build.

tresat (changes requested):

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredBuildInputsDynamicGroovyBuildScriptIntegrationTest.groovy
jbartok: Ohh, this turned out to be a huge rabbit hole. I've made a ton of changes due to this observation, and it makes much more sense now. Thanks!

subprojects/core/src/main/java/org/gradle/internal/classpath/FileUtils.java
jbartok: Fixed.

platforms/core-runtime/classpath/src/main/java/org/gradle/internal/classpath/declarations/KotlinStdlibFileInterceptors.java
jbartok: Not sure what the best course of action is here; for now, I've just added a comment acknowledging the problem.

subprojects/core/src/main/java/org/gradle/internal/classpath/declarations/NioFileInterceptors.java
jbartok: You are right, addressed.

platforms/core-runtime/classpath/src/main/java/org/gradle/internal/classpath/declarations/KotlinStdlibFileInterceptors.java
tresat: @mlopatkin - Any thoughts on this?

platforms/core-runtime/classpath/src/main/java/org/gradle/internal/classpath/declarations/KotlinStdlibFileInterceptors.java
```diff
@@ -124,6 +130,66 @@ public static Object intercept_useLines(
             : invokeKotlinStaticDefault(USE_LINES_DEFAULT, mask, self, charset, block);
     }
 
+    @InterceptCalls
+    @StaticMethod(ofClass = FilesKt.class)
```
bamboo, line 134: 🤔 Intercepting at this level doesn't feel right. I wonder if we couldn't intercept the underlying `java.io.File#listFiles()` calls these library functions are based on somehow.

PR #37482, opened 2026-04-09
Remove ProblemsProgressEventEmitterHolder static holder

hegyibalint:
## Summary
- Replaces the `ProblemsProgressEventEmitterHolder` static service locator with per-job binding of `InternalProblems` via `ServiceRegistry`
- `WorkerDaemonFactory` lazily resolves `InternalProblems` from the project-scoped `ServiceRegistry` at job-execution time and binds it to the worker daemon client
- If no problems service is available, problem reports from workers are silently dropped
- Follows the same `ServiceRegistry` injection pattern used by `NoIsolationWorkerFactory` and `IsolatedClassloaderWorkerFactory`


hegyibalint:
@bot-gradle test this

hegyibalint:
@bot-gradle test this

platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonFactory.java
```diff
@@ -55,6 +59,9 @@ public DefaultWorkResult execute(IsolatedParametersActionExecutionSpec<?> spec,
                 // wrap in build operation for logging startup failures
                 final WorkerDaemonClient client = CurrentBuildOperationRef.instance().with(parentBuildOperation, this::reserveClient);
                 try {
+                    // Lazily resolve the build-tree scoped ProblemsInternal service for this job.
+                    ProblemsInternal problems = (ProblemsInternal) serviceRegistry.find(ProblemsInternal.class);
+                    client.bindProblems(problems);
```
eskatos: ❓ What are the situations where this will be null?

donat (changes requested):
I have the same question as Paul: What's a null `Problems` service? My understanding is that by the time any worker can run, the service should always be available. Failing a build may be too much in this situation, but we should warn the user about potentially missing reports.

platforms/core-execution/worker-process-services/src/main/java/org/gradle/process/internal/worker/DefaultWorkerProblemProtocol.java
```diff
@@ -31,9 +31,20 @@
 @NullMarked
 public class DefaultWorkerProblemProtocol implements WorkerProblemProtocol {
 
+    private @Nullable ProblemsInternal problems;
+
+    /**
+     * Binds the problems service to use for reporting problems received from the worker.
+     * If {@code null}, problems will be silently dropped.
+     */
+    public void bindProblems(@Nullable ProblemsInternal problems) {
```
reinsch82: ❓ I think we could make this part of the `WorkerProblemProtocol` interface.

platforms/core-execution/worker-process-services/src/main/java/org/gradle/process/internal/worker/DefaultWorkerProblemProtocol.java
```diff
@@ -31,9 +31,20 @@
 @NullMarked
 public class DefaultWorkerProblemProtocol implements WorkerProblemProtocol {
 
+    private @Nullable ProblemsInternal problems;
```
reinsch82: ❌ since this is used in different threads, it's probably a good idea.
```suggestion
    private volatile @Nullable ProblemsInternal problems;
```

platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonFactory.java
reinsch82: ❌ this needs to be cleared out in `finally`, otherwise I think `problems` is kept in memory longer than needed. `WorkerDaemonFactory` is a longer lasting scope than `ProblemsInternal`.

reinsch82 (changes requested):
> If no problems service is available, problem reports from workers are silently dropped

❌ I think we should at least log something. Is logging available there?

platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonFactory.java
```diff
@@ -55,6 +59,9 @@ public DefaultWorkResult execute(IsolatedParametersActionExecutionSpec<?> spec,
                 // wrap in build operation for logging startup failures
                 final WorkerDaemonClient client = CurrentBuildOperationRef.instance().with(parentBuildOperation, this::reserveClient);
                 try {
+                    // Lazily resolve the build-tree scoped ProblemsInternal service for this job.
+                    ProblemsInternal problems = (ProblemsInternal) serviceRegistry.find(ProblemsInternal.class);
```
reinsch82: 🤔 maybe use `org.gradle.internal.lazy.Lazy` here?

platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonFactory.java
hegyibalint: Supported, and intentional: `null` means we didn't found a service. Otherwise, we would keep an old `InternalProblems` service around, and we don't want that.

See how we handle it on the other side here: https://github.com/gradle/gradle/pull/37482/changes#diff-09545880c8a92c7e012c46ee2d83f3427e7a2c2c8eea555628fc9703e47cbda7R46-R48

platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonFactory.java
hegyibalint: That would exactly be the opposite what we want to achieve here: we want to always get _the actually registered_ problem instance.

platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonFactory.java
hegyibalint: However, based on Reinhold's [idea below](https://github.com/gradle/gradle/pull/37482/changes#r3287041191) I think we can have an `if (problems != null`) gate here, and always reset in the `finally` block to none.

hegyibalint: Good catch! That would indeed keep a Problems service until the worker is destroyed, or a new work unit is dispatched.

platforms/core-execution/worker-process-services/src/main/java/org/gradle/process/internal/worker/DefaultWorkerProblemProtocol.java
hegyibalint: Good point, let me do that!

platforms/core-execution/worker-process-services/src/main/java/org/gradle/process/internal/worker/DefaultWorkerProblemProtocol.java
hegyibalint: Addressed, I've added:
 - volatile to the field on the class
 - plus a local variable capturing the reference of this volatile problems, so it stays the same under the problem reporting.

platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonFactory.java
eskatos: What are the situations where we wouldn't have the service available?

platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonFactory.java
hegyibalint: Hmm, I might be actually wrong about this: this factory is project scoped, depending on `Problems`, that is build-tree. The reference for this class should be the same under it's compelete lifetime. We should just get the service here, and store it.

The whole point is that the `WorkerDaemonClient` itself might be one outliving this instance.

Let me simplify this approach.

platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonFactory.java
eskatos: If we know there is no such situation, then we should throw instead of ignoring problems. Ignoring problems is a red flag.

eskatos: Ah, I see now that you addressed that in [1ccb0f1](https://github.com/gradle/gradle/pull/37482/commits/1ccb0f1c78de4d9b76cbb6dbdbe7064b8fda81e8)

platforms/core-execution/worker-process-services/src/main/java/org/gradle/process/internal/worker/DefaultWorkerProblemProtocol.java
```diff
@@ -31,9 +31,24 @@
...
+        this.problems = null;
+    }
+
     @Override
     public void reportProblem(Problem problem, OperationIdentifier id) {
-        ProblemReporterInternal problemsService = ProblemsProgressEventEmitterHolder.get().getInternalReporter();
-        problemsService.report(problem, id);
+        // Read once into a local to avoid a NPE if another thread clears the binding between the null check and the dereference.
+        ProblemsInternal p = problems;
+        if (p != null) {
```
eskatos: ❓ Shouldn't we fail if `problems` is null at this stage? What would be a situation where it is null?

platforms/core-execution/worker-process-services/src/main/java/org/gradle/process/internal/worker/request/Receiver.java
```diff
@@ -50,14 +50,24 @@ public class Receiver implements ResponseProtocol, StreamCompletion, StreamFailu
 
     // Sub-handlers for the different protocols implemented by ResponseProtocol
     private final WorkerLoggingProtocol loggingProtocol;
-    private final WorkerProblemProtocol problemProtocol;
+    private final DefaultWorkerProblemProtocol problemProtocol;
```
eskatos: ❌  Now that the methods have been moved to `WorkerProblemProtocol` this is an unnecessary change

eskatos (changes requested):
This is looking good.

There still is a lenient handling of the availability of the problems service that we should clear out before merging.

platforms/core-execution/worker-process-services/src/main/java/org/gradle/process/internal/worker/request/Receiver.java
hegyibalint: Good point, thanks for noticing!

platforms/core-execution/worker-process-services/src/main/java/org/gradle/process/internal/worker/DefaultWorkerProblemProtocol.java
hegyibalint: That is a good question: as I've pointed out, we should not have a case where the worker is `null`, and it technically points to an invalid configuration.

I think I have enough confidence that this should not happen, and supported by CI we should make quite certain that we should have no `null` here.

platforms/core-execution/worker-process-services/src/main/java/org/gradle/process/internal/worker/WorkerProblemDispatcher.java
```diff
@@ -17,23 +17,46 @@
...
+public class DefaultWorkerProblemProtocol implements WorkerProblemProtocol, WorkerProblemServiceManager {
+
+    private volatile @Nullable ProblemsInternal problems;
+
+    @Override
+    public void bindProblemsService(ProblemsInternal problems) {
+        Objects.requireNonNull(problems, "Problems service must not be null");
+        if (this.problems != null) {
+            throw new IllegalStateException("A problems service is already bound; previous worker run did not clear it");
+        }
```
hegyibalint, line 46: ℹ️ I've implemented a much tighter check on the service incoming, and on the current state.
Anything else than this is a serious bug.

eskatos (approved):
LGTM

reinsch82 (approved):

donat (approved):

Issue #25328, opened 2023-06-07
Update Project Isolation documentation page


bamboo:
Add information about our current plans, known caveats, etc.

ljacomet:
Moved to 8.10 RC1.

If this was delivered with 8.9, please close and update the milestone.

alllex:
We need to make sure the updated page contains the words "project isolation" verbatim.

Even though we decided to go with the "Isolated Projects" name for the feature, folks outside Gradle who are aware of it keep calling it "project isolation". To me this is a signal we'll never get away from people referring to the feature like this.

We did discuss that "project isolation" is something the users will do to make their build IP compatible. So it can fit naturally in one of the first paragraphs on the page.

FWIW, search engines today show the right result for googling "gradle project isolation", but I'd like us to have the alternative spelling in clear text.

For the record, our own in-docs search shows nothing relevant with the "project isolation" lookup


![Image](https://github.com/user-attachments/assets/048e10d3-8a5b-4b44-8787-977dcd45ebcf)

Issue #36940, opened 2026-03-04
Lack of clarity on Isolated Projects status


smplio:
### Issue type

Missing information

### Problem description

Isolated Projects criticaly lacks publicly available status. 
Issue on a roadmap board is dead: https://github.com/gradle/build-tool-roadmap/issues/76
Issue about Incubating IP keeps changing year in the description without any visible progress: https://github.com/gradle/build-tool-roadmap/issues/76
IP in docs is still described as pre-alpha feature since it's initial release

It's IMO the most anticipated feature for large projects, but it seems like there is no progress done for last two years. Could you, please, bring at least some clarity to it's development. Or is it dead on arrival?

### Context (optional)

Yes, I know it's not a documentation issue per se, but it was the closest topic

### Page with the problem

https://docs.gradle.org/current/userguide/isolated_projects.html

dsvoronin:
Totally agree. 
I understand that this project involves multiple big projects and companies to do their part to make this effective to the end user, but we can't even understand the status in the Gradle part. 😕 

Community started to gather pieces of information, like this project: https://agp-status.frybits.com/gradle-project-isolation/ maybe it's a good idea to review it and contribute on the Gradle status? Shouldn't take long

ljacomet:
This issue needs a decision from the team responsible for that area. They have been informed. Response time may vary.

***

The project is very much alive but indeed we are missing updates in the roadmap items at least.

mlopatkin:
The issue is in the backlog of the relevant team and is prioritized by them.

***

We'll look into updating the docs in 9.5.0.

alllex:
We've made steps to address this request, though not fully.

As Louis points out, the initiative is indeed actively being worked on. However, there are still challenges with preparing Gradle internals to handle parallelism and caching in the new parts of project lifecycles. This delays changes directly visible to users.

Having said that, even in the current state, the feature is already showing concrete benefits and a lot of further potential. About two months ago, the Gradle team adopted Isolated Projects as a daily driver for local builds of the Gradle repo. We are seeing ~2x speedup of IDE sync times and general improvements for task-running workflows. All of this comes from the parallel project configuration alone. We will probably share a more detailed blog post about this in the near future.

We are also hearing from early adopters who have successfully onboarded and are also seeing significant performance improvements.

In Gradle 9.5.0, we reworked the userguide for Isolated Projects, adding more clarity to what can be expected:
* https://docs.gradle.org/9.5.0/userguide/isolated_projects.html

[...]

PR #37491, opened 2026-04-09
Refresh Isolated Projects docs

alllex:
Fixes https://github.com/gradle/gradle/issues/25328
Related: https://github.com/gradle/gradle/issues/36940

This is a rewrite of the Isolated Projects userguide page:

- Manage user expectations about performance benefits by explaining how the feature works in different scenarios and what affects performance
- Provide a more holistic description of IP constraints
- Describe a general migration path with initial steps and recommend build logic refactoring strategies
- Clarify the relation between Isolated Projects and other features
- Remove claims that are not currently valid, such as claims about project model caching, which was disabled in 9.3

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
@@ -18,152 +18,529 @@
 = Isolated Projects
 :keywords: ip
 
-Isolated Projects is a pre-alpha Gradle feature that extends the configuration cache to further improve performance, particularly the performance of Android Studio and IDEA sync.
+Isolated Projects is an experimental feature aimed at improving Gradle scalability and performance, especially for large builds and IDE sync (Android Studio / IntelliJ IDEA).
 
-When Isolated Projects is enabled, the configuration model of Gradle projects are "isolated" from each other.
-This means that build logic, such as build scripts or plugins, applied to a project cannot directly access the mutable state of another project.
-This allows configuration and tooling model creation for each project to safely run in parallel, with the result cached and invalidated independently for each project.
+When the Isolated Projects feature is enabled, the projects in a <<multi_project_builds.adoc#multi_project_builds,multi-project build>> are "isolated" from each other.
+This means that build logic, such as build scripts or plugins, applied to a project cannot directly access the mutable state of other projects.
+This constraint unlocks two things: parallel project configuration and finer-grained caching of configuration results.
 
-== Status as of Gradle 8.11
-
-When Isolated Projects is enabled, Gradle applies two levels of caching during IDE sync:
-
-1. Gradle starts by applying coarse-grained caching.
-+
-To do this, Gradle caches the result of the entire sync operation and reuses it when nothing that affects the IDE model has changed.
-When the cache entry can be reused, Gradle short-circuits the entire sync operation and returns the cached result to the IDE.
-
-2. Generally, the settings and build scripts affect the IDE model, but the project's source code does not.
-So, when these scripts change, the cache entry cannot be reused.
-When this happens, Gradle falls back to fine-grained caching.
-+
-To do this, Gradle caches the result of creating the tooling models for each project and reuses these models when nothing that affects them has changed.
-When a project's cached models can be reused, Gradle short-circuits all work on that project, including the configuration phase and other work such as dependency resolution.
-
-This means that Gradle will only configure and create tooling models for projects whose configuration has changed.
-This work is done in parallel for each project.
-
-== Current limitations
-
-Isolated Projects is a pre-alpha feature, and as such, the current implementation has a number of limitations.
-These will be addressed in future Gradle releases:
-
-* Gradle, IntelliJ IDEA, Android Studio, and Kotlin plugins are not yet 100% compatible with Isolated Projects, so you should expect to see some violations reported. The teams are actively working on fixing these incompatibilities.
-* Parallel configuration does not support <<configuration_on_demand.adoc#sec:configuration_on_demand, Configuration on Demand>>. All projects will be configured, even if they aren't running tasks.
-* Changes to included builds invalidate all cached results, even when the change would not affect the cached results.
-* The implementation does not exploit the isolation to limit peak memory consumption. Currently, peak memory consumption is a function of how many projects must be configured.
-* All caching, including the configuration cache, is done on the local machine. Remote caching is not supported yet.
-
-== How do I use it?
-
-You will need Gradle 8.5 or later to use Isolated Projects, preferably a recent nightly.
-You should also use the most recent version of IDEA or Android Studio.
-
-The feature is off by default.
-You can enable it by setting the `org.gradle.unsafe.isolated-projects` Gradle property to `true`.
-For example:
+You can enable the feature by setting the `org.gradle.unsafe.isolated-projects` to `true` in `gradle.properties`,
+or passing a flag on the command line:
 
 [source,bash]
 ----
 $ ./gradlew build -Dorg.gradle.unsafe.isolated-projects=true
 ----
 
-When enabled, Gradle will fail the build whenever build logic attempts to cross project boundaries and access the model of another project.
-Gradle collects all of these access problems in the configuration cache report, as it does other problems.
+Follow the <<sec:ip_migration_path,recommended migration steps>> to make your build logic compatible with <<sec:ip_constraints,Isolated Projects constraints>>.
 
-The configuration cache command-line options can be used to control how Gradle handles these problems. For example:
+[WARNING]
+====
+Isolated Projects is an experimental feature that has not yet reached <<feature_lifecycle.adoc#sec:incubating_state,incubating>> state.
+APIs and behavior may change at any time without notice.
+The feature will enter incubation once Gradle's core has stronger support for this mode and more common use cases are covered.
+====
 
-* `--configuration-cache-problems=warn` can be used to treat access problems as warnings instead of errors.
-* `-Dorg.gradle.configuration-cache.max-problems=x` can be used to increase the maximum number of problems included in the report.
+[[sec:ip_changelog]]
+== Recent changes
 
-You can also use `-Dorg.gradle.internal.invalidate-coupled-projects=false` to force parallel configuration when there are access problems.
+While the feature is experimental, notable changes are documented here rather than in release notes:
 
-Note that these options disable the validation that makes execution parallel- and caching-safe when Isolated Projects is enabled, so you may see some unexpected behavior when using them.
+* Gradle 9.4.0
+** Isolated Projects violations are now deduplicated, making the Configuration Cache HTML reports significantly smaller.
 
-== Build logic constraints
+* Gradle 9.3.0
+** Tooling model caching is now disabled, as it currently has known correctness problems.
 
-Isolated Projects prevents build logic from accessing the state of another project.
-This includes:
+* Gradle 9.2.0
+** Additional APIs on `TaskContainer` are treated as incompatible: `findByPath(String)` and `getByPath(String)`
+methods now emit violations when a task from another project is requested.
 
-* Using most methods on the `Project` type. A small number of methods that return immutable information about the project are allowed:
-** `getName()`
-** `getPath()`
-** `getBuildTreePath()`
-** `getProjectDir()`
-** `getRootDir()`
-** `getChildProjects()`
-** `getSubprojects()`
-** `getAllProjects()`
-** `project()` overloads
-** `subprojects()` overloads
-** `allprojects()` overloads
+[[sec:ip_performance]]
+== Performance benefits and limitations
 
-Note that Isolated Projects is a pre-alpha feature.
-These constraints are not final and can change at any time.
+The main goal of Isolated Projects is to *speed up or entirely avoid running the <<build_lifecycle.adoc#configuration,configuration phase>>*.
+This is <<sec:ip_vs_cc,similar to the Configuration Cache>> feature, which allows skipping configuration when you rerun a task without changing the build configuration inputs.
+In fact, Isolated Projects uses Configuration Cache infrastructure as the foundation.
 
-== ChangeLog
+Builds that will get the most benefit are those that have a long configuration phase, such as large multi-project builds.
+Even for builds with short configuration times, keeping that phase under a few seconds helps developers stay in flow and avoid losing momentum.
 
-=== Gradle 9.3.0
+Here is how Isolated Projects affects the <<build_lifecycle.adoc#build_phases,lifecycle of the build>>:
 
-==== Disable tooling model caching by default
+1. Initialization phase
++
+Running init scripts and evaluating settings remains sequential with Isolated Projects.
+Compared to other phases, this phase is relatively quick, especially for large builds.
 
-Tooling model caching applies to scenarios like IDE sync.
-As it currently has known problems, we disable it for now to simplify early adoption.
+2. Configuration phase
++
+Isolated Projects directly improves the performance of this phase by *running project configuration in parallel*.
 
-=== Gradle 9.2.0
+3. Execution phase
++
+Isolated Projects has no effect on this phase, as the tasks are already isolated from each other and run in parallel with Configuration Cache.
 
-==== Additional APIs treated as incompatible with Isolated Projects
+[NOTE]
+====
+In case of a Configuration Cache hit, the initialization and configuration are <<configuration_cache.adoc#subsequent_runs,skipped entirely>>.
+This applies with Isolated Projects as well.
+====
 
-The following two methods now emit IP violations when the requested task lives in a different project:
+An important takeaway is that enabling Isolated Projects on a machine with N cores will not make your build N times faster.
+If the configuration phase is only 40% of the build time,
+then that's the ceiling for improvement with Isolated Projects.
+That said, for large builds, even a few-percent improvement can translate into minutes or more of actual time,
+which can have a substantial impact on developer productivity.
 
-- `TaskContainer#findByPath(String)`
-- `TaskContainer#getByPath(String)`
+Overall, the performance benefits of Isolated Projects depend on a variety of factors,
+such as the size and shape of your build, hardware resources, as well as your workflow.
+It also depends on the way Gradle is invoked.
 
-=== Gradle 8.11
+[[sec:gradle_invocation_modes]]
+=== Gradle invocation modes
 
-==== Projects configured in parallel
+Most users interact with Gradle in different ways:
 
-When building from the command line (i.e. for task execution) with Isolated Projects enabled, projects are now configured in parallel.
+1. Running tasks (either via command-line or via IDE)
+2. Building <<tooling_api.adoc#tooling_api,tooling models>> (this happens behind the scenes when you do an IDE sync)
 
-Parallel configuration does not yet support Configuration on Demand. All projects will be configured, even those which have no tasks to run.
+The configuration phase runs in both cases, but it is significantly different when model building is involved.
+Because of this, the performance benefits of Isolated Projects also depend on the way of interacting.
 
-=== Gradle 8.9
+[NOTE]
+====
+Performance improvements for Isolated Projects are released gradually across Gradle versions.
+Some improvements might bring benefits only for running tasks, while others for model building (and IDE sync by extension).
+As the Isolated Projects feature matures, the improvements will serve more scenarios.
+====
 
-==== Lift restriction on string-based task dependency notation
+[[sec:ip_perf_tasks]]
+==== Performance when running tasks
 
-Depending on a task from another project in string-notated form is a common idiom:
+When running tasks, the configuration phase consists of multiple steps.
 
-[source,text]
+1. Configure projects
++
+Isolated Projects makes this step run in **parallel**, which can significantly improve performance for builds with many projects.
+In the current implementation, all projects are always configured; there is no project configuration avoidance yet, whether by skipping unneeded projects or by using cached results.
+
+2. Discover the work graph
++
+The work graph is constructed by starting with the requested tasks and following the dependencies until all required work is discovered.
+This is not a directly parallelizable process.
+In the current implementation, even with Isolated Projects, this step is **sequential**.
+For invocations that result in having large work graphs, this step can become a performance bottleneck.
+An example is running checks or tests on <<command_line_interface.adoc#executing_tasks_in_multi_project_builds,all projects via an unqualified task>>, such as `./gradlew check`, which is typical on CI.
+
+3. Store the work graph (Configuration Cache)
++
+Isolated Projects makes this step run in **parallel**.
+This is the <<sec:ip_vs_parallel_cc,same as with Parallel Configuration Cache>>,
+but with the additional safety provided by the <<sec:ip_constraints,Isolated Projects constraints>>.
+
+4. Load the work graph (Configuration Cache)
++
+This step is **already parallel** with Configuration Cache; Isolated Projects has no additional effect here.
+
+[[sec:ip_perf_models]]
+==== Performance when building models (IDE Sync)
+
+For scenarios such as IDE sync, Gradle provides a special mechanism of building <<tooling_api.adoc#tooling_api,tooling models>>.
+The models describe the structure of the build, tell IDEs where to find production sources, and more.
+
+From Gradle's point of view, an IDE sync or a tooling invocation in general consists of the following parts (_simplified_):
+
+1. Configure projects
++
+Isolated Projects makes this step run in **parallel**, which can significantly improve performance for builds with many projects.
+In the current implementation, all projects are always configured; there is no project configuration avoidance yet, whether by skipping unneeded projects or by using cached results.
+
+2. Run model builders
++
+The invoking tool (e.g., the IDE) controls how individual models are built.
+Because of this, enabling Isolated Projects alone might not be enough to build models in parallel.
+However, in some cases it's possible to achieve parallel model building even without Isolated Projects,
+as explained below.
+
+3. Return the full model
++
+Similarly to how the task-running invocation finishes after all tasks have executed,
+the model-building invocation finishes after returning the final model.
+The model content is fully defined by the tool, e.g. the IDE.
+
+[NOTE]
+====
+There is a lot more nuance to this, but it's out of scope of this page.
+Individual models can be streamed back to the consumer before the final return.
+A tooling invocation can also request running tasks and build models before and after.
+All this and more affect the performance of a model building invocation and how Isolated Projects can improve it.
+====
+
+Without Isolated Projects, it is not possible to get a "Configuration Cache hit" for IDE sync
+because Configuration Cache does not currently support model building invocations.
+When Isolated Projects is enabled, the *full model can be cached and returned immediately* on later invocations
+if the build configuration hasn't changed.
+
+[[sec:parallel_model_building]]
+===== Parallel model building for IDE sync
+
+Parallel model building can significantly speed up IDE sync.
+However, both Gradle and the IDE need to be configured to allow it.
+
+On the Gradle side, parallel model building must be allowed in the build definition.
+It is *automatically allowed with Isolated Projects*.
+It can also be <<performance.adoc#sec:configure_tooling_api_actions_parallelism,allowed without Isolated Projects>>
+via `org.gradle.tooling.parallel` or `org.gradle.parallel` Gradle properties.
+
+[WARNING]
+====
+Enabling parallel model fetching without Isolated Projects can be unsafe.
+Parts of the configuration logic will run in parallel without supporting validations, potentially leading to flakiness or unreliable results.
+Isolated Projects makes this safe by enforcing <<sec:ip_constraints,constraints>> that prevent projects from sharing mutable state.
+====
+
+On the IDE side, the setup depends on the IDE:
+
+* Android Studio
++
+Parallel model building has been enabled automatically since Android Studio Electric Eel (2022.1.1).
+
+* IntelliJ IDEA
++
+Parallel model building is never enabled automatically in IntelliJ IDEA as of the latest 2026.1 release.
+It has to be manually enabled in the UI at `Setting › 'Build, Execution, Deployment' › Build Tools › Gradle`
+via the `Enable parallel Gradle model fetching` checkbox.
+The setting is only applied to the current project opened in the IDE.
+
+If you have already been using parallel model building for IDE sync, then Isolated Projects will not provide a performance benefit in that part of the configuration phase.
+However, Isolated Projects will catch violations as you change your build logic, preventing correctness issues from creeping in silently.
+
+[[sec:ip_parallelism_controls]]
+=== Parallelism controls
+
+Much of the performance benefit of Isolated Projects comes from running project configuration in parallel.
+The degree of parallelism is controlled by the <<build_environment.adoc#sec:property:workers_max,`org.gradle.workers.max` Gradle property>>,
+which sets the maximum number of parallel units of work across all <<build_lifecycle.adoc#build_lifecycle,build phases>>,
+including configuration.
+
+The default value is the number of CPU cores on your machine, which is usually a good choice.
+If you've explicitly set it lower for any reason, keep in mind that it will also limit how much parallelism Isolated Projects can use.
+
+[[sec:ip_current_limitations]]
+=== Current limitations
+
+The current implementation of Isolated Projects has a number of limitations that might be addressed in the future.
+
+* All projects always get configured, and <<sec:ip_vs_cod,Configuration on Demand has no effect>>.
+* A subproject can't start configuring until all its parent projects are done, which limits the parallelism.
+* The configuration phase might require more memory to process more work in parallel.
+* Changes to included builds invalidate all cached results, even unrelated ones.
+* Tooling model caching is local only: no sharing across checkouts or remote caching.
+
+[[sec:ip_constraints]]
+== Isolated Projects constraints
+
+During configuration, Gradle runs lots of smaller units of work (evaluating settings, configuring individual projects, etc.),
+and each one mutates live state on objects like `Project`, `Settings`, `Gradle`.
+
+Isolated Projects doesn't change how individual projects or <<composite_builds.adoc#composite_builds,included builds>> are configured;
+<<plugins.adoc#project_vs_settings_vs_init_plugins,plugins>> still mutate the state of the target they're applied to.
+What it adds is strict boundaries that prevent projects and builds from observing or mutating each other's state.
+
+With Isolated Projects, it is **not allowed to observe or change**:
+
+* In project-scope, mutable state of other projects (<<sec:ip_cross_project_access,cross-project access>>)
+* In project-scope, mutable state of the owner build (<<sec:ip_project_to_build_access,project-to-build access>>)
+* In build-scope, mutable state of other builds (<<sec:ip_cross_build_access,cross-build access>>)
+* In build-scope, mutable state of the build that is passed into project-scope via callbacks (<<sec:ip_build_to_project_access,build-to-project access>>)
+
+Among the build logic patterns described by these constraints, <<sec:ip_cross_project_access,cross-project access>> is by far the most common.
+
+The constraints are designed to allow safely running the individual configuration units in parallel and also caching the results of their execution.
+
+[[sec:ip_constraint_violations]]
+=== Constraint violations
+
+The violations of Isolated Projects constraints are detected at runtime during an invocation.
+When at least one violation is detected, the build results cannot be considered reliable and the build will fail.
+
+Here is an example of the build logic that contains a violation:
+
+[source,kotlin]
+.app/build.gradle.kts
 ----
-foo.dependsOn(":a:bar")
+// ⚠️ `Project.version` is mutable state of the root project and cannot be accessed from a subproject
+val commonVersion = rootProject.version
 ----
 
-Starting with this release, this is no longer considered a violation of Isolated Projects boundaries.
+[source,bash]
+----
+$ ./gradlew -Dorg.gradle.unsafe.isolated-projects=true
+...
+- Build file 'app/build.gradle.kts': Project ':app' cannot access 'Project.version' functionality on another project ':'
+...
+BUILD FAILED in 0s
+----
 
-==== `gradle init` generates Isolated Projects compatible projects
+[[sec:ip_cross_project_access]]
+=== Cross-project access constraints
 
-The <<build_init_plugin.adoc#build_init_plugin, Build Init Plugin>> supports creating multi-module projects.
+The cross-project access constraints apply during project configuration when a `Project` instance representing another project is involved.
+Each instance of `Project` has both mutable and immutable state.
+Accessing mutable state will result in a violation, but accessing immutable state is allowed.
 
-Starting with this release, `gradle init` generates projects compatible with Isolated Projects restrictions.
+[[sec:ip_mutable_cross_project]]
+==== Restricted access to mutable data of other projects
 
-==== `IsolatedProject` provides a project identifier in composite builds
+Most parts of the `Project` state are directly or indirectly mutable via the methods on the interface.
+Accessing that state on _another_ project is not allowed with Isolated Projects.
 
-The link:{javadocPath}/org/gradle/api/project/IsolatedProject.html[IsolatedProject] type was introduced in Gradle 8.8 to explicitly mark the project state that is safe to access across projects.
+Examples of mutable state (not exhaustive)::
+* Basic state: `description`, `group`, `version`, `status`, `buildDir`
+* Containers: `tasks`, `dependencies`, `configurations`, `artifacts`, `repositories`, `components`, `plugins`, `extensions`, `ext`
+* Services: `layout`, `providers`, `objects`, `dependencyFactory`
 
-Gradle 8.9 adds a link:{javadocPath}/org/gradle/api/project/IsolatedProject.html#getBuildTreePath()[`buildTreePath`] member, which serves as a unique project identifier in <<composite_builds.adoc#composite_builds, composite build>> setups.
+The `Project` interface also provides many convenience methods that allow configuring it, or registering callbacks.
+Most of those methods indirectly access the mutable state and are also not allowed to be called on another project.
 
-=== Gradle 8.8
+In general, it is easier to assume calling any method on another `Project` is forbidden,
+unless it is one of the handful getters that provide _immutable data_.
 
-==== New Gradle lifecycle callbacks
+[[sec:ip_immutable_cross_project]]
+==== Unrestricted access to immutable data of other projects
 
-This release introduces a new link:{javadocPath}/org/gradle/api/invocation/GradleLifecycle.html[`GradleLifecycle`] API, accessible via `gradle.lifecycle`, which plugin authors and build engineers can use to register actions to be executed at certain points in the build lifecycle.
+Immutable project state includes data that was configured during settings evaluation and finalized before any project configuration has begun.
+This includes basic information about the project, but also the structure of the project hierarchy, since it can no longer be changed during project configuration.
 
-Actions registered as `GradleLifecycle` callbacks (currently, `beforeProject` and `afterProject`) are *link:{javadocPath}/org/gradle/api/IsolatedAction.html)[isolated]*, running in an isolated context that is private to every project.
-This will allow Gradle to perform additional performance optimizations and will be required in the future to take advantage of parallelism during the build configuration phase.
+Immutable project data::
+* Basics: `name`, `displayName`, `path`, `buildTreePath`, `depth`
+** Caution: `group` and `version` are **mutable**!
+* Directories/files: `projectDir`, `buildFile`, `rootDir`
+** Caution: `buildDir` and `layout` are **mutable**!
 
-While the existing callbacks continue to work, we encourage everyone to adopt the new API and provide us with early feedback.
+Project hierarchy navigation::
+* `isolated: IsolatedProject`
+* `rootProject: Project`
+* `parent: Project?`
+* `project(): Project` overloads
+* `childProjects: Map<String, Project>`
+* `subprojects()` and `allprojects()` overloads
 
-The example below shows how this new API could be used in a settings script or <<plugins.adoc#project_vs_settings_vs_init_plugins, settings plugins>> to apply configuration to all projects,
+It is allowed to traverse the project structure and access immutable data of other projects.
+However, it is best to reduce such interactions to a minimum and instead express them through dependencies between projects.
+
+[[sec:ip_project_to_build_access]]
+=== Project-to-build access constraints
+
+[WARNING]
+====
+Project-to-build access constraints are not fully enforced yet.
+====
+
+Project configuration logic has access to the build-scoped state shared across projects via the `Gradle` interface.
+
+A large part of the `Gradle` interface is dedicated to callback registration.
+Many of the callbacks are either no-ops or throw an exception if you try registering them after settings have been evaluated.
+Callbacks that do get successfully registered may execute in a different order per invocation due to parallel project configuration.
+
+[[sec:ip_shared_build_services]]
+==== Shared Build Services
+
+<<build_services.adoc#build_services,Shared Build Services>> are also part of the build-scoped state.
+As the name suggests, they are shared between projects in the build.
+However, the API for registering and observing their registrations was not designed with parallel project configuration in mind.
+
+Generally speaking, registering a build service via `gradle.sharedServices.registerIfAbsent(...)` is safe with Isolated Projects.
+Only the first registration succeeds, and any later registrations return the previously registered instance.
+However, the parameters of the build service can become shared mutable state.
+Any mutation of the build service parameters outside of `registerIfAbsent(...) { parameters { ... } }` should be avoided.
+
+Accessing the set of build service registrations at `gradle.sharedServices.registrations` is unsafe.
+
+[[sec:ip_cross_build_access]]
+=== Cross-build access constraints
+
+[WARNING]
+====
+Cross-build access constraints are not fully enforced yet.
+====
+
+Build configuration logic (init scripts and settings) has access to the state of other builds via `gradle.parent`.
+
+Since (included) builds can be configured in parallel, access to the shared mutable state is also problematic.
+
+With Isolated Projects, using `gradle.parent` should be avoided.
+
+[[sec:ip_build_to_project_access]]
+=== Build-to-project access constraints
+
+[WARNING]
+====
+Build-to-project access constraints are not fully enforced yet.
+====
+
+Build configuration logic (init script and settings) can affect project configuration by registering callbacks that will run later in the build lifecycle.
+
+Some of the callbacks are executed around project evaluation time: immediately before or after the project configuration.
+With Isolated Projects, when project configuration runs in parallel, these callbacks might execute in parallel as well.
+If the callbacks capture any mutable state from the build-scope, that state becomes shared mutable state, leading to concurrency problems.
+
+See the section on <<sec:isolating_lifecycle_callbacks,state-isolating callbacks>> for a better alternative.
+
+[[sec:ip_migration]]
+== Migration
+
+This section covers how to migrate your build to be compatible with Isolated Projects.
+
+[WARNING]
+====
+Isolated Projects is an experimental feature, and it is actively changing.
+Make sure to use the latest available version of Gradle and that you are reading the documentation for that version.
+====
+
+[[sec:ip_before_migration]]
+=== Before the migration
+
+* Upgrade the build to the latest version of Gradle
++
+If available, consider a pre-release version such as a milestone or a nightly to get the best experience while Isolated Projects is experimental.
+
+* Migrate third-party plugins to their latest versions
++
+Ecosystem plugins and community plugins gradually improve support for Isolated Projects.
+By using the latest versions you reduce the risks of dealing with problems that have already been fixed.
+
+* Use the latest tooling, such as the latest versions of Android Studio and IntelliJ IDEA
++
+The IDE support for Isolated Projects is already in a good state for 2025.3 and later releases.
+However, newer IDE versions can bring better performance as well as a better user experience, when going through the migration.
+
+* Make your build <<configuration_cache_enabling.adoc#config_cache:adoption,compatible with Configuration Cache>>
++
+The Isolated Projects feature builds on top of the Configuration Cache feature.
+
+[[sec:ip_enable]]
+=== Enabling Isolated Projects
+
+Isolated Projects can be enabled by setting the `org.gradle.unsafe.isolated-projects` Gradle option to `true`.
+
+You can pass it as an argument when running tasks from the command line:
+
+[source,bash]
+----
+$ ./gradlew help -Dorg.gradle.unsafe.isolated-projects=true
+----
+
+To make use of the feature in IDE sync scenarios, it has to be enabled in the `gradle.properties` file:
+
+[source,properties]
+.gradle.properties
+----
+org.gradle.unsafe.isolated-projects=true
+----
+
+If you have never enabled the feature in the build before, it's likely that the build is not compatible and contains Isolated Projects <<sec:ip_constraint_violations,constraint violations>>.
+In that case the build will fail, and one or more violations will be shown in the error message.
+
+[[sec:ip_ignore_violations]]
+=== Temporarily ignoring violations
+
+To guarantee reliable build results, Gradle will fail the build when any violations of <<sec:ip_constraints,Isolated Project constraints>> are detected.
+However, in the initial stages of the Isolated Projects adoption journey,
+it can be useful to gauge an estimate of the speed-up provided by the feature in your particular setup and workflow.
+
+This can be achieved by temporarily ignoring the violations and not failing the build on the basis of their presence.
+It is especially relevant for the IDE sync scenarios, where the violations are preventing a sync from completion.
+
+Since Isolated Projects violations are reported via the same mechanism as <<configuration_cache_enabling.adoc#config_cache:usage:ignore_problems,Configuration Cache problems>>,
+enabling Configuration Cache _warning mode_ achieves the desired result.
+
+[source,properties]
+.gradle.properties
+----
+# ⚠️ Temporary measure of ignoring Isolated Projects violations. Do not enable permanently!
+org.gradle.configuration-cache.problems=warn
+----
+
+[NOTE]
+====
+This does not guarantee that the build will succeed.
+Concurrency issues may still cause exceptions which fail the build.
+====
+
+Alternatively, a flag can be used on the command line:
+
+[source,bash]
+----
+$ ./gradlew --configuration-cache-problems=warn
+----
+
+[WARNING]
+====
+Warning mode is a migration and troubleshooting aid and not intended as a persistent way of ignoring incompatibilities.
+It will also not prevent new incompatibilities being accidentally added to your build later.
+====
+
+[[sec:ip_migration_path]]
+=== Recommended migration path
+
+This section describes how to start making your build compatible with Isolated Projects.
+
+1. Run `./gradlew help -Dorg.gradle.unsafe.isolated-projects=true` to get the lay of the land
++
+Running the help task makes Gradle configure all projects.
+This way Gradle can detect Isolated Projects violations that are present in any workflow and should be addressed first.
+There can be more violations hidden by laziness, such as in the task configuration logic, but those can be addressed later.
++
+You can find the Isolated Projects violations in the Configuration Cache HTML report.
+Use the report to pinpoint the incompatible pieces of build logic and plugins.
+
+2. Report the violations coming from the community plugins to their maintainers.
++
+An incompatible plugin can prevent you from getting the performance benefits of Isolated Projects.
+It is best to report the incompatibilities as soon as possible so the plugin maintainers can address them, while you make your own build logic compatible.
+
+3. Follow the <<sec:ip_adoption_strategies,build-logic refactoring strategies>> to refactor your build logic and make it compatible.
++
+Remember that Isolated Projects violations only detect concrete instances of the problem, e.g. one specific project touching some mutable state of another.
+It might not be enough to replace one API call with another to address the problem.
+
+[[sec:ip_adoption_strategies]]
+=== Build-logic refactoring strategies
+
+[[sec:ip_convention_plugins]]
+==== Use convention plugins
+
+The most common cause of Isolated Projects violations is projects trying to configure other projects.
+This is often done to have a single place where some configuration logic is defined
+and then reused by applying it to many projects via callbacks like `allprojects {}`.
+
+<<sharing_build_logic_between_subprojects.adoc#sec:sharing_logic_via_convention_plugins,Convention plugins>> address the same need from a different perspective.
+The configuration logic is defined once in a <<plugins.adoc#sec:convention_plugins,convention plugin>> and then explicitly applied in projects that need it.
+Because each project applies the plugins to itself, no Isolated Projects constraints are violated.
+
+Additionally, convention plugins can be composed.
+This allows for more elaborate reuse of build logic without compromising Isolated Projects compatibility.
+
+[[sec:isolating_lifecycle_callbacks]]
+==== Use state-isolating lifecycle callbacks
+
+Sometimes it is still desirable to apply configuration to many projects at once.
+This is usually done with a callback such as `allprojects {}` called from a project or registered in the build scope via `gradle`.
+
+One problem with these callbacks is that the owner of the callback ends up touching state of another scope.
+However, there is another subtle yet important issue: callbacks can inadvertently share mutable state of the owner with all targets of the callback.
+When the target projects are configured in parallel, it's not possible to safely run the corresponding callbacks.
+
+The solution Gradle offers is _state-isolating_ callbacks.
+This works on the same principles as the isolation of task state performed by Configuration Cache to allow running tasks in parallel.
+
+The link:{javadocPath}/org/gradle/api/invocation/GradleLifecycle.html[`GradleLifecycle`] API,
+accessible via `gradle.lifecycle`, offers `beforeProject` and `afterProject` callbacks.
+Actions registered as `GradleLifecycle` callbacks are *link:{javadocPath}/org/gradle/api/IsolatedAction.html[isolated]* and will run in an isolated context that is private to each project.
+
+The example below shows how this API could be used in a settings script or <<plugins.adoc#project_vs_settings_vs_init_plugins, settings plugins>> to apply configuration to all projects,
 while avoiding <<sharing_build_logic_between_subprojects.adoc#sec:convention_plugins_vs_cross_configuration, cross-project configuration>>:
 
 [source,kotlin]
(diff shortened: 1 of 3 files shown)


alllex:
@bot-gradle test BD

alllex:
@bot-gradle test BD

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+
+The main goal of Isolated Projects is to *speed up or entirely avoid running the <<build_lifecycle.adoc#configuration,configuration phase>>*.
+This is <<sec:ip_vs_cc,similar to the Configuration Cache>> feature, which allows skipping configuration when you rerun a task without changing the build configuration inputs.
+In fact, Isolated Projects uses Configuration Cache infrastructure as the foundation.
+
+Builds that will get the most benefit are those that have a long configuration phase, such as large multi-project builds.
+If configuration time in your build is short, the performance gains will likely be small.
+Even so, keeping configuration time under a few seconds helps developers stay in the flow while working with the build.
+
+Here is how Isolated Projects affect the <<build_lifecycle.adoc#build_phases,lifecycle of the build>>:
```
octylFractal: Weird subject-verb agreement here: while Isolated Projects looks plural which would be "affect", as a term we tend to use it in the singular as it refers to a single feature, so it should be "affects".
```suggestion
Here is how Isolated Projects affects the <<build_lifecycle.adoc#build_phases,lifecycle of the build>>:
```

This is consistent with some of the earlier parts of the document.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+
+1. Initialization phase
++
+Running init scripts and evaluating settings remains sequential with Isolated Projects.
+Compared to other phases, this phase is relatively quick, especially for large builds.
+
+2. Configuration phase
++
+Isolated Projects directly improve the performance of this phase by *running project configuration in parallel*.
```
octylFractal: Same plural/singular confusion here.
```suggestion
Isolated Projects directly improves the performance of this phase by *running project configuration in parallel*.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+
+3. Execution phase
++
+Isolated Projects have no effect on this phase, as the tasks are already isolated from each other and run in parallel with Configuration Cache.
```
octylFractal: Same plural/singular confusion here.
```suggestion
Isolated Projects has no effect on this phase, as the tasks are already isolated from each other and run in parallel with Configuration Cache.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+
+[NOTE]
+====
+In case of a Configuration Cache hit, the initialization and configuration phases are skipped entirely.
+This applies with Isolated Projects as well.
+====
+
+An important takeaway is that enabling Isolated Projects on a machine with N cores will not make your build N times faster.
```
octylFractal, line 90: Should we be even clearer here and emphasize that it also won't reduce the _configuration_ time by N times, due to overhead and some unparallelizable work?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
-2. Generally, the settings and build scripts affect the IDE model, but the project's source code does not.
-So, when these scripts change, the cache entry cannot be reused.
-When this happens, Gradle falls back to fine-grained caching.
+2. Discover the work graph
 +
-To do this, Gradle caches the result of creating the tooling models for each project and reuses these models when nothing that affects them has changed.
-When a project's cached models can be reused, Gradle short-circuits all work on that project, including the configuration phase and other work such as dependency resolution.
+The work graph is constructed by starting with the requested tasks and following the dependencies until all required work is discovered.
+This is not a directly parallelizable process.
+In the current implementation, **even with Isolated Projects, this step is sequential**.
```
octylFractal: Opinion: the bold works better if applied only the the part that really needs emphasizing:
```suggestion
In the current implementation, **even with Isolated Projects**, this step is sequential.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+For invocations that result in having large work graphs, this step can become a performance bottleneck.
+An example is running checks or tests on all projects via an unqualified task, such as `./gradlew check`, which is typical on CI.
 
-This means that Gradle will only configure and create tooling models for projects whose configuration has changed.
-This work is done in parallel for each project.
+3. Store the work graph (Configuration Cache)
++
+Isolated Projects makes this step **parallel**.
```
octylFractal:
```suggestion
Isolated Projects makes this step run in **parallel**.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+
+[NOTE]
+====
+There is a lot more nuance to this, but it's out of scope of this page.
+Individual models can be streamed back to the consumer before the final return.
+A tooling invocation can also request running tasks and build models before and after.
+All this and more affect the performance of a model building invocation and how Isolated Projects can improve it.
+====
+
+Without Isolated Projects, it is not possible to get "Configuration Cache hit" for IDE sync
```
octylFractal: I think this is right?
```suggestion
Without Isolated Projects, it is not possible to get a "Configuration Cache hit" for IDE sync
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+However, it requires both Gradle and IDE settings to allow for it.
+
+On the Gradle side, parallel model building must be allowed in the build definition.
+It is *automatically allowed with Isolated Projects*.
+It can also be <<performance.adoc#sec:configure_tooling_api_actions_parallelism,allowed without Isolated Projects>>
+via `org.gradle.tooling.parallel` or `org.gradle.parallel` Gradle properties.
+
+[WARNING]
+====
+Enabling parallel model fetching without Isolated Projects can be unsafe.
```
octylFractal, line 198: Tangential, but does this mean we should be calling it `org.gradle.unsafe.tooling.parallel`?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+
+* IntelliJ IDEA
++
+Parallel model building is never enabled automatically in IntelliJ IDEA as of the latest 2026.1 release.
+It has to be manually enabled in the UI at `Setting › 'Build, Execution, Deployment' › Build Tools › Gradle`
+via the `Enable parallel Gradle model fetching` checkbox.
+The setting is only applied to the current project opened in the IDE.
+
+If you have already been using parallel model building for IDE sync, then Isolated Projects will not provide a performance benefit in that part of the configuration phase.
+However, *Isolated Projects provide the safety guarantees* via the <<sec:ip_constraints,Isolated Projects constraints>>, allowing you to evolve your build logic without silently introducing correctness issues.
```
octylFractal: Same plural/singular confusion here.
```suggestion
However, *Isolated Projects provides safety guarantees* via the <<sec:ip_constraints,Isolated Projects constraints>>, allowing you to evolve your build logic without silently introducing correctness issues.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+
+[[sec:ip_parallelism_controls]]
+=== Parallelism controls
+
+A large part of the performance benefits that Isolated Projects bring is in the ability to run more parts of the configuration phase in parallel.
```
octylFractal: Same plural/singular confusion here.
```suggestion
A large part of the performance benefits that Isolated Projects brings is in the ability to run more parts of the configuration phase in parallel.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+
+The current implementation of Isolated Projects has a number of limitations that might be addressed in the future.
+
+* All projects are always configured, and <<sec:ip_vs_cod,Configuration on Demand has no effect>>.
+* Projects are configured in parallel, but for any subproject all of its parents have to be configured first, which can limit the parallelism.
+* The configuration phase might require more memory to process more work in parallel.
 * Changes to included builds invalidate all cached results, even when the change would not affect the cached results.
-* The implementation does not exploit the isolation to limit peak memory consumption. Currently, peak memory consumption is a function of how many projects must be configured.
-* All caching, including the configuration cache, is done on the local machine. Remote caching is not supported yet.
+* Caching happens locally in a given build: the cache is not shared between different project checkouts, and remote caching is not supported.
```
octylFractal: I am confused by what this means. Does the build cache not work? It certainly seems to on my machine.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+
+The configuration phase consists of many smaller units of work, such as executing settings and configuring individual projects.
+When <<composite_builds.adoc#composite_builds,composite builds>> are involved, such as for build logic, similar work is happening within them as well.
+Each of those units of work is executable code that runs against the live state of the Gradle model, working with objects such as `Project`, `Settings`, `Gradle`, etc.
+The configuration happens through the mutation of state, e.g. `project.version = "1.0.0"`.
+
+With Isolated Projects, the approach to the configuration of the individual parts of the model stays the same.
+For instance, a `Project` plugin is still expected to mutate the state of the project it was applied to, e.g. by registering extensions and tasks.
+
+What Isolated Projects brings is the **strict boundaries between configuration units** to avoid observing and mutating shared mutable state.
```
octylFractal:
```suggestion
What Isolated Projects brings is **strict boundaries between configuration units** that forbid observing and mutating shared mutable state.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
 ----
 
-When enabled, Gradle will fail the build whenever build logic attempts to cross project boundaries and access the model of another project.
-Gradle collects all of these access problems in the configuration cache report, as it does other problems.
+[[sec:ip_cross_project_access]]
+=== Cross-project access constraints
+
+The cross-project access constraints apply during project configuration when a `Project` instance representing another project is involved.
+Each instance of `Project` has both mutable and immutable state.
+Access to a mutable state will result in a violation, while access to an immutable state is allowed.
```
octylFractal: This reads a little weird to me, perhaps:
```suggestion
Accessing mutable state will result in a violation, but accessing immutable state is allowed.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
-** `getRootDir()`
-** `getChildProjects()`
-** `getSubprojects()`
-** `getAllProjects()`
-** `project()` overloads
-** `subprojects()` overloads
-** `allprojects()` overloads
+Immutable project data::
+* Basics: `name`, `displayName`, `path`, `buildTreePath`, `depth`
+* Locations: `projectDir`, `buildFile`, `rootDir`
```
octylFractal: Should we use a different term than "locations" here, e.g. "Some locations"? `buildDir` is notably mutable.

Also, why isn't "path" or "buildTreePath" also a "location"?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+However, it is best to reduce such interactions to a minimum and instead express them through dependencies between projects.
 
-=== Gradle 9.3.0
+[[sec:ip_project_to_build_access]]
+=== Project-to-build access constraints
 
-==== Disable tooling model caching by default
+[WARNING]
+====
+Project-to-build access constraints are not fully enforced yet
```
octylFractal:
```suggestion
Project-to-build access constraints are not fully enforced yet.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+Accessing the set of build service registrations at `gradle.sharedServices.registrations` is unsafe.
 
-==== Projects configured in parallel
+[[sec:ip_cross_build_access]]
+=== Cross-build access constraints
 
-When building from the command line (i.e. for task execution) with Isolated Projects enabled, projects are now configured in parallel.
+[WARNING]
+====
+Cross-build access constraints are not fully enforced yet
```
octylFractal:
```suggestion
Cross-build access constraints are not fully enforced yet.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+====
 
-Parallel configuration does not yet support Configuration on Demand. All projects will be configured, even those which have no tasks to run.
+Build configuration logic (init scripts and settings) has access to the state of other builds via `gradle.parent`.
 
-=== Gradle 8.9
+Since (included) builds can be configured in parallel, access to the shared mutable state is also problematic.
 
-==== Lift restriction on string-based task dependency notation
+With Isolated Projects, using `gradle.parent` should be avoided.
```
octylFractal, line 373: Tangential: why is it that project hierarchy navigation is OK, but not build hierarchy?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
 
-Depending on a task from another project in string-notated form is a common idiom:
+[[sec:ip_build_to_project_access]]
+=== Build-to-project access constraints
 
-[source,text]
+[WARNING]
+====
+Build-to-project access constraints are not fully enforced yet
```
octylFractal:
```suggestion
Build-to-project access constraints are not fully enforced yet.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+====
+
+Build configuration logic (init script and settings) can affect project configuration by registering callbacks that will run later in the build lifecycle.
+
+Some of the callbacks are executed around project evaluation time: immediately before or after the project configuration.
+With Isolated Projects, when project configuration runs in parallel, these callbacks might execute in parallel as well.
+If the callbacks capture any mutable state from the build-scope, that state becomes a shared mutable state, leading to concurrency problems.
```
octylFractal:
```suggestion
If the callbacks capture any mutable state from the build-scope, that state becomes shared mutable state, leading to concurrency problems.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+Isolated Projects is an experimental feature, and it is actively changing.
+Make sure to use the latest available version of Gradle and that you are reading the documentation for that version.
+====
+
+[[sec:ip_before_migration]]
+=== Before the migration
+
+* Upgrade the build to the latest version of Gradle
++
+If available, consider a pre-release version such as a milestone or a nightly to get the best experience while the Isolated Projects is experimental.
```
octylFractal:
```suggestion
If available, consider a pre-release version such as a milestone or a nightly to get the best experience while Isolated Projects is experimental.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+
+* Migrate third-party plugins to their latest versions
++
+Ecosystem plugins and community plugins gradually improve support for Isolated Projects.
+By using the latest versions you reduce the risks of dealing with problems that have already been fixed.
+
+* Use the latest tooling, such as the latest versions of Android Studio and IntelliJ IDEA
++
+The IDE support for Isolated Projects is already in a good state for 2025.3 and later releases.
+However, newer IDE versions can bring better performance as well as better user experience, when going through the migration.
```
octylFractal:
```suggestion
However, newer IDE versions can bring better performance as well as a better user experience, when going through the migration.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+
+* Make your build <<configuration_cache_enabling.adoc#config_cache:adoption,compatible with Configuration Cache>>
++
+Isolated Projects feature builds on top of the Configuration Cache foundation.
```
octylFractal:
```suggestion
The Isolated Projects feature builds on top of the Configuration Cache feature.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+The Isolated Projects constraints extend the constraints introduced by Configuration Cache.
```
octylFractal: Is this necessary? It feels like it duplicates the previous line.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+org.gradle.unsafe.isolated-projects=true
+----
+
+If you have never enabled the feature in the build before, it's likely that the build is not compatible and contains Isolated Projects <<sec:ip_constraint_violations,constraint violations>>.
+In that case the build will fail, and one or more violations will be shown in the error message.
+
+[[sec:ip_ignore_violations]]
+=== Temporarily ignoring violations
+
+To guarantee reliable build results, Gradle will fail the build when any violations of Isolated Project <<sec:ip_constraints,constraints>> are detected.
```
octylFractal:
```suggestion
To guarantee reliable build results, Gradle will fail the build when any violations of <<sec:ip_constraints,Isolated Project constraints>> are detected.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,544 @@
...
+enabling Configuration Cache _warning mode_ achieves the desired result.
+
+[source,properties]
+.gradle.properties
+----
+# ⚠️ Temporary measure of ignoring Isolated Projects violations. Do not enable permanently!
+org.gradle.configuration-cache.problems=warn
+----
+
+NOTE: This does not guarantee that the build will succeed.
```
octylFractal:
```suggestion
NOTE: This does not guarantee that the build will succeed. Concurrency issues may still cause exceptions which fail the build.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -172,26 +564,95 @@ while avoiding <<sharing_build_logic_between_subprojects.adoc#sec:convention_plu
...
+This requires an opt-in because not all builds are compatible with this behavior, and there is no explicit validation of violations.
+
+Generally speaking, Parallel Execution is superseded by Configuration Cache because it enables running tasks in parallel even within a single project.
+
+However, Parallel Execution has an additional effect of allowing parallel tooling model building in scenarios like IDE sync.
+That is why many teams enable it alongside Configuration Cache.
+
+On the one hand, Isolated Projects extend Configuration Cache, enabling intra-project task parallelism.
+On the other hand, Isolated Projects constraints guarantee safe parallel tooling model building.
+That is why enabling Parallel Execution alongside Isolated Projects has no benefit.
```
octylFractal: I don't think these two things are opposite each other?
```suggestion
Isolated Projects extends the Configuration Cache, which enables intra-project task parallelism.
Additionally, Isolated Projects constraints guarantee safe parallel tooling model building, which allows parallel model building to be enabled by default.
Therefore, there is no benefit to enabling Parallel Execution alongside Isolated Projects.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -172,26 +564,95 @@ while avoiding <<sharing_build_logic_between_subprojects.adoc#sec:convention_plu
...
+
+The Parallel Execution flag is _ignored_ when Isolated Projects is enabled.
+
+[[sec:ip_vs_cod]]
+=== Configuration On Demand
```
octylFractal: Consistency:
```suggestion
=== Configuration on Demand
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I was going back and forth between the two variants. What trips me up is that IntelliJ's built-in grammar checker doesn't see through this and squiggle-highlights every singluar verb that comes after "Isolated Projects"

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I was going for bolding out things that might help repeat readers or people who skim the docs to get the idea without reading all paragraphs. In this case, it's reading the bullet point and then the bold text. If we bold only the "even with Isolated Projects", then the pattern breaks and you need to read all the words.

Maybe it's a wrong goal to pursue

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: The "unsafe" prefix was considered in the naming. It also sparked an extensive discussion in the Architecture Council after it was proposed as a general naming rule for "unsafe" features as part of the [Gradle Property naming ADR](https://github.com/gradle/gradle/blob/36589f07fe221bb66dcea9a449100ec4d8798a09/architecture/standards/0010-gradle-properties-naming.md). 

The discussion concluded that in order to have stable flags with the "unsafe" prefix, we should have a clear rule on how to decide whether something is safe or not. At the time, no such obvious rule was identified.

At least two other flags enable features of similar "unsafety" level: Parallel Execution (`org.gradle.parallel`) and Parallel Configuration Cache (`org.gradle.configuraiton-cache.parallel`). There is also Configure on Demand. None of them use the "unsafe" prefix.

It was argued that Configuration Cache by itself also has degrees of unsafety, which aren't explicitly verified, such as completely disregarding static state of classes as part of (de)serialization pipeline. I think there were other examples as well.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I don't feel strongly about your suggestion. We also don't want to undersell the feature too much.

If you feel strongly that we should clarify this, how would you phrase it?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: This is not about the build cache, but about tooling model caching. I updated the sentence. Is it more clear now?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: > Should we use a different term than "locations" here, e.g. "Some locations"?

The semantic group is "directories and files", and I went for "locations" as a common word. Does "Some locations" work well in that context?

> buildDir is notably mutable.

It is mutable, and that's why it's in the list in the other section, and here we have `buildFile`, which is immutable

> Also, why isn't "path" or "buildTreePath" also a "location"?

The "paths" that we expose on the interface are actually identifiers, such as `:foo`, so they don't belong into the "locations" sections. Though, I can see how it's easy to think that they should, based only on their names.

---

Given this context, do you still think we should rename the category?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: Agreed. Removed the line

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: There are valid use cases of walking the hierarchy, e.g. to add all projects or children projects as dependencies. On the build level it's very much not clear that walking the hierarchy is useful.

When a child build's settings are evaluated, it's possible the child build evaluation was triggered while we were evaluating the settings of the main build. In this case, accessing `gradle.parent` might even give you access to a not fully configured state.

The advice here is not wrong, but intentionally vague. We'll make it more precise together with introducing the runtime-enforced constraints.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
octylFractal: If you want to do that, I would bold the whole sentence then.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
octylFractal: I just remember this being a point of potential confusion that we discussed. I don't feel too strongly, and perhaps you're right that for now we may undersell it. I think we can reconsider when it reaches incubating and we know more about what the expected benefit generally is.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
octylFractal: Hmm, okay.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
octylFractal: Yes.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
octylFractal: The concern I have here is that someone may jump to the "immutable data" section, not read the mutable section, and infer that any "location" is immutable data. I don't know if providing the categories here has much benefit given how small they are.

"Some locations" is meant to clarify that not all of the location data is immutable state.

I'm fine with paths as they are I suppose, but if you really want to keep the category I think I would rename it to "Some filesystem locations".

blindpirate:
@bot-gradle test BD

blindpirate:
@bot-gradle test BD

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
 = Isolated Projects
 :keywords: ip
 
-Isolated Projects is a pre-alpha Gradle feature that extends the configuration cache to further improve performance, particularly the performance of Android Studio and IDEA sync.
+Isolated Projects is an experimental feature aimed at improving Gradle scalability and performance, particularly the performance of Android Studio and IDEA sync.
```
lkasso:
```suggestion
Isolated Projects is an experimental feature aimed at improving Gradle scalability and performance, especially for large builds and IDE sync (Android Studio / IntelliJ IDEA).
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
 
-When Isolated Projects is enabled, the configuration model of Gradle projects are "isolated" from each other.
-This means that build logic, such as build scripts or plugins, applied to a project cannot directly access the mutable state of another project.
-This allows configuration and tooling model creation for each project to safely run in parallel, with the result cached and invalidated independently for each project.
+When the Isolated Projects feature is enabled, the projects in a <<multi_project_builds.adoc#multi_project_builds,multi-project build>> are "isolated" from each other.
+This means that build logic, such as build scripts or plugins, applied to a project cannot directly access the mutable state of other projects.
+This constraint allows safely running project configuration in parallel and also caching the results of the configuration at a finer granularity, improving performance across a wider range of workflows.
```
lkasso:
```suggestion
This constraint unlocks two things: parallel project configuration and finer-grained caching of configuration results.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
 
-== Status as of Gradle 8.11
+You can enable the feature by setting the `org.gradle.unsafe.isolated-projects` to `true` in `gradle.properties`,
+or passing the flag on the command line:
```
lkasso:
```suggestion
or passing a flag on the command line:
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+$ ./gradlew build -Dorg.gradle.unsafe.isolated-projects=true
+----
+
+Follow the <<sec:ip_migration_path,recommended migration path>> to make your build logic compatible with the <<sec:ip_constraints,Isolated Projects constraints>>.
+
+[WARNING]
+====
+Isolated Projects is an experimental feature that has not yet reached <<feature_lifecycle.adoc#sec:incubating_state,incubating>> state.
+APIs and behavior may change at any time without notice.
+Once the core of Gradle better supports this mode of execution and enough common use cases are covered, the feature will enter incubation.
```
lkasso: I don't think this needs to be said.
```suggestion
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+You can enable the feature by setting the `org.gradle.unsafe.isolated-projects` to `true` in `gradle.properties`,
+or passing the flag on the command line:
 
-When Isolated Projects is enabled, Gradle applies two levels of caching during IDE sync:
+[source,bash]
+----
+$ ./gradlew build -Dorg.gradle.unsafe.isolated-projects=true
+----
+
+Follow the <<sec:ip_migration_path,recommended migration path>> to make your build logic compatible with the <<sec:ip_constraints,Isolated Projects constraints>>.
```
lkasso:
```suggestion
Follow the <<sec:ip_migration_path,recommended migration steps>> to make your build logic compatible with <<sec:ip_constraints,Isolated Projects constraints>>.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+====
+Isolated Projects is an experimental feature that has not yet reached <<feature_lifecycle.adoc#sec:incubating_state,incubating>> state.
+APIs and behavior may change at any time without notice.
+Once the core of Gradle better supports this mode of execution and enough common use cases are covered, the feature will enter incubation.
+====
+
+[[sec:ip_changelog]]
+== Recent changes
+
+While the feature stays experimental, notable changes are reported here instead of release notes.
```
lkasso:
```suggestion
While the feature is experimental, notable changes are documented here rather than in release notes:
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+[[sec:ip_performance]]
+== Performance benefits and limitations
+
+The main goal of Isolated Projects is to *speed up or entirely avoid running the <<build_lifecycle.adoc#configuration,configuration phase>>*.
+This is <<sec:ip_vs_cc,similar to the Configuration Cache>> feature, which allows skipping configuration when you rerun a task without changing the build configuration inputs.
+In fact, Isolated Projects uses Configuration Cache infrastructure as the foundation.
+
+Builds that will get the most benefit are those that have a long configuration phase, such as large multi-project builds.
+If configuration time in your build is short, the performance gains will likely be small.
```
lkasso: Let's combine this sentence with the next one.
```suggestion
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+Even so, keeping configuration time under a few seconds helps developers stay in the flow while working with the build.
```
lkasso:
```suggestion
Even for builds with short configuration times, keeping that phase under a few seconds helps developers stay in flow and avoid losing momentum.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
lkasso: What about something like:

```suggestion
Because configuration is only one phase of the build, enabling Isolated Projects on a machine with N cores will not make your build N times faster.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+Isolated Projects has no effect on this phase, as the tasks are already isolated from each other and run in parallel with Configuration Cache.
+
+[NOTE]
+====
+In case of a Configuration Cache hit, the initialization and configuration phases are skipped entirely.
+This applies with Isolated Projects as well.
+====
+
+An important takeaway is that enabling Isolated Projects on a machine with N cores will not make your build N times faster.
+If the configuration phase takes 40% of the time of a build invocation in your workflow,
```
lkasso:
```suggestion
If the configuration phase is only 40% of the build time,
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+then this boundary will limit the performance benefit of enabling Isolated Projects in that workflow.
```
lkasso:
```suggestion
then that's the ceiling for improvement with Isolated Projects.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+Having said that, for large builds even a few-percent improvement can translate into minutes or more of actual time,
```
lkasso:
```suggestion
That said, for large builds, even a few-percent improvement can translate into minutes or more of actual time,
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+which can have a substantial impact on developer productivity.
+
+Overall, the performance benefits of Isolated Projects depend on a variety of factors,
+such as the size and shape of your build, hardware resources, as well as your workflow.
```
lkasso, line 97: Is there a better word than "shape" here?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+It also depends on the way Gradle is invoked.
+
+[[sec:gradle_invocation_modes]]
+=== Gradle invocation modes
+
+Most users interact with Gradle in two quite different ways:
```
lkasso:
```suggestion
Most users interact with Gradle in different ways:
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+1. Running tasks (either via command-line or via IDE)
+2. Building <<tooling_api.adoc#tooling_api,tooling models>> -- this happens behind the scenes when you do an IDE sync
 
-1. Gradle starts by applying coarse-grained caching.
+The configuration phase runs in both cases, but it is significantly different when model building is involved.
```
lkasso, line 108: I feel like we need to define "model building" somewhere in our docs and link to it here. WDYT?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+Isolated Projects aims to speed up both scenarios, but the performance characteristics differ depending on how Gradle is invoked.
```
lkasso: We should reword this; it's a little awkward as-is. This suggestion also introduces the next sections, wydt:
```suggestion
Isolated Projects aims to speed up both scenarios, but the gains come from different mechanisms — parallel configuration when running tasks, and model caching when doing IDE sync.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+such as the size and shape of your build, hardware resources, as well as your workflow.
+It also depends on the way Gradle is invoked.
+
+[[sec:gradle_invocation_modes]]
+=== Gradle invocation modes
+
+Most users interact with Gradle in two quite different ways:
+
+1. Running tasks (either via command-line or via IDE)
+2. Building <<tooling_api.adoc#tooling_api,tooling models>> -- this happens behind the scenes when you do an IDE sync
```
lkasso:
```suggestion
2. Building <<tooling_api.adoc#tooling_api,tooling models>> (this happens behind the scenes when you do an IDE sync)
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
-When this happens, Gradle falls back to fine-grained caching.
+2. Discover the work graph
 +
-To do this, Gradle caches the result of creating the tooling models for each project and reuses these models when nothing that affects them has changed.
-When a project's cached models can be reused, Gradle short-circuits all work on that project, including the configuration phase and other work such as dependency resolution.
+The work graph is constructed by starting with the requested tasks and following the dependencies until all required work is discovered.
+This is not a directly parallelizable process.
+In the current implementation, **even with Isolated Projects, this step is sequential**.
+For invocations that result in having large work graphs, this step can become a performance bottleneck.
+An example is running checks or tests on all projects via an unqualified task, such as `./gradlew check`, which is typical on CI.
```
lkasso: ok this might be silly, but I don't actually know what "unqualified" task means.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
++
+This step is **already parallel** with Configuration Cache; Isolated Projects has no additional effect here.
+
+[[sec:ip_perf_models]]
+==== Performance when building models (IDE Sync)
 
-Isolated Projects is a pre-alpha feature, and as such, the current implementation has a number of limitations.
-These will be addressed in future Gradle releases:
+For scenarios such as IDE sync, Gradle provides a special mechanism of building <<tooling_api.adoc#tooling_api,tooling models>>.
+The models describe the structure of the build, tell IDEs where to find production sources and much more.
```
lkasso:
```suggestion
The models describe the structure of the build, tell IDEs where to find production sources, and more.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
 
-* Gradle, IntelliJ IDEA, Android Studio, and Kotlin plugins are not yet 100% compatible with Isolated Projects, so you should expect to see some violations reported. The teams are actively working on fixing these incompatibilities.
-* Parallel configuration does not support <<configuration_on_demand.adoc#sec:configuration_on_demand, Configuration on Demand>>. All projects will be configured, even if they aren't running tasks.
+From Gradle's point of view, an IDE sync or a tooling invocation in general consists of the following parts (simplified):
```
lkasso:
```suggestion
From Gradle's point of view, an IDE sync or a tooling invocation in general consists of the following parts (_simplified_):
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+1. Configure projects
++
+Isolated Projects makes this step run in **parallel**, which can significantly improve performance for builds with many projects.
+In the current implementation, all projects are always configured.
```
lkasso: This doesn't make sense to me. The first sentence is about parallelism, so you would think this second sentence would say something like: 

```suggestion
In the current implementation, all projects are always configured serially.
```

Is this what you meant?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+2. Run model builders
++
+The flow of building individual models is dictated by the invoking tool (e.g. the IDE).
```
lkasso:
```suggestion
+
The invoking tool (e.g., the IDE) controls how individual models are built.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+Because of this, enabling Isolated Projects alone might not be enough to build models in parallel.
+However, in some cases it's possible to achieve parallel model building even without Isolated Projects,
+as explained further in this section.
```
lkasso:
```suggestion
as explained below.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+3. Return the full model
++
+Similarly to how the task-running invocation finishes after all tasks have executed,
+the model-building invocation finishes after returning the final model.
```
lkasso, line 169: Again, I wish we had "model" defined somewhere. Maybe I should do a PR alongside this one?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+Without Isolated Projects, it is not possible to get a "Configuration Cache hit" for IDE sync
+because Configuration Cache does not currently support model building invocations.
+When Isolated Projects is enabled, the *full model can be cached and returned immediately* on later invocations
+if the build configuration hasn't changed.
+
+[[sec:parallel_model_building]]
+===== Parallel model building for IDE sync
+
+Parallel model building can significantly speed up IDE sync.
+However, it requires both Gradle and IDE settings to allow for it.
```
lkasso:
```suggestion
However, both Gradle and the IDE need to be configured to support it.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+On the Gradle side, parallel model building must be allowed in the build definition.
+It is *automatically allowed with Isolated Projects*.
+It can also be <<performance.adoc#sec:configure_tooling_api_actions_parallelism,allowed without Isolated Projects>>
+via `org.gradle.tooling.parallel` or `org.gradle.parallel` Gradle properties.
```
lkasso, line 194: We should link to either the Properties or Parallel enabling pages here.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+[WARNING]
+====
+Enabling parallel model fetching without Isolated Projects can be unsafe.
+Parts of the configuration logic will run in parallel without supporting validations, potentially leading to flakiness or unreliable results.
+With Isolated Projects it becomes safe due to the additional <<sec:ip_constraints,constraints>> that are always validated.
```
lkasso:
```suggestion
Isolated Projects makes this safe by enforcing <<sec:ip_constraints,constraints>> that prevent projects from sharing mutable state.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+* IntelliJ IDEA
++
+Parallel model building is never enabled automatically in IntelliJ IDEA as of the latest 2026.1 release.
+It has to be manually enabled in the UI at `Setting › 'Build, Execution, Deployment' › Build Tools › Gradle`
+via the `Enable parallel Gradle model fetching` checkbox.
+The setting is only applied to the current project opened in the IDE.
+
+If you have already been using parallel model building for IDE sync, then Isolated Projects will not provide a performance benefit in that part of the configuration phase.
+However, *Isolated Projects provides safety guarantees* via the <<sec:ip_constraints,Isolated Projects constraints>>, allowing you to evolve your build logic without silently introducing correctness issues.
```
lkasso:
```suggestion
However, Isolated Projects will catch violations as you change your build logic, preventing correctness issues from creeping in silently.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+[[sec:ip_parallelism_controls]]
+=== Parallelism controls
```
lkasso, line 220: This section is too complex for what it needs to convey. Here's what I propose instead:


```suggestion
=== Parallelism controls

Much of the performance benefit from Isolated Projects comes from running configuration in parallel. 
The degree of parallelism is controlled by the `org.gradle.workers.max` Gradle property, which sets the maximum number of parallel units of work across all build phases, including configuration.

The default value is the number of CPU cores on your machine, which is usually a good choice. 
If you've explicitly set it lower for any reason, keep in mind that it will also limit how much parallelism Isolated Projects can use.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+This property is usually configured in the Gradle User Home properties,
+because its value is normally tied to the machine running the builds rather than the builds themselves.
+====
+
+[[sec:ip_current_limitations]]
+=== Current limitations
+
+The current implementation of Isolated Projects has a number of limitations that might be addressed in the future.
+
+* All projects are always configured, and <<sec:ip_vs_cod,Configuration on Demand has no effect>>.
```
lkasso:
```suggestion
* All projects always get configured, and <<sec:ip_vs_cod,Configuration on Demand has no effect>>.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+* Projects are configured in parallel, but for any subproject all of its parents have to be configured first, which can limit the parallelism.
```
lkasso:
```suggestion
* A subproject can't start configuring until all its parent projects are done, which limits the parallelism.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+* The configuration phase might require more memory to process more work in parallel.
 * Changes to included builds invalidate all cached results, even when the change would not affect the cached results.
```
lkasso:
```suggestion
* Changes to included builds invalidate all cached results, even unrelated ones.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
-* The implementation does not exploit the isolation to limit peak memory consumption. Currently, peak memory consumption is a function of how many projects must be configured.
-* All caching, including the configuration cache, is done on the local machine. Remote caching is not supported yet.
+* Tooling model caching happens locally in a given build:
```
lkasso:
```suggestion
* Tooling model caching is local only:
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+same as for Configuration Cache, cached models are not shared between different project checkouts, and remote caching is not supported.
```
lkasso:
```suggestion
no sharing across checkouts or remote caching.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+[[sec:ip_constraints]]
+== Isolated Projects constraints
+
+The configuration phase consists of many smaller units of work, such as executing settings and configuring individual projects.
+When <<composite_builds.adoc#composite_builds,composite builds>> are involved, such as for build logic, similar work is happening within them as well.
```
lkasso:
```suggestion
When <<composite_builds.adoc#composite_builds,composite builds>> are involved, similar work is happening within them as well.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+Each of those units of work is executable code that runs against the live state of the Gradle model, working with objects such as `Project`, `Settings`, `Gradle`, etc.
```
lkasso: Still feel terrible "Gradle model" isn't properly defined anywhere else in the docs...

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+The configuration happens through the mutation of state, e.g. `project.version = "1.0.0"`.
```
lkasso: I don't understand this sentence. Maybe this entire first paragraph can be summarized to:

```
During configuration, Gradle runs lots of smaller units of work (evaluating settings, configuring individual projects, etc.), and each one mutates live state on objects like `Project`, `Settings`, `Gradle`.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+With Isolated Projects, the approach to the configuration of the individual parts of the model stays the same.
+For instance, a `Project` plugin is still expected to mutate the state of the project it was applied to, e.g. by registering extensions and tasks.
+
+What Isolated Projects brings is **strict boundaries between configuration units** that forbid observing and mutating shared mutable state.
```
lkasso: And this entire second paragraph can be:

```
Isolated Projects doesn't change how individual projects are configured; plugins still mutate the state of the project they're applied to. What it adds is strict boundaries that prevent projects from observing or mutating each other's state.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+
+It is **not allowed to observe or change**:
+
+* In project-scope, mutable state of other projects (<<sec:ip_cross_project_access,cross-project access>>)
```
lkasso, line 253: I like that you added scope to these. Nice!

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
lkasso: I personally prefer the category @alllex mentioned: "Directories/Files"

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+====
 
-Tooling model caching applies to scenarios like IDE sync.
-As it currently has known problems, we disable it for now to simplify early adoption.
+Project configuration logic has access to the build-scoped state shared across projects via the `Gradle` interface.
 
-=== Gradle 9.2.0
+A large part of the `Gradle` interface is dedicated to callback registration.
+Many of the callbacks are either no-ops or throw an exception if you try registering them after settings have been evaluated.
+Callbacks that do get successfully registered may execute in a different order per invocation due to parallel project configuration.
```
lkasso, line 345: Could we give a code example here?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
 
-==== Additional APIs treated as incompatible with Isolated Projects
+[[sec:ip_shared_build_services]]
+==== Shared Build Services
 
-The following two methods now emit IP violations when the requested task lives in a different project:
+Shared Build Services are also part of the build-scoped state.
```
lkasso: We should link to the doc page here.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -18,152 +18,548 @@
...
+Some of the callbacks are executed around project evaluation time: immediately before or after the project configuration.
+With Isolated Projects, when project configuration runs in parallel, these callbacks might execute in parallel as well.
+If the callbacks capture any mutable state from the build-scope, that state becomes shared mutable state, leading to concurrency problems.
+
+See the section on <<sec:isolating_lifecycle_callbacks,state-isolating callbacks>> for a better alternative.
+
+[[sec:ip_migration]]
+== Migration
+
+This section provides recommendations on how to approach the migration of your build to make and keep it compatible with Isolated Projects.
```
lkasso:
```suggestion
This section covers how to migrate your build to be compatible with Isolated Projects.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -172,26 +568,95 @@ while avoiding <<sharing_build_logic_between_subprojects.adoc#sec:convention_plu
...
+==== Use `IsolatedProject` instead of `Project`
 
-There is now support for obtaining an isolated view of a project as an   link:{javadocPath}/org/gradle/api/project/IsolatedProject.html[`IsolatedProject`] via link:{javadocPath}/org/gradle/api/Project.html#getIsolated--[`Project.getIsolated()`].
+You can get a simplified view of a project by calling link:{javadocPath}/org/gradle/api/Project.html#getIsolated--[`project.isolated`],
+which returns link:{javadocPath}/org/gradle/api/project/IsolatedProject.html[`IsolatedProject`] type.
+This works in all contexts and is especially useful when accessing an instance of another project,
+such as `project(":foo").isolated` or `rootProject.isolated`.
 
-The view exposes only those properties that are safe to access across project boundaries when running the build configuration phase in parallel (to be supported in a future release).
+The data provided by `IsolatedProject` is limited, but the view exposes only those properties that are safe to access across project boundaries.
```
lkasso:
```suggestion
The data provided by `IsolatedProject` is limited and exposes only properties that are safe to access across project boundaries.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -172,26 +568,95 @@ while avoiding <<sharing_build_logic_between_subprojects.adoc#sec:convention_plu
...
+<<configuration_cache.adoc#config_cache,Configuration Cache>> allows skipping the entire configuration phase when no build configuration inputs have changed.
+However, when any input has changed, the configuration has to be reexecuted in full.
+When executed, the configuration phase runs sequentially, configuring one project after another.
+
+Isolated Projects relies on the Configuration Cache infrastructure, and it also extends the set of constraints imposed by Configuration Cache.
+
+Configuration Cache is enabled automatically when Isolated Projects is enabled.
+You should make your build <<configuration_cache_enabling.adoc#config_cache:adoption,compatible with Configuration Cache>> first, before migrating to Isolated Projects.
+
+It is an error to enable Isolated Projects and explicitly disable Configuration Cache.
```
lkasso, line 607: important callout?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
```diff
@@ -172,26 +568,95 @@ while avoiding <<sharing_build_logic_between_subprojects.adoc#sec:convention_plu
...
+Isolated Projects extends the Configuration Cache, which enables intra-project task parallelism.
+Additionally, Isolated Projects constraints guarantee safe parallel tooling model building, which allows parallel model building to be enabled by default.
+Therefore, there is no benefit to enabling Parallel Execution alongside Isolated Projects.
+
+The Parallel Execution flag is _ignored_ when Isolated Projects is enabled.
+
+[[sec:ip_vs_cod]]
+=== Configuration on Demand
+
+<<configuration_on_demand.adoc#sec:configuration_on_demand,Configuration on Demand>> allows avoiding project configuration of projects that aren't explicitly required to execute requested tasks.
```
lkasso:
```suggestion
<<configuration_on_demand.adoc#sec:configuration_on_demand,Configuration on Demand>> lets Gradle skip configuring projects that aren't needed for the requested tasks.
```

lkasso (commented):
TLDR: This doc is amazing, but some sections are wordy and repetitive.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I have multiple reasons for adding this.

Firstly, there is no such feature state as "experimental" that is described on the Feature Lifecycle page. The wider team hasn't yet agreed to formalize this state, but I personally think that calling IP "experimental" is friendlier than saying "pre-alpha" as it was before. Therefore, some explanation is owed to the readers.

Secondly, we don't yet have a concrete public-facing plan on what needs to happen before IP actually becomes incubating. Therefore, providing at least some high-level reasoning is helpful. At the moment this refers to some known problems of "Gradle core" as well as some missing APIs.  

Given this context, would you still suggest we remove the sentence? If not, should we rephrase it?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I'll go with "Directories/files" then.

I think Octavia's concern is still valid. I'll just add a sub-point that explicitly says that `buildDir` and `layout` are mutable

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I agree that we should, but I'd rather do that in a follow-up. I think all constraints sections except the project-to-project need further explanation.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: Yes, and the "model" work is also overloaded. That's why I opted for clarifying what is meant by it here.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: It was an example that shows a simple state mutation. I agree that it might be more confusing than helping without more context.

I like your simplification. I'll use it as the first paragraph

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I like this phrasing. The only thing I'd add is that it's not only about the projects, but also builds

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I already linked to a relevant section in the line above. I think it's better than linking to just the properties, because the section explains the difference.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I think "support" is not the right word here, but I'm not sure "allow" is the best either.

In my book, "support" means whether Gradle or IDE shipped the capability of enabling/disabling something. Once something is supported, then the user can enable/disable the thing in their build, provided they use a new enough version of the tool. 

With "allow" here, I am going for a variant of "enable". I didn't use "enable" because what I'm describing in the section is a secondary effect. What you "enable" is Isolated Projects or Parallel Execution. Both of which _also_ give you parallel model building.

Can you suggest another verb that would work here? Or will "enable" cover it regardless of my concerns?

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: That's not what I meant. The projects aren't configured serially, they are configured in parallel.

What I wanted here is to be explicit that there is no "project configuration avoidance" yet, in simpler terms. This is partially related to Configurtion on Demand, but not fully. With CoD, you don't configure projects that aren't required for the requested tasks. IP might allow that in the future. But IP can also allow skipping project configuration _because it's cached_. So it technically "runs", but we just its result from the cache, which is not what CoD does.

I added a clarification and aligned the wording with that of `Performance when running tasks`.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: A qualified task is one where you explicitly specify the project, in which you want the task to run, e.g. `:check` for root project or `:foo:check` for the `foo` subprojct.

An unqualified task is a task that will be executed for all projects where it exists. So by running `./gradlew check`, the `check` task of every project is scheduled for execution.

---

We have some docs on this: https://docs.gradle.org/current/userguide/command_line_interface.html#executing_tasks_in_multi_project_builds

However, we don't introduce the concept of "unqualified" tasks, and instead we only describe the mechanism:

> run a task for all subprojects using a task selector that consists of only the task name

Btw, I don't think this is technically correct, because an unqualified task is resolved for in "allprojects", and not only "subprojects".

---

Maybe it would be good to introduce a term, "unqualified" or something else, so it can be referred to from other parts of the documentation, without having to explain what that is.

For now, I'll just link to that section of the docs.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
lkasso: Ok sounds good, let's rephrase it to:
```suggestion
The feature will enter incubation once Gradle's core has stronger support for this mode and more common use cases are covered.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I don't feel like bolding the whole sentence is what I want. I'll emphasize just the `sequential` instead.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
lkasso: Ok how about:

```suggestion
In the current implementation, all projects are always configured; there is no project configuration avoidance yet, whether by skipping unneeded projects or by using cached results.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
lkasso: How about:
```suggestion
However, both Gradle and the IDE need to be configured to allow it.
```

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
lkasso: Ok, sounds good to me.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
lkasso: PR coming!

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
lkasso: Sounds good.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: Not that comes to my mind. This refers to the topology of the build logic, included builds, project hierarchies within them and so on.

Ironically, we can't give more a precise prognosis because there is no isolation between builds and projects, and because all of them can be interconnected.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I agree. I was surprised to find out how underdocumented Tooling API is.

For now, I linked to the root of the Tooling API docs in the bullet point above. Once we have better docs for modeling building, it would be nice to link them here.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I reworded and simplified the sentence. Though, I didn't go with your suggestion.

Parallel project configuration already applies to both task-running and model-building. There will be a per-project model caching in the future, but there will also be per-project task graph caching equivalent for task-running scenarios.

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: I like the phrasing. Ideally, we'd also explain what an "unneeded project" is, but for now I think it's intuitive enough

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: Feel free, but I don't want to make it a condition for merging this PR, since I want to land this as part of 9.5.0 GA

platforms/documentation/docs/src/docs/userguide/reference/structuring-builds/isolated_projects.adoc
alllex: Thanks for providing a conrete suggestion on how to simplify!

alllex: @lkasso  Btw, I think the docs for the property itself are not very precise: https://docs.gradle.org/current/userguide/build_environment.html#gradle_properties_reference

> org.gradle.workers.max=(max # of worker processes)

It controls much more than the number of worker processes. E.g., it controls the number of tasks running in parallel. It also controls number of tooling models build in parallel, and so on. And with IP now, number of projects configured in parallel.

It would be nice to give the docs of that Gradle option a refresh.

---

Also, is it not possible to link to the docs of a specific Gradle option, like this one? I found that I can only link to a section, in which all properties are defined, which is not optimal

alllex:
@lkasso, I believed I address all of your feedback. For most of it, I applied your suggestions. For some, I explained the reasoning.

Thank you for providing very concrete phrasing of sentences and even entire paragraphs! I think the page is now much simpler to read.

alllex:
@bot-gradle test BD

alllex:
@bot-gradle test BD

Issue #34469, opened 2025-07-29
Investigate internal cross-project accesses


jvandort:
### Issue type

Other

### Problem description

There are a number of ways to access mutable project instances across Gradle projects within Gradle internal code:

[ModelContainer](https://github.com/gradle/gradle/blob/cb0c615fb8e3690971bb7f89ad80f58943360624/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/model/ModelContainer.java#L27) (which ProjectState implements as `ModelContainer<ProjectInternal>`) has a number of methods to access the wrapped model.

`fromMutableState` and `applyToMutableState` provide access to the mutable project while acquiring the project lock. However, these do not ensure the project has been configured.

[ProjectState.getMutableModel](https://github.com/gradle/gradle/blob/cb0c615fb8e3690971bb7f89ad80f58943360624/subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectState.java#L145) provides access to the mutable project while _not_ acquiring the project lock.

[ProjectRegistry](https://github.com/gradle/gradle/blob/c77df1f03636048365483d361f5d0b699c1b4a34/subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectRegistry.java#L24) provides access to mutable project instances. This should be replaced with ProjectStateRegistry

### Context (optional)

We need to go through all usages of these APIs to ensure they are being called in a safe manner. 

1. All APIs which do not acquire project locks should be removed. Usages should be migrated to APIs which acquire project locks or should be deprecated/removed. In some cases we might "know" that the project lock is already acquired. In these cases we should be operating on the mutable _Project_ instance itself, which was acquired from the point at which we acquired the lock. 
2. All APIs which acquire project locks should also ensure projects are configured. We may choose to add a wrapper around `ModelContainer` locking methods which also ensure the target build/project is configured.

If we can, we should strive to _remove_ as many cross project accesses as possible. We should only resort to configuring/locking if necessary. The more cross-project accesses we have, the more contention on locks. Part of this work should be evaluating whether we even _need_ to keep using these cross-project APIs in some places.

octylFractal:
More work related to this:
- Clean up `CrossProjectModelAccess`, e.g. https://github.com/gradle/gradle/pull/37659#discussion_r3154425850 and removing `access` using `ProjectInternal`.
  - https://github.com/gradle/gradle/pull/37794
[...]

PR #37493, opened 2026-04-09
Wrap simple usages of mutable project

octylFractal:
Related to #34469 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/AllProjectsLock.java
+++ b/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/AllProjectsLock.java
@@ -16,14 +16,18 @@
 
 package org.gradle.internal.resources;
 
+import java.util.function.Predicate;
+
 class AllProjectsLock extends ExclusiveAccessResourceLock {
-    public AllProjectsLock(String displayName, ResourceLockCoordinationService coordinationService, ResourceLockContainer owner) {
+    private final Predicate<AllProjectsLock> anyProjectLockHeldByOtherThreads;
+
+    public AllProjectsLock(String displayName, ResourceLockCoordinationService coordinationService, ResourceLockContainer owner, Predicate<AllProjectsLock> anyProjectLockHeldByOtherThreads) {
         super(displayName, coordinationService, owner);
+        this.anyProjectLockHeldByOtherThreads = anyProjectLockHeldByOtherThreads;
     }
 
     @Override
     protected boolean canAcquire() {
-        // TODO - should block while some other thread holds a project lock
-        return true;
+        return !anyProjectLockHeldByOtherThreads.test(this);
     }
 }
--- a/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/AllProjectsLock.java
+++ b/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/AllProjectsLock.java
@@ -16,18 +16,27 @@
 
 package org.gradle.internal.resources;
 
-import java.util.function.Predicate;
+import java.util.function.Supplier;
 
 class AllProjectsLock extends ExclusiveAccessResourceLock {
-    private final Predicate<AllProjectsLock> anyProjectLockHeldByOtherThreads;
+    private final Supplier<Iterable<ProjectLock>> cachedProjectLocks;
 
-    public AllProjectsLock(String displayName, ResourceLockCoordinationService coordinationService, ResourceLockContainer owner, Predicate<AllProjectsLock> anyProjectLockHeldByOtherThreads) {
+    public AllProjectsLock(String displayName, ResourceLockCoordinationService coordinationService, ResourceLockContainer owner, Supplier<Iterable<ProjectLock>> cachedProjectLocks) {
         super(displayName, coordinationService, owner);
-        this.anyProjectLockHeldByOtherThreads = anyProjectLockHeldByOtherThreads;
+        this.cachedProjectLocks = cachedProjectLocks;
     }
 
     @Override
     protected boolean canAcquire() {
-        return !anyProjectLockHeldByOtherThreads.test(this);
+        for (ProjectLock projectLock : cachedProjectLocks.get()) {
+            if (projectLock.getAllProjectsLock() != this) {
+                // We only care about locks that are associated with this "all projects" lock.
+                continue;
+            }
+            if (projectLock.isLocked() && !projectLock.isLockedByCurrentThread()) {
+                return false;
+            }
+        }
+        return true;
     }
 }
--- a/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLockRegistry.java
+++ b/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLockRegistry.java
@@ -25,7 +25,26 @@ public class ProjectLockRegistry extends AbstractResourceLockRegistry<Path, Proj
     public ProjectLockRegistry(ResourceLockCoordinationService coordinationService, boolean parallelEnabled) {
         super(coordinationService);
         this.parallelEnabled = parallelEnabled;
-        allProjectsLocks = new LockCache<Path, AllProjectsLock>(coordinationService, this);
+        allProjectsLocks = new LockCache<>(coordinationService, this);
+    }
+
+    /**
+     * Checks if any project lock is currently held that is associated with the given "all projects" lock, and is held by another thread.
+     * This is effectively a proxy for checking if any of the project locks have the same {@code buildIdentityPath} as the given "all projects" lock.
+     *
+     * @param relevantAllProjectsLock the "all projects" lock for which to check if any project locks are currently held by another thread
+     * @return {@code true} if any project lock is currently held by another thread that is associated with the given "all projects" lock, {@code false} otherwise
+     */
+    private boolean isAnyProjectLockHeldByAnotherThread(AllProjectsLock relevantAllProjectsLock) {
+        for (ProjectLock projectLock : getAllResourceLocks()) {
+            if (projectLock.getAllProjectsLock() != relevantAllProjectsLock) {
+                continue;
+            }
+            if (projectLock.isLocked() && !projectLock.isLockedByCurrentThread()) {
+                return true;
+            }
+        }
+        return false;
     }
 
     public boolean getAllowsParallelExecution() {
@@ -33,12 +52,9 @@ public class ProjectLockRegistry extends AbstractResourceLockRegistry<Path, Proj
     }
 
     public ResourceLock getAllProjectsLock(final Path buildIdentityPath) {
-        return allProjectsLocks.getOrRegisterResourceLock(buildIdentityPath, new ResourceLockProducer<Path, AllProjectsLock>() {
-            @Override
-            public AllProjectsLock create(Path key, ResourceLockCoordinationService coordinationService, ResourceLockContainer owner) {
-                String displayName = "All projects of " + buildIdentityPath;
-                return new AllProjectsLock(displayName, coordinationService, owner);
-            }
+        return allProjectsLocks.getOrRegisterResourceLock(buildIdentityPath, (key, coordinationService, owner) -> {
+            String displayName = "All projects of " + buildIdentityPath;
+            return new AllProjectsLock(displayName, coordinationService, owner, this::isAnyProjectLockHeldByAnotherThread);
         });
     }
 
@@ -47,12 +63,9 @@ public class ProjectLockRegistry extends AbstractResourceLockRegistry<Path, Proj
     }
 
     private ProjectLock doGetResourceLock(final Path buildIdentityPath, final Path lockPath) {
-        return getOrRegisterResourceLock(lockPath, new ResourceLockProducer<Path, ProjectLock>() {
-            @Override
-            public ProjectLock create(Path projectPath, ResourceLockCoordinationService coordinationService, ResourceLockContainer owner) {
-                String displayName = parallelEnabled ? "state of project " + lockPath : "state of build " + lockPath;
-                return new ProjectLock(displayName, coordinationService, owner, getAllProjectsLock(buildIdentityPath));
-            }
+        return getOrRegisterResourceLock(lockPath, (projectPath, coordinationService, owner) -> {
+            String displayName = parallelEnabled ? "state of project " + lockPath : "state of build " + lockPath;
+            return new ProjectLock(displayName, coordinationService, owner, getAllProjectsLock(buildIdentityPath));
         });
     }
 }
--- a/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLockRegistry.java
+++ b/platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLockRegistry.java
@@ -28,33 +28,19 @@ public class ProjectLockRegistry extends AbstractResourceLockRegistry<Path, Proj
         allProjectsLocks = new LockCache<>(coordinationService, this);
     }
 
-    /**
-     * Checks if any project lock is currently held that is associated with the given "all projects" lock, and is held by another thread.
-     * This is effectively a proxy for checking if any of the project locks have the same {@code buildIdentityPath} as the given "all projects" lock.
-     *
-     * @param relevantAllProjectsLock the "all projects" lock for which to check if any project locks are currently held by another thread
-     * @return {@code true} if any project lock is currently held by another thread that is associated with the given "all projects" lock, {@code false} otherwise
-     */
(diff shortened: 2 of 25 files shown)


jvandort:
I think this is fine as long as it doesn't affect performance. 

It seems we're looping over projects and potentially acquiring the same locks for the same projects more than once.

Is this a scenario where we would want to grab the all projects lock before doing the loop? Or, can we loop over the projects once and perform the `applyProjectStates` and `collectProjects` within the same project lock acquisition? 

I imagine grabbing the locks multiple times should not really be an issue, as long as there is no other contention, so I don't think its a big deal if its not easy to combine these

octylFractal:
The two are used in different sides of the protocol, apply is on read and collect is on write. So I don't think there's much of double-acquisition going on, but perhaps it would be nicer to just pull the all projects lock for each of these.

octylFractal:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/initialization/ProjectPropertySettingBuildLoader.java
```diff
@@ -44,7 +45,10 @@ public ProjectPropertySettingBuildLoader(
     @Override
     public void load(SettingsInternal settings, GradleInternal gradle) {
         buildLoader.load(settings, gradle);
-        setProjectProperties(gradle.getOwner().getProjects().getRootProject());
+        BuildProjectRegistry projectRegistry = gradle.getOwner().getProjects();
+        projectRegistry.withMutableStateOfAllProjects(() ->
```
jvandort: How come we grab the all project locks if `addPropertiesToProject` acquires the lock of each individual project? Is it a matter of efficiency since we know we will be visiting all the projects so we avoid grabbing and dropping a bunch of locks in succession?

platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/idea/internal/IdeaIsolatedProjectsWorkarounds.java
```diff
@@ -36,8 +36,9 @@ class IdeaIsolatedProjectsWorkarounds {
      * The check is done bypassing the Isolated Projects validations.
```
jvandort, line 36: We no longer "bypass" IP violations, but safely acquire the lock for the project to determine if it has a project applied. 

I am curious about the use case here, why would someone without a project lock need to know that some other project has a plugin applied? 

Looking at where this is used in `IdeaScalaConfigurer`, this looks like a job for the all project locks. We are iterating over all projects and seeing if any have both `idea` and `scala` and throwing an exception if so. 

This whole pattern seems a bit fishy. Why are idea and scala not IP compatible thogether? Is this something we should fix separately? 

In any case, I think `IdeaIsolatedProjectsWorkarounds` seem a bit suspicious as a concept to me

subprojects/core/src/main/java/org/gradle/initialization/ProjectPropertySettingBuildLoader.java
jvandort: So I'm seeing this pattern elsewhere, and I suppose its since there is no other way to get the actual `Project` instance without resorting to `getMutableModel`. I wonder if `withMutableStateOfAllProjects` could pass a parameter to the lambda where you could lookup `Project` instances for the duration of the lambda.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLockRegistry.java
```diff
@@ -25,20 +25,36 @@ public class ProjectLockRegistry extends AbstractResourceLockRegistry<Path, Proj
...
+    /**
+     * Checks if any project lock is currently held that is associated with the given "all projects" lock, and is held by another thread.
+     * This is effectively a proxy for checking if any of the project locks have the same {@code buildIdentityPath} as the given "all projects" lock.
+     *
+     * @param relevantAllProjectsLock the "all projects" lock for which to check if any project locks are currently held by another thread
+     * @return {@code true} if any project lock is currently held by another thread that is associated with the given "all projects" lock, {@code false} otherwise
+     */
+    private boolean isAnyProjectLockHeldByAnotherThread(AllProjectsLock relevantAllProjectsLock) {
+        for (ProjectLock projectLock : getAllResourceLocks()) {
+            if (projectLock.getAllProjectsLock() != relevantAllProjectsLock) {
```
jvandort: Can we pass the build identity path into both the project lock and the all projects lock so we don't have to rely on the object identity of the all projects lock to "proxy" the actual comparison we want to make?

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLockRegistry.java
```diff
@@ -25,20 +25,36 @@ public class ProjectLockRegistry extends AbstractResourceLockRegistry<Path, Proj
...
+    }
+
+    /**
+     * Checks if any project lock is currently held that is associated with the given "all projects" lock, and is held by another thread.
+     * This is effectively a proxy for checking if any of the project locks have the same {@code buildIdentityPath} as the given "all projects" lock.
+     *
+     * @param relevantAllProjectsLock the "all projects" lock for which to check if any project locks are currently held by another thread
+     * @return {@code true} if any project lock is currently held by another thread that is associated with the given "all projects" lock, {@code false} otherwise
+     */
+    private boolean isAnyProjectLockHeldByAnotherThread(AllProjectsLock relevantAllProjectsLock) {
```
jvandort: IMO this feels like it should be a function inside the all projects lock, and the all projects lock should instead accept a function as a constructor parameter which provides all other single-project locks

subprojects/core/src/test/groovy/org/gradle/api/internal/project/DefaultProjectStateRegistryTest.groovy
```diff
@@ -311,6 +311,65 @@ class DefaultProjectStateRegistryTest extends ConcurrentSpec {
...
+                    assert state1.hasMutableState()
+                    assert !state2.hasMutableState()
+                }
+            }
+        }
+
+        then:
+        noExceptionThrown()
+    }
+
```
jvandort, line 372: Some other tests which might be interesting:
1. `can lock project while current thread holds all project lock`
2. `can lock project in one build while another thread holds all projects lock in another build`
3. `can lock all projects lock in one build while another thread holds a projct lock in another build`

jvandort (approved):
No major comments -- though I'd mostly like to see us not rely on all projects object identity and instead compare build identity paths

subprojects/core/src/main/java/org/gradle/initialization/ProjectPropertySettingBuildLoader.java
octylFractal: Yes, I did it for efficiency since we won't bother pulling the lock at all if we hold the all projects lock. Having a lambda to allow converting `ProjectState` -> `Project` would be nice, I'll look at it.

platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/idea/internal/IdeaIsolatedProjectsWorkarounds.java
octylFractal: I'll talk about it with @alllex.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLockRegistry.java
octylFractal: Yeah, that would work I think.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLockRegistry.java
octylFractal: Sure.

octylFractal: After thinking about it some more, especially with the refactor I'll push up for the other comment, I think it makes more sense to just directly compare the object identity here. Since the project lock references a specific all projects lock, we also want to only check those project locks that are directly related to the all projects lock, regardless of build identity, in order to be consistent here.

jvandort: To me then this defines a single project's all projects lock to instead be _some_ lock (not necessarily all projects in any particular build) that grants access to the single project in question. It just so happens that we construct this lock such that the set of projects which reference it are all the projects in a given build. 

I guess im fine with using object identity but it weakens the justification for calling it an "all projects lock"

octylFractal: This is consistent with how the `ProjectLock` checks for the `AllProjectsLock`, so I think the definition is unchanged. If we want to change it I think we should adjust both this and the `ProjectLock`, because technically we can currently have multiple different instances of the `AllProjectsLock` when the `LockCache` cleans up unused ones due to GC, so there is already some kind of possibility that a change to `LockCache` could in fact construct `ProjectLock`s that use different `AllProjectsLock`s for the same build.

jvandort: > could in fact construct ProjectLocks that use different AllProjectsLocks for the same build.

This sounds a bit scary. You're saying someone could obtain an all projects lock that doesn't necessarily prevent other projects in that build from being acquired while the all projects lock is locked? 

If we continue with this identity based approach I think a build should allocate an all projects lock early and retain it for the entire build's lifetime. These should not be expensive to keep around, right?

jvandort (approved):

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLockRegistry.java
octylFractal: If there was a bad change to `LockCache`, yes. We discard locks when not referenced using weak values currently, but if it was some other mechanism and it had a bug, perhaps we would have issues. It won't happen currently.

I'm in favor of just allocating one early and retaining it, as I think there's actually a lot more overhead for keeping the whole `LockCache` around. But that can be done separately.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/AllProjectsLock.java
```diff
@@ -16,14 +16,27 @@
...
         super(displayName, coordinationService, owner);
+        this.cachedProjectLocks = cachedProjectLocks;
     }
 
     @Override
     protected boolean canAcquire() {
-        // TODO - should block while some other thread holds a project lock
+        for (ProjectLock projectLock : cachedProjectLocks.get()) {
+            if (projectLock.getAllProjectsLock() != this) {
+                // We only care about locks that are associated with this "all projects" lock.
```
alllex, line 33: ❔ Why? Is the implication that we might have gotten project locks from another build? If so, I'd be explicit about it.

Also, if it's projects-belonging-to-the-same build question, I wonder if it would be better to actually carry the build identity in the locks to be able to check their equality, rather than doing so implicitly through the all-projects-lock equality.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/AllProjectsLock.java
```diff
@@ -16,14 +16,27 @@
 
 package org.gradle.internal.resources;
 
+import java.util.function.Supplier;
+
 class AllProjectsLock extends ExclusiveAccessResourceLock {
-    public AllProjectsLock(String displayName, ResourceLockCoordinationService coordinationService, ResourceLockContainer owner) {
+    private final Supplier<Iterable<ProjectLock>> cachedProjectLocks;
```
alllex, line 22: 🤔 I find it somewhat concerning that a rather fragile concept of weakly cached locks is travelling in a very generic container far away from its origin. I think it would have been better if it stayed at least inside the class that owns it. We can do that by moving the logic of `canAcquire` into the `ProjectLockRegistry` and only call it here. Alternatively, we can move the entire `AllProjectsLock` type in there and make it private to the class.

platforms/core-runtime/build-profile/build.gradle.kts
```diff
@@ -19,7 +19,6 @@ dependencies {
     implementation(projects.logging)
     implementation(projects.loggingApi)
     implementation(projects.reportRendering)
-    implementation(projects.serviceLookup)
```
alllex, line 22: 🎉

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
```diff
@@ -243,18 +245,49 @@ public Set<? extends ProjectState> getAllProjects() {
         }
 
         @Override
-        public void withMutableStateOfAllProjects(Runnable runnable) {
-            withMutableStateOfAllProjects(Factories.toFactory(runnable));
+        public <T extends @Nullable Object> T fromMutableStateOfAllProjects(Function<AllProjectsAccess, T> factory) {
+            ResourceLock allProjectsLock = workerLeaseService.getAllProjectsLock(owner.getIdentityPath());
+            Collection<? extends ResourceLock> locks = workerLeaseService.getCurrentProjectLocks();
+            return workerLeaseService.withReplacedLocks(locks, allProjectsLock, () -> {
+                AllProjectsAccessImpl fetcher = new AllProjectsAccessImpl(owner);
```
alllex: 💅 Align the variable name with the type

subprojects/core/src/main/java/org/gradle/internal/build/AllProjectsAccess.java
```diff
@@ -0,0 +1,30 @@
...
+import org.gradle.api.internal.project.ProjectState;
+
+/**
+ * Given out when taking the all projects lock to allow retrieving the mutable state of
+ * {@link org.gradle.api.internal.project.ProjectState} without extra locking checks.
+ * Instances must never live longer than the lambda, and calls on this interface will
+ * fail after the lock has been released.
+ */
+public interface AllProjectsAccess {
+    ProjectInternal getMutableModel(ProjectState project);
```
alllex, line 44: 👍 This is a really nice idea.

💅 We should at least add a note to `ProjectState.getMutableModel` that it's not the right way of obtaining the model. Otherwise, it's very non-trivial to understand why shouldn't I simply call `projectState.getMutableModel` to get the same thing.

platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/idea/internal/IdeaIsolatedProjectsWorkarounds.java
alllex: For some reason, the Scala plugin treats the root project very specially, and also aggregates the info on all subprojects where Scala is applied, at configuration time (after evaluate).

I'd keep this as a workaround, despite the appearance of the safe access. IIUC, one thing this doesn't do is making sure the projects are configured before checking for plugins. That's also for another day.

alllex (approved):
LGTM!

I left some suggestions

subprojects/core/src/main/java/org/gradle/internal/build/AllProjectsAccess.java
octylFractal: I think the note belongs on this method, but yeah.

alllex: The note on this method is not discoverable, if you just have an instance of `ProjectState`

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLockRegistry.java
alllex: I missed this comment and wrote a similar one to Justin's: https://github.com/gradle/gradle/pull/37493#discussion_r3113224938

I think having a pre-allocated AllProjectsLock per build makes a lot of sense. Then it can also be tied to the identity of the build, which can be used for comparing them. In the same sense, the ProjectLock should probably be tied to the identity of a project.

If this refactoring also allows us to get rid of the `LockCache`, even better.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/AllProjectsLock.java
alllex: Ah, this is already discussed in https://github.com/gradle/gradle/pull/37493#discussion_r3090880006

PR #37499, opened 2026-04-10
Extract classpath packages from core into new module

ghale:
Core and core-api are large monoliths that contain api and implementation classes that cross multiple platforms. Having these classes lumped together encourages tangling and circular relationships between platforms and can lead to relationships that do not comply with our platform rules. It also leads to extremely coarse dependencies that drag in much more than is necessary and can cause classes to be misplaced solely in order to workaround the limitations of these coarse relationships. Breaking up core/core-api into smaller projects helps to prevent tangling and allows for narrower, more granular dependency relationships. It also clears the path for us to start enforcing platform relationships in a way that we can't easily do so today.

Move classpath manipulation, instrumentation, and bytecode transformation infrastructure from subprojects/core into a new platforms/core-runtime/classpath module.

Production code moved:
- org.gradle.api.internal.classpath (3 classes)
- org.gradle.internal.classpath (72 classes including subpackages)
- ClasspathFingerprinter interface
- ConventionInterfaceGenerator
- JarUtil (moved to base-services)

All associated tests, test fixtures, and test interceptors move with the production code. 

CommonTypes is relocated from transforms to types subpackage to break a nasty package cycle that required "org.gradle.**" as an accepted cycle.  There are other cycles in the package, but they are all within the classpath subpackages, so resolving them is left as future work.

Downstream projects updated to declare direct dependency on the new classpath module where needed.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ghale:
@bot-gradle test this

bamboo (approved):
Looks good 👍

Issue #37365, opened 2026-03-31
In DCL schema building, omit issues for types that are only used from other problematic types


h0tk3y:
### Expected Behavior

If a type is only referenced from other types that are illegal in a Declarative definition, issues arising from it should be deprioritized. If there are issues in types used from other valid types, those should be reported first. Then there might be a mention of other issues in types used only from non-Declarative types.

### Current Behavior (optional)

If the DCL schema accidentally includes a type that is not meant to be declarative and has numerous problems, as well as references to numerous other problematic types, then the error report is unproportionally long, and it is hard to find out where the problem is coming from. 

For example, using an `ObjectFactory` somewhere in the definition leads to the following (huge!) problem report:

<details><summary>Details</summary>
<p>

```
Failures in building the schema:
    Type 'org.gradle.api.provider.Provider' is a hidden type and cannot be directly used.
      Appears as hidden:
        - in the supertypes of 'org.gradle.api.provider.Property'
        - in the supertypes of 'org.gradle.api.provider.SetProperty'
        - type 'org.gradle.api.provider.Provider' is annotated as hidden
        - in the supertypes of 'org.gradle.api.provider.ListProperty'
        - in the supertypes of 'org.gradle.api.provider.MapProperty'
        - in the supertypes of 'org.gradle.api.file.DirectoryProperty'
        - in the supertypes of 'org.gradle.api.file.RegularFileProperty'
      Illegal usages:
        - referenced from member 'fun org.gradle.api.DomainObjectCollection<T>.addAllLater(org.gradle.api.provider.Provider<out kotlin.collections.(Mutable)Iterable<T!>!>!): kotlin.Unit'
        - referenced from member 'fun org.gradle.api.DomainObjectCollection<T>.addLater(org.gradle.api.provider.Provider<out T!>!): kotlin.Unit'
    Type 'org.gradle.api.provider.SetProperty' is a hidden type and cannot be directly used.
      Appears as hidden:
        - type 'org.gradle.api.provider.SetProperty' is annotated as hidden
      Illegal usages:
        - referenced from member 'fun org.gradle.api.model.ObjectFactory.setProperty(java.lang.Class<T!>!): org.gradle.api.provider.SetProperty<T!>!'
    Type 'org.gradle.api.file.SourceDirectorySet' is a hidden type and cannot be directly used.
      Appears as hidden:
        - type 'org.gradle.api.file.SourceDirectorySet' is annotated as hidden
      Illegal usages:
[...]

PR #37517, opened 2026-04-13
Prune DCL schema building failures

h0tk3y:
Report only the DCL schema building failures occurring in types that are used
from non-erroneous types. If a type is only used in other erroneous types,
omit failures in it, but report a summary that says that such failures are also
present.

Fixes:
* https://github.com/gradle/gradle/issues/37365

### DIFF
--- a/platforms/core-configuration/declarative-dsl-core/src/main/kotlin/org/gradle/internal/declarativedsl/schemaBuilder/DataSchemaBuilder.kt
+++ b/platforms/core-configuration/declarative-dsl-core/src/main/kotlin/org/gradle/internal/declarativedsl/schemaBuilder/DataSchemaBuilder.kt
@@ -50,6 +50,7 @@ import org.gradle.internal.declarativedsl.schemaBuilder.TypeDiscovery.Discovered
 import org.gradle.internal.declarativedsl.schemaBuilder.TypeDiscovery.DiscoveredClass.DiscoveryTag.ProjectFeatureDefinition
 import org.gradle.internal.declarativedsl.schemaBuilder.TypeDiscovery.DiscoveredClass.DiscoveryTag.Supertype
 import java.util.concurrent.atomic.AtomicLong
+import kotlin.collections.mutableSetOf
 import kotlin.reflect.KCallable
 import kotlin.reflect.KClass
 import kotlin.reflect.KFunction
@@ -65,8 +66,8 @@ import kotlin.reflect.typeOf
 interface SchemaBuildingHost {
     val topLevelReceiverClass: KClass<*>
 
-    val typeFailures: Iterable<SchemaResult.Failure>
-    fun recordTypeFailure(failure: SchemaResult.Failure)
+    val typeFailures: Map<FqName, Set<SchemaResult.Failure>>
+    fun recordTypeDiscoveryFailure(inTypeByName: FqName, failure: SchemaResult.Failure)
 
     fun recordClaimedMember(kClass: KClass<*>, member: SupportedCallable)
     fun recordMemberWithFailure(kClass: KClass<*>, member: SupportedCallable, failure: SchemaResult.Failure)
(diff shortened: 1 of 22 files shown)


h0tk3y:
@bot-gradle test RFN

platforms/core-configuration/declarative-dsl-core/src/main/kotlin/org/gradle/internal/declarativedsl/schemaBuilder/DataSchemaBuilder.kt
```diff
@@ -161,6 +162,8 @@ object SchemaBuildingTags {
                 featureNames.joinToString { "'$it'" } + if (pluginId != null) " (plugin '${pluginId}')" else ""
             }
     )
+
+    fun erroneousTypes(typeNames: Set<String>) = TagContextElement("types that are only used in other erroneous types: ${typeNames.joinToString(limit = 3) { "'$it'" }}")
```
eskatos, line 166: ❓ Shouldn't this somehow convey that only the first three of them are displayed and that there might be others?

platforms/core-configuration/declarative-dsl-core/src/main/kotlin/org/gradle/internal/declarativedsl/schemaBuilder/SchemaResult.kt
```diff
@@ -208,20 +213,21 @@ object SchemaFailureMessageFormatter {
                 is UnsafeInjectProperty -> ": injected service property"
                 is UnsafeJavaBeanProperty -> ": unsafe property"
                 is UnsafeNonPureFunction -> ": function relying on side effects or custom implementation"
-                is UnsafeBecauseHasHiddenMembers -> ": hidden member${if (cause.memberNames.size > 1) "s" else ""} ${cause.memberNames.joinToString(limit = 3) { "'$it'" }}"
-                is UnsafeBecauseHasNonPublicMembers -> ": non-public member${if (cause.memberNames.size > 1) "s" else ""} ${cause.memberNames.joinToString(limit = 3) { "'$it'" }}"
+                is UnsafeBecauseHasHiddenMembers -> ": hidden member${if (cause.memberNames.size > 1) "s" else ""} ${cause.memberNames.sorted().joinToString(limit = 3) { "'$it'" }}"
+                is UnsafeBecauseHasNonPublicMembers -> ": non-public member${if (cause.memberNames.size > 1) "s" else ""} ${cause.memberNames.sorted().joinToString(limit = 3) { "'$it'" }}"
```
eskatos, line 217: ❓ Shouldn't these somehow convey that only the first three of them are displayed and that there might be others?

platforms/core-configuration/declarative-dsl-core/src/test/kotlin/org/gradle/internal/declarativedsl/schemaBuidler/SchemaSupertypeMemberVisibilityTest.kt
```diff
@@ -60,10 +60,10 @@ class SchemaSupertypeMemberVisibilityTest {
             |Multiple failures in building the declarative schema:
             |
             |* Conflicting annotations: @VisibleInDefinition and @HiddenInDefinition are present
-            |  in class 'org.gradle.internal.declarativedsl.schemaBuidler.VisibleHidden'
-            |
-            |* Conflicting annotations: @VisibleInDefinition and @HiddenInDefinition are present
             |  in class 'org.gradle.internal.declarativedsl.schemaBuidler.AnotherVisibleHiddenSub'
+            |
+            |* More failures that are not reported now
+            |  in types that are only used in other erroneous types: 'org.gradle.internal.declarativedsl.schemaBuidler.VisibleHidden'
```
eskatos, line 66: ❓ I wonder if we should even report these to the user. It sounds more like debugging information that could be just logged. WDYT?

platforms/core-configuration/declarative-dsl-provider/src/main/kotlin/org/gradle/internal/declarativedsl/common/MinimalSchemaBuildingComponent.kt
```diff
@@ -54,7 +55,7 @@ class MinimalSchemaBuildingComponent : AnalysisSchemaComponent {
     override fun typeDiscovery(): List<TypeDiscovery> = listOf(
         FunctionLambdaTypeDiscovery(gradleConfigureLambdas),
         FunctionParameterTypeDiscovery(),
-        FunctionReturnTypeDiscovery()
+        FunctionReturnTypeDiscovery { it != Property::class } // not interested in bringing `Property` to the schema from functions returning it, it only produces noise in failure reports
```
eskatos, line 58: ❓ Is `Property` the only type that is useful to filter here? What about e.g. `ListProperty`?

platforms/core-configuration/declarative-dsl-core/src/main/kotlin/org/gradle/internal/declarativedsl/schemaBuilder/DataSchemaBuilder.kt
h0tk3y: `limit = 3` produces an ellipsis as in `foo, bar, baz, ...`.

platforms/core-configuration/declarative-dsl-core/src/test/kotlin/org/gradle/internal/declarativedsl/schemaBuidler/SchemaSupertypeMemberVisibilityTest.kt
h0tk3y: I think it makes sense to report this. Users who are experimenting with the schema and expecting a failure in some type would be surprised if that failure is silently ignored (because of other failures). It makes sense to me to report this so users are aware that other failures are present. It will also make them less frustrated when they discover those failures after fixing the immediately reported ones. :)

PR #37524, opened 2026-04-14
Prevent library.jansi.path from leaking across JVMs

### LINKED ISSUE #37769: Gradle TAPI Client library sets library.jansi.path system property
### Current Behavior

Gradle Tooling API client library sets system property `library.jansi.path` value pointing to native components of jansi 1.18 and never clears it.

### Expected Behavior

No system properties are adjusted by a library in the runtime

### Context (optional)

Our integration tests in IntelliJ IDEA stumbled upon this problem. When IDEA's process uses another, newer version 2.4.1 of `jansi` e.g. in the next test, the system property still points to an old version of the native component. This results in the JVM native crash.

PFA the project. Execute `gradlew run` , observe the system property value 

[tapi-test.zip](https://github.com/user-attachments/files/27280569/tapi-test.zip)

in the output:
```
system property library.jansi.path = 
```
It should be null, but points somewhere

### Gradle version

9.3

### Build scan URL (optional)

_No response_

_No response_

reinsch82:
## Summary

When an IDE (e.g. IntelliJ IDEA) uses Tooling API 9.x to connect to an old Gradle project (e.g. Gradle 7.6), the cross-version compatibility layer loads the old provider's JARs into the IDE's JVM. The old provider's `NativeServices` / `JansiBootPathConfigurer` sets the JVM-global system property `library.jansi.path` to point at Gradle's pre-extracted jansi **1.18** native directory. After `configure()` returns, this stale value remains set. When the IDE's own jansi **2.x** later reads the property, it attempts to load the 1.18 native library with 2.x JNI bindings, causing a **JVM crash**.

### Root cause

`JansiBootPathConfigurer.configure()` calls `System.setProperty("library.jansi.path", ...)` to tell jansi where to find the pre-extracted native library (avoiding extraction to a potentially noexec-mounted tmpdir). This property is JVM-global and was never cleaned up after use.

### Fix

**Provider side (`JansiBootPathConfigurer`):** The property is now set temporarily — save the previous value, set the path, eagerly call `JansiLoader.initialize()` to load the native library, then restore the previous value in a `finally` block. Once jansi's internal `loaded` flag is set, the property is no longer consulted, so clearing it is safe. This prevents future Gradle versions from leaking the property when loaded as TAPI providers.

**Consumer side (`DefaultToolingImplementationLoader`):** Save and restore `library.jansi.path` around the provider's `configure()` call. This is the defense-in-depth fix that protects against **old** provider versions (e.g. Gradle 7.6) whose code cannot be changed.

## Test plan

- [ ] `JansiBootPathConfigurerTest` — verifies the property is not leaked after `configure()` and that a pre-existing value is restored
- [ ] `DefaultToolingImplementationLoaderTest` — two new tests using a mock connection that simulates an old provider setting the property; verifies cleanup when no prior value exists and restoration when one does
- [ ] Existing `JansiStorageLocatorTest`, `JansiLibraryFactoryIntegrationTest`, and `JansiEndUserIntegrationTest` still pass

## Bug Fixes
* Enhanced Jansi library path management to properly restore system state after initialization, ensuring the native library path property is only temporarily set during configuration and restored to its previous state afterward.

## Tests
* Added comprehensive test coverage for system property cleanup in Jansi initialization and tooling connection creation workflows.

### DIFF
--- a/platforms/core-runtime/native/src/main/java/org/gradle/internal/nativeintegration/jansi/JansiBootPathConfigurer.java
+++ b/platforms/core-runtime/native/src/main/java/org/gradle/internal/nativeintegration/jansi/JansiBootPathConfigurer.java
@@ -17,6 +17,7 @@
 package org.gradle.internal.nativeintegration.jansi;
 
 import org.apache.commons.io.IOUtils;
+import org.fusesource.jansi.internal.JansiLoader;
 import org.gradle.internal.IoActions;
 import org.gradle.internal.nativeintegration.NativeIntegrationException;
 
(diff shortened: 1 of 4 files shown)


reinsch82:
@coderabbitai review

PR #37526, opened 2026-04-14
Fix AccessTrackingProperties to return a proper clone

### LINKED ISSUE #17344: org.gradle.internal.classpath.Instrumented.DecoratingProperties.clone() does not return an independent instance
A call to `org.gradle.internal.classpath.Instrumented.DecoratingProperties.clone()` (used for example to get an isolated copy of the JVM system properties) returns a `DecoratingProperties` instance that uses the same delegate. Calls to `setProperty(...)` will change the original (and all other "cloned") instances.

### Expected Behavior
Cloning should also clone the delegate `Properties` instance.

### Current Behavior
Cloned Properties-Instances will be "coupled" through their delegate. Changes in one cloned instance will be visible in all other cloned instances.

### Context
https://github.com/OpenAPITools/openapi-generator uses clone() to create an independent copy of the JVM system properties. This works well outside of Gradle, but fails when the generator is used in Gradle.

### Steps to Reproduce 
The following code should print `World`, but it prints `Bug` when Properties instances are instrumented by Gradle:

```
System.setProperty("Hello", "World");
Properties copy = (Properties) System.getProperties().clone();
copy.setProperty("Hello", "Bug");
System.out.println(System.getProperty("Hello"));
```

Workaround: Create a new Properties instance and use `putAll(...)` to add the original property values.

### Your Environment
Gradle 6.7.1 (also happens with 7.0.2) on Windows 10

ov7a:
Fixes #17344

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


blindpirate (approved):

platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java
```diff
@@ -483,6 +483,12 @@ public void list(PrintWriter out) {
         delegate.list(out);
     }
 
+    @Override
+    public synchronized Object clone() {
+        Properties clonedDelegate = (Properties) delegate.clone();
+        return new AccessTrackingProperties(clonedDelegate, listener);
```
alllex: 💭 I don't think it's directly obvious that we continue tracking instead of reporting aggregate access. This is a decision about an optimization: not immediately reporting all properties, when somebody calls `System.getProperties().clone()`. Maybe an inline comment can help clarify this?

The current class seems a bit like prepature abstraction, given that we only ever use it for tracking JVM system properties, and those have very special meaning for the running JVM, as opposed to a random `Properties` instance. But I don't propose to change that in the current PR.

Though I do believe we should add integration tests that demonstrate the behavior with **`System.getProperties().clone()`**, and not just unit tests, even though those are good to have. The integration tests can also reference the issue, while unit tests aren't necessarily appropriate for this.

platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java
```diff
@@ -483,6 +483,12 @@ public void list(PrintWriter out) {
         delegate.list(out);
     }
 
+    @Override
+    public synchronized Object clone() {
```
alllex, line 488: 💅 We suppessed a warning in Checkstyle, but I still get one in the IDE. It's probably good to suppress it as well to solidify the intent.

alllex (changes requested):

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredBuildInputsIntegrationTest.groovy
```diff
@@ -955,6 +954,46 @@ internal property sun.java.command changes.
         outputContains("Execution: build-logic-value")
     }
 
+    @Issue("https://github.com/gradle/gradle/issues/17344")
+    def "cloned system properties still track individual access"() {
```
alllex: We should also validate that changing cloned system properties doesn't change the global shared system properties, as plugins observe them. That's the original problem indicated by the issue, IIUC.

alllex (approved):
LGTM

Though, we should check that cloned system property modifications don't afffect global state

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredBuildInputsIntegrationTest.groovy
ov7a: @alllex Thank you for your insights. After writing an integration test, I realized that we don't need to track mutations on a clone, so I've updated the code accordingly.

platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java
```diff
@@ -649,4 +659,37 @@ public Object setValue(Object value) {
             return oldValue;
         }
     }
+
+    /**
+     * A listener wrapper that only forwards read access notifications, ignoring mutations.
+     * Used by {@link #clone()} so that reads on the cloned properties are still tracked
+     * but writes to the clone don't affect the original properties' fingerprint.
```
alllex: ❓ I thought that it's already how this worked. Was it not the case?

My understanding was that the current System property interception infra already knows that when a user does `System.setProperty()` inside build logic, then any reads of that system property shouldn't make it part of the fingerprint (they no longer come from the environment).

I am also suspicious of this half-forwarding listener, because it might _interfere_ with the logic above. My guess is that if you now do `System.getProperties().clone.clear().put("foo",value).get("foo")` (pseudo-code), then `foo` becomes part of the fingerprint, because the original listener didn't know the property was modified, before reading it.

If we are continuing to track the cloned properties in a more fine-grained way, then we have to think (and test!) about all these scenarios. That's why I originally thought it'd be easier to do the report-aggregated.

My current thinking is that we should add more test coverage before merging.

ov7a: > My understanding was that the current System property interception infra already knows that when a user does System.setProperty() inside build logic, then any reads of that system property shouldn't make it part of the fingerprint (they no longer come from the environment).

Hmm, I wasn't aware of that. I stumbled across the issue when the sample from the original issue
```
            System.setProperty("Hello", "World")
            def copy = (Properties) System.getProperties().clone()
            copy.setProperty("Hello", "Bug")
            println("Hello = ${System.getProperty("Hello")}")
 ```
 didn't load CC entry, assuming that hello got changed. Given what you said, I think it's best to revert my last commit for now

Comment (ov7a) [reply platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java]: So, I can identify the following scenarios:
1. A read is performed on cloned properties - it should include the read property into fingerprint.
   1.1. A read is performend on cloned properties after clear or removal - ideally, it should not be included, because it's not the original property. Realistically, if you clone and then remove properties - you're doing something quite weird and it's ok to have cache miss in this scenario
   1.2. A read is performend on cloned properties, but this property was not present in the original map. - This is the scenario that rely on `setProperty()` extra logic. That case justifies that we should track writes as well.
2. A write is performed on clone. It should not contribute to the fingerprint, because it's performed on a clone.

Does it makes sense? Any imporant scenarios I missed?

Comment (alllex) [reply platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java]: Let me prefix my answer with this: I don't know whether system property tracking logic in CC _today_ is smart enough to avoid adding runtime-changed system properties to the fingerprint. I was under the impression that it is, but I might be wrong. It would be great if you could clarify that by investigating the code / finding tests. Alternatively, we should wait for Mike or ping Rodrigo directly, who authored and touched that part of the code base most recently, if memory serves.

---

> 1.2. A read is performend on cloned properties, but this property was not present in the original map. - This is the scenario that rely on setProperty() extra logic.

You didn't clarify the fingerprinting behavior for this case. The answer will depend on whether the property was set or not. When reading a non-existing property from the original map, we should add the key to the fingerprint as well.

---

I think maybe a different description of the scenarios can help to be more exhaustive.

We are dealing with "original" values and "modified" values. The value can be either a null or a string. We can assume that the original map consists of null values for all possible keys. The value is considered _modified_ (at runtime) if it was either set or removed (via remove or via clear, doesn't matter).

1. The act of modifying any property by itself doesn't affect the fingerprint
2. Once a value was modified, its key should never be added to the fingerprint
3. Any read of the original value should add the key-value pair to the fingerprint (even if there was no value in the backing map)

If you clone the initial properties, all the same rules should apply. Right?

This makes me think that we should just have shared test coverage for CC behavior with original system properties _and_ with cloned system properties.

Comment (ov7a) [reply platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java]: Thanks. I added a bunch of integ tests and the approach when only reads are tracked on clone still looks sane to me. I'll leave it to @mlopatkin as SME for the final review

File: platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredSystemPropertiesIntegrationTest.groovy [hunk 5]
```diff
@@ -0,0 +1,417 @@
...
+            this.groovyExpression = expr
+        }
+
+        @Override
+        String toString() {
+            return name().toLowerCase()
+        }
+    }
+
+    @Unroll
```
tresat, line 42: ❗ `@Unroll` is no longer necessary.

Applies to several tests.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredSystemPropertiesIntegrationTest.groovy
```diff
@@ -0,0 +1,417 @@
...
+    def "read of existing original key on #source source is tracked as fingerprint input"() {
+        given:
+        buildFile("""
+            def props = ${source.groovyExpression}
+            println("foo = \${props.getProperty("foo")}")
+            tasks.register("printProperty") { doLast { println("done") } }
```
tresat, line 48: ❓ Would it be more realistic to have the `printProperty` task print a value from `props`, and use the Properties with the task?

Applies to several tests.

tresat (changes requested):
Looks good overall, just minor cleanup.

platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java
mlopatkin: >whether system property tracking logic in CC today is smart enough to avoid adding runtime-changed system properties to the fingerprint. 

It is smart, but in a different way. Aside from recording incoming values for fingerprinting, we also want to record all runtime-set system properties and restore them on cache hit, so the execution phase sees the modified values every time. Here comes the clever trick Rodrigo implemented: rather than tracking mutations, we record them as fingerprint entries, effectively replaying all mutations while checking. E.g. you may have:
```java
System.getProperty("foo"); // records read("foo", expect: externally-set-value-of-foo)
System.setProperty("foo", "bar"); // records write("foo", "bar")
System.getProperty("foo"); // records read("foo", expect: "bar")
```

When checking we verify, that the first `read` sees the same externally set value, modify the value, verify that the second read sees the expected value. The latter is, obviously, always a match, but it simplifies the internals.

Now, I haven't looked at the implementation, maybe it already works this way, but here is my understanding of the desired behavior: with `clone()`, writes to the cloned instance shouldn't write anything to the fingerprint (modulo exposing old value through the return value of `put`); reading updated values shouldn't write anything either. However, reading a non-modified value (that comes from the original map) should emit a fingerprint entry. Thanks to the replay logic, we don't have to care if the original map has been modified by the time `clone` has been called.

platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java
```diff
@@ -649,4 +660,38 @@ public Object setValue(Object value) {
...
+    private static class ReadOnlyListener implements Listener {
+        private final Listener delegate;
+
+        ReadOnlyListener(Listener delegate) {
+            this.delegate = delegate;
+        }
+
+        @Override
+        public void onAccess(Object key, @Nullable Object value) {
+            delegate.onAccess(key, value);
```
mlopatkin, line 702: :x: This isn't entirely correct. If `key` has been mutated on the `clone()`, its reads shouldn't be reported. You're somewhat protected from it when you don't modify the original map, because FP code ignores duplicate reads, but for some reason that I cannot recall, reads after modification are still recorded with new value being expected.

This check-after-replay also gets in the way in another scenario:
1. `-Dfoo=bar`
2. `clone()`
3. `System.setProperty("foo", "baz") // replay will set `foo` to `baz`
4. `clone.get("foo")` - // We record `foo expect: bar` into fingerprint, but the check will compare it with `baz`

Another false miss.

It looks like that both stem from us re-fingerprinting the mutated property. @bamboo do you remember why we need this `remove`?
https://github.com/gradle/gradle/blob/4c407f5207bfb23b98f833a91ec41ef92f5555e5/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/fingerprint/ConfigurationCacheFingerprintWriter.kt#L900-L903

mlopatkin (changes requested):
There are a couple false misses now. It is better than false hits or invalid behavior we had before, but I wonder if we can fix it. So far, debugging these misses wouldn't be easy.

platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java
bamboo: We need the `remove` to ensure that we record new reads by user code, e.g. value sources. 

The problem here seems that we lose the distinction between reading a property from a **snapshot** of the shared mutable state at a point in time versus reading it straight from the source.

ov7a: I think the `clone` scenario isn't worth introducing complex logic (at least for now).

For now, I think reporting reads is a fine compromise:
Pros: We will catch indirect reads. 
Cons: There will be false misses in some (weird) scenarios.

Anything I'm missing? We can issue a log warning when there's a read from cloned properties, implying that the build logic should be improved. Would it be ok?

mlopatkin: Snapshotting mutated properties seems to matter only when we replay concurrently, then there may be a race between tracked reads and updates. Value Sources may also introduce untracked reads, which are even harder to get right when replaying them in parallel with updates.

Nevertheless, user may snapshot with `clone()` at any time, and this complicates the matter a lot. I thought a bit more about my scenario from https://github.com/gradle/gradle/pull/37526#discussion_r3123136866 and even removing tracking of the mutated properties wouldn't help (it will actually make things worse). Imagine we got rid of it, so we now get the proper hit. Now add another step:
5. `-Dfoo=qux` 

The read through `clone()` would change its value, but we don't have anything recorded, so we don't invalidate.

Arguably all this is likely rare, and people don't usually mix reading the clone and mutating the env. One exception that comes to mind is rolling back some temporary changes, but does it really have a place in build scripts? So the most common scenario, highlighted in the issue would be:
- get a clone
- perform some mutations on it
- pass it further, e.g. as `Test.systemProperties`.

To optimize for that, we can just ignore the reads of keys changed on the clone, and propagate reads on whatever wasn't snapshotted. Out-of-order reads will cause no-hit builds due to false invalidations, but that's the price we pay for the reasonably complex implementation.

ov7a: 👍 See the latest [commit](https://github.com/gradle/gradle/pull/37526/commits/0da190854eaa7012b40a5936523e59d047f57a6c) that covers this scenario

platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java
```diff
@@ -649,4 +662,44 @@ public Object setValue(Object value) {
...
+        }
+
+        @Override
+        public void onRemove(Object key) {
+            mutatedKeys.add(key);
+        }
+
+        @Override
+        public void onClear() {
+            cleared = true;
```
mlopatkin, line 721: 💅 I'd purge `mutatedKeys` here too.

platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java
```diff
@@ -483,6 +484,18 @@ public void list(PrintWriter out) {
...
+    @Override
+    @SuppressWarnings("MethodDoesntCallSuperMethod")
+    public synchronized Object clone() {
+        Properties clonedDelegate = (Properties) delegate.clone();
+        // Continue tracking individual property access on the clone, not reporting aggregate access.
+        // Cloning alone doesn't mean the caller will read all properties — they may only access specific ones.
+        // Mutations on the clone must NOT be reported: they only affect the clone's delegate, not the real system properties.
+        // Reads of mutated properties are not reported as well.
+        // See https://github.com/gradle/gradle/issues/17344
+        return new AccessTrackingProperties(clonedDelegate, new ReadUnchangedListener(listener));
```
mlopatkin: :x: This isn't entirely correct, let me show you a corner case. Let's start with system property `foo: bar` 
```
def clone = System.getProperties().clone()
```

Now there are two scenarios. First:
```
clone.putAll(foo: baz)
def cloneOfClone = clone.clone()
println(cloneOfClone.get("foo")) // baz. No inputs recorded today, because delegate remembers the change. This is correct.
```
Second:
```
def cloneOfClone = clone.clone()
clone.putAll(foo: baz)
println(cloneOfClone.get("foo")) // bar! No inputs recorded today, because delegate listener remembers the change, but it is unrelated.
// This is incorrect ‼️ 
```

As a mitigation, we'll have to "clone" the listeners too so they don't rely on its parent's state.

mlopatkin (changes requested):
There is one more false hit scenario I found, but otherwise the approach is sound.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredSystemPropertiesIntegrationTest.groovy
```diff
@@ -0,0 +1,564 @@
...
+
+        when:
+        configurationCacheRun("-Dfoo=external", "printProperty")
+
+        then:
+        configurationCache.assertStateLoaded()
+    }
+
+    @Issue("https://github.com/gradle/gradle/pull/37526")
+    def "mutate then read same key on clone is not tracked"() {
```
mlopatkin: ❓ How does this test verify "not tracked"?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredSystemPropertiesIntegrationTest.groovy
```diff
@@ -0,0 +1,564 @@
...
+    def "read of existing original key on #source source is tracked as fingerprint input"() {
+        given:
+        buildFile("""
+            def props = ${source.groovyExpression}
+            def captured = props.getProperty("foo")
+            tasks.register("printProperty") {
+                doLast { println("foo = \${captured}") }
+            }
+        """)
+        def configurationCache = newConfigurationCacheFixture()
```
mlopatkin: 💅 Can `configurationCache` become a field so you don't have to repeat it in every test?

platforms/core-configuration/input-tracking/src/main/java/org/gradle/internal/configuration/inputs/AccessTrackingProperties.java
```diff
@@ -649,4 +664,58 @@ public Object setValue(Object value) {
...
+
+        @Override
+        public void onChange(Object key, Object newValue) {
+            mutatedKeys.add(key);
+        }
+
+        @Override
+        public void onRemove(Object key) {
+            mutatedKeys.add(key);
+        }
```
mlopatkin, line 717: 💅 Late realization: we can also gate on `!cleared` here so the set no longer grows upon mutations.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredSystemPropertiesIntegrationTest.groovy
```diff
@@ -0,0 +1,564 @@
...
+
+        then:
+        configurationCache.assertStateLoaded()
+        outputContains("foo = null")
+
+        where:
+        source << PropertySource.values()
+    }
+
+    def "clear then getProperty on real source returns null and hits cache"() {
```
mlopatkin: 💅 Why aren't this and the next test folded together into a parameterized one?

mlopatkin (commented):
Looks good, though one test is suspicious.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/inputs/undeclared/UndeclaredSystemPropertiesIntegrationTest.groovy
ov7a: If it was tracked, then it wouldn't load state from cache, right? 

Having said that, I'll add a third case with `foo=different` - would it be enough?

ov7a: Aw, shoot. 

`setProperty` does a read internally that goes nowhere :/ (the test works with putAll though)

ov7a: Should be more clear now

mlopatkin (approved):
LGTM!

PR #37533, opened 2026-04-15
Fix flaky CapturingUserInputCrossVersionSpec tests

blindpirate:
## Problem

`CapturingUserInputCrossVersionSpec` and `CapturingMultipleUserInputCrossVersionSpec` have been failing intermittently in `:tooling-api:embeddedCrossVersionTest`.

Root cause: `DaemonClientInputForwarder` uses a **single-thread executor** shared between two competing tasks:
- `readAndForwardText()` — reads a line from stdin and delivers it as a `UserResponse` to `DefaultUserInputReader`
- `readAndForwardStdin()` — reads raw bytes from stdin and forwards them as `ForwardInput` to `StdInStream` (e.g. triggered by buildSrc compilation reading `System.in`)

Both tasks share the same `BufferedReader` wrapping the provided `InputStream`. If `readAndForwardStdin()` is queued first (blocking on `reader.read()`), it can consume the bytes written by the test before `readAndForwardText()` gets a turn. Those bytes are forwarded as raw `ForwardInput` rather than a `UserResponse`, causing `askYesNoQuestion`/`askQuestion` to receive EOF and return `null`/default instead of the provided answer.

The previous test approach — `PipedInputStream` + `poll(60)` waiting for the prompt + writing bytes asynchronously — has a window where this race can occur between the poll succeeding (prompt rendered) and the byte-write reaching the executor.

## Fix

Replace `PipedInputStream`/`PipedOutputStream` + async `run(resultHandler)` + `poll()` with a `ByteArrayInputStream` pre-loaded with all answers, and synchronous `run()`. Since `ByteArrayInputStream.readLine()` returns immediately (non-blocking), the executor task completes correctly regardless of scheduling order — no window for `readAndForwardStdin()` to win the race.

## Test plan

- [ ] `CapturingUserInputCrossVersionSpec` — "can capture user input if standard input was provided"
- [ ] `CapturingUserInputCrossVersionSpec` — "cannot capture user input if standard input was not provided" (unchanged)
- [ ] `CapturingMultipleUserInputCrossVersionSpec` — all three test methods


blindpirate:
@bot-gradle test this

reinsch82 (approved):

PR #37541, opened 2026-04-15
Bump profiler to 0.25.0-alpha-1

6hundreds:
Adopting the latest version of the Profiler. Its main feature is IDEA Sync support, although this PR doesn't try to integrate it so far. This PR is aiming to have working performance-test suite.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


6hundreds:
@bot-gradle test apt

testing/internal-performance-testing/src/main/groovy/org/gradle/performance/fixture/GradleBuildExperimentRunner.java
```diff
@@ -89,11 +91,10 @@ public void doRun(String testId, BuildExperimentSpec experiment, MeasuredOperati
         InvocationSpec invocationSpec = experiment.getInvocation();
         File workingDirectory = invocationSpec.getWorkingDirectory();
 
-        if (!(invocationSpec instanceof GradleInvocationSpec)) {
+        if (!(invocationSpec instanceof GradleInvocationSpec invocation)) {
```
alllex, line 94: 💅  Do we need an unused variable?

alllex (approved):

testing/internal-performance-testing/src/main/groovy/org/gradle/performance/fixture/GradleBuildExperimentRunner.java
6hundreds: We use it at the line 102

alllex: Ah, I was not ready for the size of this method. But also, I am questioning the readability of this code, when the pattern-matching is used with negation. The variable is introduced _inside_ an if expression, but is used in the rest of the method body.

I think the previous version of the code with a cast was actually more readable.

Looking at this specific example, I wonder if Java language designers should have forbidden such a code pattern.

PR #37543, opened 2026-04-15
Make all None parameters behave identically

### LINKED ISSUE #30182: Unclear message when ValueSource parameters are accessed for a source with None parameters
Consider the following snippet:

```kotlin
abstract class MyValueSource : ValueSource<String, ValueSourceParameters.None> {
    override fun obtain(): String? {
        println("Parameters: $parameters")
        return "foo"
    }
}
```

This currently fails with:
```
> org.gradle.internal.service.UnknownServiceException: No service of type ValueSourceParameters$None available in value source services.
  > No service of type ValueSourceParameters$None available in value source services.
```

There seems to be no contract for accessing the parameters for a value source that has `ValueSourceParameters.None`, e.g. in any of the javadocs.

It is currently forbidden by failing, but there seems to be no reason why this shouldn't be supported.

There are at least two use-cases:
* A user was experimenting and reduced their parameters to `None` without updating the implementation code that still accesses the parameters
* A user has an abstract value source class that prints parameters for debugging. It should not matter whether children of this class use `ValueSourceParameters.None` as a parameter type.

---

Build scan: https://ge.gradle.org/s/yxps2f3m5husm/failure#1

ov7a:
Fixes #30182

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed behavior for work actions, transforms, value sources, build services, and flow actions with no parameters. `getParameters()` now returns a singleton instance instead of throwing an exception.

* **Tests**
  * Added integration tests validating that no-parameter actions can safely access their parameters.

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheValueSourceIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheValueSourceIntegrationTest.groovy
@@ -24,6 +24,36 @@ import spock.lang.Issue
 
 class ConfigurationCacheValueSourceIntegrationTest extends AbstractConfigurationCacheIntegrationTest {
 
+    @Issue("https://github.com/gradle/gradle/issues/30182")
+    def "value source without parameters can access None parameters"() {
+        given:
+        def configurationCache = newConfigurationCacheFixture()
+
+        buildFile("""
+            import org.gradle.api.provider.*
+
+            abstract class GreetValueSource implements ValueSource<String, ValueSourceParameters.None> {
+                String obtain() {
+                    println("Parameters: " + getParameters())
+                    return "Hello!"
+                }
+            }
+
+            def greetValueSource = providers.of(GreetValueSource) {}
+            tasks.register("greet") {
+                doLast { println greetValueSource.get() }
+            }
+        """)
+
+        when:
+        configurationCacheRun "greet"
+
+        then:
+        configurationCache.assertStateStored()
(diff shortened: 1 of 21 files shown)


ov7a:
@bot-gradle test this

ov7a:
@coderabbitai review

tresat (approved):
Should we make a common `Parameters` super-interface for all these parameter interfaces to implement, to unify the hierarchy?

ov7a:
@tresat I don't think we should. These parameters are from different domains. A marker interface is no better than proper naming here, IMO

mlopatkin (approved):
Looks good!

🤔  It might be beneficial to also look at the configuration side of things, e.g. at what you get when calling `getParameters()` or `parameters` when registering build service, transform, etc.

So far, there is certain level of inconsistency as well:
- registering `ValueSource` and `BuildService` return `null` from `get` and throw when calling `parameters {}`
- registering `TransformAction` throws `IllegalStateException`
- `FlowAction` and `WorkAction` just doesn't run configuration closure, so there is no place to call the methods.

Aligning first two and the last in either direction would be a breaking change, unfortunately. However, if we're to change, I vote for:
- Running configuration closure regardless of type (changing `FlowAction` and `WorkAction` behavior)
- Returning/supplying `None` instance rather than `null` or throwing.

We can do it separately, not in this PR.

💭  My knee-jerk reaction was to avoid exposing `INSTANCE` through public API, but it would make unit testing of `None`-taking classes harder, so I'm fine with it now.

ov7a:
Thanks @mlopatkin !

I've created an issue from your comment
- https://github.com/gradle/gradle/issues/37638

and will proceed with merging this PR.

PR #37552, opened 2026-04-16
Fix addJvmArguments() system property propagation to startParameter

### LINKED ISSUE #37550: BuildLauncher.addJvmArguments() does not propagate system properties to daemon start parameters
### Current Behavior

System properties passed via `LongRunningOperation.addJvmArguments()` (e.g. `-Dorg.gradle.tooling.parallel.ignore-legacy-default=true`) are not visible in `startParameter.systemPropertiesArgs` and are not available via `System.getProperty()` early in the daemon lifecycle.

This was surfaced by JetBrains' transparent Parallel Model Fetching (PMF) enablement in IDEA 2026.1.1. IDEA was passing `org.gradle.tooling.parallel.ignore-legacy-default=true` via `addJvmArguments()`, but the Gradle daemon did not see it. If shipped that way, PMF would have activated incorrectly for users who have `org.gradle.parallel=true` in their `gradle.properties` but have not opted into PMF, potentially making IDE sync unreliable. JetBrains caught this before GA and switched to `withArguments()` as a workaround.

The property _is_ visible later when read from a build script via `System.getProperty()`, but that is too late — PMF configuration needs it before the build action runs.

### Expected Behavior

System properties passed via `addJvmArguments("-Dfoo=bar")` should be available in `startParameter.systemPropertiesArgs` and behave identically to properties passed via `addArguments("-Dfoo=bar")` / `withArguments("-Dfoo=bar")`.

### Context

JetBrains caught the issue before shipping IDEA 2026.1.1 and worked around it by using `withArguments()` instead of `addJvmArguments()`. However, the underlying TAPI bug remains.

Existing test coverage did not catch this because:
- `ParallelActionExecutionCrossVersionSpec` uses `addArguments()` rather than `addJvmArguments()`, so it sidesteps the bug.
- `JvmArgumentPassingCrossVersionSpec` only reads system properties via `System.getProperties()` from the build script, which is too late in the lifecycle to catch early-read properties. It also contains a dead `jvmArgs` local variable that is read but never asserted.

Using the Tooling API, run a build action that reads a system property early (before the build script phase):

```java
ProjectConnection connection = GradleConnector.newConnector()
    .forProjectDirectory(projectDir)
    .connect();

BuildLauncher launcher = connection.newBuild();
launcher.addJvmArguments("-Dorg.gradle.tooling.parallel.ignore-legacy-default=true");
launcher.run();
```

The property `org.gradle.tooling.parallel.ignore-legacy-default` will **not** be present in `startParameter.systemPropertiesArgs`.

Replacing `addJvmArguments()` with `addArguments()` or `withArguments()` makes the property visible.

### Gradle version

9.5.0-RC2 (also reproduces on `master`)

hegyibalint:
Fixes #37550

### DIFF
--- a/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r960/JvmArgumentSystemPropertyCrossVersionSpec.groovy
+++ b/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r960/JvmArgumentSystemPropertyCrossVersionSpec.groovy
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.integtests.tooling.r960
+
+import org.gradle.integtests.tooling.fixture.TargetGradleVersion
+import org.gradle.integtests.tooling.fixture.ToolingApiSpecification
+import org.gradle.integtests.tooling.fixture.ToolingApiVersion
+import org.gradle.test.precondition.Requires
+import org.gradle.test.preconditions.TestExecutionPreconditions
+import org.gradle.tooling.BuildAction
+import org.gradle.tooling.BuildController
+import spock.lang.Issue
+
+@Issue("https://github.com/gradle/gradle/issues/37550")
+@ToolingApiVersion(">=8.13")
+@TargetGradleVersion(">=9.6")
+@Requires(
+    value = TestExecutionPreconditions.NotEmbeddedExecutor,
+    reason = "JVM arguments require the external executor to be passed to the Gradle daemon."
+)
+class JvmArgumentSystemPropertyCrossVersionSpec extends ToolingApiSpecification {
+
+    interface StartParameterPropertyModel {
+        String getPropertyValue()
+    }
+
+    static class FetchStartParameterPropertyAction implements BuildAction<StartParameterPropertyModel>, Serializable {
+        @Override
+        StartParameterPropertyModel execute(BuildController controller) {
+            return controller.getModel(StartParameterPropertyModel)
+        }
+    }
+
+    def setup() {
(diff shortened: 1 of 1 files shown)


platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r960/JvmArgumentSystemPropertyCrossVersionSpec.groovy
```diff
@@ -0,0 +1,99 @@
...
+import org.gradle.integtests.tooling.fixture.ToolingApiVersion
+import org.gradle.test.precondition.Requires
+import org.gradle.test.preconditions.TestExecutionPreconditions
+import org.gradle.tooling.BuildAction
+import org.gradle.tooling.BuildController
+import spock.lang.Issue
+
+@Issue("https://github.com/gradle/gradle/issues/37550")
+@ToolingApiVersion(">=8.13")
+@TargetGradleVersion(">=9.6")
```
alllex, line 30: 🤔  This test should cover pretty much all Gradle versions supported by TAPI, imo
Especially, if we are intending to fix (aka change) the behavior in one of the upcoming Gradle versions.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r960/JvmArgumentSystemPropertyCrossVersionSpec.groovy
```diff
@@ -0,0 +1,99 @@
...
+                }
+            }
+        """
+
+        buildFile """
+            apply plugin: StartParameterPropertyPlugin
+        """
+    }
+
+    def "system properties passed via addJvmArguments are available in startParameter.systemPropertiesArgs"() {
```
alllex, line 88: 💭  We should be explicit on whether we are testing our internal mechanisms vs what is available to the user.

If it's for the user (aka Tooling Model Builder author), then with these tests we shown an intent that that's how these users are _expected_ to read system properties if they want to or need to. In that case, we should definitely cover using `System.getProperty()`.

If it's for the internal mechanism, I'd rather be very explicit in the comment, because then it's not about the user contract, and we can and probably should actually test it differently, since then it's not about tooling model builders at all (but about tooling invocations).

hegyibalint:
@bot-gradle test this

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r960/JvmArgumentSystemPropertyCrossVersionSpec.groovy
hegyibalint: Fair point! Let me loosen the requirements

hegyibalint:
We will drop this PR, as there are older tests that could be fixed and re-used.

PR #37553, opened 2026-04-16
Upgrade Spock from 2.3-groovy-4.0 to 2.4-groovy-4.0

cobexer:
## Summary
- Upgrade Spock framework dependency from 2.3-groovy-4.0 to 2.4-groovy-4.0
- Adapt `IntegrationTestTimeoutInterceptor` to Spock 2.4's changed `TimeoutInterceptor(Duration, TimeoutConfiguration)` constructor and remove the now unnecessary `TimeoutAdapter`
- Use Spock 2.4's new `combined:` blocks to replace manually expanded cartesian-product `where:` tables in tests
- Convert `target.with { assert ... }` call sites across unit and integration tests to `verifyAll(target) { ... }`. Spock 2.4 implicit-asserts every top-level expression in `then:`/`expect:`, and `with(...)` returns the closure's last value — `null` when the body ends in an `assert` — so the outer expression now fails. `verifyAll` is the Spock-aware grouping construct already used elsewhere in the repo.
- `@Ignore GcsClientTest > should include uri when file not found`: Spock 2.4's strict `ByteBuddyMockInteractionValidator` rejects stubbing the `final` `GoogleJsonResponseException.getStatusCode()`. A per-mock `mockMaker: MockMakers.mockito` opt-in compiles but breaks task execution at runtime because the production code does not throw an exception for a 404.

## References
- [Spock 2.4 release notes](https://spockframework.org/spock/docs/2.4/release_notes.html)
- [Cross-multiplying data providers](https://spockframework.org/spock/docs/2.4/data_driven_testing.html#_cross_multiplying_data_providers)


cobexer:
@bot-gradle test this

cobexer:
@bot-gradle test RfR

cobexer:
@bot-gradle test RfR

ov7a (approved):

PR #37556, opened 2026-04-16
Deprecate Kotlin DSL delegated properties

eskatos:
Deprecates all Kotlin DSL delegated property extensions (`by tasks.registering`, `by configurations.creating`, `by extra`, etc.) in favor of explicit API calls.

* Specification: [Removing Kotlin DSL Property Delegates](https://docs.google.com/document/d/140dOuBcCEDCKLGTiY80NNx2bCQ7dUv8EVFsn4ph86fU/edit?tab=t.0)
* Fixes https://github.com/gradle/gradle/issues/37555
* Fixes #9367

#### Changes

- Emit deprecation warnings for all Kotlin DSL delegate extensions
- Add `@Deprecated` annotations to delegate APIs
- Stop generating delegate syntax in `:build-init`
- Replace delegate syntax in build logic, documentation, tests, and performance tests
- Add deprecation and upgrade guide documentation

#### Tests

- Existing delegate unit tests updated to expect deprecation warnings
- Integration tests verify deprecation warnings are emitted
- Documentation snippets and samples tests pass without delegates

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #37557, opened 2026-04-16
Remove Kotlin DSL delegated property extensions

eskatos:
Remove all delegate provider classes, extension functions, operators, and related tests from the Kotlin DSL:

- PropertyDelegate, MutablePropertyDelegate interfaces
- registering, creating, existing, getting delegates
- provideDelegate operators on Project, Settings, containers
- getValue/setValue on Property, ConfigurableFileCollection, ExtensionContainer, ExtraPropertiesExtension

Keep the catch-all KT-25810 `T.provideDelegate` workaround as `@Deprecated(level = HIDDEN)` so plugins compiled against earlier Gradle versions that baked an INVOKESTATIC reference into their bytecode (e.g. via `by lazy { }` after
`import org.gradle.kotlin.dsl.provideDelegate`) keep loading. Covered by ProvideDelegateBinaryCompatibilityIntegrationTest, which publishes such a plugin with a pinned previous Gradle and applies it with the current one.

Delete tests that exclusively exercised delegate syntax. Rewrite tests that mixed delegate and non-delegate coverage to keep only the non-delegate tests.

* Follows up https://github.com/gradle/gradle/pull/37556

This is a draft for Gradle 10.0.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


eskatos:
@bot-gradle test RFN

PR #37558, opened 2026-04-16
Refactor ide smoke tests

ov7a:
1. Provision installations via IntelliJ Platform and caching repo
2. Move versions to properties
3. Reduce duplication
4. Add a task for automatic upgrade

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ov7a:
@bot-gradle test AST

ov7a:
@bot-gradle  test APT

ov7a:
@bot-gradle test this

testing/smoke-ide-test/build.gradle.kts
```diff
@@ -30,6 +32,8 @@ dependencyAnalysis {
 
 addDependenciesAndConfigurations("smokeIde")
 
+IdeProvisioningPlugin.registerTestIdes(project)
```
cobexer: ❓ could/should the application of the plugin `id("gradlebuild.ide-provisioning")` set this up automaticaylly?

testing/smoke-ide-test/build.gradle.kts
```diff
@@ -92,6 +96,9 @@ tasks {
                 ideStarterDir = ideStarterBuildDir
             }
         )
+        jvmArgumentProviders.add(
+            IdeProvisioningPlugin.ideArchivesProvider(project)
```
cobexer, line 98: 💅 this looks out of place considering the manually passed project is there a nicer way?

gradle.properties
```diff
@@ -49,6 +49,11 @@ org.jetbrains.dokka.internal.enableWorkaroundKT80551=true
 # https://github.com/GradleUp/shadow/issues/1835
 com.gradleup.shadow.enableDevelocityIntegration=false
 
+# Redirect IntelliJ Platform Gradle Plugin cache under build/ so it is gitignored
+org.jetbrains.intellij.platform.intellijPlatformCache=build/intellijPlatform
```
cobexer, line 53: 💅 not sure if that is the right place

build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/ide/IdeProvisioningPlugin.kt
```diff
@@ -0,0 +1,129 @@
...
+                }
+                jetbrainsIdeInstallers {
+                    url = uri("https://repo.grdev.net/artifactory/jetbrains-ide-installers")
+                }
+                releases {
+                    url = uri("https://repo.grdev.net/artifactory/jetbrains")
+                }
+                jetbrainsRuntime {
+                    url = uri("https://repo.grdev.net/artifactory/intellij-jbr")
+                }
```
cobexer: ❌ Need to think about this, I'd prefer a solution that uses external upstream URLs and get's overriden via `REPO_MIRROR_URLS` when running in CI environments.

cobexer (changes requested):

testing/smoke-ide-test/build.gradle.kts
ov7a: That was the initial approach, but it caused dependency resolution errors like
```
org.gradle.tooling.BuildException: Could not fetch model of type 'IdeaProject' using connection to Gradle installation '/home/tcagent1/agent/work/f63322e10dd6b396/packaging/distributions-full/build/bin distribution'.
> org.gradle.internal.exceptions.LocationAwareException: Failed to query the value of property 'languageVersion'.
   > org.gradle.internal.buildtree.DefaultBuildTreeLifecycleController$BuildActionExecutionException: Failed to query the value of property 'languageVersion'.
      > org.gradle.api.internal.provider.AbstractProperty$PropertyQueryException: Failed to query the value of property 'languageVersion'.
         > org.gradle.api.internal.artifacts.configurations.DefaultConfiguration$IllegalResolutionException: Resolution of the configuration ':enterprise-plugin-performance:intellijPlatformDependency' was attempted without an exclusive lock. This is unsafe and not allowed.
```
I'll try one more time to resolve this, but I don't feel it's worth to invest too much time into it at this point, TBH

testing/smoke-ide-test/build.gradle.kts
ov7a: IDK, I wanted to keep the internals consolidated in IdeProvisioningPlugin as much as possible. Any suggestions?

gradle.properties
ov7a: by default, it was a root of the repo :harold: Happy to move it somewhere else, but I think it's fine?

build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/ide/IdeProvisioningPlugin.kt
ov7a: Hmm, I can shove this configuration behind a CI flag - so on CI it will use repo.grdev.net, and defaults without it - would it be ok?

gradle.properties
cobexer: @blindpirate suggestions?

build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/ide/IdeProvisioningPlugin.kt
cobexer: `REPO_MIRROR_URLS` should do this automatically, let's figure out what needs to be set up on TC and decide then

6hundreds:
@ov7a Could you provide a motivation behind this? I see we hide some archive downloading logic behind IntelliJPlatform plugin and add some automation like auto-update of the verification file. In general, the IDE provisioning process became more complex in my opinion, with no clear benefits in compare with vanilla solution we had before.

I've cooked a draft PR which is doing less(it doesn't update verification file or does automatic update) and keeps all the bits mostly in place, while adding IDEA support to it, and also prevents duplication: https://github.com/gradle/gradle/pull/37602

So I wonder, do we have a strong reason to keep all this complexity?

6hundreds (changes requested):
See my previous  comment

ov7a:
@6hundreds So the original PR was triggered by the AS URL changes and related problems with the caching repo. (Also manifested via [ide-starter](https://github.com/gradle/gradle-private/issues/5157)).
DV had fewer issues with that because they were using the IntelliJ platform plugin for tests.

So I started this work to 
a) have a single way to provide AS (and ide) 
b) encapsulate url construction (via the IntelliJ platfrom plugin)
c) perform downloads via caching proxy 
d) perform dependency verification

This PR does one more thing - it extracts versions into properties and adds a task to upgrade these versions automatically. It's a cherry on top I wanted to do for a while.

Looking back, the IJ platform is very opinionated, and it was not trivial to make it work without side effects. However, it encapsulates all the dances around downloading and unpacking IDEs.

I took a quick liik at your PR and IIUC, it only addresses a).

6hundreds:
@ov7a I'm not 100% sold that AS URL changes justifies introducing this complexity.
We don't know how often such changes is going to happen, but even with IntelliJPlatform leverage, to pick up the updated URLs we'll need to update IntelliPlatform plugin, I believe.
> c) perform downloads via caching proxy

Could you share how do we benefit of it?
> task to upgrade these versions automatically

Well, I just wonder how often do we need to do that. In my experience it's not so often, but this tasks contributes the significant part to the complexity.

> I took a quick liik at your PR and IIUC, it only addresses a). 

Yes, that's true

ov7a:
> We don't know how often such changes is going to happen, but even with IntelliJPlatform leverage, to pick up the updated URLs we'll need to update IntelliPlatform plugin, I believe.

Yeah, I'm not sure either, but IJ plugin update should be trivial, and I think it's easier to delegate the maintenance of this logic...

> Well, I just wonder how often do we need to do that. In my experience it's not so often, but this tasks contributes the significant part to the complexity.

Every release. But this task does not depend on IJ platform, and can be extracted from this PR (as long as versions are picked from properties/build logic).




 > c) perform downloads via caching proxy
> Could you share how do we benefit of it?

Being good citizen + stability (@cobexer is a better person to reply on this question)

6hundreds:
@ov7a 
> Yeah, I'm not sure either, but IJ plugin update should be trivial, and I think it's easier to delegate the maintenance of this logic...

Well, this triviality of the updating is not free - we're paying for this by dancing around opinions that Intellij plugin has, and supporting this build logic in future. I wouldn't expect URL changing so often that it justifies this load.

The plugin's opinions here are not subtle that we can accept easily and live with it. It requires us to have additional entries in `gradle.properties` and custom plumbing in `mirrors.settings.gradle.kts`. All plugins assumptions about caching, resolution, and repository layout — all opaque and harder to debug when CI breaks.  

Having said that, it's hard to agree to me that we getting any benefits from IntelliJ plugin in compare with vanilla Ivy setup(I have to admit that I'm not sure how good Ivy citizen with caching proxy usage you mentioned).

> Every release

I'd say IDE dependencies aren't special in terms of smoke testing, and if we do update others on each release, then yes,  automated update is a thing.

Both, automated update, and performing dependency verification - I think it make sense to create a separate PR(s) for these.

build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/ide/IdeProvisioningPlugin.kt
```diff
@@ -0,0 +1,118 @@
...
+            require(file.exists()) { "IDE versions file not found: $file" }
+            return Properties().apply {
+                file.inputStream().use { load(it) }
+            }
+        }
+    }
+
+    override fun apply(target: Project) {
+        with(target) {
+            pluginManager.apply("org.jetbrains.intellij.platform.module")
```
cobexer: ❌ looking at what the plugins do, I think we should only apply the ~~base~~ settings plugin:
https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/b2e749110acf602c6d56f76014c5342fefcafdb5/build.gradle.kts#L243

https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/b2e749110acf602c6d56f76014c5342fefcafdb5/src/main/kotlin/org/jetbrains/intellij/platform/gradle/plugins/settings/IntelliJPlatformSettingsPlugin.kt#L13-L24

if we can get away with that!

cobexer (changes requested):

cobexer:
IIUC using the settings plugin I discovered should allow for a much simpler integration than what either PR currently has.

The IJ plugin would completely hide`*` the download URL complexity from us which AFAIU would simplify both PRs.

Mirroring integration is a must especially since the download archives are huge.

I agree that separate PRs may be preferred - but they all depend on the agreed upon solution.


Let's attempt using the settings plugin from IJ only and see if we can utilize it to abstract the download handling from us - if that works I am in favor of using the IJ plugin and have IJ make sure their plugin can download their IDE - that **should** be simpler.

`*` aside from the mirror setup in CI systems

6hundreds:
@cobexer @ov7a Sounds good.
> IJ make sure their plugin can download their IDE - that should be simpler.

It might be a case that IDEA downloading will be simpler, but AS will require some dancing. In this case I'd prefer same approach for both IDEs

6hundreds:
@ov7a Is it ready for review?

build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/ide/IdeProvisioningPlugin.kt
```diff
@@ -0,0 +1,183 @@
...
+                androidStudioInstallers()
+                jetbrainsIdeInstallers()
+            }
+
+            val androidStudioArchive = configurations.register(ANDROID_STUDIO_ARCHIVE_CONFIGURATION) { isCanBeConsumed = false }
+            val intellijIdeaArchive = configurations.register(INTELLIJ_IDEA_ARCHIVE_CONFIGURATION) {
+                isCanBeConsumed = false
+                // JetBrains CDN intermittently serves gzip-encoded HTML for missing `.asc` URLs,
+                // which Gradle caches and then fails to parse as a PGP signature (ArmoredInputException).
+                resolutionStrategy.disableDependencyVerification()
```
ov7a, line 88: I plan to fix this in gradle, but until then, there's no easy workaround for this

ov7a:
@6hundreds I want to explore an idea of shoving coordinates (`androidStudioDependencyCoordinates` and `intellijIdeaInstallerCoordinates`) into .properties (so there's no need to recompute them every time), but it wouldn't change the general approach, so you can take a look.

ov7a:
> I want to explore an idea of shoving coordinates (androidStudioDependencyCoordinates and intellijIdeaInstallerCoordinates) into .properties

Meh, it didn't went well

@6hundreds @cobexer, please review. I think the current approach is better than my previous version and ticks all the boxes. 

I don't like the dances around classifiers, but I haven't found a decent way to make it simpler.

build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/ide/IdeProvisioningPlugin.kt
ov7a: * https://github.com/gradle/gradle/pull/37667

ov7a: Filed an issue to JB 
* https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/2140
And they have already fixed it on their side

cobexer:
> It might be a case that IDEA downloading will be simpler, but AS will require some dancing.

The IntelliJ platform Gradle plugin is supposed to take care of the dance for us, for example the updated URL schema of Android Studio was implemented with: JetBrains/intellij-platform-gradle-plugin#2087

build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/ide/IdeProvisioningPlugin.kt
cobexer: could we create and link an issue to track fixing this after the next release of the IJ platform plugin?

cobexer (approved):
LGTM!

build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/ide/IdeProvisioningPlugin.kt
```diff
@@ -0,0 +1,183 @@
...
+                targetDir.set(layout.buildDirectory.dir("android-studio"))
+            }
+
+            extensions.create<IdeProvisioningExtension>("ideProvisioning").apply {
+                androidStudioInstallPath.set(extractAndroidStudio.flatMap { it.targetDir })
+            }
+        }
+    }
+
+    private fun Project.androidStudioDependencyCoordinates(asVersion: String): Provider<String> {
```
cobexer, line 110: ❓ I'm wondering if we can utilize https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/0574034e5f5395d6190122280e932edac284930e/src/main/kotlin/org/jetbrains/intellij/platform/gradle/IntelliJPlatformType.kt#L22 (and connected code) to simplify this logic here

6hundreds:
@ov7a Speaking only about provisioning logic. This approach is much concise, I certainly like it more. However, can you point an area that was significantly improved by using IntelliJPlugin in your opinion?

The things I see are: using `ExtractionService` that allowed us to get rid of custom extraction task, and elimination of codename declaration for AndroidStudio, as the downloadLink is provided by `AndroidStudioValueSource` now. 

But we still do own [link parsing](https://github.com/gradle/gradle/pull/37558/changes#diff-ca47132e42e3fd1b78b39f311906733bed845aa24eef3739bf1f2771ce017786R119-R127), what can be potentially broken by URL schema change(talking about this as it was a main driver for these changes). As a side note - we also [bake this type of logic to automated update task](https://github.com/gradle/gradle/pull/37558/changes#diff-b2cf3568799f3ac242c23c9984f3cf99c47f9548148e16190778d10d6111e1cdR51-R77). Speaking differently - we're as fragile as before to URL change.

I didn't found justification for using IntelliJPlatfrom in the current shape, but I don't want to gatekeep the merge.

Speaking about other aspects of the PR - autoupdate and verification metadata file change - I'd prefer to see those as a separate PR(s?)

build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/ide/IdeProvisioningPlugin.kt
ov7a: This is what I tried, see above, but it was quite hard to push through private methods

ov7a:
@6hundreds 

> The things I see are: using ExtractionService that allowed us to get rid of custom extraction task, and elimination of codename declaration for AndroidStudio, as the downloadLink is provided by AndroidStudioValueSource now.

I think these things are good enough to justify the change. 

> But we still do own [link parsing](https://github.com/gradle/gradle/pull/37558/changes#diff-ca47132e42e3fd1b78b39f311906733bed845aa24eef3739bf1f2771ce017786R119-R127), what can be potentially broken by URL schema change(talking about this as it was a main driver for these changes).

That's true. I struggled to simplify this more. However, it's not a full reconstruction from zero, just an obscure way to get a classifier and extension. In a perfect world, gradle module metadata would solve it. It may be worth raising this issue in JB. 

> As a side note - we also [bake this type of logic to automated update task](https://github.com/gradle/gradle/pull/37558/changes#diff-b2cf3568799f3ac242c23c9984f3cf99c47f9548148e16190778d10d6111e1cdR51-R77).

I'd treat the update version logic as a glorified script, so I think it's acceptable. I don't think there's a easier way since it's private in the plugin (see https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/ce833688b128794994620f1eb49cc17cc529567b/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt#L954)

> Speaking differently - we're as fragile as before to URL change.

I'd say we would be less fragile:
1. We would do less custom logic
2. It's easier to adapt something from the IJ platform plugin if something breaks.

> I didn't found justification for using IntelliJPlatfrom in the current shape, but I don't want to gatekeep the merge.
> Speaking about other aspects of the PR - autoupdate and verification metadata file change - I'd prefer to see those as a separate PR(s?)

I'd say migrating to IJP plugin made it easier to verify dependencies (thanks to repo setup & (partial) info about coordinates). A caching proxy is another clear benefit. While it's possible to achive without IJP, it would require more code to maintain

PR #37562, opened 2026-04-17
Strip `JVM *_OPTIONS` stderr echoes from execution results

blindpirate:
JDK prints lines such as "Picked up JAVA_TOOL_OPTIONS:" to stderr when those environment variables are set. CI often injects them, which made command-line stderr differ from Tooling API captures in cross-version tests.

We now drops those lines, and OutputScrapingExecutionResult applies this when wrapping process stderr so integration tests see stable error output. A unit test covers the stripping behavior.

See also https://github.com/gradle/bt-dev-prod-infrastructure/pull/300


blindpirate:
@coderabbitai review

blindpirate:
@bot-gradle test rfr

PR #37568, opened 2026-04-17
Enrich file visitor with file size info

snoopcheri:
This PR adds file size reporting to the file property visitor API, enabling consumers (such as the DV plugin) to retrieve file lengths when traversing resolved artifacts during Gradle's snapshot-based build operations.

The change is very small and doesn't have any measurable impact on the performance.

The DV plugin uses the new method to skip system calls for determining the file size on its own (see [this draft PR in DV](https://github.com/gradle/dv/pull/64736)).


subprojects/core/src/main/java/org/gradle/api/internal/tasks/BaseFilePropertyVisitState.java
```diff
@@ -129,6 +137,9 @@ public SnapshotVisitResult visitEntry(FileSystemLocationSnapshot snapshot) {
         this.path = snapshot.getAbsolutePath();
         this.name = snapshot.getName();
         this.hash = fingerprint.getNormalizedContentHash();
+        this.length = snapshot instanceof RegularFileSnapshot
```
bamboo: ❓ Can it be anything other than a `RegularFileSnapshot` at this point?

snoopcheri: I believe it could also be of type `MissingFileSnapshot`.

According to Claude, a missing file with `AbsolutePathFingerprintingStrategy` could pass the fingerprint check in line 131, and we could end up with a `MissingFileSnapshot` here.

bamboo: ❌ I'd make it an explicit expectation then. I suggest introducing a member function `getFileSnapshotLength(FileSystemLocationSnapshot snapshot)` that extracts the length from `RegularFileSnapshot`, return `0` for `MissingFileSnapshot` and fails otherwise. Consider the PR accepted after that change.

bamboo (changes requested):

subprojects/core/src/main/java/org/gradle/api/internal/tasks/BaseFilePropertyVisitState.java
snoopcheri: DONE (2f7c6270)

bamboo (approved):

PR #37575, opened 2026-04-18
Deprecate accessing Task passed to task dependency closure

jvandort:
This way, starting in Gradle 10, we can avoid wiring the Task through the whole task dependency resolution chain, permitting us to begin to 'un taskify' the work graph traversal logic

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/DefaultTaskDependency.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/DefaultTaskDependency.java
@@ -30,7 +30,6 @@ import org.gradle.internal.typeconversion.UnsupportedNotationException;
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 
-import java.lang.reflect.Proxy;
 import java.nio.file.Path;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -114,9 +113,18 @@ public class DefaultTaskDependency extends AbstractTaskDependency {
                 ((TaskDependencyContainer) dependency).visitDependencies(context);
             } else if (dependency instanceof Closure) {
                 Closure closure = (Closure) dependency;
-                Task task = context.getTask();
-                Object taskProxy = wrapTaskWithDeprecatingProxy(task);
-                Object closureResult = closure.call(taskProxy);
+                Object closureResult;
+                try {
+                    closureResult = closure.call(new Object[]{ null });
+                } catch (NullPointerException e) {
+                    DeprecationLogger.deprecateAction("Accessing tasks provided to task dependency closures")
+                        .willBecomeAnErrorInGradle10()
+                        .withUpgradeGuideSection(9, "task_in_task_dependency_closure")
+                        .nagUser();
+                    Task task = context.getTask();
+                    closureResult = closure.call(task);
+                }
+
                 if (closureResult != null) {
                     queue.addFirst(closureResult);
                 }
@@ -165,39 +173,11 @@ public class DefaultTaskDependency extends AbstractTaskDependency {
                 formats.add("A TaskDependency instance");
                 formats.add("A Provider that represents a task output");
                 formats.add("A Provider instance that returns any of these types");
-                formats.add("A Closure instance that returns any of these types");
                 formats.add("A Callable instance that returns any of these types");
                 formats.add("An Iterable, Collection, Map or array instance that contains any of these types");
                 throw new UnsupportedNotationException(dependency, String.format("Cannot convert %s to a task.", dependency), null, formats);
             }
         }
-    }
-
-    private static Object wrapTaskWithDeprecatingProxy(@Nullable Task task) {
-        if (task == null) {
-            return null;
-        }
-        return Proxy.newProxyInstance(
-            task.getClass().getClassLoader(),
-            task.getClass().getInterfaces(),
-            (proxy, method, args) -> {
-                if ("getMetaClass".equals(method.getName()) && method.getParameterCount() == 0) {
-                    // Return a MetaClass for the proxy's own class, not the delegate's.
-                    // Groovy's invokedynamic call site asks for the MetaClass to determine
-                    // the argument type. If we delegate to the real task, Groovy will fail
-                    // due to the type mismatch.
-                    return org.codehaus.groovy.runtime.InvokerHelper.getMetaClass(proxy.getClass());
-                }
-
-                DeprecationLogger.deprecateAction("Accessing tasks provided to task dependency closures")
-                    .withContext("Cannot call method '" + method.getName() + "' on task passed to task dependency closure.")
-                    .willBecomeAnErrorInGradle10()
-                    .withUpgradeGuideSection(9, "task_in_task_dependency_closure")
-                    .nagUser();
-
-                return method.invoke(task, args);
-            }
-        );
     }
 
     private static void addAllFirst(Deque<Object> queue, Object[] items) {
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/DefaultTaskDependency.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/DefaultTaskDependency.java
@@ -173,6 +173,7 @@ public class DefaultTaskDependency extends AbstractTaskDependency {
                 formats.add("A TaskDependency instance");
                 formats.add("A Provider that represents a task output");
                 formats.add("A Provider instance that returns any of these types");
+                formats.add("A Closure instance that returns any of these types");
                 formats.add("A Callable instance that returns any of these types");
                 formats.add("An Iterable, Collection, Map or array instance that contains any of these types");
                 throw new UnsupportedNotationException(dependency, String.format("Cannot convert %s to a task.", dependency), null, formats);
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/DefaultTaskDependency.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/DefaultTaskDependency.java
@@ -25,10 +25,12 @@ import org.gradle.api.internal.provider.ProviderInternal;
 import org.gradle.api.internal.provider.ValueSupplier;
 import org.gradle.api.tasks.TaskDependency;
 import org.gradle.internal.Cast;
+import org.gradle.internal.deprecation.DeprecationLogger;
 import org.gradle.internal.typeconversion.UnsupportedNotationException;
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 
+import java.lang.reflect.Proxy;
 import java.nio.file.Path;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -112,7 +114,9 @@ public class DefaultTaskDependency extends AbstractTaskDependency {
                 ((TaskDependencyContainer) dependency).visitDependencies(context);
             } else if (dependency instanceof Closure) {
                 Closure closure = (Closure) dependency;
-                Object closureResult = closure.call(context.getTask());
+                Task task = context.getTask();
+                Object taskProxy = wrapTaskWithDeprecatingProxy(task);
+                Object closureResult = closure.call(taskProxy);
                 if (closureResult != null) {
                     queue.addFirst(closureResult);
                 }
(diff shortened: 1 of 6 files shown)


jvandort:
@bot-gradle test this

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -93,6 +93,39 @@ Problem severity can no longer be set explicitly when creating a new instance.
 Instead, it is determined by the reporting method: `ProblemReporter.report()` produces warnings and `ProblemReporter.throwing()` produces errors.
 Calling `.severity()` on `ProblemSpec` is now a no-op and will be removed in Gradle 10.0.
 
+[[task_in_task_dependency_closure]]
+==== Accessing Task in dependsOn Closure
+
+Accessing a task provided to closures passed to the link:{javadocPath}/org/gradle/api/Task.html#dependsOn(java.lang.Object...)[Task.dependsOn] method has been deprecated.
+Starting in Gradle 10.0, Gradle will no longer provide the Task as an argument to closures passed to the Task.dependsOn method.
+
+The following code will not be permitted starting in Gradle 10.0
```
lkasso, line 102:
```suggestion
The following code will not be permitted starting in Gradle 10:

```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -93,6 +93,39 @@ Problem severity can no longer be set explicitly when creating a new instance.
 Instead, it is determined by the reporting method: `ProblemReporter.report()` produces warnings and `ProblemReporter.throwing()` produces errors.
 Calling `.severity()` on `ProblemSpec` is now a no-op and will be removed in Gradle 10.0.
 
+[[task_in_task_dependency_closure]]
+==== Accessing Task in dependsOn Closure
+
+Accessing a task provided to closures passed to the link:{javadocPath}/org/gradle/api/Task.html#dependsOn(java.lang.Object...)[Task.dependsOn] method has been deprecated.
+Starting in Gradle 10.0, Gradle will no longer provide the Task as an argument to closures passed to the Task.dependsOn method.
+
```
lkasso, line 101: Can we just add a brief line about the Kotlin DSL/lambdas?

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -93,6 +93,39 @@ Problem severity can no longer be set explicitly when creating a new instance.
 Instead, it is determined by the reporting method: `ProblemReporter.report()` produces warnings and `ProblemReporter.throwing()` produces errors.
 Calling `.severity()` on `ProblemSpec` is now a no-op and will be removed in Gradle 10.0.
 
+[[task_in_task_dependency_closure]]
+==== Accessing Task in dependsOn Closure
+
+Accessing a task provided to closures passed to the link:{javadocPath}/org/gradle/api/Task.html#dependsOn(java.lang.Object...)[Task.dependsOn] method has been deprecated.
+Starting in Gradle 10.0, Gradle will no longer provide the Task as an argument to closures passed to the Task.dependsOn method.
```
lkasso, line 100:
```suggestion
Starting in Gradle 10, Gradle will no longer provide the `Task` as an argument to closures passed to the `Task.dependsOn` method.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -93,6 +93,39 @@ Problem severity can no longer be set explicitly when creating a new instance.
 Instead, it is determined by the reporting method: `ProblemReporter.report()` produces warnings and `ProblemReporter.throwing()` produces errors.
 Calling `.severity()` on `ProblemSpec` is now a no-op and will be removed in Gradle 10.0.
 
+[[task_in_task_dependency_closure]]
+==== Accessing Task in dependsOn Closure
+
+Accessing a task provided to closures passed to the link:{javadocPath}/org/gradle/api/Task.html#dependsOn(java.lang.Object...)[Task.dependsOn] method has been deprecated.
```
lkasso, line 99:
```suggestion
Accessing a task provided to closures passed to the link:{javadocPath}/org/gradle/api/Task.html#dependsOn(java.lang.Object...)[`Task.dependsOn`] method has been deprecated.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -93,6 +93,39 @@ Problem severity can no longer be set explicitly when creating a new instance.
 Instead, it is determined by the reporting method: `ProblemReporter.report()` produces warnings and `ProblemReporter.throwing()` produces errors.
 Calling `.severity()` on `ProblemSpec` is now a no-op and will be removed in Gradle 10.0.
 
+[[task_in_task_dependency_closure]]
+==== Accessing Task in dependsOn Closure
```
lkasso, line 97:
```suggestion
==== Accessing Task in `dependsOn` Closure
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -93,6 +93,39 @@ Problem severity can no longer be set explicitly when creating a new instance.
...
+
+Accessing a task provided to closures passed to the link:{javadocPath}/org/gradle/api/Task.html#dependsOn(java.lang.Object...)[Task.dependsOn] method has been deprecated.
+Starting in Gradle 10.0, Gradle will no longer provide the Task as an argument to closures passed to the Task.dependsOn method.
+
+The following code will not be permitted starting in Gradle 10.0
+[source,groovy]
```
lkasso, line 103: This coded example should also show how the users should update the build logic for Gradle 10. For the rare legitimate case where someone actually wanted that Task parameter, what should they do?

lkasso (approved):
I left a few suggestions in the upgrade guide.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
jvandort: This change does not affect the Kotlin DSL since we are deprecating functionality related to a Groovy closure. The task is not accessible from anywhere in Kotlin task dependency callbacks, unless you implement a Closure in Kotlin, which is very odd

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
jvandort: You can technically access the enclosing Task like this:
```
    def "can access enclosing task from within task dependency closure"() {
        buildFile << """
            def bar = tasks.register("bar")
            tasks.register("foo") {
                dependsOn {
                    println("Name: " + name)
                    bar
                }
            }
        """

        when:
        succeeds("foo")

        then:
        outputContains("Name: foo")
    }
```

Though I don't think we should suggest this. I can't really think of a use case for accessing the task-being-resolved inside a closure meant to describe what that task depends on

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
lkasso: Sounds fair. Feel free to ignore this comment then.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
lkasso: Ok that's fair. Let's go ahead and not add this.

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/DefaultTaskDependency.java
```diff
@@ -112,7 +114,9 @@ public void visitDependencies(final TaskDependencyResolveContext context) {
                 ((TaskDependencyContainer) dependency).visitDependencies(context);
             } else if (dependency instanceof Closure) {
                 Closure closure = (Closure) dependency;
-                Object closureResult = closure.call(context.getTask());
+                Task task = context.getTask();
```
jvandort: TODO: pass null, use try/catch, deprecate + call again with getTask
TODO: Test with Kotlin
TODO: Update Javadoc on Task.dependsOn

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/DefaultTaskDependency.java
```diff
@@ -161,7 +173,6 @@ public void visitDependencies(final TaskDependencyResolveContext context) {
                 formats.add("A TaskDependency instance");
                 formats.add("A Provider that represents a task output");
                 formats.add("A Provider instance that returns any of these types");
-                formats.add("A Closure instance that returns any of these types");
```
big-guy, line 164: This should stay because we still accept Closure, we just don't want to accept a Closure that uses the parameter.

subprojects/core-api/src/main/java/org/gradle/api/Task.java
```diff
@@ -114,10 +114,6 @@
  * <li>A {@code Callable}. The {@code call()} method may return any of the types listed here. Its return value is
  * recursively converted to tasks. A {@code null} return value is treated as an empty collection.</li>
  *
- * <li>A Groovy {@code Closure} or Kotlin function. The closure may take a {@code Task} as parameter.
```
big-guy, line 117: We still accept Closures and Kotlin functions.  You could argue that Closures are also a kind of Callable, but I would leave this here and we can make a separate pass to remove explicit mentions of Groovy Closures. 

I would just remove "The closure may take a Task as parameter"

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/DefaultTaskDependency.java
```diff
@@ -112,7 +113,18 @@ public void visitDependencies(final TaskDependencyResolveContext context) {
                 ((TaskDependencyContainer) dependency).visitDependencies(context);
             } else if (dependency instanceof Closure) {
                 Closure closure = (Closure) dependency;
-                Object closureResult = closure.call(context.getTask());
+                Object closureResult;
+                try {
+                    closureResult = closure.call(new Object[]{null});
+                } catch (NullPointerException e) {
+                    DeprecationLogger.deprecateAction("Accessing tasks provided to task dependency closures")
```
big-guy, line 120: This has two annoying problems.

_Any_ NPE will trigger this deprecation, not just one that relies on the parameter. So something like this would trigger it too:

```
foo {
  dependsOn { 
       def task = getSomeTaskOrNull()
       task.name // NPE sometimes
       task
  }
}
```

Plugins could be anticipating null and treating it special, this behavior would change immediately:
```
foo {
   dependsOn { 
      if (it != null) {
          return bar
      }
      return null
   }
}
```

I'm not convinced that either of these are actually problems.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -86,46 +86,46 @@ ProjectDependency dependency = project.getDependencyFactory().createProjectDepen
...
 
-There is no direct replacement for this method.
-If you were using it to reach tasks across project boundaries, <<artifact_views.adoc#artifact-views, Artifact Views>> let you select artifacts from dependent projects and let Gradle wire the producer tasks through normal dependency resolution.
+[source,groovy]
+----
+def foo = tasks.register("foo")
 
-[[deprecate_problem_spec_severity]]
-==== Deprecation of `ProblemSpec.severity()`
+// The below usages of dependsOn access the Task object provided as a parameter.
```
big-guy, line 108: I would explicitly say "This is deprecated behavior" on both of these.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -86,46 +86,46 @@ ProjectDependency dependency = project.getDependencyFactory().createProjectDepen
...
-Problem severity can no longer be set explicitly when creating a new instance.
-Instead, it is determined by the reporting method: `ProblemReporter.report()` produces warnings and `ProblemReporter.throwing()` produces errors.
-Calling `.severity()` on `ProblemSpec` is now a no-op and will be removed in Gradle 10.0.
+// Build logic may continue to pass closures to dependsOn without accessing the provided Task.
+tasks.register("okay") {
+    dependsOn {
+        foo
+    }
+}
+----
```
big-guy, line 129: We should consider a more realistic scenario and how you'd deal with it.

e.g., 
```

tasks.register("bar", Bar) {
    dependsOn { bar ->
        // deprecated
        if (bar.someProperty) {
            foo
        } else {
            []
        }
    }
}

tasks.register("bar", Bar) {
    dependsOn { bar ->
        // OK because this falls back to bar?
        if (someProperty) {
            foo
        } else {
            []
        }
    }
}
```

big-guy (changes requested):

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/DefaultTaskDependency.java
jvandort: Yup makes sense -- I was over eager to remove this

subprojects/core-api/src/main/java/org/gradle/api/Task.java
jvandort: Agreed

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/DefaultTaskDependency.java
jvandort: I think what we have now is the best solution to deprecate this behavior. I don't see any of these being big issues

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
jvandort: yes, we _could_ recommend this, but what would the user actually be doing here? 

Your suggested replacement is acceptable given the scenario, but if you are looking at the current task and making a decision about your dependencies, shouldn't you be using a Provider to wire these values?

jvandort:
@bot-gradle test this

big-guy (approved):

PR #37578, opened 2026-04-18
Fix issue with excluding test by class

### LINKED ISSUE #37539: `filter.excludeTest` doesn't filter JUnit5 nested test classes and excluding top-level class also excludes nested classes
### Current Behavior

A test in class `A$B` is not excluded when test task is configured with a `filter.excludeTest("A$B", null)`, while using `filter.excludeTest("A", null)` will exclude tests in `A$B` and `A$B$C` as well.
Running tasks `:testExcludeAll :testExcludeAllNested :testExcludeTopLevel :testExcludeFirstLevelNested :testExcludeSecondLevelNested` in the reproducer project gives following result:
```
> Task :testExcludeAll
===========================================
===========================================

> Task :testExcludeAllNested
===========================================
test()
firstLevelNestedTest()
secondLevelNestedTest()
===========================================

> Task :testExcludeTopLevel
===========================================
===========================================

> Task :testExcludeFirstLevelNested
===========================================
test()
firstLevelNestedTest()
secondLevelNestedTest()
===========================================

> Task :testExcludeSecondLevelNested
===========================================
test()
firstLevelNestedTest()
secondLevelNestedTest()
===========================================
```
Only `:testExcludeAll` behaves as expected. Compare to the output in https://github.com/gradle/gradle/issues/31304 where only `:testExcludeSecondLevelNested` gives unexpected results.

### Expected Behavior

`filter.excludeTest("A$B", null)` excludes tests in `A$B`
`filter.excludeTest("A", null)` excludes tests in `A`, but not in `A$B` and `A$B$C`

### Context (optional)

TeamCity uses excludes when [running tests in parallel](https://www.jetbrains.com/help/teamcity/parallel-tests.html) to determine which tests are run in which batch. Tests are included/excluded by test classes. When running Gradle 8, we made a workaround for https://github.com/gradle/gradle/issues/31304 which disabled tests that weren't excluded by Gradle.
After upgrading Gradle, if the test class `A` was excluded in batch 1, and test class `A$B` was excluded in batch 2, the tests from `A$B` didn't run in batch 1 since excluding `A` excluded them too, and didn't run in batch 2 since the workaround made up for ignored exclude entry. In the end, the tests from nested class weren't run at all.
Disabling the workaround helps, but
1) running the reproducer gives different results on different versions of Gradle (e. g. on 9.3.0 excluding nested classes works but excluding top-level class also excludes nested classes) so I'm not sure what will happen after next upgrade,
2) having tests run in the batch where the class is formally excluded is confusing when debugging.

[test-filter-bug-gradle-9.4.1.zip](https://github.com/user-attachments/files/26748489/test-filter-bug-gradle-9.4.1.zip)
The project is practically the same as provided in https://github.com/gradle/gradle/issues/31304, so the results can be compared, but modified to run on Gradle 9.

### Gradle version

9.4.1

8.11 (except for the bug https://github.com/gradle/gradle/issues/31304)

### Build scan URL (optional)

_No response_

_No response_

ghale:
Calling filter.excludeTest("SampleTest", null) didn't just exclude `SampleTest` — it also silently excluded all of its `@Nested` children.  Compounding this, the test that should have caught this regression was broken in such a way that it continued to pass even though the behavior was wrong. 
                                                                                                                                                                                                                 
Three interacting bugs were responsible:                                                                                                                          
   
- Scan-time over-pruning. Excluded outer classes were dropped before JUnit Platform discovery. Because JUnit Platform discovers `@Nested` children through their outer class as a selector, dropping the outer left no way to reach the nested tests.
- Ancestor walk-up re-inclusion. The post-discovery filter's walk up the descriptor chain treated any non-excluded ancestor as sufficient to include a test, so a nested  class exclude could be bypassed by its non-excluded enclosing class.                                                                                                      
- Fuzzy exclude matching. The exclude-match heuristic treated a pattern like SampleTest$NestedTestClass as also matching SampleTest.  This was accidentally patching over the walk-up bug in some cases.                                   
                                                            
Fixing any single bug in isolation moves the failures around, so a correct fix involved fixing all three together.                                                           
                                                            
The following fixes were made:
- The test that should have caught this regression was rewritten to be more resilient.                                                                                                                                                                    
- Stopped pruning excluded classes at scan time that have nested classes. Excluded outer classes remain available as discovery selectors; filtering moves to post-discovery where it can be correct.  
- Rewrote the walk-up so that an exclude on a nested class is never bypassed by its enclosing class, while unrelated ancestors may still re-include.                                                                                                                                                         
- Added an exact-match exclude query distinct from the existing fuzzy heuristic. The exact query is used where a definitive "this class is excluded" answer is required; the fuzzy one is retained for the JUnit 4 filter, which relies on it to drill into individual methods.                                                                    

Fixes #37539 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/ClassMethodNameFilter.java
+++ b/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/ClassMethodNameFilter.java
@@ -83,8 +83,8 @@ public final class ClassMethodNameFilter implements PostDiscoveryFilter {
     private boolean shouldRun(TestDescriptor descriptor, boolean checkingParent, ClassSource classSource) {
         String className = classSource.getClassName();
         if (matcher.matchesExcludeClass(className)) {
-            // This class exactly matches an exclude pattern (not merely a fuzzy may-match).
-            // Don't let children-recursion re-include the container. Ancestors that are
+            // This class exactly matches an exclude pattern.
+            // Return immediately to prevent children from re-including the container. Ancestors that are
             // themselves included by pattern (e.g. a test suite) are handled by classMatch.
             return false;
         }
--- a/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/ClassMethodNameFilter.java
+++ b/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/ClassMethodNameFilter.java
@@ -139,6 +139,9 @@ public final class ClassMethodNameFilter implements PostDiscoveryFilter {
     private boolean classMatch(TestDescriptor descriptor) {
         TestDescriptor current = descriptor;
         String methodName = null;
+        // Records the lowest-level class in the chain that matches an exclude pattern.
+        // Used to block re-inclusion via the walk-up only for enclosing-class ancestors;
+        // unrelated ancestors (e.g. test suites) can still re-include.
         String excludedClassName = null;
         while (true) {
             Optional<TestDescriptor> parent = current.getParent();
@@ -152,9 +155,11 @@ public final class ClassMethodNameFilter implements PostDiscoveryFilter {
                 if (excludedClassName == null && matcher.matchesExcludeTest(name, methodName)) {
                     excludedClassName = name;
                 }
-                boolean encloseExclude = excludedClassName != null
+                // True when this ancestor is an enclosing class of the excluded class
+                // (same class, or a $-parent), as opposed to an unrelated ancestor.
+                boolean ancestorEnclosesExclude = excludedClassName != null
                     && (excludedClassName.equals(name) || excludedClassName.startsWith(name + "$"));
-                if (!encloseExclude && matcher.matchesIncludeTest(name, methodName)) {
+                if (!ancestorEnclosesExclude && matcher.matchesIncludeTest(name, methodName)) {
                     return true;
                 }
             }
--- a/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/ClassMethodNameFilter.java
+++ b/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/ClassMethodNameFilter.java
@@ -73,7 +73,7 @@ public final class ClassMethodNameFilter implements PostDiscoveryFilter {
         while (current != null) {
             Optional<String> enclosingClassName = className(current);
             if (enclosingClassName.isPresent()) {
-                return !matcher.matchesExcludeClass(enclosingClassName.get());
+                return !matcher.matchesExcludeClassExactly(enclosingClassName.get());
             }
             current = current.getParent().orElse(null);
         }
@@ -82,7 +82,7 @@ public final class ClassMethodNameFilter implements PostDiscoveryFilter {
 
     private boolean shouldRun(TestDescriptor descriptor, boolean checkingParent, ClassSource classSource) {
         String className = classSource.getClassName();
-        if (matcher.matchesExcludeClass(className)) {
+        if (matcher.matchesExcludeClassExactly(className)) {
             // This class exactly matches an exclude pattern.
             // Return immediately to prevent children from re-including the container. Ancestors that are
             // themselves included by pattern (e.g. a test suite) are handled by classMatch.
--- a/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/ClassMethodNameFilter.java
+++ b/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/ClassMethodNameFilter.java
@@ -54,24 +54,40 @@ public final class ClassMethodNameFilter implements PostDiscoveryFilter {
 
     private boolean shouldRun(TestDescriptor descriptor, boolean checkingParent) {
         Optional<TestSource> source = descriptor.getSource();
-        if (!source.isPresent()) {
-            return true;
+        if (source.isPresent()) {
+            TestSource testSource = source.get();
+            if (testSource instanceof MethodSource) {
+                return shouldRun(descriptor, (MethodSource) testSource);
+            }
+            if (testSource instanceof ClassSource) {
+                return shouldRun(descriptor, checkingParent, (ClassSource) testSource);
+            }
         }
 
-        TestSource testSource = source.get();
-        if (testSource instanceof MethodSource) {
-            return shouldRun(descriptor, (MethodSource) testSource);
+        // Source is absent or of a custom type (e.g. ArchUnit field-based tests).
+        // Walk up to the first ancestor with a class source and honor its exclude status:
+        // if that enclosing class exactly matches an exclude pattern, this descriptor is also
+        // excluded (as a member of the class). Otherwise default to included (original behavior
+        // preserved — the filter's status quo for custom sources is inclusive).
+        TestDescriptor current = descriptor.getParent().orElse(null);
+        while (current != null) {
+            Optional<String> enclosingClassName = className(current);
+            if (enclosingClassName.isPresent()) {
+                return !matcher.matchesExcludeClass(enclosingClassName.get());
+            }
+            current = current.getParent().orElse(null);
         }
-
-        if (testSource instanceof ClassSource) {
-            return shouldRun(descriptor, checkingParent, (ClassSource) testSource);
-        }
-
-        Optional<TestDescriptor> parent = descriptor.getParent();
-        return parent.isPresent() && shouldRun(parent.get(), true);
+        return true;
     }
 
     private boolean shouldRun(TestDescriptor descriptor, boolean checkingParent, ClassSource classSource) {
+        String className = classSource.getClassName();
+        if (matcher.matchesExcludeClass(className)) {
+            // This class exactly matches an exclude pattern (not merely a fuzzy may-match).
+            // Don't let children-recursion re-include the container. Ancestors that are
+            // themselves included by pattern (e.g. a test suite) are handled by classMatch.
+            return false;
+        }
         Set<? extends TestDescriptor> children = descriptor.getChildren();
         if (!checkingParent) {
             for (TestDescriptor child : children) {
@@ -81,7 +97,6 @@ public final class ClassMethodNameFilter implements PostDiscoveryFilter {
             }
         }
         if (children.isEmpty()) {
-            String className = classSource.getClassName();
             return matcher.matchesTest(className, null)
                 || matcher.matchesTest(className, descriptor.getLegacyReportingName());
         }
--- a/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/ClassMethodNameFilter.java
+++ b/platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/ClassMethodNameFilter.java
@@ -101,20 +101,45 @@ public final class ClassMethodNameFilter implements PostDiscoveryFilter {
             .isPresent();
     }
 
+    /**
+     * Walks up the descriptor chain looking for an ancestor class whose name matches an include
+     * pattern. An ancestor-level include may re-include a descendant, but an exclude on a nested
+     * class must not be bypassed by walking up to its enclosing class.
+     *
+     * <p>An ancestor is an <em>enclosing class</em> of the excluded class when the excluded
+     * class's fully qualified name starts with the ancestor's name followed by {@code $}
+     * (e.g. {@code SampleTest} encloses {@code SampleTest$NestedTestClass}). Re-inclusion
+     * via an enclosing ancestor is forbidden. Re-inclusion via an unrelated ancestor
+     * (e.g. a JUnit Platform suite that references a separate excluded class) is allowed.
+     *
+     * <p>This asymmetry is intentional:
+     * <ul>
+     *   <li>{@code --tests Parent} should pull in tests from {@code Parent$Nested} transitively.</li>
+     *   <li>{@code excludeTest("Parent$Nested", null)} must not be bypassed because the
+     *       enclosing class {@code Parent} happens not to match the exclude.</li>
+     *   <li>{@code excludeTestsMatching("Foo")} excludes standalone {@code Foo} tests, but
+     *       should still allow {@code Foo} tests to run when wrapped in an unrelated suite.</li>
+     * </ul>
+     */
     private boolean classMatch(TestDescriptor descriptor) {
         TestDescriptor current = descriptor;
         String methodName = null;
+        String excludedClassName = null;
         while (true) {
-
             Optional<TestDescriptor> parent = current.getParent();
             if (!parent.isPresent()) {
                 break;
             }
 
-            // If the current descriptor is a class, check if it matches the test selection criteria
             Optional<String> className = className(current);
             if (className.isPresent()) {
-                if (matcher.matchesTest(className.get(), methodName)) {
+                String name = className.get();
+                if (excludedClassName == null && matcher.matchesExcludeTest(name, methodName)) {
+                    excludedClassName = name;
+                }
+                boolean encloseExclude = excludedClassName != null
+                    && (excludedClassName.equals(name) || excludedClassName.startsWith(name + "$"));
+                if (!encloseExclude && matcher.matchesIncludeTest(name, methodName)) {
                     return true;
                 }
             }
--- a/platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
+++ b/platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
@@ -37,13 +37,12 @@ import static org.apache.commons.lang3.StringUtils.substringAfterLast;
  *       class possibly contribute tests that match the includes? Used to prune scanning; false
  *       positives are acceptable, false negatives are not.</li>
  *   <li><strong>Combined test-level check</strong> ({@link #matchesTest(String, String)}) — the
- *       AND of {@link #matchesIncludeTest(String, String)} and the negation of
+ *       result of {@link #matchesIncludeTest(String, String)} ANDed with the negation of
  *       {@link #matchesExcludeTest(String, String)}. What most callers want for a leaf test.</li>
  *   <li><strong>Separated include / exclude queries</strong>
  *       ({@link #matchesIncludeTest(String, String)} / {@link #matchesExcludeTest(String, String)}
  *       / {@link #matchesExcludeClass(String)}) — lets callers reason about include and exclude
- *       semantics independently. Needed by walkers that must distinguish "excluded at this
- *       descriptor" from "include matched via a non-excluded ancestor".</li>
+ *       semantics independently.</li>
  * </ul>
  *
  * <p>Pattern interpretation:
@@ -77,8 +76,8 @@ class ClassTestSelectionMatcher {
     }
 
     /**
-     * Returns true if the given (className, methodName) pair matches the include patterns and is not excluded by the
-     * exclude patterns.
+     * {@return true if the given (className, methodName) pair matches the include patterns and is not excluded by the
+     * exclude patterns}
      */
     public boolean matchesTest(String className, @Nullable String methodName) {
         return matchesIncludeTest(className, methodName) && !matchesExcludeTest(className, methodName);
@@ -152,8 +151,7 @@ class ClassTestSelectionMatcher {
      *
      * <p>Unlike {@link #matchesExcludeTest(String, String)}, this does <em>not</em> consult
      * the fuzzy {@code mayMatchClass} heuristic: pattern {@code Parent} does not match
-     * class {@code Parent$Nested} here. Used by callers that need a definitive
-     * "is this class excluded?" answer rather than a "might this class match?" hint.</p>
+     * class {@code Parent$Nested} here.</p>
      */
     public boolean matchesExcludeClass(String className) {
         for (ClassTestPattern pattern : buildScriptExcludePatterns) {
--- a/platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
+++ b/platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
@@ -29,17 +29,34 @@ import static org.apache.commons.lang3.StringUtils.splitPreserveAllTokens;
 import static org.apache.commons.lang3.StringUtils.substringAfterLast;
 
 /**
- * This class has two responsibilities:
+ * Class-and-method pattern matcher used by {@link TestSelectionMatcher}.
  *
+ * <p>Separates three kinds of queries:
  * <ul>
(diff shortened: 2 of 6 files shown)


ghale:
@bot-gradle test this

platforms/jvm/testing-jvm-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/filters/ClassMethodNameFilter.java
```diff
@@ -54,24 +54,40 @@ private boolean shouldRun(TestDescriptor descriptor) {
...
-        return parent.isPresent() && shouldRun(parent.get(), true);
+        return true;
     }
 
     private boolean shouldRun(TestDescriptor descriptor, boolean checkingParent, ClassSource classSource) {
+        String className = classSource.getClassName();
+        if (matcher.matchesExcludeClass(className)) {
+            // This class exactly matches an exclude pattern (not merely a fuzzy may-match).
+            // Don't let children-recursion re-include the container. Ancestors that are
+            // themselves included by pattern (e.g. a test suite) are handled by classMatch.
```
octylFractal, line 88: 💅 This comment seems a little off, wdyt?
```suggestion
            // This class exactly matches an exclude pattern.
            // Return immediately to prevent children from re-including the container. Ancestors that are
            // themselves included by pattern (e.g. a test suite) are handled by classMatch.
```

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -29,17 +29,34 @@
...
  * <ul>
- * <li>Judge whether a test class might be included. For example, class 'org.gradle.Test' can't
- * be included by pattern 'org.apache.Test'
- * <li>Judge whether a test method is matched exactly.
+ *   <li><strong>Permissive scan-time check</strong> ({@link #mayIncludeClass(String)}) — may this
+ *       class possibly contribute tests that match the includes? Used to prune scanning; false
+ *       positives are acceptable, false negatives are not.</li>
+ *   <li><strong>Combined test-level check</strong> ({@link #matchesTest(String, String)}) — the
+ *       AND of {@link #matchesIncludeTest(String, String)} and the negation of
+ *       {@link #matchesExcludeTest(String, String)}. What most callers want for a leaf test.</li>
```
octylFractal, line 41:
```suggestion
 *       result of {@link #matchesIncludeTest(String, String)} ANDed with the negation of
 *       {@link #matchesExcludeTest(String, String)}. What most callers want for a leaf test.</li>
```

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -29,17 +29,34 @@
...
 /**
- * This class has two responsibilities:
+ * Class-and-method pattern matcher used by {@link TestSelectionMatcher}.
  *
+ * <p>Separates three kinds of queries:
  * <ul>
- * <li>Judge whether a test class might be included. For example, class 'org.gradle.Test' can't
- * be included by pattern 'org.apache.Test'
- * <li>Judge whether a test method is matched exactly.
+ *   <li><strong>Permissive scan-time check</strong> ({@link #mayIncludeClass(String)}) — may this
```
octylFractal, line 36: ❓ Why is this separated from the "Separated include / exclude queries"?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -29,17 +29,34 @@
...
+ *       class possibly contribute tests that match the includes? Used to prune scanning; false
+ *       positives are acceptable, false negatives are not.</li>
+ *   <li><strong>Combined test-level check</strong> ({@link #matchesTest(String, String)}) — the
+ *       AND of {@link #matchesIncludeTest(String, String)} and the negation of
+ *       {@link #matchesExcludeTest(String, String)}. What most callers want for a leaf test.</li>
+ *   <li><strong>Separated include / exclude queries</strong>
+ *       ({@link #matchesIncludeTest(String, String)} / {@link #matchesExcludeTest(String, String)}
+ *       / {@link #matchesExcludeClass(String)}) — lets callers reason about include and exclude
+ *       semantics independently. Needed by walkers that must distinguish "excluded at this
+ *       descriptor" from "include matched via a non-excluded ancestor".</li>
```
octylFractal, line 46: ❌ I think the extra sentence here is just noisy.
```suggestion
 *       semantics independently.</li>
```

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -29,17 +29,34 @@
...
  * </ul>
  *
- * In both cases, if the pattern starts with an upper-case letter, it will be used to match
- * the simple class name;
- * otherwise, it will be used to match the fully qualified class name.
+ * <p>Pattern interpretation:
+ * <ul>
+ *   <li>A pattern that starts with an upper-case letter matches the simple class name.</li>
```
octylFractal, line 51: 💭 Maybe not for now, but we should formalize exactly what locale we use to determine "upper-case", as currently it's based on the default locale. It should probably just use ASCII?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -29,17 +29,34 @@
...
+ *   <li>Otherwise, it matches the fully qualified class name.</li>
+ *   <li>{@link #matchesExcludeTest(String, String)} applies a fuzzy {@code mayMatchClass}
+ *       heuristic at the class level (when method name is null), so callers that iterate
+ *       into individual methods can still find exact exclude matches without giving up the
+ *       entire class prematurely.</li>
```
octylFractal, line 56: ❓ I don't quite understand this, if it only applies when `null`, how does iterating into individual methods relate? What does "giving up the entire class prematurely" mean in this context?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -29,17 +29,34 @@
...
  * </ul>
  *
- * In both cases, if the pattern starts with an upper-case letter, it will be used to match
- * the simple class name;
- * otherwise, it will be used to match the fully qualified class name.
+ * <p>Pattern interpretation:
+ * <ul>
+ *   <li>A pattern that starts with an upper-case letter matches the simple class name.</li>
+ *   <li>Otherwise, it matches the fully qualified class name.</li>
+ *   <li>{@link #matchesExcludeTest(String, String)} applies a fuzzy {@code mayMatchClass}
```
octylFractal, line 53: :x: This should be explaining exactly what "fuzzy" means here. Referencing `mayMatchClass` is pointless in Javadoc since it's private, so someone viewing the Javadoc can't read the implementation.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -59,17 +76,38 @@ private static List<ClassTestPattern> prepareClassBasedPatternList(Collection<St
             .collect(Collectors.toList());
     }
 
-
+    /**
+     * Returns true if the given (className, methodName) pair matches the include patterns and is not excluded by the
+     * exclude patterns.
```
octylFractal, line 81: 💅 For this and others, this pattern gives a proper `@return` with the same text otherwise:
```suggestion
     * {@return true if the given (className, methodName) pair matches the include patterns and is not excluded by the
     * exclude patterns}
```

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -59,17 +76,38 @@ private static List<ClassTestPattern> prepareClassBasedPatternList(Collection<St
...
-            && !matchesExcludePattern(className, methodName);
+            && matchesPattern(commandLineIncludePatterns, className, methodName);
+    }
+
+    /**
+     * Returns true if the given (className, methodName) pair matches any exclude pattern.
+     *
+     * <p>An empty exclude set returns false. Include patterns are not consulted.</p>
+     */
+    public boolean matchesExcludeTest(String className, @Nullable String methodName) {
```
octylFractal, line 104: :x: I think this method's Javadoc should mention the special behavior called out in the class Javadoc.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -125,6 +147,23 @@ private boolean matchesExcludePattern(String className, @Nullable String methodN
         return matchesClassAndMethod(buildScriptExcludePatterns, className, methodName);
     }
 
+    /**
+     * Returns true if the class name exactly matches an exclude pattern's class component.
+     *
+     * <p>Unlike {@link #matchesExcludeTest(String, String)}, this does <em>not</em> consult
+     * the fuzzy {@code mayMatchClass} heuristic: pattern {@code Parent} does not match
+     * class {@code Parent$Nested} here. Used by callers that need a definitive
+     * "is this class excluded?" answer rather than a "might this class match?" hint.</p>
```
octylFractal, line 156: :x: I don't think this extra sentence is needed.
```suggestion
     * class {@code Parent$Nested} here.</p>
```

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
```diff
@@ -125,6 +147,23 @@ private boolean matchesExcludePattern(String className, @Nullable String methodN
...
+     */
+    public boolean matchesExcludeClass(String className) {
```
octylFractal, line 158: 💭 Perhaps this should be called `matchesExcludeClassExactly`? Or some other term to differentiate it more.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcher.java
```diff
@@ -24,22 +24,31 @@
 import java.util.Collections;
 
 /**
- * This class has three public APIs:
+ * Entry point for matching test descriptors against include and exclude patterns.
  *
+ * <p>Provides the following queries:
```
octylFractal, line 29: :x: I don't think listing these here is useful. Javadoc already lists all the methods with a short summary.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcher.java
```diff
@@ -81,6 +90,36 @@ public boolean matchesTest(String className, @Nullable String methodName) {
         return classTestSelectionMatcher.matchesTest(className, methodName);
     }
 
+    /**
+     * Returns true if the given class and method matches the include patterns (build-script and command-line).
```
octylFractal, line 94: :x: Mentioning "build-script and command-line" feels like a violation of the abstraction here. This class doesn't even own those fields, they're in the `classTestSelectionMatcher`.

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcherTest.groovy
```diff
@@ -257,6 +258,71 @@ class TestSelectionMatcherTest extends Specification {
         ['org.gradle.FooTest*'] | ['*org.gradle.BarTest*']        | 'org.gradle.FooTest'  | true
     }
 
+    def "matchesIncludeTest ignores exclude patterns"() {
```
octylFractal, line 261: :x: This test name is confusing, it contains cases with no `excludes` at all.

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcherTest.groovy
```diff
@@ -257,6 +258,71 @@ class TestSelectionMatcherTest extends Specification {
...
+        where:
+        includes           | excludes           | className         | methodName       | match
+        []                 | []                 | "FooTest"         | "aaa"            | true
+        []                 | ["FooTest"]        | "FooTest"         | "aaa"            | true
+        ["FooTest"]        | []                 | "FooTest"         | "aaa"            | true
+        ["FooTest"]        | ["FooTest"]        | "FooTest"         | "aaa"            | true
+        ["FooTest"]        | []                 | "BarTest"         | "aaa"            | false
+    }
+
+    def "matchesIncludeTest combines build-script and command-line includes"() {
```
octylFractal, line 274: :x: Should be explicit about combining meaning "AND"
```suggestion
    def "matchesIncludeTest returns the AND of build-script and command-line includes"() {
```

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcherTest.groovy
```diff
@@ -257,6 +258,71 @@ class TestSelectionMatcherTest extends Specification {
...
+        buildScript        | commandLine       | className         | match
+        []                 | []                | "FooTest"         | true
+        ["FooTest"]        | []                | "FooTest"         | true
+        []                 | ["FooTest"]       | "FooTest"         | true
+        ["FooTest"]        | ["FooTest"]       | "FooTest"         | true
+        ["FooTest"]        | ["BarTest"]       | "FooTest"         | false
+        ["FooTest"]        | ["BarTest"]       | "BarTest"         | false
+    }
+
+    def "matchesExcludeTest ignores include patterns"() {
```
octylFractal, line 288: :x: This also contains tests with no `includes`, which doesn't make sense to me.

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcherTest.groovy
```diff
@@ -257,6 +258,71 @@ class TestSelectionMatcherTest extends Specification {
...
+        []                 | ["FooTest"]                | "BarTest"                        | null               | false
+        ["BarTest"]        | ["FooTest"]                | "BarTest"                        | null               | false
+        []                 | ["FooTest.doThing"]        | "FooTest"                        | "doThing"          | true
+        ["BarTest"]        | ["FooTest.doThing"]        | "FooTest"                        | "doThing"          | true
+        []                 | ["FooTest.doThing"]        | "FooTest"                        | "doOther"          | false
+        ["BarTest"]        | ["FooTest.doThing"]        | "FooTest"                        | "doOther"          | false
+    }
+
+    @Issue("https://github.com/gradle/gradle/issues/37539")
+    def "matchesExcludeClass is exact and does not treat parent pattern as matching nested class"() {
```
octylFractal, line 306: :x: This seems like it should be two tests, one for the exactness, and one for the nested class handling.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
ghale: Because `mayIncludeClass` is used fundamentally differently than the others.  This is used at scan time to do a cheap "should we even look at this class at all" check.  In other words, "if we already know the includes don't match the class, we can just omit it from the set of classes we send to the worker".  We don't look at excludes or try to understand the hierarchy of nested tests or anything, we just exclude things that don't match the includes.  Then at execution time, in the worker, the separate include/exclude queries are used to do a more exact check of whether a given test that was discovered should be executed or not.   

I think it makes sense for this to be called out separately.  Do you disagree?

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
ghale: I don't _think_ this can occur with JUnit5, but for JUnit4, you can potentially create tests that have `ClassSource` descriptors without children.  However, they can stuff the "method name" into legacyReportingName, so you have a bunch of `ClassSource` descriptors with different legacy names.  So we return false only when 1) there are no excludes, 2) method is not null and we don't have an exact match on method and class name or 3) method _is_ null, but we're sure the class can't still match an exclude on a method name.

If we returned false when the method is null and we think it could still match an exclude based on method name, we would "accept" the test and never get to the check against legacy reporting name (which would exclude it).

I find this all hard to reason about (lots of double negatives) and the verbiage in the javadoc is admittedly confusing.  I've attempted to write this more clearly.  Let me know if you think it needs further polishing.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
ghale: See https://github.com/gradle/gradle/pull/37578#discussion_r3125064208.  I've rewritten this - hopefully it's a little better now.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
ghale: Good point.  Done.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
ghale: Yeah, I like that.  Done.  Same for `TestSelectionMatcher`.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcher.java
ghale: Fair enough.  Removed.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcher.java
ghale: Removed.

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcherTest.groovy
ghale: Well, the idea was to show that it behaves the same with excludes and without, but I see your point.  I've removed the non-exclude iterations and added a few more iterations to be comprehensive.

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcherTest.groovy
ghale: Done.

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcherTest.groovy
ghale: Same as above.  I' ve split this into two tests - one to test `matchesExcludeTest()` in isolation and another to test just that it ignores excludes.

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcherTest.groovy
ghale: I think I prefer seeing all the cases together where it's clear what will and will not be excluded given each exclude.

However, I don't feel strongly about it.  If you do, then I don't mind changing this.  But if not, I'd rather just keep it the way it is.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
ghale: Agreed, but I think this is outside of the scope of this change.

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
ghale: Maybe as a follow up we could ask an agent to analyze this and try to simplify it all.  I suspect it _can_ be simplified, but I don't currently see a clear path to doing so.  I think this should be done as follow up work, though, and we should confine the scope of this PR as much as possible (i.e. it already introduces some substantial changes).

platforms/software/testing-base-infrastructure/src/main/java/org/gradle/api/internal/tasks/testing/filter/ClassTestSelectionMatcher.java
octylFractal: I guess it's fine. I think I take issue with the name being too close but that's not new in this PR, so I'll leave it.

platforms/software/testing-base-infrastructure/src/test/groovy/org/gradle/api/internal/tasks/testing/filter/TestSelectionMatcherTest.groovy
octylFractal: It's mostly that I find `true`/`false` in these tests to be hard to read. If you want to point out what will and won't match, I think it would be clearer with `excludePattern | matchingClassNames | unmatchingClassNames`, so each `excludePattern` is just listed once.

ghale: Great idea.  I think it makes these tests much easier to reason about.  I've updated all of the tests to use this pattern since this is a good general improvement and it keeps the pattern consistent.

octylFractal (approved):

PR #37579, opened 2026-04-18
Consolidate build action execution against a build-tree

alllex:
### Context

The session-to-tree transition for build action execution is currently spread across a chain of single-purpose `BuildSessionActionExecutor` decorators (`RunAsWorkerThreadBuildActionExecutor` → `RunAsBuildOperationBuildActionExecutor` → `BuildTreeLifecycleBuildActionExecutor`), making the execution path hard to follow and `LauncherServices` bloated with pass-through dependencies.

Additionally, `BuildTreeState` — the lifecycle owner of the tree's service registry — is injected into every `AbstractBuildState` subclass just to access the tree's services. This conflates service access with lifecycle ownership.

There is also a latent bug: a prior refactoring made the cleanup-failure combination branch dead code, since `result` is always `null` when `BuildTreeState.close()` throws.

### Approach

Introduce a `BuildTreeActionExecutor` service (scoped to `BuildSession`) that owns the full session-to-tree transition — worker-thread acquisition, build-operation wrapping, tree lifecycle, and failure handling — in one place. Introduce `BuildTreeServices` as a thin wrapper so build states access the tree's services without depending on its lifecycle.

### Changes

- Add `BuildTreeActionExecutor` interface; collapse the three decorator classes into `BuildTreeLifecycleBuildActionExecutor`
- Delete `RunAsWorkerThreadBuildActionExecutor` and `RunAsBuildOperationBuildActionExecutor`
- Add `BuildTreeServices`; replace `BuildTreeState` usage in `AbstractBuildState` and all concrete build states/factories
- Remove `BuildTreeState` from the service registry (it stays as internal lifecycle owner only)
- Fix the dead-code bug in cleanup-failure handling
- Null-mark `org.gradle.launcher.exec`


alllex:

alllex:

platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/LauncherServices.java
```diff
@@ -170,14 +157,7 @@ BuildSessionActionExecutor createActionExecutor(
...
                     virtualFileSystem,
-                    new RunAsWorkerThreadBuildActionExecutor(
-                        workerLeaseService,
-                        new RunAsBuildOperationBuildActionExecutor(
-                            new BuildTreeLifecycleBuildActionExecutor(buildModelParametersFactory, buildLayoutValidator, valueSnapshotter, options),
-                            buildOperationRunner,
-                            loggingBuildOperationProgressBroadcaster,
-                            buildOperationNotificationValve
-                        ))));
+                    buildTreeActionExecutor));
```
alllex, line 160: ℹ️  This is the main change: what was previously some wrappers around the build session action executor, is now a more explicitly defined abstraction of `BuildTreeActionExecutor` dedicated to running an action against the build tree.

platforms/core-runtime/build-state/src/main/java/org/gradle/internal/buildtree/BuildTreeActionExecutor.java
```diff
@@ -0,0 +1,73 @@
...
+     * exceptions happens at the session level, while result-based failure logging happens
+     * within the build tree scope.
+     *
+     * @param action the build action to execute
+     * @param buildSessionServices the build-session-scoped service registry,
+     *     used as the parent for build-tree services
+     * @return the result of executing the action, containing either a client result
+     *     or a build failure
+     */
+    BuildActionRunner.Result runBuildTreeAction(BuildAction action, ServiceRegistry buildSessionServices);
```
alllex, line 71: ℹ️  This now represents a deeper abstraction of running an action against a build tree in a self-contained manner.

platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/BuildTreeLifecycleBuildActionExecutor.java
```diff
@@ -23,88 +23,118 @@
...
 import org.gradle.tooling.internal.provider.serialization.SerializedPayload;
+import org.jspecify.annotations.Nullable;
 
 import java.util.function.Supplier;
 
-/**
- * A {@link BuildActionExecutor} responsible for establishing the build tree for a single invocation of a {@link BuildAction}.
- */
-public class BuildTreeLifecycleBuildActionExecutor implements BuildSessionActionExecutor {
+public class BuildTreeLifecycleBuildActionExecutor implements BuildTreeActionExecutor {
```
alllex: ℹ️  We consolidate the complexity of running the action against a build tree in a single place to avoid the additional cognitive load of understanding separate pieces that are always arranged in the same way and are executed unconditionally.

subprojects/core/src/main/java/org/gradle/internal/build/AbstractBuildState.java
```diff
@@ -43,22 +43,19 @@ public abstract class AbstractBuildState implements BuildState, Closeable {
     private final Lazy<ProjectStateRegistry> projectStateRegistry;
     private final Lazy<BuildWorkGraphController> workGraphController;
 
-    public AbstractBuildState(BuildTreeState buildTree, BuildDefinition buildDefinition, @Nullable BuildState parent) {
+    public AbstractBuildState(BuildTreeServices buildTreeServices, BuildDefinition buildDefinition, @Nullable BuildState parent) {
```
alllex, line 47: ℹ️  Instead of requiring a direct access to a conceptually broader "build tree state", we trim it down to only requiring the build-tree service registry to use it as a parent of the build-scoped services.

In the future, we want to avoid the build state directly instantiating its services. It's not possible today, because the BuildState instance is also registered as one of the services and there are entanglements with ProjectBuilder (https://github.com/gradle/gradle/pull/37581)

platforms/core-runtime/build-state/src/main/java/org/gradle/internal/buildtree/BuildTreeActionExecutor.java
```diff
@@ -0,0 +1,73 @@
...
+
+package org.gradle.internal.buildtree;
+
+import org.gradle.internal.invocation.BuildAction;
+import org.gradle.internal.service.ServiceRegistry;
+import org.gradle.internal.service.scopes.Scope;
+import org.gradle.internal.service.scopes.ServiceScope;
+
+/**
+ * Responsible for executing a {@link BuildAction} within a freshly constructed build tree.
```
6hundreds: 💅 This `freshly` is a bit confusing me as a reader, it mentally applies a restriction on freshness of the build tree.

6hundreds (approved):
LGTM!

platforms/core-runtime/build-state/src/main/java/org/gradle/internal/buildtree/BuildTreeActionExecutor.java
alllex: What if we simplify this part to just `against a build-tree` ?
I think maybe the docstring that is already on the method would be enough to compensate for this.

6hundreds: Makes sense

PR #37581, opened 2026-04-19
Untangle ProjectBuilder lifecycle from Build-scoped services

alllex:
This is a follow-up for
- https://github.com/gradle/gradle/pull/35915

ProjectBuilder-created projects have a different lifecycle compared to regular invocation.
For instance, they don't have a settings phase, which normally houses the lifecycle of project registration.
To circumvent this, a hatch was maintained to allow ProjectBuilder to register custom Build-scope services,
where it would put services that normally live only in the Settings-scope, like ProjectDescriptorRegistry

We remove the hatch, making the vanilla Build-scope services setup more straightforward.
We also make it much more explicit in the ProjectBuilderImpl that the settings-scope services require special care.


alllex:

platforms/extensibility/unit-test-fixtures/src/main/java/org/gradle/testfixtures/internal/ProjectBuilderImpl.java
```diff
@@ -350,15 +343,31 @@ public IncludedBuildInternal getModel() {
...
 
-        @Override
-        public File getBuildRootDir() {
-            return getBuildServices().get(BuildDefinition.class).getBuildRootDir();
+    /**
+     * ProjectBuilder has its own simplified lifecycle that compresses all operational service scopes
+     * into a single operation. It also skips the Settings phase entirely, so settings-specific
+     * services (e.g. {@link ProjectDescriptorRegistry}) must be created and managed manually.
+     */
+    private record Control(
```
jvandort: I don't find `Control` to very well express the purpose of this class. IMO something like `RootProjectContext` may be more fitting

jvandort (approved):

PR #37585, opened 2026-04-19
Remove getDependenciesForInternalUse

jvandort:
getDependenciesForInternalUse resolved a TaskDependencyContainer adhoc using a new CachingTaskDependencyResolveContext.
Further, it always resolved a TaskDependencyContainer to the set of Tasks it contains and ignores all other work. In most cases, this behavior is undesirable.
    
We instead introduce a new TaskDependencyUtil.newTaskResolver method which explicitly defines this operation as first creating a new resolver and then using it to resolve tasks. We also add documentation describing that this operation should only be used to implement existing public APIs which assume task graphs only contain tasks.
    
Along the way:
- We improve type safety by enforcing that a CachingTaskDependencyResolveContext accepts TaskDependencyContainer instances instead of any Object instance. Same for TaskDependencyResolver, which delegates to a CachingTaskDependencyResolveContext.
- We inline AbstractTaskDependency into DefaultTaskDependency, which was the only class that extended AbstractTaskDependency.
- We remove TaskDependencyContainerInternal in favor of TaskDependencyInternal
    
Now, the contract for defining task dependenices is much more clear. Task dependencies are always defined by a TaskDependencyContainer.
Objects which resolve task dependencies into concrete sets always accept a TaskDependencyContainer. In some cases we discover a TaskDependency object which does not implement a TaskDependencyContainer. These scenarios are handled adhoc in the rare case they arise. Eventually this behavior should be deprecated.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

platforms/core-configuration/file-collections/src/main/java/org/gradle/api/internal/file/AbstractFileCollection.java
```diff
@@ -96,9 +96,7 @@ protected void appendContents(TreeFormatter formatter) {
     @Override
     public final TaskDependency getBuildDependencies() {
         assertCanCarryBuildDependencies();
-        DefaultTaskDependency result = taskDependencyFactory.visitingDependencies(context -> context.add(AbstractFileCollection.this));
-        result.setToStringProvider(() -> "Dependencies of " + getDisplayName());
```
big-guy, line 99: Where does this name show up?

jvandort: Calling toString on the task dependency object returned by `Buildable.taskDependencies`

subprojects/core/src/main/java/org/gradle/api/DefaultTask.java
```diff
@@ -228,7 +228,7 @@ public Task mustRunAfter(Object... paths) {
     }
 
     @Override
-    public TaskDependency getMustRunAfter() {
```
big-guy, line 232: I don't think these should change since it leaks an internal type on a public type.

Can we keep the signature the same (I think this means changing AbstractTask and TaskInternal back too) and perform ugly casts elsewhere?

big-guy (dismissed):
I don't think I have enough knowledge to judge the overall change or the history on why we added getDependenciesForInternalUse initially.

We need to get @h0tk3y or @mlopatkin or someone else from cortex to agree to these changes.

subprojects/core/src/main/java/org/gradle/api/DefaultTask.java
jvandort: My justification here is we already expose `TaskDependencyInternal getTaskDependencies()` and these are used in the same way

big-guy: getTaskDependencies() is a long standing mistake too, but one we're forced to keep because of AbstractTask.  We shouldn't add more internal APIs on public types if we can avoid it.

big-guy:
Punting to 9.7 since we need input from Cortex.  Obviously, we can pull this back if we hear something next week.

subprojects/core/src/main/java/org/gradle/api/DefaultTask.java
jvandort: Removed changes to DefaultTask. Cast internally instead.

jvandort:
Based on git history, getTaskDependenciesForInternal use is to allow internal code to avoid an IP violation for code that wanted to call getDependencies. getDependencies is a public method and can return tasks from other projects.

jvandort:
@bot-gradle test this

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/WorkDependencyResolver.java
```diff
@@ -35,12 +34,14 @@ public interface WorkDependencyResolver<T> {
      */
     WorkDependencyResolver<Task> TASK_AS_TASK = new WorkDependencyResolver<Task>() {
         @Override
+        @AllowUsingApiForExternalUse
         public boolean resolve(Task originalTask, Object node, Action<? super Task> resolveAction) {
             if (node instanceof TaskDependency) {
+                assert !(node instanceof TaskDependencyContainer) : "TaskDependencyContainer should be handled by CachingTaskDependencyResolveContext";
```
6hundreds, line 40: 🤔   The fact, that this invariant lives in a different class and not in `CachingTaskDependencyResolveContext` is confusing. However, I see this code path is around to die and restructure the walker perhaps is a waste in this case.

platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/idea/internal/IdeaModuleMetadata.java
```diff
@@ -51,7 +51,8 @@ public File getFile() {
     }
 
     @Override
-    public Set<? extends Task> getGeneratorTasks() {
-        return Collections.singleton(generatorTask.get());
+    public void visitDependencies(TaskDependencyResolveContext context) {
+        context.add(generatorTask.get());
```
6hundreds: ❓ `EclipseProjectMetadata` passes task provider. Do we want to be consistent here?

platforms/native/plugins-model-native/src/main/java/org/gradle/language/base/plugins/ComponentModelBasePlugin.java
```diff
@@ -218,7 +219,7 @@ public CheckForNotBuildableBinariesAction(List<BinarySpecInternal> notBuildable)
 
             @Override
             public void execute(Task task) {
-                Set<? extends Task> taskDependencies = TaskDependencyUtil.getDependenciesForInternalUse(task.getTaskDependencies(), task);
+                Set<? extends Task> taskDependencies = TaskDependencyUtil.newTaskResolver().getDependencies(task, (TaskDependencyInternal) task.getTaskDependencies());
```
6hundreds: 💅 We have ~7 instances of this sort of cast
`TaskDependencyUtil.newTaskResolver().getDependencies((TaskDependencyContainer/TaskDependencyInternal) task.getTaskDependencies())`
Maybe some static helpers on `TaskDependencyUtil` would reduce it?

subprojects/core/src/main/java/org/gradle/execution/plan/DefaultExecutionPlan.java
```diff
@@ -136,6 +136,12 @@ private void addEntryTasks(Collection<? extends Task> tasks, int ordinal) {
         doAddEntryNodes(nodes, ordinal);
     }
 
+    @Override
+    public void addEntryWork(TaskDependencyContainer dependencies) {
```
6hundreds, line 140: 💅 Do we have a test for this API? Say, in `DefaultExecutionPlanTest` or somewhere else?

6hundreds (approved):
Good stuff. No one of my comments is a blocker I believe. I'd like to see some static helpers to reduce amount of casting, tho

platforms/native/plugins-model-native/src/main/java/org/gradle/language/base/plugins/ComponentModelBasePlugin.java
jvandort: All these casts are a direct result of sterling's comment: https://github.com/gradle/gradle/pull/37585#discussion_r3169750566

The change in this PR to make `CachingTaskDependencyResolveContext` operate on `TaskDependencyContainer` is correct, and a `TaskInternal` _should_ provide that type. But, `TaskInternal` is not really internal as it leaks to public API via `DefaultTask`, and we don't have a "more internal" solution. Casting is a requirement. 

I've created a `TaskInternalUtil` that consolidates these casts into a well-known common location.

platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/idea/internal/IdeaModuleMetadata.java
jvandort: Yeah good call. Passing the provider is cleaner here.

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/tasks/WorkDependencyResolver.java
jvandort: To me it is confusing that this work resolver handles `TaskDependency` at all. In practice, after this PR, all Gradle created task dependency instances _are_ a `TaskDependencyInternal`, as a `TaskDependencyInternal` extends a `TaskDependencyContainer`. 

This is what the comment below specifically calls out. This condition should not even exist. The assert you commented on is just there to ensure the comment below is actually correct.

jvandort: I'm not sure there is anything to do here other than deprecating custom use-provided implementations of `TaskDependency`.

subprojects/core/src/main/java/org/gradle/execution/plan/DefaultExecutionPlan.java
jvandort: Added a unit test for this

PR #37589, opened 2026-04-20
Dependency verification keyring cleanup

### LINKED ISSUE #37275: Improve the dependency verification clean up process
### Expected Behavior

The [current documentation](https://docs.gradle.org/current/userguide/dependency_verification.html#cleaning_up_the_verification_file) about cleaning up dependency verification metadata focuses on the `verification-metadata.xml` file.

We should add a feature that allows user to prune their keyrings once they have performed the manual clean up of the verification metadata file.

While the documentation does not mention it, one can remove entries from the armored key ring by hand. But for the binary one, there is no such option.

An alternative is to remove the keyring entirely and re-generate it. But it can be a very time consuming process as keyservers may heavily rate limit requests.

### Context

Manual and hard because one must match long key IDs to full key IDs, or automated and possibly very slow because of keyserver limitations.

ljacomet:
This adds a `--prune-keys` CLI flag that causes Gradle to remove keys from the saved keyrings if they are not part of the dependency verification metadata.

Fixes #37275

### DIFF
--- a/platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
+++ b/platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
@@ -84,6 +84,7 @@ Performance:
 Security:
   --dependency-verification, -F      Configures the dependency verification mode. Supported values are 'strict', 'lenient', or 'off'.
   --export-keys                      Exports the public keys used for dependency verification.
+  --prune-keys                       Prunes the dependency verification keyring files so they only contain keys referenced by the verification metadata file.
   --refresh-keys                     Refreshes the public keys used for dependency verification.
   --update-locks                     Performs a partial update of the dependency lock. Allows passed-in module notations to change version. [incubating]
   --write-locks                      Persists dependency resolution for locked configurations. Ignores existing locking information if it exists.
--- a/platforms/core-runtime/daemon-messaging/src/main/java/org/gradle/tooling/internal/provider/action/BuildActionSerializer.java
+++ b/platforms/core-runtime/daemon-messaging/src/main/java/org/gradle/tooling/internal/provider/action/BuildActionSerializer.java
@@ -167,6 +167,7 @@ public class BuildActionSerializer {
             encoder.writeString(startParameter.getDependencyVerificationMode().name());
             encoder.writeBoolean(startParameter.isRefreshKeys());
             encoder.writeBoolean(startParameter.isExportKeys());
+            encoder.writeBoolean(startParameter.isPruneKeys());
             encoder.writeString(startParameter.getWelcomeMessageConfiguration().getWelcomeMessageDisplayMode().name());
             encoder.writeBoolean(startParameter.isPropertyUpgradeReportEnabled());
             encoder.writeBoolean(startParameter.isProblemReportGenerationEnabled());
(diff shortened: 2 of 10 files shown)


platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
```diff
@@ -84,6 +84,7 @@ Performance:
 Security:
   --dependency-verification, -F      Configures the dependency verification mode. Supported values are 'strict', 'lenient', or 'off'.
   --export-keys                      Exports the public keys used for dependency verification.
+  --prune-keys                       Prunes the dependency verification keyring files so they only contain keys referenced by the verification metadata file.
   --refresh-keys                     Refreshes the public keys used for dependency verification.
```
big-guy, line 88: Instead of adding a new CLI option, could we expand what "refresh keys" means? Or does `--export-keys` effectively do what `--prune-keys` would do already?

If I understand correctly...
`--export-keys` - write a key ring file based on what's in the metadata verification file. If a key cannot be retrieved, it's not written or checked for again for 24 hours.
`--refresh-keys` - do what `--export-keys` does again and ignore the 24 hour cache and add keys to the existing file. Again, if a key cannot be retrieved, it's not written.
`--prune-keys` - Keep the existing key ring file, but remove any keys that are not found in the metadata verification file. 

If that's right, you could run `--export-keys` and you'd get pruning behavior, but you could also remove required keys that are temporarily unavailable.  We could make export a little more complicated:
- If no key ring exists, do what we do today.
- If a key ring _does_ exist, look at the key ring before looking externally.

I think this would accomplish the same thing. `--export-keys` would always produce a key ring based on the keys required by the build.

lkasso (approved):
Docs LGTM (if we don't change anything). Please re-request review if you change it based on @big-guy's comment.

platforms/core-runtime/launcher/src/testFixtures/groovy/org/gradle/launcher/cli/HelpFixture.groovy
ljacomet: 🤔 

From looking at history and feature, this is slightly different:
* `--refresh-keys` is a command used so that a key cached as missing (for 24h) can be rechecked. It can be used with or without exported keys, but would have no impact if your configuration says `<key-servers enabled="false"/>`
* `--export-keys` is a command that can be used standalone to export additively all the keys referenced in the metadata file, without validating or adding dependencies.
* `--export-keys` can also be used while writing the dependency verification metadata file, it then forces `key-servers` to be enabled. In that usage, it can be combined with `--refresh-keys` to recheck for missing keys.

The proposal you make misses one point: As a user of the feature, I want to only add extra keys. Their presence is not a problem. If I need to trust them again later, I still have them around.

To be clear: my favorite answer is having a way to evolve Gradle CLI flags so that something like `--export-keys` remain valid, but it can also become `--export-keys=prune` or `--export-keys prune`.
But we don't have that support. And unless we mandate the use of `=`, I don't think we can add it because of parsing determinism.

But I see your point that having 3 options tied to key handling for verification can feel like a lot.

ljacomet: Another idea, but to be honest I don't like it much:
Move the prune/append decision to the `verification-metadata.xml` configuration. And then `--export-keys` behavior changes based on that setting, with `append` being the default, and `prune` an option.

@big-guy Could you clarify the nature of your comment: what I proposed must change? Or where you just looking to see if there is a better alternative?

ljacomet:
Moved back to draft, looking at doing this differently

ljacomet:
Closing, decision has been made to not add a CLI argument for this

PR #37590, opened 2026-04-20
Run perf tests on Java 21 by default

blindpirate:
We currently run all performance test against Java 17. It's time to run them with more recent Java version. We tried Java 21 and 25, and found that [groovy 2.5.22](https://github.com/gradle/gradle/blob/master/testing/internal-performance-testing/src/main/groovy/org/gradle/performance/generator/FileContentGenerator.groovy#L47) doesn't support Java 25 yet.


blindpirate:
@bot-gradle test APT

Issue #24498, opened 2023-03-25
Communication between client and daemon is broken due to file encoding issues on s390x


Crosswind:
### Expected Behavior
When running a simple `gradle init` a daemon should start and I should be presented with the regular options when running init.

### Current Behavior
The client fails with a `GradleException: Could not start Gradle daemon.`
The stack trace that is produced with the --stacktrace option:
```
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] FAILURE: Build failed with an exception.
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] 
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] * What went wrong:
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] Could not start Gradle daemon.
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] 
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] * Try:
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] > Run with --scan to get full insights.
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] 
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] * Exception is:
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] org.gradle.api.GradleException: Could not start Gradle daemon.
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] 	at org.gradle.launcher.daemon.client.DefaultDaemonStarter.startProcess(DefaultDaemonStarter.java:185)
2023-03-24T18:42:08.145-0500 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] 	at org.gradle.launcher.daemon.client.DefaultDaemonStarter.startDaemon(DefaultDaemonStarter.java:137)
[...]

PR #24779, opened 2023-04-16
Default to file.encdoing=UTF-8 on s390x arch

Crosswind:
On an IBM Mainframe (s390x architecture) the default file.encoding of the JVM is IBM-1048 (EBCDIC). Gradle expects the JVM to run in UTF-8. Issues appear if either the client or the daemon JVM run in EBCDIC.

Instead of changing the code it makes most sense to make sure the JVM is started in UTF-8 on s390x as well.

Fixes #24498

### Context
As described in #24498 this is needed to be able to start Gradle on an IBM Mainframe (s390x architecture). The JVM by default runs in EBCDIC (IBM-1047). Gradle expects to run the JVM in UTF-8 because some of the communication between client and daemon relies on it.

This change allows to run Gradle on s390x without any manual modifications or knowledge about setting the JVM option which will ease usage and (hopefully) contribute a wider usage in general.

### Contributor Checklist
- [x] [Review Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md)
- [x] Make sure that all commits are [signed off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff) to indicate that you agree to the terms of [Developer Certificate of Origin](https://developercertificate.org/).
- [x] Make sure all contributed code can be distributed under the terms of the [Apache License 2.0](https://github.com/gradle/gradle/blob/master/LICENSE), e.g. the code was written by yourself or the original code is licensed under [a license compatible to Apache License 2.0](https://apache.org/legal/resolved.html).
- [x] Check ["Allow edit from maintainers" option](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) in pull request so that additional changes can be pushed by Gradle team
- [ ] Provide integration tests (under `<subproject>/src/integTest`) to verify changes from a user perspective
- [ ] Provide unit tests (under `<subproject>/src/test`) to verify logic
- [ ] Update User Guide, DSL Reference, and Javadoc for public-facing changes
- [ ] Ensure that tests pass sanity check: `./gradlew sanityCheck`
- [ ] Ensure that tests pass locally: `./gradlew <changed-subproject>:quickTest`

### Gradle Core Team Checklist
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation

PR #37596, opened 2026-04-20
Barebones support for starting Gradle on z/OS without compromise

lux01:
#24498 

### Context

IBM Java can perform implicit conversion of process stdin/stdout/stderr streams which interferes with the daemon bootstrap by causing the parent and daemon processes to disagree about the character encoding needed for the initial handshake message. Furthermore, there's inconsistency about the character set used to encode this handshake message which can lead to even more disagreement between the two processes.

This is especially problematic on z/OS where the standard stream conversion is to/from EBCDIC codepages, but the daemon process always starts with `file.encoding=UTF-8`. Users on z/OS have historically resorted to forcing  `-Dfile.encoding=UTF-8 -Dconsole.encoding=UTF-8` but this is not a long term solution.

With this change, a user on z/OS can use gradle without forcing the encodings. I've managed to verify this on the z/OS machine I have access to and can build a simple Java project. I've unsuccessfully tried to get gradle itself to build on z/OS, but the user ID I have access to has a too low ulimit for file handles to be able to actually compile all the plugins needed... the daemon starts at least!

This fix works by directly accessing the file descriptor for `stdin` rather than using `System.in`, which will have been wrapped in some sort of `FilterInputStream` in IBM JVMs.

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-execution/worker-main/src/main/java/org/gradle/process/internal/worker/GradleWorkerMain.java
+++ b/platforms/core-execution/worker-main/src/main/java/org/gradle/process/internal/worker/GradleWorkerMain.java
@@ -19,7 +19,10 @@ package org.gradle.process.internal.worker;
 import org.gradle.internal.classloader.FilteringClassLoader;
 import org.gradle.internal.stream.EncodedStream;
 
+import java.io.BufferedInputStream;
 import java.io.DataInputStream;
+import java.io.FileDescriptor;
+import java.io.FileInputStream;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


ljacomet:
Thank you for the proposed contribution.

Turning the daemon to use UTF-8 always is a breaking change and thus cannot be accepted as-is right now, see:
* #2270

However upon following breadcrumbs, we might have made a mistake in refusing:
* #24779

Would such a change resolve the issue for you? That is a similar change for z/OS

The changes related to the OperatingSystem enum should be OK on their own, if they bring something useful.

lux01:
Thanks for your respones @ljacomet, I didn't think that this would be a breaking change as the UTF-8 usage is limited solely to the encoding used to specify the encoding to use on both sides of the initial daemon handshake. It doesn't change the default encoding of anything else from what I can tell. The rest of the changes are to directly access the underlying `stdin` stream which should side-step the automatic conversion feature in the IBM JVM.

I'll go back and double check with `GRADLE_OPTS=-Dfile.encoding=UTF-8`. It wasn't working for me in the past which is why I ended up going down this rabbit hole...

PR #37606, opened 2026-04-20
Introduce DomainObjectCollection.getElements

### LINKED ISSUE #28043: Add Provider<Collection<T>> getElements() to DomainObjectCollection to lazy map/transform its elements
### Expected Behavior

```kotlin
val myDomains = objects.domainObjectSet(String::class)
val outputProvider = objects.listProperty(String::class)

// actual, there is no method, but you could accidentally use the `map` method of Kotlin, which is wrong
outputProvider.set(myDomains.map { "$it$it" })// this `map` is part of Kotlin stdlib Collection.map which maps the realized elements, and it is not a provider

// expected
outputProvider.set(myDomains.elements.map { it.map { "$it$it" }}) // lazily transform the elements

myDomains.add("foo")
println(outputProvider.get()) // expected "[foofoo]"

// works with
inline val <reified T : Any> DomainObjectCollection<T>.elements: Provider<out Collection<T>> get() = project.provider {
    toList()
}


No lazy method to transform a `DomainObjectCollection` without realizing it. You can write an `getElemets` function by yourself easily, but I expect a built-in solution.

There is also already a `FileCollection.getElements()` to support the same use-case, lazily mapping its content.

### Context

I want to pass a `DomainObjectCollection` of an extension to a Java SourceSet. The domain of the collection contains a task provider as member, so I need to map the collection to its member. 
The SourceSet supports `Provider<Iterable<TaskProvider>` as input, but there is no lazy option to map the collection. If you accidental use `map` of the Kotlin stdlib, you are realizing the collection early, resulting into a empty list, because the domains of the collection are added lazily later (using `register`). 

So you need to get the elements of the collection lazily as Provider.

### PR DESCRIPTION
**Pull request by jvandort**
This allows getting a Provider<Collection<T>> of the elements within a DOC, without realizing the container. The returned provider carries the build dependencies of the providers added to the container via `addLater` and `addAllLater`.

Fixes https://github.com/gradle/gradle/issues/28043

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DelegatingDomainObjectSet.java
+++ b/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DelegatingDomainObjectSet.java
@@ -130,6 +130,7 @@ public class DelegatingDomainObjectSet<T> implements DomainObjectSet<T>, DomainO
 
     @Override
     public Provider<? extends Collection<T>> getElements() {
+        onMethodCall("getElements()");
         return delegate.getElements();
     }
 
--- a/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DelegatingDomainObjectSet.java
+++ b/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DelegatingDomainObjectSet.java
@@ -129,6 +129,11 @@ public class DelegatingDomainObjectSet<T> implements DomainObjectSet<T>, DomainO
     }
 
     @Override
+    public Provider<? extends Collection<T>> getElements() {
+        return delegate.getElements();
+    }
+
+    @Override
     public boolean add(T o) {
         onMethodCall("add(Object)");
         return getDelegate().add(o);
--- a/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/collections/AbstractIterationOrderRetainingElementSource.java
+++ b/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/collections/AbstractIterationOrderRetainingElementSource.java
@@ -34,6 +34,8 @@ import org.gradle.api.internal.provider.Collectors.ElementsFromCollectionProvide
 import org.gradle.api.internal.provider.Collectors.SingleElement;
 import org.gradle.api.internal.provider.Collectors.TypedCollector;
 import org.gradle.api.internal.provider.ProviderInternal;
+import org.gradle.api.internal.tasks.TaskDependencyContainer;
+import org.gradle.api.internal.tasks.TaskDependencyResolveContext;
 import org.gradle.api.specs.Spec;
 import org.gradle.internal.Cast;
 import org.jspecify.annotations.Nullable;
@@ -47,7 +49,8 @@ import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Set;
 
-abstract public class AbstractIterationOrderRetainingElementSource<T> implements ElementSource<T> {
+abstract public class AbstractIterationOrderRetainingElementSource<T> implements ElementSource<T>, TaskDependencyContainer {
+
     // This set represents the order in which elements are inserted to the store, either actual
     // or provided.  We construct a correct iteration order from this set.
     private final List<Element<T>> inserted = new ArrayList<>();
@@ -226,6 +229,13 @@ abstract public class AbstractIterationOrderRetainingElementSource<T> implements
             element.clearCache();
         }
         return added;
+    }
+
+    @Override
+    public void visitDependencies(TaskDependencyResolveContext context) {
+        for (Element<T> element : inserted) {
+            context.add(element.getProducer());
+        }
     }
 
     @Override
@@ -447,4 +457,5 @@ abstract public class AbstractIterationOrderRetainingElementSource<T> implements
             return index;
         }
     }
+
 }
--- a/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/collections/AbstractIterationOrderRetainingElementSource.java
+++ b/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/collections/AbstractIterationOrderRetainingElementSource.java
@@ -26,6 +26,7 @@ import it.unimi.dsi.fastutil.ints.IntSets;
 import org.gradle.api.Action;
 import org.gradle.api.internal.DefaultMutationGuard;
 import org.gradle.api.internal.MutationGuard;
+import org.gradle.api.internal.lambdas.SerializableLambdas;
 import org.gradle.api.internal.provider.ChangingValue;
 import org.gradle.api.internal.provider.CollectionProviderInternal;
 import org.gradle.api.internal.provider.Collector;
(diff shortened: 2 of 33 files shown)

### CODE REVIEW

**Comment by jvandort** [comment]
@bot-gradle test this

**Comment by jvandort** [comment]
@bot-gradle test ast

**Comment by jvandort** [comment]
@bot-gradle test ast

**Review by lkasso — APPROVED** [review]
Docs LGTM. It might be nice to add the "motivation" behind adding this API (maybe the example I added can reflect better how our users should use this API).

File: platforms/core-configuration/domain-object-collections/src/testFixtures/groovy/org/gradle/api/internal/AbstractDomainObjectCollectionSpec.groovy [hunk 1]
```diff
@@ -1945,6 +1949,52 @@ abstract class AbstractDomainObjectCollectionSpec<T> extends Specification {
         toList(container) == [a, b, c]
     }
 
+    def "can get elements of a container"() {
```
octylFractal: ❓ Can there also be a test for if you call `getElements()`, add something to the original DOC via `addAllLater`, then call `get()` on the elements, it still works? Just to ensure it's properly being lazy.

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/collections/AbstractIterationOrderRetainingElementSource.java
```diff
@@ -47,7 +49,8 @@
 import java.util.NoSuchElementException;
 import java.util.Set;
 
-abstract public class AbstractIterationOrderRetainingElementSource<T> implements ElementSource<T> {
+abstract public class AbstractIterationOrderRetainingElementSource<T> implements ElementSource<T>, TaskDependencyContainer {
```
asodja: ❓ Could we avoid `TaskDependencyContainer` at this level and move it `TaskDependencyContainer` one level above where `AbstractIterationOrderRetainingElementSource` is used? 

It feels like `AbstractIterationOrderRetainingElementSource` is relatively "generic" collection, but we are adding to it Task context. It feels like `AbstractIterationOrderRetainingElementSource` shouldn't know about Tasks

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/collections/SortedSetElementSource.java
```diff
@@ -245,6 +247,25 @@ public boolean removePendingCollection(CollectionProviderInternal<T, ? extends I
...
+                // dependencies if elements are realized early. To fix this, we would need
+                // to retain the pending elements for the lifetime of this element source.
+                for (Collectors.TypedCollector<T> collector : pending) {
+                    context.add(collector.getProducer());
+                }
+            },
+            Cast.uncheckedCast(SortedSet.class),
+            () -> {
+                realizePending();
+                return values;
```
asodja: ❓    Supposedly this leaks original TreeSet while others return immutable snapshot, is that a problem?

So if you call `tasks.elements.get().clear()` is it possible to clear this tree?

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/collections/SortedSetElementSource.java
```diff
@@ -245,6 +247,25 @@ public boolean removePendingCollection(CollectionProviderInternal<T, ? extends I
         return removeByProvider(provider);
     }
 
+    @Override
+    public ProviderInternal<SortedSet<T>> getElements() {
+        return new BuildableBackedProvider<>(
+            context -> {
+                // TODO: As this is implemented, the returned provider will "lose" build
```
asodja: ❓ According to AI this loses dependencies in:
  - Anything that iterates the container (contains, iterator(), a plugin looping over configurations at configuration time — extremely common) drains pending and silently drops all build dependencies.
  - Calling elements.get() itself realizes pending — so querying the provider once at configuration time destroys its own future dependency information.

So this may be a foot gun, how could we fix this in the future?

asodja: I think we should mention this to public Javadoc of  `getElements`, so users are aware of this. Maybe something like:
```
The returned provider retains the build dependencies of providers added by {@link #addLater(Provider)} or {@link #addAllLater(Provider)}.
However, if a pending element has already been realized — for example, because this collection was iterated or otherwise queried — its build dependencies may no longer be carried by the returned provider.
```

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DelegatingDomainObjectSet.java
```diff
@@ -128,6 +128,11 @@ public void addAllLater(Provider<? extends Iterable<T>> provider) {
         getDelegate().addAllLater(provider);
     }
 
+    @Override
+    public Provider<? extends Collection<T>> getElements() {
+        return delegate.getElements();
```
asodja: ❓ Do we need to call `onMethodCall("getElements()")` as in other methods. I guess others use it to check IP constraints

asodja (changes requested):

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/collections/SortedSetElementSource.java
jvandort: Good catch. Generally its best practice that providers contain immutable values. I've updated this to return a copy of the underling tree set.

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/collections/SortedSetElementSource.java
jvandort: So `getElements` is generally useful in certain scenarios when you have something in the existing API that is modeled by a DOC but should be modeled as a `SetProperty`. The biggest examples of these today are:
- `PublishArtifact`
- `PublicationArtifact` 

Querying which artifacts a configuration publishes requires task dependencies, as the artifacts may come from the output of a task.

The point being that for cases where there _were_ task dependencies, querying the container before those tasks are executed should be illegal. You are asking for the value of a provider before the task that produced that value has been executed. 

Your examples of iterating over configurations at configuration time is okay. Configurations never have task dependencies -- configurations are not produced by tasks. Same goes for most configuration-time constructs like tasks, source sets, dependencies. None of these should ever have task dependencies.

jvandort: As for fixing this in the future, we could update `SortedSetElementSource` to not clear its `pending` set when those providers are realized.

However I think it would be better to lean more the other way that if you _have_ providers with task dependencies, we should make it illegal to query the value of the provider early. 

In any case, we can improve the documentation.

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DelegatingDomainObjectSet.java
jvandort: Yes, good catch

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/collections/AbstractIterationOrderRetainingElementSource.java
jvandort: > one level above where AbstractIterationOrderRetainingElementSource is used?

Can you be more precise with what you mean here? 

> It feels like AbstractIterationOrderRetainingElementSource is relatively "generic" collection

It is a relatively generic collection -- of potentially eager and lazy elements. 

Looking at the interface it implements, `ElementSource`, it has method like `boolean addPending(ProviderInternal<? extends T> provider);`

Well, `ProviderInternal`, a generic producer of some value extends `TaskDependencyContainer`. 

> It feels like AbstractIterationOrderRetainingElementSource shouldn't know about Tasks

A provider carries a piece of data and a reference to the work that may need to be performed to _produce_ that piece of data. The reference to work on a work graph is modeled by a `TaskDependencyContainer`. Note, `Task` in TDC is a pretty poor name now, as this type is better named something like `WorkDependencyContainer`.

Any type that "knows" about work or the work graph generally implements this type. The two most important ones being:
- `FileCollectionInternal`
- `ProviderInternal` (and thus all of our collection properties)

IMO it makes sense that our generic collections of things, where those things may contain values produced by work in the work graph, are also collections of work.

I would argue the opposite, in that it would make more sense for `ElementSource` to be a TDC, as that is more consistent with our other generic containers

subprojects/core-api/src/main/java/org/gradle/api/DomainObjectCollection.java
```diff
@@ -54,6 +54,35 @@ public interface DomainObjectCollection<T> extends Collection<T> {
...
+     * The value produced by a provider with build dependencies, by definition,
+     * is not valid until those build dependencies are executed. For example, if a task
+     * produces an output file, the value of a provider returning that file is not valid
+     * until the task executes.
+     * <p>
+     * Querying the contents of this provider before the build dependencies of providers
+     * added by {@link #addLater(Provider)} and {@link #addAllLater(Provider)} will result
+     * in undefined behavior. In general, the returned provider should only be used to
+     * wire values between tasks. Avoid directly calling {@link Provider#get()} or any other
+     * method that realizes the value of the returned provier.
```
asodja: 💅 Typo `provier`

asodja (approved):
LGTM

subprojects/core-api/src/main/java/org/gradle/api/DomainObjectCollection.java
jvandort:
```suggestion
     * method that realizes the value of the returned provider.
```

jvandort:
@bot-gradle test this

PR #37618, opened 2026-04-21
Support precompiled script plugins KotlinDSL IP model building

6hundreds:
This PR adds precompiled scripts support to KDSL IP model builder and fix sourcePath computation for buildscripts.
- Precompiled scripts wasn't supported before. This PR is not adding any new logic of computation precompiled scripts classpath or so, it reuses vintage implementation bits. IP compatible discovery of projects scripts was implemented [before](https://github.com/gradle/gradle/blob/dc2206026547c2c4bcac0646dc7ded5037c5235f/platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinDslScriptsModelBuilder.kt#L193-L199).
- SourcePath for buildscripts [are hierarchical](https://github.com/gradle/gradle/blob/34a75f15027340be92d88923d19fe68f6e997ea4/platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt#L384-L386) in vintage. This PR adds hierarchy to IP model building also. [The test](https://github.com/gradle/gradle/blob/ca7472cc73dcf8422d0ed9422afa28699b9ae18b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiKotlinDslIntegrationTest.groovy#L62) exercises it.
- Added bunch of tests for KDSL model building for composite builds. When we [fetch KDSL model](https://github.com/gradle/gradle/blob/ca7472cc73dcf8422d0ed9422afa28699b9ae18b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiKotlinDslIntegrationTest.groovy#L38) via TAPI, we fetch it only for root build. For the picture completeness we need to fetch it for all editable builds
- Added a smoke test that fetches KDSL IP model for G/G build

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive integration tests for Kotlin DSL scripts with isolated projects mode, including composite build scenarios.
  * Added smoke test validating Kotlin DSL script models remain consistent across isolated and standard modes.

* **Refactor**
  * Refactored implicit import derivation logic for precompiled script plugins into reusable helper functions.


testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsKotlinDslSmokeTest.groovy
```diff
@@ -0,0 +1,83 @@
...
+import org.gradle.integtests.fixtures.executer.IntegrationTestBuildContext
+import org.gradle.tooling.GradleConnector
+import org.gradle.tooling.ProjectConnection
+import org.gradle.tooling.model.kotlin.dsl.KotlinDslScriptModel
+import org.gradle.tooling.model.kotlin.dsl.KotlinDslScriptsModel
+import org.gradle.util.internal.ToBeImplemented
+
+class GradleBuildIsolatedProjectsKotlinDslSmokeTest extends AbstractGradleceptionSmokeTest {
+
+    @ToBeImplemented
```
alllex: 💅  Does this take a parameter where we can put the issue link?

alllex (approved):

6hundreds:
@bot-gradle test this

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
```diff
@@ -160,21 +165,24 @@ fun buildScriptModelsInHierarchy(
...
-    fun collect(models: IsolatedScriptsModel) {
+    fun collect(models: IsolatedScriptsModel, parentSourcePath: ClassPath) {
         for (childScriptModel in models.models) {
-            outputModels[childScriptModel.scriptFile] = buildOutputModel(base, childScriptModel)
+            outputModels[childScriptModel.scriptFile] = buildOutputModel(base, childScriptModel, parentSourcePath)
         }
     }
 
-    fun visit(projectState: ProjectState) {
+    fun visitChildren(projectState: ProjectState, parentSourcePath: ClassPath) {
```
6hundreds, line 179: ℹ️ `visit` -> `visitChildren` has changed for clarity.
`parentSourcePath` is being propagated in the hierarchy now, similarly as vintage KDSL model building does

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
```diff
@@ -329,13 +344,29 @@ fun buildScriptModelFor(project: ProjectInternal): IntermediateScriptModel? {
...
+    val sourceSets = project.sourceSets ?: return emptyList()
+    val metadataDir = project.precompiledScriptPluginsMetadataDir
+
+    val classPathBySourceSet = mutableMapOf<String, ClassPath>()
+    val pluginSpecImports = metadataDir.implicitPluginSpecBuildersImports
+
+    return scripts.mapNotNull { scriptFile ->
+        val sourceSet = sourceSets.find { scriptFile in it.allSource } ?: return@mapNotNull null
+        val classPath = classPathBySourceSet.getOrPut(sourceSet.name) { DefaultClassPath.of(sourceSet.compileClasspath) }
+        val accessorImports = metadataDir.implicitAccessorsImports(scriptFile)
```
6hundreds, line 353: ℹ️ We're reusing vintage precompiled script models calculation logic for each project. `project.discoverPrecompiledScriptPluginScripts()` above produces scripts for the project in IP compatible manner

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt
```diff
@@ -172,10 +173,23 @@ fun Project.findSourceSetOf(file: File): EnclosingSourceSet? =
...
+        resolve("plugin-spec-builders").resolve(kotlinDslPluginSpecBuildersImplicitImports)
+    ) + implicitImportsFrom(
+        // Gradle <= 8.12 was using this other name with a dash but this was incompatible with moving to kotlin-scripting-host API
+        // Keeping it for compatibility with previous Gradle versions
+        resolve("plugin-spec-builders").resolve("implicit-imports")
+    )
+
+internal
+fun MetadataDir.implicitAccessorsImports(scriptFile: File): List<String> =
+    implicitImportsFrom(resolve("accessors").resolve(hashOf(scriptFile)))
```
6hundreds: ℹ️ Making it reusable in the IP model builder

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/KotlinDslModelChecker.groovy
```diff
@@ -0,0 +1,64 @@
...
+
+class KotlinDslModelChecker {
+
+    static void checkKotlinDslScriptModel(KotlinDslScriptModel actual, KotlinDslScriptModel expected) {
+        checkModel(actual, expected, [
+            { withNormalizedAccessorHash(it.classPath) },
+            { withNormalizedAccessorHash(it.sourcePath) },
+            { it.implicitImports },
+            // TODO:isolated support editor reports
+//            [{ it.editorReports }, { a, e -> checkEditorReport(a, e) }],
```
6hundreds, line 48: ℹ️ It's not really hard to support in IP editorReports as well, I've failed to create a test covering this.

6hundreds:
@coderabbitai review

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/FetchKotlinDslScriptsModelForAllBuilds.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Map;
+
+/**
+ * Fetches {@link KotlinDslScriptsModel} for the root build and all editable builds,
+ * mirroring what the IDE does during sync.
+ */
+public class FetchKotlinDslScriptsModelForAllBuilds implements BuildAction<Map<File, KotlinDslScriptModel>>, Serializable {
+
+    @Override
+    public Map<File, KotlinDslScriptModel> execute(BuildController controller) {
```
alllex: 💭 I would not flatmap all scripts into a single map. It's better to maintain clear separation of scripts per included build. We'd also then be validating per-included-build model in the assertions.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/FetchKotlinDslScriptsModelForAllBuilds.java
```diff
@@ -0,0 +1,52 @@
...
+        GradleBuild rootBuild = controller.getModel(GradleBuild.class);
+        Map<File, KotlinDslScriptModel> allModels = new LinkedHashMap<>();
+
+        collectScriptModels(controller, rootBuild, allModels);
+        for (GradleBuild build : rootBuild.getEditableBuilds()) {
```
alllex: 🤔 I think we should really use the `controller.run()` to additionall ensure parallel model building with IP doesn't affect the results.

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/GradleBuildIsolatedProjectsKotlinDslSmokeTest.groovy
```diff
@@ -0,0 +1,69 @@
...
+                .setStandardError(System.err)
+            if (extraArgs) {
+                actionBuilder.addArguments(extraArgs as List)
+            }
+            return actionBuilder.run()
+        }
+    }
+
+    static void checkBuildTreeScriptModels(Map<File, KotlinDslScriptModel> actual, Map<File, KotlinDslScriptModel> expected) {
+        assert actual.size() == expected.size()
```
alllex: 💅 The key set check below validates this as well, but will also provide a better error message, in case it fails

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt
```diff
@@ -67,6 +67,7 @@ data class KotlinBuildScriptModelParameter(
     val correlationId: String?
 )
 
+internal typealias MetadataDir = File
```
alllex: 💭 We could make `MetadataDir`  a (inlined) value class, so that extension functions defined on it provide additional type-safety (not letting to call them on a random `File`).

I think a short docstring on this type would also go a long way.

alllex: Thinking more about this, the extension functions defined on this type actually scatter around the expected structure of that directory. I think it's better to go the other direction, and consolidate these functions directly on this type. Then it's not even that important it's a value class.

Bringing the structure (names of subdirectories) into a single place will reduce obscurity, making it easier to understand the contracts around this directory. It will also help with the docstring, because if the contract changes, it's would also be clear the docstring must be kept up-to-date

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt
```diff
@@ -189,22 +203,14 @@ fun precompiledScriptPluginModelBuilder(
...
+            implicitAccessorsImports(scriptFile) + implicitPluginSpecBuildersImports
         }
     }
 )
 
 
-private
-val Project.precompiledScriptPluginsMetadataDir: File
+internal
+val Project.precompiledScriptPluginsMetadataDir: MetadataDir
```
alllex: 💭 It would be nice to co-locate the definition of that directory with the type. At this point, I think it's more important that the relative path of that directory (relative to project) is closer to the rest of the structure. My personal preference would be to have `MetadataDir.of(Project)`, because then the callsites are much easier to read, then guessing where the extension function is coming from.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
```diff
@@ -274,12 +283,14 @@ data class IntermediateScriptModel(
...
     val localImplicitImports: List<String> = emptyList(),
+    val includeParentSourcePath: Boolean = true,
 )
 
 
 internal
 data class IsolatedScriptsModel(
-    val models: List<IntermediateScriptModel>
+    val models: List<IntermediateScriptModel>,
+    val buildscriptSourcePath: ClassPath
```
alllex: 💅 `buildScript...` for consistency with the surrounding code

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
alllex: 💭 The `visitChildren` for me reads as we are going to look only at the children. If I were to rename this, I'd go with `visitSubprojects` or `visitRecursively` to highlight that aspect. But also, it's a local function, so it can be named more concisely than a top-level function. So I'd personally stick with `visit`.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
```diff
@@ -186,10 +194,11 @@ fun IntermediateToolingModelProvider.getIsolatedModels(requester: ProjectState,
 
 
 private
-fun buildOutputModel(base: ScriptModelBase, model: IntermediateScriptModel): StandardKotlinDslScriptModel {
+fun buildOutputModel(base: ScriptModelBase, model: IntermediateScriptModel, parentSourcePath: ClassPath): StandardKotlinDslScriptModel {
```
alllex: 💭 Understanding this function requires understanding the caller (and the other way around). It's also called only in a single place. Given this, I'd inline the definition

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/KotlinDslModelChecker.groovy
alllex: Do the tests still pass, if we uncomment this line?

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/KotlinDslModelChecker.groovy
```diff
@@ -0,0 +1,64 @@
...
+            { it.severity },
+            { it.message },
+            [{ it.position }, [
+                { it.line },
+                { it.column },
+            ]]
+        ])
+    }
+
+    // replaces `kotlin-dsl/accessors/*hash*/` with `kotlin-dsl/accessors/<hash>`
```
alllex, line 67: 🤔 It's hard to grasp (without knowing the implementation) that this doesn't accidentally conflate multiple files into one.

Do we know why the hashes are different without/with IP?

alllex (commented):
This looks good in general, I left a couple of questions.

I also suggest consolidating some of the logic around the structure on disk, instead of having it scattered around in extension functions. I believe this would improve readability, but this departs from the current style the code is written. I'll leave it to you (and to the @gradle/bt-tide) team to decide whether it's something we should do in the current PR.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/KotlinDslModelChecker.groovy
6hundreds: @alllex Yes

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/KotlinDslModelChecker.groovy
6hundreds: @alllex 
> Do we know why the hashes are different without/with IP?

Well, I assume hashes are different each time we build the model?

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt
6hundreds: @alllex Yes, certainly a good thing to do, I'll apply it to this PR

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/FetchKotlinDslScriptsModelForAllBuilds.java
6hundreds: @alllex Could you guide me a bit here, how it should looks like?

6hundreds: I see https://github.com/gradle/gradle/blob/fc2a0ec0e67f404adb62ed7c89147549247ee7eb/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/FetchCustomModelForSameProjectInParallel.java#L26 as a reference

alllex: Yep, that's what I meant

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/KotlinDslModelChecker.groovy
alllex: > Well, I assume hashes are different each time we build the model?

We can quickly check this by having the test request the vintage model twice. My gut feel is that these hashes should be content-based, not time-based. Otherwise, they aren't very cacheable.

I saw some of the content-hashing in related functions: https://github.com/gradle/gradle/blob/34a75f15027340be92d88923d19fe68f6e997ea4/platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt#L224

Though, I didn't look what else goes into those hashes.

mlopatkin:
architecture-council LGTM.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/FetchKotlinDslScriptsModelForAllBuilds.java
6hundreds: @alllex Something like `Map<BuildIdentifier, KotlinDslScriptsModel>`? 

I'm struggling with finding a key for such a map, nor `GradleBuild` or `BuildIdentifier` are really good for this

alllex: Keeping the actual `KotlinDslScriptsModel` as a value is the best 👍 

Since this is only for testing, we don't need to be overly type-safe about the keys. I think a build-tree path of the included build as a `String` should be enough.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/KotlinDslModelChecker.groovy
6hundreds: @alllex see https://github.com/gradle/gradle/issues/37719

PR #37622, opened 2026-04-21
Forbid BuildServicesRegistry.getRegistrations with IP

### LINKED ISSUE #32239: gradle.sharedServices.registrations.getByName() is not thread safe
### Current Behavior

`gradle.sharedServices.registrations.getByName()` is the common way for users to obtain already-registered `BuildService`s.

However, according to [this comment](https://github.com/gradle/gradle/issues/18587#issuecomment-1071138302) from @bamboo, `gradle.sharedServices.registrations.getByName()` is **not thread safe**.

In the Android Gradle plugin, we had to work around this issue by [avoiding using that API](https://android.googlesource.com/platform/tools/base/+/bf19b466c3c656769801e65c8238c2cdbdfbc9b3/build-system/gradle-core/src/main/java/com/android/build/gradle/internal/services/buildServices.kt#54).

Unfortunately, not all plugin authors are aware of this issue, and until it is fixed it will continue to cause flaky, hard-to-debug build failures (e.g., [KT-73842](https://youtrack.jetbrains.com/issue/KT-73842/)).

### Expected Behavior

Please consider:
  - Make `gradle.sharedServices.registrations.getByName()` thread safe
  - Or deprecate that API, and replace it with a thread-safe one.

### Context (optional)

_No response_

N/A

### Gradle version

N/A

### Build scan URL (optional)

_No response_

_No response_

jvandort:
This API return a NamedDomainObjectSet which is not thread safe. Further, even if it was thread safe, it would expose the parameters of each build service as shared mutable state. For this reason, we forbid this method unconditionally with IP

Fixes #32239

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache-base/src/main/kotlin/org/gradle/internal/cc/base/problems/IgnoringProblemsListener.kt
+++ b/platforms/core-configuration/configuration-cache-base/src/main/kotlin/org/gradle/internal/cc/base/problems/IgnoringProblemsListener.kt
@@ -25,7 +25,7 @@ import org.gradle.internal.configuration.problems.StructuredMessageBuilder
 /**
  * A stub implementation of ProblemsListener.
  */
-class IgnoringProblemsListener : ProblemsListener {
+object IgnoringProblemsListener : ProblemsListener {
     override fun onProblem(problem: PropertyProblem) = Unit
 
     override fun onError(trace: PropertyTrace, error: Exception, message: StructuredMessageBuilder) = Unit
--- a/platforms/core-configuration/configuration-cache-base/src/main/kotlin/org/gradle/internal/cc/base/problems/IgnoringProblemsListener.kt
+++ b/platforms/core-configuration/configuration-cache-base/src/main/kotlin/org/gradle/internal/cc/base/problems/IgnoringProblemsListener.kt
@@ -17,6 +17,7 @@
 package org.gradle.internal.cc.base.problems
 
 import org.gradle.api.Task
+import org.gradle.internal.configuration.problems.IsolatedProjectsProblemsListener
 import org.gradle.internal.configuration.problems.ProblemsListener
 import org.gradle.internal.configuration.problems.PropertyProblem
 import org.gradle.internal.configuration.problems.PropertyTrace
@@ -25,7 +26,7 @@ import org.gradle.internal.configuration.problems.StructuredMessageBuilder
 /**
  * A stub implementation of ProblemsListener.
  */
-object IgnoringProblemsListener : ProblemsListener {
+class IgnoringProblemsListener : ProblemsListener, IsolatedProjectsProblemsListener {
     override fun onProblem(problem: PropertyProblem) = Unit
 
     override fun onError(trace: PropertyTrace, error: Exception, message: StructuredMessageBuilder) = Unit
@@ -35,4 +36,6 @@ object IgnoringProblemsListener : ProblemsListener {
     override fun forTask(task: Task): ProblemsListener = this
 
     override fun onExecutionTimeProblem(problem: PropertyProblem) = Unit
+
+    override fun onIsolatedProjectsProblem(problem: PropertyProblem) = Unit
 }
--- a/platforms/core-configuration/configuration-cache-base/src/main/kotlin/org/gradle/internal/cc/base/problems/IgnoringProblemsListener.kt
+++ b/platforms/core-configuration/configuration-cache-base/src/main/kotlin/org/gradle/internal/cc/base/problems/IgnoringProblemsListener.kt
@@ -17,7 +17,6 @@
 package org.gradle.internal.cc.base.problems
 
 import org.gradle.api.Task
-import org.gradle.internal.configuration.problems.IsolatedProjectsProblemsListener
 import org.gradle.internal.configuration.problems.ProblemsListener
 import org.gradle.internal.configuration.problems.PropertyProblem
 import org.gradle.internal.configuration.problems.PropertyTrace
@@ -26,7 +25,7 @@ import org.gradle.internal.configuration.problems.StructuredMessageBuilder
 /**
  * A stub implementation of ProblemsListener.
  */
-class IgnoringProblemsListener : ProblemsListener, IsolatedProjectsProblemsListener {
+class IgnoringProblemsListener : ProblemsListener {
     override fun onProblem(problem: PropertyProblem) = Unit
 
     override fun onError(trace: PropertyTrace, error: Exception, message: StructuredMessageBuilder) = Unit
@@ -36,6 +35,4 @@ class IgnoringProblemsListener : ProblemsListener, IsolatedProjectsProblemsListe
     override fun forTask(task: Task): ProblemsListener = this
 
     override fun onExecutionTimeProblem(problem: PropertyProblem) = Unit
-
-    override fun onIsolatedProjectsProblem(problem: PropertyProblem) = Unit
 }
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/BuildTreeModelControllerServices.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/BuildTreeModelControllerServices.kt
@@ -118,7 +118,7 @@ object BuildTreeModelControllerServices : ServiceRegistrationProvider {
             // Allow nudging to enable CC if it is off and there is no explicit decision. CC doesn't work for model building so do not nudge there.
             !requirements.startParameter.configurationCache.isExplicit && !requirements.isCreatesModel -> add(ConfigurationCachePromoHandler::class.java)
             // Do not nudge if CC is explicitly disabled or if models are requested.
-            else -> add(ProblemsListener::class.java, IgnoringProblemsListener::class.java)
+            else -> add(ProblemsListener::class.java, IgnoringProblemsListener)
         }
 
         add(InstrumentedExecutionAccessListenerRegistry::class.java)
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/BuildTreeModelControllerServices.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/BuildTreeModelControllerServices.kt
@@ -59,7 +59,6 @@ import org.gradle.internal.concurrent.ExecutorFactory
 import org.gradle.internal.configuration.problems.CommonReport
 import org.gradle.internal.configuration.problems.DefaultIsolatedProjectsProblemsReporter
 import org.gradle.internal.configuration.problems.DefaultProblemFactory
-import org.gradle.internal.configuration.problems.IsolatedProjectsProblemsListener
 import org.gradle.internal.configuration.problems.IsolatedProjectsProblemsReporter
 import org.gradle.internal.configuration.problems.NoOpIsolatedProjectsProblemsReporter
 import org.gradle.internal.configuration.problems.ProblemFactory
@@ -119,7 +118,7 @@ object BuildTreeModelControllerServices : ServiceRegistrationProvider {
             // Allow nudging to enable CC if it is off and there is no explicit decision. CC doesn't work for model building so do not nudge there.
             !requirements.startParameter.configurationCache.isExplicit && !requirements.isCreatesModel -> add(ConfigurationCachePromoHandler::class.java)
             // Do not nudge if CC is explicitly disabled or if models are requested.
-            else -> add(ProblemsListener::class.java, IsolatedProjectsProblemsListener::class.java, IgnoringProblemsListener::class.java)
+            else -> add(ProblemsListener::class.java, IgnoringProblemsListener::class.java)
         }
 
         add(InstrumentedExecutionAccessListenerRegistry::class.java)
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/BuildTreeModelControllerServices.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/BuildTreeModelControllerServices.kt
@@ -59,6 +59,7 @@ import org.gradle.internal.concurrent.ExecutorFactory
 import org.gradle.internal.configuration.problems.CommonReport
 import org.gradle.internal.configuration.problems.DefaultIsolatedProjectsProblemsReporter
 import org.gradle.internal.configuration.problems.DefaultProblemFactory
+import org.gradle.internal.configuration.problems.IsolatedProjectsProblemsListener
 import org.gradle.internal.configuration.problems.IsolatedProjectsProblemsReporter
 import org.gradle.internal.configuration.problems.NoOpIsolatedProjectsProblemsReporter
 import org.gradle.internal.configuration.problems.ProblemFactory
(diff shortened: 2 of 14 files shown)


platforms/core-configuration/configuration-problems-base/src/main/kotlin/org/gradle/internal/cc/impl/problems/BuildNameProvider.kt
```diff
@@ -1,58 +0,0 @@
...
-    // The issue is that settingsEvaluated relies on the BuildListener which is not supported in CC. So we can not throw in case it is null.
-    // the tests that break if you do are in ConfigurationCacheDependencyResolutionIntegrationTest
-    // a nicer way to do this would be to use buildStateRegistry.rootBuild.projects.rootProject.name but that also doesn't work with CC :(
-    // one of the tests that fail if you try this is org.gradle.integtests.resolve.catalog.TomlDependenciesExtensionIntegrationTest.should throw an error if 'from' is called multiple times
-    fun buildName(): String? {
-        return rootBuildName
-    }
-
-    override fun settingsEvaluated(settings: Settings) {
-        if ((settings as SettingsInternal).gradle.isRootBuild) {
```
jvandort, line 54: The dependency on `SettingsInternal` required us to depend on `:core`. Replacing this class with `BuildStateRegistry` allows us to remove the `:core` dependency and allow `:core` to depend on this project

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -53,8 +53,13 @@ While the feature is experimental, notable changes are documented here rather th
 ** Tooling model caching is now disabled, as it currently has known correctness problems.
 
 * Gradle 9.2.0
-** Additional APIs on `TaskContainer` are treated as incompatible: `findByPath(String)` and `getByPath(String)`
-methods now emit violations when a task from another project is requested.
+** Additional APIs on `TaskContainer` are treated as incompatible:
+`findByPath(String)` and `getByPath(String)` now emit violations when a task from another project is requested.
+
+* Gradle 9.6.0
```
alllex: ❌ The timeline is recent-first

subprojects/core/src/main/java/org/gradle/api/services/internal/DefaultBuildServicesRegistry.java
```diff
@@ -390,4 +429,68 @@ public void buildFinished(BuildResult result) {
...
+
+        @Inject
+        public IsolatedProjectsReportingRegistrationsContainer(
+            NamedDomainObjectSet<BuildServiceRegistration<?, ?>> delegate,
+            IsolatedProjectsProblemsReporter problems,
+            BuildModelParameters buildModelParameters,
+            FunctionRunner synchronizedRunner
+        ) {
+            //noinspection DataFlowIssue
+            super(null);
```
alllex: 🤔 This feels a bit shaky. Can we instead make `DelegatingDomainObjectSet` and `DelegatingNamedDomainObjectSet` parameteric on the delegate, so that child classes don't actually need to override `getDelegate()` just to do a cast?

This should make it so it's enough to pass-in the `delegate` to the `super()` call to then be able to call `getDelegate()` receiving a subtype directly.

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DelegatingNamedDomainObjectSet.java
```diff
@@ -51,118 +52,142 @@ protected NamedDomainObjectSet<T> getDelegate() {
 
     @Override
     public <S extends T> NamedDomainObjectSet<S> withType(Class<S> type) {
+        onMethodCall("withType(Class)");
```
alllex, line 55: ❔ Is it a concern that all delegating sets are now calling these methods to do nothing?

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/CompositeDomainObjectSet.java
```diff
@@ -101,11 +103,21 @@ protected DomainObjectCompositeCollection<T> getStore() {
         return (DomainObjectCompositeCollection) getDelegate().getStore();
     }
 
+    @Override
+    public void whenObjectAdded(Closure action) {
+        whenObjectAdded(ConfigureUtil.configureUsing(action));
```
alllex, line 108: ❓ Are we skipping the delegate here? Why? Why weren't we doing it previously?

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DelegatingNamedDomainObjectSet.java
jvandort: If perf tests pass -- I don't think so
I would expect the JVM would see the empty method body and optimize the call out, but if it doesn't and it affects performance measurably we should see that

platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/CompositeDomainObjectSet.java
jvandort: This is `CompositeDomainObjectSet`, it is not a delegate implementation.
This implementation overrides the normal `whenObjectAdded` so if we didn't implement that we would skip the overridden logic.
Previously, the `Delegating` implementations skipped the delegates and called `whenObjectAdded`. This changed the `Delegating` implementations to be more "pure", properly going through the delegate.

subprojects/core/src/main/java/org/gradle/api/services/internal/DefaultBuildServicesRegistry.java
jvandort: This one here is a "special" case -- The delegating sets _do_ accept a delegate. This implementation needs a mutable delegate due to the existing logic in this class.

jvandort: We can instantiate a new instance of the container instead

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
lkasso: +1

lkasso (approved):
Docs LGTM. Just have to be moved.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/BuildTreeModelControllerServices.kt
```diff
@@ -109,7 +111,7 @@ object BuildTreeModelControllerServices : ServiceRegistrationProvider {
             // Allow nudging to enable CC if it is off and there is no explicit decision. CC doesn't work for model building so do not nudge there.
             !requirements.startParameter.configurationCache.isExplicit && !requirements.isCreatesModel -> add(ConfigurationCachePromoHandler::class.java)
             // Do not nudge if CC is explicitly disabled or if models are requested.
-            else -> add(ProblemsListener::class.java, IgnoringProblemsListener)
+            else -> add(ProblemsListener::class.java, IsolatedProjectsProblemsListener::class.java, IgnoringProblemsListener::class.java)
```
alllex: ❌ Instead of making the _listener_ a cross-mode service, we should instead provide different implementations of the _reporter_. 

We already have this in Reinhold's PR: https://github.com/gradle/gradle/pull/37782

6hundreds:
@bot-gradle test this

6hundreds:
@bot-gradle test this

subprojects/core/src/integTest/groovy/org/gradle/api/services/BuildServiceIntegrationTest.groovy
```diff
@@ -1157,6 +1158,7 @@ Hello, subproject1
         outputContains("service: closed with value 13")
     }
 
+    @ToBeFixedForIsolatedProjects(because = "Cannot modify registration parameters with IP enabled")
```
octylFractal: ❓ Is this really a "to be _fixed_"? I would assume we would just not ever support this?

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -519,9 +519,6 @@ processResources {
 ----
 |===
 
-[[changes_9.6.0]]
-== Upgrading from 9.5.0 and earlier
-
```
octylFractal, line 548: :x: This change, if needed, should be done separately IMO.

lkasso: OK with that. I'll clean up in a separate PR if this doesn't make it.

lkasso (approved):

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/promo/ConfigurationCachePromoHandler.kt
```diff
@@ -138,4 +140,7 @@ internal class ConfigurationCachePromoHandler(
     private fun TaskExecutionGraph.hasIncompatibleTasks() = allTasks.any { !(it as TaskInternal).isCompatibleWithConfigurationCache }
 
     private fun runWithoutBuildDefinition() = rootBuildLayout.isBuildDefinitionMissing
+
+    override fun onIsolatedProjectsProblem(problem: PropertyProblem) = Unit
```
alllex: ❓ Why is this change necessary?

platforms/core-configuration/configuration-cache-base/src/main/kotlin/org/gradle/internal/cc/base/problems/IgnoringProblemsListener.kt
```diff
@@ -25,7 +25,7 @@ import org.gradle.internal.configuration.problems.StructuredMessageBuilder
 /**
  * A stub implementation of ProblemsListener.
  */
-object IgnoringProblemsListener : ProblemsListener {
+class IgnoringProblemsListener : ProblemsListener {
```
alllex: ❔ Why is this change necessary?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/BuildTreeModelControllerServices.kt
```diff
@@ -119,7 +119,7 @@ object BuildTreeModelControllerServices : ServiceRegistrationProvider {
             // Allow nudging to enable CC if it is off and there is no explicit decision. CC doesn't work for model building so do not nudge there.
             !requirements.startParameter.configurationCache.isExplicit && !requirements.isCreatesModel -> add(ConfigurationCachePromoHandler::class.java)
             // Do not nudge if CC is explicitly disabled or if models are requested.
-            else -> add(ProblemsListener::class.java, IgnoringProblemsListener)
+            else -> add(ProblemsListener::class.java, IgnoringProblemsListener::class.java)
```
alllex: ❔ Why is this change necessary?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -368,7 +373,9 @@ Only the first registration succeeds, and any later registrations return the pre
 However, the parameters of the build service can become shared mutable state.
 Any mutation of the build service parameters outside of `registerIfAbsent(...) { parameters { ... } }` should be avoided.
 
-Accessing the set of build service registrations at `gradle.sharedServices.registrations` is unsafe.
+Accessing the set of build service registrations at `gradle.sharedServices.registrations` is restricted with Isolated Projects.
+Only `findByName(String)` is permitted on the returned set; all other methods emit IP violations.
```
alllex: ❌ We should not be using "IP" as abbreviation in the prose of the user guide. It's not the only place, where it's been added in this PR

alllex (changes requested):

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/BuildTreeModelControllerServices.kt
6hundreds: Because of IgnoringProblemsListener became a class, above

platforms/core-configuration/configuration-cache-base/src/main/kotlin/org/gradle/internal/cc/base/problems/IgnoringProblemsListener.kt
6hundreds: Not sure, I'd revert it back

subprojects/core/src/integTest/groovy/org/gradle/api/services/BuildServiceIntegrationTest.groovy
6hundreds: That would be my understanding as well. There is no `UnsupportedWithIP` annotation so far tho. I can introduce it in follow up PR

octylFractal: That works for me.

octylFractal: Actually, what about `@Requires` with `NotIsolatedProjects` + reason?

6hundreds: Sounds good!

6hundreds:
@alllex @octylFractal Thanks for review, pushed the changes! PTAL

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
6hundreds: Fixed

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/BuildTreeModelControllerServices.kt
6hundreds: Reverted

platforms/core-configuration/configuration-cache-base/src/main/kotlin/org/gradle/internal/cc/base/problems/IgnoringProblemsListener.kt
6hundreds: Reverted

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/promo/ConfigurationCachePromoHandler.kt
6hundreds: Reverted

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/promo/ConfigurationCachePromoHandler.kt
```diff
@@ -55,6 +55,7 @@ internal class ConfigurationCachePromoHandler(
     private val degradationController: DefaultConfigurationCacheDegradationController,
     private val documentationRegistry: DocumentationRegistry
 ) : RootBuildLifecycleListener, ProblemsListener {
+
```
octylFractal: 💅 Let's remove this file from the diff entirely.
```suggestion
```

octylFractal (approved):

alllex (approved):

PR #37632, opened 2026-04-22
Implicitly finalize List, Set and Map Properties with CC

### LINKED ISSUE #36664: ListProperty and MapPropery should be implicitly finalized when the configuration cache is used
Related:
* https://github.com/gradle/gradle/issues/32591
* https://github.com/gradle/gradle/issues/25516

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/CollectionPropertyIntegrationTest.groovy
+++ b/platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/CollectionPropertyIntegrationTest.groovy
@@ -17,6 +17,7 @@
 package org.gradle.api.provider
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
+import org.gradle.integtests.fixtures.executer.GradleContextualExecuter
 
 class CollectionPropertyIntegrationTest extends AbstractIntegrationSpec {
     def setup() {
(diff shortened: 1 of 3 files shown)


ov7a:
@bot-gradle test APT

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/CollectionPropertyIntegrationTest.groovy
mlopatkin, line 1: ❓ Could you please add a test that explicitly verifies property identity being shared inside the task and not shared across task?

Something like:
```kotlin
tasks {
    val sharedListProp = objects.listProperty<String>()
    register("foo") {
        val listProp = objects.listProperty<String>()
        doFirst {
            listProp.add("A")
            sharedListProp.add("A")
        }
        doLast {
            listProp.add("B")
            sharedListProp.add("B")
            println("listProp = ${listProp.get()}")  // [A, B]
            println("FOO: sharedListProp = ${sharedListProp.get()}")  // [A, B]
        }
    }
    register("bar") {
        dependsOn("foo")
        doFirst {
            sharedListProp.add("C")
            println("BAR: sharedListProp = ${sharedListProp.get()}")  // [C]
        }
    }
}

```

It has to be in Kotlin, BTW, because of Groovy's wrapping `Reference` instance shenanigans.

Same for `MapProperty`

mlopatkin (dismissed):
LGTM, but it would be great to have a test that verifies identities without relying on side effects of finalization.

While we may deprecate late modifications later, it is important to maintain the current behavior until then.

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/MapPropertyIntegrationTest.groovy
```diff
@@ -831,4 +812,46 @@ task thing {
         "keySet provider" | 'keySet().map { it.sort().join(",")}' | true        | [":foo"]                 | "Entry is 1,42"
         "keySet provider" | 'keySet().map { it.sort().join(",")}' | false       | [":bar", ":baz", ":foo"] | "Entry is 1,42"
     }
+
+    def "map property has separate identity per task"() {
```
mlopatkin: 💅 Name is a bit confusing, given that the test runs with and without CC. 
```suggestion
    def "map property has separate identity per task with CC only"() {
```

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/CollectionPropertyIntegrationTest.groovy
```diff
@@ -547,6 +546,86 @@ task wrongPropertyElementTypeApi(type: MyTask) {
         "set" | "setProperty"
     }
 
+    def "list property has separate identity per task"() {
```
mlopatkin: 💅 Name is a bit confusing, given that the test runs with and without CC. 
```suggestion
    def "list property has separate identity per task with CC only"() {
```

platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/CollectionPropertyIntegrationTest.groovy
```diff
@@ -547,6 +546,86 @@ task wrongPropertyElementTypeApi(type: MyTask) {
...
+        outputContains("listProp = [A, B]")
+        outputContains("FOO: sharedListProp = [A, B]")
+        if (GradleContextualExecuter.configCache) {
+            outputContains("BAR: sharedListProp = [C]")
+        } else {
+            outputContains("BAR: sharedListProp = [A, B, C]")
+        }
+    }
+
+    def "set property has separate identity per task"() {
```
mlopatkin: 💅 Name is a bit confusing, given that the test runs with and without CC. 
```suggestion
    def "set property has separate identity per task with CC only"() {
```

mlopatkin: 💅 Given how similar is this and the list test, perhaps it would be beneficial to fold them into a parameterized test with the actual property type in `where:`.

mlopatkin (commented):
LGTM, with some non-blocking suggestions for tests.

FWIW, a similar test is named `retains Property identity for each task`.

PR #37639, opened 2026-04-22
Render problem solutions inline instead of in resolution section

### LINKED ISSUE #36380: Problem resolutions are dissociated from their description in the build failure message
### Current Behavior

The Problems API allows clients to fail the build by throwing an exception with the `ProblemReporter.throwing(..)` methods. 

The build failure message has a specific rendering: the first part renders the description, then separately, there's a dedicated resolution section (ie the `Try:` section) containing the proposed solutions. When the build failure has associated Problems the solutions property from the Problems are also rendered in the resolution section. 

The problem arises when there's more than one Problem associated with the build failure. The solutions are dissociated from the descriptions. See the example build failure message:
```

* Where:
Build file '/Users/donat/Development/git/gradle/problems-playground/build.gradle.kts' line: 346

* What went wrong:
java.lang.Exception: Boom failure
> Problem 1
    Context for Problem 1
      Details about Problem 1
  Problem 2
    Context for Problem 2
      Details about Problem 2

* Try:
> Solution A
> Solution B
> Run with --scan to generate a Build Scan (powered by Develocity).

BUILD FAILED in 401ms
```

Which solution belongs to Problem 1 and to Problem 2?

### Expected Behavior

There are multiple ways we can fix this problem. One option is simply to get rid of the separation of problems and solutions and render them as a single unit. That is reasonable, however we have to revisit the existing failure template and figure out what to do with the generic advice under the Try section. 

Alternatively, the failure message could i) number the rendered problems and use those as references in the Try  section, ii) inline the solutions into the problem rendering, or something else. The numbering-based approach seems more streamlined: it would not result in a different format in different scenarios (0, 1, or more problem occurrences associated with a build failure). Also numbers can be omitted when there’s exactly one associated problem occurrence, so most build failure messages would be kept unchanged.

donat:
Fixes #36380                                                                                             

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/build-configuration/src/main/java/org/gradle/internal/buildconfiguration/DaemonJvmPropertiesConfigurator.java
+++ b/platforms/core-runtime/build-configuration/src/main/java/org/gradle/internal/buildconfiguration/DaemonJvmPropertiesConfigurator.java
@@ -86,11 +86,10 @@ public class DaemonJvmPropertiesConfigurator implements ProjectConfigureAction {
                             JavaToolchainResolverService resolverService = project.getServices().get(JavaToolchainResolverService.class);
                             if (!resolverService.hasConfiguredToolchainRepositories()) {
                                 UnconfiguredToolchainRepositoriesResolver exception = new UnconfiguredToolchainRepositoriesResolver();
-                                throw reporter.throwing(exception, UpdateDaemonJvm.TASK_CONFIGURATION_PROBLEM_ID,
-                                    problemSpec -> {
-                                        problemSpec.solution("Configure toolchain download repositories in your build settings.");
-                                        problemSpec.documentedAt(Documentation.userManual("toolchains", "sub:download_repositories").getUrl());
-                                });
+                                throw reporter.throwing(exception, reporter.create(UpdateDaemonJvm.TASK_CONFIGURATION_PROBLEM_ID, problemSpec -> {
+                                    problemSpec.solution("Configure toolchain download repositories in your build settings.");
+                                    problemSpec.documentedAt(Documentation.userManual("toolchains", "sub:download_repositories").getUrl());
+                                }));
                             }
                             Map<BuildPlatform, Optional<URI>> buildPlatformOptionalUriMap = platforms.stream()
                                 .collect(Collectors.toMap(platform -> platform,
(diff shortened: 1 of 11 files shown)


donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test this

platforms/ide/problems-rendering/src/main/java/org/gradle/problems/internal/rendering/RenderOptions.java
```diff
@@ -23,16 +23,13 @@ class RenderOptions {
 
     private final String prefix;
     private final boolean renderId;
-    private final boolean renderSolutions;
```
donat, line 26: Removed as solutions are now rendered unconditionally.

donat:
@bot-gradle test this

platforms/core-runtime/build-configuration/src/main/java/org/gradle/internal/buildconfiguration/DaemonJvmPropertiesConfigurator.java
```diff
@@ -54,7 +54,7 @@ public class DaemonJvmPropertiesConfigurator implements ProjectConfigureAction {
     public void execute(ProjectInternal project) {
         // Only useful for the root project
         if (project.getParent() == null) {
-            ProblemReporter reporter = project.getServices().get(Problems.class).getReporter();
+            ProblemReporterInternal reporter = project.getServices().get(ProblemsInternal.class).getInternalReporter();
```
eskatos: ❓ Why is the internal API required here?

platforms/ide/problems-api/src/integTest/groovy/org/gradle/api/problems/ProblemsServiceIntegrationTest.groovy
```diff
@@ -477,9 +477,12 @@ class ProblemsServiceIntegrationTest extends AbstractIntegrationSpec {
...
     Complex build logic like the Problems API usage should be integrated into plugins
-    Solution: Look up the samples index for real-life examples
-    Documentation: https://example.com/some-problem
-    Location: /path/to/script
+
+    Possible solution: Look up the samples index for real-life examples.
+
+    For more information, please refer to https://example.com/some-problem.
+
+    Location: /path/to/script line 20
```
eskatos, line 481: 💅  Why the blank lines?
💅  Why change how the documentation is displayed?

💅 What follows is not strictly about the current changes but triggered by them.

The documentation is about the problem, not the specific solution.
In the HTML report we put the documentation link on the problem label.
Also, in the HTML report we put the solution(s) last, below the location(s), and I think it makes sense.

Maybe it would be better if it looked like this:
```
Problem found: Project is a prototype (id: sample-problems:prototype-project)
  This is a prototype and not a guideline for modeling real-life projects
    Complex build logic like the Problems API usage should be integrated into plugins
    For more information, please refer to https://example.com/some-problem.
    Location: /path/to/script
    Possible solution: Look up the samples index for real-life examples
```

```
Problem found: Project is a prototype (id: sample-problems:prototype-project)
  This is a prototype and not a guideline for modeling real-life projects
    Complex build logic like the Problems API usage should be integrated into plugins
    For more information, please refer to https://example.com/some-problem.
    Location: /path/to/script
    Location: /path/to/another_script
    Possible solutions:
      1. Look up the samples index for real-life examples
      2. Do something else
```

WDYT?

donat: I chose the formatting ad-hoc, basically to loosely match what we already have for version catalog/type validation. I actually like your proposal better (wo/the extra newlines). 

Re: documentation links. They have the same issue as the solutions: they are hoisted to the `Try: ` section and so it makes sense to inline them as well.

platforms/core-runtime/build-configuration/src/main/java/org/gradle/internal/buildconfiguration/DaemonJvmPropertiesConfigurator.java
donat: To be able to configure a `DocLink` reference to to Problem (see the removal of `getUrl()`) call.

donat:
@bot-gradle test this

donat:
@eskatos your proposed update on the rendering is implemented. PTAL.

platforms/core-runtime/build-configuration/src/main/java/org/gradle/internal/buildconfiguration/DaemonJvmPropertiesConfigurator.java
eskatos: So, we can avoid the internals by just adding `.getUrl()`, correct?
If yes, I would avoid the internals. It's not worth it.

eskatos (approved):
Mostly LGTM!

There is just this topic about using internals just for avoiding the creation of URLs upfront.
If this is a real issue (e.g.  performance), then it looks like a signal that we need something in the API that takes a `Supplier<String>` for the documentation URL. But if there is no issue and that's just convenience, I think we should just remove the convenience from the internals. The more we use the public API internally the better.

donat:
@bot-gradle test this

donat:
@bot-gradle test this

donat:
@bot-gradle test this

PR #37640, opened 2026-04-22
Remove Project.getProperties() and script getProperties()

### LINKED ISSUE #37385: Remove `getProperties()` API for Project and scripts
The current implementation of the `Project.getProperties()` method is flawed because it leaks Gradle internals, including internal accessors and Project state that is not part of the public API.

Apart from accessing and exposing internals, the API also eagerly realizes various sources of data, such as Gradle Properties, making them a build configuration input, if `getProperties()` is called at configuration time. This reduces the effectiveness of Configuration Cache by making it sensitive to irrelevant inputs, hurting CC hit rates.

Last but not least, `getProperties()` collects state (including mutable state) across the project hierarchy. This makes it fundamentally incompatible with the vision of Isolated Projects.

Given the combination of this problems, we are prosing to deprecate and later remove this method from Gradle entirely. This will allow users making less mistakes that can cost them performance, and it will also make it simpler for them to migrate to Isolated Projects in the future.

What needs to be done:
- [x] https://github.com/gradle/gradle/issues/35132
- [ ] Address deprecations, if any, in the gradle/gradle build
- [ ] Prepare the removal PR, targeting 10.0.0
- [ ] Make calling this method a direct IP error

<details><summary>Here is an example of the state of the returned map in a very minimal project</summary>
<p>

Image

Image

</p>
</details> 

---

Related issues:
- https://github.com/gradle/gradle/issues/35797

reinsch82:
Fixes #37385

## Summary

Remove `Project.getProperties()` and `BasicScript.getProperties()`, deprecated in 9.6.0 by #37291. Both returned a merged map that leaked internal state, eagerly materialized Gradle properties (hurting Configuration Cache hit rates), and walked the parent-project chain in a way that was fundamentally incompatible with Isolated Projects.

## Changes

1. **Remove `Project.getProperties()`** from the public API. Also remove the overrides in `DefaultProject`, `MutableStateAccessAwareProject`, and `ProjectDelegate`. Rewrite the one internal caller — `PropertyReportTask` — to read through the dynamic-object chain directly, so `./gradlew properties` continues to produce the same output. Add the `METHOD_REMOVED` entry to `accepted-public-api-changes.json` and refreeze the ArchUnit `public-api-mutable-properties` baseline.

2. **Remove `BasicScript.getProperties()`** and the matching `ScriptDynamicObject.getProperties()` override.

3. **Throw a helpful error for dynamic `properties` access.** After removing both methods, Groovy DSL `properties` / `project.properties` would otherwise (a) return Groovy's inherited `Script.getProperties()` (binding variables — silent behaviour change) or (b) throw an opaque `MissingPropertyException`. Preempt the lookup at `DefaultProject.property()` and `BasicScript.getProperty()`: when the name is `properties`, throw `InvalidUserCodeException` naming the replacement APIs (`providers.gradleProperty`, `findProperty`, `ext`) with a link to the new 10.0 upgrade guide.

4. **Tests.** Delete `GetPropertiesDeprecationIntegrationTest`. Add `GetPropertiesRemovalIntegrationTest` covering:
   - Groovy build/settings/init script `properties` → `InvalidUserCodeException` with the migration message.
   - Groovy `project.properties` → same.
   - Kotlin `build.gradle.kts` `properties` → script compilation failure.

   Drop the deprecation-warning expectations and the now-unreachable `properties` row from `IsolatedProjectsAccessFromGroovyDslIntegrationTest` (the new upfront error fires before the cross-project wrapper reports the IP problem — so the `properties` case is covered by the new removal test, not by the IP data table). Update `DynamicObjectIntegrationTest`, `DefaultProjectTest`, `GradleKotlinDslIntegrationTest`, `BomSupportPluginsSmokeTest`, `PaparazziPluginSmokeTest` similarly.

5. **Docs.** New `upgrading_major_version_10.adoc` with a `[[removed_get_properties]]` section including the full migration table inlined (users jumping from 8.x to 10.x should not have to chase the 9.x guide), a plugin-author note, and a note that `./gradlew properties` continues to work. Register the new file in `userguide_single.adoc`. Repoint ten `@see Project#getProperties()` Javadoc references in `TaskContainer` to `findProperty(String)`.

## Out of scope (potential follow-up)

The issue's separate bullet *\"Make calling this method a direct IP error\"* is largely absorbed by commit 3 — the new error fires uniformly, IP or not. A dedicated IP-error path for cross-project dynamic access via Groovy `otherProject.properties` remains possible as a follow-up if there's value in reporting it specifically as an IP violation rather than as a removed-API access.

## Test plan

- [x] `./gradlew compileAll`
- [x] `./gradlew sanityCheck` (excluding two pre-existing `ANDROID_SDK_ROOT`-dependent performance tasks)
- [x] `./gradlew :architecture-test:checkBinaryCompatibility`
- [x] `./gradlew :core:forkingIntegTest --tests \"*GetPropertiesRemovalIntegrationTest*\"`
- [x] `./gradlew :core:configCacheIntegTest --tests \"*GetPropertiesRemovalIntegrationTest*\"`
- [x] `./gradlew :core:forkingIntegTest --tests \"*DynamicObjectIntegrationTest*\"`
- [x] `./gradlew :core:test --tests \"*DefaultProjectTest*\"`
- [x] `./gradlew :configuration-cache:forkingIntegTest --tests \"*IsolatedProjectsAccessFromGroovyDslIntegrationTest*\"`
- [x] `./gradlew :configuration-cache:configCacheIntegTest --tests \"*IsolatedProjectsAccessFromGroovyDslIntegrationTest*\"`
- [x] `./gradlew :base-diagnostics:forkingIntegTest`
- [x] `./gradlew :kotlin-dsl-integ-tests:forkingIntegTest --tests \"*GradleKotlinDslIntegrationTest.can use kotlin java8 inline-only methods*\"`
- [ ] Paparazzi and BOM-support smoke tests — not run locally; the `maybeExpectLegacyDeprecationWarning` lines have been removed but the plugins themselves may still call `project.getProperties()` on Gradle 10. If CI catches an actual failure, we'll pin a newer plugin version or `@Ignore` with a pointer here.
- [ ] CI


reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test RFN

PR #37646, opened 2026-04-23
Suppress OSC taskbar reset on plain/piped stdout

### LINKED ISSUE #37611: Extra OSC characters leaking to output in iTerm
### Current Behavior

Running a task that prints a single line of plain text produces the expected text, a newline, and then the extra escape sequence:

```bash
./gradlew :printRawVersion -q | od -c
0000000    v   0   .   1   .   0   -   6   0   f   a   8   d   8   8   3
0000020    f  \n 033   ]   9   ;   4   ;   0  \a
```

The same extra bytes are present with `--console=plain`:

```bash
./gradlew --console=plain :printRawVersion -q | od -c
0000000    v   0   .   1   .   0   -   6   0   f   a   8   d   8   8   3
0000020    f  \n 033   ]   9   ;   4   ;   0  \a
0000032
```

The trailing sequence is:

```text
\u001b]9;4;0\u0007
```

### Expected Behavior

When Gradle output is piped, and especially when `--console=plain` is specified, stdout should contain only the task's plain output:

```text
v0.1.0-60fa8d883f\n
```

No terminal control sequence should be appended.

### Context (optional)

I want to be able to run a task and not have extra characters at the end that need to be stripped off.
This problem didn't exist in the previous version of iTerm that I was using (3.6.10).  I believe the main change is the top progress bar.

Any task that writes a plain string should be enough. For example:

```groovy
tasks.register("printRawVersion") {
    doLast {
        println("v0.1.0-60fa8d883f")
    }
}
```

Then run:

```bash
./gradlew --console=plain :printRawVersion -q | od -c
```

### Gradle version

9.4.1

### Build scan URL (optional)

_No response_

- Gradle wrapper distribution URL: `gradle-9.4.1-bin.zip`
- OS: macOS
- Terminal: iTerm2 3.6.10
- Shell: `bash`

eskatos:
The shutdown hook that emits ESC]9;4;0 BEL to clear iTerm/ConEmu/ Ghostty/Kitty taskbar progress was registered whenever the host terminal was detected via env vars, ignoring the active console mode and whether stdout was actually a terminal. That leaked the trailing control sequence into piped output and into --console=plain runs.

Gate the hook on the configuration that actually renders progress: skip it for ConsoleOutput.Plain, and for ConsoleOutput.Auto when stdout is not a TTY. Rich, Verbose, Colored, and Auto-with-TTY are unchanged, preserving the previous behavior.

* Fixes https://github.com/gradle/gradle/issues/37611

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/logging/src/integTest/groovy/org/gradle/internal/logging/console/TaskbarProgressResetFunctionalTest.groovy
+++ b/platforms/core-runtime/logging/src/integTest/groovy/org/gradle/internal/logging/console/TaskbarProgressResetFunctionalTest.groovy
@@ -98,4 +98,40 @@ class TaskbarProgressResetFunctionalTest extends AbstractIntegrationSpec {
         then:
         result.output.contains(OSC_RESET)
     }
+
+    @Issue("https://github.com/gradle/gradle/issues/37611")
+    @SuppressWarnings("IntegrationTestFixtures")
+    @Requires(value = IntegTestPreconditions.NotEmbeddedExecutor,
+        reason = "OSC taskbar progress sequences are only emitted by the forked client JVM")
+    def "does not emit OSC 9;4 sequences when --console=plain"() {
+        given:
+        executer.withConsole(ConsoleOutput.Plain)
+        buildFile << """
+            task ok { }
+        """
+
+        when:
+        result = succeeds("ok")
+
+        then:
+        !result.output.contains(OSC_PROGRESS_PREFIX)
+    }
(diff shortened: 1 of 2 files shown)


eskatos:
@bot-gradle help

jbartok (approved):

PR #37647, opened 2026-04-23
Avoid incidental IP problem reporting

### LINKED ISSUE #37643: Duplicate IP problems reported on `rootProject` and `parent` accessors
When a mutable state of another project is accessed from the **sub-sub** project, then the IP violation is reported twice, instead of once.

Reproducer:

```
        settingsFile """
            include(":lib:sub-lib")
        """
        buildFile """
            ext.myProp = "bar"
        """

        buildFile "lib/sub-lib/build.gradle", """
            rootProject.findProperty("myProp")
        """
```

Output with `rootProject`:
```
- Build file 'lib/sub-lib/build.gradle': line 2: Project ':lib' cannot access 'Project.findProperty' functionality on another project ':'
- Build file 'lib/sub-lib/build.gradle': line 2: Project ':lib:sub-lib' cannot access 'Project.findProperty' functionality on another project ':'
```

Output with `parent`: (notice how the target project of the violation is different on the second line)
```
- Build file 'lib/sub-lib/build.gradle': line 2: Project ':lib' cannot dynamically look up a property in the parent project ':'
- Build file 'lib/sub-lib/build.gradle': line 2: Project ':lib:sub-lib' cannot access 'Project.findProperty' functionality on another project ':lib'
```

The behavior is reproducible with both Groovy and Kotlin DSL

---

The root cause seems to be the way we look-up a root project instance that is attached to every subproject. Because we do it by calling the public `getRootProject()` method, we get a wrapped (problem-reporting) instance.

https://github.com/gradle/gradle/blob/f208a7cf2d154fe86f9d3300d8a90f846dda0248/subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java#L229

alllex:
- Fixes: https://github.com/gradle/gradle/issues/37643

We introduce `IsolatedProjectsProblemsReporter` that from now on should be used to report IP problems. Compared to the existing `IsolatedProjectsProblemsListener`, the reporter also supports ignoring incidental problems. Incidental problems are caused by previously reported problems, but create noise for the user in identifying the root cause.

The listener remains temporarily as glue between the reporter and the CC reporting infrastructure. We'll reconsider the need for it as we decouple IP from CC problem reporting.

The reporter implements a much simpler problem-ignoring mechanism compared to what was provided by some of the infrastructure removed in this PR, such as `DynamicLookupRoutine` and `DynamicCallProblemReporting`. We might need to have those in the future, but for now they introduce significant complexity with very low benefit.


alllex:

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsAccessFromGroovyDslIntegrationTest.groovy
```diff
@@ -633,10 +638,9 @@ class IsolatedProjectsAccessFromGroovyDslIntegrationTest extends AbstractIsolate
         then:
         outputContains("My property: hello")
 
-        // an additional subproject demonstrates that the problems are duplicated as the property lookup traverses up the project hierarchy
+        // an additional subproject demonstrates that the problems are not duplicated as the property lookup traverses up the project hierarchy
         fixture.assertStateStoredAndDiscarded {
             projectsConfigured(":", ":a", ":a:aa")
-            problem("Script 'a/aa/myscript.gradle': line 4: Project ':a' cannot dynamically look up a property in the parent project ':'")
             problem("Script 'a/aa/myscript.gradle': line 4: Project ':a:aa' cannot dynamically look up a property in the parent project ':a'")
```
alllex, line 644: ℹ️  This shows the main user-facing effect of the PR on one of the existing tests -- we are no longer reporting incidental problems for property lookup.

platforms/core-configuration/configuration-problems-base/src/main/kotlin/org/gradle/internal/configuration/problems/IsolatedProjectsProblemsListener.kt
```diff
@@ -16,9 +16,24 @@
...
 
+import org.gradle.internal.service.scopes.Scope
+import org.gradle.internal.service.scopes.ServiceScope
 
+
+/**
+ * Receives and registers Isolated Projects violation problems.
+ *
+ * NOTE: Use [IsolatedProjectsProblemsReporter] instead of this interface
+ * to ensure the mechanism of ignoring problems is honered by all reported problems.
```
reinsch82: 💅 
```suggestion
 * to ensure the mechanism of ignoring problems is honored by all reported problems.
```

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingTaskExecutionGraph.kt
```diff
@@ -147,7 +145,17 @@ class CrossProjectConfigurationReportingTaskExecutionGraph(
...
-        reportProjectIsolationProblem(requestPath)
+
+        ipProblems.report {
+            problem {
+                text("Project ")
+                reference(referrerProject.identityPath.toString())
+                text(" cannot access the tasks in the task graph that were created by other projects")
+            }.exception { message ->
+                // As the exception message is not used for grouping, we can safely add the exact task name to it:
+                message.capitalized() + if (requestPath != null) "; tried to access '$requestPath'" else '"'
```
reinsch82: ❌ Since we touch that code, this looks odd to me

```suggestion
                message.capitalized() + if (requestPath != null) "; tried to access '$requestPath'" else ""
```

platforms/core-configuration/configuration-problems-base/src/main/kotlin/org/gradle/internal/configuration/problems/IsolatedProjectsProblemsReporter.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.gradle.internal.service.scopes.ServiceScope
+
+/**
+ * Reports Isolated Projects violation problems.
+ *
+ * This must be used instead of [IsolatedProjectsProblemsListener] to ensure the mechanism of ignoring problems
+ * is honered by all reported problems.
+ */
+@ServiceScope(Scope.BuildTree::class)
+class IsolatedProjectsProblemsReporter(
```
reinsch82, line 29: ❌ I think this deserves some minimalistic unit test

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectModelAccessTrackingParentDynamicObject.kt
```diff
@@ -117,29 +141,25 @@ class CrossProjectModelAccessTrackingParentDynamicObject(
         PROPERTY, METHOD
     }
 
+    @Suppress("ThrowingExceptionsWithoutMessageOrCause")
```
reinsch82: 🤔 this is actually a caused by a false positive right?
`exception()` isn't a exception constructor call.
Maybe we should add a comment explaining that.

reinsch82 (commented):
2 things,
but overall good 😃

platforms/core-configuration/configuration-problems-base/src/main/kotlin/org/gradle/internal/configuration/problems/IsolatedProjectsProblemsReporter.kt
alllex: Fair point

alllex: Added some tests

reinsch82 (approved):

PR #37656, opened 2026-04-23
Upgrade BouncyCastle to 1.84

ljacomet:
This makes sure Gradle uses a version where https://github.com/advisories/GHSA-p93r-85wp-75v3 is no longer
applicable.

Note that Gradle does not use FrodoKEM and was thus not impacted by that
vulnerability.


build-logic/uber-plugins/src/main/kotlin/gradlebuild.ci-lifecycle.gradle.kts
```diff
@@ -39,7 +39,7 @@ fun TaskContainer.registerEarlyFeedbackLifecycleTasks() {
     register("sanityCheck") {
         description = "Run all basic checks (without tests) - to be run locally and on CI for early feedback"
         group = "verification"
-        dependsOn("compileAll", "codeQuality", "projectHealth")
+        dependsOn("compileAll", "codeQuality")
```
jvandort, line 42: How come we made this change?

ljacomet: See explanation in first commit of the PR

ov7a (approved):

build-logic/uber-plugins/src/main/kotlin/gradlebuild.ci-lifecycle.gradle.kts
jvandort: Thats unfortunate. Is there an associated issue on the dependency analysis plugin?

ljacomet:
I've added a Jackson version bump to the PR. The versions are on two consecutive lines and so separate PRs would conflict.

build-logic/uber-plugins/src/main/kotlin/gradlebuild.ci-lifecycle.gradle.kts
ljacomet: No, I am going to check if the latest version has the same issue. But in Gradle 8.x we are running version 1.33 which is the latest in that line.

I could file an issue and see if Tony would take a fix and make a 1.34 release.

ljacomet: So I ended up upgrading and finding a workaround: #37664 

And I filed an issue on the plugin itself: autonomousapps/dependency-analysis-gradle-plugin#1692

PR #37659, opened 2026-04-23
Replace ProjectInternal with ProjectState in GradleInternal

octylFractal:
This ensures we acquire locks in more places

Part of #34469 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsBuildStateAccessIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsBuildStateAccessIntegrationTest.groovy
@@ -68,7 +68,7 @@ class IsolatedProjectsBuildStateAccessIntegrationTest extends AbstractIsolatedPr
         "rootProject {gradle.parent.getRootProject()}"                          | ["Build ':build-logic' cannot access Gradle.getRootProject on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
         "rootProject {gradle.parent.getIncludedBuilds()}"                       | ["Build ':build-logic' cannot access Gradle.getIncludedBuilds on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
         "rootProject {gradle.parent.includedBuild('build-logic')}"              | ["Build ':build-logic' cannot access Gradle.includedBuild on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
-        "rootProject {gradle.parent.getDefaultProject()}"                       | ["Build ':build-logic' cannot access Gradle.getDefaultProject on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
+        "rootProject {gradle.parent.getDefaultProjectState()}"                  | ["Build ':build-logic' cannot access Gradle.getDefaultProjectState on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
 
         // Internal API
         "getProjectEvaluationBroadcaster()"                                     | ["Build ':build-logic' cannot access Gradle.getProjectEvaluationBroadcaster on build ':'"]
@@ -113,12 +113,12 @@ class IsolatedProjectsBuildStateAccessIntegrationTest extends AbstractIsolatedPr
             "parent.getOwner()",
 
             // Internal API, available on settings level only when rootProject is evaluated.
-            // In the end, we need to get a ProjectInternal instance somehow to try to set it to the parent
+            // In the end, we need to get a ProjectState instance somehow to try to set it to the parent.
 
             // We use "pre-wrapped" parentGradle because calling `gradle.parent` within the closure will return cross-project reporting Gradle,
             // which doesn't prohibit these calls.
-            "rootProject {parentGradle.setRootProject(it)}",
-            "rootProject {parentGradle.setDefaultProject(it)} ",
+            "rootProject {parentGradle.setRootProjectState(((org.gradle.api.internal.project.ProjectInternal) it).owner)}",
+            "rootProject {parentGradle.setDefaultProjectState(((org.gradle.api.internal.project.ProjectInternal) it).owner)} ",
         ]
     }
 
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsBuildStateAccessIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsBuildStateAccessIntegrationTest.groovy
@@ -68,7 +68,6 @@ class IsolatedProjectsBuildStateAccessIntegrationTest extends AbstractIsolatedPr
         "rootProject {gradle.parent.getRootProject()}"                          | ["Build ':build-logic' cannot access Gradle.getRootProject on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
         "rootProject {gradle.parent.getIncludedBuilds()}"                       | ["Build ':build-logic' cannot access Gradle.getIncludedBuilds on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
         "rootProject {gradle.parent.includedBuild('build-logic')}"              | ["Build ':build-logic' cannot access Gradle.includedBuild on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
-        "rootProject {gradle.parent.getDefaultProjectState()}"                  | ["Build ':build-logic' cannot access Gradle.getDefaultProjectState on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
 
         // Internal API
         "getProjectEvaluationBroadcaster()"                                     | ["Build ':build-logic' cannot access Gradle.getProjectEvaluationBroadcaster on build ':'"]
@@ -117,7 +116,6 @@ class IsolatedProjectsBuildStateAccessIntegrationTest extends AbstractIsolatedPr
 
             // We use "pre-wrapped" parentGradle because calling `gradle.parent` within the closure will return cross-project reporting Gradle,
             // which doesn't prohibit these calls.
-            "rootProject {parentGradle.setRootProjectState(((org.gradle.api.internal.project.ProjectInternal) it).owner)}",
             "rootProject {parentGradle.setDefaultProjectState(((org.gradle.api.internal.project.ProjectInternal) it).owner)} ",
         ]
     }
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossBuildConfigurationReportingGradle.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossBuildConfigurationReportingGradle.kt
@@ -31,6 +31,7 @@ import org.gradle.api.internal.initialization.ClassLoaderScope
 import org.gradle.api.internal.plugins.PluginManagerInternal
 import org.gradle.api.internal.project.ProjectInternal
 import org.gradle.api.internal.project.ProjectRegistry
+import org.gradle.api.internal.project.ProjectState
 import org.gradle.api.invocation.Gradle
 import org.gradle.api.invocation.GradleLifecycle
 import org.gradle.api.plugins.ExtensionContainer
@@ -119,10 +120,15 @@ class CrossBuildConfigurationReportingGradle(
     // region mutable state
     override fun getRootProject(): ProjectInternal {
         onBuildMutableStateAccess("getRootProject")
-        return delegate.getRootProject()
+        return delegate.rootProjectState.mutableModel
     }
 
-    override fun setRootProject(rootProject: ProjectInternal) {
+    override fun getRootProjectState(): ProjectState {
+        onBuildMutableStateAccess("getRootProjectState")
+        return delegate.rootProjectState
+    }
+
+    override fun setRootProjectState(rootProject: ProjectState) {
         shouldNotBeUsed()
     }
 
@@ -135,12 +141,12 @@ class CrossBuildConfigurationReportingGradle(
         return delegate.taskGraph
     }
 
-    override fun getDefaultProject(): ProjectInternal {
-        onBuildMutableStateAccess("getDefaultProject")
-        return delegate.getDefaultProject()
+    override fun getDefaultProjectState(): ProjectState {
+        onBuildMutableStateAccess("getDefaultProjectState")
+        return delegate.defaultProjectState
     }
 
-    override fun setDefaultProject(defaultProject: ProjectInternal) {
+    override fun setDefaultProjectState(defaultProject: ProjectState) {
         shouldNotBeUsed()
     }
 
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossBuildConfigurationReportingGradle.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossBuildConfigurationReportingGradle.kt
@@ -120,16 +120,7 @@ class CrossBuildConfigurationReportingGradle(
     // region mutable state
     override fun getRootProject(): ProjectInternal {
         onBuildMutableStateAccess("getRootProject")
-        return delegate.rootProjectState.mutableModel
-    }
-
-    override fun getRootProjectState(): ProjectState {
-        onBuildMutableStateAccess("getRootProjectState")
-        return delegate.rootProjectState
-    }
-
-    override fun setRootProjectState(rootProject: ProjectState) {
-        shouldNotBeUsed()
+        return delegate.owner.rootProject.mutableModel
     }
 
     override fun getOwner(): BuildState {
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
@@ -86,10 +86,8 @@ class CrossProjectConfigurationReportingGradle(
         getCrossProjectRootProject()
 
     // Split out so it's clear we're not calling the @ForExternalUse method.
-    private fun getCrossProjectRootProject(): ProjectInternal = delegate.owner.rootProject.fromMutableState {
-        // Intentionally leak mutable state here, as we wrapped it in our mutable violation checks already.
-        crossProjectModelAccess.access(referrerProject, it)
-    }
+    private fun getCrossProjectRootProject(): ProjectInternal =
+        crossProjectModelAccess.accessFromState(referrerProject, delegate.owner.rootProject)
 
     override fun rootProject(action: Action<in Project>) {
         delegate.rootProject(action.withCrossProjectModelAccessCheck())
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
@@ -34,6 +34,7 @@ import org.gradle.api.internal.project.CrossProjectConfigurator
 import org.gradle.api.internal.project.CrossProjectModelAccess
 import org.gradle.api.internal.project.ProjectInternal
 import org.gradle.api.internal.project.ProjectRegistry
+import org.gradle.api.internal.project.ProjectState as InternalProjectState
 import org.gradle.api.invocation.Gradle
 import org.gradle.api.invocation.GradleLifecycle
 import org.gradle.api.plugins.ExtensionContainer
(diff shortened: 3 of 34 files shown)


subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java
```diff
@@ -280,14 +283,21 @@ public void attachSettings(@Nullable SettingsState settings) {
 
     @Override
     public ProjectInternal getRootProject() {
+        // At the very least, verify we have the lock at the time of access.
+        // In most cases other Gradle implementations will wrap the project in mutable state checks.
+        return getRootProjectState().fromMutableState(m -> m);
```
jvandort: If we go through `CrossProjectModelAccess` we shouldn't need to lock at all -- as it will give us a wrapped version during IP and an unwrapped version during vintage

subprojects/core/src/main/java/org/gradle/initialization/InstantiatingBuildLoader.java
```diff
@@ -32,7 +31,7 @@ public void load(SettingsInternal settings, GradleInternal gradle) {
 
     private void attachDefaultProject(GradleInternal gradle, ProjectDescriptorInternal defaultProjectDescriptor) {
         ProjectState defaultProject = gradle.getOwner().getProjects().getProject(defaultProjectDescriptor.path());
-        gradle.setDefaultProject(defaultProject.getMutableModel());
+        gradle.setDefaultProjectState(defaultProject);
```
jvandort, line 34: What the heck even is the "default project"?

subprojects/core/src/main/java/org/gradle/api/internal/GradleInternal.java
```diff
@@ -46,10 +46,11 @@
 @ServiceScope(Scope.Build.class)
 public interface GradleInternal extends Gradle, PluginAwareInternal {
     /**
-     * {@inheritDoc}
+     * Returns the state of the root project.
+     *
+     * @throws IllegalStateException when the root project is not yet available, see {@link #setRootProjectState(ProjectState)}
      */
-    @Override
-    ProjectInternal getRootProject() throws IllegalStateException;
```
jvandort, line 52: We might be able to override and annotate with `@ForExternalUse` -- and I think have have some build logic which prevents us from calling the method internally

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossBuildConfigurationReportingGradle.kt
```diff
@@ -119,10 +120,15 @@ class CrossBuildConfigurationReportingGradle(
     // region mutable state
     override fun getRootProject(): ProjectInternal {
         onBuildMutableStateAccess("getRootProject")
-        return delegate.getRootProject()
+        return delegate.rootProjectState.mutableModel
     }
 
-    override fun setRootProject(rootProject: ProjectInternal) {
+    override fun getRootProjectState(): ProjectState {
+        onBuildMutableStateAccess("getRootProjectState")
```
jvandort: It should be safe to access `ProjectState` instances as long as we don't call getMutableModel on it, right? -- same for `getDefaultProjectState`

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossBuildConfigurationReportingGradle.kt
```diff
@@ -119,10 +120,15 @@ class CrossBuildConfigurationReportingGradle(
     // region mutable state
     override fun getRootProject(): ProjectInternal {
         onBuildMutableStateAccess("getRootProject")
-        return delegate.getRootProject()
+        return delegate.rootProjectState.mutableModel
```
jvandort: This is not necessarily your change, but any reason we don't return a wrapped project instead of unconditionally issuing a violation? 

 I should be able to call `gradle.rootProject.name` or something from a child project without violation

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -82,7 +83,13 @@ class CrossProjectConfigurationReportingGradle(
         }
 
     override fun getRootProject(): ProjectInternal =
-        crossProjectModelAccess.access(referrerProject, delegate.rootProject)
+        delegate.rootProjectState.fromMutableState {
```
jvandort: I think it would be cleaner if `crossProjectModelAccess` allowed us to get a `Project` reference without acquiring a lock first. `crossProjectModelAccess`'s whole _job_ is to return `Project` instances that will be provided to users. It would be nice if it abstracted away locking and such

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsBuildStateAccessIntegrationTest.groovy
```diff
@@ -68,7 +68,7 @@ class IsolatedProjectsBuildStateAccessIntegrationTest extends AbstractIsolatedPr
         "rootProject {gradle.parent.getRootProject()}"                          | ["Build ':build-logic' cannot access Gradle.getRootProject on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
         "rootProject {gradle.parent.getIncludedBuilds()}"                       | ["Build ':build-logic' cannot access Gradle.getIncludedBuilds on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
         "rootProject {gradle.parent.includedBuild('build-logic')}"              | ["Build ':build-logic' cannot access Gradle.includedBuild on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
-        "rootProject {gradle.parent.getDefaultProject()}"                       | ["Build ':build-logic' cannot access Gradle.getDefaultProject on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
+        "rootProject {gradle.parent.getDefaultProjectState()}"                  | ["Build ':build-logic' cannot access Gradle.getDefaultProjectState on build ':'", "Build ':build-logic' cannot access Gradle.rootProject on build ':'"]
```
jvandort: Strange we even test for this internal API since default project is not accessible from public API -- and like I mentioned earlier I think accessing a ProjectState is "safe" but accessing getMutableModel is the unsafe part

subprojects/core/src/test/groovy/org/gradle/invocation/DefaultGradleSpec.groovy
```diff
@@ -20,6 +20,7 @@ import org.gradle.api.Action
 import org.gradle.api.Project
 import org.gradle.api.ProjectEvaluationListener
 import org.gradle.api.ProjectState
+import org.gradle.api.internal.project.ProjectState as InternalProjectState
```
jvandort: Nitpick, but I think it might be nicer if we imported this normally and referenced the public `ProjectState` as a fully qualified class name in the single place it was used. seeing `InternalProjectState` confused me for a sec

subprojects/core/src/main/java/org/gradle/execution/DefaultTasksBuildTaskScheduler.java
```diff
@@ -50,13 +49,12 @@ public void scheduleRequestedTasks(GradleInternal gradle, @Nullable EntryTaskSel
...
-            ProjectInternal project = gradle.getDefaultProject();
-
-            //so that we don't miss out default tasks
-            projectConfigurer.configure(project);
-
-            List<String> defaultTasks = project.getDefaultTasks();
-            if (defaultTasks.size() == 0) {
+            List<String> defaultTasks = gradle.getDefaultProjectState().fromMutableState(project -> {
+                //so that we don't miss out default tasks
+                projectConfigurer.configure(project);
```
jvandort, line 54: Not necessarily anything to do here, but 
`projectConfigurer.cofigure` just calls `project.getOwner().ensureConfigured();`

Might make sense to call that outside the lock (and im curious who else calls this method)

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsBuildStateAccessIntegrationTest.groovy
alllex: 💅 This was not a public API call

subprojects/core/src/main/java/org/gradle/api/internal/GradleInternal.java
```diff
@@ -71,9 +72,9 @@ public interface GradleInternal extends Gradle, PluginAwareInternal {
     TaskExecutionGraphInternal getTaskGraph();
 
     /**
-     * Returns the default project. This is used to resolve relative names and paths provided on the UI.
+     * Returns the state of the default project. This is used to resolve relative names and paths provided on the UI.
```
jvandort, line 69: Should add `@throws` for consistency with `getRootProjectState`

subprojects/core/src/main/java/org/gradle/configuration/DefaultProjectsPreparer.java
```diff
@@ -44,10 +45,11 @@ public void prepareProjects(GradleInternal gradle) {
             return;
         }
 
+        ProjectInternal rootProject = gradle.getRootProjectState().getMutableModel();
```
alllex: 💭 `ProjectConfigurer` should take `ProjectState` for all arguments, probably. There is even an overload for one of the methods. Maybe we can polish this in a follow-up

jvandort (commented):
LGTM for the most part, some minor comments

subprojects/core/src/main/java/org/gradle/api/internal/GradleInternal.java
```diff
@@ -46,10 +46,11 @@
...
 public interface GradleInternal extends Gradle, PluginAwareInternal {
     /**
-     * {@inheritDoc}
+     * Returns the state of the root project.
+     *
+     * @throws IllegalStateException when the root project is not yet available, see {@link #setRootProjectState(ProjectState)}
      */
-    @Override
-    ProjectInternal getRootProject() throws IllegalStateException;
+    ProjectState getRootProjectState() throws IllegalStateException;
```
alllex: 🤔 I wonder whether we need a dedicated method for this at all. 

There is already `getOwner().getProjects().getRootProject()`, which returns `ProjectState`. I think I'd rather be in favor of adding a shortcut method to `BuildState.getRootProject()`, because there are other places where that would be useful.

subprojects/core/src/main/java/org/gradle/api/internal/GradleInternal.java
```diff
@@ -71,9 +72,9 @@ public interface GradleInternal extends Gradle, PluginAwareInternal {
     TaskExecutionGraphInternal getTaskGraph();
 
     /**
-     * Returns the default project. This is used to resolve relative names and paths provided on the UI.
+     * Returns the state of the default project. This is used to resolve relative names and paths provided on the UI.
      */
-    ProjectInternal getDefaultProject();
+    ProjectState getDefaultProjectState();
```
alllex: 💭 I am wondering whether it would be better to avoid storing any project state in the `Gradle` instance, and instead only store the project identity of the default project. Then, whoever needs the actual state of the default project (which shouldn't be many places) can explicitly look it up and use it.

This also applies to the `getRootProjectState()`. Though, it's a bit special, because the identity of the root project is always known and doesn't even have to be stored in the `Gradle`

subprojects/core/src/main/java/org/gradle/internal/buildtree/ProblemReportingBuildActionRunner.java
```diff
@@ -92,7 +93,10 @@ public RootProjectBuildDirCollectingListener(File defaultBuildDir) {
 
         @Override
         public void projectsEvaluated(Gradle gradle) {
-            rootProjectBuildDir = gradle.getRootProject().getLayout().getBuildDirectory().getAsFile().get();
+            GradleInternal gradleInternal = (GradleInternal) gradle;
+            rootProjectBuildDir = gradleInternal.getRootProjectState().fromMutableState(rootProject ->
```
alllex: 🤔 Side note: it feels conceptually incorrect to use the build directory of the root project for the build-scope (or even build-tree) concerns.

alllex: 🤔 Side note: it feels conceptually incorrect to use the build directory of the root project for the build-scope (or even build-tree) concerns.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossBuildConfigurationReportingGradle.kt
alllex: I think in this case me might even want to throw?
The idea is that users should never call this internal API, while Gradle internals should only operate on unwrapped instances (of everything)

subprojects/core/src/main/java/org/gradle/initialization/InstantiatingBuildLoader.java
octylFractal: So if you run in a specific directory or specify `-p`, then Gradle uses that as the "default project", and will match relative task paths from there. For example, `gradle -p platforms/jvm/testing-jvm test` only runs `:testing-jvm:test`.

subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java
octylFractal: This occurs normally if we end up using one of the other implementations. Are you saying that we should drop e.g. `CrossProjectConfigurationReportingGradle` and just wrap it here all the time?

subprojects/core/src/main/java/org/gradle/configuration/DefaultProjectsPreparer.java
octylFractal: Yes, this is next on my list. I think it may be more involved, so I wanted to keep it separate.

subprojects/core/src/main/java/org/gradle/execution/DefaultTasksBuildTaskScheduler.java
octylFractal: I don't know if `ensureConfigured()` actually uses the project lock? It seems like it uses some separate lock from `WorkerLeaseService.newResource()`. Is it safe to call that separately?

subprojects/core/src/main/java/org/gradle/api/internal/GradleInternal.java
octylFractal: Currently all of the places that use this need the state. I don't think refactoring this would be very useful at the moment.

subprojects/core/src/main/java/org/gradle/api/internal/GradleInternal.java
octylFractal: I did this. It seems like it's supposed to go on `Gradle` though, so I put it there.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossBuildConfigurationReportingGradle.kt
octylFractal: I'd like to know too, let's talk about this later.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
octylFractal: So from trying this out, it seems like there might be some issues with migrating all users of `CrossProjectModelAccess` to a new `ProjectState`-based `access`, as resilient TAPI uses some of it and must be able to access the mutable model "after failure". I think it might be better for us to migrate all of that at once, rather than to make just this one call do it for now. WDYT?

jvandort: If it's not simple or straightforward to take care of this now we can address it later. I think the general pattern we want to get to is if you have a Project instance, it is either wrapped (it came from CrossProjectModelAccess) or it is unwrapped and you have a lock on that project and got it from an applyToMutableState (or similar).

What we have here should be safe, though it seems odd imo to pass a Project instance into the thing that gives you a wrapped instance, as the instance you are holding now should either be already wrapped or you should already hold the lock.

subprojects/core/src/main/java/org/gradle/execution/DefaultTasksBuildTaskScheduler.java
jvandort: Good question. This might be something we should investigate. If you do need the project lock to call ensureConfigured, it should not be on the ProjectState type. ProjectState should contain only behavior that you can use safely without a lock.

subprojects/core/src/main/java/org/gradle/api/internal/project/CrossProjectModelAccess.java
```diff
@@ -46,10 +46,30 @@ public interface CrossProjectModelAccess {
...
+     * Variant of {@link #access(ProjectInternal, ProjectInternal)} that takes a {@link ProjectState},
+     * to avoid needing to have a reference to the mutable state of the project in the caller.
+     *
+     * <p>
+     * This should be preferred over {@code access(ProjectInternal, ProjectInternal)} where possible.
+     * </p>
+     *
+     * @param referrer The project from which the return value will be used.
+     */
+    ProjectInternal accessFromState(ProjectInternal referrer, ProjectState projectState);
```
alllex, line 71: 💭 This makes me think that the `CrossProjectModelAccess` interface should only accept `ProjectIdentity` as an argument. I didn't verify all uses, but I have a hunch that we don't really need the full-fledged `ProjectInternal` instance as a referrer. Same for the access target.

Extending this even further, I now question the existance of this interface at all. Gradle developers need to remember not to use this extra step of project lookup, somewhat on top of `ProjectRegistry` or `ProjectStateRegistry`.

I think there is potential here in reducing the number of internal entities and juggling/carrying less of full-blown state types like `Project` or `ProjectState`, when immutable identities would do.

That's obviously out of scope of this PR.

subprojects/core/src/main/java/org/gradle/api/internal/GradleInternal.java
```diff
@@ -97,20 +94,12 @@ public interface GradleInternal extends Gradle, PluginAwareInternal {
...
-     * Called by the BuildLoader after the root project is determined.  Until the BuildLoader
-     * is executed, {@link #getRootProject()} will throw {@link IllegalStateException}.
+     * Called by the BuildLoader after the default project is determined. Until the BuildLoader
+     * is executed, {@link #getDefaultProjectState()} will throw {@link IllegalStateException}.
      *
-     * @param rootProject The root project for this build.
+     * @param defaultProject The state of the default project for this build.
      */
-    void setRootProject(ProjectInternal rootProject);
+    void setDefaultProjectState(ProjectState defaultProject);
```
alllex: 💅 Since we are touching this, would it be better to move the setter next to the getter?

subprojects/core/src/main/java/org/gradle/internal/build/BuildState.java
```diff
@@ -97,6 +97,17 @@ default BuildIdentifier getBuildIdentifier() {
      */
     BuildProjectRegistry getProjects();
 
+    /**
+     * {@return the root project of this build}
+     *
+     * <p>
+     * This is a convenience method for {@link #getProjects()}{@code .}{@link BuildProjectRegistry#getRootProject() getRootProject()}.
+     * </p>
```
alllex: 💅 We can drop the closing tag, according to our javadoc styleguide

alllex (approved):

subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java
```diff
@@ -136,7 +138,11 @@ public void projectsLoaded(Gradle gradle) {
 
     @Override
     public String toString() {
-        return rootProject == null ? "build" : ("build '" + rootProject.getName() + "'");
+        if (!buildState.isProjectsLoaded()) {
+            return "build";
+        }
+        // Does the rootProject name really make sense here? Wouldn't it be better to use the build identity path?
```
jvandort, line 144: I think we should be free to change this

subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java
```diff
@@ -280,15 +285,9 @@ public void attachSettings(@Nullable SettingsState settings) {
...
-        }
-        return rootProject;
-    }
-
-    @Override
-    public void setRootProject(ProjectInternal rootProject) {
-        this.rootProject = rootProject;
+        // At the very least, verify we have the lock at the time of access.
+        // In most cases other Gradle implementations will wrap the project in mutable state checks.
+        return buildState.getRootProject().fromMutableState(m -> m);
```
jvandort: I would sorta prefer `getMutableModel` here since at least this "escaped" unwrapped project is discoverable

jvandort (approved):

blindpirate:
This is consistently failing: https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_QuickFeedbackCrossVersion_5_bucket7/112143241

subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java
octylFractal: This ended up being too much for this PR, since it affects quite a few tests. I stashed some progress in `ot/debt/change-gradle-to-string`.

PR #34767, opened 2025-08-22
Build: Update Dependency Analysis Gradle Plugin to version 3.0.0

cobexer:
Signed-off-by: Christoph Obexer <cobexer@gradle.com>

PR #37664, opened 2026-04-24
Re-introduce dependency analysis plugin

ljacomet:
Bump version to 3.4.1, including an explicit dependency to the latest shaded ASM version, to not fail on multi-release JAR with Java 25 classes.


ljacomet:
Since there are no code changes, only build setup changes and that the verification changes match was was done in #34767, sending for merge directly.

cobexer (approved):

cobexer:
Looks like updating the dependency analysis plugin causes Kotlin problems:

https://ge.gradle.org/c/j5cyv4jwdx5bo/yy5qvshulryfs/dependencies (A = `release8x`, B = this branch)

:build-logic:root-build compileClasspath: 
org.jetbrains.kotlin:kotlin-stdlib:`2.0.21` -> `2.1.20`

And later:
>   Error:
  e: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected version is 2.0.0. (kotlin-metadata.kotlin_module)
  e: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected version is 2.0.0. (metadata.jvm.kotlin_module)
  e: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected version is 2.0.0. (kotlin-metadata-jvm.kotlin_module)

mlopatkin (commented):
`/core-*` LGTM, but I haven't reviewed anything substantial.

PR #37665, opened 2026-04-24
Decouple ProjectState impl from containing registry

alllex:
The logic of managing Project-scoped `ProjectState` and corresponding mutable model (`Project`) today is scattered around many classes, making it hard to understand and maintain. We should treat the user-facing mutable model `Project` with care and generally avoid using it internally, preferring `ProjectState` or, where possible, more lightweight alternatives such as `ProjectIdentity`.

In order to be able to consolidate the Project-state management logic, we are decoupling the `ProjectState` class from its surrounding registry. The registry's purpose can be limited to the creation of the `ProjectState` objects and management of project- and build-scoped locks. This change will accommodate making the implementation of `ProjectState` deeper in the future, while keeping clear separation of concerns.

### Changes

- Introduce a more targeted `ProjectStateLookup` interface that is useful in many contexts, normally, after the projects were configured
- Relying on the new interface, decouple the impl of `ProjectState` from depending on the project state registry impl
- Finally, move the impl of `ProjectState` into `DefaultProjectState`

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectState.java
+++ b/subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectState.java
@@ -38,12 +38,75 @@ import java.util.function.Function;
  */
 @ThreadSafe
 public interface ProjectState extends ModelContainer<ProjectInternal> {
-    DisplayName getDisplayName();
 
     /**
      * Returns the containing build of this project.
      */
     BuildState getOwner();
+
+    // region About this project
+
+    /**
+     * Gets the identity of this project, containing the identity within the build tree and the owning build.
+     */
+    ProjectIdentity getIdentity();
+
+    /**
+     * Returns an identifying path for this project in the build tree.
+     */
+    default Path getIdentityPath() {
+        return getIdentity().getBuildTreePath();
+    }
+
+    /**
+     * Returns a path for this project within its containing build. These are not unique within a build tree. Use instead {@link #getIdentityPath()} to uniquely this project.
+     */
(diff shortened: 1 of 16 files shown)


alllex:
@coderabbitai review

alllex:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectState.java
```diff
@@ -38,81 +38,104 @@
...
-     * Returns the project directory.
+     * Returns the identifier of the default component produced by this project.
      */
-    File getProjectDir();
+    ProjectComponentIdentifier getComponentIdentifier();
 
     /**
      * Returns the descriptor for this project.
+     * <p>
+     * Prefer the direct methods on this class, such as {@link #getIdentity()}, {@link #getParent()}, {@link #getChildProjects()}, etc.
```
6hundreds, line 103: 🤔 `Prefer the direct methods on this class` Why?

6hundreds (approved):
LGTM!

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectState.java
alllex: For brevity, but also because this method is currently only used in the `ProjectBuilder`. There seems to be no easy way to pluck out its usage there, and I didn't want to over-explain it here. There is no harm in using the descriptor, but the direct methods on `ProjectState` should be suitable for all use cases.

Issue #4605, opened 2018-03-06
JUnit Platform: Reporting entries should be displayed on failing tests


jlink:
When a failing test also publishes report entries, those should be displayed.

### Expected Behavior
The junit platform `TestExecutionListener` has a method `reportingEntryPublished` which is called when tests report additional data. Those entries should be displayed/logged in gradle's output when a test fails. They should also show up in the generated html report.

### Current Behavior

Nothing is shown.

### Context

IntelliJ, Eclipse, Maven and the junit-platform-gradle-plugin actually report the entries.

### Steps to Reproduce (for bugs)

Running the following test through gradle should not only report a failure but also something
like `timestamp = 2018-03-06T15:40:33.704, key = value` (IntelliJ's output)

	@Test
	void failingTestWithReporting(TestReporter reporter) {
		reporter.publishEntry("key", "value");
		org.junit.jupiter.api.Assertions.fail("Fails with an additional report entry");
	}

### Your Environment
OSX, Java 8, Gradle 4.6

oehme:
As mentioned on the #4716: 

We should overhaul our test logging infrastructure holistically. We have already added display names, but only to the html report and console log. Now we are considering JUnit 5's reporting entries.

We should decide what things we want to support in a cross-cutting way (so that other test frameworks can be adapted to it). A display name makes a lot of sense. A `Map<String, String>` of "additional data" also seems like something other test frameworks might be interested in.

This information should be exposed to

- Build scans
- TestListeners
- Tooling API (so IDEs can display it)
- Console
- HTML report

@blindpirate before we continue here we should have a small design spec that addresses the above and proposes an API. We should make sure this API also makes sense for the native tests frameworks for instance. 

We should not just add reporting entries as test output, because otherwise people might start depending on this implementation detail and try to parse that output. Then we'd be stuck if we wanted to change it to something more strongly modeled. Let's do some design first.

sbrannen:
Please keep in mind that reporting features like display names and reporting entries (via a map of key-value pairs) will end up in the new reporting format from the Open Test Alliance. So that should make them mainstream... once that new format is supported.

oehme:
@sbrannen is there already a spec for the format?

sbrannen:
No, unfortunately not (yet).

That will be discussed here: https://github.com/ota4j-team/opentest4j/issues/9

So feel free to chime in there with comments.

jlink:
Any updates on this one?

[...]

PR #37669, opened 2026-04-25
Spike for adding jqwik for property-based testing

octylFractal:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


subprojects/core/src/test/groovy/org/gradle/util/internal/VersionNumberTest.groovy
```diff
@@ -16,150 +16,370 @@
...
+                        .appendText(", patch=").appendValue(patch)
+                        .appendText(", qualifier=").appendValue(qualifier)
+                }
+            })
+    }
 
-        VersionNumber.parse("1.2") == new VersionNumber(1, 2, 0, null)
-        VersionNumber.parse("1.2.3") == new VersionNumber(1, 2, 3, null)
+    @Property
+    def "construction"(@ForAll int major, @ForAll int minor) {
```
tresat, line 77: 💅 Shouldn't these be `@Positive` only?

subprojects/core/src/test/groovy/org/gradle/util/internal/VersionNumberTest.groovy
```diff
@@ -16,150 +16,370 @@
...
-        VersionNumber.parse("1.rc1-SNAPSHOT") == new VersionNumber(1, 0, 0, "rc1-SNAPSHOT")
-        VersionNumber.parse("1.2.rc1-SNAPSHOT") == new VersionNumber(1, 2, 0, "rc1-SNAPSHOT")
-        VersionNumber.parse("1.2.3.rc1-SNAPSHOT") == new VersionNumber(1, 2, 3, "rc1-SNAPSHOT")
+        def fullVersion = "$major.$minor.$micro"
+        assertThat(
+            "major, minor and micro", VersionNumber.parse(fullVersion),
+            isVersion(major, minor, micro, null),
+        )
+        assertThat(
+            "major, minor and micro, doubled", VersionNumber.parse(fullVersion + "." + fullVersion),
```
tresat, line 104: ❓ What is this format?  This is testing that if you have more than the expected number of periods in the format, everything else just gets dumped into the patch?  

It's still parsing, but maybe should be a separate test: "degenerate parsing".  Also, shouldn't patch be tested as a `String` as well - we can put anything at all in there, right?

testing/internal-testing/src/main/groovy/org/gradle/util/internal/MoreArbitraries.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Provides extra edge cases to an existing arbitrary.
+     * Primarily used when a specific test requires edge cases that are not generally applicable to all tests
+     * using the base arbitrary.
+     *
+     * @param base the base arbitrary to which the extra edge cases should be added
+     * @param extraEdgeCases the extra edge cases to include in the resulting arbitrary
+     * @return a new arbitrary that includes the extra edge cases in addition to the values generated by the base arbitrary
+     */
+    public static <T extends @Nullable Object> Arbitrary<T> includeExtraEdgeCases(
```
tresat, line 37: This is a great idea.  I remember adding additional non-property-based unit tests to cover things like this.

subprojects/core/src/test/groovy/org/gradle/util/internal/VersionNumberTest.groovy
```diff
@@ -16,150 +16,370 @@
...
+            Tuple.of("a.1"),
+            Tuple.of("1_2"),
+            Tuple.of("1_2_2"),
+            Tuple.of("1.2.3_4"),
+        )
+    }
+
+    @Property
+    @FromData("unparseableStrings")
+    def "unparseable input returns UNKNOWN"(@ForAll String s) {
```
tresat, line 265: ❔ It's kind of annoying that you have to specify a method name in a string to reference it.  Is there any more refactoring-safe way to reference the method?

subprojects/core/src/test/groovy/org/gradle/util/internal/VersionNumberTest.groovy
```diff
@@ -16,150 +16,370 @@
...
+            isVersion(major, minor, micro, patch, qualifier),
+        )
+        assertThat(
+            "major, minor, micro and patch with '.'", VersionNumber.withPatchNumber().parse("$major.$minor.$micro.$patch.$qualifier"),
+            isVersion(major, minor, micro, patch, qualifier),
+        )
+    }
+
+    @Data
+    Iterable<Tuple.Tuple1<String>> unparseableStrings() {
```
tresat, line 245: ❔ Is `Table` of `Tuple` required here?  Can't this just be a list of strings?

Also, isn't this not property-based testing and just a typical parameterized test?  Shouldn't this be a `@Provide` method that generates a bunch of arbitrary strings and filters the known good ones?

tresat (commented):
I think Property-based testing is super-interesting and this is a great experiment.

A few thoughts:

1. In the user guide here: https://jqwik.net/docs/current/user-guide.html#jqwik-configuration itt mentions: 

Seeing jqwik Reporting in Gradle Output
Since Gradle does not yet support JUnit platform reporting – [see this Github issue](https://github.com/gradle/gradle/issues/4605) – jqwik has switched to do its own reporting by default. This behaviour [can be configured](https://jqwik.net/docs/current/user-guide.html#jqwik-configuration) through parameter jqwik.reporting.usejunitplatform (default: false).

Have you looked at this at all?  We fixed this in 9.4.

2. Fixed seeds - Is the idea that each time this runs, new random seeds are used to generate properties?  Isn't that a recipe for flakiness?  Shouldn't this used a fixed seed?

3. Compacting/Shrinking - 1000 param values seems excessive for these tests, even though they run fairly quick.  Have you tried Shrinking the set of tested properties to only interesting boundary values?

I feel like the primary benefit of a property-based testing approach is not just to easily throw a variety of different inputs at a test, but to discover **interesting classes of inputs** via compressing a large random sample into representatives of each class + boundary conditions.  Testing 1000 different `major` version numbers is probably mostly not useful.  

Have you though about how this could work here?

octylFractal:
Unfortunately I now can't in good faith recommend using this library because of https://github.com/jqwik-team/jqwik/issues/708. We'll have to find some other property testing library.

PR #37705, opened 2026-04-27
Fix release7x builds

blindpirate:
Now we can't merge anything to `release7x` because:

- JDK 19 used in the builds is not LTS and removed by most build agents.
- `openjdk:11-jre-slim` -> `eclipse-temurin:11-jre`
- Don't check snapshot repository for distributions

This PR fixes it by:

- Introducing `jdks.yaml` so it can have JDK 19 from the plugin.
- Disabled the failed test.


blindpirate:
@bot-gradle test this

PR #37723, opened 2026-04-28
Make parameters-related configuration more consistent

### LINKED ISSUE #37638: Make parameters-related configuration more consistent
> 🤔  It might be beneficial to also look at the configuration side of things, e.g. at what you get when calling `getParameters()` or `parameters` when registering build service, transform, etc.
> 
> So far, there is certain level of inconsistency as well:
> - registering `ValueSource` and `BuildService` return `null` from `get` and throw when calling `parameters {}`
> - registering `TransformAction` throws `IllegalStateException`
> - `FlowAction` and `WorkAction` just doesn't run configuration closure, so there is no place to call the methods.
> 
> Aligning first two and the last in either direction would be a breaking change, unfortunately. However, if we're to change, I vote for:
> - Running configuration closure regardless of type (changing `FlowAction` and `WorkAction` behavior)
> - Returning/supplying `None` instance rather than `null` or throwing.

_Originally posted by @mlopatkin in https://github.com/gradle/gradle/pull/37543#pullrequestreview-4153212349_

ov7a:
Standardize params: always run configuration even if params are None and return None instead of null for getParams.

On the serialization level, still treat None params as null.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/core-execution/daemon-server-worker/src/main/java/org/gradle/workers/WorkParameters.java
```diff
@@ -42,15 +42,18 @@ public interface WorkParameters {
      *
      * @since 5.6
      */
-    final class None implements WorkParameters {
+    @Incubating
+    enum None implements WorkParameters {
```
ov7a: 🤔 it's weird that binary compatibility check is not triggered here

mlopatkin: > 🤔 it's weird that binary compatibility check is not triggered here

Why would it? Converting `final class Foo extends Object` to `enum` is a source- and binary-compatible change. Or do you refer to adding `@Incubating`?

ov7a:
@coderabbitai review

platforms/core-execution/daemon-server-worker/src/main/java/org/gradle/workers/WorkParameters.java
ov7a: > Converting final class Foo extends Object to enum is a source- and binary-compatible chang

I wasn't aware of that.

>  Or do you refer to adding @Incubating?

No, it was triggered properly

platforms/core-execution/daemon-server-worker/src/main/java/org/gradle/workers/WorkParameters.java
```diff
@@ -42,15 +42,18 @@ public interface WorkParameters {
      *
      * @since 5.6
      */
-    final class None implements WorkParameters {
+    @Incubating
```
mlopatkin: 💅 We can move the annotation to the `INSTANCE` constant instead of the type.

platforms/core-execution/workers/src/main/java/org/gradle/workers/WorkQueue.java
```diff
@@ -34,6 +34,9 @@ public interface WorkQueue {
      * Work submitted using {@link WorkerExecutor#processIsolation()} will execute in an idle daemon that meets the requirements set
      * in the {@link ProcessWorkerSpec}.  If no idle daemons are available, a new daemon will be started.  Any errors
      * will be thrown from {@link #await()} or from the surrounding task action if {@link #await()} is not used.
+     *
+     * <p>The {@code parameterAction} is invoked even when the work action declares {@link WorkParameters.None} as its
+     * parameters type; in that case it receives the {@link WorkParameters.None#INSTANCE singleton}.</p>
      */
     <T extends WorkParameters> void submit(Class<? extends WorkAction<T>> workActionClass, Action<? super T> parameterAction);
```
mlopatkin, line 41: 🤔 This change in behavior is likely upgrade-guide worthy.

Also applies to other places where we start invoking callbacks.

subprojects/core-api/src/main/java/org/gradle/api/artifacts/transform/TransformParameters.java
```diff
@@ -47,15 +47,18 @@ public interface TransformParameters {
      *
      * @since 5.3
      */
-    final class None implements TransformParameters {
+    @Incubating
```
mlopatkin: 💅 Same, `Incubating` can be next to `INSTANCE`.

subprojects/core-api/src/main/java/org/gradle/api/provider/ValueSourceParameters.java
```diff
@@ -38,16 +38,18 @@ public interface ValueSourceParameters {
      *
      * @since 6.1
      */
-    final class None implements ValueSourceParameters {
+    @Incubating
```
mlopatkin: 💅 Same, `Incubating` can be next to `INSTANCE`.

subprojects/core/src/main/java/org/gradle/api/services/internal/BuildServiceDetails.java
```diff
@@ -44,7 +45,7 @@ public BuildServiceDetails(BuildIdentifier buildIdentifier, String name, Class<T
         this.buildIdentifier = buildIdentifier;
         this.name = name;
         this.implementationType = implementationType;
-        this.parameters = null;
+        this.parameters = Cast.uncheckedNonnullCast(BuildServiceParameters.None.INSTANCE);
```
mlopatkin: 🤔 This is semantically different. This represents an unresolved service, for which parameters aren't available, but it isn't necessarily `None`. I'd make the `parameters` `@Nullable`, keep the `null` assignment here and add `requireNonNull` to `getParameters` below - we don't and must not call the method for an unresolved service.

mlopatkin (commented):
LGTM, though I'd like to see the upgrade guide section.

It would also be great to have a second opinion about enum singletons we're using here.

platforms/core-execution/daemon-server-worker/src/main/java/org/gradle/workers/WorkParameters.java
ov7a: Unfortunately, that's cumbersome because of
>  Method org.gradle.workers.WorkParameters$None.values(): Is not annotated with @Incubating.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -675,11 +528,22 @@ The default version of CodeNarc has been updated from 3.6.0 to link:https://gith
...
+
+1. **The configuration `Action` is now invoked even when the action declares `None` as its parameters type.** Calls like `WorkQueue.submit(MyAction.class, p -> ...)`, `FlowActionSpec.parameters { ... }`, `ValueSourceSpec.parameters { ... }`, `BuildServiceSpec.parameters { ... }` and `TransformSpec.parameters { ... }` now always run the configuration block. When the parameters type is `None`, the block receives the `None.INSTANCE` singleton instead of being skipped. Existing blocks that just configured properties on a real parameters type are unaffected; any code that relied on the block being silently skipped for `None` will now run.
```
```suggestion
1. **The configuration `Action` is now invoked even when the action declares `None` as its parameters type.**
Calls like `WorkQueue.submit(MyAction.class, p -> ...)`, `FlowActionSpec.parameters { ... }`, `ValueSourceSpec.parameters { ... }`, `BuildServiceSpec.parameters { ... }` and `TransformSpec.parameters { ... }` now always run the configuration block.
When the parameters type is `None`, the block receives the `None` singleton instead of being skipped.
Existing blocks that just configured properties on a real parameters type are unaffected; any code that relied on the block being silently skipped for `None` will now run.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -675,11 +528,22 @@ The default version of CodeNarc has been updated from 3.6.0 to link:https://gith
...
+
+2. **`getParameters()` returns `None.INSTANCE` instead of `null` for parameterless actions.** Inside a `WorkAction`, `TransformAction`, `BuildService`, `ValueSource` or `FlowAction` that declares `None` as its parameters type, `getParameters()` now returns the `None.INSTANCE` singleton. Previously it could return `null`. Null-checks against the result are no longer required.
```
```suggestion
2. **`getParameters()` returns the `None` singleton instead of `null` for parameterless actions.**
Inside a `WorkAction`, `TransformAction`, `BuildService`, `ValueSource` or `FlowAction` that declares `None` as its parameters type, `getParameters()` now returns the `None` singleton.
Previously it could return `null`.
Null-checks against the result are no longer required.
```

lkasso (approved):
Docs LGTM

mlopatkin (approved):
LGTM!

PR #37726, opened 2026-04-28
Deprecate artifactUrls on MavenArtifactRepository

big-guy:
The artifactUrls family of methods on MavenArtifactRepository is a Gradle-only feature with no equivalent in Maven repository declarations. Nag users via DeprecationLogger and document removal in Gradle 10 in the upgrade guide. 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


big-guy:
@bot-gradle test this

subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/RepositoryHandler.java
jvandort, line 152: Not here necessarily, but we should deprecate this method eventually as well

jvandort (approved):

subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/RepositoryHandler.java
big-guy: Yeah, already thought of that.

PR #37731, opened 2026-04-28
Retire pre-stable CC task access warnings

### LINKED ISSUE #37567: Get rid of `task-execution-access-pre-stable` internal CC flag
The `org.gradle.internal.configuration-cache.task-execution-access-pre-stable` internal flag usage is scattered around the integration tests:

https://github.com/search?q=repo%3Agradle%2Fgradle+org.gradle.internal.configuration-cache.task-execution-access-pre-stable&type=code

There might be a connection with KGP and AGP, if the older versions of the plugins relied on the CC behavior that this flag allows to preserve.

We should fix our own integration tests, currently relying on the flag.

We should also investigate whether the removal of this flag is possible in a minor release or requires a major release due to (potentially) breaking CC compatibility of older KGP/AGP versions.

mlopatkin:
Fixes #37567

### DIFF
--- a/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AndroidPluginsSmokeTest.groovy
+++ b/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AndroidPluginsSmokeTest.groovy
@@ -40,17 +40,6 @@ class AndroidPluginsSmokeTest extends AbstractPluginValidatingSmokeTest implemen
         AndroidHome.assertIsSet()
     }
 
-    @Override
-    SmokeTestGradleRunner runner(String... tasks) {
-        def runner = super.runner(tasks)
-        // TODO: AGP's ShaderCompile uses Task.project after the configuration barrier to compute inputs
-        return runner.withJvmArguments(runner.jvmArguments + [
-            // A workaround for this has been added to TaskExecutionAccessCheckers;
-            // TODO once we remove it, uncomment the flag below or upgrade AGP
-            // "-Dorg.gradle.internal.configuration-cache.task-execution-access-pre-stable=true"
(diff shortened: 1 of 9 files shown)


mlopatkin:
@bot-gradle test ACC

mlopatkin:
@bot-gradle test PRF

mlopatkin:

mlopatkin:

alllex (approved):
I didn't expect it to be as easy as no longer passing the flag in the tests 😅

alllex:
Though, some relevant tests seem to fail after the change

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AndroidPluginsSmokeTest.groovy
```diff
@@ -44,11 +44,8 @@ class AndroidPluginsSmokeTest extends AbstractPluginValidatingSmokeTest implemen
     SmokeTestGradleRunner runner(String... tasks) {
```
cobexer: ❌ looks like this override is now unused and should be removed

mlopatkin:
@bot-gradle test this

mlopatkin:
> Though, some relevant tests seem to fail after the change

These are flaky :(

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AndroidPluginsSmokeTest.groovy
mlopatkin: > ❌ looks like this override is now unused and should be removed

Thanks, I removed it

PR #37741, opened 2026-04-29
Add internal flag to fail builds on implicit parent-project lookup

reinsch82:
## Motivation

This is intended to gather early feedback from partners on deprecating parent lookup. having it fail makes sure in test sets that ignore deprecation warnings we still get a signal.

## Summary

Adds the internal option `org.gradle.internal.fail-on-parent-property-lookup`. When enabled, any property or method that resolves through the parent-project chain at the lookup site throws `InvalidUserCodeException` and fails the build immediately, instead of emitting a deprecation warning.

Intended as a CI-side enforcement / pre-flight-check tool aligned with the eventual Gradle 10 behavior in which implicit parent-walking is removed entirely.

- Wired into `ExtensibleDynamicObject` by `DefaultProject` only when the project's parent is set, so non-Project dynamic objects (extension hierarchies, etc.) are unaffected.
- Covers all four parent-walking paths: `tryGetProperty`, `hasProperty`, `hasMethod`, and the inner `doTryInvokeMethod` called by `tryInvokeMethod`.
- Default is off; existing behavior is unchanged. The option is internal and intentionally undocumented in the upgrade guide.

Failure message format:
> Implicit parent-project property lookup is not allowed: property 'foo' was resolved from root project 'root' for project ':child'. Define 'foo' explicitly in project ':child', or unset the org.gradle.internal.fail-on-parent-property-lookup system property.

## Test plan

- [x] `./gradlew :core:forkingIntegTest --tests '*ParentPropertyLookupFailOnAccessIntegrationTest*'` (8/8 pass)
- [x] `./gradlew :model-core:test --tests '*ExtensibleDynamicObject*'` (no regressions)
- [ ] CI green

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
@@ -19,6 +19,7 @@ import groovy.lang.Closure;
 import groovy.lang.MissingMethodException;
 import groovy.lang.MissingPropertyException;
 import org.codehaus.groovy.runtime.GeneratedClosure;
+import org.gradle.api.InvalidUserCodeException;
 import org.gradle.api.NamedDomainObjectContainer;
 import org.gradle.api.plugins.ExtensionContainer;
 import org.gradle.api.plugins.ExtraPropertiesExtension;
@@ -68,6 +69,7 @@ public class ExtensibleDynamicObject extends AbstractDynamicObject implements Hi
     private CompositeDynamicObject delegateObjects;
     @Nullable
     private HierarchicalDynamicObject parent;
+    private boolean failOnParentAccess;
 
     public ExtensibleDynamicObject(Object delegate, Class<?> publicType, InstanceGenerator instanceGenerator) {
         this(delegate, new BeanDynamicObject(delegate, publicType), new DefaultExtensionContainer(instanceGenerator));
@@ -132,6 +134,18 @@ public class ExtensibleDynamicObject extends AbstractDynamicObject implements Hi
         this.parent = parent;
     }
 
+    /**
+     * Enables strict mode for parent-project property/method lookup. When set, any property or
+     * method that resolves through the parent chain causes the build to fail with an
+     * {@link InvalidUserCodeException} at the lookup site. Used to opt into the eventual
+     * Gradle 10 behavior early or to enforce zero parent-walking in CI builds.
+     *
+     * @see org.gradle.api.internal.project.DefaultProject#FAIL_ON_PARENT_PROPERTY_LOOKUP
+     */
+    public void setFailOnParentAccess(boolean failOnParentAccess) {
+        this.failOnParentAccess = failOnParentAccess;
+    }
+
     public ExtensionContainer getExtensions() {
         return extensionContainer;
     }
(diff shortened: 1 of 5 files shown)


platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -260,13 +283,26 @@ private DynamicInvokeResult doTryInvokeMethod(String name, @Nullable Object... a
         while (parent != null) {
             result = parent.tryInvokeMethod(name, arguments);
             if (result.isFound()) {
+                if (failOnParentAccess) {
+                    throw failingParentAccess("method", name, parent);
+                }
```
6hundreds: 💅 I think conditioning and throwing may be included to method like 
```
fun `failOnParentAcessIfNeeded(String memberKind, String memberName, DynamicObject resolvedParent) {
    if (enabled) {
        throw
    }
}
```

subprojects/core/src/integTest/groovy/org/gradle/api/ParentPropertyLookupFailOnAccessIntegrationTest.groovy
```diff
@@ -0,0 +1,136 @@
...
+class ParentPropertyLookupFailOnAccessIntegrationTest extends AbstractIntegrationSpec {
+
+    private static final String FAIL_ON_PARENT_LOOKUP = "-Dorg.gradle.internal.fail-on-parent-property-lookup=true"
+
+    def setup() {
+        createDirs("child")
+        settingsFile << """
+            rootProject.name = "root"
+            include "child"
+        """
```
6hundreds: 💅 Use syntax-highlightin overrides
https://github.com/gradle/gradle/blob/1439b8dbe998010e410b4926acbb9bfcc9e6a4fd/testing/internal-distribution-testing/src/main/groovy/org/gradle/integtests/fixtures/LanguageSpecificTestFileFixture.groovy#L43

6hundreds: Here and everywhere below

subprojects/core/src/integTest/groovy/org/gradle/api/ParentPropertyLookupFailOnAccessIntegrationTest.groovy
```diff
@@ -0,0 +1,136 @@
...
+            def fooMethod() { "from parent" }
+        """
+    }
+
+    def "fails on implicit property reference from child"() {
+        file("child/build.gradle") << "println foo"
+
+        when:
+        executer.withArgument(FAIL_ON_PARENT_LOOKUP)
+        fails "help"
```
6hundreds: 💅 I think such invocations might be inlined `fails "help", FAIL_ON_PARENT_LOOKUP`

subprojects/core/src/integTest/groovy/org/gradle/api/ParentPropertyLookupFailOnAccessIntegrationTest.groovy
```diff
@@ -0,0 +1,136 @@
...
+
+        then:
+        failure.assertHasCause("Implicit parent-project property lookup is not allowed: property 'foo' was resolved from root project 'root' for project ':child'.")
+    }
+
+    def "fails on findProperty from child"() {
```
6hundreds: 💅 We have a bunch of similar tests that differ only by API used for trigger failure. I think we might apply Spock data driven test here

6hundreds (approved):
LGTM! I'd also be consistent when use parenthesis in integ tests, where it's possible.

alllex (approved):

PR #37746, opened 2026-04-29
Document why Android Studio sync setup is not in setup()

reinsch82:
## Summary

- Follow-up to #37714, explaining the reason for moving the initialization to the test methods.
- `:performance:writeTmpPerformanceScenarioDefinitions` (run by `sanityCheck`) executes `setup()` for every feature method even when the bodies are skipped. `AndroidSyncPerformanceTestFixture.configureStudio()` asserts `ANDROID_SDK_ROOT`, so a plain `setup()` would break `sanityCheck` on machines without that variable.
- Adds the same comment to both `IsolatedProjectsAndroidSyncPerformanceRegressionTest` and `IsolatedProjectsAndroidSyncPerformanceComparisonTest`.

## Test plan

- [x] Comment-only change; no behavior impact.
- [ ] CI green.

### DIFF
--- a/testing/performance/src/performanceTest/groovy/org/gradle/performance/crossbuild/ip/IsolatedProjectsAndroidSyncPerformanceComparisonTest.groovy
+++ b/testing/performance/src/performanceTest/groovy/org/gradle/performance/crossbuild/ip/IsolatedProjectsAndroidSyncPerformanceComparisonTest.groovy
@@ -34,6 +34,11 @@ class IsolatedProjectsAndroidSyncPerformanceComparisonTest extends AbstractCross
 
     private static int maxWorkers = 8
 
+    // Invoked from feature methods rather than Spock's setup() because
+    // :performance:writeTmpPerformanceScenarioDefinitions (run by sanityCheck)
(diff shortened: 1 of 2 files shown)


testing/performance/src/performanceTest/groovy/org/gradle/performance/crossbuild/ip/IsolatedProjectsAndroidSyncPerformanceComparisonTest.groovy
alllex, line 1: The size of the comment makes me want to deduplicate it 😅 
I also feel like it hints that the documentation-driven approach might not be the best here, because somebody calling the configureStudio method elsewhere, might bump into the same problem.

Can we instead enforce that the configureStudio method is only called from within the test method body?
This way, the usages stay the same, but any new usage will get automatically validated. And the explanation (as wild as it is!) for why this can't be called from setup() will be centralized.

I'd image Spock has some info we can check that the test body is executing. The fixture also doesn't have to be static but be a trait instead, if this helps access the info.

alllex (approved):

Issue #31372, opened 2024-11-21
Decide and add support for some type conversions for Property.setFromAnyValue


asodja:
We should add some conversions that works with eager property to also work with lazy properties for Groovy.

Current candidates:
- [] Integer -> Long, usecase: `Test.forkEvery`, to avoud `test.forkEvery = 42L`
- [] String -> Url, usecase: `GitVersionControlSpec.url = "String"` and `UrlArtifactRepository.url = "String"
- [] String -> Enum, usecase: `TestLogging.exceptionFormat = "full"` and similar

asodja:
Maybe `Integer -> Long` should be implemented as [NumberSanitizer](https://github.com/gradle/gradle/blob/9239189ef8d307b43b765f5b9637f509d8f5a436/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/ValueSanitizers.java#L26).

asodja:
Integer -> Long was implemented in https://github.com/gradle/gradle/pull/31453

mlopatkin:
It looks like the configuration team can accept implicit conversions for Groovy DSL. It should not leak into Kotlin/Java API (e.g. `set()`) even when considering type erasure. There is no strong feelings for or against deprecation of these conversions.

Supporting most of them in Kotlin DSL should be a separate decision.

asodja:
We think we support enough Property type coercion cases for Provider API migration. I will close this issue

PR #37750, opened 2026-04-29
Resolve relative GitVersionControlSpec URL strings via FileResolver

abstratt:
`setUrl(String)` now routes through `FileResolver.resolveUri()` so relative paths resolve against the settings directory, more or less consistent with `Project.uri(String)`.

Issue: #31372

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/version-control/src/main/java/org/gradle/vcs/git/internal/DefaultGitVersionControlSpec.java
+++ b/platforms/software/version-control/src/main/java/org/gradle/vcs/git/internal/DefaultGitVersionControlSpec.java
@@ -16,15 +16,21 @@
 
 package org.gradle.vcs.git.internal;
 
-import org.gradle.internal.UncheckedException;
+import org.gradle.api.internal.file.FileResolver;
 import org.gradle.vcs.git.GitVersionControlSpec;
 import org.gradle.vcs.internal.spec.AbstractVersionControlSpec;
 
+import javax.inject.Inject;
 import java.net.URI;
-import java.net.URISyntaxException;
 
 public class DefaultGitVersionControlSpec extends AbstractVersionControlSpec implements GitVersionControlSpec {
+    private final FileResolver fileResolver;
     private URI url;
+
+    @Inject
+    public DefaultGitVersionControlSpec(FileResolver fileResolver) {
+        this.fileResolver = fileResolver;
+    }
 
     @Override
     public URI getUrl() {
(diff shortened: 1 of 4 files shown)


abstratt:
@bot-gradle test this

platforms/software/version-control/src/main/java/org/gradle/vcs/git/internal/DefaultGitVersionControlSpec.java
```diff
@@ -38,12 +44,7 @@ public void setUrl(URI url) {
 
     @Override
     public void setUrl(String url) {
-        // TODO - should use a resolver so that this method is consistent with Project.uri(string)
-        try {
-            setUrl(new URI(url));
-        } catch (URISyntaxException e) {
-            throw UncheckedException.throwAsUncheckedException(e);
-        }
+        setUrl(fileResolver.resolveUri(url));
```
asodja, line 47: ❓ Can you explain a bit why this is required?

abstratt: Do you mean, why to use a resolver instead of handling it in an ad-hoc way?

The removed TODO said:

```
// TODO - should use a resolver so that this method is consistent with Project.uri(string)
```

It made sense to me (but it will have a different base path to resolve relative paths than the `Project`'s resolver). This test shows the effect:

https://github.com/gradle/gradle/blob/fbdaf914cd0d2c0801c24172541e8c652ae28156/subprojects/core/src/test/groovy/org/gradle/api/internal/file/BaseDirFileResolverTest.groovy#L244

This is going to be a behavior change, but sounds like a good one to me:
* before: relative URIs were silently accepted, but would cause a failure when attempting  to perform Git operations that needed the URI, such as [`clone`](https://github.com/gradle/gradle/blob/4cd06d872d5a12295718dbb8c0442f0292d1532a/platforms/software/version-control/src/main/java/org/gradle/vcs/git/internal/GitVersionControlSystem.java#L131) and [`ls-remote`](https://github.com/gradle/gradle/blob/4cd06d872d5a12295718dbb8c0442f0292d1532a/platforms/software/version-control/src/main/java/org/gradle/vcs/git/internal/GitVersionControlSystem.java#L118)
* now: relative URIs, if ever used, will be resolved based on the build's settings dir.

 Now, thinking about it, that was never needed before, AFAIK, so maybe a failure is better than falling back to a build-based directory. 

Maybe we should force an early failure here (say, by invoking `URI.toUrl()`). 

Or just leave it as is, and add the TODO back, which may be your preference.

platforms/software/version-control/src/main/java/org/gradle/vcs/git/internal/DefaultGitVersionControlSpec.java
```diff
@@ -38,12 +44,7 @@ public void setUrl(URI url) {
 
     @Override
     public void setUrl(String url) {
-        // TODO - should use a resolver so that this method is consistent with Project.uri(string)
-        try {
-            setUrl(new URI(url));
```
abstratt, line 43: This does not fail with a relative URI string

PR #37751, opened 2026-04-29
Use project lock to control lifecycle state management

octylFractal:
This ensures that calls like `ensureConfigured` hold the project lock when they're doing their work. Part of #34469.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectState.java
```diff
@@ -72,7 +73,7 @@ class DefaultProjectState implements ProjectState, Closeable {
         this.descriptor = descriptor;
         this.identity = descriptor.getIdentity();
         this.projectFactory = projectFactory;
-        this.controller = new ProjectLifecycleController(getDisplayName(), stateTransitionControllerFactory, buildServices);
+        this.controller = new ProjectLifecycleController(getDisplayName(), stateTransitionControllerFactory, new ModelContainerSynchronizer(this), buildServices);
```
alllex: ❓ Can we avoid sharing `this` from inside the constructor?

subprojects/core/src/main/java/org/gradle/internal/model/StateTransitionControllerFactory.java
```diff
@@ -47,6 +47,23 @@ public <T extends StateTransitionController.State> StateTransitionController<T>
...
             synchronizer = new WaitBuildOperationFiringSynchronizer(displayName, synchronizer, buildOperationRunner);
         }
+        return newController(displayName, initialState, synchronizer);
+    }
+
+    /**
+     * Creates a new controller with the given synchronizer.
+     *
+     * <p>
+     * The synchronizer is not wrapped to report locking operations, the caller must do that if needed.
```
alllex, line 57: ❓ It seems natural to be able to observe the waits if configuration of child projects blocks on their parent configuration. Does the current implementation/wiring make this work?

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectState.java
octylFractal: I could make the new `Synchronizer` more explicitly about project locks and move all the logic for `runWithModelLock` into it.

subprojects/core/src/main/java/org/gradle/internal/model/StateTransitionControllerFactory.java
octylFractal: It would be observed by waiting to acquire the parent's lock. The child lock isn't involved at all as parent configuration happens before involving the controller.

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectState.java
alllex: Do you think it will result in a better design?

octylFractal: I'm not 100% sure. I thought we could re-use it for `BuildState` if we were going to make it also a `ModelContainer`, so that its controller would also sync on the same lock. But maybe that's looking too far forward for a slightly worse design now. I'll take a look at changing it and see how it looks.

alllex: I am honestly wary of generalizations, especially if they're forward-looking. Both Project and Build have very distinct roles and lifecycles. I'd much rather we actually not treat them as any kind of shared abstraction, and instead have their handling entirely explicit. Only then, if we see that yes, the generalization simplifies things without making things too obscure, would I suggest reusing abstractions.

Regarding the suggested change, please take a look. If it's too complex to make at the moment, we don't have to do it now.

octylFractal: It seems pretty trivial, I pushed it up.

alllex: The separation of concerns now seems much better

alllex (approved):

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectStateSynchronizer.java
```diff
@@ -0,0 +1,96 @@
...
+
+final class ProjectStateSynchronizer implements Synchronizer {
+    private final WorkerLeaseService workerLeaseService;
+
+    private final ResourceLock allProjectsLock;
+    private final ResourceLock projectLock;
+
+    private final Set<Thread> canDoAnythingToThisProject = new CopyOnWriteArraySet<>();
+
+    public ProjectStateSynchronizer(BuildState owner, ProjectIdentity identity, WorkerLeaseService workerLeaseService) {
```
jvandort, line 39: Not to address here but it would be nice to eventually have a `BuildIdentity`. We have `BuildIdentifier` but this is a public type and it is a "dependency management" type so it doesn't really suffice

jvandort (approved):

octylFractal:
@alllex and @jvandort can you take another look at the latest commit to see if it makes sense to you?

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectLifecycleController.java
```diff
@@ -94,6 +94,12 @@ public ProjectInternal getMutableModelEvenAfterFailure() {
     }
 
     public void ensureSelfConfigured() {
+        // Avoid taking the controller lock if already configured, to avoid contention and deadlocks.
+        // ProjectState#ensureConfigured tries to configure parent projects, and child projects shouldn't
+        // need to all take the lock if the parent project(s) are already configured.
+        if (controller.inStateOrLater(State.Configured)) {
```
jvandort: This seems like it should be a general "feature" of `maybeTransitionIfNotCurrentlyTransitioning`. `maybeTransitionIfNotCurrentlyTransitioning` already has all the information it needs to be "smart" and not grab the lock if it doesn't need to, we should be able to put

```
        if (inStateOrLater(toState)) {
            return;
        }
```

at the top of the method. 

This way we don't need to expose a public method documented with `You should try to not use this method`

jvandort: Given that, the method would look like:

```
        if (inStateOrLater(toState)) {
            return;
        }
        synchronizer.withLock(() -> {
            if (state.inStateOrTransitioningTo(toState)) {
                return;
            }
            doTransition(fromState, toState, action);
        });
```

Which expands to 

```
        CurrentState<T> current = state;
        if (current.hasSeenStateIgnoringTransitions(toState)) {
            return;
        }
        synchronizer.withLock(() -> {
            if (state.inStateOrTransitioningTo(toState)) {
                return;
            }
            doTransition(fromState, toState, action);
        });
```

This implementation seems a bit off. Why are we short-circuiting with `hasSeenStateIgnoringTransitions` before acquiring the lock and short-circuiting with `inStateOrTransitioningTo` after acquiring the lock? 

My first intuition says we should be checking the same condition in both places

jvandort: Thinking about this a bit more deeply, there are a few scenarios I can think of:
1. We are not in the target state and we are not transitioning to the target state. We should acquire the lock
2. We are not in the target state and another thread is transitioning to the target state. We should block waiting for the other thread to transition. 
3. We are not in the target state and _this thread_ is transitioning to the target state. This is some sort of suspect reentrant behavior
4. We are in the target state or later. We should short-circuit.

Without digging deeper atm I am not quite sure if it makes sense or not that we are checking different conditions inside and outside the lock

octylFractal: This type of question is why I left it outside the controller for now, only to be used in this specific case.

I don't think we can use a "transitioning to" state outside of the lock properly. In that case, we cannot be assured that the transition must currently be being done by the current thread, and will later fail or succeed. The other thread's transition may fail, and in that case we'd like to throw rather than return silently. We can use a successful state and return fast as we know that at least this transition has not failed, and it will still throw if there is a failed state.

`maybeTransition` and `transitionIfNotPreviously` are similar methods and would also need similar consideration. I don't think it's worth trying to make sure this is perfectly safe for all of them right now.

octylFractal:
@bot-gradle test PT

Issue #37730, opened 2026-04-28
DCL: Project feature apply actions fire in wrong order under declarative DSL


ghale:
### Current Behavior

When a single project feature plugin binds to both a project type and a nested
definition of that project type (using two `bindsFeatureTo` calls), and the
feature is configured at both locations in the build script, the apply action
for the **nested definition** binding fires **before** the apply action for the
**project type** binding under the **declarative DSL**.

Under the **Kotlin DSL** the project type apply action fires first, as expected.

Concretely, with a feature `spring` bound to both `TestProjectType` and to the
`Dependencies` nested definition of `TestProjectType`, configured as:

```
testProjectType {
    spring {
        foo = "top-spring"
    }
    dependencies {
        spring {
            foo = "deps-spring"
        }
    }
}
```

the apply actions in the declarative-DSL build print:

```
spring applied to dependencies   <-- runs first
spring applied to type           <-- runs second
```

### Expected Behavior

The apply action for the project-type-level binding (`testProjectType.spring`)
should run before the apply action for the nested-definition-level binding
(`testProjectType.dependencies.spring`) — matching the declarative tree
structure and matching the behaviour observed under the Kotlin DSL.

### Context

This was caught by the new integration test
`ProjectFeatureDeclarationIntegrationTest.applies a project feature to a project feature to a project type before applying it to a nested dependencies definition`.
The test passes under the Kotlin DSL run and fails under the declarative DSL run:

```
Condition not satisfied:

output.indexOf("spring applied to type") < output.indexOf("spring applied to dependencies")
|      |                                 | |      |
|      480                               | |      365
```

### Self-contained Reproducer

```
    def 'applies a project feature to a project type before applying it to a nested dependencies definition'() {
        given:
        PluginBuilder pluginBuilder = testScenario {
            def type = projectType("testProjectType") {
                definition {
                    nested("dependencies", "Dependencies") {
                        implementsDefinition("DependenciesBuildModel") {
                            property "foo", String
                        }
                        property "foo", String
                    }
                }
            }
            projectFeature("spring") {
                definition {
                    property "foo", String
                    buildModel {
                        property "foo", String
                    }
                }
                plugin {
                    bindsFeatureTo type
                    bindsFeatureTo type.definition.className + ".Dependencies"
                    applyActionCode """
                        String marker = (parent instanceof TestProjectTypeDefinition.Dependencies) ? "dependencies" : "type";
                        System.out.println("spring applied to " + marker);
                    """
                }
            }
[...]

PR #37758, opened 2026-04-30
Fix DCL effects application order: don't group resolution results by receiver

h0tk3y:
Fixes:
* https://github.com/gradle/gradle/issues/37730


platforms/core-configuration/project-features/src/integTest/groovy/org/gradle/features/ProjectFeatureDeclarationIntegrationTest.groovy
```diff
@@ -1606,6 +1606,96 @@ class ProjectFeatureDeclarationIntegrationTest extends AbstractIntegrationSpec i
...
+            }
+        """ << DeclarativeTestUtils.nonDeclarativeSuffixForKotlinDsl
+
+        when:
+        run(":printSpringDefinition1Configuration", ":printSpringDefinition2Configuration")
+
+        then:
+        outputContains("spring applied to type")
+        outputContains("spring applied to dependencies")
+        output.indexOf("spring applied to type") < output.indexOf("spring applied to dependencies") == (!dependenciesFirst && !dependenciesInDefaults)
```
ghale, line 1684: 💭 This might read better if it was just a helper assert method: 
```
void assertApplicationOrder(List<String> applicationsInOrder) {
    def previous = null
    for (application : applicationsInOrder) {
        outputContains(application)
        if (previous) {
            assert output.indexOf(application) < output.indexOf(previous)
        }
        previous = application
    }
}
```
Then your iteration just contains a list of apply messages in the expected order.  You could even use that to get rid of the awkward `dependenciesFirst` boolean in the dcl script, but I don't know if it would be more than incrementally better.

Just a thought - your call.

ghale (approved):
Looks good to me.  I left one suggestion for making the test a little less complicated, but that's up to you.

platforms/core-configuration/project-features/src/integTest/groovy/org/gradle/features/ProjectFeatureDeclarationIntegrationTest.groovy
h0tk3y: Thanks for the suggestion! I think I'd rather keep this test plain like it is now. If we add more cases to it, your approach looks much better, but with just these three, I'd prefer keeping it all in the test method body.

PR #37759, opened 2026-04-30
Improve docs and diagnostics for beforeProject + pluginManagement.includeBuild

### LINKED ISSUE #36461: gradle.lifecycle.beforeProject does not work with convention plugins as claimed in docs
### Current Behavior

The current docs at https://docs.gradle.org/current/userguide/isolated_projects.html#new_gradle_lifecycle_callbacks recommend using this pattern to automatically apply a convention plugin to subprojects in a PI-friendly way, but in a very trivial setup this doesn't actually appear to be possible as it seems this block doesn't see the included build classpath

Image

### Expected Behavior

I expect it to work as the docs expect

### Context (optional)

_No response_

https://github.com/ZacSweers/metro/pull/1745

revert [927c00f](https://github.com/ZacSweers/metro/pull/1745/commits/927c00f0dfde50fec00ee25bc0cbec033af81f38), which puts it back into a reproducing state where it fails to sync or configure

```
Plugin with id 'metro.spotless' not found.

* Try:

* Exception is:
org.gradle.api.plugins.UnknownPluginException: Plugin with id 'metro.spotless' not found.
	at org.gradle.api.internal.plugins.DefaultPluginManager.apply(DefaultPluginManager.java:146)
	at org.gradle.api.internal.plugins.DefaultObjectConfigurationAction.applyType(DefaultObjectConfigurationAction.java:162)
	at org.gradle.api.internal.plugins.DefaultObjectConfigurationAction.access$200(DefaultObjectConfigurationAction.java:44)
	at org.gradle.api.internal.plugins.DefaultObjectConfigurationAction$3.run(DefaultObjectConfigurationAction.java:99)
	at org.gradle.api.internal.plugins.DefaultObjectConfigurationAction.execute(DefaultObjectConfigurationAction.java:185)
	at org.gradle.api.internal.project.AbstractPluginAware.apply(AbstractPluginAware.java:42)
	at org.gradle.api.internal.project.DefaultProject.apply(DefaultProject.java:1461)
	at org.gradle.kotlin.dsl.PluginAwareExtensionsKt.apply(PluginAwareExtensions.kt:33)
	at org.gradle.kotlin.dsl.PluginAwareExtensionsKt.apply$default(PluginAwareExtensions.kt:31)
	at Settings_gradle$4.execute(settings.gradle.kts:77)
	at Settings_gradle$4.execute(settings.gradle.kts:76)
```

### Gradle version

9.3.0

### Build scan URL (optional)

_No response_

_No response_

reinsch82:
## Summary

Closes the immediate doc/diagnostic gap reported in #36461: `gradle.lifecycle.beforeProject { apply(plugin = "...") }` does not work for convention plugins from `pluginManagement.includeBuild(...)` in the way the Isolated Projects guide implies, and the failure today is opaque.

Reference reproducer: [autonomousapps/composite-build-gradle-lifecycle](https://github.com/autonomousapps/composite-build-gradle-lifecycle) (Tony Robalik).

This is a Phase 1 stop-the-bleeding change: docs + diagnostic + tests, no behavior change. The ergonomic fix is tracked separately in #38181.

## Changes

- **Docs** (`isolated_projects.adoc`): list the limitation under `[[sec:current_limitations]]` and add a subsection explaining the asymmetry, recommending a settings convention plugin from the included build (with `plugins { id("X") apply false }` in settings as a simpler alternative).
- **Diagnostic** (`DefaultIsolatedProjectEvaluationListenerProvider.kt`): when an isolated `beforeProject`/`afterProject` callback fails with `UnknownPluginException` and a `pluginManagement.includeBuild(...)` is registered, rethrow with a hint pointing at the `apply false` workaround, substituting the actual missing plugin id.
- **Tests** (`IsolatedProjectsBeforeProjectPluginApplicationIntegrationTest`): negative case with the hint, both workarounds succeeding, and a configuration-cache store/load round-trip.

Why the workarounds work: plugins from `pluginManagement.includeBuild(...)` are resolved lazily and their classpath is exported only into the scope of the script whose `plugins {}` block triggered resolution — so a project-side `apply(plugin = "X")` never sees them. Both workarounds force resolution during settings evaluation, exporting the classpath into the settings scope, which project scopes inherit from.

## Test plan

- [x] `./gradlew sanityCheck`
- [x] `./gradlew :configuration-cache:forkingIntegTest --tests "*IsolatedProjectsBeforeProjectPluginApplicationIntegrationTest*"`
- [x] CI on this branch (Ready for Nightly green on an earlier head; Quick Feedback green on the current head)


reinsch82:
@bot-gradle test rfn

lkasso (approved):

alllex:
@bot-gradle test BD

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -251,6 +251,8 @@ The current implementation of Isolated Projects has a number of limitations that
 * The configuration phase might require more memory to process more work in parallel.
 * Changes to included builds invalidate all cached results, even unrelated ones.
 * Tooling model caching is local only: no sharing across checkouts or remote caching.
+* `gradle.lifecycle.beforeProject` callbacks registered in `settings.gradle(.kts)` cannot directly apply convention plugins from a build registered via `pluginManagement.includeBuild(...)`, even though the same plugins are usable from a `plugins {}` block in the same settings script.
```
alllex: > even though the same plugins are usable from a `plugins {}` block in the same settings script

This doesn't sound right. The plugins applied inside `beforeProject` are project plugins, while the `plugins {}` in settings script have to be setting plugins to be "usable". I think the fact that project plugin can appear there with `apply false` is a nuance that is not easy to guess from this wording.

subprojects/core/src/integTest/groovy/org/gradle/api/invocation/GradleLifecycleConventionPluginApplicationIntegrationTest.groovy
```diff
@@ -0,0 +1,164 @@
...
+    def "fails with hint when applying included-build plugin via beforeProject without settings plugins {} declaration"() {
+        given:
+        file("settings.gradle.kts") << """
+            pluginManagement {
+                includeBuild("build-logic")
+            }
+            rootProject.name = "consumer"
+            include("a")
+            include("b")
+            gradle.lifecycle.beforeProject {
```
alllex: 💅 This whole header can be moved into the setup. At least the pluginManagement part, I think, should be there to highlight that was is written in the setup is always part of the test build.

Fun fact, you can have imperative statements like `include("a")` appear before the `plugins {}` block, and the plugins-block will still be applied correctly. This can help us with reuse here.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsBeforeProjectPluginApplicationIntegrationTest.groovy
```diff
@@ -0,0 +1,164 @@
...
+            rootProject.name = "consumer"
+            include("a")
+            include("b")
+            gradle.lifecycle.beforeProject {
+                apply(plugin = "${CONVENTION_PLUGIN_ID}")
+            }
+        """
+
+        when:
+        isolatedProjectsRun(":a:verifyConvention", ":b:verifyConvention")
```
alllex: 💅 Since we don't do any filtering, the convention also applies to the root project. Because of it, I'd run an unqualified `verifyConvention` and validate that root project is also affected/works. Maybe we don't even need "a" and "b" for these tests and one "sub" is enough.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsBeforeProjectPluginApplicationIntegrationTest.groovy
```diff
@@ -0,0 +1,164 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl.isolated
+
+import spock.lang.Issue
+
+@Issue("https://github.com/gradle/gradle/issues/36461")
+class IsolatedProjectsBeforeProjectPluginApplicationIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
```
alllex: ❓ Should we add a test showing that things work with buildSrc?

subprojects/core/src/integTest/groovy/org/gradle/api/invocation/GradleLifecycleConventionPluginApplicationIntegrationTest.groovy
```diff
@@ -0,0 +1,164 @@
...
+        isolatedProjectsRun(":a:verifyConvention", ":b:verifyConvention")
+
+        then:
+        outputContains("convention applied to :a")
+        outputContains("convention applied to :b")
+    }
+
+    def "applying included-build plugin via beforeProject works from a settings convention plugin"() {
+        given:
+        file("build-logic/src/main/groovy/my.lifecycle.settings.gradle") << """
```
alllex, line 177: ❓ Should we also have a test where the precompiled plugin is writting in Kotlin?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsBeforeProjectPluginApplicationIntegrationTest.groovy
```diff
@@ -0,0 +1,164 @@
...
+
+        when:
+        isolatedProjectsRun(":a:verifyConvention", ":b:verifyConvention")
+
+        then:
+        outputContains("convention applied to :a")
+        outputContains("convention applied to :b")
+    }
+
+    def "applying included-build plugin via beforeProject reuses configuration cache"() {
```
alllex: ❌ it doesn't feel like this test belongs here.

Actually, thinking about it, the behavior of the lifecycle callbacks should be the same across all modes (VT, CC, IP). Hence the test coverage shouldn't be IP specific, and instead live somewhere, where it is executed with all three modes.

This test specifically would then probably work best by _assuming_ CC is enabled via the contextual executor and then validating the cache hit. Then it will be check both with IP _and_ CC.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -701,6 +703,62 @@ include::sample[dir="snippets/structuring-builds/isolatedProjects-lifecycleApi/k
...
+When Gradle sees `plugins { id("my.convention") }`, it builds `my.convention` from the included build and exports the resulting classpath into the surrounding script's classloader scope.
+
+There is a limitation however:
+
+- Using a `plugins {}` block in `settings.gradle(.kts)` or any project script *works*, because the block itself triggers the build of the plugin.
+- Using the plugin from inside a `gradle.lifecycle.beforeProject` (or `afterProject`) callback registered in the settings script *does not work*.
+The callback only registers an action; it doesn't request anything.
+By the time the callback runs at project configuration, nothing has asked Gradle to build the plugin, so `apply(plugin = "...")` fails with `Plugin with id '...' not found`.
+
+By contrast, `buildSrc/` is built eagerly and its classpath is automatically exported to all projects.
```
alllex: ❓ Isn't this one of the viable alternatives?

It might not be. I think mentioning it is definitely important, but I'd add a note about the best practice that recommends against it.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -701,6 +703,62 @@ include::sample[dir="snippets/structuring-builds/isolatedProjects-lifecycleApi/k
...
+The limitation above does not apply there.
+
+Two workarounds are available.
+
+[[sec:lifecycle_via_settings_convention_plugin]]
+====== Recommended: register `beforeProject` from a settings convention plugin
+
```
alllex: 💅 I am not certain h6 is worth it. The header looks smaller than regular text, when rendered:

https://builds.gradle.org/repository/download/Gradle_Master_Check_BuildDistributions/113887719:id/distributions/gradle-9.7.0-docs.zip!/gradle-9.7.0-20260612082444%2B0000/docs/userguide/isolated_projects.html#sec:lifecycle_via_apply_false

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -701,6 +703,62 @@ include::sample[dir="snippets/structuring-builds/isolatedProjects-lifecycleApi/k
...
+For simple cases where adding a settings plugin to your included build is unwarranted, declare the project convention plugin in the settings `plugins {}` block with `apply false`.
+This forces resolution at settings evaluation time and exports the classpath into the settings classloader scope, the parent of every project's scope, so the `beforeProject` callback can find the plugin:
+
+include::sample[dir="snippets/isolatedProjects/lifecycleApplyFalse/kotlin",files="settings.gradle.kts[tags=settings-file]"]
+include::sample[dir="snippets/isolatedProjects/lifecycleApplyFalse/groovy",files="settings.gradle[tags=settings-file]"]
+
+[NOTE]
+====
+Both patterns are workarounds for the current implementation limitation listed under <<sec:current_limitations>>.
+A more direct way to make convention plugins from `pluginManagement.includeBuild` available to `gradle.lifecycle` callbacks is planned.
```
alllex: 🤔 We should avoid making promises in the userguide. Let's keep it informational about the current state of things

alllex (changes requested):
Great finding that there is more than one workaround 👍 

In this review round I mostly focused on the userguide and tests. I haven't looked at the implementation of the hint yet

reinsch82:
@bot-gradle test this

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -702,6 +704,59 @@ include::sample[dir="snippets/structuring-builds/isolatedProjects-lifecycleApi/k
...
+- Using the plugin from inside a `gradle.lifecycle.beforeProject` (or `afterProject`) callback registered in the settings script *does not work*.
+The callback only registers an action; it doesn't request anything.
+By the time the callback runs at project configuration, nothing has asked Gradle to build the plugin, so `apply(plugin = "...")` fails with `Plugin with id '...' not found`.
+
+This limitation does not apply to `buildSrc/`: it is built eagerly and its classpath is exported to every project, so a `gradle.lifecycle.beforeProject` callback can apply a `buildSrc` convention plugin directly, with none of the workarounds below.
+Moving build logic into `buildSrc` solely to sidestep this limitation is not recommended, however; Gradle best practice is to <<best_practices_structuring_builds.adoc#favor_composite_builds,favor `build-logic` composite builds over `buildSrc`>>.
+
+Two workarounds are available.
+
+=== Recommended: register `beforeProject` from a settings convention plugin
```
alllex: ❌ The heading depth seems wrong. It should be within the current section, not define a new one

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -702,6 +704,59 @@ include::sample[dir="snippets/structuring-builds/isolatedProjects-lifecycleApi/k
...
+[NOTE]
+====
+The example above applies the `base` plugin, which ships with Gradle and is therefore always on the project's plugin classpath.
+Applying a convention plugin from a build registered via `pluginManagement.includeBuild(...)` is subject to a current limitation; see <<sec:lifecycle_callbacks_with_included_plugin_builds>>.
+====
+
+[[sec:lifecycle_callbacks_with_included_plugin_builds]]
+===== Applying convention plugins from an included build
+
+Plugins from a build registered via `pluginManagement.includeBuild(...)` are resolved lazily so they are built only when explicitly requested in a `plugins {}` block.
```
alllex, line 716: ❓ Is `pluginManagement` part essential? What about regular included builds added via `includeBuild` calls directly on settings?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -702,6 +704,59 @@ include::sample[dir="snippets/structuring-builds/isolatedProjects-lifecycleApi/k
 include::sample[dir="snippets/structuring-builds/isolatedProjects-lifecycleApi/groovy",files="settings.gradle[tags=lifecycle-before-project]"]
 ====
```
alllex, line 705: 💭 The Kotlin and Groovy samples have diverged. Not sure when this happened, but it would be good to align them, given we are reworking this section.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -702,6 +704,59 @@ include::sample[dir="snippets/structuring-builds/isolatedProjects-lifecycleApi/k
...
+=== Recommended: register `beforeProject` from a settings convention plugin
+
+Move the `gradle.lifecycle.beforeProject` registration into a <<implementing_gradle_plugins_precompiled.adoc#settings_pre_compiled_plugins,settings convention plugin>> published from your included build, then apply that settings plugin from `settings.gradle(.kts)`.
+Inside the settings plugin, the included build's classpath is naturally available, and the project plugin it applies is visible to all projects via the standard plugin-resolution flow.
+
+[source,kotlin]
+.build-logic/src/main/kotlin/my.lifecycle.settings.gradle.kts
+----
+gradle.lifecycle.beforeProject {
+    apply(plugin = "my.convention")
```
alllex: 💅 Given the readers don't see full sample projects, it's not immediately clear where the `my.convention` file is located.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
alllex, line 1: 💭 It's likely that users have also applied 3rd-party plugins in the `allprojects {}` blocks, so they'd need guidance for that case as well. We should check what the current experience is (is it any different from the convention plugins?) and ideally document it as well.

platforms/core-configuration/isolated-action-services/src/main/kotlin/org/gradle/internal/isolate/actions/services/DefaultIsolatedProjectEvaluationListenerProvider.kt
```diff
@@ -186,19 +190,74 @@ class IsolatedProjectEvaluationListener(
...
+private
+tailrec fun findUnknownPluginInCauseChain(t: Throwable?): UnknownPluginException? = when (t) {
+    null -> null
+    is UnknownPluginException -> t
+    else -> findUnknownPluginInCauseChain(t.cause.takeUnless { it === t })
+}
+
+
+private
+val unknownPluginIdRegex = Regex("""Plugin with id '([^']+)' not found\.""")
```
alllex: ❌ We shouldn't parse our own messages. We can represent the ID structurally either by adding information directly to the UnknownPluginException or by subclassing

platforms/core-configuration/isolated-action-services/src/main/kotlin/org/gradle/internal/isolate/actions/services/DefaultIsolatedProjectEvaluationListenerProvider.kt
```diff
@@ -186,19 +190,74 @@ class IsolatedProjectEvaluationListener(
...
+fun rethrowWithLifecyclePluginHintIfApplicable(t: Throwable, gradle: Gradle): Nothing {
+    val toThrow = lifecyclePluginHintFor(t, gradle) ?: t
+    throw toThrow
+}
+
+
+private
+fun lifecyclePluginHintFor(t: Throwable, gradle: Gradle): Throwable? {
+    val gradleInternal = gradle as? GradleInternal ?: return null
+    if (!hasIncludedPluginBuilds(gradleInternal)) return null
```
alllex: 💭 I'd reconsider this check, depending on the experience with the 3rd party plugins

platforms/core-configuration/isolated-action-services/src/main/kotlin/org/gradle/internal/isolate/actions/services/DefaultIsolatedProjectEvaluationListenerProvider.kt
```diff
@@ -186,19 +190,74 @@ class IsolatedProjectEvaluationListener(
...
+
+
+private
+fun extractMissingPluginId(message: String): String? =
+    unknownPluginIdRegex.find(message)?.groupValues?.get(1)
+
+
+private
+fun lifecyclePluginHint(pluginId: String, documentationRegistry: DocumentationRegistry): String = "\n" +
+    "If this plugin is provided by a build registered via `pluginManagement.includeBuild(...)`, " +
```
alllex: ❓ Same question as in the userguide, is `pluginManagement` essential or does regular `includeBuild` have the same problem?

alllex (changes requested):

reinsch82:
@bot-gradle test BD

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
reinsch82: yes, the workaround doesn't work for plain `includedBuild`

platforms/core-configuration/isolated-action-services/src/main/kotlin/org/gradle/internal/isolate/actions/services/DefaultIsolatedProjectEvaluationListenerProvider.kt
reinsch82: see above.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
reinsch82: Can we move this to a follow up PR?

alllex: We can, but then we should continue the investigation and follow-up as part of 9.7

reinsch82:
@bot-gradle test this

platforms/documentation/docs/src/snippets/isolatedProjects/lifecycleApplyFalse/common/gradle.properties
```diff
@@ -0,0 +1 @@
+org.gradle.unsafe.isolated-projects=true
```
alllex: 💅 Drop `unsafe`. There are other places like this

testing/architecture-test/src/test/java/org/gradle/architecture/test/ProviderMigrationArchitectureTest.java
```diff
@@ -93,6 +93,7 @@ public boolean test(JavaClass input) {
         .and(not(declaredIn(Configuration.class)))
         .and(not(declaredIn(FileCollection.class)))
         .and(not(declaredIn(ConfigurableFileCollection.class)))
+        .and(not(declaredIn(assignableTo(Exception.class)))) // exceptions are not migrated to lazy properties
```
alllex, line 96: ❔ What's motivating this change?

platforms/core-configuration/isolated-action-services/src/main/kotlin/org/gradle/internal/isolate/actions/services/DefaultIsolatedProjectEvaluationListenerProvider.kt
```diff
@@ -186,19 +190,77 @@ class IsolatedProjectEvaluationListener(
...
+private
+fun rethrowWithLifecyclePluginHintIfApplicable(t: Throwable, gradle: Gradle): Nothing {
+    val toThrow = lifecyclePluginHintFor(t, gradle) ?: t
+    throw toThrow
+}
+
+
+private
+fun lifecyclePluginHintFor(t: Throwable, gradle: Gradle): Throwable? {
+    val gradleInternal = gradle as? GradleInternal ?: return null
```
alllex: 💅 That's always true, can cast at the very beginning

subprojects/core/src/integTest/groovy/org/gradle/api/invocation/GradleLifecycleConventionPluginApplicationIntegrationTest.groovy
```diff
@@ -0,0 +1,219 @@
...
+
+        when:
+        fails("help")
+
+        then:
+        failureCauseContains("Plugin with id '${CONVENTION_PLUGIN_ID}' not found.")
+        failureCauseContains("If this plugin is provided by a build registered via `pluginManagement.includeBuild(...)`")
+        failureCauseContains("settings convention plugin")
+        failureCauseContains("plugins { id(\"${CONVENTION_PLUGIN_ID}\") apply false }")
+        failureCauseContains("userguide/isolated_projects.html#sec:lifecycle_callbacks_with_included_plugin_builds")
```
alllex: 💅 🤖

```suggestion
        failure.assertHasCause(
            "Plugin with id '${CONVENTION_PLUGIN_ID}' not found.\n" +
            "The plugin must be on the project's plugin classpath before the `gradle.lifecycle` callback runs. " +
            "Declare it in the settings `plugins {}` block with `apply false` " +
            "(e.g. `plugins { id(\"${CONVENTION_PLUGIN_ID}\") apply false }`) so its classpath is exported to all projects. " +
            "For a plugin resolved from a repository, include its version (e.g. `id(\"${CONVENTION_PLUGIN_ID}\") version \"...\" apply false`)." +
            "\nIf this plugin is provided by a build registered via `pluginManagement.includeBuild(...)`, " +
            "you can instead move the `gradle.lifecycle` callback registration into a settings convention plugin " +
            "published from that build (recommended).\n" +
            documentationRegistry.getDocumentationRecommendationFor("information", "isolated_projects", "sec:lifecycle_callbacks_with_included_plugin_builds")
        )
```

`documentationRegistry` is the canonical fixture (it is the same `DocumentationRegistry` the production code uses), so the doc link is exact and version-correct rather than a raw substring. Asserting the full cause is usually better too — it shows the exact message the user sees.

subprojects/core/src/integTest/groovy/org/gradle/api/invocation/GradleLifecycleConventionPluginApplicationIntegrationTest.groovy
```diff
@@ -0,0 +1,219 @@
...
+        """
+
+        when:
+        fails("help")
+
+        then:
+        failureCauseContains("Plugin with id 'com.example.absent' not found.")
+        failureCauseContains("Declare it in the settings `plugins {}` block with `apply false`")
+        failureCauseContains("userguide/isolated_projects.html#sec:lifecycle_callbacks_with_included_plugin_builds")
+        failure.assertThatCause(not(containsString("settings convention plugin")))
```
alllex, line 116: 💅 🤖

```suggestion
        failure.assertHasCause(
            "Plugin with id 'com.example.absent' not found.\n" +
            "The plugin must be on the project's plugin classpath before the `gradle.lifecycle` callback runs. " +
            "Declare it in the settings `plugins {}` block with `apply false` " +
            "(e.g. `plugins { id(\"com.example.absent\") apply false }`) so its classpath is exported to all projects. " +
            "For a plugin resolved from a repository, include its version (e.g. `id(\"com.example.absent\") version \"...\" apply false`).\n" +
            documentationRegistry.getDocumentationRecommendationFor("information", "isolated_projects", "sec:lifecycle_callbacks_with_included_plugin_builds")
        )
```

Same idea: `documentationRegistry` is the canonical fixture for the doc link, and asserting the whole cause demonstrates exactly what the user sees. Matching the full message here also makes the missing included-build advice explicit, so the separate `not(containsString("settings convention plugin"))` check on the next line becomes redundant and can be dropped.

alllex (approved):
LGTM!

testing/architecture-test/src/test/java/org/gradle/architecture/test/ProviderMigrationArchitectureTest.java
reinsch82: [this](https://github.com/gradle/gradle/pull/37759/changes#diff-6d5b1a19390b28fa4a4e168548a005783dc04b2c0c31edf84d6b7ccabeb82ba8R55) fails the architecture test. I could have added another exclusion or fix it for good, since I don't think this is relevant for any exception type.
@asodja Do you agree?

platforms/documentation/docs/src/snippets/isolatedProjects/lifecycleApplyFalse/common/gradle.properties
reinsch82: I see 2 in my change set.

I'll follow up with another PR that cleans up the rest.

- https://github.com/gradle/gradle/issues/38400

testing/architecture-test/src/test/java/org/gradle/architecture/test/ProviderMigrationArchitectureTest.java
asodja: That makes sense, yes

PR #37766, opened 2026-04-30
Change DefaultGradle.toString

octylFractal:
An extra change from #37659 -- this aligns with how we mention builds elsewhere, and removes the awkward unnamed `"build"` when there is no root project yet.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildLookupIntegrationTest.groovy
+++ b/subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildLookupIntegrationTest.groovy
@@ -42,7 +42,7 @@ class CompositeBuildLookupIntegrationTest extends AbstractCompositeBuildValidati
         fails(buildA, "broken")
 
         then:
-        failure.assertHasCause("Included build 'unknown' not found in build 'buildA'.")
+        failure.assertHasCause("Included build 'unknown' not found in build ':'.")
     }
 
     def "dir name is used as build name"() {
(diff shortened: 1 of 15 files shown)


subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildLookupIntegrationTest.groovy
```diff
@@ -108,7 +108,7 @@ class CompositeBuildLookupIntegrationTest extends AbstractCompositeBuildValidati
         fails(buildA, "broken")
 
         then:
-        failure.assertHasCause("Included build 'buildB' not found in build 'buildA'.")
+        failure.assertHasCause("Included build 'buildB' not found in build ':'.")
```
jvandort, line 111: The to sting for the root build seems like a step backward.

In the project display name we put the name of the root project in quotes instead of the path. Can we do something similar here? It seems there might be a name we can use in the BuildDefinition, though not sure if that's normally null for the root

jvandort (approved):
LGTM, even if we can't fix the root build name comment.

One thing that's surprising is changing the toString effects so many messages. We should be using the build's getDisplayName for that.

If it's not too annoying we should consider aligning the builds toString and display name, and then move existing users of toString to the display name

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildLookupIntegrationTest.groovy
octylFractal: I think it usually is `null`, and we use the name of the directory. I don't think that's super helpful tbh, I think the root build should be obvious?

eskatos (approved):

octylFractal:
@bot-gradle test this

Issue #3972, opened 2018-01-05
provide robust credentials/secrets management


marmax:
### Expected Behavior
as a gradle user, I would like to specify secrets (credentials, ssh keys, ...) at possibly one place, with gradle providing variable binding, so I can later use any given secret only by reference.

I should be able to use it **consistently** everywhere I need to., for example:
- in settings.gradle
```
pluginManagement {
    repositories {
        maven {
            url 'https://myrepo.local/nexus/content/repositories/gradle-plugins/'
            credentials(id: 'my-nexus-credentials')
        }
    }
}
```
- in build.gradle
```
repositories {
    maven {
        url 'https://myrepo.local/nexus/content/repositories/gradle-plugins/'
        
        credentials(id: 'my-nexus-credentials', usernameVariable: 'NEXUS_USER', passwordVariable: 'NEXUS_USER_PASSWORD') {
            username = NEXUS_USER
            password = NEXUS_USER_PASSWORD
        }
    }
}

// or:
def apiToken = credentials(id: 'ext-system-token')
task callAPI {
	doLast {
	  def json = new JsonSlurper()
	  return json.parse(addr.toURL().newReader(requestProperties: ["Authorization": "Bearer ${apiToken}".toString(), "Accept": "application/json"]))
	}
}
```

### Current Behavior
currently, a common practice is IMHO to use something like this:
```
credentials {
    username = hasProperty('repoUsername') ? "$repoUsername" : 'anonymous'
    password = hasProperty('repoPassword') ? "$repoPassword" : ''
}
```
using $HOME/.gradle/gradle.properties

### Context
I can also imagine external systems like Hashicorp Vault or MacOSX keychain used for storing the secrets

bmuschko:
Did you try the https://github.com/etiennestuder/gradle-credentials-plugin yet?

marmax:
Hi @bmuschko, 
thank you for the link. I will evaluate it further.

Anyway:
- it's a thirdparty
- it's a plugin and as a such, it cannot be used e.g. in pluginManagement DSL

I would like to see Gradle providing robust secrets management as first class citizen.
& there are IMHO two aspects which should get covered: 

1) **consistent access** to "credentials" (be it even still unencrypted in ~/.gradle/gradle.properties) 
2) pluggable/configurable storage and support for established solutions

Or do you find this to be a nonsense?

daniel-shuy:
> Did you try the https://github.com/etiennestuder/gradle-credentials-plugin yet?

[...]

PR #37767, opened 2026-05-01
Update Documentation: Configuration Cache main page

### LINKED ISSUE #15805: Configuration Cache documentation states build will fail on all violations, but doesn't
The page https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements states
> Each of these requirements is treated as a configuration cache "problem" and fails the build if violations are present.

But this is untrue.
For example the undeclared usage of environment variables and gradle properties does not yet trigger a build failure but just don't work as exepected.
Please update the docs so that it clearly states which violations will / should cause a build failure and for which there is no check implemented yet and whether it is planned to be added.

### LINKED ISSUE #24253: Clarify documentation about configuration cache and parallel builds
### Expected Behavior

It should be very clear that enabling the configuration cache will also enable parallel builds _independently_ of the `org.gradle.parallel` and `--no-parallel` flags.

### Current Behavior

Currently the documentation states:

image

There are issues with this:

1. it is a behavior change. Builds which rely on projects _not_ being executed in parallel, or tasks not being executed in parallel, [will suddenly start failing](https://gradle-community.slack.com/archives/C013WEPGQF9/p1678350560763329).
2. "Executes all tasks in parallel, even those in the same subproject" can be interpreted as "previously, with `--parallel`, only projects were executed in parallel, but now tasks within a single project are executed in parallel".
3. this behavior change is a potential breaking change, so it should be higher in the docs, listed as a behavior change

In addition, I am not convinced that ignoring `--no-parallel` is intuitive. We can argue about `org.gradle.parallel`, because it's "only" a property, but a CLI flag like `--no-parallel` should, IMHO, be honored, or, minimally, should trigger a warning that it is ignored because of the configuration cache.

### Context

I am working on adapting our build plugins for the configuration cache. As part of this work I discovered that tasks were executed in parallel despite neither `--parallel` nor `org.gradle.parallel` being set to `true`. Even after reading the docs, it wasn't obvious if it should have worked with an explicit `org.gradle.parallel=false` or `--no-parallel`.

### LINKED ISSUE #27622: Task with private Ktor client causes Configuration Cache error `Unable to make field private long java.lang.ref.SoftReference.timestamp accessible`
### Current Behavior

Task with private Ktor client causes Configuration Cache error.

### Expected Behavior

Task with private Ktor client should not cause Configuration Cache error.

### Context (optional)

_No response_

### Steps to Reproduce

```kotlin
// build.gradle.kts
import io.ktor.client.*
import io.ktor.client.engine.cio.*
import io.ktor.client.plugins.*
import io.ktor.client.plugins.contentnegotiation.*
import io.ktor.client.request.*
import io.ktor.client.statement.*
import io.ktor.serialization.kotlinx.json.*
import kotlinx.coroutines.runBlocking

buildscript {
  repositories {
    mavenCentral()
    gradlePluginPortal()
  }
  dependencies {
    classpath("io.ktor:ktor-client-core:2.3.7")
    classpath("io.ktor:ktor-client-auth:2.3.7")
    classpath("io.ktor:ktor-client-cio:2.3.7")
    classpath("io.ktor:ktor-client-content-negotiation:2.3.7")
    classpath("io.ktor:ktor-serialization-kotlinx-json:2.3.7")
  }
}

abstract class HttpFetchTask : DefaultTask() {
  @get:Input
  abstract val url: Property<String>

  private val client = HttpClient(CIO) {
    install(HttpRequestRetry) {
      retryOnServerErrors(maxRetries = 5)
      exponentialDelay()
    }
    install(ContentNegotiation) {
      json()
    }
  }

  @TaskAction
  private fun fetch(): Unit = runBlocking {
    val url = url.get()

    val body = client.get(url).bodyAsText()

    println("got $body")
  }
}

val httpFetchTask by tasks.registering(HttpFetchTask::class) {
  url.set("https://gradle.com/")
}
```

Run `gradle httpFetchTask --configuration-cache`

Observe error

```text

* What went wrong:
Configuration cache state could not be cached: field `value` of `kotlin.reflect.jvm.internal.ReflectProperties$LazySoftVal` bean found in field `descriptor$delegate` of `kotlin.reflect.jvm.internal.KClassImpl$Data` bean found in field `value` of `kotlin.InitializedLazyImpl` bean found in field `data` of `kotlin.reflect.jvm.internal.KClassImpl` bean found in field `ignoredTypes` of `io.ktor.client.plugins.contentnegotiation.ContentNegotiation` bean found in field `map` of `io.ktor.util.ConcurrentSafeAttributes` bean found in field `map` of `io.ktor.util.ConcurrentSafeAttributes` bean found in field `attributes` of `io.ktor.client.HttpClient` bean found in field `client` of task `:httpFetchTask` of type `Build_gradle$HttpFetchTask`: error writing value of type 'java.lang.ref.SoftReference'
> Unable to make field private long java.lang.ref.SoftReference.timestamp accessible: module java.base does not "opens java.lang.ref" to unnamed module @440bc22a

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --scan to get full insights.

BUILD FAILED in 729ms
```

### Gradle version

8.5

### Build scan URL (optional)

_No response_

_No response_

### LINKED ISSUE #33620: Should explain expected runtime differences between Configuration cache and vanilla Gradle
### Issue type

Missing information

### Problem description

The configuration cache documentation should describe expected behavior differences between running with Configuration Cache enabled versus vanilla Gradle. Things that are missing:

* dependency resolution happens eagerly during configuration (or before execution), no matter whether dependencies are actually needed (the tasks that need them were skipped). Builds that got away with declaring unused invalid dependencies or failing to configure a repository will fail with CC, when they used to pass without CC.
* there are other kinds of configuration which may be ignored without CC but will be validated (and potentially fail) with CC, and could be documented as well, but dependencies alone seems worth starting this new section. 
 

### Context (optional)

_No response_

### Page with the problem

https://docs.gradle.org/current/userguide/configuration_cache.html

### LINKED ISSUE #36606: Document early dependency resolution when CC is enabled
### Issue type

Missing information

### Problem description

For example, when a plugin applied to project with no sources, its dependencies are not resolved

> The dependencies are (mostly) resolved during serialization, we don't want to carry state needed for dependency resolution to execution phase.
> For WellBehavedPluginTest, I suppose they don't have any sources, which causes early "skip" at execution time. It includes skipping input fingerprinting, which is where dependencies would be resolved in vintage. With CC we want to get cache hits even if someone adds sources later, so we resolve regardless (and the implementation is simpler that way).

### Context (optional)

Related:
* https://github.com/gradle/gradle/pull/36605

### Page with the problem

https://docs.gradle.org/current/userguide/configuration_cache.html or a special page related to CC nuances

lkasso:
### Details

### Context
- Docs for https://github.com/gradle/gradle/issues/15805
- Docs for https://github.com/gradle/gradle/issues/24253
- Docs for https://github.com/gradle/gradle/issues/27622
- Docs for https://github.com/gradle/gradle/issues/33620
- Docs for https://github.com/gradle/gradle/issues/36606
- Docs for https://github.com/gradle/configuration-cache/issues/587
- Updates page to include how CC impacts the Build Lifecycle
   - configuration_cache.adoc
- Updates pages to include the security impact of CC
  - configuration_cache.adoc
  - configuration_cache_status.adoc
  - supported_repository_protocols.adoc

### Documentation Checklist


lkasso:
@bot-gradle test BD

lkasso:
@bot-gradle test BD

lkasso:
@bot-gradle test BD

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -26,6 +26,18 @@ The Configuration Cache improves build performance by caching the result of the
 
 image::configuration-cache-4.png[]
 
+[IMPORTANT]
+====
+This feature is _not enabled by default_ and has the following limitations:
```
tresat: 💭 We've committed to enabled by default in Gradle 10, so maybe we want to hint that this is coming soon?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -45,78 +57,121 @@ To do this, Gradle does the following:
 
 [cols="~,~,~,~,~"]
 |===
-| Phase | Cache | Cache Key | Work Avoidance | Parallelism and Performance
+| Cache | What It Caches | Cache Key | Work Avoidance | Parallelism Win
```
tresat: 💅 "Parallelism Win" sounds perhaps a bit too informal.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -45,78 +57,121 @@ To do this, Gradle does the following:
...
-====
-This feature is _not enabled by default_ and has the following limitations:
+[[config_cache:intro:how_does_it_work]]
+== How It Works
 
-- It does not yet support all <<configuration_cache_status.adoc#config_cache:plugins:core, Core Gradle Plugins>> and <<configuration_cache_status.adoc#config_cache:not_yet_implemented, features>>. Full support is in progress.
-- Your build and its plugins may require modifications to meet the <<configuration_cache_requirements.adoc#config_cache:requirements, Configuration Cache requirements>>.
-- <<config_cache:ide,IDE imports and syncs>> do not yet use the Configuration Cache.
-- Continuous Integration (CI) support is actively evolving.
+When the Configuration Cache is enabled, and you run Gradle for a set of tasks (for example, `./gradlew check`), Gradle's job is to avoid re-running the configuration phase if nothing about the build configuration has changed.
```
tresat:
```suggestion
When the Configuration Cache is enabled, and you run Gradle for a set of tasks (for example, `./gradlew check`) that you have already run once, Gradle will avoid re-configuring the build if no relevant inputs to configuration have changed.
```

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -45,78 +57,121 @@ To do this, Gradle does the following:
...
+| Runs the scheduled tasks against the just-computed task graph.
+| Runs the scheduled tasks against the deserialized task graph. Tasks are isolated from each other through serialization (see <<config_cache:intro:serialization, Serialization>>), which enables intra-project parallel execution.
+|===
+
+image::configuration-cache-6.png[]
+
+=== The Hit-or-Miss Decision
 
-=== First-Time Execution
+At the start of every invocation, Gradle uses the requested tasks (and their command-line arguments) as a *cache key*, then checks whether a stored entry matches that key.
```
tresat: A cache key contains much more than just the requested tasks and arguments.  I think this needs to say something like "the requested tasks, their arguments, and some other environmental information".  Some other things are added to the cache key computation and aren't just left to the fingerprint.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -45,78 +57,121 @@ To do this, Gradle does the following:
...
 
-On subsequent executions of the same tasks (e.g., `./gradlew check` again), Gradle:
+On a cache hit (a subsequent invocation of the same tasks where no configuration inputs have changed), Gradle skips the configuration phase entirely.
+It does *not*:
+
+- Run init scripts.
+- Evaluate the settings script.
+- Configure or rebuild `buildSrc` (unless its inputs changed, which is itself a miss).
+- Apply project plugins.
+- Compute the task graph.
```
tresat: ❗ This is confusing to me.  Is it trying to say that Gradle will act as if init scripts are not present?  That project plugins won't be applied?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -45,78 +57,121 @@ To do this, Gradle does the following:
...
 
-- Skips the configuration phase entirely.
-- Loads the task graph from the Configuration Cache instead.
+Instead, Gradle restores the task graph by deserializing the cache entry, then proceeds straight to execution.
 
-To match a saved entry in the cache, Gradle verifies that no build configuration inputs have changed.
-If any input has changed, Gradle reruns the configuration phase and stores a new entry in the cache.
+[NOTE]
+====
+The lifecycle change has direct consequences for how build logic should be written: some configuration-time code only runs on a cache miss, and some lifecycle callbacks do not fire on a cache hit.
```
tresat: ❗ This makes me nervous.  Which ones?  I can't count on lifecycle callbacks firing?  This makes me not want to enable CC.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+==== What gets serialized
+
+A Configuration Cache entry contains:
+
+- The set of tasks scheduled to run, in dependency order.
+- Each task instance's complete state: every reachable field is serialized, including private fields and fields that are not annotated as `@Input`, `@OutputFile`, or `@Internal`.
+- The graph of objects reachable through those fields (file collections, providers, nested beans, dependency results, build service references).
+
+Anything *not* reachable from a scheduled task is excluded.
+References to types that cannot survive the configuration-to-execution boundary (`Project`, `Gradle`, `ClassLoader`, `Configuration`, and others enumerated in <<configuration_cache_requirements.adoc#config_cache:requirements, the requirements>>) are rejected with a Configuration Cache problem at store time, before they can cause a runtime failure on a later load.
```
tresat: "cannot survive the configuration-to-execution boundary" is weird.  These types "survive" fine when running without CC.

How about something like "some complex types that are intimately connected to the Gradle lifecycle cannot be serialized".

Also, be careful about `Configuration`.  It can _sort of_ be serialized, as a collection of files.  Many examples show this.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+The serialized task graph is stored alongside a separate **fingerprint**: the recorded list of configuration inputs the build observed (see <<config_cache:intro:build_configuration_inputs, Build Configuration Inputs>>).
+The two are distinct: the entry holds the *result* of configuration, the fingerprint holds the *evidence* that the result is still valid.
+Loading an entry on a hit uses both: Gradle replays the fingerprint to confirm validity, then deserializes the entry to reconstruct the task graph.
+
+image::configuration-cache-7.png[]
+
+==== Codecs
+
+Gradle has a dedicated *codec* for each type or family of types it knows how to serialize.
+Three codecs are worth understanding:
```
tresat: I think the `Configuration` codec and serialization process is also vital to understand.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+This isolation is what makes intra-project parallel task execution safe: tasks cannot leak mutable state to each other through a shared reference.
+
+image::configuration-cache-8.png[]
 
-While Configuration Cache serialization doesn't rely on Java Serialization, it understands <<configuration_cache_status.adoc#config_cache:not_yet_implemented:java_serialization, some of its features>>.
-This can be used to customize serialization behavior, but incurs a performance penalty and should be avoided.
+Reference equality *is* preserved for instances of user-defined classes, but *is not* preserved for `String`, `File`, or many `java.util.Collection` types.
+<<configuration_cache_requirements.adoc#config_cache:requirements:shared_objects, Sharing Mutable Objects>> walks through the implications with a runnable example.
+To share state across tasks under the Configuration Cache, use a <<build_services#build_services, Build Service>>.
+Each task is otherwise its own isolated realm.
```
tresat: I think "each task is otherwise its own isolated realm" deserves even more emphasis - perhaps a callout?  Misunderstanding exactly the scenario in the image is a common failure point - then at execution time task a adds an element to the list, but task b doesn't see it.  Maybe adding exactly this example scenario to the callout would help.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+==== Detecting configuration inputs
+
+While the task graph is being serialized, Gradle is also recording the build's <<config_cache:intro:build_configuration_inputs, configuration inputs>>: every environment variable, system property, file read, and file-system check the build performed during configuration.
+To capture those reads without requiring plugins to declare them, Gradle uses a Java agent and rewrites the bytecode of build-script and plugin classes at load time.
+<<configuration_cache_requirements.adoc#config_cache:requirements:bytecode_modifications_and_isolation, Bytecode Modifications and Java Agent>> describes the implications and the Worker API escape hatch for libraries that can't tolerate rewriting.
+
+[[config_cache:intro:behavior_differences]]
+== Behavioral Differences
+
+Some builds behave differently when the Configuration Cache is enabled than when it is disabled.
```
tresat: 💅 Really, all builds.  But for some (well-structured) builds, you just won't care about the observable differences.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+These differences are inherent to how the cache works and are worth knowing about, because a build that previously succeeded without the Configuration Cache may now fail with it enabled.
+
+=== Dependency resolution is eager
+
+Without the Configuration Cache, dependency resolution is lazy: a configuration is resolved only when a task that needs its files actually runs.
+A configuration that declares an invalid dependency, or one that depends on a misconfigured repository, may pass silently if no scheduled task ever touches it.
```
tresat:
```suggestion
A build containing a configuration that cannot be resolved due to an invalid dependency or a misconfigured repository may still succeed if no scheduled task ever resolves that configuration.
```

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+
+With the Configuration Cache enabled, the resolved state of every declared configuration must be captured before execution begins so that it can be serialized into the cache entry.
+Resolution therefore happens during configuration, or at the boundary between configuration and execution, regardless of whether the tasks that need a given configuration are scheduled to run.
```
tresat: Should we say "Resolution happens during serialization, at the end of the configuration phase"?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+
+This includes tasks that would otherwise skip early at execution time.
```
tresat:
```suggestion
This includes tasks that would otherwise be skipped at execution time.
```

I don't like the words "early skip" - earlier than what?  The skip happens when the task would be executed.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+For example, the Java plugin's compile task is skipped when its source set is empty; without the Configuration Cache, the early skip also bypasses input fingerprinting, so the task's compile classpath is never resolved (a configuration with an invalid dependency may go undetected).
```
tresat:
```suggestion
For example, the Java plugin's compile task is skipped when its source set is empty; without the Configuration Cache the task's compile classpath is never resolved (so a configuration with an invalid dependency may go undetected).
```

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+With the Configuration Cache enabled, resolution still happens at store time because the resolved state must be captured into the cache entry.
+A practical benefit of this: cache hits remain stable when sources are added to a previously empty project, since the dependency graph has already been resolved and serialized.
```
tresat: ❓ I'm not sure what this benefit is.  Source being source files added to a project?  Why would that affect configuration?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+
+The practical consequence: a build that contained latent dependency-declaration errors (an invalid coordinate in an unused configuration, a typo'd repository URL, missing credentials for a repository nobody actually fetched from) may pass without the Configuration Cache and fail with it enabled.
```
tresat: This paragraph is just repeating what was already explained.  Could the whole thing be dropped (or made into a short WARNING callout)?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
-- *Parallel Task Execution*: When parallel execution is enabled, even tasks within the same project can be run in parallel, subject to dependency constraints.
-- *Cached Dependency Resolution*: Dependency resolution results are stored and reused.
+Beyond skipping the configuration phase on a cache hit, the Configuration Cache enables two performance optimizations that are not available without it.
 
 [[config_cache:in_action]]
 image::configuration-cache/running-help.gif[]
 
+=== Parallel Task Execution
+
+When the Configuration Cache is enabled, tasks within the same project can run in parallel, subject to their declared input/output dependencies.
```
tresat: 💅 Add emphasis on WITHIN here too

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+=== Cached Dependency Resolution
+
+Dependency resolution is one of the more expensive parts of a build's configuration phase.
+Gradle walks the dependency graph, resolves versions, and computes the resulting classpaths.
+
+When the Configuration Cache is enabled, the *resolved* dependency state becomes part of the cache entry: the resolved configurations, their files, and the artifact metadata are all serialized and replayed on a cache hit.
+On a hit, dependency resolution does not re-run; tasks see the same resolved classpaths that the previous build computed.
+On a miss, resolution runs as part of the configuration phase, and the result is stored alongside the rest of the entry.
+
+Resolution under the Configuration Cache also covers every declared configuration, regardless of whether the tasks needing it run.
```
tresat: ❗ Not every declared configuration - but every resolvable configuration used as a task input.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_status.adoc
```diff
@@ -255,3 +255,14 @@ Check the <<configuration_cache_enabling#config_cache:adoption:changes_in_behavi
 build.
 
 See link:{gradle-issues}33772[gradle/gradle#33772].
+
+[[config_cache:not_yet_implemented:secret_exclusion]]
+=== Selective Exclusion of Sensitive Values from Cache Entries
+
+The Configuration Cache currently has no API to mark a task field or property as "do not serialize."
```
tresat: ❗ You can mark a field as `transient` to exclude it from serialization - "the Configuration Cache considers all but transient fields serializable;"

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_status.adoc
```diff
@@ -255,3 +255,14 @@ Check the <<configuration_cache_enabling#config_cache:adoption:changes_in_behavi
...
+Sensitive values stored in a task's state, such as credentials, tokens, and API keys, are captured in the cache entry alongside the rest of the build configuration.
+
+Gradle encrypts cache entries on disk to mitigate the risk of accidental exposure, but the encryption is per-machine; entries can still be read by anything that has access to both the project's `.gradle/configuration-cache` directory and the encryption key stored in `<<directory_layout.adoc#dir:gradle_user_home,GRADLE_USER_HOME>>`.
+
+TIP: The current workaround is to store secrets in `<<directory_layout.adoc#dir:gradle_user_home,GRADLE_USER_HOME>>/gradle.properties`, whose contents are not included in cache entries; only the file's fingerprint is recorded.
```
tresat: I wouldn't call this a workaround, more like a best practice

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +189,125 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+The full set of rules build logic must follow under the Configuration Cache is enumerated in <<configuration_cache_requirements.adoc#config_cache:requirements, the requirements>>.
 
 [[config_cache:intro:performance_improvements]]
 == Performance Improvements
 
-Beyond skipping the configuration phase, the Configuration Cache enhances performance in the following ways:
-
-- *Parallel Task Execution*: When parallel execution is enabled, even tasks within the same project can be run in parallel, subject to dependency constraints.
-- *Cached Dependency Resolution*: Dependency resolution results are stored and reused.
+Beyond skipping the configuration phase on a cache hit, the Configuration Cache enables two performance optimizations that are not available without it.
```
tresat, line 267: Maybe the Performance Improvements section should come before the Behavioral Differences section - to show the "good" before the "bad".

tresat (changes requested):
Overall I think it is an improvement.  The diagrams are very helpful.  A few things seem confusing or are oversimplified.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
lkasso: removed.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -278,6 +278,8 @@ The Configuration Cache eliminates this contention by giving each task an isolat
 This complements Gradle's existing `--parallel` flag, which executes tasks across *different* projects in parallel.
 With the Configuration Cache, parallelism applies *within* a single project as well.
 
+IMPORTANT: Intra-project parallel execution is always enabled when the Configuration Cache is active. The `--no-parallel` flag and `org.gradle.parallel=false` do not disable it. Builds that depend on tasks within the same project running sequentially may need to declare explicit task dependencies or use <<build_services.adoc#build_services, shared build services>> to coordinate access to shared resources.
```
tresat: Instead of "may need to declare explicit task dependencies" how about "should properly wire task inputs and outputs to enforce task order"

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -146,20 +147,11 @@ This is why we recommend testing your plugin with `--configuration-cache` from t
...
-- Apply project plugins.
-- Compute the task graph.
-
 Instead, Gradle restores the task graph by deserializing the cache entry, then proceeds straight to execution.
 
-[NOTE]
-====
-The lifecycle change has direct consequences for how build logic should be written: some configuration-time code only runs on a cache miss, and some lifecycle callbacks do not fire on a cache hit.
-====
+Because the configuration phase is skipped on a hit, any logic registered through configuration-phase callbacks, such as `BuildListener`, `Gradle.beforeProject`, `Gradle.afterProject`, or `Gradle.taskGraph.whenReady`, will not execute.
```
tresat: will not **re-execute**

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -220,15 +213,23 @@ image::configuration-cache-7.png[]
 ==== Codecs
 
 Gradle has a dedicated *codec* for each type or family of types it knows how to serialize.
-Three codecs are worth understanding:
+Four codecs are worth understanding:
 
 - The *task node codec* writes and reads the units of the task graph itself.
-- The *bean codec* is the fallback for arbitrary user-defined classes; it walks reflective fields and recursively serializes each one. Most plugin-defined task and extension types end up handled by the bean codec without any extra work from the author.
-- The *provider codec* handles `Provider` and `Property` instances. The provider codec **freezes lazy values**: the value a provider would compute is captured at store time and replayed at load time, so providers do not re-evaluate their original sources during execution. This is why properties wired through `Provider` and `Property` survive the cache round-trip cleanly even when the original source (e.g., a file or environment variable) is no longer accessible at execution time.
+- The *bean codec* is the fallback for arbitrary user-defined classes; it walks reflective fields and recursively serializes each one.
```
tresat: it walks reflective fields -> it walks fields using reflection

tresat (approved):
I think you should still seek a CC expert for a final review, but everything LGTM now.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -45,78 +58,112 @@ To do this, Gradle does the following:
...
+| Task outputs and artifact transform outputs
+| Each task's declared inputs (sources, classpath, action implementation)
+| Skips task execution by reusing previous outputs
+| Composes with `--parallel` for cross-project parallel execution
+
+| *Configuration Cache*
+| The result of the configuration phase: the task graph, each task's configured state, and dependency information
+| Build logic and environment (build scripts, init scripts, environment variables and properties read at configuration time, and more)
+| Skips the configuration phase entirely on a cache hit
+| Enables intra-project parallel task execution
```
mlopatkin: 💅Maybe "Enable parallel task execution within the same project", to make it easier for those beginning their English journey?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -45,78 +58,112 @@ To do this, Gradle does the following:
...
+| *Execution*
+| Runs the scheduled tasks against the just-computed task graph.
+| Runs the scheduled tasks against the deserialized task graph. Tasks are isolated from each other through serialization (see <<config_cache:intro:serialization, Serialization>>), which enables intra-project parallel execution.
+|===
+
+image::configuration-cache-6.png[]
+
+=== The Hit-or-Miss Decision
+
+At the start of every invocation, Gradle uses the requested tasks, their arguments, and some other environmental information as a *cache key*, then checks whether a stored entry matches that key.
```
mlopatkin: :thinking: The way I read "a stored entry matches that key" kind of implies that there is only one entry, though maybe `a` should have told me that there are many. I'd put it 
"check whether there is a stored entry matching the key" to emphasize there is a lookup going on.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -45,78 +58,112 @@ To do this, Gradle does the following:
...
+If a candidate entry exists, Gradle compares the build's current <<config_cache:intro:build_configuration_inputs, configuration inputs>> against the entry's recorded fingerprint.
```
mlopatkin: :thinking: Maybe clarify what "fingerprint" is? What Gradle actually does, it checks if the build inputs discovered when preparing the entry still have the same values.

UPD: I see it is covered later in the doc, but maybe it is still worth explaining here.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -45,78 +58,112 @@ To do this, Gradle does the following:
...
 4. Run build scripts, applying any requested project plugins. If plugins come from included builds, Gradle builds them first.
 5. Calculate the task graph, executing deferred configuration actions.
 
-Gradle then stores a snapshot of the task graph in the Configuration Cache for future use.
-After this, Gradle loads the task graph from the cache and proceeds with task execution.
+Gradle then **serializes the resulting task graph and writes it to the cache** before execution begins.
 
-NOTE: When it is enabled, Gradle always loads from the configuration cache - even on a cache miss - to ensure that hits and misses are handled consistently.
+Even on a miss, Gradle then **loads the task graph back from the cache it just wrote** and executes against the loaded copy.
+This always-load behavior is intentional: it guarantees that hits and misses execute through the same code path, so any serialization-related issue surfaces on the first run rather than later.
```
mlopatkin, line 136: :thinking: it is worth noting that it also allows parallel task execution.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -125,7 +172,7 @@ The following elements determine whether a Configuration Cache entry is valid:
 ** `GRADLE_USER_HOME`
 ** Gradle Daemon JVM
 2. Init scripts
-3. `buildSrc` and included build logic build contents (build scripts, sources, and intermediate build outputs)
+3. `buildSrc` and included build logic, build contents (build scripts, sources, and intermediate build outputs)
```
mlopatkin: :x: this should be parsed as "contents of buildSrc and included build logic builds", so the comma is wrong. Our own "buffalo buffalo buffalo".

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -45,78 +58,112 @@ To do this, Gradle does the following:
...
+
+=== Configuration Cache Entries
 
-To match a saved entry in the cache, Gradle verifies that no build configuration inputs have changed.
-If any input has changed, Gradle reruns the configuration phase and stores a new entry in the cache.
+A Configuration Cache entry stores:
 
+- The set of tasks to run
+- Their configuration details
+- Dependency information
```
mlopatkin: ❔Does this mean "task dependency information"?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+At load time, the task sees the resolved files, not the original `Configuration` object.
+Task authors should still <<configuration_cache_requirements.adoc#config_cache:requirements:disallowed_types, prefer `FileCollection`>> over `Configuration` in their task inputs.
+
+Built-in codecs also exist for collections, files, file collections, dependency results, build services, lambdas, and many Gradle-specific types.
+When none of these match, the bean codec is the fallback.
+Types that can't be safely serialized (`ClassLoader`, `Thread`, and others) are rejected at store time rather than producing a failure on later load.
+
+Gradle does not use Java Serialization, but it understands <<configuration_cache_status.adoc#config_cache:not_yet_implemented:java_serialization, a subset of its features>> for classes that opt in via `Externalizable` or `writeObject` / `readObject`.
+Using those features is supported but incurs a performance penalty and should be avoided in new code.
```
mlopatkin: :x: Codecs are an implementation detail. We shouldn't expose them to users. From what I see, we're trying to describe transformations/simplifications that CC applies to the data. Let's phrase the docs in these terms instead.

I've added some comments about each codec on what to salvage and what can be dropped.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+On a hit, dependency resolution does not re-run; tasks see the same resolved classpaths that the previous build computed.
+On a miss, resolution runs as part of the configuration phase, and the result is stored alongside the rest of the entry.
+
+Resolution under the Configuration Cache covers every resolvable configuration used as a task input, regardless of whether the tasks needing it run.
+See <<config_cache:intro:behavior_differences, Behavioral Differences>> for the implications.
+
+[[config_cache:intro:behavior_differences]]
+== Behavioral Differences
+
+Enabling the Configuration Cache changes how every build executes — task isolation, eager dependency resolution, and the skipped configuration phase all apply universally.
```
mlopatkin: ❓  Not sure if I understand what "all apply universally" tries to convey.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+== Security Considerations
+
+The Configuration Cache serializes the configured state of every scheduled task into a cache entry stored under `.gradle/configuration-cache` in the project directory.
+This includes any values reachable from task fields — which may contain credentials, tokens, API keys, or other sensitive data if they are held in task state.
+
+Gradle encrypts Configuration Cache entries on disk using a machine-specific key stored in `<<directory_layout.adoc#dir:gradle_user_home,GRADLE_USER_HOME>>`.
+However, anything with access to both the cache directory and the encryption key can read the entries.
+
+Additionally, system properties and environment variables referenced during the configuration phase are recorded in the cache entry's fingerprint.
+If credentials are passed through these mechanisms, their values become part of the stored fingerprint data.
```
mlopatkin: ❓ The order somewhat implies fingerprint data is not encrypted, but it actually is.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+
+To reduce the risk of exposing sensitive values through the Configuration Cache:
```
mlopatkin, line 338: 🤔 I'd add "use lazy wiring, like `providers.environmentVariable`, to feed sensitive data to tasks and only obtain the values at execution time".

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -19,7 +19,10 @@
 
 [[config_cache:requirements]]
 To capture and reload the task graph state using the Configuration Cache, Gradle enforces specific requirements on tasks and build logic.
-Any violation of these requirements is reported as a Configuration Cache "problem," which causes the build to fail.
+
+Many violations of these requirements are detected and reported as Configuration Cache "problems," which cause the build to fail.
+However, not all violations can be detected automatically.
+Some (such as reading an environment variable or system property through a code path that bypasses Gradle's bytecode instrumentation), are not reported, but will produce incorrect results on a cache hit because the value is not tracked as a configuration input.
```
mlopatkin: 🤔We no longer consider reading environment vars or system properties a violation, so I'd use "starting an external process at configuration time" as an example here.

TBH, maybe it's time to split "input tracking" part out of enforced requirements. Its current location is 7.x legacy where calling System.getenv was an actual CC problem. But that's out of scope for this PR.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -554,21 +557,27 @@ In simple cases, when the libraries also provide command-line entry points (`pub
...
+See <<configuration_cache_status.adoc#config_cache:not_yet_implemented:secret_exclusion, Selective Exclusion of Sensitive Values>> for the current status.
 
 To mitigate the risk of accidental exposure, Gradle encrypts the Configuration Cache.
 When required, Gradle transparently generates a machine-specific secret key, caches it under the
 `<<directory_layout.adoc#dir:gradle_user_home,GRADLE_USER_HOME>>` directory, and uses it to encrypt data in the project-specific caches.
 
 To further enhance security, follow these recommendations:
 
-- Secure access to Configuration Cache entries.
-- Use `<<directory_layout.adoc#dir:gradle_user_home,GRADLE_USER_HOME>>/gradle.properties` to store secrets.
```
mlopatkin, line 701: 🤔Lazy wiring mentioned [there](https://github.com/gradle/gradle/pull/37767/changes#r3355522324) applies here as well.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -554,21 +557,27 @@ In simple cases, when the libraries also provide command-line entry points (`pub
...
+- **Store credentials in `<<directory_layout.adoc#dir:gradle_user_home,GRADLE_USER_HOME>>/gradle.properties`.**
+The content of this file is **not** included in the Configuration Cache — only its fingerprint is.
 If storing secrets in this file, ensure access is **properly restricted**.
+- **Use the <<supported_repository_protocols.adoc#sec:handling-credentials, Gradle credentials API>>** for repository authentication, so that credential values are looked up from Gradle properties rather than hardcoded in build scripts.
+- **Mark fields holding runtime secrets as `transient`** if the value is not needed after deserialization.
+- **Restrict access** to the `.gradle/configuration-cache` directory and the encryption key in `<<directory_layout.adoc#dir:gradle_user_home,GRADLE_USER_HOME>>`.
 
-See link:{gradle-issues}22618[gradle/gradle#22618].
+For an overview, see <<configuration_cache.adoc#config_cache:security, Security Considerations>>.
+See also link:{gradle-issues}22618[gradle/gradle#22618].
```
mlopatkin: :nail_care: Let's point to the top-level issue: https://github.com/gradle/gradle/issues/3972

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_status.adoc
```diff
@@ -255,3 +255,16 @@ Check the <<configuration_cache_enabling#config_cache:adoption:changes_in_behavi
...
+=== Selective Exclusion of Sensitive Values from Cache Entries
+
+While Java's `transient` keyword can be used to exclude individual fields from Configuration Cache serialization, there is no Gradle-level API to selectively mark a property as "do not serialize."
+Sensitive values stored in non-transient task state, such as credentials, tokens, and API keys, are captured in the cache entry alongside the rest of the build configuration.
+
+Gradle encrypts cache entries on disk to mitigate the risk of accidental exposure, but the encryption is per-machine; entries can still be read by anything that has access to both the project's `.gradle/configuration-cache` directory and the encryption key stored in `<<directory_layout.adoc#dir:gradle_user_home,GRADLE_USER_HOME>>`.
+
+TIP: The recommended practice is to store secrets in `<<directory_layout.adoc#dir:gradle_user_home,GRADLE_USER_HOME>>/gradle.properties`, whose contents are not included in cache entries; only the file's fingerprint is recorded.
+
+See <<configuration_cache.adoc#config_cache:security, Security Considerations>> for additional guidance.
```
mlopatkin, line 262: :nail_care: Let's point to the top-level issue here too: https://github.com/gradle/gradle/issues/3972

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+Without the Configuration Cache, intra-project parallelism is constrained because tasks within a project share access to the `Project` instance and contend on project-level locks.
+The Configuration Cache eliminates this contention by giving each task an isolated, deserialized copy of its state (see <<config_cache:intro:serialization, Serialization>>), so tasks no longer need to acquire project-level locks at execution time.
+
+This complements Gradle's existing `--parallel` flag, which executes tasks across *different* projects in parallel.
+With the Configuration Cache, parallelism applies *within* a single project as well.
+
+IMPORTANT: Intra-project parallel execution is always enabled when the Configuration Cache is active. The `--no-parallel` flag and `org.gradle.parallel=false` do not disable it. Builds that depend on tasks within the same project running sequentially should properly wire task inputs and outputs to enforce task order or use <<build_services.adoc#build_services, shared build services>> to coordinate access to shared resources.
```
mlopatkin: :thinking: maybe also point readers to https://docs.gradle.org/current/userguide/build_services.html#accessing_the_build_service_concurrently directly?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+
+Without the Configuration Cache, dependency resolution is lazy: a configuration is resolved only when a task that needs its files actually runs.
+A build containing a configuration that cannot be resolved due to an invalid dependency or a misconfigured repository may still succeed if no scheduled task ever resolves that configuration.
+
+With the Configuration Cache enabled, the resolved state of every declared configuration must be captured before execution begins so that it can be serialized into the cache entry.
+Resolution happens during serialization, at the end of the configuration phase, regardless of whether the tasks that need a given configuration are scheduled to run.
+
+This includes tasks that would otherwise be skipped at execution time.
+For example, the Java plugin's compile task is skipped when its source set is empty; without the Configuration Cache the task's compile classpath is never resolved (so a configuration with an invalid dependency may go undetected).
+With the Configuration Cache enabled, resolution still happens at store time because the resolved state must be captured into the cache entry.
```
mlopatkin, line 312: :thinking: I think we should remind the reader that the general contract of "you should resolve stuff at execution time only" still applies. Something along the lines:
>While dependency resolution by Configuration Cache happens at what is considered configuration time, this doesn't mean that the build logic should no longer obey the general rule of only resolving configurations at execution time. Gradle can apply different optimizations to dependency resolution performed by the Configuration Cache that are not possible for general build logic, so a well-formed build is still likely to perform better.

This can be a good place to mention parallel store as one of the potential optimizations.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -45,78 +58,112 @@ To do this, Gradle does the following:
...
-If a matching entry is found, Gradle reuses it and skips the configuration phase entirely.
-A cache hit is essential for avoiding configuration work.
+[cols="~,~,~"]
+|===
+| Phase | Without the Configuration Cache | With the Configuration Cache
 
-=== Configuration Cache Entries
+| *Initialization*
+| Runs every build. Detects the settings file, creates the `Settings` and `Project` instances, and determines the structure of the build.
+| Runs every build.
```
mlopatkin: :nail_care: Initialization with CC is very different. Maybe we can elaborate a bit here - 
>Runs every build. Locates the cache entry, verifies if it is up-to-date and deserializes the task graph, or falls back to full initialization if there is no valid entry.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+Loading an entry on a hit uses both: Gradle replays the fingerprint to confirm validity, then deserializes the entry to reconstruct the task graph.
+
+image::configuration-cache-7.png[]
+
+==== Codecs
+
+Gradle has a dedicated *codec* for each type or family of types it knows how to serialize.
+Four codecs are worth understanding:
+
+- The *task node codec* writes and reads the units of the task graph itself.
```
mlopatkin: :x: Seems irrelevant (Nodes are an internal concept). We can say that we serialize all scheduled task instances.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+- The *bean codec* is the fallback for arbitrary user-defined classes; it walks fields using reflection and recursively serializes each one.
```
mlopatkin: :x: This is all valid and important, but I'd rather not refer to it as "bean". 

Maybe "the default serialization" and then "some types are subjects to more specific handling".

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+Most plugin-defined task and extension types end up handled by the bean codec without any extra work from the author.
+- The *provider codec* handles `Provider` and `Property` instances.
```
mlopatkin: :x: Providers (and Properties) are handled specially by the configuration cache. CC attempts to pre-compute and simplify provider chains it discovers while walking the graph. Then we can point readers to the Provider javadocs where the idea is explained in more detail.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+The provider codec **freezes lazy values**: the value a provider would compute is captured at store time and replayed at load time, so providers do not re-evaluate their original sources during execution.
```
mlopatkin: :x: This is not always true, and may give wrong ideas to users. We should be clear that only some providers are subject to this treatment.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+This is why properties wired through `Provider` and `Property` survive the cache round-trip cleanly even when the original source (e.g., a file or environment variable) is no longer accessible at execution time.
```
mlopatkin: :x: I don't understand this. As written, it is completely wrong: if something is gone by execution time, then CC entry is invalidated.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+- The *configuration codec* handles `Configuration` instances by resolving them at store time and serializing the result as a file collection.
+This is why dependency resolution is <<config_cache:intro:behavior_differences, eager under the Configuration Cache>>: every declared configuration must be resolved at store time so its files can be captured into the cache entry.
+At load time, the task sees the resolved files, not the original `Configuration` object.
```
mlopatkin: :thinking: One important consequence is that the restored object is a `FileCollection` but not `Configuration`, so authors who only need files (for classpath/other processing) should declare their tasks' field as `FileCollection`s.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+Task authors should still <<configuration_cache_requirements.adoc#config_cache:requirements:disallowed_types, prefer `FileCollection`>> over `Configuration` in their task inputs.
+
+Built-in codecs also exist for collections, files, file collections, dependency results, build services, lambdas, and many Gradle-specific types.
```
mlopatkin: :thinking: this is good, but instead of talking codecs we can just say that Gradle handles many standard Java and Gradle-specific types in a customized fashion - and add a link to to the requirement page - `config_cache:requirements:shared_objects`.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+If two tasks shared the same `ArrayList` at configuration time, they each get an independent copy at execution time.
+When task A adds an element to its list during execution, task B will *not* see the change — its copy is a separate object.
+
+This is a common source of subtle bugs: code that worked without the Configuration Cache (where both tasks referenced the same in-memory list) silently breaks when the cache isolates them.
+To share mutable state across tasks, use a <<build_services#build_services, Build Service>>.
+====
+
+==== Detecting configuration inputs
+
+While the task graph is being serialized, Gradle is also recording the build's <<config_cache:intro:build_configuration_inputs, configuration inputs>>: every environment variable, system property, file read, and file-system check the build performed during configuration.
```
mlopatkin: :x: We record the inputs throughout the whole configuration phase, not only during serialization.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -134,25 +181,171 @@ The following elements determine whether a Configuration Cache entry is valid:
...
+Serialization is the mechanism that makes the Configuration Cache possible.
+When Gradle stores a cache entry, it walks the configured task graph and writes a binary snapshot of every task and the state reachable from it.
+When Gradle loads a cache entry, it reverses the process to reconstruct the task graph for execution.
+
+==== What gets serialized
+
+A Configuration Cache entry contains:
+
+- The set of tasks scheduled to run, in dependency order.
+- Each task instance's complete state: every reachable field is serialized, including private fields and fields that are not annotated as `@Input`, `@OutputFile`, or `@Internal`.
```
mlopatkin: :nail_care: maybe drop "reachable"? An "instance field" (as opposed to a static one) would be more relevant.

mlopatkin (changes requested):
Thank you for taking this on! 🎉 
This is a monumental rewrite, and it addresses so many missing bits folks were stumbling upon for years.

I really like the structure, now it is much easier to evolve. I left quite a number of comments, but these are mostly about particular details.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
lkasso: Incorrect codec information was removed entirely.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
lkasso: Codec information was removed entirely.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
lkasso: Codec information was removed entirely.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
lkasso: Incorrect codec information was removed entirely.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
lkasso: Codec information was removed entirely.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache.adoc
```diff
@@ -324,11 +328,12 @@ The full set of rules build logic must follow under the Configuration Cache is e
...
 
-Gradle encrypts Configuration Cache entries on disk using a machine-specific key stored in `<<directory_layout.adoc#dir:gradle_user_home,GRADLE_USER_HOME>>`.
+Gradle encrypts all Configuration Cache data on disk, both the serialized task graph and the fingerprint, using a machine-specific key stored in `<<directory_layout.adoc#dir:gradle_user_home,GRADLE_USER_HOME>>`.
 However, anything with access to both the cache directory and the encryption key can read the entries.
 
-Additionally, system properties and environment variables referenced during the configuration phase are recorded in the cache entry's fingerprint.
-If credentials are passed through these mechanisms, their values become part of the stored fingerprint data.
+System properties and environment variables referenced during the configuration phase are recorded in the fingerprint.
+If credentials are passed through these mechanisms, their values become part of the stored data.
+Use lazy wiring, like `providers.environmentVariable`, to feed sensitive data to tasks and only obtain the values at execution time.
```
mlopatkin: 💅 Maybe add a link to [ProviderFactory](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/ProviderFactory.html#environmentVariable(java.lang.String))?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -573,11 +707,12 @@ To further enhance security, follow these recommendations:
 The content of this file is **not** included in the Configuration Cache — only its fingerprint is.
 If storing secrets in this file, ensure access is **properly restricted**.
 - **Use the <<supported_repository_protocols.adoc#sec:handling-credentials, Gradle credentials API>>** for repository authentication, so that credential values are looked up from Gradle properties rather than hardcoded in build scripts.
+- **Use lazy wiring** like `providers.environmentVariable` to feed sensitive data to tasks and only obtain the values at execution time, keeping them out of the serialized task graph.
```
mlopatkin: 💅 Maybe add a link to [ProviderFactory](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/ProviderFactory.html#environmentVariable(java.lang.String)) too?

mlopatkin (approved):
Awesome! Let's ship it!

You may want to add links before that, but I don't insist.

PR #37771, opened 2026-05-01
Update embedded Groovy from 4.0.29 to 4.0.31

big-guy:
## Summary

Bump the embedded Groovy version from 4.0.29 to [4.0.31](https://groovy-lang.org/changelogs/changelog-4.0.31.html) (also picks up [4.0.30](https://groovy-lang.org/changelogs/changelog-4.0.30.html)).

Split out from #37770 so the Groovy upgrade and the two fixes it requires land in their own PR.

## Test-fixture compile fix

Groovy 4.0.31 newly enforces interface/abstract-base visibility consistency, surfacing a latent issue: `AbstractHttpResource.getPath()` was declared `protected` while `MetaDataArtifact` implements `RootMavenMetaData` (whose `getPath()` is public via `ModuleArtifact`). Drop the `protected` modifier on the abstract base and the matching `BasicHttpResource` override so the visibility lines up with the interface contract.

## BeanDynamicObject runtime fix

Groovy 4.0.30 removed the private `MetaClassImpl.getMetaProperty(String, boolean)` helper that `BeanDynamicObject` reflectively invokes as a faster alternative to the public `getMetaProperty(String)`. Without a fix the class fails to initialize at runtime, taking down anything that goes through Gradle's dynamic-object path (e.g. `:build-cache-core:test`'s `DirectoryBuildCacheEntryRetentionTest`). Switch to the underlying private `getMetaProperty(Class, String, boolean, boolean)` helper, which the removed 2-arg variant delegated to and which exists in both 4.0.29 and 4.0.31 — preserves the fast path without a public-API fallback.

## Test plan

- [x] `./gradlew sanityCheck`
- [x] `./gradlew :build-cache-core:test --tests DirectoryBuildCacheEntryRetentionTest`
- [ ] CI: cross-version Groovy tests (`GroovyCoverage.PREVIOUS` retains 4.0.29 as a previous-version data point — no change needed)


big-guy:
@bot-gradle merge

Issue #13506, opened 2020-06-18
Configuration cache should support source dependencies


eskatos:
The configuration cache doesn't support source dependencies yet. This issue is about adding that support.

### Expected Behavior

The configuration cache supports source dependencies.

### Current Behavior

See https://docs.gradle.org/nightly/userguide/configuration_cache.html#not_yet_implemented

Vampire:
Does that mean, source dependencies will not get supported with CC, as this issue now got closed as completed with auto-disabling CC as far as I understood?

bamboo:
Hey @Vampire, source dependencies will be supported with CC when the feature is promoted out of incubation.

abstratt:
Update for 9.0: use of source dependencies now no longer causes a build failure with CC, but instead causes CC to be disabled. See: https://github.com/gradle/gradle/pull/33873

PR #37773, opened 2026-05-02
Support source dependencies with configuration cache

big-guy:
## Summary

- Lift the source-dependency degradation gate so `--configuration-cache` works with builds that use `sourceControl.vcsMappings`
- Round-trip the implicit-vs-explicit included-build distinction through CC store/load via a new `BuildStateRegistry.addImplicitIncludedBuild(BuildDefinition, Path)` overload and one extra boolean per child build in `ConfigurationCacheState`
- Wire dynamic source-dep selectors (branch refs, `latest.integration`, version ranges) into the existing `ChangingValueDependencyResolutionListener` path so CC invalidates entries whose resolved commit can move
- Add `SourceDependencyConfigurationCacheIntegrationTest` covering store, real cache hit on reuse, settings-script invalidation, dynamic-selector invalidation, and upstream-commit invalidation

Refs: #13506

## What changed

| Area | Change |
|---|---|
| `BuildStateRegistry` (`subprojects/core`) | New `addImplicitIncludedBuild(BuildDefinition, Path)` overload — same as the existing no-arg variant but accepts a known build path for CC reload |
| `DefaultIncludedBuildRegistry` (`subprojects/composite-builds`) | Implements the overload; the existing zero-path method delegates |
| `ConfigurationCacheBuild` + `DefaultConfigurationCacheHost` | New `addImplicitIncludedBuild(...)` method on the CC build interface; routes through the registry overload |
| `ConfigurationCacheState.kt` | `writeIncludedBuild` writes `state.isImplicitBuild`; `readIncludedBuild` routes to either implicit or explicit registration. Removes the source-dep `writeChildBuilds` log-not-implemented branch and now-unused imports |
| `DefaultConfigurationCacheDegradationController.kt` | Removes `isSourceDependenciesUsed()` from feature degradation, plus the `VcsMappingsStore` constructor parameter |
| `DefaultVcsVersionWorkingDirResolver` (`platforms/software/version-control`) | Detects dynamic constraints (branch, `latest.integration`, range, etc.) and fires `ChangingValueDependencyResolutionListener.onDynamicVersionSelection` with a zero-duration `Expiry`, so CC always invalidates entries that resolved a moving ref. Mirrors the external-dynamic-version path; a future `cacheDynamicVersionsFor`-equivalent for source deps can swap the constant for a real `Duration` without changing the writer or checker |
| `VersionControlServices` | Injects `ChangingValueDependencyResolutionListener` into the resolver |
| `ConfigurationCacheGracefulDegradationIntegrationTest` | The two lazy-property-in-degraded-mode tests no longer use source deps as a degradation trigger; they now use the task-level `requireConfigurationCacheDegradation` API. The source-dep-specific feature-degradation test and the `enableSourceDependencies()` helper are removed |
| `ConfigurationCacheCompositeBuildsIntegrationTest` | Removed the assertion that source deps gracefully degrade — no longer holds |

## Testing

The new `SourceDependencyConfigurationCacheIntegrationTest` covers:
1. Store CC entry for a source-dep build
2. Reuse the entry on the second invocation with a static version (`assertStateLoaded()`)
3. Invalidate when the `vcsMappings` block changes (settings-script fingerprint already covers this)
4. Always invalidate when the source-dep selector is dynamic (`latest.integration`)
5. Invalidate when the upstream commit changes for a dynamic selector

`GitVersionSelectionIntegrationTest`: 3 dynamic-selector tests (`latest.integration`, sub-version, branch HEAD) had `@ToBeFixedForConfigurationCache(#13506)` removed and now pass under CC. 4 cases remain `Skip.FLAKY` for the unrelated per-invocation HTTP-expectation pattern.

## Known limitations / follow-ups

- **Annotated test rewrites.** ~25 source-dep tests across multiproject/nested/identity/cleanup/etc. suites are marked `@ToBeFixedForConfigurationCache`. They fail under CC for orthogonal reasons (per-invocation HTTP fixture expectations, ext properties read at execution time, build-script anti-patterns, working-dir reset relying on per-build resolution, build-operation hierarchy differences). Each annotation cites the specific reason. Properly rewriting them as CC-safe is follow-up work.
- **Working-dir reset on cache hit.** A handful of `GitVcsIntegrationTest` cases verify that local edits inside a source-dep checkout get reset on the next build. Under CC this is skipped on a cache hit because resolution doesn't run. Real correctness gap; tracked in the annotated tests and worth a separate design discussion.
- **Dead enum entry.** `NotYetImplementedSourceDependencies` in `PropertyProblem.kt` is no longer referenced; removing it requires also pruning the doc anchor.

## Test plan

- [x] `./gradlew sanityCheck`
- [x] `./gradlew :version-control:configCacheIntegTest`
- [x] `./gradlew :version-control:forkingIntegTest --tests '*SourceDependencyConfigurationCacheIntegrationTest'`
- [x] `./gradlew :version-control:forkingIntegTest --tests '*GitVersionSelectionIntegrationTest'`
- [x] `./gradlew :configuration-cache:forkingIntegTest --tests '*ConfigurationCacheGracefulDegradationIntegrationTest'`
- [ ] CI cross-version coverage

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGracefulDegradationIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGracefulDegradationIntegrationTest.groovy
@@ -51,11 +51,9 @@ class ConfigurationCacheGracefulDegradationIntegrationTest extends AbstractConfi
         def configurationCache = newConfigurationCacheFixture()
 
         given:
-        // any way to trigger graceful degradation will do
-        enableSourceDependencies()
-
         buildFile """
-        class TaskWithLazyProperty extends DefaultTask {
+        ${taskWithInjectedDegradationController()}
+        abstract class TaskWithLazyProperty extends DegradingTask {
             private String value
             @Input
             String getLazyValue() {
@@ -66,6 +64,7 @@ class ConfigurationCacheGracefulDegradationIntegrationTest extends AbstractConfi
             }
         }
         tasks.register("lazy", TaskWithLazyProperty) { task ->
+           getDegradationController().requireConfigurationCacheDegradation(task, provider { "Project access" })
            doLast {
                println("Value is " + task.lazyValue)
            }
@@ -84,17 +83,21 @@ class ConfigurationCacheGracefulDegradationIntegrationTest extends AbstractConfi
         def configurationCache = newConfigurationCacheFixture()
 
         given:
-        // any way to trigger graceful degradation will do
-        enableSourceDependencies()
-
         buildKotlinFile """
+        import javax.inject.Inject
+        import org.gradle.api.internal.ConfigurationCacheDegradationController
+
         abstract class TaskWithLazyProperty: DefaultTask() {
+            @get:Inject
+            abstract val degradationController: ConfigurationCacheDegradationController
+
             @get:Input
             val lazyValue: String by lazy {
                 this.project.name
             }
         }
         tasks.register("lazy", TaskWithLazyProperty::class) {
+            degradationController.requireConfigurationCacheDegradation(this, provider { "Project access" })
             doLast {
                 println("Value is " + lazyValue)
             }
@@ -189,27 +192,6 @@ class ConfigurationCacheGracefulDegradationIntegrationTest extends AbstractConfi
         args                                                          | expectedOutputs                                               | degradationReason
         ["-DaccessTaskProject=true"]                                  | ["Task's project accessed!"]                                  | "Project access."
         ["-DaccessTaskProject=true", "-DaccessTaskDependencies=true"] | ["Task's project accessed!", "Task's dependencies accessed!"] | "Project access, TaskDependencies access."
-    }
-
-    def "features may cause CC degradation"() {
-        enableSourceDependencies()
-
-        when:
-        configurationCacheRun("help")
-
-        then:
-        configurationCache.assertNoConfigurationCache()
-
-        and:
-        // feature degradation is reported as a report problem, but it is silently suppressed from the console
-        problems.assertResultConsoleSummaryHasNoProblems(result)
-        problems.htmlReport(result.output).assertContents {
-            totalProblemsCount = 1
-            withProblem("Feature 'source dependencies' is incompatible with the configuration cache.")
-        }
-
-        and:
-        assertConfigurationCacheDegradation()
     }
 
     def "CC problems in warning mode are not hidden by CC degradation"() {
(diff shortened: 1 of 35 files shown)


platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGracefulDegradationIntegrationTest.groovy
```diff
@@ -190,27 +193,6 @@ class ConfigurationCacheGracefulDegradationIntegrationTest extends AbstractConfi
         ["-DaccessTaskProject=true", "-DaccessTaskDependencies=true"] | ["Task's project accessed!", "Task's dependencies accessed!"] | "Project access, TaskDependencies access."
     }
 
-    def "features may cause CC degradation"() {
```
ghale, line 194: ❓My understanding is that this test is removed because source dependencies is the last "feature" that could create a degradation.  This means [collectFeatureDegradationReasons()`[https://github.com/gradle/gradle/pull/37773/changes#diff-bf9bfb6023edab8e4aed49850149bcc9a49e161e93b247b9f0042ae76b0f76d3R82] no longer serves a purpose.  Should this and its friends (`DegradationDecision.onDegradeFeature()`, `ConfigurationCacheProblems.reportDegradingFeature()`) all be cleaned up, too?  Is there any reason to expect a new "feature" to be added that is not CC-compatible?

ghale (approved):
This all looks pretty sane to me.  

I left a comment on some stuff I suspect we could remove now, but this isn't blocking.

We should mention this in the release notes along with a mention of the does-not-reset-worktree-on-next-build behavior.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGracefulDegradationIntegrationTest.groovy
big-guy: I think I was going to check if it made sense to artificially support something else, but I agree with you that this can probably just all be removed.

ghale:
@bot-gradle test this

ghale:
@bot-gradle test this

ghale:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGracefulDegradationIntegrationTest.groovy
ghale: I went ahead and removed the feature stuff.

lkasso (approved):
Docs LGTM. I made some minor updates.

PR #37782, opened 2026-05-04
Make Project.getProperties() an IP violation

### LINKED ISSUE #37385: Remove `getProperties()` API for Project and scripts

reinsch82:
## Summary
- Promote `Project.getProperties()` (and the script-level bare-name `properties` in the Groovy DSL) to a hard Isolated Projects violation. The method is already deprecated in Gradle 9 and scheduled to fail with an error in Gradle 10; under IP, treat it as a hard violation now.
- Implementation hooks the violation into `TrackingDynamicLookupRoutine.getProperties(...)`, the IP-only `DynamicLookupRoutine` impl. Both call paths — `DefaultProject.getProperties()` (direct) and `BasicScript.getProperties()` (script-level dynamic) — route through it, so a single chokepoint covers both with no new interface or cross-layer service wiring.

## Test plan
- [x] Added new test in `IsolatedProjectsAccessFromGroovyDslIntegrationTest` covering own-project access via `properties`, `project.properties`, and `project.getProperties()`.
- [x] Updated existing parent-walk test row to also expect the new general violation when `expr == "properties"`.
- [x] Updated `TrackingDynamicLookupRoutineTest` to supply the two new constructor params (Mockito deep stubs).
- [x] `:configuration-cache:test --tests "*TrackingDynamicLookupRoutineTest*"` passes.
- [x] `:configuration-cache:forkingIntegTest --tests "*IsolatedProjectsAccessFromGroovyDslIntegrationTest*"` passes.
- [x] `:core:forkingIntegTest --tests "*GetPropertiesDeprecationIntegrationTest*"` passes (non-IP deprecation behavior unchanged).
- [x] `./gradlew sanityCheck` passes.


reinsch82:
@bot-gradle test RfN

subprojects/core/src/main/java/org/gradle/api/internal/project/IsolatedProjectsApiAccessReporter.java
```diff
@@ -0,0 +1,40 @@
...
+ * Bridge that lets call sites in {@code :core} report Isolated Projects
+ * violations for selected Project APIs without depending on the
+ * {@code :configuration-problems-base} module directly.
+ * <p>
+ * The default implementation is a no-op; the Isolated Projects-aware
+ * implementation is wired in {@code :configuration-cache} when IP is
+ * enabled.
+ */
+@ServiceScope(Scope.BuildTree.class)
+public interface IsolatedProjectsApiAccessReporter {
```
reinsch82: Introduced a new interface, because otherwise we would either need to draw more stuff in to `core` or the refactoring would be even bigger.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsAccessFromGroovyDslIntegrationTest.groovy
```diff
@@ -524,17 +533,17 @@ class IsolatedProjectsAccessFromGroovyDslIntegrationTest extends AbstractIsolate
...
+            executer.expectDocumentedDeprecationWarning(SCRIPT_GET_PROPERTIES_DEPRECATION)
         }
         isolatedProjectsFails(":a:help")
 
         then:
         fixture.assertStateStoredAndDiscarded {
             projectsConfigured(":", ":a")
             problem("Build file 'a/build.gradle': line 2: Project ':a' cannot dynamically look up a $kind in the parent project ':'")
+            if (expr == "properties") {
+                problem("Build file 'a/build.gradle': line 2: use of 'Project.getProperties()' is not allowed with Isolated Projects")
```
alllex: ❓  Why does it still issue two violations?

subprojects/core/src/main/java/org/gradle/api/internal/project/IsolatedProjectsApiAccessReporter.java
alllex: 🤔  This feels like a wrong way to slice the abstaction. Can we instead make `IsolatedProjectsProblemsReporter` a Java interface that lives in core, but keep the implementation where it is?

subprojects/core/src/integTest/groovy/org/gradle/api/DynamicObjectIntegrationTest.groovy
alllex, line 1: 🤔  The conditionals look complicated. Given that our plan with `getProperties()` is different from the other Property API methods (we want to remove it), can we split out the tests for it to be separate?

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
```diff
@@ -1197,6 +1197,12 @@ public boolean hasProperty(String propertyName) {
             .willBecomeAnErrorInGradle10()
             .withUpgradeGuideSection(9, "deprecated_get_properties")
             .nagUser();
```
alllex, line 1203: ❓  Do you think it makes sense to issue a deprecation and a violation at the same time?

alllex (changes requested):

subprojects/core/src/main/java/org/gradle/api/internal/project/IsolatedProjectsApiAccessReporter.java
reinsch82: `org.gradle.internal.configuration.problems.ProblemFactory` and `org.gradle.internal.configuration.problems.PropertyProblem` are not in core yet, these would draw a lot of other things into core. I assume making core bigger is not a desired outcome.

reinsch82:
@coderabbitai review

reinsch82:

platforms/core-configuration/base-diagnostics/src/integTest/groovy/org/gradle/api/tasks/diagnostics/PropertyReportTaskIntegrationTest.groovy
```diff
@@ -17,7 +17,10 @@
 package org.gradle.api.tasks.diagnostics
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
+import org.gradle.test.precondition.Requires
+import org.gradle.test.preconditions.TestExecutionPreconditions
 
+@Requires(value = TestExecutionPreconditions.NotIsolatedProjects, reason = "The `properties` task calls Project.getProperties() internally, which is a hard violation under Isolated Projects")
```
alllex: 💅 This should be marked `@ToBeFixedForIsolatedProjects` instead, because we want the task to work, it's not an inherent limitation of IP.

Can you check if we already have an issue for that and if not, create it?

platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/BuildInitPluginIntegrationTest.groovy
alllex, line 1: 💅   Let's also mark this with `@ToBeFixedForIsolatedProjects`

alllex (approved):
LGTM!

Though we should update annotations on some tests to show the right intention

PR #37783, opened 2026-05-04
Add back setters to the Provider API migration

### LINKED ISSUE #37342: Research keeping deprecated setters for migrated properties
## Summary

Investigate the feasibility and impact of keeping old setter methods (e.g., `setClasspath()`, `setMaxParallelForks()`, `setCompilerArgs()`) as `@Deprecated` forwarders to the new Provider API (`getX().set()` / `getX().setFrom()`), rather than removing them entirely during the Provider API migration.

Key Insight: Based on our research, keeping deprecated setters is the single highest-impact mitigation available. It would eliminate ~40% of the changes required by users during the migration to the lazy API. (e.g., the Spring Framework would require zero user code changes with this mitigation alone).

## What to investigate

1. **Implementation**
  - What is the full list of setters we need to keep?
  - Decoration Logic: The current decoration infrastructure likely does not allow keeping setters alongside `Property<T>` getters. Investigate adding exceptions to permit this hybrid state.
  - Validation: Verify forwarding works correctly for properties (e.g. simple set, `setFrom`, file conversion, etc.).

2. **IDE behavior for Groovy**
  Check behavior in IntelliJ when deprecated setters are present in the Groovy DSL:
  - Are deprecated setters shown with strikethrough in code completion?
  - Does the IDE suggest the new API as a replacement?
  - Risk Assessment: Is there a risk that users continue using deprecated setters indefinitely because Groovy doesn't surface deprecation warnings as prominently as Kotlin/Java?

3. **IDE behavior for Kotlin**
  - Verify that the Kotlin compiler emits warnings for `@Deprecated` usage as expected for these hand-written setter shims.

4. **Which property types benefit from this?**
  - Verify that this really simplify upgrades 
  - Potential test: re-add setters just for the properties used in Spring Framework and check if everything works without upgrade
  - Should we also want to add `setProperty(Provider<T>)`?

5. Deprecation Timeline
  - Gradle 10.0: Keep setters (functional, not yet deprecated).
  - Gradle 10.1: Mark as **deprecated-for-removal** to provide users a migration window.

---

## Context

- Related: #36583 (Re-evaluate APIs we upgraded)
- Related: #31497 (Decide if we want to re-add `setExecutable(Object)` and `setWorkingDir(Object)`)
- Related: #33279 (Deprecate `setWorkingDir(Object)`)
- Related: #33015 (Deprecate `setCommandLine()` methods)

hegyibalint:
Fixes https://github.com/gradle/gradle/issues/37342


platforms/core-configuration/base-diagnostics/src/main/java/org/gradle/api/tasks/diagnostics/ConventionReportTask.java
```diff
@@ -79,21 +83,39 @@ protected ConventionReportTask() {
      */
     @Optional
     @OutputFile
-    @ReplacesEagerProperty
+    @ReplacesEagerProperty(replacedAccessors = @ReplacedAccessor(value = GETTER, name = "getOutputFile"))
```
asodja: 💅 Note that annotation processor already uses the right name if original getter has "getter" by convention.

So name = "<value>" is normally not needed. Just a nitpick though

platforms/core-runtime/internal-instrumentation-processor/src/main/java/org/gradle/internal/instrumentation/extensions/property/PropertyUpgradeAnnotatedMethodReader.java
```diff
@@ -470,62 +456,52 @@ private BinaryCompatibility readBinaryCompatibility(AnnotationMirror annotation)
...
     private AccessorSpec getAccessorSpec(ExecutableElement method, AnnotationMirror annotation, DeprecationSpec parentDeprecationSpec, BinaryCompatibility binaryCompatibility) {
         String methodName = AnnotationUtils.findAnnotationValue(annotation, "name")
             .map(v -> (String) v.getValue())
             .orElseThrow(() -> new AnnotationReadFailure("Missing 'name' attribute in @ReplacedAccessor"));
         AccessorType accessorType = AnnotationUtils.findAnnotationValue(annotation, "value")
             .map(v -> AccessorType.valueOf(v.getValue().toString()))
             .orElseThrow(() -> new AnnotationReadFailure("Missing 'value' attribute in @ReplacedAccessor"));
+        if (accessorType == AccessorType.SETTER) {
+            // Setter synthesis is no longer supported; explicit SETTER entries in replacedAccessors are silently ignored.
+            return null;
```
asodja: ❌ Let's do annotation processor cleanup in a separate PR and target master instead of provider API branch

I realized that is a infrastructure change, so we should keep this branch as much as possible without infrastructure changes as possible

platforms/software/maven/src/main/java/org/gradle/api/publish/maven/tasks/GenerateMavenPom.java
```diff
@@ -78,6 +78,26 @@ public void setPom(MavenPom pom) {
...
+        getDestination().fileValue(destination);
+    }
+
+    /**
+     * Sets the destination the descriptor will be written to.
+     *
+     * <p>The argument is evaluated as per {@link org.gradle.api.Project#file(Object)}.
+     */
+    public void setDestination(Object destination) {
+        ProviderApiDeprecationLogger.logDeprecation(GenerateMavenPom.class, "setDestination(Object)", "getDestination()");
```
asodja: 💅 Why we have this deprecation? Is that from before?

subprojects/core-api/src/main/java/org/gradle/caching/configuration/BuildCache.java
```diff
@@ -46,9 +56,16 @@ default Property<Boolean> getIsEnabled() {
...
      * Controls whether a given build can store outputs in the build cache.
      */
-    @ReplacesEagerProperty(originalType = boolean.class)
+    @ReplacesEagerProperty(replacedAccessors = @ReplacedAccessor(value = GETTER, name = "isPush", originalType = boolean.class))
     Property<Boolean> getPush();
 
+    /**
+     * Sets whether a given build can store outputs in the build cache.
+     */
+    default void setPush(boolean push) {
```
asodja: I read JLS
> Adding a default method, or changing a method from abstract to default, does not break compatibility with pre-existing binaries, but may cause an IncompatibleClassChangeError if a pre-existing binary attempts to invoke the method. This error occurs if the qualifying interface of the method invocation, K, is a subinterface of two interfaces, I and J, where both I and J declare a default method with the same signature and result, and neither I nor J is a subinterface of the other.

So there is some risk there. Let's just not risk it and mirror implemention of these methods as is on master: in implementation classes

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/TestReport.java
```diff
@@ -82,15 +81,19 @@ protected MetadataRendererRegistry getMetadataRendererRegistry() {
...
+    @ReplacesEagerProperty(replacedAccessors = @ReplacedAccessor(value = GETTER, name = "getDestinationDir"), deprecation = @ReplacedDeprecation(removedIn = RemovedIn.GRADLE9))
     public abstract DirectoryProperty getDestinationDirectory();
 
+    /**
+     * Sets the directory to write the HTML report to.
+     *
+     * @deprecated Use {@link #getDestinationDirectory()} instead.
+     */
+    @Deprecated
+    public void setDestinationDir(File destinationDir) {
```
asodja: ❌ This setter was actually removed with Gradle 9. We should not re-add it

platforms/jvm/jvm-compiler-worker/src/main/java/org/gradle/api/tasks/compile/CompileOptions.java
```diff
@@ -373,15 +492,19 @@ public Property<Boolean> getIsIncremental() {
...
+    @ReplacesEagerProperty(replacedAccessors = @ReplacedAccessor(value = GETTER, name = "getAnnotationProcessorGeneratedSourcesDirectory"), deprecation = @ReplacedDeprecation(removedIn = RemovedIn.GRADLE9))
     public abstract DirectoryProperty getGeneratedSourceOutputDirectory();
 
+    /**
+     * Returns the directory to place source files generated by annotation processors.
+     *
+     * @deprecated Use {@link #getGeneratedSourceOutputDirectory()} instead.
+     */
+    @Deprecated
+    public void setAnnotationProcessorGeneratedSourcesDirectory(File file) {
```
asodja: ❌ This setter was actually removed with Gradle 9. We should not re-add it

platforms/jvm/language-jvm/src/main/java/org/gradle/api/tasks/compile/AbstractCompile.java
```diff
@@ -61,16 +72,23 @@ public AbstractCompile() {
...
         return destinationDirectory;
     }
 
+    /**
+     * Sets the directory to generate the {@code .class} files into.
+     *
+     * @deprecated Use {@link #getDestinationDirectory()} instead.
+     */
+    @Deprecated
+    public void setDestinationDir(File destinationDir) {
```
asodja: ❌ This setter was actually removed with Gradle 9. We should not re-add it

asodja (changes requested):

hegyibalint:
@bot-gradle test this

hegyibalint:
@bot-gradle test this

platforms/software/maven/src/main/java/org/gradle/api/publish/maven/tasks/GenerateMavenPom.java
hegyibalint: This was present in the adapter below.
https://github.com/gradle/gradle/pull/37783/changes#diff-8d5336ce4406e7e57b8373cf0e6cff3891f544c61aa03c06fbea829b8ae33a65L99

The migration just takes those setters, and move them back into the same place where the `Property` is.

platforms/jvm/language-java/src/integTest/groovy/org/gradle/api/tasks/JavaExecSubclassingIntegrationTest.groovy
```diff
@@ -51,7 +51,7 @@ class JavaExecSubclassingIntegrationTest extends AbstractIntegrationSpec {
         """
 
         when:
-        executer.expectDocumentedDeprecationWarning("Changing property value of task ':run' property 'jvmArgs' at execution time. This behavior has been deprecated. Starting with Gradle 11, changing property value of task ':run' property 'jvmArgs' at execution time will become an error.")
+        executer.expectDocumentedDeprecationWarning("Changing property value of property 'jvmArguments' at execution time. This behavior has been deprecated. Starting with Gradle 11, changing property value of property 'jvmArguments' at execution time will become an error.")
```
asodja: 🤔 Is this change intentional? Didn't we fix a similar change at one point?

platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
```diff
@@ -197,6 +212,23 @@ public ListProperty<String> getJvmArgs() {
...
+        javaExecSpec.getJvmArguments().set(arguments);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void setJvmArgs(Iterable<?> arguments) {
+        javaExecSpec.getJvmArguments().empty();
+        jvmArgs(arguments);
```
asodja: 🤔 I see there is a different behaviour between `setJvmArgs(List<String> arguments)` and `setJvmArgs(Iterable<?> arguments)`. I think we should just do `javaExecSpec.getJvmArguments().empty();` and  call `addJvmArguments` otherwise behaviour would be different between these methods and also different if you use `getJvmArgs().set()` directly. So I believe doing `javaExecSpec.getJvmArguments().empty();` and `addJvmArguments(arguments)`. is the only compatible way even if slighlty differes from master behaviour.

Even `jvmArgs(Iterable<?> arguments)` is a bit funny, it should also call `addJvmArguments` without `javaExecSpec.checkDebugConfiguration(arguments);`, but that change is here from before so you don't need to go down `jvmArgs(Iterable<?> arguments)` rabbit hole.

platforms/jvm/antlr/src/main/java/org/gradle/api/plugins/antlr/AntlrTask.java
```diff
@@ -162,6 +204,15 @@ public Property<Boolean> getIsTraceTreeWalker() {
     @ReplacesEagerProperty
     public abstract ConfigurableFileCollection getAntlrClasspath();
 
+    /**
+     * Specifies the classpath containing the Ant ANTLR task implementation.
+     *
+     * @param antlrClasspath The Ant task implementation classpath. Must not be null.
+     */
+    public void setAntlrClasspath(FileCollection antlrClasspath) {
```
asodja: ❓ This is protected on master, do we need it public here?

platforms/core-execution/build-cache-http/src/main/java/org/gradle/caching/http/HttpBuildCache.java
asodja: 💅 This has no `throws URISyntaxException` on master. Can we also fix that and remove `throws URISyntaxException`?

asodja: 💅 Can we also remove this, it's unused and not present on master

I know it's not your change, but I noticed that here

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
```diff
@@ -1104,11 +1108,15 @@ boolean claimPropertyImplementation(PropertyMetadata property) {
...
             }
+            // When all getters are abstract, the user can't be managing a backing field themselves;
+            // any concrete setter is a forwarder, so don't let it prevent claiming.
+            boolean allAbstract = allGettersAreAbstract(property);
             for (Method setter : property.setters) {
                 if (!Modifier.isAbstract(setter.getModifiers())
                     // TODO: fix/remove in Gradle 9.0, this takes care of Property<URI> in special way
                     //    Note: we have a similar logic in AsmBackedClassGenerator#isLazyUriPropertyWithSetObject
-                    && !isLazyUriProperty(setter, property)) {
+                    && !isLazyUriProperty(setter, property)
```
asodja: ❓ Do we still need `!isLazyUriProperty(setter, property)`? 

Can you maybe open an issue to remove `isLazyUriProperty` once we merge this?

platforms/jvm/jacoco/src/integTest/groovy/org/gradle/testing/jacoco/plugins/JacocoCachingIntegrationTest.groovy
```diff
@@ -69,7 +69,7 @@ class JacocoCachingIntegrationTest extends AbstractIntegrationSpec implements Di
         javaProjectUnderTest.writeIntegrationTestSourceFiles()
         buildFile << """
             integrationTest.jacoco {
-                destinationFile = test.jacoco.destinationFile
+                destinationFile.set(test.jacoco.destinationFile)
```
asodja: ❓ Why is this change needed?

platforms/jvm/language-jvm/src/main/java/org/gradle/api/tasks/compile/AbstractCompile.java
```diff
@@ -53,6 +53,15 @@ public AbstractCompile() {
     @ReplacesEagerProperty
     public abstract ConfigurableFileCollection getClasspath();
 
+    /**
+     * Sets the classpath to use to compile the source files.
+     *
+     * @param configuration The classpath. Must not be null, but may be empty.
+     */
+    public void setClasspath(org.gradle.api.file.FileCollection configuration) {
```
asodja: 💅 Do we need Full type name, i.e. `org.gradle.api.file.FileCollection` or can it be `FileCollection`

platforms/software/ivy/src/integTest/groovy/org/gradle/api/publish/ivy/IvyPublishDescriptorCustomizationIntegTest.groovy
```diff
@@ -123,6 +123,7 @@ class IvyPublishDescriptorCustomizationIntegTest extends AbstractIvyPublishInteg
         """
 
         when:
+        executer.expectDocumentedDeprecationWarning("The GenerateIvyDescriptor.setDestination(Object) method has been deprecated. This is scheduled to be removed in Gradle 10. Use getDestination() property instead Consult the upgrading guide for further information: https://docs.gradle.org/current/userguide/upgrading_version_9.html#provider_api_migration")
```
asodja: ❓ How come we have this deprecation now?

platforms/software/ivy/src/main/java/org/gradle/api/publish/ivy/tasks/GenerateIvyDescriptor.java
```diff
@@ -78,6 +78,26 @@ public void setDescriptor(IvyModuleDescriptorSpec descriptor) {
...
+        getDestination().fileValue(destination);
+    }
+
+    /**
+     * Sets the destination the descriptor will be written to.
+     *
+     * <p>The argument is evaluated as per {@link org.gradle.api.Project#file(Object)}.
+     */
+    public void setDestination(Object destination) {
+        ProviderApiDeprecationLogger.logDeprecation(GenerateIvyDescriptor.class, "setDestination(Object)", "getDestination()");
```
asodja: 🤔 Let's just remove this deprecation, we can add deprecations later

platforms/software/maven/src/integTest/groovy/org/gradle/api/publish/maven/MavenPublishPomCustomizationIntegTest.groovy
```diff
@@ -270,6 +270,7 @@ class MavenPublishPomCustomizationIntegTest extends AbstractMavenPublishIntegTes
         """
 
         when:
+        executer.expectDocumentedDeprecationWarning("The GenerateMavenPom.setDestination(Object) method has been deprecated. This is scheduled to be removed in Gradle 10. Use getDestination() property instead Consult the upgrading guide for further information: https://docs.gradle.org/current/userguide/upgrading_version_9.html#provider_api_migration")
```
asodja: 🤔 Same as for Ivy, let's just remove deprecation for now, we can add it later

platforms/software/maven/src/main/java/org/gradle/api/publish/maven/tasks/GenerateMavenPom.java
asodja: I see, but let's remove the deprection for now and we can add it later in a different effort

testing/architecture-test/src/changes/accepted-changes/javaForkOptions-accepted-public-api-changes.json
```diff
@@ -1,12 +1,3 @@
...
-        {
-            "type": "org.gradle.kotlin.dsl.GradleApiKotlinDslExtensions_4vqe7vkwi3h45wskb54l2vuiuKt",
-            "member": "Method org.gradle.kotlin.dsl.GradleApiKotlinDslExtensions_4vqe7vkwi3h45wskb54l2vuiuKt.setSystemProperties(org.gradle.process.JavaForkOptions,kotlin.Pair[])",
-            "acceptation": "JavaForkOptions.getSystemProperties() upgrade",
-            "changes": [
-                "Method has been removed"
-            ]
-        }
-    ]
+    "acceptedApiChanges": []
```
asodja: 💅 Can we remove this file now?

testing/architecture-test/src/changes/accepted-changes/processForkOptions-accepted-public-api-changes.json
```diff
@@ -1,12 +1,3 @@
...
-        {
-            "type": "org.gradle.kotlin.dsl.GradleApiKotlinDslExtensions_4i8m7ubpazumv0o7fg1f6fdogKt",
-            "member": "Method org.gradle.kotlin.dsl.GradleApiKotlinDslExtensions_4i8m7ubpazumv0o7fg1f6fdogKt.setEnvironment(org.gradle.process.ProcessForkOptions,kotlin.Pair[])",
-            "acceptation": "Auto generated accessors",
-            "changes": [
-                "Method has been removed"
-            ]
-        }
-    ]
+    "acceptedApiChanges": []
```
asodja: 💅 Can we remove this file now?

testing/architecture-test/src/changes/accepted-changes/checkstyleTask-accepted-public-api-changes.json
```diff
@@ -1,20 +1,3 @@
...
-        {
-            "type": "org.gradle.kotlin.dsl.GradleApiKotlinDslExtensions_erf2xhrqr36mq5qab8vprh3ecKt",
-            "member": "Method org.gradle.kotlin.dsl.GradleApiKotlinDslExtensions_erf2xhrqr36mq5qab8vprh3ecKt.setConfigProperties(org.gradle.api.plugins.quality.Checkstyle,kotlin.Pair[])",
-            "acceptation": "Checkstyle.getConfigProperties() upgrade",
-            "changes": [
-                "Method has been removed"
-            ]
-        }
-    ]
+    "acceptedApiChanges": []
```
asodja: 💅 Can we remove this file now?

testing/architecture-test/src/changes/accepted-changes/checkstyleExtension-accepted-public-api-changes.json
```diff
@@ -1,20 +1,3 @@
...
-        {
-            "type": "org.gradle.kotlin.dsl.GradleApiKotlinDslExtensions_3nmb2qjzin72lydud9myfmci6Kt",
-            "member": "Method org.gradle.kotlin.dsl.GradleApiKotlinDslExtensions_3nmb2qjzin72lydud9myfmci6Kt.setConfigProperties(org.gradle.api.plugins.quality.CheckstyleExtension,kotlin.Pair[])",
-            "acceptation": "CheckstyleExtension.getConfigProperties() upgrade",
-            "changes": [
-                "Method has been removed"
-            ]
-        }
-    ]
+    "acceptedApiChanges": []
```
asodja: 💅 Can we remove this file now?

asodja (changes requested):

platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
hegyibalint: Hmm, now looking into `JavaExec`, I don't really understand why this behavior was done like this in the `provider-api-migration`. For me it looks like the `master` version already uses providers, and we would need almost no modifications to that one.

What stops us from just adapting exactly what `master` does today? 
These properties don't even need the annotation: they were introduced 8.1.

platforms/jvm/language-java/src/integTest/groovy/org/gradle/api/tasks/JavaExecSubclassingIntegrationTest.groovy
hegyibalint: This is going to be the same problem that we fixed a few weeks ago: instrumentation runs in the wrong place, and we don't associate the task object with the property automatically.

hegyibalint: This might be influenced about the decision in https://github.com/gradle/gradle/pull/37783#discussion_r3310716585 — let's nail that one first

platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
asodja: `getJvmArgs()` is not a `ListProperty<String>` on master. On provider-api-migration we made it ListProperty and basically unified `getJvmArgs()` and `getJvmArguments()`. And setJvmArgs should just work as jvmArguments.set(value) to have consistent behaviour for `=`

platforms/core-configuration/base-diagnostics/src/main/java/org/gradle/api/tasks/diagnostics/ConventionReportTask.java
hegyibalint: Solved, thanks!

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
hegyibalint: I can create an issue.
BTW, there is no advertised `AsmBackedClassGenerator#isLazyUriPropertyWithSetObject` anymore (?)

platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
hegyibalint: As discussed, I am changing all the behavior to just set the properties.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
hegyibalint: https://github.com/gradle/gradle/issues/38038

platforms/jvm/language-java/src/integTest/groovy/org/gradle/api/tasks/JavaExecSubclassingIntegrationTest.groovy
hegyibalint: This is an interesting case. Let me refresh our memory.
So this happens because `JavaExec` has a fully implemented getter, meaning that we won't apply the [`attachOwner` instrumentation](https://github.com/gradle/gradle/blob/48faacae3de1e82fe3804bde0785167b4038f7a6/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AsmBackedClassGenerator.java#L1565-L1577) here in the `JavaExec` class; it will be done in the lower levels.

The simplest solution now was to just do it, and use `ManagedObjectFactory.attachOwner` to glue on the task where we are. But this is probably worth an issue, because there might be structural solutions for this.

.idea/gradle.xml
```diff
@@ -0,0 +1,343 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
asodja: ❓ Do we need to commit this file?

platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/JavaExec.java
```diff
@@ -189,30 +192,62 @@ public Provider<List<String>> getAllJvmArgs() {
...
      * {@inheritDoc}
      */
     @Override
     public ListProperty<String> getJvmArgs() {
-        return javaExecSpec.getJvmArguments();
+        // Manual owner-attach: JavaExec.getJvmArgs() is a concrete delegating getter,
+        // so the class generator's automatic instrumentation does not fire here. The
+        // returned property would otherwise carry only its spec-level display name
+        // (jvmArguments), losing the task context in error/deprecation messages.
+        return ManagedObjectFactory.attachOwner(javaExecSpec.getJvmArguments(), (ModelObject) this, "jvmArgs");
```
asodja: 🤔 What if we just call
```
return this.getJvmArguments();
```
would that trigger `ManagedObjectFactory.attachOwner` implicitly and we can avoid manually calling ManagedObjectFactory?

testing/architecture-test/src/changes/accepted-changes/jacoco-accepted-public-api-changes.json
```diff
@@ -0,0 +1,12 @@
+{
+    "acceptedApiChanges": [
+        {
+            "type": "org.gradle.testing.jacoco.plugins.JacocoTaskExtension",
+            "member": "Method org.gradle.testing.jacoco.plugins.JacocoTaskExtension.setDestinationFile(org.gradle.api.provider.Provider)",
```
asodja: ❓ I think we shouldn't removed that but keep it as is on master. Or is that not possible?

asodja (approved):
LGTM

PR #37494, opened 2026-04-09
Inspect Kotlin delegates for unsupported types usage

tresat:
> [!NOTE]
> Merge https://github.com/gradle/gradle/pull/37466 first.

Using types like Configuration or SourceDirectorySet as a value type
for a Property causes confusing type-mismatch errors during cache
loading. This change introduces a check during serialization to fail
fast with a clear diagnostic message and resolution suggestion.

Fixes:
- https://github.com/gradle/gradle/issues/16177

Issue #16177, opened 2021-02-16
Disallowed types in configuration cache are not detected when using Kotlin lazy properties


Vampire:
In Gradle 6.8.2 with a task property like
```kt
private val classPath = project.configurations.detachedConfiguration(
    project.dependencies.create("com.sun.xml.ws:jaxws-tools:2.3.3")
)
```
the configuration cache complains that `Configuration` is not allowed.

When using
```kt
private val classPath: FileCollection = project.configurations.detachedConfiguration(
    project.dependencies.create("com.sun.xml.ws:jaxws-tools:2.3.3")
)
```
it works.

But when using
```kt
private val classPath by lazy {
    project.configurations.detachedConfiguration(
        project.dependencies.create("com.sun.xml.ws:jaxws-tools:2.3.3")
    )
}
```
there is no complaint, storing the configuration cache works, but then on reusing you get a cast exception.

Instead using
```kt
private val classPath: FileCollection by lazy {
    project.configurations.detachedConfiguration(
        project.dependencies.create("com.sun.xml.ws:jaxws-tools:2.3.3")
    )
}
```
works.

bamboo:
This is still relevant.

bamboo:
Related to #17821

PR #37786, opened 2026-05-04
Detect incompatible type roundtrips at CC store time

### LINKED ISSUE #31139: Emit serialization error if the bean property value cannot be assigned to the field after deserialization
There are some cases where the value type changes after the configuration cache serialization roundtrip. For example, a `Configuration` becomes `FileCollection`. This isn't always easy to debug because configuration cache complains at deserialization time and there is no hint that the value type has changed.

Ideally, this should happen at serialization time and be properly explained.

### LINKED ISSUE #35129: SourceDirectorySet containing `provider.flatMap { ... }.map { ... }` elements breaks configuration cache
### Context

This issue was first reported at https://issuetracker.google.com/446189433. After debugging, I found that this is probably caused by a bug in Gradle, as demonstrated in the project below.

### Steps to reproduce

1. Open the attached project: [SourceDirectorySetAndFlatMapCCIssueProject.zip](https://github.com/user-attachments/files/22538924/SourceDirectorySetAndFlatMapCCIssueProject.zip)

2. Run ` ./gradlew clean :compileSources --configuration-cache -s`, the build will fail with:

```
* What went wrong:
Configuration cache state could not be cached: field `__sourceFiles__` of task `:compileSources` of type `Build_gradle$CompileSourcesTask`: error writing value of type 'org.gradle.api.internal.file.collections.DefaultConfigurableFileCollection'
> Querying the mapped value of flatmap(provider(task 'generateSources', class Build_gradle$GenerateSourcesTask)) before task ':generateSources' has completed is not supported

* Try:
> Run with --scan to generate a Build Scan (Powered by Develocity).

* Exception is:
org.gradle.internal.cc.base.exceptions.ConfigurationCacheError: Configuration cache state could not be cached: field `__sourceFiles__` of task `:compileSources` of type `Build_gradle$CompileSourcesTask`: error writing value of type 'org.gradle.api.internal.file.collections.DefaultConfigurableFileCollection'
        at org.gradle.internal.cc.impl.problems.ConfigurationCacheProblems.onError(ConfigurationCacheProblems.kt:175)
        at org.gradle.internal.serialize.graph.AbstractIsolateContext.onError(Contexts.kt:396)
        at org.gradle.internal.serialize.graph.BeanPropertyExtensionsKt.writePropertyValue(BeanPropertyExtensions.kt:36)
        at org.gradle.internal.serialize.beans.services.BeanPropertyWriter.writeStateOf(BeanPropertyWriter.kt:54)
...
Caused by: org.gradle.api.InvalidUserCodeException: Querying the mapped value of flatmap(provider(task 'generateSources', class Build_gradle$GenerateSourcesTask)) before task ':generateSources' has completed is not supported
        at org.gradle.api.internal.provider.TransformBackedProvider.lambda$beforeRead$0(TransformBackedProvider.java:98)
        at org.gradle.api.internal.provider.ValueSupplier$TaskProducer.visitContentProducerTasks(ValueSupplier.java:134)
        at org.gradle.api.internal.provider.TransformBackedProvider.beforeRead(TransformBackedProvider.java:95)
        at org.gradle.api.internal.provider.TransformBackedProvider.calculateOwnValue(TransformBackedProvider.java:80)
        at org.gradle.api.internal.provider.AbstractMinimalProvider.calculateValue(AbstractMinimalProvider.java:115)
```

The project is set up such that the input files to a task is set to a `SourceDirectorySet`, and the `SourceDirectorySet` contains elements that are created from a `provider.flatMap { ... }.map { ... }` call. This setup is similar to the one in https://issuetracker.google.com/446189433, where [KotlinCompilation.defaultSourceSet.kotlin](https://cs.android.com/android-studio/platform/tools/base/+/4ea2a821fe74f8551cc3182cbd1c87824c541085:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/services/BuiltInKotlinServices.kt;l=330) is a `SourceDirectorySet`, and its value is a [flatMap { ... }.map { ...}](https://cs.android.com/android-studio/platform/tools/base/+/4c51f715beee1625cfdc9a16facb68eecfaddc92:build-system/gradle-core/src/main/java/com/android/build/api/variant/impl/TaskProviderBasedDirectoryEntryImpl.kt;l=49) (`directoryProvider` at this link is a `flatMap`).

Note:
  - If we don't use a `SourceDirectorySet`, the issue does not reproduce.
  - If we use `provider.map { ... }.map { ... }` instead of `provider.flatMap { ... }.map { ... }`, the issue does not reproduce.

### Workaround

`provider.map { ... }.map { ... }` fixes the issue in the sample project, but in the Android Gradle plugin we use `flatMap` a lot, and either it's not possible or there isn't a good reason to change them to `map`.

### Priority

This issue was a corner case before, but starting with AGP 9.0 the code path that exposes this issue will be exercised, so it will be very common (it will happen with AGP 9.0 when a source generating task is used).

So we would appreciate it if this issue was fixed in Gradle.

### Gradle version

9.1.0

tresat:
The configuration cache (CC) now detects, at store time, when a value's serialization codec would decode back as a type the receiving slot cannot accept — e.g. a `Configuration`-typed field is fed a value whose codec only decodes into `FileCollection`. Previously these mismatches passed the store silently and failed on load with an obscure type error far from the actual misuse site.

The new check covers three sites:

- **Bean fields** — codec decoded-type vs. the declared field type.
- **Managed properties** — codec decoded-type for the `T` of `Property<T>`, `ListProperty<T>`, `SetProperty<T>`, and (key + value) `MapProperty<K, V>`.
- **Kotlin property delegates** (`by lazy { ... }`, observable properties, etc.) — inspects the wrapped value via reflection and checks against the Kotlin getter's declared return type.

In `fail` mode each problem surfaces as the cause of the eventual `ConfigurationCacheProblemsException` with an `UnsupportedTypeException` carrying actionable resolution advice. In `warn` mode the build still completes: the offending value is dropped (bean fields get `null`, managed properties get `ExecutionTimeValue.missing()`) and a deferred problem is emitted, so `--configuration-cache-problems=warn` continues to be a viable escape hatch.

## Behavior changes

- Builds that previously cold-stored a CC entry then failed on the *next* load with an obscure deserialization error now fail (or warn) on the *first* run with a message naming the offending field/property and its incompatible type.
- A `MapProperty<Bad, Bad>` reports two problems — one for the key, one for the value — each with a position-specific resolution ("Avoid using X as a MapProperty key." vs. "...as a MapProperty value."). The HTML report dedupes by message+trace, so identical-text reports collapse to one entry.

### 1. The `WideningCodec` opt-in

A new interface `WideningCodec<T> : Codec<T>` marks codecs whose `decode` returns a type strictly wider than the runtime values they encode. Implementations declare:

- `decodedType: Class<T>` — the exact type produced by `decode`.
- `publicDecodedType: Class<*>` — display type for error messages (defaults to `decodedType`; lets internal types like `FileCollectionInternal` render as `FileCollection`).
- `wideningFix: String` — required user-facing fix line.

Two opt-ins in this PR:

- **`ConfigurationCodec`** — extends `FileCollectionCodec`, decodes into `FileCollectionInternal`, declares `publicDecodedType = FileCollection`. Must be registered before `FileCollectionCodec` — first match wins.
- **`SourceDirectorySetCodec`** — same shape; recommends `ConfigurableFileCollection` or `ConfigurableFileTree`.

`FileCollectionCodec` and `FileTreeCodec` were changed from `final` to `open class` to allow these subclasses.

### 2. Centralized widening helpers

All the widening dispatch and reporting helpers live in `WideningTypeSerializationFailureHelper.kt` (graph-serialization, `codecs` package):

```kotlin
fun WriteContext.findCodecThatWidensIncompatibly(
    declaredType: Class<*>,
    runtimeType: Class<*> = declaredType
): WideningCodec<*>?

suspend fun WriteContext.reportSerializationProblem(exception: Exception)

suspend fun WriteContext.reportIfIncompatibleRoundtrip(
    field: Field, fieldName: String, fieldValue: Any?
): Boolean

suspend fun WriteContext.reportIfUnsupportedPropertyValueType(
    propertyKind: Class<*>,
    valueType: Class<*>,
    resolutionFor: (widening: WideningCodec<*>, valueType: Class<*>) -> String = { w, _ -> w.wideningFix }
): Boolean
```

`findCodecThatWidensIncompatibly`'s `runtimeType` parameter exists because bean fields have both a declared type and a runtime value — looking up the codec by runtime type catches the case where `Configuration` flows into a `FileCollection` field before `FileCollectionCodec` resolves the encoding. Single-type call sites (lambda parameters, `Property<T>` value types) pass `declaredType` alone.

The `D.isAssignableFrom(widening.decodedType)` carve-out means the helper only flags genuine widenings — a `FileCollection`-declared field receiving a `Configuration` is fine; only the reverse arrow gets flagged.

### 3. Per-runtime-type codec lookup, cached

`BindingsBackedCodec` exposes `encodingForType(type: Class<*>): Encoding?` via a `CodecLookup` interface that `WriteContext` provides. The cache is shared with `encode`'s `taggedEncodingFor` via a `noMatch` sentinel:

- Hits store the matching `TaggedEncoding`.
- Misses store `noMatch`.
- `taggedEncodingFor` turns observed `noMatch` into `IllegalArgumentException`.
- `encodingForType` turns it into `null`.

Diagnostic call sites only need the `Encoding`, so `encodingForType` returns `Encoding?` not `TaggedEncoding?` — the tag is a serialization-internal concern.

### 4. Bean fields

`BeanPropertyWriter.writeStateOf` runs two checks per field:

- `reportIfIncompatibleRoundtrip` (centralized helper) — feeds `(field.type, fieldValue.javaClass)` to `findCodecThatWidensIncompatibly`. Emits a deferred problem and signals "drop this value" on incompatibility.
- `reportIfUnsupportedKotlinDelegate` (private to BeanPropertyWriter) — only fires for fields whose name ends in `$delegate` AND whose value `instanceof`-checks as `Lazy` / `ObservableProperty` / `ReadWriteProperty`. Resolves the Kotlin property's declared getter return type via reflection (`KotlinDelegateInspector.kotlinPropertyGetterReturnType`), unwraps the delegate's stored value, then runs the same widening lookup against `(getterReturnType, wrappedValue.javaClass)`.

  **Uninitialized `by lazy` delegates are deliberately skipped** — there is no value yet to type-check, and the underlying Lazy codec forces the lazy via `writeReplace` during the subsequent encode pass, so the type concern is moot. The skip is logged at info level for diagnostic visibility.

When either check fires, `writeStateOf` substitutes `null` for the field value before calling `writePropertyValue`. `BeanPropertyReader.set` already tolerates `null` for object fields, so no read-side changes are needed.

### 5. Managed properties

Each property codec in `ProviderCodecs.kt` calls `reportIfUnsupportedPropertyValueType` for the relevant type argument(s) and routes around the unsupported value:

```kotlin
val unsupported = reportIfUnsupportedPropertyValueType(Property::class.java, type)
writeClass(type)
if (unsupported) {
    providerCodec.run { encodeValue(ExecutionTimeValue.missing<Any>()) }
} else {
    providerCodec.run { encodeProvider(value.provider) }
}
```

`MapPropertyCodec` reports key and value types separately, passing a resolution lambda that pins the position-specific wording (`"Avoid using X as a MapProperty key."` / `"...value."`).

`encodePreservingIdentityOf` is unaffected — the identity frame opens and closes around a self-consistent payload (either the encoded provider or the `missing()` placeholder).

### 6. Reporting via `reportSerializationProblem`

```kotlin
suspend fun WriteContext.reportSerializationProblem(exception: Exception)
```

Builds a `PropertyProblem(trace, structuredMessage, exception = exception, stackTracingFailure = failureFactory.create(exception))` and calls `onProblem(...)`. Three design points:

- **The exception is attached, not thrown.** This is what keeps warn mode honest — `onProblem` routes through the deferred-problem aggregation; only `onError` is fatal. The attached `UnsupportedTypeException` becomes a cause of `ConfigurationCacheProblemsException` in fail mode, so `failure.assertHasCause(...)` and `failure.assertHasResolution(...)` keep working.
- **`stackTracingFailure` is populated explicitly** via `ownerService<FailureFactory>().create(exception)` so the HTML report includes a stack trace for warn-mode problems.
- **Trace handling is the caller's responsibility.** Bean callers wrap with `withPropertyTrace(PropertyKind.Field, fieldName)`; property callers already have the correct trace open via `encodePreservingIdentityOf`.

### 7. Placeholder semantics

The cache must remain self-consistent after dropping a value, so each site writes a stable placeholder:

- **Bean fields** → `null`. The reader's reflection-based setter already accepts null for object fields.
- **Managed property values** → `ExecutionTimeValue.missing<Any>()`. The reader reconstructs an empty `Property` / `ListProperty` / `SetProperty` / `MapProperty` without further branching.

After load, the property reports `isPresent == false` or evaluates to an empty collection — code paths that already handle "missing convention value" continue to work.

### 8. Pre-existing analogues unified

`JavaRecordCodec` and `SerializedLambdaParametersCheckingCodec` both had ad-hoc versions of this check before this PR; they now both call `findCodecThatWidensIncompatibly`.

Supercedes:
- https://github.com/gradle/gradle/pull/37494
- https://github.com/gradle/gradle/pull/37466

And fixes:
- https://github.com/gradle/gradle/issues/19122
- https://github.com/gradle/gradle/issues/16177
- https://github.com/gradle/gradle/issues/31139

There is an in-progress PR to migrate the Shadow plugin away from relying on serialization `Configuration` to get at its files:
- https://github.com/GradleUp/shadow/pull/2003.


tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/NarrowingCodec.kt
```diff
@@ -0,0 +1,87 @@
...
+ * the field, serialization fails immediately rather than producing a cache
+ * entry that cannot be loaded.
+ *
+ * Example: `ConfigurationCodec` extends `FileCollectionCodec` and is bound
+ * for `Configuration`, but its `decode` (inherited from `FileCollectionCodec`)
+ * always returns a `FileCollectionInternal`.  A field declared as
+ * `Configuration` cannot accept that decoded value, so `ConfigurationCodec`
+ * opts in.
+ */
+interface NarrowingCodec<T : Any> : Codec<T> {
```
mlopatkin: ❓ Isn't it technically a "widening" codec? We have `Configuration`, but encode it as its parent type `FileCollection`, which is wider. Obviously, it will be some subtype of `FileCollection`, but a sibling. Narrowing (returning a subtype of `Configuration`) would be safe, as it can be assigned back.

mlopatkin (changes requested):
It is a good base, but we're getting overly strict for a minor release here. It is possible to have a working build when there is no actual value to serialize. Here is what I'd like us to change:

#### Honor `--configuration-cache-problems=warn` for populated cases

It was possible to downgrade (de)serialization problems to warnings. Now we fail fast and hard. We should allow users to shoot themselves in the foot with the warning mode :) Likely, we should just emit the warning and drop the value.

#### Deprecate, don't reject, the empty/unset case

For working builds like the Shadow plugin, we should not break them. If the holder isn't supported but happens to be empty, we can emit just a deprecation with the goal to elevate it to error in 10.

I haven't gone through the delegate dissection code yet, but plan to get to it when I'm back next week.

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/NarrowingCodec.kt
tresat: Done.

lkasso:
No documentation needed for this PR.

tresat:
@mlopatkin - This is ready for another look.

platforms/core-configuration/graph-serialization/src/test/kotlin/org/gradle/internal/serialize/graph/codecs/FindCodecThatWidensIncompatiblyTest.kt
```diff
@@ -0,0 +1,112 @@
...
+import org.gradle.internal.serialize.graph.Codec
+import org.gradle.internal.serialize.graph.ReadContext
+import org.gradle.internal.serialize.graph.WriteContext
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertSame
+import org.junit.Test
+import java.lang.reflect.Proxy
+
+
+class FindCodecThatWidensIncompatiblyTest {
```
abstratt: This naming style for the class seems unusual for the configuration platform (reads like a sentence). Claude tells me the dominant naming style for the descriptive stem is noun-first, verbs only appear as gerunds or baked into a noun (LazyAddition…, QueryAndMutate…).

| Pattern | What the stem names | ~Share | Examples |                                                                                                                                                                                  
  |---|---|---|---|                                                                                                                                                                                                                      
  | **Feature/component** | The architectural feature under test | ~45-50% | `ConfigurationCacheEnablement`, `KotlinDslScripts`, `FilePropertyLifecycle` |                                                                               
  | **Domain-type** | An internal type/container | ~5-7% | `DefaultDomainObjectCollection`, `NamedDomainObjectCollectionSchema`, `CompositeDomainObjectSet` |                                                                            
  | **Behavior/scenario** | An interaction or scenario | ~8-10% | `LazyAdditionToDomainObjectCollection`, `MutatingNamedDomainObjectContainerInHook`, `QueryAndMutateFromWithTypeGetByName` |                                            
  | **Error/failure** | An error / invalid / problem path | ~3-5% | `MutationFailureFromDisallowChanges`, `InvalidManagedModelMutation`, `ConfigurationCacheProblemReporting` |                                                          
  | **Instrumentation/InterOp** | A technical aspect across languages | ~6-8% | `ProcessInstrumentationInJava`, `PropertyKotlinInterOp`, `SystemPropertyInstrumentationInKotlin` |                                                       
  | **Serialization/Codec** | Round-tripping / codecs | ~2-3% | `UserTypesCodec`, `ConfigurationCacheJavaSerialization`, `SupplierCodec` |                                                                                               
  | **Validation/Reporting** | Checking & diagnostics | ~3-4% | `ConfigurationCacheFingerprintChecker`, `ModelPathValidation`, `SchemaBuildingFailureReporting` |

tresat: I'm happy to rename.  What should it be called?

`WideningCodecUnsupportedReportingTest`?

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/WideningTypeSerializationFailureHelper.kt
```diff
@@ -0,0 +1,165 @@
...
+ * Callers who need a new trace frame (e.g., bean-field encoders that have
+ * not yet entered the field's property trace) should wrap with
+ * [withPropertyTrace] themselves; this keeps the trace surface explicit at
+ * the call site instead of conflating it with the problem-reporting helper.
+ */
+suspend fun WriteContext.reportSerializationProblem(exception: Exception) {
+    val failureFactory = ownerService<FailureFactory>()
+    val message = StructuredMessage.build {
+        text("failed to serialize value of ")
+        reference(trace.toString())
```
mlopatkin: :x: this doesn't render correctly - it will be a structured message rendered as reference. So instead of
image

you get something way off: 

image

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/ConfigurationCacheCodecs.kt
```diff
@@ -433,9 +437,15 @@ class DefaultConfigurationCacheCodecs(
...
         bind(ConfigurableFileTreeCodec(fileCollectionFactory))
+        // SourceDirectorySetCodec must precede FileTreeCodec: SourceDirectorySet is a
+        // FileTreeInternal subtype, and the binding-walk picks the first match.
+        bind(SourceDirectorySet::class, SourceDirectorySetCodec(fileCollectionFactory, directoryFileTreeFactory, fileOperations))
         bind(FileTreeCodec(fileCollectionFactory, directoryFileTreeFactory, fileOperations))
         val fileCollectionCodec = FileCollectionCodec(fileCollectionFactory, artifactSetConverter, taskDependencyFactory)
         bind(ConfigurableFileCollectionCodec(fileCollectionCodec, fileCollectionFactory))
+        // ConfigurationCodec must precede FileCollectionCodec: Configuration is a
+        // FileCollectionInternal subtype, and the binding-walk picks the first match.
+        bind(Configuration::class, ConfigurationCodec(fileCollectionFactory, artifactSetConverter, taskDependencyFactory))
```
mlopatkin, line 456: :nail_care: can we move it before `val fileCollectionCodec`?

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/Contexts.kt
```diff
@@ -78,6 +78,14 @@ class DefaultWriteContext(
     specialEncoders: SpecialEncoders = SpecialEncoders()
 
 ) : AbstractIsolateContext<WriteIsolate>(codec, problemsListener, name), CloseableWriteContext, Encoder by encoder {
+    /**
+     * Root codec captured at construction. `getCodec()` returns the currently-pushed
+     * codec, which during nested encodes may be a transient codec that does not carry
+     * the binding registry. The widening-type lookup must query the binding registry
```
mlopatkin: :x: 
>The widening-type lookup must query the binding registry regardless of what is pushed

This isn't correct. The current codec is solely responsible for encoding the values in this context while it is active. If there is no widening codec available, there will be no widening, so there is no point in emitting anything.

platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/BeanPropertyWriter.kt
```diff
@@ -47,15 +62,63 @@ class BeanPropertyWriter(
...
+     */
+    private
+    suspend fun WriteContext.reportIfUnsupportedKotlinDelegate(field: Field, fieldName: String, fieldValue: Any?): Boolean {
+        // A Kotlin `by`-delegate is identified by BOTH the field name (compiler emits `<name>$delegate`)
+        // AND the value type. A regular field declared as `Lazy<T>` is not a delegate even though its
+        // value satisfies isKotlinDelegate(); skip it so normal codec-driven serialization handles it.
+        if (!field.name.endsWith("\$delegate") || !KotlinDelegateInspector.isKotlinDelegate(fieldValue)) return false
+        if (fieldValue is Lazy<*> && !fieldValue.isInitialized()) {
+            // Deliberate skip — see KDoc above.
+            logger.info(
```
mlopatkin: :x: I think `--info` is too high for such a check, let's lower it to debug.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGradlePropertiesIntegrationTest.groovy
```diff
@@ -743,4 +748,484 @@ class ConfigurationCacheGradlePropertiesIntegrationTest extends AbstractConfigur
         outputContains "Bar: one"
         // Must be cache miss
     }
+
```
mlopatkin, line 745: :x: These tests do not belong to this file. I suppose AI picked up "Properties" but hadn't considered what this means. `ConfigurationCacheUnsupportedTypesIntegrationTest` should be a good destination.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGradlePropertiesIntegrationTest.groovy
```diff
@@ -743,4 +748,484 @@ class ConfigurationCacheGradlePropertiesIntegrationTest extends AbstractConfigur
...
+        """
+
+        when:
+        configurationCacheFails "tasks"
+
+        then:
+        assertHasUnsupportedPropertyValueType(Property, Configuration, "tasks", "TaskReportTask")
+    }
+
+    def "reports error only once when multiple properties of same unsupported type exist"() {
```
mlopatkin: :thinking: Is it a good idea to limit error reporting to one per task? And is it actually true?

What I see is
```
4 problems were found storing the configuration cache.
- Task `:multiConf` of type `MultiConfTask`: failed to serialize value of 'field `__first__` of task `:multiConf` of type `MultiConfTask`'
- Task `:multiConf` of type `MultiConfTask`: failed to serialize value of 'field `__fourth__` of task `:multiConf` of type `MultiConfTask`'
- Task `:multiConf` of type `MultiConfTask`: failed to serialize value of 'field `__second__` of task `:multiConf` of type `MultiConfTask`'
- Task `:multiConf` of type `MultiConfTask`: failed to serialize value of 'field `__third__` of task `:multiConf` of type `MultiConfTask`'
```

Looks like the assertion is too weak for the intended purpose.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGradlePropertiesIntegrationTest.groovy
```diff
@@ -743,4 +748,484 @@ class ConfigurationCacheGradlePropertiesIntegrationTest extends AbstractConfigur
...
+        """
+
+        when:
+        configurationCacheFails "printFiles"
+
+        then:
+        assertHasUnsupportedPropertyValueType(Property, Configuration, "printFiles", "BrokenPrintTask")
+
+        where:
+        annotation << ["@Input", "@Internal", "@InputFiles", "@Classpath"]
```
mlopatkin: :thinking: CC doesn't rely on annotations for serialization, I'd keep just `@Internal` and un-parameterize the test to save CI time.

Here, and in other places in the file.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGradlePropertiesIntegrationTest.groovy
```diff
@@ -743,4 +748,484 @@ class ConfigurationCacheGradlePropertiesIntegrationTest extends AbstractConfigur
...
+        assertHasUnsupportedPropertyValueType(Property, Configuration, "beanHolder", "BeanHolderTask")
+    }
+
+    def "reports sensible error when Property of unsupported type is captured without custom task"() {
+        buildFile << """
+            interface ConfHolder {
+                Property<Configuration> getConf()
+            }
+
+            def holder = objects.newInstance(ConfHolder)
```
mlopatkin: :nail_care: I'd move this into `named` clause to avoid capturing the script instance.

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/WideningTypeSerializationFailureHelper.kt
```diff
@@ -0,0 +1,165 @@
...
+    }
+    onProblem(
+        PropertyProblem(
+            trace = trace,
+            message = message,
+            exception = exception,
+            stackTracingFailure = failureFactory.create(exception)
```
mlopatkin: :question: This adds a stacktrace to the report, but it is rather useless there - it is full of Gradle internals. We can get away with only specifying the exception - then it will be attached to the build failure (and will provide advices), but will not end up in the report.

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/WideningTypeSerializationFailureHelper.kt
```diff
@@ -0,0 +1,165 @@
...
+ * the call site instead of conflating it with the problem-reporting helper.
+ */
+suspend fun WriteContext.reportSerializationProblem(exception: Exception) {
+    val failureFactory = ownerService<FailureFactory>()
+    val message = StructuredMessage.build {
+        text("failed to serialize value of ")
+        reference(trace.toString())
+    }
+    onProblem(
+        PropertyProblem(
```
mlopatkin, line 92: :x: We're losing the documentation link in the report. We should at least keep pointing at `RequirementsDisallowedTypes`

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/WideningTypeSerializationFailureHelper.kt
```diff
@@ -0,0 +1,165 @@
...
+ * build later fails due to deferred problems, and its
+ * [resolutions][org.gradle.internal.exceptions.ResolutionProvider]
+ * remain visible to `failure.assertHasResolution(...)`.
+ *
+ * Callers who need a new trace frame (e.g., bean-field encoders that have
+ * not yet entered the field's property trace) should wrap with
+ * [withPropertyTrace] themselves; this keeps the trace surface explicit at
+ * the call site instead of conflating it with the problem-reporting helper.
+ */
+suspend fun WriteContext.reportSerializationProblem(exception: Exception) {
```
mlopatkin: :nail_care: we can omit `suspend` here.

platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/KotlinDelegateInspector.kt
```diff
@@ -0,0 +1,190 @@
...
+     */
+    fun kotlinPropertyGetterReturnType(delegateField: Field): Class<*> {
+        val propertyName = delegateField.name.removeSuffix("\$delegate")
+        if (propertyName == delegateField.name) {
+            throw DelegateInspectionException(
+                "Field '${delegateField.name}' on ${delegateField.declaringClass.name} " +
+                    "does not follow the Kotlin delegate naming convention (<name>\$delegate)."
+            )
+        }
+        val getterName = "get${propertyName.replaceFirstChar { it.uppercase() }}"
```
mlopatkin: :x: this heuristic isn't correct, at least not for boolean properties. E.g. `val isReady by lazy {}` generates 
```
private final kotlin.Lazy isReady$delegate;
public final boolean isReady();
```

I suggest we do a quick check for "are there any `$delegate` fields" and then use `kotlin-reflect` to interrogate them.

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/WideningTypeSerializationFailureHelper.kt
```diff
@@ -0,0 +1,165 @@
...
+    // The helper has already rejected the case where field.type is a supertype of
+    // decodedType (load-side value fits the field). This branch handles the OPPOSITE
+    // subtype relation: field.type is a subtype of decodedType — the codec may produce
+    // a concrete instance of that subtype at runtime (codecs declare a broad interface
+    // but generally construct via a factory that yields the expected concrete class).
+    // Only flag when the types share no subtyping relation at all — then reassignment
+    // is definitely impossible.
+    if (widening.decodedType.isAssignableFrom(field.type)) return false
+    val exception = UnsupportedTypeException(
+        "Cannot serialize value of type ${fieldValue.javaClass.name} into field " +
```
mlopatkin: :thinking: I really like how descriptive the message is, but delivering it through exception message leaves much to be desired in the presentation department. It is buried in the report, and the console output only lists the cause under the summary (`1 problem found`), where it becomes somewhat detached. Especially when there are multiple problems.
```

* What went wrong:
Configuration cache problems found in this build.

- Task `:listConfTask` of type `ListConfTask`: failed to serialize value of 'field `__confs__` of task `:listConfTask` of type `ListConfTask`'

See the complete report at file:///home/user/Projects/gradle/build/reports/configuration-cache/1nqkvnjp46job5cgjf2lumv97/ba2dhs9rb1i2i83piezwjyoz0/configuration-cache-report.html
> Cannot serialize ListProperty<Configuration> in task :listConfTask of type ListConfTask. The value type of this property (org.gradle.api.artifacts.Configuration) is not supported with the configuration cache: its codec produces org.gradle.api.file.FileCollection on load.
```
Can we fold this information into the problem message somehow?

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/WideningTypeSerializationFailureHelper.kt
```diff
@@ -0,0 +1,165 @@
...
+            "${field.name} of ${field.declaringClass.name} in ${trace.taskDescription()}: " +
+            "its codec produces ${widening.publicDecodedType.name} on load, " +
```
mlopatkin: :x: I don't think we should talk about codecs in user-facing messages.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/ConfigurationCodec.kt
```diff
@@ -0,0 +1,51 @@
...
+    artifactSetConverter: ArtifactSetToFileCollectionFactory,
+    taskDependencyFactory: TaskDependencyFactory
+) : FileCollectionCodec(fileCollectionFactory, artifactSetConverter, taskDependencyFactory),
+    WideningCodec<FileCollectionInternal> {
+
+    override val decodedType: Class<FileCollectionInternal> = FileCollectionInternal::class.java
+
+    override val publicDecodedType: Class<*> = FileCollection::class.java
+
+    override val wideningFix: String = "Use a ConfigurableFileCollection instead."
```
mlopatkin: :thinking: It might not always be the best solution, e.g. for stuff captured in the lambda. We can add something like "Or change the captured type to FileCollection".

platforms/core-configuration/model-reflect/src/main/java/org/gradle/internal/reflect/UnsupportedTypeException.java
```diff
@@ -0,0 +1,49 @@
...
+        super(message);
+        this.resolutions = Collections.unmodifiableList(resolutions);
+    }
+
+    public UnsupportedTypeException(String message, Throwable cause) {
+        super(message, cause);
+        this.resolutions = Collections.emptyList();
+    }
+
+    @NonNull
```
mlopatkin, line 53: :nail_care: No need to add NonNull.

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/CodecLookup.kt
```diff
@@ -0,0 +1,36 @@
...
+ * dependencies on `.graph.codecs`.
+ */
+interface CodecLookup {
+    /**
+     * Returns the encoding registered for the given runtime [type], or null
+     * when none matches. The returned value is typed as `Any?` at this layer
+     * so callers cast to concrete codec interfaces (for example
+     * `WideningCodec<*>`) without forcing this interface to depend on them.
+     */
+    fun encodingForType(type: Class<*>): Any?
```
mlopatkin: :nail_care: Can the returned type be at lease `EncodingProvider<*>?` instead of `Any?`

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGradlePropertiesIntegrationTest.groovy
```diff
@@ -743,4 +748,484 @@ class ConfigurationCacheGradlePropertiesIntegrationTest extends AbstractConfigur
...
+
+        when:
+        configurationCacheFails "multiConf"
+
+        then:
+        assertHasUnsupportedPropertyValueType(Property, Configuration, "multiConf", "MultiConfTask")
+    }
+
+    def "unset Property of unsupported type does not cause an error"() {
+        // An unset abstract managed property's backing field is null (lazy initialization never triggered).
```
mlopatkin: :x: while this test passes due to the stated reason, it is a very weak form of "allow unset `Property<Configuration>` as a compatibility measure". The implementation doesn't work for e.g. `Property<Configuration> conf = objects.property(Configuration)` or if anyone calls `getConf` for whatever reason.

platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/KotlinDelegateInspector.kt
```diff
@@ -0,0 +1,190 @@
...
+
+package org.gradle.internal.serialize.beans.services
+
+import java.lang.reflect.Field
+import kotlin.properties.ReadOnlyProperty
+import kotlin.properties.ReadWriteProperty
+
+
+/**
+ * Reflective utilities for inspecting Kotlin property delegates ahead of
```
mlopatkin, line 28: :thinking: TBH, I don't like our deep reflection over delegates. AFAICT, we only need the value to gate on null/non-null (hard failure vs. deprecation); if we ditch that, we can just complain as soon as we see the type of the delegated field.

I suggest we limit ourselves to `Lazy` handling for now, and obtain the lazy value regardless of whether it is available - we're going to compute it anyway to store the value.

We can still emit deprecations for other delegates in the future, effectively making
`foo: Configuration by <something>` eventually forbidden.

mlopatkin (changes requested):
I'm sorry for taking so long to look through this. I don't think we can merge the PR in its current state. My primary concerns are:
- The rendering of failures.
- The handling of empty properties.
- The handling of delegates.

We talked about emitting deprecations instead of problems for empty fields/properties, and I don't see it in the PR, but this is something we can add later.

platforms/core-configuration/model-reflect/src/main/java/org/gradle/internal/reflect/UnsupportedTypeException.java
tresat: It silences an IDE warning: "Not annotated method overrides method annotated with @NullMarked".

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/WideningTypeSerializationFailureHelper.kt
tresat: I put it into the Problem output.  Let me know how you think the message looks now:

```
* What went wrong:
Configuration cache problems found in this build.

- Task `:printFiles` of type `BrokenPrintTask`: failed to serialize value of field '__conf__' of task ':printFiles' of type 'BrokenPrintTask'
  Cannot serialize Property<Configuration> in task :printFiles of type BrokenPrintTask.
  The value type of this property (org.gradle.api.artifacts.Configuration) is not supported with the configuration cache: values of this type are restored from the configuration cache as org.gradle.api.file.FileCollection.
  See https://docs.gradle.org/9.7.0-19700101000000+0000/userguide/configuration_cache_requirements.html#config_cache:requirements:disallowed_types

See the complete report at file:///Users/ttresansky/Projects/gradle-worktrees/master-worktrees/tt/96/value-type-serialization-errors/platforms/core-configuration/configuration-cache/build/tmp/te%C5%9Dt%20files/Configurati.Test/ddgd4/build/reports/configuration-cache/b56j4yvb7dswn45fnma1uollw/b4hts3xwz9bvj7k840hws1yst/configuration-cache-report.html
```

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheGradlePropertiesIntegrationTest.groovy
tresat: Yeah, I don't know why I thought this was important.  It certainly isn't working.  And if there are multiple problems, and we can display them all, I think we should just do that.

I'll reverse this misleading test to verify ALL are shown, with proper asserts.

platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/KotlinDelegateInspector.kt
tresat: I understand wanting to avoid so much brittle custom reflection code.  I tried to use kotlin-reflect to do this work in order to reduce the code and increase the reliability.  I left the functionality the same, because I still think it is useful and would prefer to keep it.  Please take a look and see what you think.

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
The ShadowJar plugin is failing its own smoke test, and calling the Gradle build smoke tests to fail.  It does this:

```
val configurations: SetProperty<Configuration>
```

[here](https://github.com/GradleUp/shadow/blob/c38115710eacaf3b914dc07016dc78c618dd815e/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt#L180) which fails the new widening check.

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/KotlinDelegateInspector.kt
```diff
@@ -0,0 +1,178 @@
...
+        else -> throw DelegateInspectionException(
+            "Not a recognised Kotlin property delegate: ${delegate::class.java.name}. " +
+                "Callers must guard with isKotlinDelegate() before calling extractValue()."
+        )
+    }
+
+    /**
+     * Returns `true` when [value] is a recognised Kotlin property delegate type.
+     */
+    fun isKotlinDelegate(value: Any?): Boolean =
```
mlopatkin, line 92: 💅 As we only recognize a subset of the delegates, it makes sense to highlight that in the name:
```suggestion
    fun isSupportedKotlinDelegate(value: Any?): Boolean =
```

platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/KotlinDelegateInspector.kt
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    /**
+     * Returns the declared return type of the Kotlin property backed by the given
+     * `$delegate` field.
+     *
+     * Uses [kotlin-reflect][kotlin.reflect] to interrogate the declaring class,
+     * which correctly handles Kotlin's JVM naming conventions — notably boolean
+     * `val isReady by lazy {}` compiles to a getter named `isReady()` (not
+     * `getIsReady()`) and would defeat any name-based getter lookup.
```
mlopatkin, line 118: 💅 These are implementation details and do not belong to the kdoc

platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/BeanPropertyWriter.kt
```diff
@@ -49,15 +64,62 @@ class BeanPropertyWriter(
...
     }
 
+    /**
+     * Inspects a Kotlin property delegate field's value for an unsupported
+     * roundtrip into the property's declared getter return type.
+     *
+     * Uninitialized `by lazy` delegates are deliberately skipped: there is no
+     * value yet to type-check, and the underlying Lazy codec forces the lazy
+     * via `writeReplace` during the subsequent encode pass, so the type
+     * concern is moot — the value that ends up in the cache is the forced one.
```
mlopatkin, line 84: ❓ I still don't understand why we skip. The value will be computed and stored nevertheless, and may be widened. But instead of a useful problem we'll get a ClassCastException at the use site, which is what we're trying to eliminate in the first place.

Something like:
```kotlin
abstract class FooTask : DefaultTask() {
  @get:Internal
  val foo: Configuration by lazy { project.configurations.create("x") }


   @TaskAction fun action() {
     println (foo.name)
   }
}


tasks.register<FooTask>("foo") {}
```

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/SerializedLambdaParametersCheckingCodec.kt
```diff
@@ -101,19 +104,45 @@ class SerializedLambdaParametersCheckingCodec(
...
+     * has no runtime type-assignment fallback: if the codec produces the
+     * supertype where the lambda body expects the subtype, the call site
+     * fails with a `ClassCastException` that is much harder to diagnose than
+     * a store-time rejection.
+     */
+    private
+    fun WriteContext.unsupportedWidenedTypeFor(paramType: Type): KClass<*>? {
+        if (paramType.sort != Type.OBJECT) return null
+        val paramClass = try {
+            Class.forName(paramType.className, false, this.javaClass.classLoader)
```
mlopatkin, line 138: :nail_care: We actually have the parameter value itself. It is probably more reliable to use its classloader - as the actual value's type is going to be a subtype of the captured type, the captured type should be reachable.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/JavaRecordCodec.kt
```diff
@@ -96,14 +94,31 @@ class JavaRecordEncoding(
...
+     * because the codec that handles [fieldValue] declares a decoded type that
+     * the field cannot accept. Returns null when the roundtrip is compatible.
+     */
+    private fun WriteContext.unsupportedWidenedTypeFor(field: Field, fieldValue: Any?): KClass<*>? {
+        val lookupType = fieldValue?.javaClass ?: field.type
+        val widening = findCodecThatWidensIncompatibly(field.type, lookupType) ?: return null
+        // The helper has already rejected the case where field.type is a supertype of
+        // decodedType. This branch handles the OPPOSITE subtype relation: field.type
+        // is a subtype of decodedType — the codec may produce a concrete instance of
+        // that subtype at runtime.
```
mlopatkin, line 118: :thinking: This can benefit from a concrete example.

platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/BeanPropertyWriter.kt
```diff
@@ -49,15 +64,62 @@ class BeanPropertyWriter(
                     null -> field.get(bean)
                     else -> conventionValueOf(bean, field, isExplicitValue)
                 }
-            relevantField.unsupportedFieldType?.let {
-                reportUnsupportedFieldType(it, "serialize", fieldName, fieldValue)
-            }
+            val dropped =
+                reportIfIncompatibleRoundtrip(field, fieldName, fieldValue) ||
```
mlopatkin, line 68: :question: It looks like we're now allowing `null` values for unsupported fields without emitting a problem?

platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/BeanPropertyWriter.kt
```diff
@@ -49,15 +64,62 @@ class BeanPropertyWriter(
...
+     * without forcing.
+     *
+     * @return `true` when the field's value must be dropped from the cache.
+     */
+    private
+    suspend fun WriteContext.reportIfUnsupportedKotlinDelegate(field: Field, fieldName: String, fieldValue: Any?): Boolean {
+        // A Kotlin `by`-delegate is identified by BOTH the field name (compiler emits `<name>$delegate`)
+        // AND the value type. A regular field declared as `Lazy<T>` is not a delegate even though its
+        // value satisfies isKotlinDelegate(); skip it so normal codec-driven serialization handles it.
+        if (!field.name.endsWith("\$delegate") || !KotlinDelegateInspector.isKotlinDelegate(fieldValue)) return false
```
mlopatkin, line 98: :nail_care: Can we fold this check into KotlinDelegateInspector? Now we're spilling the knowledge of delegate implementation internals across two classes - repeated `$delegate` suggests that.

mlopatkin (changes requested):
I'm afraid, the Shadow's failure indicates that we should adjust our approach. I underestimated how relaxed our checks in the `Property` land are - rather than sanitizing the restored value to conform to the expected type, we accept basically anything and let the user figure that out when they get CCE upon `get()`.

Restricting this would be a breaking change, so - not an easy decision - we'll have to just deprecate for now. 

This is my summary of the changes (`Configuration` implies any widened type):
- plain `Configuration` field of a POJO - problem in `master`, problem if non-null in the PR -> should remain a problem regardless of nullness.
- `Configuration` field of a Record - problem in `master`, problem in the PR -> let's keep it.
- `Configuration` captured by a lambda - problem in `master`, problem in the PR -> let's keep it.
- Delegate backed by `Configuration` - nothing in `master`, problem in the PR -> relax to deprecation.
- non-initialized `Configuration by lazy {}` - nothing in `master`, nothing in the PR -> promote to deprecation.
- `Property<Configuration>` and collection properties - nothing in `master`, problem if non-null/non-empty in the PR -> relax to deprecation regardless of nullness/emptiness.


I can pick this PR up as part of the Cortex team's platform work if you don't have bandwidth to move it forward.

tresat:
> I'm afraid, the Shadow's failure indicates that we should adjust our approach. I underestimated how relaxed our checks in the `Property` land are - rather than sanitizing the restored value to conform to the expected type, we accept basically anything and let the user figure that out when they get CCE upon `get()`.
> 
> Restricting this would be a breaking change, so - not an easy decision - we'll have to just deprecate for now.
> 
> This is my summary of the changes (`Configuration` implies any widened type):
> 
> * plain `Configuration` field of a POJO - problem in `master`, problem if non-null in the PR -> should remain a problem regardless of nullness.
> * `Configuration` field of a Record - problem in `master`, problem in the PR -> let's keep it.
> * `Configuration` captured by a lambda - problem in `master`, problem in the PR -> let's keep it.
> * Delegate backed by `Configuration` - nothing in `master`, problem in the PR -> relax to deprecation.
> * non-initialized `Configuration by lazy {}` - nothing in `master`, nothing in the PR -> promote to deprecation.
> * `Property<Configuration>` and collection properties - nothing in `master`, problem if non-null/non-empty in the PR -> relax to deprecation regardless of nullness/emptiness.
> 
> I can pick this PR up as part of the Cortex team's platform work if you don't have bandwidth to move it forward.

@mlopatkin Thanks for looking.  I probably won't be able to return to this for a while, so if Cortex can finish it, that's fine with me.

big-guy:
@mlopatkin do you intend to pick this up for 9.8 (IOW, this week)?

PR #37788, opened 2026-05-04
Ensure constraints attach to replaced node

jvandort:
A constraint by definition attaches to all nodes in the target component, as a constraint contains the version of a module and therfore does not involve itself with selecting individual variants.

However, when a constraint targets a module and a node of a component of that module loses a capability conflict, the node in that module is not selected. The node has a replacement that is selected.

Therefore, we ensure we look for replacements of nodes that we attach constraints to so they properly follow node replacements

Fixes failures found in: https://github.com/gradlex-org/jvm-dependency-conflict-resolution/pull/403

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

ljacomet (commented):
Approved 😄

PR #37790, opened 2026-05-04
Add archive task file timestamp property

### LINKED ISSUE #14819: allow to configure timestamp used for preserveFileTimestamps
### Expected Behavior

preserveFileTimestamps = false
fixedTimestamp = <users expected stable timestamp, i.e. UTC time from last git commit>

### Current Behavior

current behavior all files are set to 1980ies date/time.

### Context

I use reproducible builds in jbang but users don't like that all releases has a timestamp back in 1980.

Would be nice being able to just state that the timestamp should be the time of last commit so the timestamp is still fixed between builds; but at least reflect reality/expectations better.

nicklauslittle-gov:
Fixes #14819. (See issue for related issues and discussions.)

This change adds a new `fileTimestamp` property to archive tasks. If specified, this property will set all timestamps in the archive to the specified value. If not specified, existing functionality is retained, and timestamps are determined by the `preserveFileTimestamps` property. Integration tests have been added to [ReproducibleArchivesIntegrationTest.groovy](https://github.com/gradle/gradle/compare/master...nicklauslittle-gov:gradle:add-archive-timestamp) to check new functionality, and the documentation has been updated.

### DIFF
--- a/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/TarCopyAction.java
+++ b/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/TarCopyAction.java
@@ -35,6 +35,7 @@ import org.gradle.internal.IoActions;
 import java.io.File;
 import java.io.OutputStream;
 import java.util.OptionalLong;
+import java.util.function.LongUnaryOperator;
 
 public class TarCopyAction implements CopyAction {
     /**
@@ -49,12 +50,18 @@ public class TarCopyAction implements CopyAction {
 
     private final File tarFile;
     private final ArchiveOutputStreamFactory compressor;
-    private final OptionalLong fileTimestampMillis;
+    private final OptionalLong reproducibleFileTimestampMs;
 
     public TarCopyAction(File tarFile, ArchiveOutputStreamFactory compressor, boolean preserveFileTimestamps, Provider<Long> reproducibleFileTimestamp) {
         this.tarFile = tarFile;
         this.compressor = compressor;
-        this.fileTimestampMillis = CopyActionUtil.computeReproducibleTimestamp(preserveFileTimestamps, reproducibleFileTimestamp, CONSTANT_TIME_FOR_TAR_ENTRIES);
+        this.reproducibleFileTimestampMs = CopyActionUtil.computeReproducibleTimestamp(
+            preserveFileTimestamps,
+            reproducibleFileTimestamp,
+            LongUnaryOperator.identity(),
+            CONSTANT_TIME_FOR_TAR_ENTRIES,
+            Long.MAX_VALUE
+        );
     }
 
     @Override
@@ -136,6 +143,6 @@ public class TarCopyAction implements CopyAction {
     }
 
     private long getArchiveTimeFor(FileCopyDetails details) {
-        return fileTimestampMillis.orElseGet(details::getLastModified);
+        return reproducibleFileTimestampMs.orElseGet(details::getLastModified);
     }
 }
--- a/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/TarCopyAction.java
+++ b/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/TarCopyAction.java
@@ -24,7 +24,9 @@ import org.gradle.api.internal.file.CopyActionProcessingStreamAction;
 import org.gradle.api.internal.file.archive.compression.ArchiveOutputStreamFactory;
 import org.gradle.api.internal.file.copy.CopyAction;
 import org.gradle.api.internal.file.copy.CopyActionProcessingStream;
+import org.gradle.api.internal.file.copy.CopyActionUtil;
 import org.gradle.api.internal.file.copy.FileCopyDetailsInternal;
+import org.gradle.api.provider.Provider;
 import org.gradle.api.tasks.WorkResult;
 import org.gradle.api.tasks.WorkResults;
 import org.gradle.internal.ErroringAction;
@@ -32,6 +34,7 @@ import org.gradle.internal.IoActions;
 
 import java.io.File;
 import java.io.OutputStream;
+import java.util.OptionalLong;
 
 public class TarCopyAction implements CopyAction {
     /**
@@ -46,12 +49,12 @@ public class TarCopyAction implements CopyAction {
 
     private final File tarFile;
     private final ArchiveOutputStreamFactory compressor;
-    private final boolean preserveFileTimestamps;
+    private final OptionalLong fileTimestampMillis;
 
-    public TarCopyAction(File tarFile, ArchiveOutputStreamFactory compressor, boolean preserveFileTimestamps) {
+    public TarCopyAction(File tarFile, ArchiveOutputStreamFactory compressor, boolean preserveFileTimestamps, Provider<Long> reproducibleFileTimestamp) {
         this.tarFile = tarFile;
         this.compressor = compressor;
-        this.preserveFileTimestamps = preserveFileTimestamps;
+        this.fileTimestampMillis = CopyActionUtil.computeReproducibleTimestamp(preserveFileTimestamps, reproducibleFileTimestamp, CONSTANT_TIME_FOR_TAR_ENTRIES);
     }
 
     @Override
@@ -133,6 +136,6 @@ public class TarCopyAction implements CopyAction {
     }
 
     private long getArchiveTimeFor(FileCopyDetails details) {
-        return preserveFileTimestamps ? details.getLastModified() : CONSTANT_TIME_FOR_TAR_ENTRIES;
+        return fileTimestampMillis.orElseGet(details::getLastModified);
     }
 }
--- a/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
+++ b/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
@@ -25,13 +25,18 @@ import org.gradle.api.internal.DocumentationRegistry;
 import org.gradle.api.internal.file.CopyActionProcessingStreamAction;
 import org.gradle.api.internal.file.copy.CopyAction;
 import org.gradle.api.internal.file.copy.CopyActionProcessingStream;
+import org.gradle.api.internal.file.copy.CopyActionUtil;
 import org.gradle.api.internal.file.copy.FileCopyDetailsInternal;
 import org.gradle.api.internal.file.copy.ZipCompressor;
+import org.gradle.api.provider.Provider;
 import org.gradle.api.tasks.WorkResult;
 import org.gradle.api.tasks.WorkResults;
 import org.gradle.internal.IoActions;
 
 import java.io.File;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.util.OptionalLong;
 
 public class ZipCopyAction implements CopyAction {
 
@@ -39,14 +44,17 @@ public class ZipCopyAction implements CopyAction {
     private final ZipCompressor compressor;
     private final DocumentationRegistry documentationRegistry;
     private final String encoding;
-    private final boolean preserveFileTimestamps;
+    private final OptionalLong fileTimestampMillis;
 
-    public ZipCopyAction(File zipFile, ZipCompressor compressor, DocumentationRegistry documentationRegistry, String encoding, boolean preserveFileTimestamps) {
+    public ZipCopyAction(File zipFile, ZipCompressor compressor, DocumentationRegistry documentationRegistry, String encoding, boolean preserveFileTimestamps, Provider<Long> reproducibleFileTimestamp) {
         this.zipFile = zipFile;
         this.compressor = compressor;
         this.documentationRegistry = documentationRegistry;
         this.encoding = encoding;
-        this.preserveFileTimestamps = preserveFileTimestamps;
+        this.fileTimestampMillis = CopyActionUtil.computeReproducibleTimestamp(preserveFileTimestamps, reproducibleFileTimestamp.map(ms -> {
+            final Instant utcTime = Instant.ofEpochMilli(ms);
+            return utcTime.minusSeconds(ZoneId.systemDefault().getRules().getOffset(utcTime).getTotalSeconds()).toEpochMilli();
+        }), ZipEntryConstants.CONSTANT_TIME_FOR_ZIP_ENTRIES);
     }
 
     @Override
@@ -124,6 +132,6 @@ public class ZipCopyAction implements CopyAction {
     }
 
     private long getArchiveTimeFor(FileCopyDetails details) {
-        return preserveFileTimestamps ? details.getLastModified() : ZipEntryConstants.CONSTANT_TIME_FOR_ZIP_ENTRIES;
+        return fileTimestampMillis.orElseGet(details::getLastModified);
     }
 }
--- a/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
+++ b/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
@@ -35,7 +35,9 @@ import org.gradle.internal.IoActions;
 
 import java.io.File;
 import java.time.Instant;
+import java.time.LocalDateTime;
 import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.OptionalLong;
 
 public class ZipCopyAction implements CopyAction {
@@ -51,10 +53,23 @@ public class ZipCopyAction implements CopyAction {
         this.compressor = compressor;
         this.documentationRegistry = documentationRegistry;
         this.encoding = encoding;
-        this.fileTimestampMillis = CopyActionUtil.computeReproducibleTimestamp(preserveFileTimestamps, reproducibleFileTimestamp.map(ms -> {
-            final Instant utcTime = Instant.ofEpochMilli(ms);
-            return utcTime.minusSeconds(ZoneId.systemDefault().getRules().getOffset(utcTime).getTotalSeconds()).toEpochMilli();
-        }), ZipEntryConstants.CONSTANT_TIME_FOR_ZIP_ENTRIES);
+        this.fileTimestampMillis = CopyActionUtil.computeReproducibleTimestamp(
+            preserveFileTimestamps,
+            reproducibleFileTimestamp.map(ZipCopyAction::normalizeTimestamp),
+            ZipEntryConstants.CONSTANT_TIME_FOR_ZIP_ENTRIES
+        );
+    }
+
+    /**
+     * Zip stores entry times as MS-DOS date/time rendered in the default time zone (see ZipUtil.toDosTime),
+     * so shift the timestamp to make the stored fields equal its UTC wall clock, independent of the zone.
+     */
+    private static long normalizeTimestamp(long timestampMillis) {
+        Instant utcTime = Instant.ofEpochMilli(timestampMillis);
+        return LocalDateTime.ofInstant(utcTime, ZoneOffset.UTC)
+            .atZone(ZoneId.systemDefault())
+            .toInstant()
+            .toEpochMilli();
     }
 
     @Override
--- a/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
+++ b/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
@@ -20,6 +20,7 @@ import org.apache.commons.compress.archivers.zip.Zip64RequiredException;
 import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
 import org.apache.commons.compress.archivers.zip.ZipArchiveOutputStream;
 import org.gradle.api.GradleException;
+import org.gradle.api.InvalidUserDataException;
 import org.gradle.api.file.FileCopyDetails;
 import org.gradle.api.internal.DocumentationRegistry;
 import org.gradle.api.internal.file.CopyActionProcessingStreamAction;
@@ -55,7 +56,14 @@ public class ZipCopyAction implements CopyAction {
         this.encoding = encoding;
         this.fileTimestampMillis = CopyActionUtil.computeReproducibleTimestamp(
             preserveFileTimestamps,
-            reproducibleFileTimestamp.map(ZipCopyAction::normalizeTimestamp),
+            reproducibleFileTimestamp.map(ms -> {
+                if (ms > ZipEntryConstants.MAXIMUM_TIME_FOR_ZIP_ENTRIES) {
+                    throw new InvalidUserDataException(String.format(
+                        "The reproducible file timestamp %s is greater than the maximum timestamp %s supported for ZIP archives.",
+                        Instant.ofEpochMilli(ms), Instant.ofEpochMilli(ZipEntryConstants.MAXIMUM_TIME_FOR_ZIP_ENTRIES)));
+                }
+                return normalizeTimestamp(ms);
+            }),
             ZipEntryConstants.CONSTANT_TIME_FOR_ZIP_ENTRIES
         );
     }
--- a/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
+++ b/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
@@ -20,7 +20,6 @@ import org.apache.commons.compress.archivers.zip.Zip64RequiredException;
 import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
 import org.apache.commons.compress.archivers.zip.ZipArchiveOutputStream;
 import org.gradle.api.GradleException;
-import org.gradle.api.InvalidUserDataException;
 import org.gradle.api.file.FileCopyDetails;
 import org.gradle.api.internal.DocumentationRegistry;
 import org.gradle.api.internal.file.CopyActionProcessingStreamAction;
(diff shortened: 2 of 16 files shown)


ov7a:

nicklauslittle-gov:
Making updates to fix sanityCheck...

subprojects/core/src/main/java/org/gradle/api/tasks/bundling/AbstractArchiveTask.java
```diff
@@ -239,6 +243,20 @@ public CopySpec into(Object destPath, Action<? super CopySpec> copySpec) {
...
+     * If <code>null</code> each archive entry will use the value as determined by the <code>preserveFileTimestamps</code> property.
+     * If not <code>null</code> each archive entry will use the file timestamp, ignoring the <code>preserveFileTimestamps</code> property.
+     *
+     * @return the file timestamp used for each entry in the archive
+     * @since 9.6.0
+     */
+    @Incubating
+    @Input
+    @Optional
+    public abstract Property<Instant> getFileTimestamp();
```
asodja: ❌  Let's just use Long. `Instant` is "Java modern" but from DSL point of view it's clunky

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/TarCopyAction.java
```diff
@@ -46,12 +48,21 @@ public class TarCopyAction implements CopyAction {
...
+    private final OptionalLong fileTimestampMillis;
 
-    public TarCopyAction(File tarFile, ArchiveOutputStreamFactory compressor, boolean preserveFileTimestamps) {
+    public TarCopyAction(File tarFile, ArchiveOutputStreamFactory compressor, boolean preserveFileTimestamps, Instant fileTimestamp) {
         this.tarFile = tarFile;
         this.compressor = compressor;
-        this.preserveFileTimestamps = preserveFileTimestamps;
+
+        if (fileTimestamp != null) {
+            this.fileTimestampMillis = OptionalLong.of(fileTimestamp.toEpochMilli());
```
asodja: ❓ What if `fileTimestamp < CONSTANT_TIME_FOR_TAR_ENTRIES`?
We should handle that correctly. I believe CONSTANT_TIME_FOR_TAR_ENTRIES should be the minimum of timestamps.

We should document that min value will be whatever is for TAR and whatever is for ZIP, or alternatively we should fail if value is less than these two values

subprojects/core/src/main/java/org/gradle/api/tasks/bundling/AbstractArchiveTask.java
```diff
@@ -239,6 +243,20 @@ public CopySpec into(Object destPath, Action<? super CopySpec> copySpec) {
         return this;
     }
 
+    /**
+     * Specifies the file timestamp used for each entry in the archive.
+     * <p>
+     * If <code>null</code> each archive entry will use the value as determined by the <code>preserveFileTimestamps</code> property.
```
asodja: 💅 Instead of "If <code>null</code>" it would be better would to say: if not set, since Property is never null

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
```diff
@@ -32,21 +32,32 @@
...
-    public ZipCopyAction(File zipFile, ZipCompressor compressor, DocumentationRegistry documentationRegistry, String encoding, boolean preserveFileTimestamps) {
+    public ZipCopyAction(File zipFile, ZipCompressor compressor, DocumentationRegistry documentationRegistry, String encoding, boolean preserveFileTimestamps, Instant fileTimestamp) {
         this.zipFile = zipFile;
         this.compressor = compressor;
         this.documentationRegistry = documentationRegistry;
         this.encoding = encoding;
-        this.preserveFileTimestamps = preserveFileTimestamps;
+
+        if (fileTimestamp != null) {
+            this.fileTimestampMillis = OptionalLong.of(fileTimestamp.toEpochMilli());
```
asodja: Also here: 
❓ What if fileTimestamp < CONSTANT_TIME_FOR_ZIP_ENTRIES?
We should handle that correctly. I believe CONSTANT_TIME_FOR_ZIP_ENTRIES should be the minimum of timestamps.

We should document that min value will be whatever is for TAR and whatever is for ZIP, or alternatively we should fail if value is less than these two values.

subprojects/core/src/main/java/org/gradle/api/tasks/bundling/AbstractArchiveTask.java
```diff
@@ -239,6 +243,20 @@ public CopySpec into(Object destPath, Action<? super CopySpec> copySpec) {
...
+     * If not <code>null</code> each archive entry will use the file timestamp, ignoring the <code>preserveFileTimestamps</code> property.
+     *
+     * @return the file timestamp used for each entry in the archive
+     * @since 9.6.0
```
asodja: 💅 Let's change that to 9.7.0

subprojects/core/src/main/java/org/gradle/api/tasks/bundling/AbstractArchiveTask.java
asodja: ❌ Let's rename it to `getReproducibleFileTimestamp()` so it's more alligned with `isReproducibleFileOrder` and also fix Javadoc and add `reproducible`

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
```diff
@@ -32,21 +32,32 @@
...
+        } else {
+            this.fileTimestampMillis = preserveFileTimestamps ? OptionalLong.empty() : OptionalLong.of(ZipEntryConstants.CONSTANT_TIME_FOR_ZIP_ENTRIES);
+        }
+    }
+
+    public ZipCopyAction(File zipFile, ZipCompressor compressor, DocumentationRegistry documentationRegistry, String encoding, boolean preserveFileTimestamps) {
```
asodja: ❓ Why we need this constructor?

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/TarCopyAction.java
```diff
@@ -46,12 +48,21 @@ public class TarCopyAction implements CopyAction {
...
+        } else {
+            this.fileTimestampMillis = preserveFileTimestamps ? OptionalLong.empty() : OptionalLong.of(CONSTANT_TIME_FOR_TAR_ENTRIES);
+        }
+    }
+
+    public TarCopyAction(File tarFile, ArchiveOutputStreamFactory compressor, boolean preserveFileTimestamps) {
```
asodja: ❓ Why we need this constructor?

subprojects/core/src/main/java/org/gradle/api/tasks/bundling/AbstractArchiveTask.java
asodja, line 273: 🤔 I think it would be nice we would have some validation for misconfiguraiton. If `isPreserveFileTimestamps` is set to true and `getReproducibleFileTimestamp()` is set we should throw an error since users intent is not clear

asodja (changes requested):

lkasso (changes requested):
For documentation purposes, can you:
  1. Add a usage example in working_with_files.adoc showing `fileTimestamp` in a build script
  2. Add a release notes entry for the new incubating feature

nicklauslittle-gov:
Rebasing...

subprojects/core/src/main/java/org/gradle/api/tasks/bundling/AbstractArchiveTask.java
nicklauslittle-gov: I've made these changes. I do think `Instant` makes the API easier to use, but if `Long` is preferred, we can go with that.

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/TarCopyAction.java
nicklauslittle-gov: Updated to use `Math.max(reproducibleFileTimestamp.get(), CONSTANT_TIME_FOR_TAR_ENTRIES)`

subprojects/core/src/main/java/org/gradle/api/tasks/bundling/AbstractArchiveTask.java
nicklauslittle-gov: Updated.

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
nicklauslittle-gov: Updated to use `Math.max(reproducibleFileTimestamp.get(), ZipEntryConstants.CONSTANT_TIME_FOR_ZIP_ENTRIES)`.

subprojects/core/src/main/java/org/gradle/api/tasks/bundling/AbstractArchiveTask.java
nicklauslittle-gov: Done.

subprojects/core/src/main/java/org/gradle/api/tasks/bundling/AbstractArchiveTask.java
nicklauslittle-gov: Done.

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
nicklauslittle-gov: Removed constructor.

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/TarCopyAction.java
nicklauslittle-gov: Removed constructor.

subprojects/core/src/main/java/org/gradle/api/tasks/bundling/AbstractArchiveTask.java
nicklauslittle-gov: I added a check and an exception now gets thrown in this case. Let me know if I need to change the exception type or message.

nicklauslittle-gov:
> For documentation purposes, can you:
> 
> 1. Add a usage example in working_with_files.adoc showing `fileTimestamp` in a build script
> 2. Add a release notes entry for the new incubating feature

I've completed these, but I'm not sure if I did this correctly. Let me know if changes are needed.

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -79,6 +79,9 @@ Gradle provides an intuitive [command-line interface](userguide/command_line_int
 ### Build authoring improvements
 Gradle provides [rich APIs](userguide/getting_started_dev.html) for build engineers and plugin authors, enabling the creation of custom, reusable build logic and better maintainability.
 
+#### Archive tasks allow custom timestamp
```
lkasso: Can you update the release notes so they are a little more detailed, for example:

````
#### Custom timestamps for reproducible archives

Gradle produces [reproducible archives](userguide/working_with_files.html#sec:reproducible_archives) by default, using fixed timestamps for all entries. 
However, some environments, such as those following the [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/) specification, require a meaningful, verifiable timestamp rather than a fixed default.

Archive tasks now support a [`reproducibleFileTimestamp`](userguide/working_with_files.html#sec:reproducible_timestamp) property that lets you set a custom timestamp for every entry in the archive:

```kotlin
tasks.withType<AbstractArchiveTask>().configureEach {
    reproducibleFileTimestamp = java.time.Instant.ofEpochSecond(System.getenv("SOURCE_DATE_EPOCH").toLong()).toEpochMilli()
}
```

See the [Timestamp for files inside archives](userguide/working_with_files.html#sec:reproducible_timestamp) section in the Gradle User Manual for more details.
````

File: platforms/documentation/docs/src/docs/userguide/reference/gradle-types/working_with_files.adoc [hunk 9]
```diff
@@ -1090,6 +1091,18 @@ For such configuration see <<#sec:revert_reproducible_archives,Preserve file sys
 2. Case where the permissions of files and directories in the archive should be set to specific values, for example to set the executable bit on a script file.
 For such configuration see <<#sec:setting_file_permissions,Setting file permissions>> section.
 
+[[sec:reproducible_timestamp]]
+== Timestamp for files inside archives
+
+The timestamp for files inside archives can be specified. This can be used to complement reproducibility by using a meaningful, recent timestamp rather than the default.
```
lkasso:
```suggestion
The link:{groovyDslPath}/org.gradle.api.tasks.bundling.AbstractArchiveTask.html#org.gradle.api.tasks.bundling.AbstractArchiveTask:reproducibleFileTimestamp[`reproducibleFileTimestamp`] property allows you to set a custom timestamp for every entry in the archive.
When set, it takes precedence over the `preserveFileTimestamps` property.
When not set, timestamps are determined by `preserveFileTimestamps` as usual.

This can be used to complement reproducibility by using a meaningful, verifiable timestamp rather than the default.
```

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/working_with_files.adoc
```diff
@@ -1090,6 +1091,18 @@ For such configuration see <<#sec:revert_reproducible_archives,Preserve file sys
 2. Case where the permissions of files and directories in the archive should be set to specific values, for example to set the executable bit on a script file.
 For such configuration see <<#sec:setting_file_permissions,Setting file permissions>> section.
 
+[[sec:reproducible_timestamp]]
+== Timestamp for files inside archives
```
lkasso:
```suggestion
== Timestamp for files inside archives incubating-label:[]
```

lkasso (changes requested):
Some doc changes requested.
Also, for the new snippets, make sure they pass with: `./gradlew :docs:docsTest --tests '*archives*'`

platforms/documentation/docs/src/docs/release/notes.md
nicklauslittle-gov: Updated the release notes with more details.

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/working_with_files.adoc
nicklauslittle-gov: Updated this text. I changed some of it to match the code changes. (`reproducibleFileTimestamp` no longer overrides the `preserveFileTimestamps` property. It throws an exception if `preserveFileTimestamps` is `true` and `reproducibleFileTimestamp` is set.)

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/working_with_files.adoc
nicklauslittle-gov: Done.

nicklauslittle-gov:
> Some doc changes requested. Also, for the new snippets, make sure they pass with: `./gradlew :docs:docsTest --tests '*archives*'`

Thanks for the tip. The documentation tests now pass.

platforms/documentation/docs/src/snippets/integration-tests/files/archives/kotlin/build.gradle.kts
```diff
@@ -58,3 +60,16 @@ tasks.withType<AbstractArchiveTask>().configureEach {
...
+    isPreserveFileTimestamps = false // Reset to the default value
+}
+
+val sourceDateEpoch = 1248072216L
+
+// tag::reproducible-timestamp[]
+// import java.time.Instant
+
+tasks.withType<AbstractArchiveTask>().configureEach {
+    reproducibleFileTimestamp = Instant.ofEpochSecond(sourceDateEpoch).toEpochMilli()
```
nicklauslittle-gov: This is a little bit ugly. Let me know if you want me to change it to something like:

```suggestion
tasks.withType<AbstractArchiveTask>().configureEach {
    reproducibleFileTimestamp = sourceDateEpoch * 1000L
```

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/TarCopyAction.java
```diff
@@ -46,12 +49,19 @@ public class TarCopyAction implements CopyAction {
...
-    private final boolean preserveFileTimestamps;
+    private final OptionalLong fileTimestampMillis;
 
-    public TarCopyAction(File tarFile, ArchiveOutputStreamFactory compressor, boolean preserveFileTimestamps) {
+    public TarCopyAction(File tarFile, ArchiveOutputStreamFactory compressor, boolean preserveFileTimestamps, Provider<Long> reproducibleFileTimestamp) {
         this.tarFile = tarFile;
         this.compressor = compressor;
-        this.preserveFileTimestamps = preserveFileTimestamps;
+
+        if (!reproducibleFileTimestamp.isPresent()) {
```
asodja: 💅 I see this code is duplicated between TarCopyAction and ZipCopyAction. Could we extract it to some internal CopyActionUtil (or a similar class if there is already one available) ?

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -79,6 +79,21 @@ Gradle provides an intuitive [command-line interface](userguide/command_line_int
...
+#### Custom timestamps for reproducible archives
+
+Gradle produces [reproducible archives](userguide/working_with_files.html#sec:reproducible_archives) by default, using fixed timestamps for all entries.
+However, some environments, such as those following the [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/) specification, require a meaningful, verifiable timestamp rather than a fixed default.
+
+Archive tasks now support a [`reproducibleFileTimestamp`](userguide/working_with_files.html#sec:reproducible_timestamp) property that lets you set a custom timestamp for every entry in the archive:
+
+```kotlin
+tasks.withType<AbstractArchiveTask>().configureEach {
+    reproducibleFileTimestamp = java.time.Instant.ofEpochSecond(System.getenv("SOURCE_DATE_EPOCH").toLong()).toEpochMilli()
```
asodja: 💅 Note that better option is to use for CC friendliness:
```
reproducibleFileTimestamp = providers.systemProperty("SOURCE_DATE_EPOCH").map { 
    java.time.Instant.ofEpochSecond(it.toLong()).toEpochMilli()
}
```

platforms/documentation/docs/src/snippets/integration-tests/files/archives/kotlin/build.gradle.kts
asodja: I would not do any calculation, it's not essential, I would just do it in a variable, e.g. `
```
val fileTimestampMs = 1248072216 * 1000L

// tag::reproducible-timestamp[]
tasks.withType<AbstractArchiveTask>().configureEach {
    reproducibleFileTimestamp = fileTimestampMs
}

// end::reproducible-timestamp[]
```

asodja (changes requested):
I think we are very close to the finish line and the solution, nice work.

I need to just check if using any milliseconds is ok with Zip from reproducible point of view, or we should round it to an even seconds aka DOS timestamp (see this [comment](https://github.com/gradle/gradle/blob/8ea3fb514acd669e7d8bdab806e58f13efb30cde/platforms/core-runtime/files/src/main/java/org/gradle/api/internal/file/archive/ZipEntryConstants.java#L30-L35)). But I won't be able to check this today, so will have to wait for next week

platforms/documentation/docs/src/snippets/integration-tests/files/archives/kotlin/build.gradle.kts
asodja: Or alternatively we reuse the same example as we have in release notes

lkasso (approved):
Docs LGTM.

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/TarCopyAction.java
nicklauslittle-gov: Done.

platforms/documentation/docs/src/docs/release/notes.md
nicklauslittle-gov: I updated the release notes and the snippets to use the provider instead of a computed value.

platforms/documentation/docs/src/snippets/integration-tests/files/archives/kotlin/build.gradle.kts
nicklauslittle-gov: Updated the release notes example using the suggestion below and then used that in all the snippets.

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
```diff
@@ -25,28 +25,31 @@
...
+    private final OptionalLong fileTimestampMillis;
 
-    public ZipCopyAction(File zipFile, ZipCompressor compressor, DocumentationRegistry documentationRegistry, String encoding, boolean preserveFileTimestamps) {
+    public ZipCopyAction(File zipFile, ZipCompressor compressor, DocumentationRegistry documentationRegistry, String encoding, boolean preserveFileTimestamps, Provider<Long> reproducibleFileTimestamp) {
         this.zipFile = zipFile;
         this.compressor = compressor;
         this.documentationRegistry = documentationRegistry;
         this.encoding = encoding;
-        this.preserveFileTimestamps = preserveFileTimestamps;
+        this.fileTimestampMillis = CopyActionUtil.computeReproducibleTimestamp(preserveFileTimestamps, reproducibleFileTimestamp, ZipEntryConstants.CONSTANT_TIME_FOR_ZIP_ENTRIES);
```
asodja: 🤔 I did some research and ZIP stores time as timezone-local MS-DOS field, so for reproducibility we need to normalize the millis for ZIP (but not TAR), something like:

```
  private static final LocalDateTime MIN_ZIP_ENTRY_TIME = LocalDateTime.of(1980, 2, 1, 0, 0, 0);
  private static long toReproducibleZipFileTimestamp(long timeEpochMillis) {
      LocalDateTime timestamp = LocalDateTime
          .ofInstant(Instant.ofEpochMilli(timeEpochMillis), ZoneOffset.UTC)
          .withNano(0);
      timestamp = timestamp.withSecond(timestamp.getSecond() - (timestamp.getSecond() % 2));

      if (timestamp.isBefore(MIN_ZIP_ENTRY_TIME)) {                 
          return ZipEntryConstants.CONSTANT_TIME_FOR_ZIP_ENTRIES;
      }
      return timestamp.atZone(ZoneId.systemDefault()).toInstant().toEpochMilli();
  }
```

This should normalize any arbitrary epoch-millis into a timestamp that commons-compress should write identically on every machine, regardless of the build's timezone (we also floor seconds to the nearest even value, matching DOS time's 2-second resolution)

Would be possible to do this adjustment? 
It would be great to have also integration test but this may require more work, so you can also leave it to me

asodja (changes requested):

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java
nicklauslittle-gov: Yes, ZIP stores time with 2-second resolution. TAR has a similar limitation with 1-second resolution. Currently, this is handled automatically when writing the files, regardless of if timestamps are preserved from the original files or set to the reproducible timestamp value. (I've confirmed that modifying the ReproducibleArchivesIntegrationTest by adding 1ms to the timestamp results in the same hash for both ZIP and TAR, adding 1s to the timestamp results in just the TAR hashes failing, and adding 2s to the timestamp results in both ZIP and TAR hashes failing.) So, the timestamp may not be reversible from the archive file, but it's still deterministic and reproducible.

There are extensions for both ZIP and TAR that allow higher resolution timestamps, but Gradle doesn't use them currently. My preference would be to leave it as-is, since it already handles trimming down the timestamp to the appropriate resolution, and then, if Gradle chooses to use the higher resolution timestamp extensions in the future, the information would already be available.

Maybe the documentation just needs to be updated to state that archive formats have limitations when it comes to timestamp resolution, so the timestamp in the archive may be less precise than what's specified? Thoughts?

nicklauslittle-gov: As for the time zone information, there is no way to preserve it. The archives are reproducible because they store the UTC value. Anyone who unzips the content in any time zone other than UTC will end up with non-matching timestamps. This is just a limitation of the ZIP file format.

asodja: I think you are right about the time resolution, so I agree with that. So adding some docs would be nice. 

For timezone: From my tests it seems ZIP archives are not reproducible in different timezones. So basically if you use the same `reproducibleFileTimestamp` you get different zip hash, depending on the timezone you are running your build. 
TAR archives are reproducible. 

Check this test:
```
def "reproducible #taskName with reproducibleFileTimestamp is independent of the build timezone"() {
    given:
    files.each {
        file("src/${it}").text = it
    }
    buildFile << """
        task ${taskName}(type: ${taskType}) {
            reproducibleFileTimestamp = java.time.Instant.ofEpochSecond(1248072216).toEpochMilli()
            from 'src'
            destinationDirectory = buildDir
            archiveFileName = 'test.${fileExtension}'
            def layout = project.layout
            doFirst {
                java.util.TimeZone.setDefault(java.util.TimeZone.getTimeZone(System.getProperty("archiveBuildTimeZone")))
            }
            doLast {
                layout.buildDirectory.file("effective-tz.txt").get().asFile.text = java.util.TimeZone.getDefault().getID()
            }
        }
        """

    when: "the archive is built on machines in different timezones"
    def utc = runArchiveInTimeZone(taskName, fileExtension, "UTC")
    def kolkata = runArchiveInTimeZone(taskName, fileExtension, "Asia/Kolkata")

    then: "the timezone actually changed in the build JVM and the produced archives are identical"
    kolkata.tz == "Asia/Kolkata"
    utc.tz == "UTC"
    utc.hash == kolkata.hash

    where:
    taskName << ['zip', 'tar']
    taskType = taskName.capitalize() as String
    fileExtension = taskName as String
    files = ['DIR1/FILE11.txt', 'dir2/file22.txt', 'DIR3/file33.txt']
}

private Map<String, String> runArchiveInTimeZone(String taskName, String fileExtension, String timeZoneId) {
    executer.requireDaemon().requireIsolatedDaemons()
    executer.withArgument("-DarchiveBuildTimeZone=${timeZoneId}")
    succeeds(taskName, "--rerun")
    return [hash: file("build/test.${fileExtension}").md5Hash, tz: file("build/effective-tz.txt").text.trim()]
}
```  

Or am I missing something?

Note that `ZipEntryConstants.CONSTANT_TIME_FOR_ZIP_ENTRIES` works ok in production right now, since CONSTANT_TIME_FOR_ZIP_ENTRIES is calculated with whatever the default timezone is. It would also fail with this artifical test, but that is just because I hack timezone change via doFirst, which is not the case in production.

nicklauslittle-gov: Yes. I see now that the ZIP archive timestamps are based on the local time instead of UTC like I originally thought. It looks like the `atZone()` method on `Instant` doesn't actually offset the timestamp to local time. It just adds time zone information to the timestamp, so converting it back to an `Instant` results in the same value.

I changed this to offset the timestamp using the following code, which seems to work:

```java
final Instant utcTime = Instant.ofEpochMilli(ms);
return utcTime.minusSeconds(ZoneId.systemDefault().getRules().getOffset(utcTime).getTotalSeconds()).toEpochMilli();
```

I've added the integration test you posted above and added a note to the documentation. Let me know if there's anything else you want me to update. Thanks!

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

cobexer:

asodja:

asodja (approved):
LGTM

@nicklauslittle-gov thanks for your contribution!

PR #37794, opened 2026-05-05
Clean up some usages of ProjectInternal in CrossProjectModelAccess

octylFractal:
Mostly adjusts referrer to be ProjectIdentity, but also cleans up some target/base usages where trivially possible.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/.idea/inspectionProfiles/Gradle.xml
+++ b/.idea/inspectionProfiles/Gradle.xml
@@ -158,6 +158,14 @@
         <constraint name="receiver" nameOfExprType="InternalOptions" exprTypeWithinHierarchy="true" within="" contains="" />
         <constraint name="opt" nameOfExprType="InternalOption&lt;.*Nullable String&gt;" exprTypeWithinHierarchy="true" within="" contains="" />
       </replaceConfiguration>
+      <replaceConfiguration name="Use Path.asString() instead of toString()" suppressId="use_path_asString" problemDescriptor="Use `asString()` instead of `toString()` on `org.gradle.util.Path`" text="$receiver$.toString()" recursive="false" caseInsensitive="false" type="JAVA" pattern_context="default" reformatAccordingToStyle="true" shortenFQN="true" useStaticImport="true" replacement="$receiver$.asString()">
+        <constraint name="__context__" within="" contains="" />
+        <constraint name="receiver" nameOfExprType="org\.gradle\.util\.Path" exprTypeWithinHierarchy="true" within="" contains="" />
+      </replaceConfiguration>
+      <replaceConfiguration name="Use Path.asString() instead of toString() (Kotlin)" suppressId="use_path_asString" problemDescriptor="Use `asString()` instead of `toString()` on `org.gradle.util.Path`" text="$receiver$.toString()" recursive="false" caseInsensitive="false" type="Kotlin" pattern_context="default" reformatAccordingToStyle="true" shortenFQN="true" useStaticImport="true" replacement="$receiver$.asString()">
+        <constraint name="__context__" within="" contains="" />
+        <constraint name="receiver" nameOfExprType="org\.gradle\.util\.Path" exprTypeWithinHierarchy="true" within="" contains="" />
+      </replaceConfiguration>
       <replaceConfiguration name="Use @ToBeImplemented annotation" suppressId="not_yet_implemented" problemDescriptor="Prefer '@ToBeImplemented' annotation" text="@NotYetImplemented" recursive="false" caseInsensitive="false" type="Groovy" pattern_context="File" reformatAccordingToStyle="true" shortenFQN="true" useStaticImport="true" replacement="@org.gradle.util.internal.ToBeImplemented">
         <constraint name="__context__" within="" contains="" />
       </replaceConfiguration>
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingTaskExecutionGraph.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingTaskExecutionGraph.kt
@@ -18,12 +18,14 @@ package org.gradle.internal.cc.impl
 
 import groovy.lang.Closure
 import org.gradle.api.Action
-import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.execution.TaskExecutionGraph
 import org.gradle.api.execution.TaskExecutionGraphListener
+import org.gradle.api.internal.TaskInternal
 import org.gradle.api.internal.project.CrossProjectModelAccess
+import org.gradle.api.internal.project.ProjectIdentity
 import org.gradle.api.internal.project.ProjectInternal
+import org.gradle.api.internal.project.ProjectStateLookup
 import org.gradle.execution.plan.FinalizedExecutionPlan
 import org.gradle.execution.plan.ScheduledWork
 import org.gradle.execution.taskgraph.TaskExecutionGraphExecutionListener
@@ -38,10 +40,11 @@ import java.util.Objects
 internal
 class CrossProjectConfigurationReportingTaskExecutionGraph(
     taskGraph: TaskExecutionGraphInternal,
-    private val referrerProject: ProjectInternal,
+    private val referrerProject: ProjectIdentity,
     private val ipProblems: IsolatedProjectsProblemsReporter,
     private val crossProjectModelAccess: CrossProjectModelAccess,
     private val coupledProjectsListener: CoupledProjectsListener,
+    private val projectStateLookup: ProjectStateLookup,
 ) : TaskExecutionGraphInternal {
 
     private
@@ -80,10 +83,12 @@ class CrossProjectConfigurationReportingTaskExecutionGraph(
             if (task == null) {
                 // check whether the path refers to a different project
                 val parentPath = Path.path(path).parent?.asString()
-                if (parentPath != referrerProject.path) {
+                if (parentPath != referrerProject.projectPath.asString()) {
                     // even though the task was not found, the current project is coupled with the other project:
                     // if the configuration of that project changes, the result of this call might be different
-                    val coupledProjects = listOfNotNull(parentPath?.let { referrerProject.findProject(it) })
+                    val coupledProjects = listOfNotNull(parentPath?.let { p ->
+                        projectStateLookup.findProject(referrerProject.resolveProjectPath(p))?.identity
+                    })
                     reportCrossProjectTaskAccess(coupledProjects, path)
                 }
             } else {
@@ -98,8 +103,8 @@ class CrossProjectConfigurationReportingTaskExecutionGraph(
 
     private
     fun checkCrossProjectTaskAccess(task: Task) {
-        val taskOwner = task.project as ProjectInternal
-        if (!taskOwner.isReferrerProject) {
+        val taskOwner = (task as TaskInternal).taskIdentity.projectIdentity
+        if (taskOwner != referrerProject) {
             val coupledProjects = listOf(taskOwner)
             reportCrossProjectTaskAccess(coupledProjects, task.path)
         }
@@ -133,23 +138,21 @@ class CrossProjectConfigurationReportingTaskExecutionGraph(
     private
     fun observingTasksMaybeFromOtherProjects(tasks: Collection<Task>) {
         val otherProjects = tasks.mapNotNullTo(LinkedHashSet(tasks.size / 8)) { task ->
-            (task.project as? ProjectInternal)?.takeIf { project -> !project.isReferrerProject }
+            (task.project as ProjectInternal).owner.identity.takeIf {
+                it != referrerProject
+            }
         }
         reportCrossProjectTaskAccess(otherProjects)
     }
 
     private
-    val Project.isReferrerProject: Boolean
-        get() = this is ProjectInternal && identityPath == referrerProject.identityPath
-
-    private
-    fun reportCrossProjectTaskAccess(coupledProjects: Iterable<ProjectInternal>, requestPath: String? = null) {
+    fun reportCrossProjectTaskAccess(coupledProjects: Iterable<ProjectIdentity>, requestPath: String? = null) {
         reportCoupledProjects(coupledProjects)
 
         ipProblems.report {
             problem {
                 text("Project ")
-                reference(referrerProject.identityPath.toString())
+                reference(referrerProject.buildTreePath.toString())
                 text(" cannot access the tasks in the task graph that were created by other projects")
             }.exception { message ->
                 // As the exception message is not used for grouping, we can safely add the exact task name to it:
@@ -159,9 +162,9 @@ class CrossProjectConfigurationReportingTaskExecutionGraph(
     }
 
     private
-    fun reportCoupledProjects(coupledProjects: Iterable<ProjectInternal>) {
+    fun reportCoupledProjects(coupledProjects: Iterable<ProjectIdentity>) {
         coupledProjects.forEach { other ->
-            coupledProjectsListener.onProjectReference(referrerProject.owner, other.owner)
+            coupledProjectsListener.onProjectReference(referrerProject, other)
         }
     }
 
@@ -171,7 +174,7 @@ class CrossProjectConfigurationReportingTaskExecutionGraph(
     private
     class CrossProjectAccessTrackingTaskExecutionGraphListener(
         private val delegate: TaskExecutionGraphListener,
-        private val referrerProject: ProjectInternal,
+        private val referrerProject: ProjectIdentity,
         private val crossProjectModelAccess: CrossProjectModelAccess
     ) : TaskExecutionGraphListener {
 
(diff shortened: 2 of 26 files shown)


octylFractal:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectModelAccessTrackingParentDynamicObject.kt
```diff
@@ -129,17 +129,17 @@ class CrossProjectModelAccessTrackingParentDynamicObject(
...
                 text(" in the parent project ")
-                reference(ownerProject.identityPath.toString())
+                reference(ownerProject.buildTreePath.toString())
             }
                 .mapLocation { location ->
                     when (memberKind) {
                         MemberKind.PROPERTY -> {
                             if (memberName != null)
-                                PropertyTrace.Property(PropertyKind.PropertyUsage, memberName, PropertyTrace.Project(referrerProject.path, location))
+                                PropertyTrace.Property(PropertyKind.PropertyUsage, memberName, PropertyTrace.Project(referrerProject.projectPath.toString(), location))
```
alllex: ❌  Here and in many other places it should be `.asString()`

alllex: One thing we can do to provide Gradle developers with feedback on this is to introduce a structural-search-based warning. It's going to be in-IDE-only, but it's better than nothing.

Claude is very good at this, and this particular case is also very simple, compared to some warnings that we already have. PTAL:
* https://github.com/gradle/gradle/blob/d344da35819b887ce376a79b32148fd7626f44b1/.idea/inspectionProfiles/Gradle.xml

alllex: Additionally, I would have added a `reference(Path)` overload, since it seems to be a common need.

octylFractal: Done.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
```diff
@@ -580,13 +581,16 @@ class ProblemReportingCrossProjectModelAccess(
...
+            coupledProjectsListener.onProjectReference(referrer, delegate.projectIdentity)
             // Configure the target project, if it would normally be configured before the referring project
-            if (delegate < referrer && delegate.parent != null && buildModelParameters.isInvalidateCoupledProjects) {
+            if (isDelegateBeforeReferrer() && delegate.parent != null && buildModelParameters.isInvalidateCoupledProjects) {
                 delegate.owner.ensureConfigured()
             }
         }
 
+        private
+        fun isDelegateBeforeReferrer(): Boolean = ProjectOrderingUtil.compare(delegate.projectIdentity, referrer) < 0
```
alllex, line 597: 💅 It's a good idea. It's more readable as a function with a name. I'd improve it further by adding a clarifying comment that "before" refers to Vintage's configuration order (that defines total order among projects)

.idea/inspectionProfiles/Gradle.xml
```diff
@@ -158,6 +158,14 @@
         <constraint name="receiver" nameOfExprType="InternalOptions" exprTypeWithinHierarchy="true" within="" contains="" />
         <constraint name="opt" nameOfExprType="InternalOption&lt;.*Nullable String&gt;" exprTypeWithinHierarchy="true" within="" contains="" />
       </replaceConfiguration>
+      <replaceConfiguration name="Use Path.asString() instead of toString()" suppressId="use_path_asString" problemDescriptor="Use `asString()` instead of `toString()` on `org.gradle.util.Path`" text="$receiver$.toString()" recursive="false" caseInsensitive="false" type="JAVA" pattern_context="default" reformatAccordingToStyle="true" shortenFQN="true" useStaticImport="true" replacement="$receiver$.asString()">
```
alllex, line 161: 👍 This is very nice

alllex (approved):

.idea/inspectionProfiles/Gradle.xml
jvandort: Aren't we getting rid of this file?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingTaskExecutionGraph.kt
```diff
@@ -80,10 +81,12 @@ class CrossProjectConfigurationReportingTaskExecutionGraph(
...
                 val parentPath = Path.path(path).parent?.asString()
-                if (parentPath != referrerProject.path) {
+                if (parentPath != referrerProject.projectPath.asString()) {
                     // even though the task was not found, the current project is coupled with the other project:
                     // if the configuration of that project changes, the result of this call might be different
-                    val coupledProjects = listOfNotNull(parentPath?.let { referrerProject.findProject(it) })
+                    val coupledProjects = listOfNotNull(parentPath?.let { p ->
+                        crossProjectModelAccess.findProject(referrerProject, referrerProject, p)
+                    })
                     reportCrossProjectTaskAccess(coupledProjects, path)
```
jvandort, line 92: Seems we should be able to update this method to not require a Project instance if we update `reportCoupledProjects` to not accept a Project instance (it only requires its identity now) unless I'm missing something. 

This way we can avoid calling into `crossProjectModelAccess`

.idea/inspectionProfiles/Gradle.xml
octylFractal: No, that is `.idea/gradle.xml`, this is `.idea/inspectionProfiles/Gradle.xml`.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
```diff
@@ -92,66 +94,65 @@ class ProblemReportingCrossProjectModelAccess(
...
+        return CrossProjectConfigurationReportingTaskExecutionGraph(taskGraph, referrer, ipProblems, this, coupledProjectsListener)
     }
 
-    override fun parentProjectDynamicInheritedScope(referrerProject: ProjectInternal): HierarchicalDynamicObject? {
-        val parent = referrerProject.parent ?: return null
+    override fun parentProjectDynamicInheritedScope(referrer: ProjectState): HierarchicalDynamicObject? {
+        val parent = referrer.parent ?: return null
         return CrossProjectModelAccessTrackingParentDynamicObject(
-            parent, parent.inheritedScope, referrerProject, ipProblems, coupledProjectsListener
+            parent.identity, parent.fromMutableState { it.inheritedScope }, referrer.identity, ipProblems, coupledProjectsListener
```
jvandort: It is suspect that we leak the parent's dynamic object here. I would continue to use `getMutableModel` so we can make this leak more discoverable, or further, wire ProjectState of the parent into `CrossProjectModelAccessTrackingParentDynamicObject` so we do the "leaking" even closer to where it is accessed

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
```diff
@@ -580,13 +581,16 @@ class ProblemReportingCrossProjectModelAccess(
 
         private
         fun onProjectsCoupled() {
-            coupledProjectsListener.onProjectReference(referrer.owner, delegate.owner)
+            coupledProjectsListener.onProjectReference(referrer, delegate.projectIdentity)
             // Configure the target project, if it would normally be configured before the referring project
-            if (delegate < referrer && delegate.parent != null && buildModelParameters.isInvalidateCoupledProjects) {
+            if (isDelegateBeforeReferrer() && delegate.parent != null && buildModelParameters.isInvalidateCoupledProjects) {
```
jvandort, line 588: It seems strange to rely on the implicit ordering of vintage project configuration here. IMO we should just call `delegate.owner.ensureConfigured()` which should be idempotent

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
```diff
@@ -566,7 +567,7 @@ class ProblemReportingCrossProjectModelAccess(
                 // Referent is not configured and wasn't invalidated
                 // This can be a case in an incremental sync scenario, when referrer script was changed and since re-configured,
                 // but referent wasn't changed and since wasn't configured.
-                delegate < referrer && delegate.state.isUnconfigured && !buildModelParameters.isInvalidateCoupledProjects -> {
+                isDelegateBeforeReferrer() && delegate.state.isUnconfigured && !buildModelParameters.isInvalidateCoupledProjects -> {
```
jvandort, line 572: I'm not entirely sure whats going on here. Are we only reporting problems if there is a cross project access that is "looks backward" in the "normal" vintage project configuration order? This seems strange. Why isn't this uncionditional?

subprojects/core/src/main/java/org/gradle/api/internal/project/CrossProjectModelAccess.java
```diff
@@ -43,87 +43,99 @@ public interface CrossProjectModelAccess {
...
+
+    /**
+     * Locates the project with the given path, resolving relative paths against {@code base}.
+     *
+     * @param referrer The project from which the return value will be used.
+     * @param base The project against which a relative {@code path} is resolved.
+     * @param path A path to the requested project, either absolute or relative to {@code base}.
+     */
+    @Nullable
+    default ProjectInternal findProject(ProjectIdentity referrer, ProjectIdentity base, String path) {
```
jvandort: This method seems suspect. I don't necessarily want to "normalize" not knowing the absolute path of a project when you're trying to get its Project instance. 

Id rather us instead add a `ProjectIdentity.resolvePath(String)` or something like that so we don't need to have this API act as both a Project instance lookup API and a project path resolution API

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingTaskExecutionGraph.kt
```diff
@@ -149,7 +152,7 @@ class CrossProjectConfigurationReportingTaskExecutionGraph(
         ipProblems.report {
             problem {
                 text("Project ")
-                reference(referrerProject.identityPath.toString())
+                reference(referrerProject.buildTreePath.toString())
```
alllex: ❌ This still uses `toString()`

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
alllex: We are reporting problems in all branches, this case tries to be more precise, when coupled projects clustering is not involved.

I no longer recall all the details. We'd have to revisit the projects coupling tracking and the story in general.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
alllex: The `isInvalidateCoupledProjects` is enabled by default, and according to the docstring, we have it to be able to disable it for some benchmarking. So, normally we do call `ensureConfigured()` here.

Regarding the benchmarking, @6hundreds might have more details.

Either way, I wouldn't try and change this in the current PR.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingTaskExecutionGraph.kt
octylFractal: Done.

subprojects/core/src/main/java/org/gradle/api/internal/project/CrossProjectModelAccess.java
octylFractal: Done.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
octylFractal: So, I'll leave a `parent.mutableModel.inheritedScope` in `CrossProjectModelAccessTrackingParentDynamicObject`, then? I pushed that.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingTaskExecutionGraph.kt
```diff
@@ -133,23 +137,21 @@ class CrossProjectConfigurationReportingTaskExecutionGraph(
     private
     fun observingTasksMaybeFromOtherProjects(tasks: Collection<Task>) {
         val otherProjects = tasks.mapNotNullTo(LinkedHashSet(tasks.size / 8)) { task ->
-            (task.project as? ProjectInternal)?.takeIf { project -> !project.isReferrerProject }
+            (task.project as ProjectInternal).owner.identity.takeIf {
```
alllex, line 141:
```suggestion
            (task as TaskInternal).taskIdentity.projectIdentity.takeIf {
```

Avoids going through state when all we need is identities. We possibly should add `getProjectIdentity()` directly to `TaskInternal` to encourage this pattern.

subprojects/core-api/src/main/java/org/gradle/api/internal/project/ProjectIdentity.java
```diff
@@ -152,6 +152,18 @@ public String getProjectName() {
...
 
+    /**
+     * Given a potentially relative path to a project, resolve it to an absolute path within the build,
+     * using this project as the base for resolving relative paths. This is purely a project path operation,
+     * and does not use the build path in any way.
+     *
+     * @param path a path to a project, either absolute or relative to this project
+     * @return the absolute path to the project within the build, i.e. not including the build path
+     */
+    public Path resolveProjectPath(String path) {
```
alllex, line 171: 💭 Possibly worth adding a test to demostrate build path is not involved

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectOrderingUtil.java
```diff
@@ -23,12 +23,16 @@ public static int compare(Project left, Project right) {
...
+    }
+
+    public static int compare(ProjectIdentity left, ProjectIdentity right) {
+        int buildCompare = left.getBuildPath().compareTo(right.getBuildPath());
         if (buildCompare != 0) {
             return buildCompare;
         }
 
-        int depthCompare = depthCompare(left, right);
+        int depthCompare = Integer.compare(left.getProjectPath().segmentCount(), right.getProjectPath().segmentCount());
```
alllex: 🤔 We are duplicating the definition of depth here. While it's unlikely to change meaningfully, such a change increases obscurity.

A better approach would be to move that definition into the `ProjectIdentity`. We can then even remove it from `ProjectState` altogether. I'd also give it a more descriptive internal name, given the chance -- `getProjectDepth()` -- in line with `projectPath` that doesn't take the build tree into account.

alllex (changes requested):

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingTaskExecutionGraph.kt
octylFractal: I don't think the shorthand is necessary, yet.

alllex (approved):
LGTM! Great work 👍

Issue #26340, opened 2023-09-12
8.3 - JavaExec - argumentProviders with Input Provider results into Provider is queried before task run isn't support


hfhbd:
### Current Behavior

```
Execution failed for task ':generateLexer'.
> Querying the mapped value of task ':generateLexer' property 'targetOutputDir' before task ':generateLexer' has completed is not supported
```

Build scan: https://gradle.com/s/q6xg7xsjyyqno

Repo: https://github.com/hfhbd/gradle-jvmargprovider-bug

PR: Bumping Gradle from 8.2.1 to 8.3 https://github.com/hfhbd/gradle-jvmargprovider-bug/pull/1

### Expected Behavior

Behavior from 8.2.1: 
```
Task :generateLexer FAILED
Error: Could not find or load main class jflex.Main
Caused by: java.lang.ClassNotFoundException: jflex.Main


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':generateLexer'.
0 problems were found storing the configuration cache.
> Process 'command '/usr/lib/jvm/temurin-11-jdk-amd64/bin/java'' finished with non-zero exit value 1
```

Reason for failure: The reproducer does not contain this class on the class path, and it doesn't matter for the reproducer. Gradle executes the task `generateLexer` correctly.

### Context (optional)

I have a custom JavaExec task with uses the task inputs as inputs to the called main class, using argumentProvider to enable laziness.

I enabled stable configuration cache.

### Steps to Reproduce

Repo: https://github.com/hfhbd/gradle-jvmargprovider-bug

To reproduce:
1. Bump Gradle from 8.2.1 to 8.3
2. `./gradlew generateLexer`

There is already a bumped branch: `gradle83`

### Gradle version

8.3

### Gradle version that used to work

8.2.1

### Build scan URL (optional)

https://gradle.com/s/q6xg7xsjyyqno

hfhbd:
Workaround:
Replace 
```kotlin
private val <T : FileSystemLocation> Provider<T>.path: Provider<String>
  get() = map { it.asFile.canonicalPath }
```
with
```kotlin
private val <T : FileSystemLocation> Provider<T>.path: String
  get() = get().asFile.canonicalPath
```

erdi:
[...]

PR #37801, opened 2026-05-05
Introduce lazy variants of some file properties instead of upgrading them

asodja:
## Summary

Instead of upgrading certain `File` properties to lazy `RegularFileProperty` / `DirectoryProperty` with the same name (which leaves users with a confusing `File` → `DirectoryProperty` migration in their build scripts), this PR adds a **renamed** lazy property and keeps the original eager getter as a backward-compatible bridge. 

## ConventionMapping bridge for renamed properties

Some plugins use ConventionMapping on old properties. We also add a special ProviderApiMigrationConventionHelper that makes sure convention mapping works even after rename.

## Renamed (14 properties)

| Class | Old (eager `File`) | New (lazy) |
|---|---|---|
| `JacocoTaskExtension` | `getClassDumpDir()` | `getClassDumpDirectory(): DirectoryProperty` |
| `War` | `getWebXml()` | `getWebXmlFile(): RegularFileProperty` |
| `CreateStartScripts` | `getOutputDir()` | `getOutputDirectory(): DirectoryProperty` |
| `CreateStartScripts` | `getUnixScript()` | `getUnixScriptFile(): RegularFileProperty` |
| `CreateStartScripts` | `getWindowsScript()` | `getWindowsScriptFile(): RegularFileProperty` |
| `GroovyCompileOptions` | `getConfigurationScript()` | `getConfigurationScriptFile(): RegularFileProperty` |
| `GroovyCompileOptions` | `getStubDir()` | `getStubDirectory(): DirectoryProperty` |
| `CodeQualityExtension` | `getReportsDir()` | `getReportsDirectory(): DirectoryProperty` |
| `GenerateIvyDescriptor` | `getDestination()` | `getDestinationFile(): RegularFileProperty` |
| `GenerateMavenPom` | `getDestination()` | `getDestinationFile(): RegularFileProperty` |
| `Javadoc` | `getDestinationDir()` | `getDestinationDirectory(): DirectoryProperty` |
| `Groovydoc` | `getDestinationDir()` | `getDestinationDirectory(): DirectoryProperty` |
| `ScalaDoc` | `getDestinationDir()` | `getDestinationDirectory(): DirectoryProperty` |

Done in a separate PR https://github.com/gradle/gradle/pull/38061:
| Class | Old (eager `File`) | New (lazy) |
|---|---|---|
| `ProcessForkOptions` | `getWorkingDir()` | `getWorkingDirectory(): DirectoryProperty` |

## Properties that will be upgraded with Gradle 10

These will be migrated via `@ReplacesEagerProperty`. 
The property name is unchanged.

**`File` → `DirectoryProperty`**

| Class | Property |
|---|---|
| `AntlrTask` | `getOutputDirectory` |
| `MinimalJavadocOptions` | `getDestinationDirectory` |
| `TestNGOptions` | `getOutputDirectory` |
| `DirectoryBuildCache` | `getDirectory` | Skipped by request (proposed name was `getLocation`) |
| `ConventionReportTask` | `getOutputFile` |
| `JacocoTaskExtension` | `getDestinationFile` |
| `StandardJavadocDocletOptions` | `getHelpFile` |
| `StandardJavadocDocletOptions` | `getStylesheetFile` |
| `Wrapper` | `getJarFile` |
| `Wrapper` | `getScriptFile` |


platforms/jvm/jacoco/src/main/java/org/gradle/testing/jacoco/plugins/JacocoTaskExtension.java
```diff
@@ -211,12 +211,32 @@ public Property<Boolean> getIsDumpOnExit() {
     /**
      * Path to dump all class files the agent sees are dumped to. Defaults to no dumps.
      *
-     * @since 3.4
+     * @since 9.6.0
      */
     @Optional
     @LocalState
-    @ReplacesEagerProperty
```
abstratt: I think we should still have some metadata that allows automated migration tools/agents to figure out what API in Gradle 10 replaces what API in Gradle 9.

jvandort:
Have we considered that Directory and RegularFile are more clunky than they're worth and we should just use `Property<File>` everywhere?

I assume they exist so we can enforce that a given file is a directory or not a directory, but don't we have @InputFile @InputDirectory for that?

I was always not a fan of the .get().getAsFile() mess that we force users into, and I feel these changes are going to make that even worse

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ConventionAwareHelper.java
```diff
@@ -68,12 +68,22 @@ private MappedProperty map(String propertyName, MappedPropertyImpl mapping) {
                 "You can't map a property that does not exist: propertyName=" + propertyName);
         }
 
+        Class<? extends IConventionAware> sourceType = _source.getClass();
+
+        // Route ConventionMapping("oldName") to the renamed lazy property's `.convention()` API
+        // for properties whose lazy version has been renamed during the provider-api migration
+        // (e.g. CreateStartScripts: getOutputDir File -> getOutputDirectory DirectoryProperty).
+        // The eager getter is kept as a backward-compat bridge but external plugins still register conventions with the old name.
+        String renamedProperty = ProviderApiMigrationConventionHelper.findRenamedProperty(sourceType, propertyName);
```
abstratt, line 76: 💭 I missed having some coverage for this. Are there test cases that would fail if `findRenamedProperty` always returned null? Maybe ask Claude to add a data-driven test case that covers all the renamed properties (including implementation classes and subinterfaces)?

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ProviderApiMigrationConventionHelper.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+        return null;
+    }
+
+    private static String capitalize(String s) {
+        return Character.toUpperCase(s.charAt(0)) + s.substring(1);
+    }
+
+    private static String decapitalize(String s) {
+        return Character.toLowerCase(s.charAt(0)) + s.substring(1);
```
abstratt: ❓ 

Doesn't this break with  properties that are all-caps (`getIURI()`) or have two or more capital letters(`getJFlags()`)? Claude suggested `Introspector.decapitalize()`, use widely in the codebase

platforms/jvm/plugins-application/src/main/java/org/gradle/jvm/application/tasks/CreateStartScripts.java
```diff
@@ -202,36 +202,75 @@ private String computeExitEnvironmentVar() {
...
+
     /**
      * The directory to write the scripts into.
+     *
+     * @since 9.6.0
      */
     @OutputDirectory
-    @ReplacesEagerProperty
-    public DirectoryProperty getOutputDir() {
+    @Optional
```
abstratt: ❓ The original property was `@Nullable`, and this is `Optional`. Could you explain that?

platforms/jvm/war/src/main/java/org/gradle/api/tasks/bundling/War.java
```diff
@@ -73,7 +74,7 @@ public War() {
 
         CopySpecInternal renameSpec = webInf.addChild();
         renameSpec.into("");
-        renameSpec.from((Callable<?>) () -> getWebXml().getOrNull());
+        renameSpec.from((Callable<File>) () -> getWebXmlFile().getAsFile().getOrNull());
```
abstratt, line 80: ❓ 

Couldn't this be:

```
renameSpec.from(getWebXmlFile());
```

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
```diff
@@ -115,16 +116,31 @@ public ProcessForkOptions executable(Object executable) {
...
 
     @Override
-    public DirectoryProperty getWorkingDir() {
+    public DirectoryProperty getWorkingDirectory() {
         return workingDir;
     }
 
+    @Override
+    public File getWorkingDir() {
+        return getWorkingDirectory().isPresent() ? getWorkingDirectory().get().getAsFile() : null;
```
abstratt: ❌ 

The spec says: 

> "Never returns null"

platforms/jvm/plugins-application/src/main/java/org/gradle/jvm/application/tasks/CreateStartScripts.java
```diff
@@ -202,36 +202,75 @@ private String computeExitEnvironmentVar() {
...
      * Returns the full path to the Unix script. The target directory is represented by the output directory, the file name is the application name without a file extension.
      * TODO: This should be Provider[RegularFile], but we don't support such upgrade with @ReplacesEagerProperty
+     *
+     * @since 9.6.0
      */
     @Internal
-    @ReplacesEagerProperty(replacedAccessors = @ReplacedAccessor(value = GETTER, name = "getUnixScript"))
-    public RegularFileProperty getUnixScript() {
+    public RegularFileProperty getUnixScriptFile() {
         return getObjectFactory().fileProperty().value(
```
abstratt: ❓ Why do we create a new property every time?

platforms/jvm/plugins-application/src/main/java/org/gradle/jvm/application/tasks/CreateStartScripts.java
asodja: Good catch, I need to fix that

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
asodja: Good catch

platforms/jvm/plugins-application/src/main/java/org/gradle/jvm/application/tasks/CreateStartScripts.java
asodja: Good question, I need to research why I added it that way, I will take a look

platforms/jvm/war/src/main/java/org/gradle/api/tasks/bundling/War.java
asodja: I had this initially, but this broke some tests, since it's resolved differently (eagerly?)

asodja:
@jvandort 
> Have we considered that Directory and RegularFile are more clunky than they're worth and we should just use Property<File> everywhere?

Good question. We have general consensus we will use wrappers around Java File types. They are definitely clunky. But this is a bigger discussion around Provider API evolution, that I don't have a good answer to right now. I feel we should not research that as part of Provider API migration since it could become a rabbit hole. 

Do you want to open this discussion in a Slack thread to include also others and gets some opinions there?

lkasso:
@asodja, depending on where this lands, we need to add an upgrade_guide entry:

```
[[deprecated_eager_file_properties]]
==== Deprecation of eager `File` properties in favor of lazy alternatives

Several task and extension types have had their eager `File` properties deprecated in favor of lazy `DirectoryProperty` or `RegularFileProperty` equivalents.
The deprecated properties will be removed in Gradle 10.

The following table lists all affected properties and their replacements:

[cols="1,1,1", options="header"]
|===
| Type | Deprecated property | Replacement

| `Javadoc`
| `destinationDir`
| `destinationDirectory`

| `JacocoReport`
| `executionData` (file collection)
| Use `executionData.from()` instead

| `JacocoReportBase`
| `executionData` (file collection)
| Use `executionData.from()` instead

| `JavaCompile`
| `destinationDir`
| `destinationDirectory`

| `AbstractCompile`
| `destinationDir`
| `destinationDirectory`

| `ProcessResources`
| `destinationDir`
| `destinationDirectory`

| `Copy`
| `destinationDir`
| `destinationDirectory`

| `Sync`
| `destinationDir`
| `destinationDirectory`

| `AbstractCopyTask`
| `destinationDir`
| `destinationDirectory`

| `SourceSetOutput`
| `resourcesDir`
| `resourcesDirectory`

| `SourceTask`
| `source` (file tree)
| `source.from()`

| `SourceDirectorySet`
| `outputDir`
| `destinationDirectory`

| `BasePluginExtension`
| `libsDirectory`
| `libsDirectory` (now returns `DirectoryProperty`)

| `BasePluginExtension`
| `distsDirectory`
| `distsDirectory` (now returns `DirectoryProperty`)
|===

For example, if you configure `Javadoc.destinationDir`, migrate to `destinationDirectory`:

====
[.multi-language-sample]
=====
.build.gradle
[source,groovy]
----
javadoc {
    // Deprecated
    destinationDir = file("${layout.buildDirectory}/custom-javadoc")

    // Replacement
    destinationDirectory = layout.buildDirectory.dir("custom-javadoc")
}
----
=====
[.multi-language-sample]
=====
.build.gradle.kts
[source,kotlin]
----
tasks.javadoc {
    // Deprecated
    destinationDir = file("${layout.buildDirectory}/custom-javadoc")

    // Replacement
    destinationDirectory = layout.buildDirectory.dir("custom-javadoc")
}
----
=====
====
```

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ProviderApiMigrationConventionHelper.java
asodja: Fixed

platforms/jvm/plugins-application/src/main/java/org/gradle/jvm/application/tasks/CreateStartScripts.java
asodja: Fixed

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
asodja: Fixed

platforms/jvm/javadoc/src/main/java/org/gradle/api/tasks/javadoc/Javadoc.java
```diff
@@ -139,7 +140,7 @@ public Javadoc() {
 
     @TaskAction
     protected void generate() {
-        File destinationDir = getDestinationDir().getAsFile().get();
+        File destinationDir = getDestinationDirectory().get().getAsFile();
```
hegyibalint, line 147: 🤔 Maybe out of scope for this PR, but I think for file-based properties, and especially in this case, we could add next to the `getAsFileTree()` method a `getAsFile()` method, so we don't need that intermediate `get()`.

platforms/jvm/plugins-application/src/main/java/org/gradle/jvm/application/tasks/CreateStartScripts.java
asodja: Fixed, now it's a abstract method and original value is a convention

testing/architecture-test/src/changes/accepted-changes/propertyExtension-accepted-public-api-changes.json
```diff
@@ -3,7 +3,7 @@
         {
             "type": "org.gradle.kotlin.dsl.PropertyExtensionsKt",
             "member": "Method org.gradle.kotlin.dsl.PropertyExtensionsKt.get(org.gradle.api.provider.MapProperty,java.lang.Object)",
-            "acceptation": "New += API",
+            "acceptation": "New +\u003d API",
```
hegyibalint: ❓ Huh, why this new encoding?

testing/architecture-test/src/changes/accepted-changes/provider-api-migration-accepted-public-api-changes.json
```diff
@@ -1,20 +1,157 @@
...
     "acceptedApiChanges": [
+        {
+            "type": "org.gradle.api.plugins.quality.CodeQualityExtension",
+            "member": "Method org.gradle.api.plugins.quality.CodeQualityExtension.getReportsDirectory()",
+            "acceptation": "Renamed lazy property added; eager getReportsDir() retained as a bridge",
+            "changes": [
+                "Method added to public class",
+                "Abstract method has been added to this class"
+            ]
+        },
```
hegyibalint: 🤔 Are these happening because we skipped the `@Incubating` phase?

testing/architecture-test/src/changes/accepted-changes/provider-api-migration-accepted-public-api-changes.json
```diff
@@ -1,20 +1,157 @@
...
+                "Method added to public class",
+                "Abstract method has been added to this class",
+                "ANNOTATION_ADDED"
+            ]
+        },
         {
             "type": "org.gradle.api.tasks.wrapper.Wrapper",
             "member": "Method org.gradle.api.tasks.wrapper.Wrapper.setJarFile(java.lang.Object)",
-            "acceptation": "Removed since it's an Object setter",
+            "acceptation": "Removed since it\u0027s an Object setter",
```
hegyibalint: ❓ Huh, why this new encoding?

hegyibalint (changes requested):
Renaming LGTM, some questions and a concern about some weird encoding

testing/architecture-test/src/changes/accepted-changes/propertyExtension-accepted-public-api-changes.json
asodja: Good catch. Will fix it, alphabetic sort task does that

testing/architecture-test/src/changes/accepted-changes/provider-api-migration-accepted-public-api-changes.json
asodja: I will take a look

testing/architecture-test/src/changes/accepted-changes/provider-api-migration-accepted-public-api-changes.json
asodja: Good catch. Will fix it, alphabetic sort task does that

testing/architecture-test/src/changes/accepted-changes/provider-api-migration-accepted-public-api-changes.json
asodja: Yes, this was the reason. I added `@Incubating` for now

testing/architecture-test/src/changes/accepted-changes/provider-api-migration-accepted-public-api-changes.json
asodja: This was fixed

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ProviderApiMigrationConventionHelper.java
```diff
@@ -0,0 +1,133 @@
...
+ * with a different name (e.g. {@code outputDir} -> {@code outputDirectory}).
+ *
+ * <p>The eager getter is kept as a backward-compatible bridge, but ConventionMapping
+ * registration uses the old property name and would otherwise never propagate to the lazy
+ * property. This helper holds a hardcoded set of known renames so {@link ConventionAwareHelper}
+ * can route the convention to the lazy property's {@code .convention()} API directly.</p>
+ *
+ * <p>Entries should be removed when the eager bridge is removed in a future major release.</p>
+ */
+class ProviderApiMigrationConventionHelper {
```
big-guy, line 42: Can this not just look at `ReplacedBy`?

platforms/documentation/docs/src/snippets/reference/plugin-development/reactingToPlugins/groovy/buildSrc/src/main/java/InhouseConventionWarPlugin.java
```diff
@@ -6,7 +6,7 @@
 public class InhouseConventionWarPlugin implements Plugin<Project> {
     public void apply(Project project) {
         project.getTasks().withType(War.class).configureEach(war ->
-            war.setWebXml(project.file("src/someWeb.xml")));
+            war.getWebXmlFile().set(project.file("src/someWeb.xml")));
```
big-guy: It doesn't quite matter, but I would probably make this `.convention`

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CodeQualityExtension.java
```diff
@@ -82,16 +85,25 @@ public void setIgnoreFailures(boolean ignoreFailures) {
...
+        return getReportsDirectory().isPresent() ? getReportsDirectory().get().getAsFile() : null;
     }
 
     /**
-     * The directory where reports will be generated.
+     * Sets the directory where reports will be generated.
      */
     public void setReportsDir(File reportsDir) {
-        this.reportsDir = reportsDir;
+        getReportsDirectory().set(reportsDir);
```
big-guy, line 109: To fully mimic previous behavior, this also needs to set `getReportsDirectory().convention(reportsDir)`

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CodeQualityExtension.java
```diff
@@ -82,16 +85,25 @@ public void setIgnoreFailures(boolean ignoreFailures) {
...
+    public abstract DirectoryProperty getReportsDirectory();
+
+    /**
+     * The directory where reports will be generated.
+     */
+    @ReplacedBy("reportsDirectory")
+    @NotToBeReplacedByLazyProperty(because = "Bridge for backward compatibility, use getReportsDirectory() instead", willBeDeprecated = true)
     public File getReportsDir() {
-        return reportsDir;
+        return getReportsDirectory().isPresent() ? getReportsDirectory().get().getAsFile() : null;
```
big-guy, line 102: Why not `getReportsDirectory().getAsFile().getOrNull()`?

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CodeQualityExtension.java
```diff
@@ -82,16 +85,25 @@ public void setIgnoreFailures(boolean ignoreFailures) {
 
     /**
      * The directory where reports will be generated.
+     *
+     * @since 9.7.0
      */
-    @ToBeReplacedByLazyProperty
+    @Incubating
+    public abstract DirectoryProperty getReportsDirectory();
```
big-guy, line 94: ❌ The XML for our docs need to be updated with any new properties.

platforms/jvm/jacoco/src/main/java/org/gradle/testing/jacoco/plugins/JacocoTaskExtension.java
```diff
@@ -93,6 +97,7 @@ public JacocoTaskExtension(ObjectFactory objects, JacocoAgentJar agent, JavaFork
         this.agent = agent;
         this.task = task;
         destinationFile = objects.fileProperty();
+        classDumpDirectory = objects.directoryProperty();
```
big-guy: We should be able to make this abstract like the others too?

platforms/jvm/javadoc/src/main/java/org/gradle/api/tasks/javadoc/Javadoc.java
big-guy: FileSystemLocationProperty has a getAsFile already.

platforms/jvm/war/src/main/java/org/gradle/api/tasks/bundling/War.java
big-guy: I wonder if it's the difference between `from(null)` and `from(() -> null)`?

big-guy (changes requested):
I only left comments on the first occurrence of some things that I think should change.

Overall, I think this is in the right direction

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CodeQualityExtension.java
asodja: I wanted to avoid weird failures, like https://github.com/gradle/gradle/issues/26340 where mapping in some cases fails. While this is true only for output directories, I applied this pattern to all properties

platforms/jvm/javadoc/src/main/java/org/gradle/api/tasks/javadoc/Javadoc.java
asodja: > we could add next to the getAsFileTree() method a getAsFile() method

As Sterling said, we have `getAsFile` but it return `Provider<Directory>` so you still need to call `get()` at the end, so you would have:
```
getDestinationDirectory().getAsFile().get()
```

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CodeQualityExtension.java
asodja: Good catch

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ProviderApiMigrationConventionHelper.java
asodja: We can add support for `ReplacedBy`. It would be cleaner solution. 

It has two caveats:
- it needs to get property method via reflection and check the annotation for every convention mapping
- it also applies to all `ReplacedBy` methods that are already there, which could be seen as a bug fix

platforms/documentation/docs/src/snippets/reference/plugin-development/reactingToPlugins/groovy/buildSrc/src/main/java/InhouseConventionWarPlugin.java
asodja: Fixed

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CodeQualityExtension.java
asodja: Why is that?

platforms/jvm/war/src/main/java/org/gradle/api/tasks/bundling/War.java
asodja: So `from(null) and from(() -> null)` are both skipped -> Callable<File> is unpacked and nukll is skipped, so it should be the same.

While `from(getWebXmlFile())` as Rafael suggested is different, since getWebXmlFile() is a Property. UnpackingVisitor wraps it in a `ProviderBackedFileCollection(..., REQUIRE_PRESENT)`. On resolution it calls `provider.get()`, which throws when the `property has no value` exception.

asodja:
@bot-gradle test this BD

platforms/jvm/jacoco/src/main/java/org/gradle/testing/jacoco/plugins/JacocoTaskExtension.java
asodja: Fixed

asodja:
@bot-gradle test this APT

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ProviderApiMigrationConventionHelper.java
asodja: I changed this logic so we figure it out via `ReplacedBy` annotation as suggested

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ProviderApiMigrationConventionHelper.java
```diff
@@ -0,0 +1,95 @@
...
+ * can route legacy {@code getConventionMapping().map("oldName", ...)} calls to the lazy
+ * property's {@code .convention()} API.
+ *
+ * <p>Renames are discovered from {@link ReplacedBy} on the old getter and only routed when the
+ * replacement is a lazy ({@link SupportsConvention}) property. They are computed once per class
+ * and cached, so lookups are constant-time without reflection.</p>
+ */
+class ProviderApiMigrationConventionHelper {
+
+    private static final ClassValue<Map<String, String>> REPLACED_PROPERTIES = new ClassValue<Map<String, String>>() {
```
big-guy, line 44: I think this may cause classloaders for user provided types to never become collectable?

cc @octylFractal

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CheckstylePlugin.java
```diff
@@ -108,17 +104,10 @@ private void configureTaskConventionMapping(Configuration configuration, Checkst
...
+        Provider<Directory> reportsDir = extension.getReportsDirectory();
         task.getReports().all(action(report -> {
             report.getRequired().convention(!report.getName().equals("sarif"));
-            report.getOutputLocation().convention(
-                layout.getProjectDirectory().file(providers.provider(() -> {
-                    String reportFileName = baseName + "." + report.getName();
-                    return new File(reportsDir.get().getAsFile(), reportFileName).getAbsolutePath();
-                }))
-            );
+            report.getOutputLocation().convention(reportsDir.map(dir -> dir.file(baseName + "." + report.getName())));
```
big-guy: I think you could
```suggestion
            report.getOutputLocation().convention(reportsDir.file(baseName + "." + report.getName()));
```
if you use DirectoryProperty for reportsDir vs the Provider

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CodeQualityExtension.java
big-guy: `set(null)` means "unset the value and return to the convention", so if the convention was not cleared, you would change behavior sometimes.

Previously, if you had:
```
checkstyle.reportsDir = file("path/to/somewhere")
// far away ...
checkstyle.reportsDir = null
```
`checkstyle.reportsDir` would return null.  

With this change, if `reportsDirectory` has a convention set, this same pattern would return the convention instead of null. For file paths, especially those that are inputs/outputs, it's unlikely these are being explicitly set to null.

There's also some similarity to how convention mapping works. Once a value has been set on a property with convention mapping, we never go back to the convention mapping, so calling both `set(...)` and `convention(...)` makes the setter's value "win" always.

We ran into this problem with `Property<String>`'s on archive tasks where the convention was a non-empty string and setting the property to null used to clear it/make it an empty string.

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CodeQualityExtension.java
big-guy: Did you actually see these issues? 

For #26340 to be an issue, we would need to be calling this getter ourselves in the task action?

big-guy (changes requested):

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ProviderApiMigrationConventionHelper.java
octylFractal: The whole point of `ClassValue` is to avoid that, as the value is held in the `Class` itself and this field holds no reference to the `Class`.

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CheckstylePlugin.java
asodja: Fixed

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CodeQualityExtension.java
asodja: Fixed

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CodeQualityExtension.java
asodja: I changed this to always apply also convention, but it looks a bit funny

platforms/jvm/code-quality/src/main/java/org/gradle/api/plugins/quality/CodeQualityExtension.java
asodja: I didn't see this issues, but I applied this pattern everywhere just to not have to deal with potential issues :) 
If you prefer I do `getReportsDirectory().getAsFile().getOrNull()` I can change it.


> We would need to be calling this getter ourselves in the task action?

That is true

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

big-guy (approved):

lkasso:
@asodja, are there any docs planned?

asodja:
@lkasso I've opened https://github.com/gradle/gradle/issues/38362. Let's do it there. I will add you as a reviewer for a PR

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

hegyibalint (approved):

PR #37802, opened 2026-05-05
Fix NPE due to null path in lenient ArtifactView resolution

### LINKED ISSUE #36284: Runtime Error 'Cannot invoke "java.util.List.get(int)" because "path" is null' when getting message of ModuleVersionResolveException
### Current Behavior

Fails with RuntimeError 

```
Caused by: java.lang.NullPointerException: Cannot invoke "java.util.List.get(int)" because "path" is null
        at org.gradle.internal.resolve.ModuleVersionResolveException.getMessage(ModuleVersionResolveException.java:111)
        at Build_gradle$DependencyGraphPrinter.action(build.gradle.kts:24)
```

### Expected Behavior

Finish successfully

### Context (optional)

There is no null-check for path in this code:
https://github.com/gradle/gradle/blob/886d6d379535e5d09a8e4bab17c0c1e0b734e7cf/platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/ModuleVersionResolveException.java#L111

```
plugins {
    java
}

repositories {
    mavenCentral()
    google()
}

dependencies {
    implementation("androidx.room:room-sqlite-wrapper:2.8.3")
    implementation("androidx.room:room-runtime:2.8.3")
}

abstract class DependencyGraphPrinter : DefaultTask() {
    @get:Internal
    abstract val ac: Property<ArtifactCollection>

    @TaskAction
    fun action() {
        println("Failures: ")
        ac.get().failures.forEach {
            println(it.message)
        }
    }
}

tasks.register("dgp", DependencyGraphPrinter::class) {
    val ac = configurations.get("compileClasspath").incoming.artifactView {
        isLenient = true
    }.artifacts
    this.ac.set(ac)
}
```

1. run `./gradlew dgp`
2. it will fail with 

```
Cannot invoke "java.util.List.get(int)" because "path" is null
```

When I try to get a message of artifact resolution failure: `println(it.message)`

### Gradle version

9.4 nightly

### Build scan URL (optional)

_No response_

_No response_

tresat:
When a consumer iterates `ArtifactCollection.failures` produced by a lenient `ArtifactView` and calls `Throwable.getMessage()` on each failure, Gradle throws `NullPointerException: Cannot invoke "java.util.List.get(int)" because "path" is null` from `ModuleVersionResolveException.getMessage()` (line 111). This fixes that behavior upstream — calculatePaths must never return a
collection containing null elements — not as a guard inside
getMessage().

Adds a unit test on DependencyGraphPathResolver.calculatePaths that
synthesizes the topology where a direct dependee's upward
getDependents() BFS cannot reach the seeded root. In that case the
final assembly loop appends null to the returned paths collection,
which later flows into ModuleVersionResolveException.withIncomingPaths
and triggers an NPE in getMessage() at user-facing iteration time
(e.g. ArtifactCollection.failures.forEach { println it.message }).

Also adds an integration test reproducer "lenient ArtifactView failures expose messages without NPE for androidx.room artifacts" in `ResolutionIssuesIntegrationTest`

Issue: #36284

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ResolutionIssuesIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ResolutionIssuesIntegrationTest.groovy
@@ -20,6 +20,7 @@ import groovy.test.NotYetImplemented
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.JdkVersionTestPreconditions
+import org.gradle.test.preconditions.TestEnvironmentPreconditions
 
 import spock.lang.Ignore
 import spock.lang.Issue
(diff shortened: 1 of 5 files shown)


tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/DependencyGraphPathResolver.java
```diff
@@ -32,6 +32,24 @@
...
 public class DependencyGraphPathResolver {
 
+    /**
+     * Calculates the shortest "incoming path" from each direct dependee of a broken dependency
+     * back to the resolution root, used to render the "Required by:" tail in
+     * {@link org.gradle.internal.resolve.ModuleVersionResolveException#getMessage}.
+     *
+     * <p>Some direct dependees can have an empty {@code getDependents()} chain at the time
+     * this runs — most commonly because the dependee's module was pushed back to "pending"
+     * state during resolution and {@code ModuleResolveState.clearIncomingAttachedConstraints}
```
jvandort, line 42: I don't think we should be referencing engine internals in this javadoc.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/DependencyGraphPathResolver.java
```diff
@@ -32,6 +32,24 @@
 
 public class DependencyGraphPathResolver {
 
+    /**
+     * Calculates the shortest "incoming path" from each direct dependee of a broken dependency
+     * back to the resolution root, used to render the "Required by:" tail in
+     * {@link org.gradle.internal.resolve.ModuleVersionResolveException#getMessage}.
+     *
+     * <p>Some direct dependees can have an empty {@code getDependents()} chain at the time
```
jvandort, line 40: > Some direct dependees can have an empty {@code getDependents()} chain at the time this runs 

This sounds like a bug in the resolution engine. I don't think we should be glossing over it here. Any node in the graph other than the root should have incoming edges.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ResolutionIssuesIntegrationTest.groovy
```diff
@@ -283,4 +284,58 @@ class ResolutionIssuesIntegrationTest extends AbstractIntegrationSpec {
...
+        // DependencyGraphPathResolver.calculatePaths could return a paths
+        // collection containing a null element when a direct dependee's
+        // module had returned to "pending" state and lost all its
+        // constraint-only incoming edges.
+        buildFile << """
+            plugins {
+                id("java")
+            }
+
+            ${mavenCentralRepository()}
```
jvandort, line 304: We should try to reduce this down to the minimal graph required to reproduce this behavior rather than rely on a complex graph like this reproducer likely generates

jvandort (changes requested):
I think we need to fix the underlying problem that causes us to have nodes in a graph with no incoming edges rather than handling a malformed graph after it is built.

Though given we have a reproducer, we should be able to figure out the underlying problem

PR #34868, opened 2025-09-02
Remove mutable references to other projects in DefaultProject

jvandort:
DefaultProject previously held mutable references to the root project, its parent project, and its owning Gradle instance. Now, we access parents, roots, and builds through the 'owner' ProjectState field, which facilitates safer access to this mutable state.

See https://github.com/gradle/gradle/issues/34469

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #37804, opened 2026-05-05
Wrap IsolatedProject around ProjectState

alllex:
This is part of the effort to reduce internal use of `Project` mutable models, with the important goal of avoiding references to other projects from within the `Project` implementation.

This PR is a step in that direction. It removes one more reason for the `DefaultProject` to have a reference to `rootProject`.

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultIsolatedProject.java
+++ b/subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultIsolatedProject.java
@@ -22,44 +22,42 @@ import org.jspecify.annotations.Nullable;
 
 public final class DefaultIsolatedProject implements IsolatedProject {
 
-    private final ProjectInternal project;
-    private final ProjectInternal rootProject;
+    private final ProjectState owner;
 
-    public DefaultIsolatedProject(ProjectInternal project, ProjectInternal rootProject) {
-        this.project = project;
-        this.rootProject = rootProject;
+    public DefaultIsolatedProject(ProjectState owner) {
+        this.owner = owner;
     }
 
     @Override
     public String getName() {
-        return project.getName();
+        return owner.getName();
     }
 
     @Override
     public String getPath() {
-        return project.getPath();
+        return owner.getProjectPath().asString();
     }
 
     @Override
     public String getBuildTreePath() {
-        return project.getBuildTreePath();
+        return owner.getIdentityPath().asString();
     }
 
     @Override
     public Directory getProjectDirectory() {
-        return project.getLayout().getProjectDirectory();
+        // TODO: avoid mutable model access and have a Directory-typed project dir available on ProjectState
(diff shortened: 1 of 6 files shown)


octylFractal (approved):

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultIsolatedProject.java
```diff
@@ -22,44 +22,42 @@
...
     @Override
     public String getBuildTreePath() {
-        return project.getBuildTreePath();
+        return owner.getIdentityPath().asString();
     }
 
     @Override
     public Directory getProjectDirectory() {
-        return project.getLayout().getProjectDirectory();
+        // TODO: avoid mutable model access and have a Directory-typed project dir available on ProjectState
```
jvandort, line 48: In my prior hack attempt at doing this I injected a `FileFactory` to do so:
https://github.com/gradle/gradle/pull/34868/changes#diff-b17e703426d8edb8fa68ccbd0933dc8d9d2b35cd8663890055321e420b02cc52

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectState.java
```diff
@@ -107,6 +108,11 @@ public ImmutableProjectDescriptor getDescriptor() {
         return descriptor;
     }
 
+    @Override
+    public IsolatedProject getIsolated() {
```
jvandort, line 112: I'm not quite sure how I feel about having this on the `ProjectState`. Since `IsolatedProject` is a high-level wrapper over a `ProjectState` _for the public API_, I think it should live in the "public API" layer of our model rather than pulling public API types into the inner "internal API" layer

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultIsolatedProject.java
alllex: That's also what we do inside `ProjectLayout`. Though, it's not 100% clear from which scope the FileFactory we are using is. Or it's clear it's Project-scope, but then it's kinds of weird. I don't recall exactly. It shouldn't matter though, as the project dir should already be absolute. But I wanted to avoid duplication of this logic before we make this more straight-forward.

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectState.java
alllex: It's not much different from what `getMutableModel()` returns, but I see your point.

For now, I think it's the most convenient to keep it close to ProjectState, especially if we want to keep enriching it with capabilities.

One option is to make `IsolatedProject` a Project-scoped service. It's clearly associated 1-to-1 with the project, and then it's also clear how we can keep extending the functionality it exposes.

PR #37650, opened 2026-04-23
Preserve property identity for directory & file properties

ov7a:
### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #37818, opened 2026-05-06
Fix serialization issue for value source with a cyclic dependency

### LINKED ISSUE #32828: Extension nested properties used in value source parameters can cause configuration cache deserialization timeout
### Current Behavior

When an extension holds a reference to a `ValueSource`-backed provider, and configures that provider's parameters using objects `@Nested` in the extension, keeps a reference to that provider on the extension, then uses the provider as a task input, gradle hangs when recovering from conf cache until failing in a timeout. For example, the following extension:

```java
public abstract class A {
    public abstract static class B {}

    public abstract static class Params implements ValueSourceParameters {
        @Nested
        public abstract Property<B> getB();
    }

    public abstract static class V implements ValueSource<String, Params> {}

    private final Provider<String> calculated;

    @Inject
    public A(ProviderFactory providers) {
        var source = providers.of(V.class, it -> {
            it.getParameters().getB().set(getB());
        });
        this.calculated = source;
    }

    @Nested
    public abstract B getB();

    public Provider<String> getCalculated() {
        return calculated;
    }
}
```

Will cause an issue if `calculated` is used as a task input.

### Expected Behavior

This should either work with conf cache, or should provide an error to explain why it does not.

### Context (optional)

_No response_

https://github.com/lukebemish/gradle-issue-circular-serialization

### Gradle version

8.13

### Build scan URL (optional)

_No response_

_No response_

ov7a:
Fixes 
* #32828 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheValueSourceIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheValueSourceIntegrationTest.groovy
@@ -927,4 +927,64 @@ class ConfigurationCacheValueSourceIntegrationTest extends AbstractConfiguration
         and:
         configurationCache.assertStateLoaded()
     }
+
+    @Issue("https://github.com/gradle/gradle/issues/32828")
+    def "value source whose parameters reference an extension's nested bean does not deadlock on store/load"() {
+        given:
+        def configurationCache = newConfigurationCacheFixture()
+        buildFile """
+            import javax.inject.Inject
+
+            abstract class B {
+                abstract Property<String> getName()
+            }
+
+            abstract class A {
+                @Nested abstract B getB()
+
+                final Provider<String> calculated
+
+                @Inject
+                A(ProviderFactory providers) {
+                    getB().name.convention("hi")
+                    calculated = providers.of(MySrc) { spec ->
+                        spec.parameters.b.set(getB())
+                    }
+                }
+            }
+
+            abstract class MySrc implements ValueSource<String, Params> {
+                interface Params extends ValueSourceParameters {
+                    @Nested Property<B> getB()
+                }
+                String obtain() {
+                    parameters.b.get().name.get() + "!"
+                }
+            }
+
+            abstract class ShowTask extends DefaultTask {
+                @Input abstract Property<String> getValue()
+                @TaskAction void run() { println("v=" + value.get()) }
+            }
+
+            def aInstance = objects.newInstance(A)
+            tasks.register("show", ShowTask) {
+                value = aInstance.calculated
+            }
+        """
+
+        when:
+        configurationCacheRun "show"
+
+        then:
+        configurationCache.assertStateStored()
+        outputContains("v=hi!")
+
+        when:
+        configurationCacheRun "show"
+
+        then:
+        configurationCache.assertStateLoaded()
+        outputContains("v=hi!")
+    }
 }
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/DefaultSharedObjectCodec.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/DefaultSharedObjectCodec.kt
@@ -187,11 +187,11 @@ class DefaultSharedObjectDecoder(
             }
         } catch (t: Throwable) {
             readFailure.set(t)
+        } finally {
+            state.set(ReaderState.STOPPED)
             for (entry in values.values) {
                 if (entry is FutureValue) entry.release()
             }
-        } finally {
-            state.set(ReaderState.STOPPED)
         }
     }
 
@@ -207,7 +207,7 @@ class DefaultSharedObjectDecoder(
         throwIfReaderFailed()
         return when (val existing = values.computeIfAbsent(id) { FutureValue(id) }) {
             is FutureValue -> {
-                if (readFailure.get() != null) {
+                if (state.get() == ReaderState.STOPPED || readFailure.get() != null) {
                     existing.release()
                 }
                 existing.get()
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/DefaultSharedObjectCodec.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/DefaultSharedObjectCodec.kt
@@ -28,6 +28,7 @@ import org.gradle.internal.serialize.graph.WriteContext
 import org.gradle.internal.serialize.graph.ownerService
 import org.gradle.internal.serialize.graph.runReadOperation
 import org.gradle.internal.serialize.graph.runWriteOperation
+import java.time.Duration
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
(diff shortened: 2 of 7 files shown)


ov7a:
@bot-gradle test this

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/ProviderCodecs.kt
```diff
@@ -326,20 +326,22 @@ object ValueSourceProviderCodec : Codec<ValueSourceProvider<*, *>> {
...
-
+            var decodedParameters: ValueSourceParameters? = null
             val valueSourceProviderFactory = isolate.owner.serviceOf<ValueSourceProviderFactory>()
-            val provider =
-                valueSourceProviderFactory.instantiateValueSourceProvider<Any, ValueSourceParameters>(
-                    valueSourceType.uncheckedCast(),
-                    parametersType?.uncheckedCast(),
-                    parameters?.uncheckedCast()
-                )
+            val provider = valueSourceProviderFactory.instantiateValueSourceProvider<Any, ValueSourceParameters>(
```
ov7a: The case looks very similar to the one I discovered in APG with File/DirProp:
* https://github.com/gradle/gradle/pull/37650/changes#diff-8e35ba4350b71217bf35a5682d7be0ec0008b4c4e7086b3dd25280ba4e79a6e7R165

it feels like a general issue with identities decoding at this point

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/DefaultValueSourceProviderFactory.java
```diff
@@ -262,8 +264,9 @@ private class LazilyObtainedValue<T, P extends ValueSourceParameters> {
         @Nullable
         public final Class<P> parametersType;
 
+        private final Supplier<@Nullable P> parametersSupplier;
```
mlopatkin: :x: This is going to have a non-trivial memory cost. In Gradle's `sanityCheck` we have about 100K value sources, and with the proposed supplier chain we'll be adding at least 32 bytes per instance. 

I suggest rewriting this in a continuation-passing style. Have a non-final field for parameters, and pass a `Function<Provider, Parameters>` from the codec. The function should register the instance and `read()` and return the parameters. We have `runReadOperation` helper to call suspend functions synchronously. Will require some plumbing in the factory, but doesn't look too complex at the first glance.

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/DefaultValueSourceProviderFactory.java
```diff
@@ -296,6 +299,14 @@ public boolean hasBeenObtained() {
             return value.isFinalized();
         }
 
+        @Nullable
+        public P getParameters() {
+            if (parameters == null) {
+                parameters = parametersSupplier.get();
```
mlopatkin: :question: what if the parameters aren't ready yet? E.g. something obtains the value source when reading them. As we're dealing with cycles, this is already possible. I suggest that we fail there. Also, worth having a test.

I suppose it is problematic to implement until `None` unification lands, because `None` is represented as `null` here?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheValueSourceIntegrationTest.groovy
```diff
@@ -920,4 +920,64 @@ class ConfigurationCacheValueSourceIntegrationTest extends AbstractConfiguration
         and:
         configurationCache.assertStateLoaded()
     }
+
+    @Issue("https://github.com/gradle/gradle/issues/32828")
+    def "value source whose parameters reference an extension's nested bean does not deadlock on store/load"() {
```
mlopatkin, line 932: :thinking: There is another failure mode that may cause deadlock - suppose that the value source fails to deserialize at all. Due to the way async reading is currently implemented in `DefaultSharedObjectDecoder`, we'll never know if something has failed to read until timeout passes. Worth addressing there or in a separate PR before closing the issue - to cover all bases.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/ProviderCodecs.kt
```diff
@@ -326,20 +326,22 @@ object ValueSourceProviderCodec : Codec<ValueSourceProvider<*, *>> {
 
     private
     suspend fun ReadContext.decodeValueSource(): ValueSourceProvider<*, *> =
-        // TODO:configuration-cache `decodePreservingSharedIdentity` should be unnecessary for shared objects
```
mlopatkin, line 325: :nail_care: This TODO is still relevant.

mlopatkin (changes requested):
I have concerns about potential memory increase.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/ProviderCodecs.kt
ov7a: Oops, my mistake

mlopatkin (approved):
Looks good so far, but I would not postpone `DefaultSharedObjectDecoder` changes for too long.

ov7a:
@coderabbitai review

ov7a:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/DefaultSharedObjectCodec.kt
```diff
@@ -108,14 +113,28 @@ class DefaultSharedObjectDecoder(
...
+                )
+            }
             val state = state.get()
             // Only await if the reading thread is still running.
             // This saves us a minute in case the reading code is broken and doesn't countDown() the latch properly.
             // See the null check below.
-            if (state < ReaderState.STOPPED && !latch.await(1, TimeUnit.MINUTES)) {
-                throw TimeoutException("Timeout while waiting for value, state was $state")
+            if (state < ReaderState.STOPPED && !latch.await(valueWaitTimeout.toNanos(), TimeUnit.NANOSECONDS)) {
+                reader.join(TimeUnit.SECONDS.toMillis(5))
```
mlopatkin, line 133: ❓ Why do we need to join the reader?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/DefaultSharedObjectCodec.kt
```diff
@@ -137,26 +163,34 @@ class DefaultSharedObjectDecoder(
...
+                                is FutureValue -> value.complete(read)
+                                else -> require(value == null)
+                            }
+                            read
                         }
-                        read
                     }
                 }
             }
+        } catch (t: Throwable) {
```
mlopatkin, line 188: :x: There is a small catch (no pun intended) not yet handled. It is possible for broken code to ask for an id that is never going to be read. E.g. I have 10 Foos written, with ids 1 to 10. A decoder goes wild and asks for a Foo with id=11. It blocks. The reader thread completes successfully, but the decoder still waits for a Foo that never arrives. A pass to release existing FutureValues would be a good mitigation to avoid the 1-minute hiccup.

mlopatkin (changes requested):
A few comments, I think we should resolve them before we can merge.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/DefaultSharedObjectCodec.kt
ov7a: To give it a chance to write `readFailure`. With a small join, it will trigger a context switch to update the value, which may be useful when the thread is currently dying.

mlopatkin (approved):

PR #37820, opened 2026-05-06
Use exact version when resolving Gradle source distributions

### LINKED ISSUE #37819: KotlinDslScriptsModel can return sources that do not match the generated Gradle API classpath for branch snapshot builds
### Current Behavior

For a Gradle branch snapshot build, KotlinDslScriptsModel can return a generated Gradle API binary classpath and a sourcePath that do not describe the same API.

In the attached minimal Tooling API repro:
- KotlinDslScriptsModel.classPath contains the generated Gradle API jar:
`<GRADLE_USER_HOME>/caches/9.6.0-20260420115815+0000/generated-gradle-jars/gradle-api-9.6.0-20260420115815+0000.jar
`
- That binary has the new API: `public abstract DirectoryProperty getDestinationDir();`
- But KotlinDslScriptsModel.sourcePath points to Javadoc.java source where the same API is still old: `public File getDestinationDir()`

So, Tooling API consumers receive a model where binaries and attached sources disagree.

As suggested by @eskatos, this probably happens because the exact source distribution for the branch snapshot is not available, and Gradle appears to fall back to a nearby source distribution: https://github.com/gradle/gradle/blob/ac7725e0abd4bff2675fa1d61e7573582ae92aeb/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourceDistributionProvider.kt#L149

### Expected Behavior

`KotlinDslScriptsModel.sourcePath` should only contain source roots that match the returned binary classpath. If no exact matching source distribution exists for the Gradle build, Gradle should return no sources for that API jar instead of returning approximate but incompatible sources.

### Context (optional)

This was observed from IntelliJ IDEA as incorrect navigation from Gradle Kotlin DSL scripts to old Gradle API sources. The attached repro removes IntelliJ from the equation and queries the raw Gradle Tooling API model directly. Returning no sources would be better for IDEs than returning mismatched sources, because the IDE can then fall back to decompiled binary navigation.

```Kotlin
package org.example

import org.gradle.tooling.GradleConnector
import org.gradle.tooling.model.kotlin.dsl.KotlinDslScriptsModel
import java.io.File
import java.nio.file.Path
import java.util.zip.ZipFile
import kotlin.io.path.exists
import kotlin.io.path.isDirectory
import kotlin.io.path.isRegularFile


fun main() {
    val model = GradleConnector.newConnector()
        .forProjectDirectory(File("test-project").absoluteFile)
        .connect()
        .use { it.model(KotlinDslScriptsModel::class.java).get() }

    val scripts = model.scriptModels.entries.sortedBy { it.key.absolutePath }
    printCollection("scripts", scripts) { (script, scriptModel) ->
        println("- script=$script")
        printCollection("  classPath", scriptModel.classPath) {
            entryPath(it.toPath(), "org/gradle/api/tasks/javadoc/Javadoc.class")
                ?.let { path -> println("   $path") }
        }
        printCollection("  sourcePath", scriptModel.sourcePath) {
            entryPath(it.toPath(), "org/gradle/api/tasks/javadoc/Javadoc.java")
                ?.let { path -> println("   $path") }
        }
        println("  exceptions=" + scriptModel.exceptions)
        println("  reports=" + scriptModel.editorReports)
    }
}

private fun <T> printCollection(label: String, collection: Collection<T>, printElement: (T) -> Unit) {
    println("$label (size=" + collection.size + ")")
    for (element in collection) {
        printElement(element)
    }
}

private fun entryPath(root: Path, relative: String): Path? = when {
    root.isDirectory() -> root.resolve(relative)
        .takeIf { it.exists() }

    root.isRegularFile() ->
        ZipFile(root.toFile()).use { zip ->
            root.resolve(relative)
                .takeIf { zip.getEntry(relative) != null }
        }

    else -> null
}
```

### Gradle version

9.6.0-20260420115815+0000

### Build scan URL (optional)

_No response_

_No response_

eskatos:
Snapshot versions no longer use a version range to find the nearest available source distribution. The resolver now requests the exact version, matching the running Gradle distribution. If it is not available, IDE source navigation falls back to decompiled classes.

The rationale is that it is better to provide no sources rather than wrong sources.

This is only about nightlies.

* Fixes https://github.com/gradle/gradle/issues/37819

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/kotlin-dsl/src/integTest/kotlin/org/gradle/kotlin/dsl/resolver/SourceDistributionResolverIntegrationTest.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/integTest/kotlin/org/gradle/kotlin/dsl/resolver/SourceDistributionResolverIntegrationTest.kt
@@ -13,7 +13,6 @@ import org.gradle.util.GradleVersion
 import org.junit.Assume.assumeTrue
 import org.junit.Test
 import java.io.File
-import java.util.UUID
 
 class SourceDistributionResolverIntegrationTest : AbstractKotlinIntegrationTest() {
 
@@ -356,8 +355,9 @@ class SourceDistributionResolverIntegrationTest : AbstractKotlinIntegrationTest(
                     """
                 )
 
-                primaryServer.expectGetMissing("/${repositoryName}/")
-                fallbackServer.expectGetMissing("/${repositoryName}/")
+                val artifactFileName = gradleVersion.artifactFileName
+                primaryServer.expectHeadMissing("/$repositoryName/$artifactFileName")
+                fallbackServer.expectHeadMissing("/$repositoryName/$artifactFileName")
 
                 build(fallbackRepoOverride).apply {
                     assertOutputContains("Could not resolve Gradle distribution sources. See debug logs for details.")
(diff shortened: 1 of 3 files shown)


eskatos:
@bot-gradle test this

eskatos:
@bot-gradle test this

reinsch82 (approved):

hegyibalint (approved):
Nice work! LGTM!

PR #37823, opened 2026-05-06
Reduce running time of IP Sync Android perf tests

alllex:
Fixes https://github.com/gradle/gradle-private/issues/5197

Steps:
* Use a smaller android-100-kts project, who is a cousine of android-500-kts (see https://github.com/gradle/isolated-projects-testbed/pull/8)
* Use fewer measurement iterations, because previous runs demonstrated stable enough values
* Drop Moderne+CoD baseline  from perf comparison test, because it doesn't differ from the just-Moderne baseline for the current projects/scenarios we are targeting

---

In the updated setup, the `org.gradle.performance.regression.ip.IsolatedProjectsAndroidSyncPerformanceRegressionTest.sync Studio after build logic ABI change with cold daemon` scenario takes ~21m with 5 warmups and 5 iterations. The iterations are now ~14s each, and the main cost comes from the very first warmup that takes ~6m, downloading all dependencies.

See:
- https://builds.gradle.org/buildConfiguration/Gradle_Master_Util_Performance_AdHocPerformanceScenarioLinuxAmd64/112478360
- https://builds.gradle.org/buildConfiguration/Gradle_Master_Util_Performance_AdHocPerformanceScenarioLinuxAmd64/112479124

Let's not include this in the Ready for Nightly just yet and see how it fairs in practice in Ready For Release.

### DIFF
--- a/.teamcity/performance-tests-ci.json
+++ b/.teamcity/performance-tests-ci.json
@@ -13,7 +13,7 @@
       }
     } ]
   }, {
-    "testId" : "org.gradle.performance.crossbuild.ip.IsolatedProjectsAndroidSyncPerformanceComparisonTest.sync Studio after included build logic refactoring",
+    "testId" : "org.gradle.performance.crossbuild.ip.IsolatedProjectsAndroidSyncPerformanceComparisonTest.sync Studio after build logic ABI change",
     "groups" : [ {
       "testProject" : "android500Kts",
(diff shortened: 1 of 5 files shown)


blindpirate (approved):
LGTM if it could finish in an acceptable duration (like <1h).

cobexer:

PR #37824, opened 2026-05-06
Fix NPE when computing failure paths through a cyclic dependency graph

### LINKED ISSUE #36284: Runtime Error 'Cannot invoke "java.util.List.get(int)" because "path" is null' when getting message of ModuleVersionResolveException

big-guy:
Fixes #36284.

## Summary

`DependencyGraphPathResolver.calculatePaths` could insert `null` entries into the paths it returned to `ModuleVersionResolveException.withIncomingPaths`, causing `getMessage()` to NPE at `path.get(0)` whenever a user inspected failures from a lenient `ArtifactView`. The originally reported reproducer used `androidx.room:room-*:2.8.3`, which is published as Kotlin Multiplatform: variant-aware resolution produces a cycle in the resolved component graph (`room-runtime ↔ room-common-jvm`), and the path resolver's interleaved single-pass traversal couldn't compute shortest paths correctly when the reverse-edge graph contained cycles.

The fix replaces that traversal with an independent BFS in the reverse (`getDependents()`) graph for each directDependee. BFS handles cycles naturally and is O(V+E) per source.

## What changed

- **`DependencyGraphPathResolver.java`** — rewrote `calculatePaths` to do a per-directDependee BFS in the reverse graph, reconstructing the shortest path via a predecessor map. The output format is unchanged (`[rootDisplayName, ...intermediate componentIds, directDependee componentId]`), so existing callers and message-format tests are unaffected.
- **`ResolutionIssuesIntegrationTest`** — added the verbatim androidx.room reproducer from the issue, gated on `TestEnvironmentPreconditions.Online`.
- **`LenientArtifactViewIntegrationTest`** — added a synthetic four-module Maven cycle (`mod-a → mod-b → mod-c → mod-d → mod-a`, with `mod-b` and `mod-d` depending on a missing module) that triggers the same NPE without requiring network access.

## Test plan

- [x] `:dependency-management:embeddedIntegTest --tests "org.gradle.integtests.resolve.LenientArtifactViewIntegrationTest"` passes (synthetic cyclic test).
- [x] `:dependency-management:embeddedIntegTest --tests "org.gradle.integtests.resolve.ResolutionIssuesIntegrationTest"` passes (real androidx.room reproducer + 4 pre-existing tests, 1 `@Ignore` skipped).
- [x] `:dependency-management:test --tests "*.DependencyGraphBuilderTest"` (33 tests exercising `calculatePaths`) passes.
- [x] `:dependency-management:test --tests "*.ModuleVersionResolveExceptionTest" --tests "*.ModuleVersionNotFoundExceptionTest"` passes (message-format tests unchanged).
- [x] Verified the new synthetic test reproduces the NPE when the fix is reverted.

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/LenientArtifactViewIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/LenientArtifactViewIntegrationTest.groovy
@@ -18,6 +18,7 @@ package org.gradle.integtests.resolve
 
 import org.gradle.api.internal.artifacts.verification.DependencyVerificationFixture
 import org.gradle.integtests.fixtures.AbstractHttpDependencyResolutionTest
+import spock.lang.Issue
 
 /**
  * Tests the behavior of {@link org.gradle.api.artifacts.ArtifactView} when it is configured to be lenient.
@@ -249,6 +250,64 @@ class LenientArtifactViewIntegrationTest extends AbstractHttpDependencyResolutio
         executer.expectDocumentedDeprecationWarning("The MavenArtifactRepository.artifactUrls(Object...) method has been deprecated. This is scheduled to be removed in Gradle 10. Consult the upgrading guide for further information: https://docs.gradle.org/current/userguide/upgrading_version_9.html#deprecated_maven_artifact_urls")
         succeeds("resolveLenient")
         outputContains("[]")
+    }
+
+    @Issue("https://github.com/gradle/gradle/issues/36284")
+    def "can read failure messages from lenient artifact view when failing module's reverse-dependency graph contains a cycle"() {
+        // Two transitive modules (modB and modD) directly depend on a missing module ("broken").
+        // The four cycle modules (modA -> modB -> modC -> modD -> modA) form a cycle in the
+        // resolved graph. With this layout, modB and modD are the directDependees of broken,
+        // and modD's only path back to root passes through modC -> modB -> modA, which the
+        // pre-fix DependencyGraphPathResolver could not compute correctly because modC was
+        // popped from the worklist before modB had been resolved.
+        def broken = mavenHttpRepo.module("org", "broken", "1.0")
+        def modA = mavenHttpRepo.module("org", "mod-a", "1.0")
+            .dependsOn("org", "mod-b", "1.0")
+            .publish()
+        def modB = mavenHttpRepo.module("org", "mod-b", "1.0")
+            .dependsOn("org", "mod-c", "1.0")
+            .dependsOn("org", "broken", "1.0")
+            .publish()
+        def modC = mavenHttpRepo.module("org", "mod-c", "1.0")
+            .dependsOn("org", "mod-d", "1.0")
+            .publish()
+        def modD = mavenHttpRepo.module("org", "mod-d", "1.0")
+            .dependsOn("org", "mod-a", "1.0")
+            .dependsOn("org", "broken", "1.0")
+            .publish()
+
+        buildFile.text = """
+            plugins {
+                id("java-library")
+            }
+            repositories {
+                maven { url = '${mavenHttpRepo.uri}' }
+            }
+            dependencies {
+                implementation("org:mod-a:1.0")
+            }
+            tasks.register("printFailures") {
+                def failures = configurations.runtimeClasspath.incoming.artifactView {
+                    lenient = true
+                }.artifacts.failures
+                doLast {
+                    failures.each { println("FAILURE: \${it.message}") }
+                }
+            }
+        """
+
+        expect:
+        modA.pom.expectGet()
+        modB.pom.expectGet()
+        modC.pom.expectGet()
+        modD.pom.expectGet()
+        broken.pom.expectGetMissing()
+        modA.artifact.expectGet()
+        modB.artifact.expectGet()
+        modC.artifact.expectGet()
+        modD.artifact.expectGet()
+        succeeds("printFailures")
+        outputContains("FAILURE: Could not find org:broken:1.0.")
     }
 
     private void withRepo() {
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/LenientArtifactViewIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/LenientArtifactViewIntegrationTest.groovy
@@ -286,7 +286,7 @@ class LenientArtifactViewIntegrationTest extends AbstractHttpDependencyResolutio
                     lenient = true
                 }.artifacts.failures
                 doLast {
-                    failures.each { println("FAILURE: \${it.message}") }
+                    failures.each { println(it.message) }
                 }
             }
         """
@@ -304,7 +304,7 @@ class LenientArtifactViewIntegrationTest extends AbstractHttpDependencyResolutio
         succeeds("printFailures")
 
         then:
-        outputContains("FAILURE: Could not find org:broken:1.0.")
+        outputContains("Could not find org:broken:1.0.")
     }
 
     private void withRepo() {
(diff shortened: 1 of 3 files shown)


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/LenientArtifactViewIntegrationTest.groovy
```diff
@@ -251,6 +252,64 @@ class LenientArtifactViewIntegrationTest extends AbstractHttpDependencyResolutio
...
+                def failures = configurations.runtimeClasspath.incoming.artifactView {
+                    lenient = true
+                }.artifacts.failures
+                doLast {
+                    failures.each { println("FAILURE: \${it.message}") }
+                }
+            }
+        """
+
+        expect:
```
jvandort: Can we instead have 

```
when:
// artifact expectations
succeeds("printFailure")

then:
outputContains(...)
...

File: platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/LenientArtifactViewIntegrationTest.groovy [hunk 2]
```diff
@@ -251,6 +252,64 @@ class LenientArtifactViewIntegrationTest extends AbstractHttpDependencyResolutio
...
+            .dependsOn("org", "broken", "1.0")
+            .publish()
+
+        buildFile.text = """
+            plugins {
+                id("java-library")
+            }
+            repositories {
+                maven { url = '${mavenHttpRepo.uri}' }
+            }
```
jvandort:
```suggestion
            ${mavenHttpRepo}
```

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ResolutionIssuesIntegrationTest.groovy
```diff
@@ -232,6 +233,51 @@ class ResolutionIssuesIntegrationTest extends AbstractIntegrationSpec {
...
+    @Requires(TestEnvironmentPreconditions.Online)
+    @Issue("https://github.com/gradle/gradle/issues/36284")
+    def "lenient artifact view exposes failure messages from androidx.room reproducer without NPE"() {
+        // Verbatim reproducer from issue 36284. With the bug, getMessage() throws
+        // NPE because DependencyGraphPathResolver.calculatePaths returns a list
+        // containing a null entry for a directDependee whose incoming edges no
+        // longer chain back to the root component.
+        buildFile << """
+            plugins {
+                id("java")
```
jvandort:
```suggestion
                id("java-library")
```

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ResolutionIssuesIntegrationTest.groovy
```diff
@@ -232,6 +233,51 @@ class ResolutionIssuesIntegrationTest extends AbstractIntegrationSpec {
...
+
+            ${mavenCentralRepository()}
+            ${googleRepository()}
+
+            dependencies {
+                implementation("androidx.room:room-sqlite-wrapper:2.8.3")
+                implementation("androidx.room:room-runtime:2.8.3")
+            }
+
+            abstract class DependencyGraphPrinter extends DefaultTask {
```
jvandort: Do we need a custom task implementation for this or can we print the failures in the `doLast` like the other test?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/DependencyGraphPathResolver.java
```diff
@@ -37,57 +38,79 @@ public static Collection<List<Describable>> calculatePaths(
...
+        Deque<DependencyGraphComponent> queue = new ArrayDeque<>();
+        queue.add(start);
+        boolean reachedRoot = false;
+        while (!queue.isEmpty()) {
+            DependencyGraphComponent current = queue.removeFirst();
+            if (current == root) {
+                reachedRoot = true;
+                break;
+            }
+            for (DependencyGraphComponent next : current.getDependents()) {
```
jvandort, line 89: Probably not worth doing it now since this fixes an NPE, and changing it would likely affect a number of test expectations, but this algorithm really should be only walking non-constraint edges, since constraints don't "pull in" a node, so constraint paths aren't really useful to the user here. 

Might be worth it to add a TODO for that

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/DependencyGraphPathResolver.java
```diff
@@ -37,57 +38,79 @@ public static Collection<List<Describable>> calculatePaths(
...
         }
-        return paths;
+
+        // Output format expected by ModuleVersionResolveException#getMessage:
+        // [<rootDisplayName>, <intermediate componentIds...>, <directDependee componentId>]
+        // The chain holds [start, ..., root] so we walk it in reverse, emitting rootDescribable
+        // first and skipping root's componentId.
+        List<Describable> result = new ArrayList<>(chain.size());
+        result.add(rootDescribable);
+        DependencyGraphComponent[] reversed = chain.toArray(new DependencyGraphComponent[0]);
```
jvandort: Whats with this `toArray` here? Doesn't seem necessaray

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/DependencyGraphPathResolver.java
```diff
@@ -37,57 +38,79 @@ public static Collection<List<Describable>> calculatePaths(
...
+        for (int i = reversed.length - 2; i >= 0; i--) {
```
jvandort: Reversing a list, then iterating it in reverse? This seems more complicated than it needs to be. Am I missing something?

jvandort (approved):

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/DependencyGraphPathResolver.java
big-guy: Could a constraint change the version of what we're looking for and that would be meaningful to see?

jvandort: It could potentially be interesting though this logic finds the _first_ path to the node so I'd argue if we're showing one edge (of which there should be at least one hard edge) the hard one is the more interesting one to show

PR #37827, opened 2026-05-07
Fix dependency configurations chart column

### LINKED ISSUE #37816: 2-column table formatted with 3 columns
### Issue type

Styling or Accessibility

### Problem description

The table in https://docs.gradle.org/current/userguide/dependency_configurations.html#sub:what-are-dependency-configurations is shown with three columns even though the data only has two.

Image

### Context (optional)

_No response_

### Page with the problem

https://docs.gradle.org/current/userguide/dependency_configurations.html#sub:what-are-dependency-configurations

Dreeam-qwq:
Fix the dependency configurations chart column.

Before:
PixPin_2026-05-06_21-41-23

After:
PixPin_2026-05-06_21-40-37

### Context
Closes https://github.com/gradle/gradle/issues/37816


ljacomet (approved):

This PR has a valid DCO and is documentation only.

Issue #22949, opened 2022-12-02
IDE import fails with Project Isolation enabled: Expected unreportedProblemInCurrentCall to be called after enterDynamicCall


h0tk3y:
If Project Isolation is enabled via a Gradle property `org.gradle.unsafe.isolated-projects=true`, then importing the project in IntelliJ IDEA fails with the following message:

```
Expected unreportedProblemInCurrentCall to be called after enterDynamicCall
```

<img width="1080" alt="image" src="https://user-images.githubusercontent.com/1888526/205258147-96f71ac6-7fcf-4094-aba8-4f9ff491ddfd.png">

### Expected Behavior
IntelliJ IDEA imports the project correctly with `org.gradle.unsafe.isolated-projects=true`.

### Current Behavior
During the import, Gradle throws an exception. The stack trace of the exception is:

<details>
  <summary>Stack trace</summary>

```stacktrace
com.intellij.openapi.externalSystem.model.ExternalSystemException: Expected unreportedProblemInCurrentCall to be called after enterDynamicCall
	at org.jetbrains.plugins.gradle.model.ProjectImportAction.addBuildModels(ProjectImportAction.java:412)
	at org.jetbrains.plugins.gradle.model.ProjectImportAction.execute(ProjectImportAction.java:138)
	at org.jetbrains.plugins.gradle.model.ProjectImportAction.execute(ProjectImportAction.java:42)
	at org.gradle.tooling.internal.consumer.connection.InternalBuildActionAdapter.execute(InternalBuildActionAdapter.java:64)
	at org.gradle.tooling.internal.provider.runner.AbstractClientProvidedBuildActionRunner$ActionAdapter.runAction(AbstractClientProvidedBuildActionRunner.java:131)
	at org.gradle.tooling.internal.provider.runner.AbstractClientProvidedBuildActionRunner$ActionAdapter.fromBuildModel(AbstractClientProvidedBuildActionRunner.java:104)
	at org.gradle.tooling.internal.provider.runner.AbstractClientProvidedBuildActionRunner$ActionAdapter.fromBuildModel(AbstractClientProvidedBuildActionRunner.java:84)
	at org.gradle.internal.buildtree.DefaultBuildTreeModelCreator.fromBuildModel(DefaultBuildTreeModelCreator.java:57)
[...]

PR #37830, opened 2026-05-07
Deprecate implicit parent-project property/method lookup

### LINKED ISSUE #32611: Closures creating objects in object containers results in project isolation failure
### Current Behavior

Failure when Isolated Projects is enabled

```
2 problems were found storing the configuration cache.
- Build file 'lib/build.gradle': line 10: Project ':lib' cannot dynamically look up a property in the parent project ':'
- Build file 'lib/build.gradle': line 14: Project ':lib' cannot dynamically look up a property in the parent project ':'
```

### Expected Behavior

Success when Isolated Projects is enabled

### Context (optional)

Gradle build.gradle scripts allow you to create new objects inside a closure by:

```groovy
configurations {
    shadowed
}

sourceSets {
    customSourceset
}
```

https://github.com/liutikas/gradle-project-isolation-closure-failure-repro/

1. `./gradlew lib:dependencies`
2. `./gradlew lib:dependencies -Dorg.gradle.unsafe.isolated-projects=true`

### Gradle version

8.13

### Build scan URL (optional)

https://scans.gradle.com/s/fn736pnzv56de

n/a

reinsch82:
Fixes https://github.com/gradle/gradle/issues/37922

## Summary

Two behavior changes for child references to properties/methods that the child build script doesn't declare locally:

- **Vintage (default mode):** the lookup still walks the parent and resolves, but it now emits a deprecation warning at the lookup site. The build behaves exactly as before; users get a soft, documented signal to migrate before Gradle 10.
- **Isolated Projects:** the parent walk is fully disabled. The child's `ExtensibleDynamicObject` is not wired to the parent's inherited scope, so a reference that previously resolved through the parent now fails with the standard `MissingPropertyException` / `MissingMethodException` — the same outcome Gradle 10 will ship by default. There is no Gradle-9 deprecation warning under IP, and no separate cross-project access violation for parent-walk references: the lookup never crosses a project boundary, so there's nothing to violate.

A new upgrade-guide section `deprecated_accessing_parent_project_properties` documents both, with migration paths: `gradle.properties` + `providers.gradleProperty(...)`, convention plugins, explicit `rootProject.ext.foo`, plus a transitional `findExtraInHierarchy` helper. The Kotlin DSL interaction with `val name: Type by project` is called out separately.

Implementation:
- `ProblemReportingCrossProjectModelAccess.parentProjectDynamicInheritedScope` returns `null` under IP, so `DefaultProject` skips parent wiring entirely; the `CrossProjectModelAccessTrackingParentDynamicObject` wrapper goes away as dead code.
- For Vintage, `DefaultProject` wires the parent and the existing `setFailOnParentAccess` strict-mode flag; the parent-walk deprecation always fires unless strict mode throws first.

## Stack

This is the first of three stacked PRs:

1. **This PR** — Vintage deprecation, IP behavior change, upgrade-guide section
2. _next_ — annotate the Vintage deprecation with caller-context detail (which API or DSL syntax reached the lookup)
3. _next_ — `NO_IMPLICIT_PARENT_PROPERTY_LOOKUP` feature preview, opting Vintage builds into the eventual Gradle 10 behavior early

Based on `master`.

## Test plan

- [x] `./gradlew :core:forkingIntegTest --tests "*ParentProjectPropertyLookup*"`
- [x] `./gradlew :configuration-cache:forkingIntegTest --tests "*IsolatedProjectsAccessFromGroovyDsl*" --tests "*IsolatedProjectsAccessFromKotlinDsl*" --tests "*IsolatedProjectsProblemReporting*"`
- [x] `./gradlew :core:test --tests "*DefaultProjectSpec*" --tests "*DefaultProjectTest*"`
- [x] `./gradlew sanityCheck`


subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
```diff
@@ -266,6 +267,10 @@ public DefaultProject(
         if (parentInherited != null) {
             extensibleDynamicObject.setParent(parentInherited);
             extensibleDynamicObject.setFailOnParentAccess(services.get(InternalOptions.class).getBoolean(FAIL_ON_PARENT_PROPERTY_LOOKUP));
+            // Suppress the Vintage deprecation under Isolated Projects: IP already reports a
+            // cross-project access violation for the same lookup, so emitting both would be
+            // duplicative.
+            extensibleDynamicObject.setDeprecateParentAccess(!services.get(BuildModelParameters.class).isIsolatedProjects());
```
jvandort: How do our IP tests handle this? 
If I execute a test which emits this deprecation in vintage, I would expect a test today to fail during IP integ test since it would not emit a deprecation

reinsch82:
@bot-gradle test this

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -51,6 +51,40 @@ Earlier versions rely on Gradle's implicit parent-project property lookup, which
...
+        for (def p = project; p != null; p = p.parent) {
+            if (p.ext.has(name)) {
+                return p.ext.get(name)
+            }
+        }
+        return null
+    }
+    ----
+
+    can be added to a build script or convention plugin to mimic the original walking semantics where strictly needed.
```
alllex: ❓ The `[source,groovy]` block is nested inside a bullet item without AsciiDoc list-continuation markers (`+`). Every other source block in this file is at top level, so I suspect the bullet list terminates at the blank line on line 68 — the code block renders detached, and the trailing "can be added to a build script…" paragraph drops out of the list item.

Could you double-check this with a docs build (`./gradlew :documentation:docs:userguide`)? If it does mis-render, the fix is either:

- add `+` continuations between the list item, the code block, and the trailing paragraph, or
- restructure so the helper sits outside the bullet list.

platforms/software/dependency-management/src/testFixtures/groovy/org/gradle/integtests/resolve/AbstractProjectDependencyConflictResolutionIntegrationSpec.groovy
```diff
@@ -119,6 +119,7 @@ abstract class AbstractProjectDependencyConflictResolutionIntegrationSpec extend
 """)
 
         then:
+        executer.noDeprecationChecks()
```
alllex: ❓ This blanket-suppresses **all** deprecation checks for every parameterized row in a fixture used by multiple subclasses (multi-project and composite-build variants). Looking at the generated build scripts (lines 88-118) I don't see an obvious parent-property lookup, so it isn't clear from the code which deprecation is being suppressed.

Could you spell out in a code comment which deprecation actually fires, and on which rows? If it's the new parent-property one this PR introduces, `executer.expectDocumentedDeprecationWarning(…)` would scope the suppression to the expected message; if it's something else entirely, that's a pre-existing problem worth its own note.

Worst case it stays as a blanket suppression — but it should be an informed decision documented in the code.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsAccessFromGroovyDslIntegrationTest.groovy
```diff
@@ -567,101 +603,29 @@ class IsolatedProjectsAccessFromGroovyDslIntegrationTest extends AbstractIsolate
...
-        then:
-        outputContains("My property: hello")
-
-        // an additional subproject demonstrates that the problems are not duplicated as the property lookup traverses up the project hierarchy
-        fixture.assertStateStoredAndDiscarded {
-            projectsConfigured(":", ":a", ":a:aa")
-            problem("Script 'a/aa/myscript.gradle': line 4: Project ':a:aa' cannot dynamically look up a property in the parent project ':a'")
-        }
+        failureCauseContains("Could not get unknown property 'foo' for project ':a'")
     }
```
alllex, line 659: 🤔 The pre-rewrite version of this test (issue #22949) exercised three `GroovyObject` dispatch operations — `getProperty`, `invokeMethod`, `setProperty` — plus a local `hasProperty` assertion. The new version only checks `getProperty`.

Worth restoring all three operations, each asserted to fail locally? A regression that re-wires only `invokeMethod` or `setProperty` to walk the parent again wouldn't be caught here, and the original test's value was specifically that all `GroovyObject` entry points get routed through `DefaultProject`'s overrides.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -286,16 +289,21 @@ private DynamicInvokeResult doTryInvokeMethod(String name, @Nullable Object... a
...
+                    + "Define '" + memberName + "' explicitly in " + getDisplayName()
+                    + ", or unset the org.gradle.internal.fail-on-parent-property-lookup system property."
+            );
+        }
+        DeprecationLogger.deprecateAction("Accessing a " + memberKind + " from a parent project")
+            .withContext(capitalize(memberKind) + " '" + memberName + "' was not found in " + getDisplayName()
+                + " and was dynamically resolved from " + resolvedParent.getDisplayName() + ".")
+            .willBecomeAnErrorInGradle10()
+            .withUpgradeGuideSection(9, "deprecated_accessing_parent_project_properties")
+            .nagUser();
```
alllex: 🤔 The current rendering is `Accessing a property from a parent project has been deprecated. …`. I think this wording is a bit misleading and can be more accurate. 

Alternative: `Implicitly resolving properties in project hierarchy has been deprecated`.

The `implicitly` or some similar word is important to highlight you can still get those properties, but explicitly. The `project hierarchy` is important, because it's not just about the parent project, but the lookup chain. Feel free to adjust the exact wording, but I'd like these concerns be represented in the final message.

For methods instead of properties the message is similar.

---

For `and was dynamically resolved from` part, I don't think `dynamically` is universally applicable for all cases. Is `project.findProperty("foo")` considered a dynamic lookup? I thought we reserved "dynamic" primarily for Groovy dynamic resolution, which is only a subset of code paths.

---

Once you've given thought to this comment. Please consider how we should adjust the deprecation userguide (achors, heading, paragraph wording, etc) to align with the deprecation message.

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
```diff
@@ -260,6 +260,11 @@ public DefaultProject(
         taskContainer = services.get(TaskContainerInternal.class);
         extensibleDynamicObject = new ExtensibleDynamicObject(this, Project.class, services.get(InstantiatorFactory.class).decorateLenient(services));
 
+        // Under Isolated Projects, parentProjectDynamicInheritedScope() returns null and we skip
+        // the parent wiring entirely: a child reference to a property/method that's defined only
+        // on the parent will fail with the standard MissingPropertyException/MissingMethodException
+        // instead of falling back to the parent. This matches the eventual Gradle 10 behavior.
+        // Vintage builds still walk the parent chain, with a deprecation warning at the lookup site.
```
alllex: 💅 The 5-line IP-vs-Vintage explanation here duplicates what's in the new comment in `ProblemReportingCrossProjectModelAccess.parentProjectDynamicInheritedScope`. Consider moving the explanation to the javadoc of `CrossProjectModelAccess.parentProjectDynamicInheritedScope` so the contract — *"may return null; when it does, the child skips parent-walk wiring"* — is documented where the method is defined. The call site here can then be a one-liner pointing at it.

platforms/native/plugins-model-native/src/integTest/groovy/org/gradle/model/SoftwareModelTaskAndBuildScriptIntegrationTest.groovy
```diff
@@ -41,7 +41,7 @@ class SoftwareModelTaskAndBuildScriptIntegrationTest extends AbstractIntegration
     @ToBeFixedForIsolatedProjects(because = "project cannot dynamically look up a method in the parent project")
     def "methods defined in project build script are visible to descendant projects when script contains only methods and model block"() {
         createDirs("child1")
-        settingsFile << "include 'child1'"
+        settingsFile << "rootProject.name = 'root'\ninclude 'child1'"
```
alllex: 🤔 `settingsFile << "rootProject.name = 'root'\ninclude 'child1'"` is fine but reads poorly. Triple-quoted GString form (like the build files just below) would be consistent with the rest of the test and easier to scan. Several of these were introduced by this PR. (Same pattern at `BuildScriptVisibilityIntegrationTest.groovy:27/72/121` and `ScriptPluginClassLoadingIntegrationTest.groovy:148/219`.)

subprojects/core/src/integTest/groovy/org/gradle/api/internal/project/ParentProjectPropertyLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,150 @@
...
+        file("a/build.gradle") << """
+            ext.foo = "child-only"
+            println("foo: " + foo)
+        """
+
+        expect:
+        succeeds("help")
+        outputContains("foo: child-only")
+    }
+}
```
alllex, line 184: 💭 The new file covers implicit `foo`, `findProperty`, `property`, and implicit method invocation — but not `getProperty('foo')` or `hasProperty('foo')` against the deprecation. `ParentPropertyLookupFailOnAccessIntegrationTest:75-99` exercises them indirectly, but adding two cases to this dedicated file would make the coverage discoverable. Not blocking.

alllex (commented):
Looks good overall. We need to polish the user-facing wording and consider the test coverage.

FYI, most comments were worded by Claude, but flagged and curated by me.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
lkasso: I fixed it.

lkasso (approved):
Docs LGTM with my patch.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -51,6 +51,75 @@ Earlier versions rely on Gradle's implicit parent-project property lookup, which
 Upgrade to the link:https://plugins.gradle.org/plugin/com.gradle.develocity[latest Develocity plugin] from the Gradle Plugin Portal.
 See the link:https://gradle.com/help/gradle-plugin-compatibility/[Develocity plugin compatibility matrix] for details.
 
+[[deprecated_accessing_parent_project_properties]]
+==== Accessing properties and methods from a parent project
+
```
alllex: 💭 We should phrase this around **implicit** lookup in project **hierarchy**

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -51,6 +51,75 @@ Earlier versions rely on Gradle's implicit parent-project property lookup, which
...
+    }
+    return null
+}
+----
++
+can be added to a build script or convention plugin to mimic the original walking semantics where strictly needed.
+
+For the Kotlin DSL specifically, the `val name: Type by project` delegate syntax — already deprecated in its own right (see link:#kotlin_dsl_delegated_properties[Removing Kotlin DSL property delegates]) — is equivalent to the explicit API:
+
+- `val foo: String? by project` corresponds to `project.findProperty("foo") as String?` and will silently return `null` for parent-only properties once the parent lookup is removed in Gradle 10. Audit usages with `grep -rn "by project" --include="*.kts"` before relying on the value.
```
alllex: ❌ In Gradle 10, the `by` delegation will be removed entirely, so this advice is misleading. I'm also not a fan of the AI-style suggestionstions to run grep with some flags.

I suggest we drop this bullet list from this PR, and consider how/whether to have it, once there is a feature preview that actually changes the behavior

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -51,6 +51,75 @@ Earlier versions rely on Gradle's implicit parent-project property lookup, which
...
+The migration paths below apply equally to standard builds (to silence the deprecation) and Isolated Projects builds (to restore the value).
+
+To migrate, choose the approach that best fits your use case:
+
+- **Build-wide configuration values** — declare them in `gradle.properties` at the root, and read them in subprojects via `providers.gradleProperty("name")`. They become typed `Provider<String>` values and are available everywhere without any walking.
+- **Plugin defaults / shared logic** — extract into a <<sharing_build_logic_between_subprojects.adoc#sharing_build_logic_between_subprojects, convention plugin>> applied to each subproject. Each subproject gets its own configuration; nothing crosses project boundaries.
+- **Explicit reference to the parent's value** — write `rootProject.ext.foo` (or `parent.ext.foo`) in the child build script. This is a transitional fix: it preserves the cross-project coupling and is therefore not Isolated Projects compatible, but it removes the deprecation warning and makes the dependency explicit. A user-defined helper such as
++
+[source,groovy]
```
alllex: 💭 Since we are currently Kotlin-first, we should also provide the Kotlin snippet. I think it's easier if we pull it out of the list and make it a dual Kotlin/Groovy snippet as usual. The Kotlin one can idiomatically be an extension function.

🤔 Since the snippet is non-trivial, I think it might be worth it to extract it outside of the guide and have it tested.

alllex (changes requested):

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
```diff
@@ -260,6 +260,8 @@ public DefaultProject(
         taskContainer = services.get(TaskContainerInternal.class);
         extensibleDynamicObject = new ExtensibleDynamicObject(this, Project.class, services.get(InstantiatorFactory.class).decorateLenient(services));
 
+        // See CrossProjectModelAccess#parentProjectDynamicInheritedScope for the IP-vs-Vintage contract:
+        // null means parent-walk is disabled (Isolated Projects), so we skip the parent wiring entirely.
```
octylFractal: :x: I don't know if this comment is really helpful? If anyone wants to know why this might be `null` it's not difficult to view the method and find out why.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
```diff
@@ -145,10 +145,8 @@ class ProblemReportingCrossProjectModelAccess(
     }
 
     override fun parentProjectDynamicInheritedScope(referrer: ProjectState): HierarchicalDynamicObject? {
-        val parent = referrer.parent ?: return null
-        return CrossProjectModelAccessTrackingParentDynamicObject(
-            parent, referrer.identity, ipProblems, coupledProjectsListener
-        )
+        // Isolated Projects: parent-walk is disabled — see CrossProjectModelAccess#parentProjectDynamicInheritedScope.
```
octylFractal: :x: The "see" is pretty unhelpful here, IMO.
```suggestion
        // Isolated Projects: parent-walk is disabled to match Gradle 10 behavior.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
lkasso: +1

subprojects/core/src/integTest/groovy/org/gradle/api/internal/project/ParentProjectPropertyLookupIntegrationTest.groovy
reinsch82: addressed by

- def "explicit getProperty(String) on the child project is deprecated when resolved from parent"() at line 106
- def "explicit hasProperty(String) on the child project is deprecated when resolved from parent"() at line 123

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -51,6 +51,63 @@ Earlier versions rely on Gradle's implicit parent-project property lookup, which
...
+[source,kotlin]
+----
+val foo: String? by project
+println(foo) // Resolved through hierarchy — now deprecated
+----
+=====
+====
+
+This implicit inheritance is deprecated and will be removed in Gradle 10.
+Both an implicit reference (e.g. a bare `foo` in a Groovy build script, or `val foo by project` in a Kotlin build script) and the explicit API (`project.findProperty("foo")`, `project.property("foo")`, `project.hasProperty("foo")`) emit the deprecation warning when they resolve through the hierarchy.
```
octylFractal:
```suggestion
Both an implicit reference (e.g. a bare `foo` in a Groovy build script, or `val foo by project` in a Kotlin build script) and the explicit API (`project.findProperty("foo")`, `project.property("foo")`, `project.hasProperty("foo")`) emit a deprecation warning when they resolve through the hierarchy.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -51,6 +51,63 @@ Earlier versions rely on Gradle's implicit parent-project property lookup, which
...
+The migration paths below apply equally to standard builds (to silence the deprecation) and Isolated Projects builds (to restore the value).
+
+To migrate, choose the approach that best fits your use case:
+
+- **Build-wide configuration values** — declare them in `gradle.properties` at the root, and read them in subprojects via `providers.gradleProperty("name")`. They become typed `Provider<String>` values and are available everywhere without any walking.
+- **Plugin defaults / shared logic** — extract into a <<sharing_build_logic_between_subprojects.adoc#sharing_build_logic_between_subprojects, convention plugin>> applied to each subproject. Each subproject gets its own configuration; nothing crosses project boundaries.
+- **Explicit reference to an ancestor's value** — write `rootProject.ext.foo` (or `parent.ext.foo`) in the child build script. This is a transitional fix: it preserves the cross-project coupling and is therefore not Isolated Projects compatible, but it removes the deprecation warning and makes the dependency explicit.
+
+If you need to keep the original walking semantics, the following helper can be added to a build script or convention plugin:
```
octylFractal, line 278: ❓ Can we note that this is incompatible with IP as well?

platforms/documentation/docs/src/snippets/releases/upgrading/findExtraInHierarchy/kotlin/child/build.gradle.kts
```diff
@@ -0,0 +1,13 @@
+// tag::helper[]
+fun Project.findExtraInHierarchy(name: String): Any? =
+    if (extra.has(name)) extra[name] else parent?.findExtraInHierarchy(name)
```
alllex: ❔ Any reason why the kotlin version is recursive, but the Groovy version is a loop? I am not sure I have a preference, but it's probably better if they show the same pattern. Maybe the recursive one is nicer, because it's more concise.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -51,6 +51,63 @@ Earlier versions rely on Gradle's implicit parent-project property lookup, which
 Upgrade to the link:https://plugins.gradle.org/plugin/com.gradle.develocity[latest Develocity plugin] from the Gradle Plugin Portal.
 See the link:https://gradle.com/help/gradle-plugin-compatibility/[Develocity plugin compatibility matrix] for details.
 
+[[deprecated_accessing_parent_project_properties]]
+==== Implicit lookup of properties and methods in the project hierarchy
```
alllex, line 229: ❔ Should we start with "Deprecation of" like the other titles?

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -51,6 +51,63 @@ Earlier versions rely on Gradle's implicit parent-project property lookup, which
 Upgrade to the link:https://plugins.gradle.org/plugin/com.gradle.develocity[latest Develocity plugin] from the Gradle Plugin Portal.
 See the link:https://gradle.com/help/gradle-plugin-compatibility/[Develocity plugin compatibility matrix] for details.
 
+[[deprecated_accessing_parent_project_properties]]
```
alllex: 💅 `deprecated_implicit_project_hierarchy_lookup`

alllex (approved):
LGTM!

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
lkasso: Yes, but I will clean it up as usual before the release anyways.

platforms/documentation/docs/src/snippets/releases/upgrading/findExtraInHierarchy/kotlin/child/build.gradle.kts
reinsch82: claude suggested it as more ideomatic.
no other particular reason.

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
octylFractal: I still think this should be changed.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
octylFractal: This is still unchanged.

octylFractal (changes requested):

platforms/documentation/docs/src/snippets/releases/upgrading/findExtraInHierarchy/kotlin/child/build.gradle.kts
bamboo: I agree with Alex. Both should probably use recursion because shorter and simpler to understand. The Kotlin version should be marked `tailrec`.

octylFractal (approved):

PR #37831, opened 2026-05-07
Annotate parent-walk diagnostics with caller-context detail

reinsch82:
## Summary

When a child build script reads a property or method defined on a parent project under Vintage, the parent-walk deprecation says *what* was resolved but not *how* the user reached the lookup site. The same lookup can come from a bare `foo` reference, a Kotlin DSL `val foo by project` delegate, or one of the explicit API methods `project.property("foo")` / `findProperty("foo")` / `getProperty("foo")` / `hasProperty("foo")`.

This PR threads a `CallerContext` through to `ExtensibleDynamicObject` so the deprecation message names the caller. The caller detail is added to the **context**.

For an explicit API call the message gains a trailing sentence:

```
Implicit lookup of properties in parent projects has been deprecated. This will fail with an error in Gradle 10. Property 'foo' was not declared in project ':child' and was resolved from root project 'root'. This lookup was initiated by 'property()'. Consult the upgrading guide ...
```

A bare reference / method invocation keeps the plain message (no trailing sentence):

```
Implicit lookup of properties in parent projects has been deprecated. This will fail with an error in Gradle 10. Property 'foo' was not declared in project ':child' and was resolved from root project 'root'. Consult the upgrading guide ...
```

Vintage-only by design: under Isolated Projects the parent walk is disabled entirely, so no caller context is ever read or rendered there.

## Wiring

- `DefaultProject`'s explicit-API overloads wrap their internal lookup in `withCallerContext(...)` for `PROPERTY`, `FIND_PROPERTY`, `GET_PROPERTY`, `HAS_PROPERTY`.
- `BasicScript` wraps Groovy script-level lookups with `BUILD_SCRIPT` (the bare-reference / method-invocation path).
- `PropertyDelegate.kt` wraps Kotlin DSL `val foo by project` delegate lookups with `KOTLIN_DELEGATION`.
- The context is set on the issuing object's `ExtensibleDynamicObject` immediately before the lookup and restored in a `finally`; `buildDeprecationContext` reads it when the walk resolves from a parent and appends the caller sentence only for explicit API calls.

## Notes

- The base parent-lookup deprecation (message wording, project attribution, upgrade-guide anchor) already lands on `master`; this PR only adds the caller-context detail on top.
- Because `getProperty('foo')` / `hasProperty('foo')` in a Groovy build script resolve to the script's own methods rather than the project's, those reach the lookup via `BUILD_SCRIPT` and render as implicit — only `property('foo')` / `findProperty('foo')` (which route to the project) render the explicit sentence. This is covered by the data-table test in `ParentProjectPropertyLookupIntegrationTest`.

## Test plan

- [x] `./gradlew :core:embeddedIntegTest --tests "*ParentProjectPropertyLookupIntegrationTest*" --tests "*ParentPropertyLookupFailOnAccessIntegrationTest*" --tests "*DynamicObjectIntegrationTest*"`
- [x] `./gradlew :model-core:test --tests "*ExtensibleDynamicObjectTest*"`
- [x] New `KotlinDslParentProjectPropertyLookupIntegrationTest` covers the `val ... by project` caller context

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
@@ -88,66 +88,41 @@ public class ExtensibleDynamicObject extends AbstractDynamicObject implements Hi
      * When a lookup walks into the parent, the context currently set on the referrer's dynamic
      * object is read to enrich the deprecation message.
      */
-    public abstract static class CallerContext {
-        private CallerContext() {}
-
-        public abstract boolean isExplicitApiCall();
+    public static final class CallerContext {
 
         /**
-         * The API method name (e.g. {@code "property()"}, {@code "findProperty()"}, or
-         * {@code "val ... by project"}). Only valid for explicit API calls.
+         * The phrase describing how the lookup was initiated, slotted into the deprecation message
+         * after {@code "This lookup was initiated by "}. Includes its own quoting where appropriate.
          */
-        public abstract String apiName();
+        private final String description;
+
+        private CallerContext(String description) {
+            this.description = description;
+        }
+
+        public String getDescription() {
+            return description;
+        }
 
         /**
          * Holder class to avoid initialization-order surprises between the outer class and its
-         * private subclasses.
+         * instances.
          */
         public static final class Instances {
             /** {@code project.property("foo")}. */
-            public static final CallerContext PROPERTY = new ExplicitApiCall("property()");
+            public static final CallerContext PROPERTY = new CallerContext("'property()'");
             /** {@code project.findProperty("foo")}. */
-            public static final CallerContext FIND_PROPERTY = new ExplicitApiCall("findProperty()");
+            public static final CallerContext FIND_PROPERTY = new CallerContext("'findProperty()'");
             /** {@code project.hasProperty("foo")}. */
-            public static final CallerContext HAS_PROPERTY = new ExplicitApiCall("hasProperty()");
+            public static final CallerContext HAS_PROPERTY = new CallerContext("'hasProperty()'");
             /** {@code project.getProperty("foo")}. */
-            public static final CallerContext GET_PROPERTY = new ExplicitApiCall("getProperty()");
+            public static final CallerContext GET_PROPERTY = new CallerContext("'getProperty()'");
             /** Kotlin DSL {@code val foo by project} property delegation. */
-            public static final CallerContext KOTLIN_DELEGATION = new ExplicitApiCall("val ... by project");
+            public static final CallerContext KOTLIN_DELEGATION = new CallerContext("'val ... by project'");
             /** Implicit reference in a Groovy build script (e.g. bare {@code foo} or {@code foo()}). */
-            public static final CallerContext BUILD_SCRIPT = new BuildScriptAccess();
+            public static final CallerContext BUILD_SCRIPT = new CallerContext("a dynamic invocation in the build script");
 
             private Instances() {}
-        }
-
-        private static final class ExplicitApiCall extends CallerContext {
-            private final String api;
-
-            ExplicitApiCall(String api) {
-                this.api = api;
-            }
-
-            @Override
-            public boolean isExplicitApiCall() {
-                return true;
-            }
-
-            @Override
-            public String apiName() {
-                return api;
-            }
-        }
-
-        private static final class BuildScriptAccess extends CallerContext {
-            @Override
-            public boolean isExplicitApiCall() {
-                return false;
-            }
-
-            @Override
-            public String apiName() {
-                throw new UnsupportedOperationException();
-            }
         }
     }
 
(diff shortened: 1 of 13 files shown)


reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -298,14 +395,29 @@ private void failOnParentAccessIfNeeded(String memberKind, String memberName, Dy
                     + ", or unset the org.gradle.internal.fail-on-parent-property-lookup system property."
             );
         }
-        DeprecationLogger.deprecateAction("Implicitly resolving " + pluralize(memberKind) + " in the project hierarchy")
-            .withContext(capitalize(memberKind) + " '" + memberName + "' was not declared in " + getDisplayName()
-                + " and was resolved from " + resolvedParent.getDisplayName() + ".")
+        DeprecationLogger.deprecateAction(buildDeprecationSummary(memberKind))
```
jvandort: The action must be a static string that does not vary between deprecation instances. Otherwise develocity will consider these different deprecation types. Any dynamic information must be in `withContext` or `withAdvice`

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProject.java
```diff
@@ -1200,7 +1205,18 @@ public void setProperty(String name, @Nullable Object value) {
 
     @Override
     public boolean hasProperty(String propertyName) {
-        return extensibleDynamicObject.hasProperty(propertyName);
+        return withCallerContext(ExtensibleDynamicObject.CallerContext.Instances.HAS_PROPERTY,
+            () -> extensibleDynamicObject.hasProperty(propertyName));
+    }
+
+    private <T> T withCallerContext(ExtensibleDynamicObject.CallerContext context, Supplier<T> action) {
```
jvandort, line 1212: This looks like it could be a method on `ExtensibleDynamicObject` and `BasicScript` could reuse it. That way we don't have to expose the raw `get/setCallerContext`.

I would also consider implementing this as a static `ThreadLocal` so you don't need to find the current extensible dynamic object in `BasicScript`.

Something like
```
class DynamicCallContext {

    public static interface CallerContext {

    } ...
    
    private static ThreadLocal<CallerContext> current ...

    public static void withCallerContext(Action ...) {
   
    }

    public static CallerContext current() {
        return current.get()
    }

}

```

Then, ExtensibleDynamicObject would just call `current()`. This is similar to what we do with `UserCodeApplicationContext`

subprojects/core/src/integTest/groovy/org/gradle/api/ParentPropertyLookupFailOnAccessIntegrationTest.groovy
```diff
@@ -81,23 +81,32 @@ class ParentPropertyLookupFailOnAccessIntegrationTest extends AbstractIntegratio
             println hasProperty('foo')
             println fooMethod()
         """)
-        5.times {
-            executer.expectDocumentedDeprecationWarning("Implicit lookup of properties in parent projects has been deprecated. " +
-                "This will fail with an error in Gradle 10. " +
-                "Property 'foo' was not declared in project ':child' and was resolved from root project 'root'. " +
-                "Consult the upgrading guide for further information: " +
-                "https://docs.gradle.org/current/userguide/upgrading_version_9.html#deprecated_implicit_lookup_in_parent_projects")
+        // `println foo`, `getProperty('foo')`, `hasProperty('foo')` reach the dynamic-object path
```
jvandort, line 86: for getProperty and hasProperty, if added another call but with `project.getProperty` and `project.hasProperty` would that give us the explicit caller context message?

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -296,13 +393,21 @@ private void failOnParentAccessIfNeeded(String memberKind, String memberName, Dy
...
+            .withContext(buildDeprecationContext(memberKind, memberName, resolvedParent))
             .willBecomeAnErrorInGradle10()
             .withUpgradeGuideSection(9, "deprecated_implicit_lookup_in_parent_projects")
             .nagUser();
     }
 
+    private String buildDeprecationContext(String memberKind, String memberName, DynamicObject resolvedParent) {
+        String context = capitalize(memberKind) + " '" + memberName + "' was not declared in " + getDisplayName()
+            + " and was resolved from " + resolvedParent.getDisplayName() + ".";
+        if (callerContext != null && callerContext.isExplicitApiCall()) {
```
jvandort: What if its not an explicit API call? Shouldn't we also tell the user it was initiated by a dynamic invocation?

jvandort (approved):
This LGTM for the most part. In the interest of time its probably not worth implementing the ThreadLocal part I mentioned, as this works as is. 

Am interested to see the tests that show `project.` and verify those print the `initiated by` messages.

subprojects/core/src/integTest/groovy/org/gradle/api/ParentPropertyLookupFailOnAccessIntegrationTest.groovy
reinsch82: yes it does

reinsch82:
@bot-gradle test this

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -73,6 +73,58 @@ public enum Location {
     @Nullable
     private HierarchicalDynamicObject parent;
     private boolean failOnParentAccess;
+    // Thread safety: only accessed while the project lock is held during configuration.
+    // Callers are expected to set/clear this around a single lookup using try/finally.
```
octylFractal, line 77: Why does this need a thread-safety note and not the other fields?

PR #37834, opened 2026-05-07
Annotate URL-based repository getUrl() as @Nullable

### LINKED ISSUE #37612: IvyArtifactRepository.getUrl can return null, but marked as not-null
### Current Behavior

`UrlArtifactRepository.getUrl` annotated as not nullable, but `IvyArtifactRepository.getUrl` can return null because of
https://github.com/gradle/gradle/blob/d9cbcfab390b6384808746f48e986b86bd5c9616/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/DefaultUrlArtifactRepository.java#L54-L56

### Expected Behavior

Nullability annotations should be alligned with reality

### Context (optional)

This breakes code for mirroring, e.g. 
```kotlin
if (this is UrlArtifactRepository && this.url != null) {
...
}
```

fails with
```
Script compilation error:

  Line 52:             if (this is UrlArtifactRepository && this.url != null) {
                                                            ^ Condition is always 'true'.
```

-

### Gradle version

9.6.0-nightly

### Build scan URL (optional)

_No response_

_No response_

devareddy05:
Fixes #37612

### Context

In the `@NullMarked` `org.gradle.api.artifacts.repositories` package, `UrlArtifactRepository.getUrl()` (and the overrides on `MavenArtifactRepository` and `IvyArtifactRepository`) is declared as non-nullable, but the implementation in `DefaultUrlArtifactRepository.getUrl()` explicitly returns `null` when no URL has been configured. This causes Kotlin static analysis to flag legitimate `repo.url != null` checks as `Condition is always 'true'`.

This change aligns the annotation with reality: `getUrl()` is marked `@Nullable` on `UrlArtifactRepository`, `MavenArtifactRepository`, and `IvyArtifactRepository`, plus the corresponding `Default*` implementations. Doc comments updated to mention the null case. The setters (`setUrl(URI)` / `setUrl(Object)`) are unchanged — null isn't a valid input.

The accepted-public-api-changes registry is updated to acknowledge the nullability widening on the three public interfaces, and `public-api-symmetrical-accessors-nullability.txt` records the intentional get/set asymmetry (matching the existing precedent for `HttpBuildCache.url`).

A small symmetry test (`DefaultMavenArtifactRepositoryTest.url is null by default`) was added to mirror the existing one on `DefaultIvyArtifactRepositoryTest`.

### Contributor Checklist
- [x] [Reviewed Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md).
- [x] All commits signed off (DCO).
- [x] Code distributable under Apache License 2.0.
- [x] "Allow edit from maintainers" enabled.
- [x] Unit test added for the symmetric Maven default-url behavior; the existing Ivy test (`default values`) already covers that side.
- [ ] Integration tests — n/a (annotation/contract change only, no behavior change).
- [x] User-facing change covered by Javadoc updates.
- [ ] `./gradlew sanityCheck` — not run locally (only JDK 21 available; Gradle build requires JDK 17). Requesting CI verification.
- [ ] `./gradlew :dependency-management:quickTest` — same.

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/DefaultIvyArtifactRepository.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/DefaultIvyArtifactRepository.java
@@ -250,6 +250,7 @@ public abstract class DefaultIvyArtifactRepository extends AbstractAuthenticatio
     }
 
     @Override
+    @Nullable
     public URI getUrl() {
         return urlArtifactRepository.getUrl();
     }
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/DefaultMavenArtifactRepository.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/DefaultMavenArtifactRepository.java
@@ -71,6 +71,7 @@ import org.gradle.internal.resource.local.FileResourceRepository;
 import org.gradle.internal.resource.local.FileStore;
 import org.gradle.internal.resource.local.LocallyAvailableResourceFinder;
 import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
 
 import javax.inject.Inject;
 import java.net.URI;
@@ -144,6 +145,7 @@ public abstract class DefaultMavenArtifactRepository extends AbstractAuthenticat
     }
 
     @Override
+    @Nullable
     public URI getUrl() {
         return urlArtifactRepository.getUrl();
     }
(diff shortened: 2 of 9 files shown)


ljacomet:

***

We need to make sure that this is not experienced as a breaking change by users.

big-guy:
We've added this to the 10.0 milestone because this is a small breaking change (compilation wise) for Kotlin users. Previously, you could reference `repository.url` and treat it as non-null in Kotlin (even if this could break at runtime).  If we make this nullable during 9.x, those consumers will no longer compile.

devareddy05:
> We've added this to the 10.0 milestone because this is a small breaking change (compilation wise) for Kotlin users. Previously, you could reference `repository.url` and treat it as non-null in Kotlin (even if this could break at runtime). If we make this nullable during 9.x, those consumers will no longer compile.

Thanks for the heads-up, that makes sense. I had registered the API change in `accepted-public-api-changes.json`, but hadn’t fully considered the Kotlin platform-type/source compatibility angle (`val u: URI = repo.url`).

10.0 sounds like the right milestone for this change. Happy to leave the PR as-is until then. Let me know if there’s anything I should adjust in the meantime (rebase target, release notes, etc.); otherwise, I’ll keep it in sync with `master`.

PR #37836, opened 2026-05-07
Sync: run with empty source to delete stale outputs

### LINKED ISSUE #37597: Sync task is skipped with NO-SOURCE if inputs become empty instead of deleting stale outputs
### Current Behavior

Given this `build.gradle.kts`:

```kotlin
tasks.register<Sync>("sync") {
  from(layout.projectDirectory.dir("in"))
  into(layout.projectDirectory.dir("out"))
}
```

And this folder structure (other files are empty):

```
├── build.gradle.kts
├── settings.gradle.kts
├── in
│   └── foo.txt
└── out
```

- Run: `gradle sync`

```
├── build.gradle.kts
├── settings.gradle.kts
├── in
│   └── foo.txt
└── out
    └── foo.txt
```
- Delete: `rm in/foo.txt`
- Run: `gradle sync`

```
├── build.gradle.kts
├── settings.gradle.kts
├── in
└── out
    └── foo.txt
```

### Expected Behavior

After running `Sync` on the now empty `in` folder, an empty `out` folder is expected:

```
├── build.gradle.kts
├── settings.gradle.kts
├── in
└── out
```

### Context (optional)

The `Sync` task action does not run, because `> Task :sync NO-SOURCE`.

This is due to the configuration inherited from `AbstractCopyTask` which adds a `skipWhenEmpty()` [here](https://github.com/gradle/gradle/blob/aec782af05a4c94a3dec185903ef0b45acf99892/subprojects/core/src/main/java/org/gradle/api/tasks/AbstractCopyTask.java#L93).

----

Notes:

- I would be interested if there is a workaround. Is it possible to reconfigure the `skipWhenEmpty()` with a `skipWhenEmpty(false)` in my task configuration?
- There was an issue for this, which was closed without a clear conclusion: #834

See description above.

### Gradle version

9.4.1

### Build scan URL (optional)

_No response_

_No response_

devareddy05:
Fixes #37597

### Context

`Sync` extends `AbstractCopyTask`, which registers each source file collection with `skipWhenEmpty()`. When all of a `Sync` task's sources are empty, Gradle reports `NO-SOURCE` and skips it — which leaves stale files in the destination, contradicting `Sync`'s documented contract that "the destination directory will only contain the files copied".

### Fix

`AbstractCopyTask` now exposes an overridable `skipWhenSourceIsEmpty()` hook (defaulting to `true`). The constructor's input registration only chains `.skipWhenEmpty()` when the hook returns `true`.

`Sync` overrides the hook to return `false`, so it always runs and the existing `SyncCopyActionDecorator` deletes stale outputs as documented. `Copy`, `AbstractArchiveTask`, and any third-party `AbstractCopyTask` subclass keep the previous skip-when-empty behavior.

### Tests

Added an integration test on `SyncTaskIntegrationTest` (`deletes stale outputs when source becomes empty`) that mirrors the issue's reproducer: populate `source/foo.txt`, sync, delete the source file, sync again, assert the destination is empty.

### Upgrade guide

Added a 9.6.0 entry under "Potential breaking changes" pointing users at `onlyIf { ... }` if they relied on the previous skip behavior.

### Contributor Checklist
- [x] [Reviewed Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md).
- [x] All commits signed off (DCO).
- [x] Code distributable under Apache License 2.0.
- [x] "Allow edit from maintainers" enabled.
- [x] Integration test added under `testing/integ-test/.../SyncTaskIntegrationTest`.
- [ ] Unit tests — n/a (the existing `SyncTest` already exercises the contract, and the change is a single conditional plus a hook).
- [x] User-facing change covered by an upgrade-guide entry.
- [ ] `./gradlew sanityCheck` — not run locally (only JDK 21 available; Gradle build requires JDK 17). Requesting CI verification.
- [ ] `./gradlew :integ-test:quickTest` — same.


devareddy05:
@bot-gradle  test this

ljacomet:

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -528,6 +528,14 @@ processResources {
 
 The embedded Kotlin has been upgraded from 2.3.20 to link:https://github.com/JetBrains/kotlin/releases/tag/v2.3.21[Kotlin 2.3.21].
 
+[[sync_runs_when_source_is_empty]]
+==== `Sync` task is no longer skipped when its source files are empty
+
+Previously, a link:{javadocPath}/org/gradle/api/tasks/Sync.html[`Sync`] task with no source files was reported as `NO-SOURCE` and skipped, leaving stale files in its destination directory.
+The task now runs even when its sources are empty, ensuring stale outputs are deleted as `Sync` is documented to do.
```
```suggestion
The task now runs even when its sources are empty, ensuring stale outputs are deleted as documented.
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -528,6 +528,14 @@ processResources {
...
+
+If you previously relied on the skip behavior, configure your task to short-circuit explicitly with `onlyIf { ... }`.
```
```suggestion
If you previously relied on the skip behavior, configure your task to short-circuit explicitly:

====
[.multi-language-sample]
=====
.build.gradle.kts
[source,kotlin]
----
tasks.named<Sync>("mySync") {
    onlyIf { !source.isEmpty }
}
----
=====
[.multi-language-sample]
=====
.build.gradle
[source,groovy]
----
tasks.named('mySync') {
    onlyIf { !source.empty }
}
----
=====
====
```

lkasso (commented):
Minor doc changes requested.

devareddy05:
Thanks for the review @lkasso ! Both suggestions applied in 33add40 ,  tightened the "as documented" wording and added the Kotlin/Groovy `onlyIf` examples.

lkasso (approved):
Docs are approved.

subprojects/core/src/main/java/org/gradle/api/tasks/AbstractCopyTask.java
```diff
@@ -115,6 +117,16 @@ protected CopySpecInternal createRootSpec() {
 
     protected abstract CopyAction createCopyAction();
 
+    /**
+     * Whether this task should be skipped with NO-SOURCE when its source files are empty.
+     * {@link Sync} overrides this to always run, since deleting stale outputs is part of its contract.
+     *
+     * @since 9.7.0
+     */
+    protected boolean skipWhenSourceIsEmpty() {
```
abstratt: ❌ 

This new API should be introduced as `@Incubating`.

https://docs.gradle.org/current/userguide/feature_lifecycle.html#sec:incubating_state

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -63,6 +63,36 @@ If your build requires shared, synchronized state, use <<build_services.adoc#bui
 
 For a complete list of types that are incompatible with the Configuration Cache, see the <<configuration_cache_requirements.adoc#config_cache:requirements:disallowed_types, documentation>>.
 
+[[sync_runs_when_source_is_empty]]
```
abstratt, line 105: ❌ 

This needs to be moved into a "Upgrading from 9.6.0 and earlier" section.

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
```diff
@@ -71,6 +71,11 @@ public abstract class Sync extends AbstractCopyTask {
 
     private final PatternFilterable preserveInDestination = new PatternSet();
 
+    @Override
+    protected boolean skipWhenSourceIsEmpty() {
+        return false;
```
abstratt, line 76: 👍

subprojects/core/src/main/java/org/gradle/api/tasks/AbstractCopyTask.java
abstratt: ❓ 

@devareddy05  Why not making this a (lazy) property (then `Sync` overrides its convention with a different default).

testing/integ-test/src/integTest/groovy/org/gradle/integtests/SyncTaskIntegrationTest.groovy
```diff
@@ -28,6 +28,35 @@ import spock.lang.Issue
...
+
+        when:
+        file('source/foo.txt').delete()
+        run 'sync'
+
+        then:
+        executedAndNotSkipped ':sync'
+        !file('dest/foo.txt').exists()
+        file('dest').directory
+    }
```
abstratt, line 66: 💭 

We should add coverage for the UP-TO-DATE case (could be just another when/then pair here).

subprojects/core/src/main/java/org/gradle/api/tasks/AbstractCopyTask.java
abstratt: Also, we have the option of making this package-private for now, and not making it API at all. Did you consider that?

abstratt (changes requested):
Thanks for this. Made some suggestions for a different approach (API-wise)

subprojects/core/src/main/java/org/gradle/api/tasks/AbstractCopyTask.java
devareddy05: Thanks @abstratt. Rather than introduce new public API, I made `skipWhenSourceIsEmpty()` package-private so it is no longer part of the public surface. That removes the need for `@Incubating` entirely. Done in the latest commit.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
devareddy05: Thanks @abstratt. Good catch. Since this change ships in 9.7.0, I added a new top section `[[changes_9.7.0]] == Upgrading from 9.6.0 and earlier` with a `Potential breaking changes` subsection and moved the note there. The AGP and Configuration Cache notes stay in the 9.6.0 section since they shipped in 9.6.0.

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
devareddy05: Thanks @abstratt. Kept this override, now package-private to match the change on the base class.

subprojects/core/src/main/java/org/gradle/api/tasks/AbstractCopyTask.java
devareddy05: Thanks @abstratt. I went with the package-private option you suggested. Both `Sync` and `AbstractCopyTask` live in `org.gradle.api.tasks`, so the override still dispatches correctly without exposing anything publicly. I deliberately avoided a configurable lazy property here because whether to skip on empty source is a fixed contract of each task type, not something a build should reconfigure (a user settable flag would let someone toggle `Sync` back to the old behavior and reintroduce the stale output bug). Happy to revisit if you would prefer the property approach.

testing/integ-test/src/integTest/groovy/org/gradle/integtests/SyncTaskIntegrationTest.groovy
devareddy05: Thanks @abstratt. Added a third when/then pair: after the source stays empty, a re-run is now asserted as `skipped ':sync'` (UP-TO-DATE), confirming the task does not needlessly re-execute once the destination is already in sync.

abstratt:
@devareddy05 Sorry for the delay, but since this changes existing behavior, we need to think this through. We are discussing within the team whether we should instead have a property so folks can opt-in into the new behavior. Should have a decision about this next week.

abstratt:
This PR will be superseded by #38396.

PR #37841, opened 2026-05-07
Break configuration-problems-base depending on core

reinsch82:
## Summary
- `:configuration-problems-base` depends on `:core` only to back `BuildNameProvider` with a build-listener-based lookup. That cycle prevents `:core` from depending on `:configuration-problems-base`, which forces every `:core` call site that wants to report an Isolated Projects violation to go through a per-API bridge interface. Replacing `BuildNameProvider` with `buildStateRegistry.rootBuild.displayName.displayName` at the two call sites lets the cycle go away.
- After the cycle break, `:configuration-problems-base` no longer needs `implementation(:core)` or `api(:messaging)`.
- `:ide-problems` flips its dependency direction: `api(:core)` (because `DefaultProblemsReportCreator` now exposes `BuildStateRegistry` from `:core` in its public constructor) and `implementation(:configurationProblemsBase)` (no longer in the public surface).

Extracted from a larger change in #37622 by @jvandort. The listener-wiring changes from his PR (moving `IsolatedProjectsProblemsReporter` registration to ALL MODES, and having `IgnoringProblemsListener` / `ConfigurationCachePromoHandler` also implement `IsolatedProjectsProblemsListener`) are intentionally left out — they are needed by his BuildServices feature and by the follow-up `Project.getProperties()` IP-violation PR (#37782), but not by the cycle break itself, and treating IP problems as a CC-promo suppression signal is a separate decision that doesn't belong in this change. cc @alllex.

## Test plan
- [x] `./gradlew sanityCheck`
- [x] `./gradlew :configuration-cache:forkingIntegTest --tests \"*ConfigurationCachePromoIntegrationTest*\"`
- [x] `./gradlew :configuration-cache:forkingIntegTest --tests \"*IsolatedProjectsAccessFromGroovyDslIntegrationTest*\"`


platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/promo/ConfigurationCachePromoHandler.kt
```diff
@@ -138,4 +139,8 @@ internal class ConfigurationCachePromoHandler(
     private fun TaskExecutionGraph.hasIncompatibleTasks() = allTasks.any { !(it as TaskInternal).isCompatibleWithConfigurationCache }
 
     private fun runWithoutBuildDefinition() = rootBuildLayout.isBuildDefinitionMissing
+
+    override fun onIsolatedProjectsProblem(problem: PropertyProblem) {
+        problems.addIfNeeded(true)
```
alllex: ❌ I don't think this is right. We would still want to promote CC, if the build has IP problems. However, I think we are going a bit ahead of ourselves. We shouldn't be reporting IP problems, when we are not in the IP mode. At least, in there is no need for it in this particular change.

alllex (changes requested):

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/promo/ConfigurationCachePromoHandler.kt
reinsch82: gone

alllex (approved):

Issue #37753, opened 2026-04-30
OOM upgrading from 9.4.1 to 9.5.0


trask:
## OOM stack (daemon log)

```
Exception in thread "JNA Cleaner" java.lang.OutOfMemoryError: Java heap space
Unexpected exception thrown.
org.gradle.internal.remote.internal.MessageIOException: Could not read message from '/127.0.0.1:42332'.
    at org.gradle.internal.remote.internal.inet.SocketConnection.receive(SocketConnection.java:96)
    at org.gradle.internal.remote.internal.hub.MessageHub$ConnectionReceive.run(MessageHub.java:270)
    ...
Caused by: java.lang.OutOfMemoryError: Java heap space

Problem in daemon expiration check
java.lang.OutOfMemoryError: Java heap space
    at java.base/java.lang.StringUTF16.compress(StringUTF16.java:211)
    at java.base/java.lang.String.<init>(String.java:303)
    at java.base/java.io.DataInputStream.readUTF(DataInputStream.java:550)
    at o.g.internal.serialize.InputStreamBackedDecoder.readString(InputStreamBackedDecoder.java:78)
    at o.g.launcher.daemon.registry.DaemonStopEvent$Serializer.read(DaemonStopEvent.java:116)
    at o.g.launcher.daemon.registry.DaemonRegistryContent$Serializer.readStopEvents(DaemonRegistryContent.java:130)
    at o.g.launcher.daemon.server.DaemonRegistryUnavailableExpirationStrategy.checkExpiration(DaemonRegistryUnavailableExpirationStrategy.java:63)
    at o.g.launcher.daemon.server.MasterExpirationStrategy.checkExpiration(MasterExpirationStrategy.java:80)
    at o.g.launcher.daemon.server.Daemon$DaemonExpirationPeriodicCheck.run(Daemon.java:273)
```

The OOM is in the **daemon JVM itself** — not a test worker fork (tests are skipped via `-PskipTests=true`).

## Heap dump analysis (Eclipse MAT 1.16.1)

`HeapDumpOnOutOfMemoryError` produced a **5,452,521,642-byte** dump (objects: 82,518,156). Total live retained ≈ **3.16 GB**.

### Suspect 1 — 677 MB (21.4%): build-operation queue not draining

| | |
|---|---|
| Class | `java.util.concurrent.ThreadPoolExecutor` (Gradle's `DefaultExecutorFactory$TrackedManagedExecutor`) |
[...]

PR #37845, opened 2026-05-07
Only try to run as worker thread in DefaultBuildOperationQueue

octylFractal:
Fixes https://github.com/gradle/gradle/issues/37753

This introduces as new `tryRunAsWorkerThread` and uses it in `DefaultBuildOperationQueue` to avoid leaving worker runnables hanging. In the future, this should not rely on (only) an exponential backoff, but instead wait for a signal indicating a worker lease is available, or this whole class should be redesigned to move WorkerRunnable (or similar concept) into the context instead, so we don't need to start one for each BuildOperationQueue that is created.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ljacomet (approved):
This looks good

I also added two commits, one to fix the test and another mostly clearing up naming.
@octylFractal If you are ok with those, go ahead and merge.

octylFractal:
Thanks.

PR #37854, opened 2026-05-08
Add getLength() to FilePropertyVisitor.VisitState

snoopcheri:
This PR adds file size reporting to the file property visitor API, enabling consumers (such as the DV plugin) to retrieve file lengths when traversing resolved artifacts during Gradle's snapshot-based build operations.

The change is very small and doesn't have any measurable impact on the performance.

The DV plugin uses the new method to skip system calls for determining the file size on its own (see https://github.com/gradle/dv/pull/64736).


cobexer:
Related PRs:
* https://github.com/gradle/gradle/pull/37568
* https://github.com/gradle/gradle/pull/37848
* https://github.com/gradle/gradle/pull/37852

PR #37855, opened 2026-05-08
Fix layout access inside nested feature defaults in settings.gradle.kts

### LINKED ISSUE #37805: DCL: Settings default - layout access inside NDOC fails
### Current Behavior

```
error: ProjectLayout should be referenced only inside of project type default configuration blocks 
```
This error only occurs with settings.gradle.kts, not with settings.gradle.dcl

### Expected Behavior

No error when accessing layout inside NDOC in settings.gradle.kts in defaults

### Context (optional)

_No response_

Checkout https://github.com/hfhbd/adventOfCode/tree/settings-error

All I did was renaming `settings.gradle.dcl` to `settings.gradle.kts`

### Gradle version

9.6.0-milestone-1

### Build scan URL (optional)

_No response_

_No response_

ghale:
Fixes #37805.                                                                                                                                                                
                                                                                                                                                                               
## Context                                                                                                                                                                   
                                                                                                                                                                               
In a `settings.gradle.kts`, accessing `layout` inside a project-type  default configuration block fails with:
                                                                                                                                                                               
  > ProjectLayout should be referenced only inside of project type                                                                                                             
  > default configuration blocks
                                                                                                                                                                               
…whenever the access sits inside a *nested* project-feature block. For example:                                                                                                                                                                     
                                                                                                                                                                               
```kotlin                                                                                                                                                                  
  defaults {
      kotlinJvmLibrary {
          dokka {                                                                                                                                                              
              sourceSets {                                                                                                                                                     
                  dclDokkaSourceSet("main") {                                                                                                                                  
                      samples = listOf(layout.projectDirectory.dir("..."))                                                                                                     
                  }                                                                                                                                                            
              }                                                                                                                                                                
          }                                                                                                                                                                    
      }           
  }
```

The same content saved as settings.gradle.dcl works correctly. The  asymmetry is because DCL resolves layout via its own DefaultsTopLevelReceiver, while the Kotlin DSL resolves layout through SharedModelDefaults.getLayout() — which is backed by a ThreadLocal<ProjectLayout> set/cleared around defaults application.                                                                                                          
                                                                                                                                                                               
### Root cause                                                                                                                                                                   
                                                                                                                                                                               
ActionBasedModelDefaultsHandler.apply brackets default-action  execution with setProjectLayout(...) / clearProjectLayout() on SharedModelDefaultsInternal. When a Kotlin DSL default action contains  a nested project-feature accessor, the runtime helper applyProjectFeature invokes DefaultProjectFeatureApplicator.registerFeatureApplicationFor, which calls applyDefaultsTo(...) for the inner feature — re-entering  apply. The inner apply's finally calls clearProjectLayout(), removing the ThreadLocal entry that the outer invocation still depends on. After the inner call returns, any subsequent layout.* reference in the outer action body resolves through a now-null ThreadLocal and throws.                                                                                                                                                      
                                                                                                                                                                               
### Fix                                                                                                                                                                          
                                                                                                                                                                               
Replace the unsafe setProjectLayout / clearProjectLayout pair on SharedModelDefaultsInternal with a single scoped method:
```                                                                                                                                                                               
  void withProjectLayout(ProjectLayout projectLayout, Runnable action);
```                                                                                                                                                                               
The implementation captures the previously bound layout, sets the new one, runs the action, and restores the previous binding (or removes the  ThreadLocal entry if there was none) on completion — including when the action throws. The sole caller in ActionBasedModelDefaultsHandler is migrated to the scoped form. This encodes the save/restore invariant in the SPI rather than at the call site, eliminating the class of bug entirely: future callers cannot reintroduce it by forgetting to save the previous binding.                                                                                                                                     
                                                                                                                                                                               
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/declarative-dsl-provider/src/integTest/groovy/org/gradle/internal/declarativedsl/WorkingWithFilesIntegrationTest.groovy
+++ b/platforms/core-configuration/declarative-dsl-provider/src/integTest/groovy/org/gradle/internal/declarativedsl/WorkingWithFilesIntegrationTest.groovy
@@ -176,10 +176,13 @@ class WorkingWithFilesIntegrationTest extends AbstractIntegrationSpec implements
             def type = projectType("testProjectType") {
                 definition {
                     property "id", String
+                    property "dirForType", DirectoryProperty
                     buildModel {
                         property "id", String
+                        property "dirForType", DirectoryProperty
                         mapping """
                                 model.getId().set(definition.getId());
+                                model.getDirForType().set(definition.getDirForType());
                             """
                     }
                 }
@@ -187,13 +190,13 @@ class WorkingWithFilesIntegrationTest extends AbstractIntegrationSpec implements
             projectFeature("feature") {
                 definition {
                     property "text", String
-                    property "dir", DirectoryProperty
+                    property "dirForFeature", DirectoryProperty
                     buildModel {
                         property "text", String
-                        property "dir", DirectoryProperty
+                        property "dirForFeature", DirectoryProperty
                         mapping """
                                 model.getText().set(definition.getText());
-                                model.getDir().set(definition.getDir());
+                                model.getDirForFeature().set(definition.getDirForFeature());
                             """
                     }
                 }
@@ -208,8 +211,9 @@ class WorkingWithFilesIntegrationTest extends AbstractIntegrationSpec implements
             defaults {
                 testProjectType {
                     feature {
-                        dir = layout.projectDirectory.dir("featureDefaultDir")
+                        dirForFeature = layout.projectDirectory.dir("featureDefaultDir")
                     }
+                    dirForType = layout.projectDirectory.dir("typeDefaultDir")
                 }
             }
         """.stripIndent()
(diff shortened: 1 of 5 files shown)


ghale:
@bot-gradle test this

platforms/core-configuration/declarative-dsl-provider/src/integTest/groovy/org/gradle/internal/declarativedsl/WorkingWithFilesIntegrationTest.groovy
```diff
@@ -168,6 +169,69 @@ class WorkingWithFilesIntegrationTest extends AbstractIntegrationSpec implements
...
+                    unsafeDefinition()
+                }
+            }
+        }.prepareToExecute()
+
+        def defaultsConfig = """
+            defaults {
+                testProjectType {
+                    feature {
+                        dir = layout.projectDirectory.dir("featureDefaultDir")
```
h0tk3y: 💭 It might also be a good idea to test that `layout` is usable in the defaults after the feature's block (i.e. it's not destructively cleared).

h0tk3y (approved):
👍 LGTM, left one note on a potentially good special case to test.

platforms/core-configuration/declarative-dsl-provider/src/integTest/groovy/org/gradle/internal/declarativedsl/WorkingWithFilesIntegrationTest.groovy
ghale: Good point.  Done.

PR #37859, opened 2026-05-08
Eliminate cross-project mutable state leaks via DefaultProject

alllex:
Stops `DefaultProject` from holding direct references to other projects' mutable state, so the cross-project model boundary is expressed only in terms of `ProjectIdentity`.

- Reshape `CrossProjectModelAccess` so every method takes `ProjectIdentity` instead of `ProjectInternal` / `ProjectState`. The default and isolated-projects implementations resolve identities to states via `ProjectStateLookup`, which now exposes `stateFor(Path)`.
- Drop `DefaultProject`'s `parent` and `rootProject` fields. The parent identity is derived from `owner.getParent()`, and `getRootProject` resolves through `ProjectStateLookup` on demand.
- Retire `ProjectIdentifier.getParentIdentifier()` and walk the `ProjectState` hierarchy in `DefaultProjectRegistry` instead.
- Simplify `IProjectFactory.createProject` / `ProjectFactory.createProject` by dropping the `parent` argument; `DefaultProject` no longer needs it.

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
@@ -43,7 +43,6 @@ import org.gradle.api.internal.project.ProjectIdentity
 import org.gradle.api.internal.project.ProjectInternal
 import org.gradle.api.internal.project.ProjectOrderingUtil
 import org.gradle.api.internal.project.ProjectRegistry
-import org.gradle.api.internal.project.ProjectState
 import org.gradle.api.internal.project.ProjectStateLookup
 import org.gradle.api.internal.tasks.TaskDependencyFactory
 import org.gradle.api.internal.tasks.TaskDependencyUsageTracker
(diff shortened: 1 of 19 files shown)


platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
```diff
@@ -94,29 +92,25 @@ class ProblemReportingCrossProjectModelAccess(
...
+    override fun access(referrer: ProjectIdentity, target: ProjectIdentity): ProjectInternal {
         // We purposefully leak mutable state here, as we wrap all mutable access with immediate failures in the case of cross-project access,
         // so there's no risk of race conditions.
-        return project.wrap(referrer, CrossProjectModelAccessInstance(DIRECT, project.projectIdentity))
-    }
-
-    override fun accessFromState(referrer: ProjectIdentity, projectState: ProjectState): ProjectInternal {
-        return projectState.fromMutableState { project ->
-            access(referrer, project)
+        return projectStateLookup.stateFor(target.buildTreePath).fromMutableState { project ->
```
octylFractal, line 106: Did you test with `ResilientKotlinDslScriptsModelBuilderCrossVersionSpec` or similar resilient TAPI tests? I couldn't make this change originally because these would fail, and I suspect that they still will with the current changes.

alllex: I haven't yet. I'll try it out

PR #37875, opened 2026-05-12
Strip $ prompt prefix when copying shell commands

runningcode:
When I was following the [steps in this guide](https://docs.gradle.org/current/userguide/gradle_wrapper.html#manually_verifying_the_gradle_wrapper_jar), I couldn't just copy paste the commands in to the terminal because they all had a leading `$`.

Here's a PR to strip the leading `$`


ljacomet:

This PR has a valid DCO and is documentation only.

PR #37879, opened 2026-05-13
Reduce internal Ant usage

reinsch82:
## Summary

First batch of removing Ant from Gradle's internal code paths. Each commit is independently reviewable.

- **`StartScriptGenerator`**: replace Ant's `Chmod` taskdef with `Files.setPosixFilePermissions`. Windows is an explicit no-op (matches the pre-existing Ant behavior — `Chmod.isValidOs()` already restricted execution to the Unix family, see [Ant 1.10.17 source](https://github.com/apache/ant/blob/5db231018603fbb23a66f43304003cb1451f20cc/src/main/org/apache/tools/ant/taskdefs/Chmod.java#L265-L269)). Drops `libs.ant` and `projects.ant` from `plugins-application`.
- **`JavaExec.setArgsString`**: swap Ant's `Commandline.translateCommandline` for commons-exec `CommandLine.parse`, which has the same shell-style tokenization (whitespace separator, both `'` and `"` quote delimiters, no backslash escapes). Adds a null/empty/whitespace guard so empty inputs continue to return an empty arg list. Adds `commons-exec 1.4.0` to the distribution catalog (its PGP key was already trusted in `verification-metadata.xml`). Drops `libs.ant` from `language-java`.
- **`TestFile.createZip(Object)`**: delete — unused dead method that drove Ant's `Zip` taskdef. No callers anywhere in `testing/`, `platforms/`, or `subprojects/`.
- **`TestFileHelper` (test fixture)**: replace Ant's `Zip`/`Tar`/`Untar`/`Expand` taskdefs and `CBZip2OutputStream` with commons-compress (already a dep). Read-only archive entries use `ZipArchiveEntry.setUnixMode` / `TarArchiveEntry.setMode`. Adds a new `TestFileHelperTest` covering round-trip zip/tar/tgz/tbz2/bzip2 and the read-only mode encoding. Drops `libs.ant` from `internal-testing`.

## Test plan

- [x] `:plugins-application:test`, `forkingIntegTest`, `configCacheIntegTest` covering `*StartScript*` and `*ApplicationPlugin*`
- [x] `:language-java:forkingIntegTest` and `configCacheIntegTest` covering `*JavaExecIntegrationTest*`
- [x] `:core:forkingIntegTest` covering `*ArchiveIntegrationTest*` and `*ConcurrentArchiveIntegrationTest*`
[...]

Issue #30111, opened 2024-08-07
Upgrading from Gradle 8 to Gradle 8.9 causes DirectoryScanner.removeDefaultExclude() to not work properly on macOS.


Ernest-L:
### Current Behavior

The project configured DirectoryScanner.removeDefaultExclude('**/.gitignore') in setting.gradle. After upgrading from Gradle 8.8 to Gradle 8.9, the "copy" task runs normally under Windows environment, but it prompts the following error on MacBook.
error:
![image](https://github.com/user-attachments/assets/c4761bd7-57f2-47bd-b948-59b2d7e36eb2)

### Expected Behavior

I hope it can run normally on Mac, thank you.

### Steps to Reproduce

Code sample：
build.gradle
task abc{
    doLast {
        copy {
            from fileTree('src/main/dist')
            into("build/test")
        }
    }

}
setting.gradle:
import org.apache.tools.ant.DirectoryScanner
DirectoryScanner.removeDefaultExclude('**/.gitignore')

### Gradle version

8.9

### Your Environment (optional)

cpu: apple m3 pro
os: Sonoma14.5

cobexer:
We appreciate the effort, but we must ask for more information so that we can triage it. 

We do not see an immediate root cause, and we need a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example) for the triage. If we have a reproducer, we may be able also to suggest workarounds or ways to avoid the problem.

The ideal way to provide a reproducer is to leverage our [reproducer template](https://github.com/gradle/gradle-issue-reproducer). You can also use [Gradle Project Replicator](https://github.com/android/project-replicator) to reproduce the structure of your project. 

This issue will be closed after 7 days unless you provide more information.

***
I tried your reproducer locally with 8.8, 8.9 and 8.10-rc1 and all of them work properly.

Please investigate your build and provide a reproducer as described above.

Issue #28802, opened 2024-04-12
Spring Boot Plugin: bootJar does not work when you change default excludes


christopherpicard:
### Current Behavior

Currently the plugin fails to run the bootJar with the following error:

Execution failed for task ':app-docker:bootJar'.
> Cannot change default excludes during the build. They were changed from [**/#*#, **/%*%, **/*~, **/.#*, **/.DS_Store, **/._*, **/.bzr, **/.bzr/**, **/.bzrignore, **/.cvsignore, **/.git, **/.git/**, **/.gitattributes, **/.gitignore, **/.gitmodules, **/.hg, **/.hg/**, **/.hgignore, **/.hgsub, **/.hgsubstate, **/.hgtags, **/.svn, **/.svn/**, **/CVS, **/CVS/**, **/SCCS, **/SCCS/**, **/vssver.scc] to [**/#*#, **/%*%, **/*~, **/.#*, **/.DS_Store, **/._*, **/.bzr, **/.bzr/**, **/.bzrignore, **/.cvsignore, **/.git, **/.git/**, **/.gitattributes, **/.gitmodules, **/.hg, **/.hg/**, **/.hgignore, **/.hgsub, **/.hgsubstate, **/.hgtags, **/.svn, **/.svn/**, **/CVS, **/CVS/**, **/SCCS, **/SCCS/**, **/vssver.scc]. Configure default excludes in the settings script instead.

However, this should not happen because the exclude changes is in the setting.gralde file. This means that somehow the plugin is running something before setting.gradle is loaded. I have tested it on many other plugins and all of them do not have this error.

### Expected Behavior

The tasks should complete normally.

### Context (optional)

Currently I have projects that need to copy files is the default exclude list. I have to comment out the exclude just to run the bootJar and then put it back in to run the docker container task. This is preventing me from having this project run in our CI/CD pipeline.

### Steps to Reproduce

1. Add the following to any project settings.gradle file:

- import org.apache.tools.ant.DirectoryScanner

- DirectoryScanner.removeDefaultExclude('**/.gitignore')

2. Run bootJar task.

### Gradle version

8.5

Issue #27225, opened 2023-11-30
Removing excludes from DirectoryScanner breaks configuration cache


mlopatkin:
### Current Behavior

Builds that calls `DirectoryScanner.removeDefaultExclude` fail with exception:
```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':compileJava'.
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:38)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:204)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:199)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:157)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:59)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.call(DefaultBuildOperationRunner.java:53)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor.call(DefaultBuildOperationExecutor.java:73)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter.execute(EventFiringTaskExecuter.java:52)
        at org.gradle.execution.plan.LocalTaskNodeExecutor.execute(LocalTaskNodeExecutor.java:42)
        at org.gradle.execution.taskgraph.DefaultTaskExecutionGraph$InvokeNodeExecutorsAction.execute(DefaultTaskExecutionGraph.java:331)
[...]

PR #37882, opened 2026-05-13
Deprecate Ant `DirectoryScanner` mutation and add `Settings.fileSystemDefaultExcludes`

reinsch82:
## Summary

Decouples Gradle's default file-system exclude patterns from `org.apache.tools.ant.DirectoryScanner` — the deepest entanglement keeping Ant on Gradle's runtime classpath. Refs [#11176](https://github.com/gradle/gradle/issues/11176) and [#11177](https://github.com/gradle/gradle/issues/11177), both open since 2019.

> **Rebased onto `master`** now that the prerequisite internal Ant-usage cleanup ([#37879](https://github.com/gradle/gradle/pull/37879)) has merged. The diff above is exclusively this change.

### What this change does

1. **`GradleDefaultExcludes`** (new constant in `platforms/core-runtime/files`) — verbatim port of Ant 1.10.17's `DirectoryScanner.DEFAULTEXCLUDES`. Gradle now owns the canonical list; a parity test guards against drift if Ant ever updates its list.

2. **`Settings.getFileSystemDefaultExcludes(): SetProperty<String>`** — new `@Incubating` public API (`@since 9.7.0`) giving build authors a first-class way to configure default excludes without importing Ant:
   ```kotlin
   // settings.gradle.kts
   fileSystemDefaultExcludes.add("**/node_modules")
   fileSystemDefaultExcludes.set(fileSystemDefaultExcludes.get() - "**/.gitignore")
   ```
   The property is created explicitly from `ObjectFactory` in the `DefaultSettings` constructor rather than as an abstract managed getter: the instantiator used for `Settings` (unlike the one for `Project`/task types) carries no `ManagedObjectRegistry`, so a synthesized managed `SetProperty` would fail at runtime.

3. **Documented deprecation** of `DirectoryScanner.add/removeDefaultExclude` mutations — fires from `DefaultFileSystemDefaultExcludesProvider.settingsEvaluated` via diff against the Gradle-owned baseline. A second warning fires when both APIs are used at once; the new Settings property always wins.

4. **Internal refactor** — `DefaultFileSystemDefaultExcludesProvider` and `VirtualFileSystemServices` now source pre-settings defaults from `GradleDefaultExcludes` rather than `DirectoryScanner`. Contributions from multiple settings scripts (composite/included builds) stack via a delta accumulator, preserving the historical process-global behavior.

5. **Docs migration** — user-guide snippets and prose updated to demonstrate the new API; the link to Ant's manual replaced with a link to the new javadoc. New upgrade-guide section at anchor `directoryscanner_default_excludes_deprecation`.

### Why now

The direction was a committed team intent in 2019 but never scheduled. The intervening years produced a recurring stream of downstream bug reports rooted in the same fragile coupling:

- [#30111](https://github.com/gradle/gradle/issues/30111) (2024) — `DirectoryScanner.removeDefaultExclude()` regression on macOS after 8.9
- [#28802](https://github.com/gradle/gradle/issues/28802) (2024) — Spring Boot `bootJar` broken when changing default excludes
- [#27225](https://github.com/gradle/gradle/issues/27225) (2023) — `DirectoryScanner` mutation breaks configuration cache
- [#26600](https://github.com/gradle/gradle/issues/26600) (2023) — `removeDefaultExclude` doesn't work with Distribution plugin
- [#13665](https://github.com/gradle/gradle/issues/13665), [#13427](https://github.com/gradle/gradle/issues/13427), [#11017](https://github.com/gradle/gradle/issues/11017)

### Deliberate non-goals

- **`PatternSpecFactory` still reads `DirectoryScanner` directly** during the build. To stay binary-behavior-compatible for the deprecation window, `DefaultFileSystemDefaultExcludesProvider` mirrors the resolved value back to `DirectoryScanner` when the new Settings API is used. The follow-up removal PR refactors `PatternSpecFactory` to listen exclusively, after which the mirror can go away.
- **`DirectoryScanner` stays on the runtime classpath** for now — it's still needed by Gradle's public `AntBuilder` API. Removing it is part of the broader Ant-removal arc, tracked separately.
- **Process-global stacking is preserved.** Each `Settings` has its own `fileSystemDefaultExcludes` property, but contributions stack across settings scripts to match the old process-global `DirectoryScanner` behavior — avoiding an additional breaking change. Whether precedence (new property vs Ant mutation) should be evaluated per-settings-script or per-session is called out as open design work.
- **Configuration-cache restore is set-based, not order-preserving.** The existing `ConfigurationCacheState` plumbing (predating this PR) serializes the excludes as an ordered list but restores them into a `Set`. Excludes behave as a set everywhere they're consumed, so this is correct today; tightening it to be order-stable is left as separate follow-up.

## Test plan

- [x] `:core:test --tests "*GradleDefaultExcludesTest*"` — parity guard against `DirectoryScanner.DEFAULTEXCLUDES` + immutability check
- [x] `:core:test --tests "*DefaultFileSystemDefaultExcludesProviderTest*"` — unit coverage for the cross-settings-script accumulator and the property-vs-Ant precedence rules (re-adding a removed default, cancelling an addition, both-APIs-ignored, CC restore)
- [x] `:core:test --tests "*DefaultSettingsTest*"` — `Settings` instantiation incl. the new property
- [x] `:snapshots:forkingIntegTest --tests "*DefaultExcludesIntegrationTest*"` — incl. new tests for the Settings API and the dual-config deprecation
- [x] `:snapshots:configCacheIntegTest --tests "*DefaultExcludesIntegrationTest*"` — same suite under configuration cache
- [x] `:docs:checkDeadInternalLinks` — verifies the new upgrade-guide anchor and updated cross-links
- [x] `sanityCheck` — binary compatibility, architecture tests, etc.
- [ ] Full CI to run

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheFileSystemDefaultExcludesTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheFileSystemDefaultExcludesTest.groovy
@@ -69,7 +69,7 @@ class ConfigurationCacheFileSystemDefaultExcludesTest extends AbstractConfigurat
         def excludedByBuildScriptFileCopy = file("build/output/${excludedByBuildScriptFileName}")
         excludedByBuildScriptFile.text = "input"
 
-        getBuildFile(dsl) << DefaultExcludesFixture.DefaultExcludesLocation.addDefaultExclude(excludedByBuildScriptFileName)
+        getBuildFile(dsl) << DefaultExcludesFixture.DefaultExcludesLocation.addDefaultExcludeForBuildScript(excludedByBuildScriptFileName)
 
         when:
         configurationCacheRun spec.copyTask
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/fixtures/DefaultExcludesFixture.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/fixtures/DefaultExcludesFixture.groovy
@@ -125,7 +125,23 @@ class DefaultExcludesFixture {
             dir.file(dsl.fileNameFor("settings"))
         }
 
-        static String addDefaultExclude(String excludedFileName) {
+        /**
+         * Snippet to be inserted into a settings.gradle(.kts) file to register an additional default exclude.
+         * Uses the modern Settings.fileSystemDefaultExcludes API.
+         */
+        static String addDefaultExcludeForSettings(String excludedFileName) {
+            """
+            fileSystemDefaultExcludes.add("**/${excludedFileName}")
+            """
+        }
+
+        /**
+         * Snippet to be inserted into a build.gradle(.kts) file to register an additional default exclude.
+         * Uses the deprecated org.apache.tools.ant.DirectoryScanner static-state API because there is no
+         * project-scoped equivalent of Settings.fileSystemDefaultExcludes. This snippet still works during
+         * the deprecation cycle for the few tests that need to exercise this build-script path.
+         */
+        static String addDefaultExcludeForBuildScript(String excludedFileName) {
             """
             ${DirectoryScanner.name}.addDefaultExclude("**/${excludedFileName}")
             """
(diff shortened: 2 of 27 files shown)


reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@coderabbitai review

lkasso (approved):
Docs approved.

PR #37884, opened 2026-05-13
Configuration Cache: Superset Task Selection Lookup (partial task invocation matches)

tresat:
The Configuration Cache (CC) now finds a hit when the current request's task list is a ordered subset of a previously stored entry's task list. `gradle a b c` followed by `gradle a` previously missed; now the second invocation reuses the first entry's state, pruning the loaded work graph down to `:a` (plus its required transitive dependencies).

## Behavior changes

- A second `gradle a` after `gradle a b c` reuses the CC entry and prints `Reusing configuration cache.` `:b` and `:c` are not scheduled.
- If the requested-task subset depends transitively on a dropped task (`a.dependsOn(b)`, request `[a]` against stored `[a, b, c]`), `b` is retained as part of `a`'s dependency closure — the build runs `:b` then `:a`.

## Limitations

Cases where superset reuse might seem applicable but is deliberately skipped — by design, for correctness. The request cold-stores in the usual way; nothing about the result is wrong, the optimization just doesn't apply.

- **Reordered subset.** `[a, b, c]` stored then `[b, a]` requested is a miss — same set, but not an order-preserving subsequence of the stored list.
- **Bare vs absolute CLI tokens don't share entries.** `gradle d` and `gradle :d` resolve to the same identity path in a single-project layout but their stored `cliTokens` lists (`["d"]` vs `[":d"]`) are distinct under verbatim matching. Each gets its own entry.
- **Multi-project bare names are exact-match-only.** A bare-name CLI request in a multi-project root (e.g. `["d"]` resolving to `[":d", ":sub:d"]`) is recorded with a non-1:1 mapping. Subsequent subset/superset requests against that entry cold-store, since the dropped identity-path set can't be derived unambiguously from the stored positional pairing.
- **Dropping a task with a `@Destroys` property.** A subset match that would prune a task declaring any `@Destroys`-annotated property is refused — its filesystem-clearing side effect is what retained tasks rely on. Catches the `:cleanSecond → :second` family via `Delete` (whose `getTargetFiles()` is `@Destroys`) as well as user-defined task types with `@Destroys` properties. Dynamic `task.destroyables.register(...)` on plain `DefaultTask` instances is not detected — annotation lookup only.
- **`mustRunAfter` / finalizer dangle.** A subset match that would leave a retained task pointing at a dropped task through `mustRunAfter` or a finalizer edge is refused — the loaded plan would either deadlock or silently execute the dropped task.
- **`-`-prefixed CLI tokens.** Any task list containing `--rerun`, `-x foo`, etc. skips the superset path entirely.
- **Empty requested-task list.** Project-defaults invocations (`gradle` with no tasks) skip the superset path — what the defaults resolve to isn't known at lookup time without running configuration.
- **Configuration logic that reads `startParameter.taskRequests` / `taskNames`.** Any read of `gradle.startParameter.taskRequests` or `gradle.startParameter.taskNames` at configuration time flags the resulting CC entry as superset-ineligible — the cached configuration state depends on the original CLI request, so reusing it for a different request would silently bake stale decisions. The flagged entry still works as an exact match.

### 1. Environment keys and how they are cached

These are not computed and stored separately from the full cache keys.

`ConfigurationCacheKey` is decomposed into two collaborating types:

- **`ConfigurationCacheEnvironmentKey`** (new) hashes every cache-key component *except* the requested task names.
- **`ConfigurationCacheKey`** now **composes** the env key (HAS-A) — it takes a `ConfigurationCacheEnvironmentKey` as its first constructor argument, calls `environmentKey.appendComponents(hasher)`, then appends `requestedTaskNames`.  Both classes are `@ServiceScope(Scope.BuildTree::class)` and registered in `BuildTreeModelControllerServices.kt`.

The per-env-key index lives at `<cc-root>/superset-index/<envKey>.bin` — written by `SupersetIndexFile` in a versioned binary format (magic header `CCSI`, then per-entry `fullKey` + `cliTokens` + `entryTaskIdentityPaths` + `taskGraphAccessed` + the `mustRunAfterEdges` edge-map + the identity-path set `sideEffectingTaskIdentityPaths`). Stale rows pointing at entry directories that have been evicted are self-healed on read. The `superset-index/` directory is excluded from the daily LRU cleanup sweep.

### 2. How task subset hits are determined

The index entry stores two parallel task lists:

- **`cliTokens`** — the user's CLI request, verbatim (`"a"`, `":foo"`, `":proj:bar"`). Used as the match key. Verbatim string comparison so bare and absolute task names cannot collide (`d` does not match `:d`).
- **`entryTaskIdentityPaths`** — identity paths of the scheduled entry tasks. Used to derive the set of tasks to prune at superset reuse. When `cliTokens.size == entryTaskIdentityPaths.size` the lists are positionally paired (the CLI token at index `i` resolved to the identity path at index `i`); when the sizes differ — a multi-project build where one bare CLI token resolved to multiple identity paths across subprojects — the mapping is ambiguous and the entry is treated as exact-match-only.

`SupersetIndexLookup.selectBestMatch(entries, requested)` is the pure selection rule — no I/O. Two-tier:

1. **Exact match wins.** The deduplicated request must equal the stored `cliTokens` as a list — same order included. `[a, a, b]` requested against `[a, b]` stored is a match (dedup); `[b, a]` requested against `[a, b]` stored is **not** (same set, but not a subsequence in the stored order).
2. **Otherwise smallest strict superset where the deduplicated request appears as a subsequence of the stored `cliTokens` list** (relative order preserved). Entries flagged `taskGraphAccessed = true` are excluded from this branch — they're exact-match-only. So are entries where `hasOneToOneCliMapping == false` (multi-project bare-name entries can't safely derive the dropped identity-path set). Tie-break among smallest-size supersets is the caller's responsibility (LRU via `FileAccessTimeJournal`).

The data types `SupersetIndexEntry` and `CompatibleEntry` are **nested inside `SupersetIndexLookup`** — the algorithm owns the data shapes.

`ConfigurationCacheRepository` is the I/O + locking orchestrator. `findCacheEntry` opens the env-key's index file inside `cache.withFileLock` (re-entrant — the same lock instance also covers the subsequent entry load), runs `selectBestMatch`, and then walks each candidate through three pruning gates (stale dir, dangling `mustRunAfter`/finalizer, side-effecting dropped task) until one is usable. Any token starting with `-` in the task list (`--rerun`, `-x foo`, etc.) short-circuits the whole superset path to `null` and falls back to exact-match, as does an empty requested-task list (project defaults — unknown at lookup time without running configuration).

### 3. Pruning the loaded work graph

After a non-exact-match hit, when `findCacheEntry` returns a strict-superset entry, `DefaultConfigurationCache.loadWorkGraph` derives `tasksToDrop` from the entry's CLI ↔ identity-path positional pairing: any stored CLI token not in the current request maps to the identity path at the same index. That set is passed through `cacheIO.readRootBuildStateFrom` and ends up in `WorkGraphPruner.pruneAndRewireInPlace`.

**`WorkGraphPruner`** (new class, in `org.gradle.internal.cc.impl`):

- Drop entry nodes whose `task.identityPath.toString()` is in `tasksToDrop`.
- BFS forward from the remaining entry nodes through `Node.dependencySuccessors`.
- Reconstruct `ScheduledWork(retainedScheduled, retainedEntries)`, where `retainedEntries` includes any original entry node that BFS retained as a transitive dependency.
- Clean up stale incoming edges: each retained node's `dependencyPredecessors` set is mutated to drop references to non-retained nodes.

The pruned `ScheduledWork` flows into `DefaultExecutionPlan.setScheduledWork`, which also repopulates `OrdinalGroup.entryNodes` from the work's entry-node set (the group's entry-node membership isn't serialized — the codec writes only the ordinal int and reconstructs the group via `OrdinalGroupFactory` — so without this restoration, ordinal-aware filtering and destroyer-location resolution misfire).

This is not `-x` semantics — transitive dependencies of requested tasks survive. A stored entry `[a, b, c]` where `a.dependsOn(b)`, requested as `[a]`, runs `:a` and `:b`.

**Fingerprint-miss fallback.** If the chosen superset entry's fingerprint is stale, store writes go to a fresh directory at `cacheKey.string` instead of the routed-to entry's directory, so the existing entry isn't corrupted.

### 4. Lookup-time safety gates

Picking a candidate that's a strict subset of a stored entry isn't unconditionally safe — pruning changes the executed schedule, and a few invariants can leak across the dropped/retained boundary. The lookup applies three gates per candidate (in `ConfigurationCacheRepository.pickUsableEntry`); a failing gate skips the candidate but leaves it in the index (still valid for exact-match reuse). Exact matches bypass all three since no pruning happens.

- **Stale entry directory** — the entry dir has been removed out from under the index (e.g. by external cleanup). The stale row is dropped and the index file is rewritten on the way out (self-heal).
- **`mustRunAfter` / finalizer dangle** (`hasDanglingMustRunAfter`) — a retained task carries a non-dependency hard-ordering edge to a dropped task. The loaded plan can't honor it: either the dropped task isn't in the plan (deadlock) or our BFS would pull it back in and it would run despite not being requested (different semantics from a fresh subset request). The lookup captures these edges at store time alongside the work graph.
- **Side-effecting dropped task** (`hasSideEffectingDroppedTask`) — dropping a task can change the filesystem state that retained tasks observe at execution time. A `:cleanSecond` task originally cleared a directory before `:second` wrote into it; subset-reusing the cached `:second` output without the cleanup leaves stale content that `OverlappingOutputs` detection would normally catch and re-execute against. The gate refuses any candidate whose dropped set contains a task recorded as side-effecting. The recorded set is identity paths of scheduled tasks whose type declares any `@Destroys`-annotated property — captured at store time by `TypeMetadataStore` annotation lookup over the scheduled `LocalTaskNode` list. `Delete` is the canonical hit (`getTargetFiles()` is `@Destroys`); user-defined task types with `@Destroys` properties are caught the same way. Dynamic `task.destroyables.register(...)` registrations on a plain `DefaultTask` are not detected.

Why a coarse `@Destroys`-property gate rather than a precise output-path overlap check: capturing per-task declared output paths goes through `task.outputs.files` / `taskProperties.outputFileProperties[].propertyFiles.files`, both of which re-resolve the underlying `FileCollection` and re-invoke user-defined `@OutputFile` / `@Nested` getters. That's observable as an extra invocation count in tests like `TaskParametersIntegrationTest."input and output properties are not evaluated too often"`. The `@Destroys` detection path stays inside `TypeMetadataStore` — pure annotation metadata, no getter invocation — so it's free of that side effect while still catching the cleanX overlap scenarios. We don't even need the destroyed paths; knowing such a task was scheduled is enough to refuse pruning it away.

### 5. Detecting request-dependent configuration logic

User code that observes anything about the current build's CLI request during configuration — the realized task graph, the list of requested tasks, or the requested task arguments — must not silently share state with a build whose CLI input was different. **The whole point** is to record one bool per index entry (`taskGraphAccessed`, named after the original use case but covering the broader concept) and exclude flagged entries from strict-superset matching at lookup. Flagged entries remain valid as exact matches.

**Flag-flipping paths:**

*Listener-registration path* (task-graph observation):
- `TaskGraphListenerRegistrationTracker` is annotated `@ListenerService` so Gradle's service registry eagerly instantiates it. It implements `BuildScopeListenerRegistrationListener` and flips `ConfigurationCacheFingerprintController.taskGraphAccessed()` on every non-`InternalListener` registration that reaches the broadcast.
- `DefaultTaskExecutionGraph.whenReady` (both `Closure` and `Action<TaskExecutionGraph>` overloads) and `addTaskExecutionGraphListener` now call `notifyListenerRegistration` — they previously didn't, which was the spec's original "v1 correctness gap." The broadcast receives the **decorated `TaskExecutionGraphListener`** instance, so `ConfigurationCacheProblemsListener.isSupportedListener` recognizes the type and doesn't flag the registration as a CC violation. The `whenReady(Closure)` overload was switched from `ClosureBackedMethodInvocationDispatch` (a `Dispatch<MethodInvocation>` that does *not* implement `TaskExecutionGraphListener`) to a lambda for the same type-check reason.
- The deprecated `beforeTask` / `afterTask` / `addTaskExecutionListener` paths already broadcast — no change there.
- The `InternalListener` marker check inside `notifyListenerRegistration` keeps Gradle's own internal listeners (including CC's) from tripping the flag.

*StartParameter-read path* (taskRequests / taskNames observation):
- `StartParameterInternal.getTaskRequests()` and `getTaskNames()` are overridden to call `InstrumentedInputs.listener().startParameterTaskRequestsObserved()` before delegating to `super`. This mirrors the override mechanism used for `getProjectProperties()`, though the downstream consumer differs: `getProjectProperties` feeds the fingerprint-input listener chain (captures values as inputs that can invalidate the entry), while `taskRequests` short-cuts directly to the superset-ineligibility flag (doesn't invalidate the entry; just excludes it from strict-superset matching).
- A new method `InstrumentedInputsListener.startParameterTaskRequestsObserved()` is added to the existing input-tracking listener interface (implementations in `NoOpInputsListener`, `PromoInputsListener`, and `InstrumentedInputAccessListener`).
- `InstrumentedInputAccessListener` is the active implementation set on `InstrumentedInputs` during the configuration phase (in `DefaultConfigurationCache.prepareForWork()`). Its implementation directly calls `ConfigurationCacheFingerprintController.taskGraphAccessed()` — short-cutting the fingerprint-input chain because this isn't a fingerprint input (the read doesn't invalidate the entry; it just excludes the entry from strict-superset matching).
- **Default is tracked.** The override fires the listener on *every* call to `getTaskRequests()` / `getTaskNames()` — there's no user-code-vs-internal-Gradle discriminator at the override level. Gradle internals avoid tripping the flag by explicitly calling `getTaskRequestsUntracked()` / `getTaskNamesUntracked()`, which delegate to `super` without firing the listener. So the distinction is "tracked vs untracked accessor," not "user code vs internal code."
- **Migrated to untracked** (in-module CC code + the always-on Gradle internal callers that read on every build): `BuiltInCommand.wasInvoked`, `RunTasksRequirements.getConfigurationCacheKeyDisplayName`, `DefaultTasksBuildTaskScheduler`, `TaskNameResolvingBuildTaskScheduler`, `TaskGraphBuildExecutionAction.renderRequestedTasks`, `ConfigurationCacheState.writeStartParameterOf`, `ConfigurationCacheStartParameter.requestedTaskNames`, `DefaultProblemsReportCreator`.
- **Known unmigrated** (out-of-module callers that don't fire on the normal CC build path, but would false-flag if exercised at config time): `RunEclipseTasksBuilder` and `RunBuildDependenciesTaskBuilder` (IDE tooling-api), `BuildProfile` (`--profile` only), `TestExecutionRequestAction` and `BuildActionSerializer` (TAPI-specific code paths), and `GradleBuild.getTasks()` (a user-task type that reads a nested-build's StartParameter — different instance, but the listener is global). These should be migrated as a follow-up; current tests pass because none of them runs on the hot CC path.

**Direct queries** — `hasTask`, `findTask`, `getAllTasks`, `getFilteredTasks`, `getDependencies`, `size`, `hasScheduledWork`, `collectScheduledWork` — are deliberately NOT instrumented. At configuration time the graph is empty, so direct queries cannot encode a request-dependent decision. Inside listener callbacks they observe a real graph, but the surrounding listener registration is already what flips the flag, so the contents of those callbacks can't drive incorrect reuse.

**Read-side.** At lookup time, `SupersetIndexLookup.selectBestMatch` excludes flagged entries from its strict-superset branch but keeps them eligible for exact matches.

## Performance

- Lookup adds one index-file read inside the existing `cache.withFileLock`. File is small (one row per stored entry per env, bounded by realistic env variety) — negligible.
- On a superset load, the BFS pruning is O(scheduled nodes); same complexity class as the existing plan finalization.
- Side-effecting-task gate at lookup is a single set-membership check per dropped identity path. Capture at store time avoids the property walker entirely — no `@OutputFile` / `@Nested` getters are invoked beyond what Gradle already does for fingerprinting (verified by `TaskParametersIntegrationTest` evaluation-count assertions).


tresat:
@bot-gradle test RfR

tresat:
@bot-gradle test RfR

tresat:
@bot-gradle test RFR

tresat:
@bot-gradle test RFR

tresat:
@bot-gradle test RFR

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/WorkGraphPruner.kt
```diff
@@ -0,0 +1,107 @@
...
+ * loader sees the pruned plan via `setScheduledWork`.
+ * <p>
+ * Names in [pruneAndRewireInPlace]'s `tasksToDrop` are compared against
+ * `task.identityPath` (canonical absolute path like `":foo:bar"`). Callers are
+ * responsible for passing canonical paths — see
+ * `DefaultConfigurationCache.commitCacheEntry` (where stored identity paths
+ * originate) and the lookup side which canonicalizes CLI input.
+ */
+internal
+object WorkGraphPruner {
```
tresat, line 39: This adds a different "drop tasks from a work graph" process to the existing process for `-x` where a lazy Spec<Task> predicate is intersected into `DefaultExecutionPlan.filter` and applied during `determineExecutionPlan()`.

The original path for `-x` can keep nodes around as "filtered no-ops" is that the plan is still mutable when it is done. The CC subset path can't do this since it operates after that bookkeeping has been completed and we already have a complete  loaded graph.  `WorkGraphPruner` is reworking an already-deserialized `ScheduledWork`.

I asked and looked for another place to store the single method here.  One option is `ScheduledWork` itself, but this seems like a worse place.  Having a new type keeps this CC-specfic work out of that general-purpose type.  It also provides a convenient place to document this functionality.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/WorkGraphPruner.kt
```diff
@@ -0,0 +1,107 @@
...
+     * Returns [initiallyScheduled] with entry nodes whose canonical identity path
+     * is in [tasksToDrop] removed, plus any nodes that are no longer reachable
+     * from the remaining entry nodes through `dependencySuccessors`. Mutates
+     * retained nodes' `dependencyPredecessors` to drop dropped-node backrefs.
+     * <p>
+     * Returns the input unchanged when:
+     *  - [tasksToDrop] is empty (exact-match reuse — no pruning needed), or
+     *  - none of the entry nodes' tasks match [tasksToDrop] (no-op for this build).
+     */
+    fun pruneAndRewireInPlace(initiallyScheduled: ScheduledWork, tasksToDrop: Set<String>): ScheduledWork {
```
tresat: This uses `tasksToDrop` rather than `tasksToKeep` because drop-lists are additive.  Keep-lists are subtractive. For this process, we start by adding the tasks missing from the CLI on this invocation (minus the cached invocation), then we add to the drop list tasks which are only present as dependencies of dropped tasks.  So the list naturally increases.

Even the initial computation is already figuring out drops; it finds stored CLI tokens NOT in the current request that identity paths to remove.  So after thinking a little more about it, I think tasksToDrop is actually the better choice here.

PR #37885, opened 2026-05-13
Configuration Cache: Execution-Time-Only Task Options (`--tests` is not a CC input)

6hundreds:
# Configuration Cache: Execution-Time-Only Task Options

The Configuration Cache (CC) now reuses a stored entry across invocations that differ only in the values of CLI options marked `@Option(executionTimeOnly = true)`. `gradle test --tests A` followed by `gradle test --tests B` previously stored two separate entries; now the second invocation (after warmup) reuses the first entry's configuration and re-applies `--tests B` at execution time.

## Behavior changes

- `gradle test --tests A` then `gradle test --tests B` (warmed) prints `Reusing configuration cache.` and runs `TestB` only.
- Two tasks in the same invocation declaring an option with the same name but different `executionTimeOnly` contracts (e.g. `:test --tests A :myCustom --tests 7`, where `:myCustom`'s `--tests` is config-time): once `tests` is in the manifest, this combination fails the build with a loud, actionable error (sample under #4 below). Users either annotate `:myCustom` or rename the option.
- A task with a malformed `@Option` / `@OptionValues` declaration no longer aborts the build at manifest-collection time — it's silently skipped (its options just don't contribute to the manifest). The validation error still fires normally when the user invokes the option on that task's CLI surface.

## Limitations

- **Cold-start two-build warmup.** First-ever invocation in a project gets no strip. Documented in test `B6`.
- **Composite builds.** Included-build options don't participate in the optimization. Documented in test `B9` and asserted in `TestLauncherCompositeBuildCrossVersionSpec`; an `@Ignore("v2 scope")` stub pins the future regression check.
- **Option-name collision across executionTime contracts.** Two task types declaring the same option name with different contracts can't both live in the same invocation under v1 — the build fails with an actionable error. Test demonstrates the failure.
- **No graceful fall-through.** Validation rejection is fatal, not recoverable.

### 1. New public annotation field

`@Option` grows a single boolean field, `@Incubating`, `@since 9.7.0`, default `false` (backwards-compatible):

```java
@Option(option = "tests", executionTimeOnly = true, description = "...")
public AbstractTestTask setTestNameIncludePatterns(List<String> patterns) { ... }
```

`true` is the property author's claim that the option's value is applied at task execution time and never read during configuration. `Test.tests` (`AbstractTestTask.setTestNameIncludePatterns`) is annotated as part of this change — the marquee use case. The value is plumbed through `OptionElement` → `OptionDescriptor` (`isExecutionTimeOnly()`, default `false` on both interfaces for source compatibility). `BooleanOptionElement.oppositeOf` propagates the flag to the synthesized `--no-foo` opposite descriptor.

### 2. Per-root-build manifest of candidate option names — a single owning service

`ExecutionTimeOnlyOptionsManifestService` (`@ServiceScope(Scope.BuildTree::class)`, registered in `BuildTreeModelControllerServices`) owns the entire feature's manifest concern plus the work-graph scans the validator needs:

- **Manifest IO** — `taskOptionNames()` (read) and `write(names)`. The manifest lives at `<cc-root>/execution-time-only-options.manifest`, a UTF-8 newline-delimited file: a `#`-prefixed header line (currently `#v1`), then sorted option names. Rewritten in full at the end of every successful configuration. The service exposes `manifestFile` for diagnostics.
- **Pure helpers in the companion**: `stripFrom` / `extractFrom` (used by `ConfigurationCacheKey` and `ConfigurationCacheState` to partition CLI args around the candidate names) and `findExecutionTimeOnlyContributor` (populates `ExecutionTimeOnlyOptionsValidationException`'s contributor field on validation failure).

The previous `ExecutionTimeOnlyOptions.NAMES = setOf("tests")` constant is **deleted** — the annotation is the sole source of truth.

**Write path.** `DefaultConfigurationCache.writeExecutionTimeOnlyOptionsManifest` runs at the end of every successful Store / Update cacheAction, walks `rootBuild.mutableModel.taskGraph.allTasks`, and uses `ExecutionTimeOnlyOptionsCollector.collect(tasks, optionReader)` to compute the name set. A task whose annotation surface throws `OptionValidationException` is silently skipped rather than aborting the build.

**Write robustness.** Atomic via `Files.createTempFile` (uniquely named — concurrent-daemon safe) plus `Files.move(..., ATOMIC_MOVE)` with `AtomicMoveNotSupportedException` fallback to `REPLACE_EXISTING` (cross-device / NFS / certain Windows configurations). Try/finally cleans up the temp file on writer failure. If the write itself fails (`IOException`) after `saveWorkGraph` has already persisted the cache entry, the orchestrator calls `problems.onStoreSerializationError()` to discard the just-stored entry — preventing the stale-manifest / fresh-entry inconsistency.

**Read path.** `ConfigurationCacheKey.appendRequestedTasks` calls `manifestService.taskOptionNames()` and feeds the candidates into `ExecutionTimeOnlyOptionsManifestService.stripFrom(requestedTaskNames, candidates)` before hashing. The key depends on the manifest service — not the storage layer — preserving the identity-providing class's value-only input contract.

**Read robustness.** Only `IOException` is caught (narrowed from `Exception`) and logged at WARN; RuntimeExceptions propagate. A missing file, empty file, missing header, or unrecognized header (e.g. a future `#v2`) all return `emptySet()` — disabling key-stripping is always safer than producing wrong build results.

### 3. Load-time validation

After deserialization, `ConfigurationCacheState.validateAndApplyExecutionTimeOnlyOptions` walks each `LocalTaskNode` in the loaded work graph. For every option whose name appears in the manifest, the task's `OptionDescriptor.isExecutionTimeOnly()` must be `true`; otherwise the validation path fires.

**Deferral pattern.** Throwing inside the codec gets the exception captured as a "problem" and surfaced at finishBuildTree, bypassing the orchestrator's catch. Instead, the failure is stored on the codec's state (`pendingValidationFailure`) and re-thrown after the codec returns, both at `ConfigurationCacheState.readRootBuildState` and in `DefaultConfigurationCacheIO.readRootBuildStateFrom` (defense in depth).

**Self-contained exception.** `ExecutionTimeOnlyOptionsValidationException` extends `GradleException`, implements `ResolutionProvider`, is `@Contextual`. Fields: `executionTimeOnlyTaskPath` (contributor), `configTimeTaskPath` (violator), `optionName`. The orchestrator no longer wraps it — it propagates directly through `DefaultConfigurationCache` and Gradle's renderer surfaces the resolutions automatically.

**Loud-failure response, not graceful fall-through.** Sample:

> Configuration cache entry cannot be reused: task ':myTestLike' declares option '--tests' as execution-time-only, but another task ':myCustom' uses an option with the same name that is allowed to be read during Configuration time.  These tasks cannot be invoked together.
>
> Possible solutions:
>  1. Annotate '--tests' on ':myCustom' with @Option(executionTimeOnly = true) if its value is in fact applied at execution time and not consulted during configuration.
>  2. Use a different option name on one of the two tasks to avoid the collision.
>  3. Invoke the two tasks in separate builds.

When the contributor can't be identified (stale manifest), the message + resolutions degrade to describe the violator alone.

### 4. Composite builds are v1-out-of-scope

The manifest is scoped to the root build's task graph only. Included-build invocations through the root (e.g. `:app:test --tests X` where `app` is an included build) do not get the strip optimization — the included-build's task options aren't collected into the root's manifest. The `LocalTaskNode` narrowing on the validation path already prevents cross-build dereferences. `TestLauncherCompositeBuildCrossVersionSpec` asserts the limitation directly: `BUILD_LAUNCHER` + composite + `--tests` + CC always misses on value change, regardless of Gradle version. An `@Ignore("v2 scope")` stub pins the expected v2 behavior so the regression check is visible when the work resumes.

## Testing

- **Unit** (40 tests across 4 classes) — `ExecutionTimeOnlyOptionsTest` (strip/extract edges), `ExecutionTimeOnlyOptionsManifestServiceTest` (IO + corruption + header forward-compat), `ExecutionTimeOnlyOptionsValidationExceptionTest` (both message + resolutions branches), `ConfigurationCacheKeyTest` (manifest-driven key via mock).
- **Integration** — `ExecutionTimeOnlyOptionsIntegrationTest` (10 B-scenarios: cold-start, value-change reuse, manifest invalidation, deletion-recovery, mixed-contract loud failure, write-failure → entry discard, stale-manifest fallback). `TestLauncherCompositeBuildCrossVersionSpec` asserts composite v1 limitation + pins v2 regression.

## Performance

- Manifest read: single small file (one option name per line; today contains `"tests"` only) inside the existing CC key-calculation path. Negligible.
- Manifest write: atomic rename at end of successful configuration. Adds one task-graph walk + one option-reader invocation per task — same machinery `TaskOptionsGenerator.generate` already performs lazily during CLI option parsing.
- Validation: O(scheduled `LocalTaskNode` count × options per task), bounded by what `TaskOptionsGenerator.generate` already does for option discovery. Runs only on CC load, not on store. The contributor lookup (only on validation failure) adds one extra graph scan.
- Cold-start cost: one extra cold store before the manifest takes effect. Steady state reuses CC normally.


tresat:
@bot-gradle test RFR

tresat:
@bot-gradle test RFR

PR #37890, opened 2026-05-14
Reduce memory impact of GraphStructure

jvandort:
Ensure we hold a soft reference to ResolvedGraphResult, which ensures the wrapper over GraphStructure is deallocated when other users no longer reference it. This ensures consistent reference identities for ResolutionResult types as long as some code references a live instance of it, while also allowing the wrapper type to be deallocated if no longer referenced.

Also migrate some internal consumers of the dependency graph to use GraphStructure instead of the public wrapper, avoiding allocations of an entire public view of the graph when not necessary. This should particularly improve Android builds, as they both use consistent resolution and artifact transform dependencies.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test apt

big-guy (approved):

PR #37893, opened 2026-05-14
Consolidate queue processing of BuildOperationQueue

octylFractal:
This is a follow-up to #37845 that removes the exponential backoff and tryRunAsWorkerThread in favor of a rewrite that stops allocating a set of worker runnables per queue. This allows us to significantly simplify the implementation because the runnables do not need to be stopped at the end of the queue processing, and we do not need to care as much about preserving them or not.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


octylFractal:
@bot-gradle test PT

octylFractal:
@bot-gradle test PT

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/WorkerLeaseQueueProcessor.java
```diff
@@ -0,0 +1,240 @@
...
+
+    @Override
+    public void notifyBlockingWorkStarting() {
+        workerCounter.notifyBlockingWorkStarting();
+        trySpawnWorker();
+    }
+
+    @Override
+    public void notifyBlockingWorkFinished() {
+        workerCounter.notifyBlockingWorkFinished();
```
ghale, line 171: 💭 Should this also call `coordinationService.notifyStateChange()`?  If another worker is stuck in `tryLockWhile()`, it won't get notified right away, will it?  I don't see this as being a problem, per se, (i.e. _something_ will eventually wake that worker up) but it seems like a correctness thing.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/operations/BuildOperationQueue.java
```diff
@@ -30,20 +30,41 @@ public interface BuildOperationQueue<T extends BuildOperation> {
...
      * @param operation operation to execute
+     * @see #addUnconstrained for IO intensive operations
      */
     void add(T operation);
 
+    /**
+     * Adds an operation to be executed without holding a worker lease, potentially executing it instantly.
+     * <p>
+     * Execution is not constrained by the configured maximum number of workers, allowing as many threads as
+     * required up to a separate, higher limit. Intended for IO intensive operations. Such operations must not
```
ghale: 💅 We might clarify what the "higher limit" is or where to find it (i.e. a link to `WorkerLimits`).

ghale (approved):
This looks good to me.  I added two minor comments, but these should not be considered blocking.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/operations/BuildOperationQueue.java
```diff
@@ -30,20 +30,43 @@ public interface BuildOperationQueue<T extends BuildOperation> {
...
+     * {@linkplain org.gradle.internal.work.WorkerLeaseRegistry worker lease} before proceeding.
+     * Intended for CPU intensive operations.
      *
      * @param operation operation to execute
+     * @see #addUnconstrained for IO intensive operations
      */
     void add(T operation);
 
+    /**
+     * Adds an operation to be executed without holding a worker lease, potentially executing it instantly.
```
jvandort: `potentially executing it instantly` makes me think we execute it on the calling thread.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/WorkerLoop.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.gradle.internal.work;
+
+/**
+ * A loop that owns a worker lease until the given stopping condition.
+ */
+public interface WorkerLoop {
+    /**
+     * {@return {@code true} if the loop should continue, or {@code false} if the loop should stop}
+     * The worker lease will be dropped if it was acquired.
```
jvandort: I think the wording is confusing here. 

`The worker lease will be dropped if it was acquired.`

Who is responsible for dropping the worker lease? 

From what I can tell, this method will not drop the worker lease. On `false` returned, this method only drops the worker "slot". 

It seems the responsibility of dropping the lease lies with the _caller_ of this method. The `will` in the message does not seem to imply that

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/WorkerCounter.java
```diff
@@ -0,0 +1,135 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.work;
+
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
+
+final class WorkerCounter {
```
jvandort, line 31: I think this class could use some documentation specifically describing how it is different than a worker lease and what a "slot" actually models

subprojects/core/src/main/java/org/gradle/internal/operations/DefaultBuildOperationQueue.java
```diff
@@ -16,408 +16,258 @@
...
+        }
+
+        QueueState finishOperation() {
+            // cancel() zeroes pendingOperations; skip the decrement in that case.
+            int newPending = status == QueueStatus.CANCELED ? pendingOperations : pendingOperations - 1;
+            return new QueueState(status, newPending, inFlightOperations - 1);
+        }
+
+        QueueState cancelQueue() {
+            return new QueueState(QueueStatus.CANCELED, 0, inFlightOperations);
```
jvandort: I think we might be able to clean up some of the conditional `state` logic here by setting pending equal to inFlight here. Or alternatively, defining pending as the number of work items _not_ in flight, in which case 0 would still apply here but finishOperation would never touch pending and startOperation would need to decrement pending.

subprojects/core/src/main/java/org/gradle/internal/operations/DefaultBuildOperationQueue.java
```diff
@@ -16,408 +16,258 @@
...
+    private final CountDownLatch allOperationsComplete = new CountDownLatch(1);
+    private final AtomicReference<QueueState> state = new AtomicReference<>(new QueueState(QueueStatus.WORKING, 0, 0));
+    private final List<Throwable> failures = new CopyOnWriteArrayList<>();
 
     DefaultBuildOperationQueue(
         boolean allowAccessToProjectState,
         WorkerLeaseService workerLeases,
-        BuildOperationExecutionContext context,
+        SubmissionQueue constrainedQueue,
+        SubmissionQueue unconstrainedQueue,
```
jvandort: I think it might be more clear here if we were "real" about the fact that the unconstraint submission queue is really just an executor, and passing in the raw `Executor` here.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/DefaultWorkerLeaseService.java
```diff
@@ -138,27 +138,37 @@ public void runAsWorkerThread(Runnable action) {
...
+        } else {
+            withLocks(Collections.singletonList(workerLeaseLockRegistry.newUnmanagedLease()), action);
         }
-        return Optional.of(runAndReleaseLocks(Collections.singletonList(lease), action));
     }
 
     @Override
-    public void runAsUnmanagedWorkerThread(Runnable action) {
+    public void runWorkerLoop(WorkerLoop workerLoop) {
         Collection<? extends ResourceLock> locks = workerLeaseLockRegistry.getResourceLocksByCurrentThread();
```
jvandort: Based on the single caller of this method, I don't think we ever get into the scenario where we start a worker loop in a thread that already has a worker lease. 

I think it might be more clear if we removed this condition, then instead exposed a more generalized `tryRunAsWorkerThreadUntil(Runnable, BooleanSupplier)`, which would let us pull this loop into the build operation queue processor and get rid of the `WorkerLoop` interface

jvandort (approved):

PR #37897, opened 2026-05-15
Update Documentation: improve documentation of BuildServiceParameters with CC

### LINKED ISSUE #17141: Improve documentation of BuildServiceParameters
### Expected Behavior

Decorated class for build service parameters should generate regardless of the name (in this case getter prefix) of a build service's parameter methods.

I haven't seen any documentation regarding the `get` prefix being required on the documentation page for build services or custom gradle types. 

### Current Behavior

If a method to get a property is declared without a `get` prefix, plugin fails to run because of failing to generate decorated class.

### Context

Was working on an unreleased gradle plugin for an open source project using build services. The organization this is destined for is erring away from redundant `get` prefixes. 

### Steps to Reproduce 

Minimal test case - with build-logic included build as to replicate what I have done in my own code:
https://github.com/i509VCB/gradle-bug-test

1. Create a new build service.
2. Define interface for build service parameters
3. Add method with a non-getter name e.x. `something` with the return type `Property<String>` to parameters interface.
4. Register the build service
5. Try to build the project

### Your Environment

Build scan URL:
https://scans.gradle.com/s/jyb7abwnynlo4

### LINKED ISSUE #17813: More docs about configuration caching, esp. with build services

lkasso:
### Details

### Context
- Fix https://github.com/gradle/gradle/issues/17141
- Fix https://github.com/gradle/gradle/issues/17813

### Documentation Checklist


platforms/documentation/docs/src/docs/userguide/reference/task-development/build_services.adoc
```diff
@@ -195,3 +198,10 @@ include::{snippetsPath}/reference/plugin-development/buildServiceListener/groovy
 <1> Use <<service_injection.adoc#service_injection,service injection>> to obtain an instance of the `BuildEventsListenerRegistry`.
 <2> Register the build service as usual.
 <3> Use the service `Provider` to subscribe to the build service to build events.
+
+[[sec:configuration_cache_interaction]]
+== Build Services and the Configuration Cache
+
+When the <<configuration_cache.adoc#config_cache, Configuration Cache>> is enabled, build service parameters are serialized into the cache entry, and the service instance is recreated on every build.
+Several caveats apply.
+See <<configuration_cache_requirements.adoc#config_cache:requirements:build_services, Using Build Services>> in the Configuration Cache requirements for the full treatment.
```
mlopatkin, line 207: 🤔 Maybe we should reverse, and have the full contract here and only link it from the CC page? My rationale is that CC page is very long already, so its better to keep details out and closer to the "thing".

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -349,6 +349,30 @@ Recommended Alternatives:
...
+
+*The service instance is recreated on every build.*
+Only the parameters are cached.
+On a cache hit, Gradle reads the recorded parameters from the entry, instantiates the build service afresh, and injects them.
+Any work done in the service's constructor or first method invocation runs again on every build, regardless of whether the configuration phase ran.
+
+*A `BuildServiceProvider` cannot drive a `ValueSource` at configuration time.*
+Passing a provider of a `BuildService` (or anything derived from it via `map` or `flatMap`) as a `ValueSourceParameters` value that is read at configuration time is not supported.
+The supported alternatives are `@ServiceReference` for direct injection and storing the build service reference in an `@Internal`-annotated property on a task.
+See <<configuration_cache_status.adoc#config_cache:not_yet_implemented:build_services_in_fingerprint, the related not-yet-implemented entry>> for the current state.
```
mlopatkin: :x: This doesn't really look like an alternative. Presumably, the user needs the value at configuration time. A proper workaround would be to replace service-derived provider with opaque computation:
```kotlin
val serviceProvider = ...

val valueSource = providers.of(...) {
    parameters {
        // Doesn't work
        value = serviceProvider.map { it.getValueFromService() }
        // Workaround, preserves laziness
        value = provider { serviceProvider.get().getValueFromService() }
    }
}

println(valueSource.get())
```

The obvious downside is that the whole computation of `valueFromService` contributes configuration inputs rather than being lazily computed when checking the fingerprint and invalidating the cache only when ValueSource's result changes.

It also doesn't work when the computed value is transient and changes between builds (e.g. a publishing URL in the original issue)

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -349,6 +349,30 @@ Recommended Alternatives:
...
+== Using Build Services
+
+<<build_services#build_services, Build Services>> are the Gradle-supported way to share state across tasks, replace `BuildListener`-style hooks, and observe task execution events.
+They are also the most common feature-level integration point with the Configuration Cache, and a few caveats apply.
+
+*Parameter values are part of the cache entry.*
+The parameters interface of a build service is a custom Gradle type, and its property values are serialized into the Configuration Cache entry alongside the rest of the build's configured state.
+Parameter types must be Configuration Cache-compatible: standard managed property types (`Property<T>`, `RegularFileProperty`, `ListProperty<T>`, and so on) are safe; raw `Configuration` references, live JVM state, and `Project` references are not.
+
+*The service instance is recreated on every build.*
```
mlopatkin: 🤔This is a bit oversimplifying, IMO

I think the point is that with CC, the lifecycle of a service differ.
I'd say something along the lines of:
1. With CC, all build services that are created at configuration time (by calling `get()` on provider) are destroyed at the end of the configuration phase. The only exception are services registered as OperationCompletionListeners and created at configuration time. The same instances of those are used during the following execution phase.
2. At execution time, services are created upon first use; this includes recreating those that were destroyed at step (1).
3. On cache hit, when there is no prior configuration time, all services are created upon first use - from scratch. Registered OperationCompletionListeners are created upon first event if nothing triggers it earlier.

A consequence - a same build service may be created and destroyed twice in the same build.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -349,6 +349,30 @@ Recommended Alternatives:
...
 
+[[config_cache:requirements:build_services]]
+== Using Build Services
+
+<<build_services#build_services, Build Services>> are the Gradle-supported way to share state across tasks, replace `BuildListener`-style hooks, and observe task execution events.
+They are also the most common feature-level integration point with the Configuration Cache, and a few caveats apply.
+
+*Parameter values are part of the cache entry.*
+The parameters interface of a build service is a custom Gradle type, and its property values are serialized into the Configuration Cache entry alongside the rest of the build's configured state.
+Parameter types must be Configuration Cache-compatible: standard managed property types (`Property<T>`, `RegularFileProperty`, `ListProperty<T>`, and so on) are safe; raw `Configuration` references, live JVM state, and `Project` references are not.
```
mlopatkin: :x: `are safe` - it is important to note that they're safe as long as their value type is CC-compatible. `Property<Thread>` wouldn't work.

It might be worth noting that all usual CC serialization considerations apply, though we need to document them somewhere first 😅

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -349,6 +349,30 @@ Recommended Alternatives:
...
+
+*A `BuildServiceProvider` cannot drive a `ValueSource` at configuration time.*
+Passing a provider of a `BuildService` (or anything derived from it via `map` or `flatMap`) as a `ValueSourceParameters` value that is read at configuration time is not supported.
+The supported alternatives are `@ServiceReference` for direct injection and storing the build service reference in an `@Internal`-annotated property on a task.
+See <<configuration_cache_status.adoc#config_cache:not_yet_implemented:build_services_in_fingerprint, the related not-yet-implemented entry>> for the current state.
+
+*Register as `OperationCompletionListener` to observe task execution.*
+A build service registered via link:{javadocPath}/org/gradle/build/event/BuildEventsListenerRegistry.html#onTaskCompletion(org.gradle.api.provider.Provider)[`BuildEventsListenerRegistry.onTaskCompletion`] is the supported way to survive the configuration-to-execution boundary correctly under the Configuration Cache.
+Other ways of holding a build service reference may not survive a cache hit cleanly.
+
```
mlopatkin: 🤔This is a bit hand-wavy, we should be more concrete, as the contract is actually clear:
- If the service is a registered `OCL` and was created at configuration time - this instance will be reused during the following execution phase _of the same build_. Further builds that hit CC and skip configuration phase will rebuild the service instance from scratch when needed.
- All other active services are destroyed (i.e. their `close()` method is called and Gradle releases all references to them) at the end of the configuration phase. When the same service is requested at execution phase, it is created from scratch.

While there can be forbidden spells to stash the original instance for longer, this isn't something users should be doing or we - hinting at.

:question: Actually, after reading more into it, I don't understand what "survive a cache hit" means at all.

mlopatkin (changes requested):
The points you're highlighting are valid and important, but I think we need to work more on the message. I'll be happy to chat synchronously if you need more info.

platforms/documentation/docs/src/docs/userguide/reference/task-development/build_services.adoc
lkasso: Good suggestion, but I'd like to keep the full treatment in configuration_cache_requirements.adoc.

My reasoning: the CC requirements page is meant to be the single authoritative reference for all CC behavior. Someone auditing a build for CC compatibility, or someone training an AI on CC docs, should be able to point at the CC pages and get everything. 
I've kept the short cross-reference in build_services.adoc so readers coming from that direction can find it too. 

I would make this change only in Gradle 10. WDYT?

mlopatkin: >I would make this change only in Gradle 10. WDYT?

I'm fine either way as long as there are cross-links.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -357,21 +357,57 @@ They are also the most common feature-level integration point with the Configura
...
+Registered `OperationCompletionListener` services are created upon the first event if nothing triggers them earlier.
 
 *A `BuildServiceProvider` cannot drive a `ValueSource` at configuration time.*
 Passing a provider of a `BuildService` (or anything derived from it via `map` or `flatMap`) as a `ValueSourceParameters` value that is read at configuration time is not supported.
-The supported alternatives are `@ServiceReference` for direct injection and storing the build service reference in an `@Internal`-annotated property on a task.
+
+If you do not need the service-derived value at configuration time, use `@ServiceReference` for direct injection or store the build service reference in an `@Internal`-annotated property on a task instead.
+
+If you need the value at configuration time, you can wrap the computation in a `ValueSource` and use an opaque provider to preserve laziness:
+
```
mlopatkin: :x: This is a bit off. We actually started with a `ValueSource` that consumes a service, and now we're talking about wrapping it in another `ValueSource`, which isn't what we want the users to do. Maybe we can split the example into two, an initial bad one, and the fixed one instead. Something  like:

Sometimes you may want to use a BuildService to derive a parameter for a ValueSource, like that:
```kotlin
val serviceProvider = gradle.sharedServices.registerIfAbsent("myService", MyService::class) { /* ... */ }

val valueSource = providers.of(MyValueSource::class) {
    parameters {
        // Doesn't work when valueSource.get() is called at configuration time
        value = serviceProvider.map { it.getValueFromService() }
   }
}
```

This is only supported by the Configuration Cache if the resulting ValueSource isn't used at configuration time. If you really need that, you have to compute the value at configuration time, wrapping the computation in `provider {}` block to preserve laziness:
```kotlin
val valueSource = providers.of(MyValueSource::class) {
    parameters {
        // Workaround
        value = provider { serviceProvider.get().getValueFromService() }
   }
}
```

mlopatkin (changes requested):
Just one correction.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -379,23 +379,32 @@ Passing a provider of a `BuildService` (or anything derived from it via `map` or
...
-        // value = serviceProvider.map { it.getValueFromService() }
+        // Doesn't work when valueSource.get() is called at configuration time
+        value = serviceProvider.map { it.getValueFromService() }
+    }
+}
+----
 
-        // Workaround — preserves laziness:
+This is only supported by the Configuration Cache if the resulting `ValueSource` isn't used at configuration time.
+If you need the value at configuration time, you have to compute the value eagerly, wrapping the computation in a `provider {}` block to preserve laziness:
```
mlopatkin: I just can't stop :)

:x: Having "eagerly" and "preserve laziness" in one sentence feels confusing. And actually, that doesn't explain the difference correctly - my bad. What actually happens is that broken sample caches the computation (which is what CC cannot do) and the workaround caches the computed value - in the fingerprint data, to recompute the value source later.

So it may go like

>If you need the `ValueSource`'s value at configuration time, you have to tell the Configuration Cache to use the computed input value for fingerprinting by wrapping the eager computation in `provider {}` block. 

but I'm not totally happy with that phrasing either.

💅 Maybe we can also replace `value =` in `parameters` block with `input = ` to avoid confusion between ValueSource's value and its input value.

mlopatkin (changes requested):
We're almost there :)

mlopatkin (approved):

Issue #11177, opened 2019-10-30
Deprecate using Ant's default excludes


wolfs:
After https://github.com/gradle/gradle/issues/11176 is done, we should start issuing deprecation warnings when
- default excludes are not specified explicitly
- default excludes are actually excluded

This way we can make the default to have no default excludes in 7.0, requiring users to specify them when needed.

---
cc: @gradle/build-cache

Issue #11176, opened 2019-10-30
Allow configuring default excludes in a first class manner


wolfs:
Currently, Gradle by default uses Ant's [default excludes](https://docs.gradle.org/current/userguide/working_with_files.html) when reading directories. That is unexpected for many people and comes for "older times" when version control files (`.svn` or `CVS`) were part of of every directory in a hierarchy.

Since default excludes need to be filtered for every file system walk, they incur quite some performance overhead as well.

In order to change the default excludes, some "magic" incantations are necessary, and those need to happen early enough in the build so the file system snapshotter and the directory walker pick it up. So something like this needs to be done in `settings.gradle`:

```
org.apache.tools.ant.DirectoryScanner.removeDefaultExclude("**/CVS")
org.apache.tools.ant.DirectoryScanner.removeDefaultExclude("**/CVS/**")
```

As a first step, it would be great to configure the default excludes in e.g. `settings.gradle` or the project explicitly, instead of using the common list from Ant. That would make it easy to use no default excludes at all, or at least only use the ones required for the current project.

See also https://github.com/gradle/gradle/issues/2986.

---
cc: @gradle/build-cache

wolfs:
See #11177 and #11178 for follow ups.

sis-yoshiday:
@wolfs Hi, I tried `6.1-20191029232520+0000` and #11017 seems to be fixed, Thanks!

Diff in my example project is below.

https://github.com/sis-yoshiday/gradle-11017/commit/061fdc32d6861f0beacfb39510d7ccfb883eade0

PR #37901, opened 2026-05-15
Remove deprecated DirectoryScanner mutation path (Gradle 10)

reinsch82:
## Summary

Follow-up to [#37882](https://github.com/gradle/gradle/pull/37882) (deprecation in 9.6.0). This PR is the **Gradle 10 removal** — the deprecation message in 9.6.0 already promises `willBeRemovedInGradle10()`, and this delivers on that.

**Targets the deprecation PR's branch** so the diff shows only the 2 unique removal commits. Will need to be rebased onto `master` once #37882 merges.

> ⚠️ **Do not merge until #37882 has shipped to users in at least one Gradle release.** Users need a chance to see the deprecation warning before the code path stops working.

## What this PR does

| Commit | What |
|---|---|
| `2a5e2294de7` | Drop `DirectoryScanner` read from `PatternSpecFactory`. Listener-only now. Removes a static-state read from a hot path (every `FileCollection` exclude resolution). |
| `1cf78fc8440` | Remove `DirectoryScanner` mutation deprecation + mirror from `DefaultFileSystemDefaultExcludesProvider`. Build-script calls to `DirectoryScanner.addDefaultExclude(...)` continue to compile (class is still on the classpath via the public `AntBuilder` API) but have no effect on Gradle's resolved excludes. Updates 5 test files accordingly. |

The mid-build mutation safety check (`InvalidUserCodeException("Cannot change default excludes during the build...")`) is also dropped along with the `PatternSpecFactory` read. With Gradle no longer consulting `DirectoryScanner` static state, mid-build mutations are silent no-ops anyway. Mid-build mutations via the new `Settings.fileSystemDefaultExcludes` `SetProperty` are not expected because the property is finalized after settings evaluation.

## What this PR does *not* do

- **Drop `libs.ant` from `:core`.** Ant remains on the runtime classpath for unrelated reasons: the public `AntBuilder` API, `Copy.filter(...)` Ant `FilterReader`s, and the `GradleApiSpecProvider` allowlist that exposes `org.apache.tools.ant.*` to user scripts. Removing those is part of the broader Ant-removal arc (Sterling's `org.gradle.ant-support` plugin design).
- **Update the Gradle 10 upgrade guide.** That document doesn't exist yet; when it's created an entry should be added pointing at the v9 deprecation section.

## Test plan

- [x] `:core-api:compileJava`, `:core:compileJava` clean
- [x] `:core:test --tests "*GradleDefaultExcludesTest*"`
- [x] `:snapshots:forkingIntegTest --tests "*DefaultExcludesIntegrationTest*"` (rewritten — old deprecation tests gone, new "is silently ignored" coverage added)
- [x] `:configuration-cache:embeddedIntegTest --tests "*ConfigurationCacheFileSystemDefaultExcludesTest*"` (fixture renamed; build-script-mutation case kept as negative test)
- [x] `sanityCheck` (binary compatibility ok; no `accepted-public-api-changes.json` entry needed)
- [ ] Full CI

Refs: gradle/gradle#11176, gradle/gradle#11177

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheFileSystemDefaultExcludesTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheFileSystemDefaultExcludesTest.groovy
@@ -69,7 +69,7 @@ class ConfigurationCacheFileSystemDefaultExcludesTest extends AbstractConfigurat
         def excludedByBuildScriptFileCopy = file("build/output/${excludedByBuildScriptFileName}")
         excludedByBuildScriptFile.text = "input"
 
-        getBuildFile(dsl) << DefaultExcludesFixture.DefaultExcludesLocation.addDefaultExclude(excludedByBuildScriptFileName)
+        getBuildFile(dsl) << DefaultExcludesFixture.DefaultExcludesLocation.addDefaultExcludeForBuildScript(excludedByBuildScriptFileName)
 
         when:
         configurationCacheRun spec.copyTask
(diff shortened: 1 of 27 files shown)


reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

PR #37904, opened 2026-05-17
fix(signing): Improve error message when signatory is not configured (#19417)

### LINKED ISSUE #19417: Improve message for Cannot perform signing task ':...' because it has no configured signatory
### Expected Behavior

The message should be understandable by those who attempt the build.

For instance: "Can't create cryptographic PGP signatures for my-app.zip because .... Please follow ...docs.gradle.org... to find how to configure the signing keys"

### Current Behavior

```
Cannot perform signing task ':signArchives' because it has no configured signatory
```

In practice, the ones who launch `./gradlew build` have absolutely no idea what `signatory` is, they have no idea why the signatory must be configured and so on.

### Context

"no configured signatory" is quite popular on Google.

algojogacor:
## Summary

Fixes #19417

The error message "Cannot perform signing task ':X' because it has no configured signatory" is unhelpful for users who encounter it — most have no idea what a "signatory" is or how to configure one.

## Root Cause

`Sign.java` line 231 throws `InvalidUserDataException` with a terse message that gives no guidance on resolution. Users searching for "no configured signatory" (which is common on Google) find the error but no actionable fix.

## Fix

Extended the error message in `Sign.java` to:
1. Explain that a signatory (PGP or in-memory) must be configured in the `signing {}` block
2. Include a link to the Gradle signing plugin documentation

The test in `NoSigningCredentialsIntegrationSpec.groovy` uses `failureHasCause` which performs substring matching — the existing assertion remains compatible with the new message.

## Changes

- `platforms/software/signing/src/main/java/org/gradle/plugins/signing/Sign.java`: improved error message (`+3 -1` lines)

## Testing

- **Existing test**: `NoSigningCredentialsIntegrationSpec` — `failureHasCause` substring match still passes with new message ✅
- **No signatory configured**: error now includes link to docs ✅
- **Signatory configured**: no change in behavior (condition is behind `if (getSignatory() == null)`) ✅

### DIFF
--- a/platforms/software/signing/src/main/java/org/gradle/plugins/signing/Sign.java
+++ b/platforms/software/signing/src/main/java/org/gradle/plugins/signing/Sign.java
@@ -228,7 +228,9 @@ public abstract class Sign extends DefaultTask implements SignatureSpec {
     @TaskAction
     public void generate() {
         if (getSignatory() == null) {
-            throw new InvalidUserDataException("Cannot perform signing task \'" + getPath() + "\' because it has no configured signatory");
+            throw new InvalidUserDataException("Cannot perform signing task '" + getPath() + "' because it has no configured signatory."
+                + " Please configure a signatory (e.g., PGP or in-memory) in the signing {} block."
+                + " See https://docs.gradle.org/current/userguide/signing_plugin.html for more information.");
         }
 
         for (Signature.Generator signature : sanitizedGenerators().values()) {


jbartok:

The [DCO](https://probot.github.io/apps/dco/) check is failing, please look into that as we can't accept the PR otherwise.

***

This has to be done via the [Problems API](https://docs.gradle.org/current/userguide/reporting_problems.html).

Issue #23817, opened 2023-02-07
Support agent-based instrumentation with TestKit `withDebug(true)`


mlopatkin:
The agent-based instrumentation needs a specially crafted JVM process, because the agent is installed with the command-line argument. This works well with daemon processes when Gradle controls creation of the process, but doesn't work with TestKit tests using `withDebug(true)` (or the internal embedded TAPI mode), where the process is already created.

At the time of writing, there seems to be two options:
1. Use the runtime attach API to add the agent to the process after JVM is started
2. Tweak the `java-gradle-plugin` plugin to configure tests in such a way the agent switch is added to the test process launch command.

eskatos:
None of the options are good.

Option 1 - Use the runtime attach API to add the agent to the process after JVM is started
* Can't unload an agent, doesn't work with `withGradleVersion()` unless we force `forkEvery = 1`
* Dynamic agent attaching is deprecated and will be removed from the JVM
   (see JEP 451 "Prepare to Disallow the Dynamic Loading of Agents")

Option 2 - Tweak the `java-gradle-plugin` plugin to add the agent to the test runner VM
* Pins the agent to the outer wrapper Gradle version. Not viable wrt. `withGradleVersion()` either

Option 3 - Classloader-only instrumentation (no agent at all)
* A Jacoco agent will see the wrong bytecode and will misattribute coverage

My take so far is that it is inherently ~not possible to make TestKit embedded mode work fine with the instrumentation agent.

Moving forward, the only option I see is to not fix the embedded mode but find a way to replace it with something that work and is correct.

The usability of debugging non-embeddable TestKit is the important aspect. This would bring always correct debugging. That would be an enhancement over the embedded mode which is in fact not correct. Embedded is a different runtime that doesn't match the actual Gradle runtime.

Speed is the other factor. Today's embedded mode provides some speed with the tradeoff of incorrectness. I'd rather we focus on correctness first, and from there we make things faster.

mlopatkin:
Given that the agent API surface is intentionally small, we might be able to achieve some sort of backward/forward compatibility when using Option 2. Of course, it is the lesser evil, but still evil - it will limit our options to evolve the agent protocol to some extent and require special handling for version mismatches.

We'll be imposing two new requirements:
- Agent must support all Gradle versions its TestKit version supports.
- Users may have to update TestKit to test on recent Gradle versions properly.

eskatos:
[...]

PR #37909, opened 2026-05-18
Allow third-party Java agents on the Gradle daemon

### LINKED ISSUE #25929: False positive configuration cache violation running TestKit in IntelliJ
### Expected Behavior

It should be possible to run TestKit tests in IntelliJ without Gradle complaining about configuration cache issues.

### Current Behavior

TestKit tests fail with this error:

```
- Gradle runtime: support for using a Java agent with TestKit builds is not yet implemented with the configuration cache.
```

This is due to fuzzy-checking in org.gradle.internal.agents.AgentUtils#isThirdPartyJavaAgentSwitch combined with IntelliJ placing an agent on the class path not for instrumentation purposes but to handle command line argument length, IDE integration.

This is the output from `ManagementFactory.getRuntimeMXBean().inputArguments`, which is the input into AgentUtils.

```
[-Didea.test.cyclic.buffer.size=1048576, -javaagent:/Users/chrislee/Library/Application Support/JetBrains/Toolbox/apps/IDEA-U/ch-0/232.8660.185/IntelliJ IDEA 2022.3 EAP.app/Contents/lib/idea_rt.jar=53708:/Users/chrislee/Library/Application Support/JetBrains/Toolbox/apps/IDEA-U/ch-0/232.8660.185/IntelliJ IDEA 2022.3 EAP.app/Contents/bin, -Dfile.encoding=UTF-8]
```

It shows the `idea_rt.jar` being installed as a Java agent.

### Context (optional)

Unable to run TestKit tests in IntelliJ with configuration cache enabled.  Disabling configuration cache is not an option, as it is now a stable feature and desirable to test plugins against.

### Steps to Reproduce

Run a TestKit unit test in IntelliJ.

### Gradle version

8.2.1

### Build scan URL (optional)

_No response_

MacOS 13.x
IntelliJ 2023.2

### LINKED ISSUE #25979: Support using third-party Java agents for Gradle daemon process with configuration cache enabled
### Expected Behavior

A user can apply java agent to the Gradle daemon process, and build with configuration cache enabled. 

* Attached agent is reported as a configuration cache problem when using TestKit to run the build (see also gradle/gradle#25929), even if the agent does not register instrumentation.
* Legacy instrumentation mode exposes instrumented bytecode to third-party agents (which is problematic for e.g. Jacoco's online instrumentation).

### Context

TestKit is the primary concern, as plugin authors want to test their plugins with the configuration cache. 

FWIW, Jacoco is not fully supported with TestKit yet (see #1465, #16603).

eskatos:
The configuration cache today refuses any build whose JVM has a non-Gradle `-javaagent:` attached, breaking IntelliJ-launched builds (`idea_rt.jar`) and TestKit + Jacoco coverage runs. Vintage builds quietly drop buildscript-classpath instrumentation in the same scenario.

* Fixes https://github.com/gradle/gradle/issues/25929
* Fixes https://github.com/gradle/gradle/issues/25979

With Gradle's own instrumentation agent active (the daemon default), `--configuration-cache` no longer fails on a third-party agent, Jacoco coverage of plugin-under-test code works under CC + TestKit, and Vintage keeps bytecode upgrades and CC-promo detection enabled. The conservative behavior is preserved when Gradle's agent is disabled; TestKit `withDebug(true)` remains out of scope, see https://github.com/gradle/gradle/issues/23817 (embedded TAPI is internal API and is also out of scope).

The injected plugin classpath's classloader was substituting pre-AOT-rewritten bytes for the JVM-supplied buffer at class load, wiping any earlier `ClassFileTransformer`'s output. It now runs Gradle's bytecode instrumentation against the JVM-supplied bytes when a third-party agent is attached, composing on top of the earlier transformer instead of replacing it; everything else keeps the fast swap path.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/VintageInjectedClasspathInstrumentationStrategy.kt
```diff
@@ -16,15 +16,23 @@
...
+internal class VintageInjectedClasspathInstrumentationStrategy internal constructor(
+    private val agentStatus: AgentStatus,
+) : InjectedClasspathInstrumentationStrategy {
+
+    override fun getTransform(): TransformMode =
+        if (!agentStatus.isAgentInstrumentationEnabled() && isThirdPartyAgentPresent()) {
+            // Without Gradle's instrumentation agent, the buildscript classpath is rewritten ahead of time;
+            // disable instrumentation so the third-party transformer observes the user's bytecode, not Gradle's.
+            // Loses Gradle's bytecode upgrades and CC-promo detection on this classpath in exchange.
+            TransformMode.NONE
```
eskatos, line 34: @mlopatkin, @asodja, this looks like a trap. For CC I assume this is not that bad because it is only about the promo lines, but still. For the Provider API migration it means that with embedded execution modes of Gradle (TestKit's `withDebug(true)` and the internal TAPI embedded) the instrumentation is silently disabled. Note that this PR doesn't change anything in this regard, I just stumbled upon this and wanted to raise it to you. I'll leave it to your teams to see what should be done.

eskatos:
@bot-gradle test RFN

lkasso (approved):
Docs LGTM.

eskatos:
@bot-gradle test RFN

tresat (commented):
No objections to the JVM/Software changes.  I didn't take the time to understand this change fully.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/VintageInjectedClasspathInstrumentationStrategy.kt
hegyibalint: (Sorry, accidentally resolved instead of replied)

I second this opinion. With the Provider API, we would rely more and more on this runtime instrumentation features to keep the setters and getters working, and this difference in behavior could cause serious amount of head-scratching on the plugin developer's end.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/VintageInjectedClasspathInstrumentationStrategy.kt
```diff
@@ -16,15 +16,23 @@
...
+import org.gradle.plugin.use.resolve.service.internal.InjectedClasspathInstrumentationStrategy
 import org.gradle.plugin.use.resolve.service.internal.InjectedClasspathInstrumentationStrategy.TransformMode
 
 
-class VintageInjectedClasspathInstrumentationStrategy(
-    problems: ProblemsListener
-) : AbstractInjectedClasspathInstrumentationStrategy(problems) {
-    override fun whenThirdPartyAgentPresent(): TransformMode {
-        // For now, disable the instrumentation
-        return TransformMode.NONE
```
hegyibalint, line 28: ❓ Does this mean that Provider API binary upgrades doesn't work with CC disabled?
cc. @asodja

platforms/core-runtime/classloaders/src/main/java/org/gradle/internal/classpath/TransformedClassPath.java
```diff
@@ -199,8 +222,11 @@ private TransformedClassPath plusWithTransforms(TransformedClassPath classPath)
             }
         }
 
+        // Receiver wins on the class-load-time transform, matching the receiver-wins rule for the transforms map.
```
hegyibalint: 🤔 Here it's advertised that "matching the receiver-wins rule for the transforms map". The code above, in my interpretation, is a set addition with added semantics over how the classloader works (left-to-right). 

However this one, this is substitution: if we have a transform chain, we will keep that otherwise we will use what's coming in. I don't think we should say this if it's actually not the same.

Am I misinterpreting something here?
If not, I would just simply remove the "matching.*" bit.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTestKitIntegrationTest.groovy
```diff
@@ -27,60 +27,169 @@ import org.gradle.util.internal.TextUtil
...
+        then:
+        !output.contains(JAVA_AGENT_PROBLEM_MESSAGE)
+        !output.contains("Configuration cache problems found")
+
+        where:
+        mode << PluginResolutionMode.values()
+    }
+
+    def "JDWP debug agent does not cause a configuration-cache problem [#mode]"() {
+        // JDWP attaches via `-agentlib:`, not `-javaagent:`, so AgentUtils does not flag it as a
```
reinsch82: ❌ We should still check for `-agentlib` and `-agentpath` to our understanding.

`jdwp` might be safe, but any native agent could be attached which have the same possibilities as "regular" java agents.

reinsch82 (changes requested):
LGTM but we think there are missed command line options

platforms/core-runtime/instrumentation-agent-services/src/main/java/org/gradle/internal/instrumentation/agent/ThirdPartyAgentDetection.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Cached query that reports whether a non-Gradle {@code -javaagent:} switch was passed
+ * to the JVM at startup.
+ */
+public final class ThirdPartyAgentDetection {
+
+    private static final Lazy<Boolean> PRESENT = Lazy.locking().of(() ->
+        ManagementFactory.getRuntimeMXBean().getInputArguments().stream()
+            .anyMatch(AgentUtils::isThirdPartyJavaAgentSwitch)
+    );
```
hegyibalint, line 37: ❌️This solution memoize the value the first time we look at this value. But the JVM can attach agents[^1] dynamically, making this boolean misrepresent the situation.

Technically what could happen that a dynamically loaded agent can retransform classes (with limitations, but it could still do). That means that already loaded classes would be retransformed, but newly loaded ones might keep pulling the AOT-transformed one if we have `PRESENT=false`.

Unfortunately, there is no good way of detecting if we suddenly have a dynamically attached agent. Therefore, the next best thing is to document that **we do not support dynamically attached agents**.

[^1]: This includes the bytecode `-javaagent:` and the native `-agentpath:` or `-agentlib:` as well.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTestKitIntegrationTest.groovy
eskatos: Very good point, thanks!
Fixed, PTAL

reinsch82 (approved):

hegyibalint (changes requested):
Technical implementation LGTM, but I've pointed out an undocumented bit with dynamicly loaded agents. PTAL.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/VintageInjectedClasspathInstrumentationStrategy.kt
eskatos: See my comment above.
It means that Provider API binary upgrades don't work with 3rd party agent, TestKit's `withDebug(true)` and CC disabled.
Note that this PR doesn't change anything in this regard.

platforms/core-runtime/classloaders/src/main/java/org/gradle/internal/classpath/TransformedClassPath.java
eskatos: Good catch. Thanks!

platforms/core-runtime/instrumentation-agent-services/src/main/java/org/gradle/internal/instrumentation/agent/ThirdPartyAgentDetection.java
eskatos: Good point. Added both to javadoc here and in the user manual.

FTR, dynamically attached agents are on the way out (see JEP 451). Java 21+ comes with warnings when an agent is dynamically attached and a new flag to simply disable them (`-XX:-EnableDynamicAgentLoading`).

hegyibalint: Interesting, didn't know about JEP 451!

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -587,3 +587,18 @@ Once generated, set the Base64-encoded key as the value of the `GRADLE_ENCRYPTIO
 ----
 $ export GRADLE_ENCRYPTION_KEY="your-generated-key-here"
 ----
+
+[[config_cache:requirements:java_agent]]
+== Third-party Java Agents with the Configuration Cache
+
+The Configuration Cache supports third-party `-javaagent:` attachments to the build JVM (for example a coverage agent such as Jacoco, IntelliJ's `idea_rt.jar`, or a custom user agent) in regular daemon builds and in <<test_kit#test_kit, TestKit>>'s default (daemon) mode.
+
+Only agents attached at JVM startup are supported. Agents attached dynamically via the Attach API after the JVM has started are not detected and may cause incorrect bytecode observation on classes loaded after attachment.
```
hegyibalint, line 611: 💄 Maybe let's put this into a warning admonition box? Too much?

hegyibalint (approved):
LGTM, one 💄 comment on the docs!

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
eskatos: I'd leave it like that to not catch too much attention. This page is already content-heavy.
Also, the fact that with Java 21 you get warnings about dynamically attached agents anyway makes this less of a thing.

hegyibalint: Fine by me! 👍

subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultScriptClassPathResolver.java
```diff
@@ -156,7 +164,16 @@ public ClassPath resolveClassPath(Configuration classpathConfiguration, ScriptCl
 
             MethodInterceptionReportCollector reportCollector = propertyUpgradeReportConfig.getReportCollector();
             instrumentedClasspath.getOrDefault(INTERCEPTED_METHODS_REPORT, Collections.emptyList()).forEach(reportCollector::collect);
-            return TransformedClassPath.handleInstrumentingArtifactTransform(instrumentedClasspath.getOrDefault(ARTIFACT, Collections.emptyList()));
+            ClassPath classPath = TransformedClassPath.handleInstrumentingArtifactTransform(instrumentedClasspath.getOrDefault(ARTIFACT, Collections.emptyList()));
+            if (composeWithThirdPartyAgent && classPath instanceof TransformedClassPath) {
+                InstrumentationTypeRegistry registry = buildService.getInstrumentationTypeRegistry(contextId);
+                ClassLoadTimeTransform classLoadTimeTransform = new InstrumentingClassLoadTimeTransform(
```
asodja: ❓ It looks like here, that if we have third party agent, we first do instrumentation + bytecode upgrades on artifacts via artifact transform and then we do the same at class loading time. 

So it looks like we do some work twice, while in the end we only use class load transform judging from this code in [VisitableURLClassLoader.java‎](https://github.com/gradle/gradle/pull/37909/changes#diff-95560805d5d82e973969fc2511c4b7c48165da7212b110f1b79785b4e70b21f4):
```
if (classLoadTimeTransform != null && className != null) {
    // A third-party agent is attached; rewrite the JVM-supplied buffer so we layer
    // Gradle's instrumentation on top of the third-party transformer's output.
    return classLoadTimeTransform.transform(className, classfileBuffer);
}
// No third-party agent to compose with; substitute the bytes Gradle rewrote during artifact transform.
return replacer.getInstrumentedClass(className, protectionDomain);
```

This is probably not a problem, just an observation. Is that intentional?

subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultScriptClassPathResolver.java
```diff
@@ -156,7 +164,16 @@ public ClassPath resolveClassPath(Configuration classpathConfiguration, ScriptCl
...
+                    BytecodeInterceptorFilter.INSTRUMENTATION_AND_BYTECODE_UPGRADE,
```
asodja: ❓ I think this changes a behaviour a bit. For "Project dependencies" we do only instrumentation, but not bytecode upgrades, see this line:
```
ArtifactCollection instrumentedProjectDependencies = getInstrumentedProjectDependencies(classpathConfiguration);
```

But here if I understand correctly we do `INSTRUMENTATION_AND_BYTECODE_UPGRADE` for project dependencies too. Is that needed?

subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultScriptClassPathResolver.java
```diff
@@ -156,7 +164,16 @@ public ClassPath resolveClassPath(Configuration classpathConfiguration, ScriptCl
 
             MethodInterceptionReportCollector reportCollector = propertyUpgradeReportConfig.getReportCollector();
             instrumentedClasspath.getOrDefault(INTERCEPTED_METHODS_REPORT, Collections.emptyList()).forEach(reportCollector::collect);
-            return TransformedClassPath.handleInstrumentingArtifactTransform(instrumentedClasspath.getOrDefault(ARTIFACT, Collections.emptyList()));
+            ClassPath classPath = TransformedClassPath.handleInstrumentingArtifactTransform(instrumentedClasspath.getOrDefault(ARTIFACT, Collections.emptyList()));
+            if (composeWithThirdPartyAgent && classPath instanceof TransformedClassPath) {
```
asodja: 💅 Could we export this part to a new method? Maybe "maybeAttachClassLoadTransform" or just `attachClassLoadTransform` or something better

platforms/core-runtime/classloaders/src/main/java/org/gradle/internal/classloader/VisitableURLClassLoader.java
```diff
@@ -164,15 +165,24 @@ private static class InstrumentingVisitableURLClassLoader extends VisitableURLCl
...
         public InstrumentingVisitableURLClassLoader(String name, ClassLoader parent, TransformedClassPath classPath) {
             super(name, parent, classPath);
             replacer = new TransformReplacer(classPath);
             errorHandler = new TransformErrorHandler(name);
+            classLoadTimeTransform = classPath.getClassLoadTimeTransform();
         }
 
         @Override
         public byte @Nullable [] instrumentClass(@Nullable String className, @Nullable ProtectionDomain protectionDomain, byte[] classfileBuffer) {
+            if (classLoadTimeTransform != null && className != null) {
```
asodja, line 180: ❓ Does this condition mean we entirely ignore "instrumented classes" written on a disk in case of 3rd party agent and we reinstrument on class load?

Also is order important - If I understand correctly comments third-party transformer is run first and then we run Gradle transform?

subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultScriptClassPathResolver.java
eskatos: Good catch!
This is tricky but doable using a strategy similar to what `TransformReplacer` does.
Work in progress...

subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultScriptClassPathResolver.java
eskatos: Nice touch, done in 42873bbada8549f755208fcfa3176cd5c08d8810

subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultScriptClassPathResolver.java
eskatos: Done in 447c6f53b2b144965e012b6ee6c4aabea382dc08

subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultScriptClassPathResolver.java
eskatos: Good catch, and yes this is intentional. When a third party agent is present the artifact transforms still run and write instrumented jars, but the class loader reads the original entries and reinstruments them on load, so those instrumented jars are never used for loading. It is not entirely wasted though: the analysis pass that runs alongside still populates the type registry that the class load transform consumes, so only the jar writing is redundant. Correctness is unaffected, since the class load transform produces the bytes that actually run. Skipping the jar writing would need a separate resolution that omits the instrumenting transform output while keeping the analysis. That might be a reasonable optimization for later if it proves to be problematic. I expect third party agents to mostly be used for debugging or testing though.

I added a code comment so this is captured where it happens.

platforms/core-runtime/classloaders/src/main/java/org/gradle/internal/classloader/VisitableURLClassLoader.java
eskatos: Yes on both counts. When a third party agent is present we ignore the instrumented classes on disk and reinstrument on load from the buffer the JVM supplies.

Order matters and your reading is correct: the third party transformer runs first and Gradle layers its instrumentation on top. That is the only correct order for coverage agents, since the agent must see the user bytes so its recorded data maps back to the compiled class, and Gradle applies afterwards. This relies on Gradle's transformer being last in the JVM transformer chain, which the daemon arranges.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/VintageInjectedClasspathInstrumentationStrategy.kt
mlopatkin: >'ll leave it to your teams to see what should be done.

Yes, we'll have to enforce having the agent at some point. I don't really see any other way. As I said, we can even make it work with embedded TAPI at a cost of requiring some version alignment/backporting solution.

mlopatkin (approved):
LGTM, thanks!

asodja (approved):
LGTM

PR #37918, opened 2026-05-18
Introduce IP Diagnostics Mode

### LINKED ISSUE #34799: Isolated Projects migration mode
### Context
One of the goals of Isolated Projects is to deliver parallel project configuration. Safe parallelism requires the absence of shared mutable state accesses between the units of parallelism, which in this case are projects.

We already have measures in place to detect unwanted accesses and report them as problems. If no problems were detected during a build, it means it produces a correct outcome even with parallelism involved. However, if at least one problem is detected, we can no longer be certain about the correctness of the outcomes/outputs.

This poses a question of how Gradle should handle the case when the problems are present. This is especially important for the period of build migration to support IP, or even before, when users want to understand the potential migration costs.

### Problem

Today, IP problem reporting relies on and behaves similarly to CC serialization problem reporting. The most problematic part is that each IP problem is treated as a _deferred problem_ and the execution of the build carries on. Of course, by default, the presence of any problems results in the build failure. However, the build failure does not happen until the end of the build, meaning that the build can still be executing work with a corrupted, stale, or otherwise wrong state.

We could consider pursuing the opposite approach: immediately fail on any IP problem. This will make sure no work runs over bad state. However, this presents a challenge for the migration period. The discovery of IP problems will become sequential: there will be no way of estimating the migration costs upfront, and no way of measuring progress.

### Proposal

Introduce a dedicated **Migration Mode** with all problem detection but no parallelism. While enabling IP without the migration can be treated as the "production mode", it does not have to have a name.

The migration mode can set a reasonable user expectation that it will not provide any performance benefits. Its sole goal is to make the migration experience as easy as possible. The "production mode," on the other hand, will run with full performance benefits, at the expense of not providing a complete picture of problems.

alllex:
Fixes https://github.com/gradle/gradle/issues/34799

Adds an opt-in `org.gradle.unsafe.isolated-projects.diagnostics=true` that disables the IP optimizations which mask or reorder violations: parallel project configuration, parallel CC store, per-project tooling model reuse, and internal configure-on-demand.

**Notes on scope**

- The IP problem severity is unchanged. `onIsolatedProjectsProblem` still reports as `Deferred` in both default and Diagnostics mode. The "fail fast on IP problems" half of the issue is left for a follow-up so that landing the diagnostic surface and changing the default behavior remain separable.
- Top-level Configuration Cache reuse is *preserved* in Diagnostics mode. Violations are caught on the cache-miss run; identical re-runs replay the entry. Users invalidate CC the usual way when they want to re-exercise configuration.
- "Realize all tasks" from the issue's implementation list is not done here. Diagnostics mode inherits whatever the rest of IP does for task realization.

**Test scope**

Existing IP integration tests that depend on collecting *multiple* violations in one invocation (multi-`problem(...)` or `problem(..., count ≥ 2)` assertions) are switched to the diagnostics runner; the rest stay on default IP. New tests cover parameter wiring, end-to-end multi-violation collection, that top-level CC reuse is preserved while per-project model reuse is not, and that all projects are configured even when configure-on-demand would skip some.

### DIFF
--- a/platforms/core-runtime/start-parameter/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
+++ b/platforms/core-runtime/start-parameter/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
@@ -90,6 +90,7 @@ public class StartParameterBuildOptions extends BuildOptionSet<StartParameterInt
         new ConfigurationCacheHeapDumpDir(),
         new ConfigurationCacheFineGrainedPropertyTracking(),
         new IsolatedProjectsOption(),
+        new IsolatedProjectsDiagnosticsOption(),
         new ProblemReportGenerationOption(),
         new PropertyUpgradeReportOption(),
         new TaskGraphOption(),
@@ -687,6 +688,19 @@ public class StartParameterBuildOptions extends BuildOptionSet<StartParameterInt
         @Override
         public void applyTo(boolean value, StartParameterInternal settings, Origin origin) {
             settings.setIsolatedProjects(Option.Value.value(value));
+        }
+    }
+
+    public static class IsolatedProjectsDiagnosticsOption extends BooleanBuildOption<StartParameterInternal> {
+        public static final String PROPERTY_NAME = "org.gradle.unsafe.isolated-projects.diagnostics";
+
+        public IsolatedProjectsDiagnosticsOption() {
+            super(PROPERTY_NAME);
+        }
+
+        @Override
+        public void applyTo(boolean value, StartParameterInternal settings, Origin origin) {
+            settings.setIsolatedProjectsDiagnostics(value);
         }
     }
 
--- a/platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
@@ -463,11 +463,10 @@ In that case the build will fail, and one or more violations will be shown in th
 [[sec:diagnostics_mode]]
 === Diagnostics mode
 
-Diagnostics mode is an opt-in mode that helps you discover Isolated Projects <<sec:constraint_violations,constraint violations>> more efficiently during migration.
-It runs all of <<sec:perf_tasks,configuration>> *sequentially*, so every violation across every project is reported in a single invocation, regardless of where it occurs.
+_Diagnostics_ mode is an *opt-in* mode that helps you discover Isolated Projects <<sec:constraint_violations,constraint violations>> more efficiently during migration.
+It runs all <<sec:perf_tasks,configuration>> *sequentially*, so every violation across every project is reported in a single invocation, regardless of where it occurs.
 
-Enable Diagnostics mode by setting `org.gradle.unsafe.isolated-projects.diagnostics=true` in `gradle.properties`,
-or by passing it on the command line alongside `-Dorg.gradle.unsafe.isolated-projects=true`:
+Enable _Diagnostics_ mode by setting `org.gradle.unsafe.isolated-projects.diagnostics=true` in `gradle.properties`, or by passing it on the command line alongside `-Dorg.gradle.unsafe.isolated-projects=true`:
 
 [source,bash]
 ----
@@ -476,29 +475,25 @@ $ ./gradlew help \
     -Dorg.gradle.unsafe.isolated-projects.diagnostics=true
 ----
 
-Compared to normal Isolated Projects, Diagnostics mode disables several optimizations to maximize the completeness of the violation report:
+Compared to normal Isolated Projects, _Diagnostics_ mode disables several optimizations to maximize the completeness of the violation report:
 
 * Project configuration runs *sequentially* instead of in parallel.
 * Configuration Cache store runs *sequentially*.
 * Per-project model reuse across builds is *disabled*, so every model is rebuilt after an input change.
 * All projects are configured, ignoring any configure-on-demand setting.
 
-These trade-offs make Diagnostics mode to be unsuitable for day-to-day operations.
+These trade-offs make _Diagnostics_ mode unsuitable for day-to-day operations.
 Use it only to enumerate violations, such as during the migration, or to compare against the default Isolated Projects path when investigating a discrepancy.
 
 [NOTE]
 ====
-Diagnostics mode does not affect top-level Configuration Cache reuse.
-If you run the same invocation twice without any input changes, the second run still reuses the cached entry and reports nothing,
-because all violations were already reported on the cache-miss run.
+_Diagnostics_ mode does not affect top-level Configuration Cache reuse.
+If you run the same invocation twice without any input changes, the second run still reuses the cached entry and reports nothing, because all violations were already reported on the cache-miss run.
 Modify a build input (for example, edit a build script) between runs to re-exercise configuration.
 ====
 
-[NOTE]
-====
-Diagnostics mode does not change which patterns count as violations.
-A build that is clean under Isolated Projects is also clean under Diagnostics mode.
-====
+_Diagnostics_ mode does not change which patterns count as violations.
+A build that is clean under Isolated Projects is also clean under _Diagnostics_ mode.
 
 [[sec:ignore_violations]]
 === Temporarily ignoring violations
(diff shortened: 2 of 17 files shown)


alllex:

lkasso (approved):
Docs LGTM with my minor changes.

platforms/core-runtime/start-parameter/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
```diff
@@ -690,6 +691,19 @@ public void applyTo(boolean value, StartParameterInternal settings, Origin origi
         }
     }
 
+    public static class IsolatedProjectsDiagnosticsOption extends BooleanBuildOption<StartParameterInternal> {
+        public static final String PROPERTY_NAME = "org.gradle.unsafe.isolated-projects.diagnostics";
```
octylFractal, line 695: ❓ What's the expected evolution of this when we change to `--isolated-projects`? Also, when we add unsafe/evaluation mode, how is it going to interact with that? I feel like it would be better to have it either remain part of the existing IP option (`-Dorg.gradle.unsafe.isolated-projects=false/true/diagnostics/evaluation`?) or a new "mode" option (`-Dorg.gradle.unsafe.isolated-projects.mode=default/diagnostics/evaluation`?).

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -457,6 +460,41 @@ org.gradle.unsafe.isolated-projects=true
 If you have never enabled the feature in the build before, it's likely that the build is not compatible and contains Isolated Projects <<sec:constraint_violations,constraint violations>>.
 In that case the build will fail, and one or more violations will be shown in the error message.
 
+[[sec:diagnostics_mode]]
+=== Diagnostics mode
+
+_Diagnostics_ mode is an *opt-in* mode that helps you discover Isolated Projects <<sec:constraint_violations,constraint violations>> more efficiently during migration.
+It runs all <<sec:perf_tasks,configuration>> *sequentially*, so every violation across every project is reported in a single invocation, regardless of where it occurs.
```
octylFractal: :x: This is a non sequitur as currently worded, what about:
```suggestion
It runs all <<sec:perf_tasks,configuration>> *sequentially*, which allows the build to safely continue even after an IP violation. This means that every violation across every project can be reported in a single invocation, regardless of where it occurs.
```

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -457,6 +460,41 @@ org.gradle.unsafe.isolated-projects=true
...
+----
+
+Compared to normal Isolated Projects, _Diagnostics_ mode disables several optimizations to maximize the completeness of the violation report:
+
+* Project configuration runs *sequentially* instead of in parallel.
+* Configuration Cache store runs *sequentially*.
+* Per-project model reuse across builds is *disabled*, so every model is rebuilt after an input change.
+* All projects are configured, ignoring any configure-on-demand setting.
+
+These trade-offs make _Diagnostics_ mode unsuitable for day-to-day operations.
```
octylFractal: :x: I don't know if this is the right wording. We would like to encourage the use of this in CI to validate that no regression in IP compatibility is introduced by changes to a build, which I think is still "day-to-day". What about:
```suggestion
These trade-offs make _Diagnostics_ mode unsuitable for typical operations.
```

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -457,6 +460,41 @@ org.gradle.unsafe.isolated-projects=true
...
+Use it only to enumerate violations, such as during the migration, or to compare against the default Isolated Projects path when investigating a discrepancy.
+
+[NOTE]
+====
+_Diagnostics_ mode does not affect top-level Configuration Cache reuse.
+If you run the same invocation twice without any input changes, the second run still reuses the cached entry and reports nothing, because all violations were already reported on the cache-miss run.
+Modify a build input (for example, edit a build script) between runs to re-exercise configuration.
```
octylFractal: ❓ Why do we do this? If we're already deoptimizing on purpose, what prevents us from deoptimizing here too?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -509,6 +547,9 @@ There can be more violations hidden by laziness, such as in the task configurati
 +
 You can find the Isolated Projects violations in the Configuration Cache HTML report.
 Use the report to pinpoint the incompatible pieces of build logic and plugins.
++
+For builds with many projects, also enable <<sec:diagnostics_mode,Diagnostics mode>>
+to surface violations across every project in a single invocation.
```
octylFractal: :x: We should say something about when to remove it too.

platforms/core-runtime/start-parameter/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
6hundreds: +1

alllex: What I envision is to have a familiar `--(no-)isolated-projects` CLI switch. However, the full Gradle option might also control whether IP is applied for tasks vs models, e.g. `org.gradle.isolated-projects=tooling`. This will then control only the _enablement_ of feature/constraints it represents. Any other aspects are controlled by additional knobs.

Specifically for the unsafe mode, it seems sufficient to go with a simple flag that only controls the way we treat problems, e.g. `org.gradle.isolated-projects.dangerously-ignore-problems=true/false`. It does what it says on the tin, and also allows getting a "successful" build even in the presence of violations, which is useful for evaluating performance gains for IDE Sync.

Then, the Diagnostics mode stands on its own. Importantly, in that mode, we preserve build outcome parity with the default IP. That means Diagnostics mode fails **iff** default IP fails. This makes it easier for humans (and agents!) to iterate on the migration until the build becomes successful.

The Diagnostics mode can then combine with the ignore-problems mode in an expected manner, by allowing runs with violations to succeed. It can also be useful for troubleshooting IDE Sync.

Having said that, we don't know enough yet to judge the final form of these knobs. We are better off shipping independent toggles and then later deciding to combine them in modes, once we learn more about how they are used and what users think.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
alllex: If there are any violations, the CC entry won't be stored or will be discarded. So we are only talking about a violation-free invocation. If the build is fully IP and CC compatible, I don't see a strong reason to re-run the configuration.

> we're already deoptimizing on purpose

Our goal is not to deoptimize, but to reintroduce the determinism of project configuration, while still detecting IP violations. The part about reconfiguring all projects is about surfacing more potential issues. We might in the future decide to force-evaluate more configuration work with the same goal.

If a de-optimization doesn't contribute to the user getting more useful information, then I see no reason why we'd intentionally make things slower.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
alllex: I'm okay with the alternative wording. I'll apply it.

Regarding the recommendation of using it on CI, I'm not sure what the purpose would be.

If you've already migrated to IP, then IP should just be enabled on CI (making things faster!). If a violation is introduced, just like any problem failing some build validation, the build will fail, and that would be the signal. The D-mode can then be used to investigate.

If you are still migrating to IP and want something on CI, then you should have the IP-violation baselining mechanism that we haven't introduced yet.

Did you have something else in mind for the CI use case?

octylFractal: I was referring to the baselining mechanism.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
octylFractal: I would reword this to be clearer then. I read it as we would still store CC entry on violations for some reason.

alllex: Reworded

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
alllex: I ended up rewriting this section. I also noted that the D-mode shouldn't be enabled persistently in the middle of its own section

alllex:

lkasso (approved):
Docs look fantastic!

Issue #25488, opened 2023-06-22
Reduce exec process threads


octylFractal:
Currently, we spawn between 3 and 4 threads for every process started from Gradle, one for waiting on the process, one each for stdout and stderr, and sometimes one for stdin. This can likely be improved by using async methods such as NIO to combine all I/O work to as little as one thread, or manual JNI binding to `waitpid` to avoid the need to have a thread-per-process by watching all child exits in a single thread, or potentially not watching for the exit if that information doesn't need to be reacted to (e.g. in the case of daemon-scope workers, there's no need to wait for their exit).

Some of these options become much easier if we were to move Gradle to a newer Java version, such as `Process#onExit() -> CompleteableFuture<Process>` (JDK 9) which avoids the need to manually spawn a thread to watch for exits and does so only on demand, though it still currently adds a thread-per-process. Starting with the upcoming JDK 21, we will have virtual threads, allowing us to easily move the I/O work to virtual threads and reduce our OS thread allocations.

These improvements would reduce with the memory consumption of Gradle by at least 2-4MB per process (assuming a 1MB thread stack size, the default on my machine), and for builds that start hundreds of processes this adds up to massive savings.

octylFractal:
For minimizing the I/O work, there are a few options:
- Use https://github.com/brettwooldridge/NuProcess or write our own version of it. This is likely quite involved.
- Wait for https://bugs.openjdk.org/browse/JDK-4483582 to be resolved. It doesn't look like that's going to happen any time soon. It would also need to be resolved with "selectable" channels, and not just exposing a FileChannel as that's currently subject to the same issue as below.
- Using Virtual Threads actually doesn't help here, as there is no proper non-blocking file I/O in Java currently. So a virtual thread effectively does the same thing that we do, and just blocks on a spare thread. Perhaps some day this won't be the case and we can do this as a simple option.

PR #37924, opened 2026-05-19
Re-use stream thread for waiting for process exit 

octylFractal:
Part of #25488

This allows us to save the thread normally used for that. Unfortunately, we couldn't just use `Process.onExit()` as it always enqueues a task on the ForkJoinPool, which can lead to deadlocks or slowdowns if all ForkJoinPool threads are occupied blocking on work that requires a process exit to be reaped. The benefit of this is that even on Java 8, we will still save the thread regardless!

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


big-guy (approved):

octylFractal:
@bot-gradle test this

big-guy (approved):

PR #37931, opened 2026-05-19
Introduce MRJAR handling to GBT build

octylFractal:
This is in preparation to refactor some of our currently reflection-based or wrong-stdlib logic to be easier to write, test, and read.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic-commons/module-identity/src/main/kotlin/gradlebuild/identity/extension/GradleModuleExtension.kt
+++ b/build-logic-commons/module-identity/src/main/kotlin/gradlebuild/identity/extension/GradleModuleExtension.kt
@@ -21,10 +21,18 @@ import org.gradle.api.Project
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
+import org.gradle.api.provider.ProviderFactory
 import org.gradle.api.tasks.Nested
 import org.gradle.api.tasks.TaskContainer
 import org.gradle.util.GradleVersion
 import java.util.Optional
+import javax.inject.Inject
+
+/**
+ * The JVM version everything compiles to by default, and the minimum JVM the test
+ * tasks run on.
+ */
+const val DEFAULT_TARGET_JVM_VERSION: Int = 17
 
 /**
  * A project extension which describes a module of a Gradle distribution.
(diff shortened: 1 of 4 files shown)


.teamcity/jdks.yaml
```diff
@@ -3,6 +3,7 @@
 # ```
 # stopGradleDaemons=true
 # ```
+# When updating this file, also update the `JvmCompileExtension.AVAILABLE_JDK_VERSIONS`.
```
cobexer: ❌ https://github.com/gradle/gradle/blob/a2bfcc5455e004b820ef1aeb9bb4e48c6fee0078/.github/workflows/update-jdks.yml#L29-L39 breaks this again

But: can we avoid having to do this?

build-logic-commons/module-identity/src/main/kotlin/gradlebuild/identity/extension/GradleModuleExtension.kt
```diff
@@ -96,24 +98,43 @@ interface ModuleIdentity {
...
 
 
     /**
      * Declare that this Gradle module runs as part of worker process.
      */
-    val worker: Property<Boolean>
+    abstract val worker: Property<Boolean>
+
+    fun computeProductionJvmTargetVersion(): Provider<Int> {
+        // Should be kept in sync with org.gradle.internal.jvm.SupportedJavaVersions
```
cobexer, line 130: 💭 Could we generate a properties file with the information from below and read that back in `SupportedJavaVersions` so we don't have to sync them?

build-logic-commons/module-identity/src/main/kotlin/gradlebuild/identity/extension/GradleModuleExtension.kt
```diff
@@ -96,24 +98,43 @@ interface ModuleIdentity {
...
+            client to 8,
+            daemon to 17,
+            worker to 8
+        )
+
+        // By default, compile to 17. This ensures projects that do not declare any target runtimes
+        // can successfully resolve their classpaths. Then, `:checkTargetRuntimes` will ensure that
+        // each project declares the proper target runtimes. Note that `:checkTargetRuntimes` cannot
+        // execute unless all project classpaths can be successfully resolved.
+        return reduceBooleanFlagValues(targetRuntimeJavaVersions, ::minOf).orElse(17)
```
cobexer: ❓ Could we re-use `JvmCompileExtension.DEFAULT_TARGET_JVM_VERSION` instead of duplicating the value?

build-logic/jvm/src/main/kotlin/gradlebuild/jvm/JvmCompileExtension.kt
```diff
@@ -30,10 +45,53 @@ abstract class JvmCompileExtension {
...
+         * The JVM version everything compiles to by default, and the minimum JVM the test
+         * tasks run on. Production compilations may lower this via [JvmCompilation.targetJvmVersion].
+         */
+        const val DEFAULT_TARGET_JVM_VERSION: Int = 17
+
+        /**
+         * JDK versions provisioned in TeamCity that can be used for testing higher Java versions in multi-release JARs.
+         * This should be kept in sync with the majors listed in `.teamcity/jdks.yaml`.
+         */
+        private val AVAILABLE_JDK_VERSIONS: Set<Int> = setOf(17, 21, 25, 26)
```
cobexer: ❓ Why do we need them? How is this different in CI vs locally? Can we avoid this duplication?

build-logic/jvm/src/main/kotlin/gradlebuild/jvm/JvmCompileExtension.kt
```diff
@@ -55,4 +113,111 @@ abstract class JvmCompileExtension {
...
+            "Multi-release version ($version) must be greater than the production target JVM version " +
+                "($productionTarget); otherwise the base classes already cover it."
+        }
+        val sourceSets = the<SourceSetContainer>()
+        val main = sourceSets.getByName("main")
+        val test = sourceSets.getByName("test")
+
+        val versioned = setUpMainMultiReleaseSourceSet(version, main)
+
+        val testJvmVersion = AVAILABLE_JDK_VERSIONS.firstOrNull { it >= version }
```
cobexer: ❌ `addMultiReleaseVersion` should have a second argument that defines this version: whoever introduces version specific logic should define how it's supposed to be tested. That then defines what engineers need to have installed locally and what needs to be provisioned on CI.

cobexer (changes requested):
How many different Java versions are we dealing with here? could you link to the usecases this needs to support?

octylFractal:
The primary use case is #37932, but I'd like to change all of our reflection or stdlib-based Java API handling to use this long-term. For now, I suppose it just needs support for 9, but we would probably also want 24 somewhat soon for the Class-file API.

build-logic-commons/module-identity/src/main/kotlin/gradlebuild/identity/extension/GradleModuleExtension.kt
octylFractal: I didn't create this code, it was just moved. I don't really want to adjust it in this PR as it's a separate concern.

build-logic/jvm/src/main/kotlin/gradlebuild/jvm/JvmCompileExtension.kt
octylFractal: Sure, I guess that makes sense. I'll change to that.

cobexer (approved):
LGTM!

octylFractal:
@big-guy might have had feedback, so I'm going to wait until he responds before merging

big-guy (approved):

PR #37933, opened 2026-05-19
Provide stable JUnit XML class for non-class tests

### LINKED ISSUE #37850: Regression in Gradle 9: Cucumber JUnit Platform tests break retry execution and lose test names in Develocity Build Scans
### Current Behavior

We observe a regression when running Cucumber JUnit Platform tests with Gradle 9.x in combination with the Develocity Gradle plugin and test retry enabled.

A failing Cucumber test (expected failure) behaves as follows:

### With Gradle 9.x + retry enabled
`./gradlew clean build`

- The Cucumber test fails as expected (intentional failure)
- Test retry is triggered but fails with:
`Selected test name cannot be null or empty`
- As a result, the build fails before completion
- No Build Scan is published due to the retry failure

### With Gradle 9.x + retry disabled
- The build completes and a Build Scan is published
- However, Cucumber scenario names are shown as N/A in the Build Scan instead of the correct scenario names

### Expected Behavior

- Cucumber test failures should behave consistently across Gradle versions
- Test retry should work for failed Cucumber JUnit Platform tests
- Build Scans should always be published, even when retries occur or fail
- Scenario names should be correctly resolved and displayed in Build Scans (not N/A)
- No difference in behavior between Gradle 8 and Gradle 9 for these cases

### Context (optional)

The underlying Cucumber test failure is identical and expected in both Gradle 8 and Gradle 9.

The regression only affects:
- retry execution of failed Cucumber tests
- Build Scan publication in combination with retry
- test name resolution in Build Scans

The issue appears to be related to Gradle 9 changes in test execution / filtering combined with Develocity retry handling for JUnit Platform dynamic tests.

A minimal reproducer is available here:

https://github.com/lschwahn/cucumber-retry-reproducer

#### Gradle 9.5.0 (failing behavior)
`./gradlew clean build`

Observed:
- expected test failure occurs
- retry fails with:
`Selected test name cannot be null or empty`
- Build Scan is not published
- disabling testRetry in "app/build.gradle" allows Build Scan publication, but scenario names appear as N/A

#### Gradle 8.14 (working behavior)
```
./gradlew wrapper --gradle-version=8.14
./gradlew clean build
```

Observed:
- expected test failure occurs
- retry executes successfully with expected test failure
- Build Scan is published
- correct scenario names are shown in Build Scan

### Gradle version

9.5.0

8.14

### Build scan URL (optional)

_No response_

Develocity Gradle plugin: 4.4.1
Cucumber: 7.23.0
JUnit: 5.12.2
Java: 21

tresat:
Previously, non-class-based tests, such as Cucumber scenarios, produced empty or generic class names in JUnit XML reports. This caused the test-retry plugin to fail due to
invalid empty identifiers.

This change ensures a stable, non-empty identifier, typically the parent container's name (e.g., the feature file path), is used for these test entries.

Fixes #37850.

### DIFF
--- a/platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/CucumberNonClassBasedTestingIntegrationTest.groovy
+++ b/platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/CucumberNonClassBasedTestingIntegrationTest.groovy
@@ -150,7 +150,88 @@ class CucumberNonClassBasedTestingIntegrationTest extends AbstractIntegrationSpe
         result.testPath(":src/test/resources/helloworld.feature:Say hello /two/three").onlyRoot().assertHasResult(TestResult.ResultType.SUCCESS)
     }
 
-    private writeCucumberStepDefinitions(String path = "src/test/java") {
+    @Issue("https://github.com/gradle/gradle/issues/37850")
+    def "test-retry plugin can filter a failed Cucumber scenario by class+method name"() {
+        given:
+        settingsFile << """
+            pluginManagement {
+                repositories {
+                    gradlePluginPortal()
+                }
+            }
+        """
+        buildFile << """
+            plugins {
+                id 'java'
+                id 'org.gradle.test-retry' version '1.6.4'
+            }
+
+            ${mavenCentralRepository()}
+
+            testing.suites.test {
+                useJUnitJupiter()
+
+                dependencies {
+                    implementation("io.cucumber:cucumber-java:7.15.0")
+                    runtimeOnly("io.cucumber:cucumber-junit-platform-engine:7.15.0")
+                }
+
+                targets.all {
+                    testTask.configure {
+                        testDefinitionDirs.from("src/test/resources")
+                        retry {
+                            maxRetries = 1
+                        }
+                    }
+                }
+            }
+        """
+
+        writeCucumberFeatureFiles()
+        writeAlwaysFailingCucumberStepDefinitions()
+
+        when:
+        // The scenario always fails; we expect the build to fail. The point of this test is to verify
+        // that the test-retry plugin can *identify* the failed scenario (className non-empty) and
+        // call TestFilter.includeTest() without tripping DefaultTestFilter.validateName.
+        // Before the fix, this would throw "Selected test name cannot be null or empty." on the
+        // attempted retry, masking the actual scenario failure.
+        fails("test")
+
+        then:
+        // The retry plugin successfully formed a className+methodName pair from the descriptor —
+        // the scenario name shows up in the retry plugin's output, proving the className contract.
+        failure.assertHasErrorOutput("helloworld.feature")
+        failure.assertHasErrorOutput("Say hello /two/three")
+        // Pre-fix regression marker; must not appear.
+        failure.assertNotOutput("Selected test name cannot be null or empty")
+    }
+
+    private writeAlwaysFailingCucumberStepDefinitions(String path = "src/test/java") {
+        def stepDefFile = file("$path/HelloStepdefs.java")
+        stepDefFile.parentFile.mkdirs()
+
+        stepDefFile.text = """
+            import io.cucumber.java.en.Given;
+            import io.cucumber.java.en.Then;
+            import io.cucumber.java.en.When;
+
+            public class HelloStepdefs {
+                @Given("^I have a hello app with Howdy and /four")
+                public void I_have_a_hello_app_with() { }
+
+                @When("^I ask it to say hi and /five/six/seven")
+                public void I_ask_it_to_say_hi() { }
+
+                @Then("^it should answer with Howdy World")
+                public void it_should_answer_with() {
+                    throw new AssertionError("Intentional failure to trigger test-retry");
+                }
+            }
+        """
+    }
+
+private writeCucumberStepDefinitions(String path = "src/test/java") {
         def stepDefFile = file("$path/HelloStepdefs.java")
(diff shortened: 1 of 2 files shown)


tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/CucumberNonClassBasedTestingIntegrationTest.groovy
```diff
@@ -107,7 +108,130 @@ class CucumberNonClassBasedTestingIntegrationTest extends AbstractIntegrationSpe
...
+        // call TestFilter.includeTest() without tripping DefaultTestFilter.validateName.
+        // Before the fix, this would throw "Selected test name cannot be null or empty." on the
+        // attempted retry, masking the actual scenario failure.
+        fails("test")
+
+        then:
+        // The retry plugin successfully formed a className+methodName pair from the descriptor —
+        // the scenario name shows up in the retry plugin's output, proving the className contract.
+        failure.assertHasErrorOutput("helloworld.feature")
+        failure.assertHasErrorOutput("Say hello /two/three")
```
octylFractal, line 209: ❓ Should this assert something about the specific failure? Like, that it's a test failure specifically?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/CucumberNonClassBasedTestingIntegrationTest.groovy
```diff
@@ -107,7 +108,130 @@ class CucumberNonClassBasedTestingIntegrationTest extends AbstractIntegrationSpe
...
+        succeeds("test")
+
+        then:
+        // Before the fix, the className on each Cucumber scenario descriptor was the
+        // empty string `JUnitPlatformSupport.NON_CLASS`, which broke Build Scan reporting ("N/A") and made
+        // `DefaultTestFilter.validateName` throw "Selected test name cannot be null or empty." when the
+        // test-retry plugin tried to rerun a failed scenario. The className must be a stable, non-empty
+        // identifier; here we assert the scenario is reachable under its feature file path.
+        def result = resultsFor()
+        result.testPath(":src/test/resources/helloworld.feature:Say hello /two/three").onlyRoot().assertHasResult(TestResult.ResultType.SUCCESS)
```
octylFractal: ❓ I'm confused how this shows anything about `className`? Haven't we always reported the test path this way?

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/CucumberNonClassBasedTestingIntegrationTest.groovy
tresat: It is checking that the test task failed.  I'll add a check that this specific test failed.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/CucumberNonClassBasedTestingIntegrationTest.groovy
tresat: This is actually my fault.  I originally generated a test that DID do this, then when reviewing the change, switched to using the assertions to verify the results instead of inspecting the XML directly (to simplify it).  I'll revert this.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/CucumberNonClassBasedTestingIntegrationTest.groovy
octylFractal: Did you forget to push that up?

tresat: Reverted it by accident while restoring the other change, whoops.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/CucumberNonClassBasedTestingIntegrationTest.groovy
```diff
@@ -107,7 +109,133 @@ class CucumberNonClassBasedTestingIntegrationTest extends AbstractIntegrationSpe
...
+            pluginManagement {
+                repositories {
+                    gradlePluginPortal()
+                }
+            }
+        """
+        buildFile << """
+            plugins {
+                id 'java'
+                id 'org.gradle.test-retry' version '1.6.4'
```
octylFractal, line 167: 💭 I wonder if maybe we should keep this to a smoke test since we're testing an external plugin here. But I'm not confident.

octylFractal (approved):

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/nonclassbased/CucumberNonClassBasedTestingIntegrationTest.groovy
tresat: I think test-retry is different, since there's so many other references to it from within the Gradle codebase.  Going to leave it here for now, as I'd expect to find this with the other Cucumber tests.

PR #37935, opened 2026-05-20
Update Documentation: add docs about querying models and resilient sync

### LINKED ISSUE #36873: Document new TAPI resilient features and behaviour
With Gradle 9.4.0 we completed the main work for resilient sync features. 

We should document the new TAPI behaviour and API in https://docs.gradle.org/current/userguide/tooling_api.html.
We should mention and provide an example for BuildController.fetch. We should explain how it differes to BuildController.getModel/findModel. 
Additionally we should also mention how BuildActions behave and what were changes there with Gradle 9.4.0 regarding to running tasks.

asodja:
Fixes https://github.com/gradle/gradle/issues/36873

### DIFF
--- a/platforms/documentation/docs/src/snippets/integration/fetchModel/common/src/main/java/org/gradle/sample/FetchEclipseModelAction.java
+++ b/platforms/documentation/docs/src/snippets/integration/fetchModel/common/src/main/java/org/gradle/sample/FetchEclipseModelAction.java
@@ -1,0 +1,23 @@
+package org.gradle.sample;
+
+import org.gradle.tooling.BuildAction;
+import org.gradle.tooling.BuildController;
+import org.gradle.tooling.Failure;
+import org.gradle.tooling.FetchModelResult;
+import org.gradle.tooling.model.eclipse.EclipseProject;
+
+import java.util.Collection;
+
+// tag::fetch-action[]
+public class FetchEclipseModelAction implements BuildAction<EclipseProjectResult> {
+    @Override
+    public EclipseProjectResult execute(BuildController controller) {
+        FetchModelResult<EclipseProject> result = controller.fetch(EclipseProject.class);
+
+        EclipseProject model = result.getModel(); // may be null
+        Collection<? extends Failure> failures = result.getFailures();
+
+        return new EclipseProjectResult(model, failures);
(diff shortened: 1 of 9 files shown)


asodja:
@bot-gradle test BD

asodja:
@bot-gradle test BD

lkasso (approved):
I updated the docs. Please check for correctness and feel free to merge if everything looks good.

platforms/documentation/docs/src/snippets/integration/fetchModel/common/src/main/java/org/gradle/sample/FetchEclipseModelAction.java
```diff
@@ -0,0 +1,23 @@
...
+import org.gradle.tooling.BuildAction;
+import org.gradle.tooling.BuildController;
+import org.gradle.tooling.Failure;
+import org.gradle.tooling.FetchModelResult;
+import org.gradle.tooling.model.eclipse.EclipseProject;
+
+import java.util.Collection;
+
+// tag::fetch-action[]
+public class FetchEclipseModelAction implements BuildAction<EclipseProjectResult> {
```
eskatos, line 12: 💭 Wouldn't it be nice to show a build action that returns one base model if getting another more complete one fails to demonstrate resiliency more precisely?

eskatos: Extracted as a follow up in
* https://github.com/gradle/gradle/issues/38551

jbartok: Resilience was never about "having some complex model fail to build and if so asking for more basic ones, which hopefully don't fail to build". The complex model that fails is supposed to also contain as much partial data as can be obtained (besides the failures), so whatever we plan for as a fallback should come from that partial data, not further model requests. I think those cases where you do further model requests for more basic models are more like a special situation, not the typical way of using resilient models.

Or, phrasing it differently: yeah you can do a second model fetch, but you could do that before/without resiliency too.

PR #37941, opened 2026-05-20
[old] Fail-fast on IP violations by default

alllex:


octylFractal:
Replaced by https://github.com/gradle/gradle/pull/38011.

PR #37953, opened 2026-05-20
Replace weak-LoadingCache and CHM with ClassValue

octylFractal:
These are the easy migrations that can be done without much thought. There are more applicable locations that need more analysis.

The justification is that this is more efficient than a LoadingCache or CHM since it's generally able to load directly from the associated `Class` object, rather than storing a hashmap of them. It also doesn't need to rely on GC to empty out the weak entries -- because the values are stored on the `Class` itself, they are freed automatically.

In some cases we never even cleared the CHM, which could leak class values.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/BindingsBackedCodec.kt
+++ b/platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/BindingsBackedCodec.kt
@@ -26,7 +26,6 @@ import org.gradle.internal.serialize.graph.ReadContext
 import org.gradle.internal.serialize.graph.SerializerCodec
 import org.gradle.internal.serialize.graph.WriteContext
 import org.gradle.internal.serialize.graph.withDebugFrame
-import java.util.concurrent.ConcurrentHashMap
 import kotlin.reflect.KClass
 
 
@@ -52,7 +51,9 @@ class BindingsBackedCodec(private val bindings: List<Binding>) : Codec<Any?> {
     }
 
     private
-    val encodings = ConcurrentHashMap<Class<*>, TaggedEncoding>()
+    val encodings = object : ClassValue<TaggedEncoding>() {
+        override fun computeValue(type: Class<*>): TaggedEncoding = computeEncoding(type)
+    }
 
     override suspend fun WriteContext.encode(value: Any?) = when (value) {
         null -> writeSmallInt(NULL_VALUE)
@@ -111,7 +112,7 @@ class BindingsBackedCodec(private val bindings: List<Binding>) : Codec<Any?> {
 
     private
     fun taggedEncodingFor(type: Class<*>): TaggedEncoding =
-        encodings.computeIfAbsent(type, ::computeEncoding)
+        encodings.get(type)
 
     private
     fun computeEncoding(type: Class<*>): TaggedEncoding {
--- a/platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/BindingsBackedCodec.kt
+++ b/platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/BindingsBackedCodec.kt
@@ -26,6 +26,7 @@ import org.gradle.internal.serialize.graph.ReadContext
 import org.gradle.internal.serialize.graph.SerializerCodec
 import org.gradle.internal.serialize.graph.WriteContext
 import org.gradle.internal.serialize.graph.withDebugFrame
+import java.util.concurrent.ConcurrentHashMap
 import kotlin.reflect.KClass
 
 
@@ -51,9 +52,7 @@ class BindingsBackedCodec(private val bindings: List<Binding>) : Codec<Any?> {
     }
 
     private
-    val encodings = object : ClassValue<TaggedEncoding>() {
-        override fun computeValue(type: Class<*>): TaggedEncoding = computeEncoding(type)
-    }
+    val encodings = ConcurrentHashMap<Class<*>, TaggedEncoding>()
 
     override suspend fun WriteContext.encode(value: Any?) = when (value) {
         null -> writeSmallInt(NULL_VALUE)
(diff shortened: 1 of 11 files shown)


bamboo (approved):
💯

blindpirate:
This PR consistently trigger OOM issue:

- https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_Stage_PullRequestFeedback_Trigger/115172997
- https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_Stage_PullRequestFeedback_Trigger/115177759
- https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_Stage_PullRequestFeedback_Trigger/115177759

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/BindingsBackedCodec.kt
```diff
@@ -52,7 +51,9 @@ class BindingsBackedCodec(private val bindings: List<Binding>) : Codec<Any?> {
     }
 
     private
-    val encodings = ConcurrentHashMap<Class<*>, TaggedEncoding>()
+    val encodings = object : ClassValue<TaggedEncoding>() {
```
bamboo: ❌ This one is very suspicious because of the lifetime graph of the encodings. Probably better to split the PR to better understand the impact of each change.

bamboo (changes requested):

bamboo:
@bot-gradle test PRF

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/codecs/BindingsBackedCodec.kt
bamboo: Here's an analysis why ClassValue is not a suitable replacement for per-operation / short-lived maps due to the additional memory pressure:

**TLDR**
Even once the key is cleared, WeakHashMap doesn't drop the entry during GC. The Entry object and its strong .value stay in the table until expungeStaleEntries() runs — and that only runs on a subsequent operation on that specific Class's map (get/put/resize). GC alone never reclaims it, because there's still a strong path Class → … → value. So the stale entry for a given class is reclaimed only when that same class is serialized again by a later, still-live ClassValue after the old one has been GC-cleared.

**Full analysis**

```diff
- val encodings = ConcurrentHashMap<Class<*>, TaggedEncoding>()   // storage lives in the codec
+ val encodings = object : ClassValue<TaggedEncoding>() { … }      // storage lives on each Class
  …
- encodings.computeIfAbsent(type, ::computeEncoding)
+ encodings.get(type)
```

The answer is: the two designs put the backing map in different objects, so "the codec is discarded" frees one and not the other. Reference strength inside the map is a red herring — both hold their values strongly. What differs is where the map is rooted.

Trace it in each case

CHM (before). The map is a field of the codec. The ownership arrow points down from the codec:

codec ──(strong field 'encodings')──▶ CHM ──▶ { Class keys (strong), TaggedEncoding values (strong) }

When the operation ends and the codec becomes unreachable, the CHM becomes unreachable with it. The strong references the CHM held to its keys and values are strong references inside an unreachable object — they keep nothing alive. The Class keys survive (their ClassLoader still holds them), but the TaggedEncoding values die with the map. Retention lifetime = codec lifetime = one operation. Next operation gets a fresh codec + fresh CHM. No accumulation.

ClassValue (after). The values are not stored in the ClassValue object. The JDK stores each value on the key Class itself, in Class.classValueMap — a WeakHashMap whose key is the ClassValue (weak) and whose entry holds the value strongly. Now the ownership arrow for the value points down from the Class, and the codec only holds a weak-keyed handle into those maps:

Class (daemon-lived) ──(strong 'classValueMap')──▶ WeakHashMap ──▶ Entry ──(strong '.value')──▶ TaggedEncoding
        ▲                                                 │
        └── still GC-root-reachable via ClassLoader       └── key = weak ref to ──▶ ClassValue ◀── codec (weak-keyed)

When the operation ends and the codec (and its ClassValue) become unreachable, nothing about the value's rooting changes — the value is still held by Class → classValueMap → Entry → value, and the Class is still a live GC root. Discarding the codec only makes the ClassValue a weak key; it doesn't touch the storage, because the storage was never in the codec. Retention lifetime = key Class lifetime ≫ one operation.

**"But the ClassValue is the weak key — shouldn't it self-clean?"**

Eventually, yes — but weakly, lazily, and per-map, which is why it still accumulates:

1. GC must first clear the weak key (the dead ClassValue). Until then the entry isn't even stale.
2. Even once the key is cleared, WeakHashMap doesn't drop the entry during GC. The Entry object and its strong .value stay in the table until expungeStaleEntries() runs — and that only runs on a subsequent operation on that specific Class's map (get/put/resize). GC alone never reclaims it, because there's still a strong path Class → … → value.
3. Each Class has its own classValueMap. Touching class A's map expunges nothing on class C's map.

So the stale entry for a given class is reclaimed only when that same class is serialized again by a later, still-live ClassValue after the old one has been GC-cleared. For a class serialized in the initial sync but not re-touched by the incremental sync, that trigger never comes — its Entry.value (and the whole Build-scoped codec graph it pins) lives as long as the class does.

octylFractal: I think this one doesn't make sense because it is rebuilt so often. The rest are build tree or global scoped services. Global is obviously not problematic here, but build tree might build up a slow leak with each build? If it caches classes that are kept loaded in the daemon globally, maybe that's enough of a problem.

octylFractal: To be honest, this does feel like something the JDK should be accounting for, and maybe they need to have a reference queue or something to actively clear the `ClassValueMap` when a `ClassValue.Identity` is GC'd.

octylFractal:
@bot-gradle test PRF

bamboo (approved):
And away we go! 🧹

PR #37965, opened 2026-05-21
Fix gradlew.bat error handling not terminating script

runningcode:
NOTE: I do not have a windows machine to test this.
This was identified by Seer [here](https://github.com/EmergeTools/hackernews/pull/806#discussion_r3281074659) and from my analysis seems to be valid.

- `"%COMSPEC%" /c exit 1` in `gradlew.bat` spawns a child `cmd.exe` that exits, but the parent batch script continues running. This causes the script to fall through past the error message into `:execute`, attempting to run an undefined `JAVA_EXE` and producing cascading errors instead of a clean exit.
- Replace with `exit /b 1` which correctly terminates the current batch script with exit code 1.

### DIFF
--- a/gradlew.bat
+++ b/gradlew.bat
@@ -51,7 +51,7 @@ echo. 1>&2
 echo Please set the JAVA_HOME variable in your environment to match the 1>&2
 echo location of your Java installation. 1>&2
 
-"%COMSPEC%" /c exit 1
+exit /b 1
 
 :findJavaFromJavaHome
 set JAVA_HOME=%JAVA_HOME:"=%
@@ -65,7 +65,7 @@ echo. 1>&2
 echo Please set the JAVA_HOME variable in your environment to match the 1>&2
 echo location of your Java installation. 1>&2
 
-"%COMSPEC%" /c exit 1
+exit /b 1
 
 :execute
 @rem Setup the command line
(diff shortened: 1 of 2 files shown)


ov7a (dismissed):

:x: This is a script file generated by the application plugin. In order to have users benefit from this change, you need to update https://github.com/gradle/gradle/blob/master/platforms/jvm/plugins-application/src/main/resources/org/gradle/api/internal/plugins/windowsStartScript.txt

(I'm not sure about the change itself, though)

ov7a:
Thanks.

The relevant team for this area will confirm the implementation choices.

big-guy:
This might be valid. We'll take a look at this. We recently went through this script, but maybe we overlooked something.

octylFractal:
This is a valid issue, but going back to `exit /b` is not an appropriate fix. As mentioned in https://github.com/gradle/gradle/commit/ca80f3bef10590febb26669d651920b6428974d5, we explicitly changed away from this as the exit code doesn't propagate properly for usage with `||` and `&&`, so reverting that would break that case. I'm doing some testing to try and figure out what works.

octylFractal:
I'm going to open my own PR for the fix since it won't be reusing any of this code -- I also opened #38082 to track this.

PR #37967, opened 2026-05-21
Fix resilient sync behaviour with IP

asodja:
This PR fixes resilient sync issues with IP:
1. Fix deadlock issue with IP in some specific case we test with ResilientKotlinDslScriptsModelBuilderCrossVersionSpec (commit 2)
2. Fix issues with Kotlin DSL scripts model (commit 3)
3. Fix issues with GradleProject and IdeaProject in resilient mode with IP (commit 4 and 5)
4. Add additional resilient sync tests with IP (commit 1 and 6)


asodja:
@bot-gradle test this

asodja:
@bot-gradle help

asodja:
@bot-gradle test AllCrossVersionTests

asodja:
@bot-gradle test this ACT

asodja:
@bot-gradle test this ACT

asodja:
@bot-gradle test this ACT

asodja:
@bot-gradle test this ACT

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this ACT

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this ACT

asodja:
@bot-gradle test ACT

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
```diff
@@ -163,6 +165,16 @@ fun buildScriptModelsInHierarchy(
...
+    val classPathModeExceptionCollector = rootProject.serviceOf<ClassPathModeExceptionCollector>()
+
+    fun prepareForParallelAccess() {
+        // Avoid deadlock on stage1BlocksAccessorClassPath on root.fromMutableState call.
+        // It's wrapped in runCatching the same way the per-project and settings accessorsClassPathOf calls,
+        // so a broken root build degrades to empty accessors instead of aborting the model build.
+        classPathModeExceptionCollector.runCatching {
+            rootProject.serviceOf<Stage1BlocksAccessorClassPathGenerator>().prepareForParallelAccess()
+        }
+    }
```
6hundreds, line 182: 🤔 I wonder, when this deadlock happens? Is it only an issue in case of resilient sync?

asodja: It happens also when there is no failure, e.g. with `ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.no failure in multi-project build: resilient model is equal to non-resilient model`  with IP.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
```diff
@@ -194,8 +213,8 @@ fun buildScriptModelsInHierarchy(
 
 
 private
-fun IntermediateToolingModelProvider.getIsolatedModels(requester: ProjectState, targets: List<ProjectState>): List<IsolatedScriptsModel> =
-    getModels(requester, targets, IsolatedScriptsModel::class.java, null)
+fun IntermediateToolingModelProvider.getIsolatedModels(requester: ProjectState, targets: List<ProjectState>): List<IntermediateToolingModelResult<IsolatedScriptsModel>> =
+    getModelsAllowingFailures(requester, targets, IsolatedScriptsModel::class.java, null)
```
6hundreds: ❓ Should this "resiliency" be a part of the regular KotlinDSL IP model builder? Or resilient sync should has its own KDSL model builder, enabled by flag(or any other way how resilient sync enabled today).

From one perspective, I don't like the diverging tho, at least due to the IP team goal to have same KDSL model builder for VT and IP. But at the same time I don't feel this resiliency logic belongs to this class. Maybe this is because I don't now enough about resilient sync status in general - are we experimenting, or it's certainly a way to go?

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
6hundreds: So I assume our regular IP tests for KDSL must catch it?

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
asodja: Resilient sync is the way to go and we are not experimenting anymore. IntelliJ will use resilient sync in IntelliJ 2026.2 by default.

> Should this "resiliency" be a part of the regular KotlinDSL IP model builder? Or resilient sync should has its own KDSL model builder, enabled by flag(or any other way how resilient sync enabled today).

If we introduce a new builder then we will have three builders for Kotlin:
Vintage Kotlin DSL builder, IP Kotlin DSL builder, IP + Resilient Kotlin DSL builder  

I would like to avoid that. IP Kotlin DSL builder should have the same resilience level as Vintage Kotlin DSL builder. As agreed with IntelliJ Kotlin DSL builder should provide best effort classpath -> best effort here means that it's better to return some classpath than no classpath. This is needed otherwise we can't provide any autocompletion for broken projects.

Maybe an Important missing piece of information: any build configuration issue is caught before Kotlin DSL model builder is run, since we configure build/project before we run a builder. TAPI then returns configuration failure + Kotlin DSL model and IDE can show an error + provide code completion.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
asodja: If you have the same test as is:
```
ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.no failure in multi-project build: resilient model is equal to non-resilient model
```
 then it should catch it yes. Maybe the issue is caused because in our test we mix getModel and fetchModel methods (non-resilient mode and resilient mode)

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
6hundreds: >  IP Kotlin DSL builder should have the same resilience level as Vintage Kotlin DSL builder

100% agreed. My only, non-blocking, concern is that resiliency aspect is sort of implicit. Like an immediate question when look at the method "why do we allowing failures while getting models? Oh that because of resiliency contract". Maybe a small comment/javadoc would be helpful here

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
6hundreds: We do have `IsolatedProjectsToolingApiKotlinDslIntegrationTest` that does exactly the same - builds two models and compares them, but apparently because of the lack of multi-project scenario there we weren't facing the deadlock. I wonder if we should add such a test(not as a part of this PR ofk)

6hundreds: Well, maybe as a part of this PR 😄 , as we touching IP model building, we'd want to have some sort of confidence reflected in a test, that the change is correct

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
asodja: Good point, it's definitely unclear from reading the code. What do you think would be the best place to document that, is a comment above this method enough or do you think there is any better place?

6hundreds: `getIsolatedModels` method used in the only place, so I'd inline it, and then document `buildScriptModelsInHierarchy` method, since it's also got changed in regard of resiliency

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
asodja: Added reproducer to IsolatedProjectsToolingApiKotlinDslIntegrationTest

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
asodja: Applied suggested changes and added a comment

asodja:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiKotlinDslIntegrationTest.groovy
```diff
@@ -58,6 +62,34 @@ class IsolatedProjectsToolingApiKotlinDslIntegrationTest extends AbstractIsolate
         fixture.assertModelLoaded()
     }
 
+    def "resilient fetch of KotlinDslScripts model for multi-project build under IP does not deadlock"() {
```
6hundreds: ❓ But is it has to be resilient, to face a deadlock?

6hundreds: I mean if we'll fetch regular KotlinDslScriptsModel in IP in this build layout we have a chance for a deadlock? or we not?

asodja: I fixed reproducer and now I use non-resilient methods. I checked without resilient sync and if you remove `prepareForParallelAccess()` call it almost always deadlock even with non-resilient methods

asodja:
@bot-gradle test this

6hundreds (approved):
KotlinDSL model builder changes are LGTM!

asodja:
@bot-gradle test this ACT

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiKotlinDslIntegrationTest.groovy
6hundreds: Now I'm thinking, how didn't we face that deadlock in our codebase 🤔

asodja: As far as I understand the deadlock happens only for projects without root build script. If there is root build script `Stage1BlocksAccessorClassPathGenerator.stage1BlocksAccessorClassPath` is resolved when root build is queried and everything is ok. But if there is no root build script,  lock on a root project  + child projects query fan out create a deadlock

6hundreds: I see, thanks

asodja:
@bot-gradle test this

blindpirate:

PR #2578, opened 2017-07-24
Change test worker output based on community feedback

bmuschko:
Based on https://twitter.com/gradle/status/888424343885160448

PR #37970, opened 2026-05-22
Skip scanForTestClasses=false RunWith reproducer on JUnit 4.0

blindpirate:
Closes https://github.com/gradle/gradle-private/issues/5232

## Summary

- The `scanForTestClasses=false bypasses framework detection so include patterns alone determine which classes execute` reproducer in `JUnit4TestClassDetectionIntegrationTest` has been failing on master since #2578 against JUnit 4.0 only.
- Root cause: JUnit 4.0's `@RunWith` is not `@Inherited` (added in 4.1), so `SubTest.getAnnotation(RunWith.class)` returns null. JUnit falls back to the default class runner, finds no `@Test` methods, and fires `testStarted` with the suite description after `testRunStarted` already registered the same description as the root node. `JUnitTestEventAdapter` then trips the `No parent found for TestNode{name=example.SubTest,id=...}` assertion because the root node has no parent.
- Gate the reproducer on the existing `supportsEmptyClassWithRunner` predicate (`version >= 4.1`) via `Assume.assumeTrue`, matching the pattern used elsewhere in this file.


blindpirate:
Verified in https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_AllVersionsIntegMultiVersion_33_bucket2/112982884?hideTestsFromDependencies=true

big-guy:
Thanks Bo.  I had the exact same local change waiting for me to approve Claude to push :)

Issue #37956, opened 2026-05-20
Bring back the eager `isFoo()` accessor to reduce migration churn


abstratt:
Bring back the eager `isFoo()` accessor to reduce migration churn. Failing to do so means Groovy clients may silently try to evaluate `bar.foo` as a boolean expression.

Related: 
- https://github.com/gradle/gradle/issues/10615
- https://github.com/gradle/gradle/issues/37342

abstratt:
This is no longer needed. We went with #37984 instead

Issue #10615, opened 2018-02-13
Possibly surprising behavior with Provider API and booleans


big-guy:
Here's an example:
```
class BooleanTask extends DefaultTask {
    boolean shouldBe
}

class Naive extends BooleanTask {
    private final Property<Boolean> prop = project.objects.property(Boolean)

    Property<Boolean> isSurprising() {
        return prop
    }

    @TaskAction
    void print() {
        println "should be " + shouldBe
        // can't use 'surprising'
        println "!!surprising = " + !!isSurprising()
        println "surprising.get() = " + isSurprising().get()
    }
}

class Better extends BooleanTask {
    private final Property<Boolean> prop = project.objects.property(Boolean)

    Property<Boolean> getSurprising() {
        return prop
    }

    @TaskAction
    void print() {
        println "should be " + shouldBe
        println "!!surprising = " + !!surprising
        println "surprising.get() = " + surprising.get()
    }
}

task printNaive(type: Naive) {
    shouldBe = project.findProperty("foo") == "bar"
    // doesn't work
    // surprising = shouldBe
    isSurprising().set(shouldBe)
}

task printBetter(type: Better) {
    shouldBe = project.findProperty("foo") == "bar"
    // works now
    surprising = shouldBe
}

task print {
    dependsOn tasks.withType(BooleanTask)
}
```

Run `gradle print -Pfoo=bar`, everything should look OK.
Run again with `gradle print -Pfoo=notbar`, you'll now see something is wrong:
```
> Task :printBetter 
should be false
!!surprising = true
surprising.get() = false

> Task :printNaive 
should be false
!!surprising = true
surprising.get() = false
```

For the naive task, we can imagine someone took their existing `setSurprising`/`isSurprising` property and turned it into a `Property<Boolean>`.  This doesn't really work because I can no longer use the setter syntax (`isSurprising` is not a JavaBean).  Due to Groovy truthiness, this has another problem.... `if (isSurprising())` doesn't do what you think.  It checks that the `Property` is non-null (which it always is).

For the better task, the plugin author has figured out that `getSurprising` is the magic way to define a boolean property and get our setter magic to work.  But this has the same sort of problem as above, `if (getSurprising())` or `if (surprising)` doesn't do what you think. 

As a fix, we could implement `asBoolean` in `AbstractProvider` and try to mimic Groovy truthiness, but we would need to always be vigilant for any new `Provider` implementation to make sure we don't lose this coercion. 

[...]

PR #37984, opened 2026-05-24
Implement groovy-truthness for Provider

abstratt:
This makes a `Provider`/`Property` behave sensibly when used in a boolean context (e.g. `if (prop)`). A provider with no value is now falsy; otherwise it follows the Groovy truth of its value. Until now, any non-null provider was simply "true", which silently misled Groovy users who expected the value to be checked — the problem raised in #10615 / #37956. Since reading a provider's value in a boolean test is something we want to steer people away from, it also logs a deprecation pointing to the explicit alternative, with removal planned for Gradle 10. This works even from `@CompileStatic` code, not just dynamic Groovy.

#### Test churn

Most of the diff is Spock cleanup. In `expect:` blocks, every line is treated as a condition, so with the new behavior, bare providers and side-effecting calls would get coerced to boolean — causing eager evaluation and the deprecation warning. Those lines were moved into `when:`/`then:` so the tests check what they mean to check, instead of accidentally triggering the new behavior.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/ProviderExtensions.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/ProviderExtensions.java
@@ -64,7 +64,7 @@ public class ProviderExtensions {
         String capitalizedType = StringUtils.capitalize(type);
         DeprecationLogger.deprecateBehaviour(format("Using a `%s` where a `boolean` is expected should be avoided.", capitalizedType))
             .withAdvice(format("If evaluating this %s is intentional, do so explicitly via `get()` or `getOrNull()`.", type))
-            .startingWithGradle10("this is not recommended")
+            .willBecomeAnErrorInGradle11()
             .undocumented()
             .nagUser();
     }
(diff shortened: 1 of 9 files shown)


abstratt:
@bot-gradle test this

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/ProviderExtensions.java
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * <p>Extension method to support using a provider in a boolean context in Groovy.</p>
+     *
+     * @param self the {@link Provider}
+     * @return the Groovy truth of the provider's value
+     * @see <a href="https://groovy-lang.org/semantics.html#the-groovy-truth">the Groovy truth</a>
+     */
+    public static boolean asBoolean(Provider<?> self) {
+        DeprecationLogger.deprecateBehaviour("Using a `Provider` where a `boolean` is expected causes the provider to be evaluated.")
+            .withAdvice("The proper way to explicitly coerce a `Provider` into a `boolean` value is via `getOrNull()`.")
```
asodja: 💅 I would say that user should try to avoid using Provider where boolean is expected and only if it's intentional call getOrNull() as i suggests

asodja (approved):

abstratt:
@bot-gradle test RFR

abstratt:
@bot-gradle test RFN

PR #37991, opened 2026-05-25
Deprecate ExecSpec.setCommandLine()

### LINKED ISSUE #33015: Deprecate `ExecSpec.setCommandLine()` methods
Since in Gradle 10.0 commandLine is read only `Provider<List<String>>`.

Users can still use `commandLine()` method.

asodja:
Fixes https://github.com/gradle/gradle/issues/33015

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -1715,9 +1715,9 @@ tasks.named('myRunTask', JavaExec) {
 [[set-command-line]]
 ==== Deprecation of `ExecSpec.setCommandLine()`
 
-The link:{javadocPath}/org/gradle/process/ExecSpec.html#setCommandLine(java.util.List)[`setCommandLine()` method] on `ExecSpec` and, by inheritance, on `JavaExecSpec` and the `Exec`/`JavaExec` tasks has been deprecated and will be removed in Gradle 10.0.0.
+The link:{javadocPath}/org/gradle/process/ExecSpec.html#setCommandLine(java.util.List)[`setCommandLine()` method] on `ExecSpec` and, by inheritance, on `JavaExecSpec` and the `Exec`/`JavaExec` tasks has been deprecated and will be removed in Gradle 10.
 
-Use link:{javadocPath}/org/gradle/process/ExecSpec.html#commandLine(java.lang.Object...)[`commandLine()`] instead.
+Use link:{javadocPath}/org/gradle/process/ExecSpec.html#commandLine(java.lang.Object...)[`commandLine(...)`] instead.
 
 In Groovy build scripts, the equivalent property assignment syntax `commandLine = [...]` is also deprecated; replace it with a `commandLine([...])` call.
 
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -1712,6 +1712,38 @@ tasks.named('myRunTask', JavaExec) {
 =====
 ====
 
+[[set-command-line]]
+==== Deprecation of `ExecSpec.setCommandLine()`
+
+The link:{javadocPath}/org/gradle/process/ExecSpec.html#setCommandLine(java.util.List)[`setCommandLine()` method] on `ExecSpec` and, by inheritance, on `JavaExecSpec` and the `Exec`/`JavaExec` tasks has been deprecated and will be removed in Gradle 10.0.0.
+
+Use link:{javadocPath}/org/gradle/process/ExecSpec.html#commandLine(java.lang.Object...)[`commandLine()`] instead.
+
+In Groovy build scripts, the equivalent property assignment syntax `commandLine = [...]` is also deprecated; replace it with a `commandLine([...])` call.
+
+====
+[.multi-language-sample]
+=====
+.build.gradle.kts
+[source, kotlin]
+----
+tasks.register<Exec>("myExec") {
+    commandLine("git", "status")
+}
+----
+=====
+[.multi-language-sample]
+=====
+.build.gradle
+[source, groovy]
+----
+tasks.register('myExec', Exec) {
+    commandLine('git', 'status')
+}
+----
+=====
+====
+
 [[archives-configuration]]
 ==== Deprecation of `archives` configuration
 
--- a/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/SetCommandLineDeprecationIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/SetCommandLineDeprecationIntegrationTest.groovy
@@ -22,7 +22,7 @@ import org.gradle.process.ShellScript
 class SetCommandLineDeprecationIntegrationTest extends AbstractIntegrationSpec {
 
     private static final String SET_COMMAND_LINE_DEPRECATION = "The ExecSpec.setCommandLine() method has been deprecated. " +
-        "This is scheduled to be removed in Gradle 10. Use commandLine() instead. " +
+        "This is scheduled to be removed in Gradle 10. Use commandLine(...) instead. " +
         "Consult the upgrading guide for further information: " +
         "https://docs.gradle.org/current/userguide/upgrading_version_9.html#set-command-line"
 
--- a/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/SetCommandLineDeprecationIntegrationTest.groovy
+++ b/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/SetCommandLineDeprecationIntegrationTest.groovy
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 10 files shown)


asodja:
@bot-gradle test this

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -1712,6 +1712,38 @@ tasks.named('myRunTask', JavaExec) {
...
 ====
 
+[[set-command-line]]
+==== Deprecation of `ExecSpec.setCommandLine()`
+
+The link:{javadocPath}/org/gradle/process/ExecSpec.html#setCommandLine(java.util.List)[`setCommandLine()` method] on `ExecSpec` and, by inheritance, on `JavaExecSpec` and the `Exec`/`JavaExec` tasks has been deprecated and will be removed in Gradle 10.0.0.
+
+Use link:{javadocPath}/org/gradle/process/ExecSpec.html#commandLine(java.lang.Object...)[`commandLine()`] instead.
+
+In Groovy build scripts, the equivalent property assignment syntax `commandLine = [...]` is also deprecated; replace it with a `commandLine([...])` call.
```
abstratt, line 1722: Aren't we bringing the setter back? (CC: @hegyibalint). Are eager setters deprecated?

asodja: Yes, but these setters will be deprecated to favour `property.set()`. But commandLine actually becomes `Provider<List<String>>`, since it's concatenated via many different properties. So we deprecate a setter already now, since for `Kotlin` `commandLine = ` will stop working, even if we keep a setter

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -1712,6 +1712,38 @@ tasks.named('myRunTask', JavaExec) {
 =====
 ====
 
+[[set-command-line]]
+==== Deprecation of `ExecSpec.setCommandLine()`
+
+The link:{javadocPath}/org/gradle/process/ExecSpec.html#setCommandLine(java.util.List)[`setCommandLine()` method] on `ExecSpec` and, by inheritance, on `JavaExecSpec` and the `Exec`/`JavaExec` tasks has been deprecated and will be removed in Gradle 10.0.0.
```
```suggestion
The link:{javadocPath}/org/gradle/process/ExecSpec.html#setCommandLine(java.util.List)[`setCommandLine()` method] on `ExecSpec` and, by inheritance, on `JavaExecSpec` and the `Exec`/`JavaExec` tasks has been deprecated and will be removed in Gradle 10.
```

lkasso (approved):
Docs approved.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -1712,6 +1712,38 @@ tasks.named('myRunTask', JavaExec) {
...
+
+Use link:{javadocPath}/org/gradle/process/ExecSpec.html#commandLine(java.lang.Object...)[`commandLine()`] instead.
```
hegyibalint: 💄 I think instead of `commandLine()` we should use `commandLine(...)` to just simply denote that there are parameters to the method. Even if straightforward that we are using it to set things. Just an opinion.

hegyibalint (approved):
LGTM, one nitpick 👍

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
asodja: Fixed

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
asodja: Fixed

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

big-guy (changes requested):
I don't think this is the right direction for a couple of reasons.

1. We're trying to move away from "space assignment" APIs: https://docs.gradle.org/current/userguide/upgrading_version_8.html#groovy_space_assignment_syntax -- This tells users to migrate specifically to one of those.  IOW, we want people to write `commandLine = ...` if `commandLine` is mutable.
2. We're getting rid of some of the ways to set the command-line to prepare for replacing the getter with a read-only Provider-based one. This removes some of the API, but it doesn't resolve the conceptual clash.

IOW, what does this property represent? `commandLine` is like an aggregate property on `ExecSpec`. 

It combines:
- `executable`
- `args`
- `argumentProviders`

On `JavaExecSpec`, you have even more things to combine (module, JVM args vs application args, heap sizes, modularity, system properties, etc).

A few options:
1. Deprecate all command-line setters except one (`commandLine(...)`) and prepare to remove them in 10. This is what the PR proposes.
2. Deprecate all setters and make the property entirely read-only in the future (hopefully 10) by removing all setters and making it a `Provider<List<String>>`.
3. Leave the setters and make the property read-write in the future (a `ListProperty<String>`).
4. Deprecate getters and setters for commandLine and make it an internal/unreachable property. 

(2) leans into this being an aggregate property with a fixed convention and explicitly makes changing it invalid.
(3) leans into this being an aggregate property with a changing convention.

(1) is the odd one that looks like (2) works like (3) and makes things inconsistent with other Provider-based properties.

In a all Provider world, we'd be able to do something like:
```
commandLine.convention(executable.zip(args) { executable, args -> 
    [executable] + args
})
```

If someone set `commandLine`, it would break the connection with executable and args, but there's already today an existing weirdness when someone changes the value of `commandLine`. If something else changes the value of `executable` later, the `commandLine` still wins. I think this would only be fixed if we got rid of the property or made it read-only. 

One bad thing about the current situation is that I think a writable `commandLine` makes a lot of sense for `exec {}` blocks and a read-only `commandLine` makes a lot of sense for `Exec` tasks. IOW, I think the convenience of setting `commandLine` for `exec {}` is useful and the structure/meaning behind `commandLine` for `Exec` is useful. We would probably like to make `executable` a regular file for tasks too. 

For consistency, I would make this a `ListProperty` with a convention (option 3). Is there something I'm missing here?

asodja:
>If someone set commandLine, it would break the connection with executable and args, but there's already today an existing weirdness when someone changes the value of commandLine. If something else changes the value of executable later, the commandLine still wins.

As far as I understand current code, commandLine setters are wired through: https://github.com/gradle/gradle/blob/ec3f90ce4c481788dbc7689cd882f33263a67dc6/platforms/core-runtime/process-services-base/src/main/java/org/gradle/process/internal/ProcessArgumentsSpec.java#L59-L64

When `setCommandLine(...)` is called we actually set the `executable` and `args`. So there is no `commandLine` field, but it's already a combination of multiple of them. If you then change `executable` later it's reflected also in `commandLine` value. 

Read-only `Provider<List<String>> commandLine` follows the same logic, except that you cannot use `=` (you can still use `setCommandLine()` and `commandLine()` as deprecated options).  `ListProperty<String> commandLine` would change this, but it would allow `=`.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -1712,6 +1712,38 @@ tasks.named('myRunTask', JavaExec) {
 =====
 ====
 
+[[set-command-line]]
+==== Deprecation of `ExecSpec.setCommandLine()`
+
+The link:{javadocPath}/org/gradle/process/ExecSpec.html#setCommandLine(java.util.List)[`setCommandLine()` method] on `ExecSpec` and, by inheritance, on `JavaExecSpec` and the `Exec`/`JavaExec` tasks has been deprecated and will be removed in Gradle 10.
```
abstratt, line 1718: ❌ 

> by inheritance, on `JavaExecSpec` and the `Exec`/`JavaExec` 

These don't inherit from `ExecSpec`though, right?

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/SetCommandLineDeprecationIntegrationTest.groovy
```diff
@@ -0,0 +1,72 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.api.tasks
+
+import org.gradle.integtests.fixtures.AbstractIntegrationSpec
+import org.gradle.process.ShellScript
+
+class SetCommandLineDeprecationIntegrationTest extends AbstractIntegrationSpec {
```
abstratt, line 22: ❓ 

Should we also cover `project.exec {}`/`ExecOperations.exec {}` and `providers.exec {}`? The nag was added in `DefaultExecAction` and `DelegatingExecSpec` too, but only the `Exec` task path is asserted here.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -1712,6 +1712,38 @@ tasks.named('myRunTask', JavaExec) {
 =====
 ====
 
+[[set-command-line]]
```
abstratt, line 1715: 💭 

This needs to be moved to appear under the proper section

abstratt (changes requested):
It seems the documentation is incorrect, plus a couple of questions/suggestions

PR #37997, opened 2026-05-26
Aggregate performance report when some buckets fail or are cancelled

cobexer:
- `PerformanceTestsPass` Trigger used `%dep.<bucket>.env.BUILD_ID%` and `%dep.<bucket>.performance.baselines%` references; a single cancelled bucket left them unresolved and the aggregate report never ran.
- Source baselines from the Trigger's own `performance.baselines` parameter and derive dependency TeamCity build IDs from the `teamCityBuildId` field of each surviving `perf-results-*.json` inside `:performance:performanceTestReport`.
- Override `onDependencyCancel` to `FailureAction.ADD_PROBLEM` so a cancelled bucket still lets the Trigger run, the report still publishes, and the Trigger ends red.
- Drop the now-unused `org.gradle.performance.dependencyBuildIds` Gradle property plumbing.

### DIFF
--- a/.teamcity/src/main/kotlin/configurations/PerformanceTestsPass.kt
+++ b/.teamcity/src/main/kotlin/configurations/PerformanceTestsPass.kt
@@ -19,6 +19,7 @@ package configurations
 import common.Os
 import common.applyDefaultSettings
 import common.setArtifactRules
+import jetbrains.buildServer.configs.kotlin.FailureAction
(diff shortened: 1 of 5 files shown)


cobexer:
@bot-gradle test APT

blindpirate (approved):
LGTM. Was about to test in https://builds.gradle.org/buildConfiguration/Gradle_Xperimental_Check_PerformanceTestTestLinux_Trigger but saw you already tested it.

PR #37999, opened 2026-05-26
Cross project configuration reporting `ExtensionContainer` for `Gradle`

### LINKED ISSUE #37988: `Gradle.getExtensions` exposes mutable state  to the project scope
`ExtensionsContainer` is a mutable state that may be exposed to projects.

Real use-case: DV plugin uses `Gradle.getExtensions.getByType()` in `ProvisionDevelocityAccessKeyTask`

6hundreds:
Fixes #37988 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl.isolated
+
+class IsolatedProjectsCrossProjectGradleAccessIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
+
+    def "reports a problem on project-level mutation of Gradle.extensions via #invocation"() {
+        settingsFile """
+            include("a")
+        """
+        file("a/build.gradle") << """
+            import org.gradle.api.reflect.TypeOf;
+
+            interface Foo {}
+
+            class DefaultFoo implements Foo {}
+
+            gradle.extensions.$invocation
+        """
+
+        when:
+        isolatedProjectsFails ":a:help"
+
+        then:
+        fixture.assertStateStoredAndDiscarded {
+            projectsConfigured(":", ":a")
+            problem("Build file 'a/build.gradle': line 8: Project ':a' cannot $problemDescription on Gradle extension container")
+        }
+
+        where:
+        invocation                                              | problemDescription
+        "add('foo', new DefaultFoo())"                          | "add extension `foo` with public type `DefaultFoo`"
+        "add(Foo, 'foo', new DefaultFoo())"                     | "add extension `foo` with public type `Foo`"
+        "add(TypeOf.typeOf(Foo), 'foo', new DefaultFoo())"      | "add extension `foo` with public type `Foo`"
+
+        "create('foo', DefaultFoo)"                             | "create extension `foo` with public type `DefaultFoo`"
+        "create(Foo, 'foo', DefaultFoo)"                        | "create extension `foo` with public type `Foo`"
+        "create(TypeOf.typeOf(Foo), 'foo', DefaultFoo)"         | "create extension `foo` with public type `Foo`"
+
+        // use ExtraPropertiesExtension as the only available by default
+        "configure(ExtraPropertiesExtension) {}"                | "configure extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
+        "configure(TypeOf.typeOf(ExtraPropertiesExtension)) {}" | "configure extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
+        "configure('ext') {}"                                   | "configure extension of name `ext`"
+
+        "extraProperties.set('foo', 'bar')"                     | "set extra properties extension `foo`"
+    }
+}
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
@@ -57,5 +57,10 @@ class IsolatedProjectsCrossProjectGradleAccessIntegrationTest extends AbstractIs
         "configure('ext') {}"                                   | "configure extension of name `ext`"
 
         "extraProperties.set('foo', 'bar')"                     | "set extra properties extension `foo`"
+
+        // Groovy dynamic property setter
+        "foo = new DefaultFoo()"                                | "add extension `foo` with public type `DefaultFoo`"
+        // Groovy dynamic property setter on extra properties
+        "extraProperties.foo = 'bar'"                           | "set extra properties extension `foo`"
     }
 }
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
@@ -52,15 +52,22 @@ class IsolatedProjectsCrossProjectGradleAccessIntegrationTest extends AbstractIs
         "create(TypeOf.typeOf(Foo), 'foo', DefaultFoo)"         | "create extension `foo` with public type `Foo`"
 
         // use ExtraPropertiesExtension as the only available by default
+        "getByType(ExtraPropertiesExtension)"                   | "get extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
+        "getByType(TypeOf.typeOf(ExtraPropertiesExtension))"    | "get extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
+        "findByType(ExtraPropertiesExtension)"                  | "find extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
+        "findByType(TypeOf.typeOf(ExtraPropertiesExtension))"   | "find extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
+
+        "getByName('ext')"                                      | "get extension of name `ext`"
+        "findByName('ext')"                                     | "find extension of name `ext`"
+
         "configure(ExtraPropertiesExtension) {}"                | "configure extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
         "configure(TypeOf.typeOf(ExtraPropertiesExtension)) {}" | "configure extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
         "configure('ext') {}"                                   | "configure extension of name `ext`"
 
-        "extraProperties.set('foo', 'bar')"                     | "set extra properties extension `foo`"
+        "extraProperties"                                       | "access extra properties"
 
-        // Groovy dynamic property setter
+        // Groovy dynamic access
+        "ext"                                                   | "get extension of name `ext`"
         "foo = new DefaultFoo()"                                | "add extension `foo` with public type `DefaultFoo`"
-        // Groovy dynamic property setter on extra properties
-        "extraProperties.foo = 'bar'"                           | "set extra properties extension `foo`"
     }
 }
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
@@ -18,7 +18,7 @@ package org.gradle.internal.cc.impl.isolated
 
 class IsolatedProjectsCrossProjectGradleAccessIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
 
-    def "reports a problem on project-level mutation of Gradle.extensions via #invocation"() {
+    def "reports a problem on project-level access of Gradle.extensions via #invocation"() {
         settingsFile """
             include("a")
         """
@@ -38,36 +38,37 @@ class IsolatedProjectsCrossProjectGradleAccessIntegrationTest extends AbstractIs
         then:
         fixture.assertStateStoredAndDiscarded {
             projectsConfigured(":", ":a")
-            problem("Build file 'a/build.gradle': line 8: Project ':a' cannot $problemDescription on Gradle extension container")
+            problem("Build file 'a/build.gradle': line 8: Project ':a' cannot access Gradle.extensions")
         }
 
         where:
-        invocation                                              | problemDescription
-        "add('foo', new DefaultFoo())"                          | "add extension `foo` with public type `DefaultFoo`"
-        "add(Foo, 'foo', new DefaultFoo())"                     | "add extension `foo` with public type `Foo`"
-        "add(TypeOf.typeOf(Foo), 'foo', new DefaultFoo())"      | "add extension `foo` with public type `Foo`"
+        invocation << [
+            "add('foo', new DefaultFoo())",
+            "add(Foo, 'foo', new DefaultFoo())",
+            "add(TypeOf.typeOf(Foo), 'foo', new DefaultFoo())",
 
-        "create('foo', DefaultFoo)"                             | "create extension `foo` with public type `DefaultFoo`"
-        "create(Foo, 'foo', DefaultFoo)"                        | "create extension `foo` with public type `Foo`"
-        "create(TypeOf.typeOf(Foo), 'foo', DefaultFoo)"         | "create extension `foo` with public type `Foo`"
+            "create('foo', DefaultFoo)",
+            "create(Foo, 'foo', DefaultFoo)",
+            "create(TypeOf.typeOf(Foo), 'foo', DefaultFoo)",
 
-        // use ExtraPropertiesExtension as the only available by default
-        "getByType(ExtraPropertiesExtension)"                   | "get extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
-        "getByType(TypeOf.typeOf(ExtraPropertiesExtension))"    | "get extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
-        "findByType(ExtraPropertiesExtension)"                  | "find extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
-        "findByType(TypeOf.typeOf(ExtraPropertiesExtension))"   | "find extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
+            // use ExtraPropertiesExtension as the only available by default
+            "getByType(ExtraPropertiesExtension)",
+            "getByType(TypeOf.typeOf(ExtraPropertiesExtension))",
+            "findByType(ExtraPropertiesExtension)",
+            "findByType(TypeOf.typeOf(ExtraPropertiesExtension))",
 
-        "getByName('ext')"                                      | "get extension of name `ext`"
-        "findByName('ext')"                                     | "find extension of name `ext`"
+            "getByName('ext')",
+            "findByName('ext')",
 
-        "configure(ExtraPropertiesExtension) {}"                | "configure extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
-        "configure(TypeOf.typeOf(ExtraPropertiesExtension)) {}" | "configure extension of type `org.gradle.api.plugins.ExtraPropertiesExtension`"
-        "configure('ext') {}"                                   | "configure extension of name `ext`"
+            "configure(ExtraPropertiesExtension) {}",
+            "configure(TypeOf.typeOf(ExtraPropertiesExtension)) {}",
+            "configure('ext') {}",
 
-        "extraProperties"                                       | "access extra properties"
+            "extraProperties",
 
-        // Groovy dynamic access
-        "ext"                                                   | "get extension of name `ext`"
-        "foo = new DefaultFoo()"                                | "add extension `foo` with public type `DefaultFoo`"
+            // Groovy dynamic access
+            "ext",
+            "foo = new DefaultFoo()",
+        ]
     }
 }
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingExtensionsContainer.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingExtensionsContainer.kt
@@ -17,7 +17,6 @@
 package org.gradle.internal.cc.impl
 
 import com.google.common.collect.ImmutableMap
-import groovy.lang.ReadOnlyPropertyException
 import org.gradle.api.Action
 import org.gradle.api.internal.plugins.ExtensionContainerInternal
 import org.gradle.api.internal.project.ProjectIdentity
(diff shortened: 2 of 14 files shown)


6hundreds:
@bot-gradle test prf

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingExtensionsContainer.kt
```diff
@@ -0,0 +1,129 @@
...
+        return delegate.create(publicType, name, instanceType, *constructionArguments)
+    }
+
+    override fun <T : Any> create(name: String, type: Class<T>, vararg constructionArguments: Any): T {
+        onMutableStateAccess("create extension `$name` with public type `${type.simpleName}`")
+        return delegate.create(name, type, *constructionArguments)
+    }
+
+    override fun <T : Any> configure(type: Class<T>, action: Action<in T>) {
+        onMutableStateAccess("configure extension of type `${type.name}`")
```
6hundreds: ℹ️ Use `Class.name`(fqn) for disambiguity when no extension name passed. Use `Class.simpleName` for brevity when extension name passed

6hundreds:
@bot-gradle test prf

6hundreds:
@bot-gradle test prf

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingExtensionsContainer.kt
```diff
@@ -0,0 +1,156 @@
...
+    fun propertyMissing(name: String, value: Any) {
+        require(findByName(name) == null) {
+            "There's an extension registered with name '$name'. You should not reassign it via a property setter."
+        }
+        add(name, value)
+    }
+    // endregion Groovy support
+}
+
+private class CrossProjectConfigurationReportingExtraPropertiesExtension(
```
alllex: ❌  Let's not wrap deeper nested state unless we have evidence prevents us from introducing more coarse-grained violation. So, I'd start by making the `getExtraProperties()` call itself to be a violation.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingExtensionsContainer.kt
```diff
@@ -0,0 +1,156 @@
...
+import org.gradle.api.internal.plugins.ExtensionContainerInternal
+import org.gradle.api.internal.project.ProjectIdentity
+import org.gradle.api.plugins.ExtensionsSchema
+import org.gradle.api.plugins.ExtraPropertiesExtension
+import org.gradle.api.reflect.TypeOf
+import org.gradle.internal.configuration.problems.IsolatedProjectsProblemsReporter
+import org.gradle.internal.extensibility.DefaultExtensionsSchema
+import org.gradle.internal.extensions.stdlib.capitalized
+
+internal class CrossProjectConfigurationReportingExtensionsContainer(
```
alllex: 💅  How about `IsolatedProjectsGradleExtensionContainer` ?

This makes it easy to find, when looking for IP stuff, and the `Gradle` part clearly achors the purpose of the class to wrap specifically the extensions of the `Gradle` instance, and not an arbitrary extension container (we don't yet have a need for that)

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingExtensionsContainer.kt
```diff
@@ -0,0 +1,156 @@
...
+    private val delegate: ExtensionContainerInternal,
+    private val referrer: ProjectIdentity,
+    private val ipProblems: IsolatedProjectsProblemsReporter
+) : ExtensionContainerInternal by delegate {
```
alllex: ❌  I think we should avoid `by delegate` for such classes to ensure we are explicit about all methods: whether they are mutable state or not. If we add anew new methods to the super interfaces in the future, we want to get a compilation error and be prompted to make a decision for the new method.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -240,7 +240,7 @@ class CrossProjectConfigurationReportingGradle(
         delegate.pluginManager
 
     override fun getExtensions(): ExtensionContainer =
-        delegate.extensions
+        crossProjectModelAccess.getExtensionContainerForProject(referrerProject, delegate.extensions)
```
alllex: 🤔  I think the additional layer of abstraction makes this code harder to reason about. The way I see it, we have a clear need to wrap the extension container of the Gradle instance and we are already in the wrapped Gradle instance. So let's call the constructor of the wrapper directly here.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradleExtensionsContainer.kt
alllex, line 1: ❌  This class currently doesn't guard against the access to the read-like method, e.g. getByType and so on. We should change that.

Our ultimate goal is to make `gradle.getExtensions()` a very coarse-grained violation. We don't want to do it directly today, because it breaks the DV integration. So should ship the implementation that still captures our intention, but carves out just the minimal piece that DV needs and that we know is "safe".

So we should make all of the methods on the wrapper violations. Then, specifically for that one method that DV plugin calls, we issue the violation conditionally -- if any other extension type is requested except the `StateAccess` one. We have the `Workarounds` classes for this.

alllex (changes requested):

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingExtensionsContainer.kt
6hundreds: `CrossProject....` naming is literally the pattern in the codebase:
[CrossProjectModelAccessProjectEvaluationListener](https://github.com/gradle/gradle/blob/57d2fbbac2b2e9efe0a02a8b593d34086f6accdf/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt#L203
)
[CrossProjectModelAccessTrackingClosure](https://github.com/gradle/gradle/blob/57d2fbbac2b2e9efe0a02a8b593d34086f6accdf/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectModelAccessTrackingClosure.kt#L28)
[CrossProjectConfigurationReportingTaskExecutionGraph](https://github.com/gradle/gradle/blob/c1ba5318f322305c6296617eae966b4e5fb5b589/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingTaskExecutionGraph.kt#L41)
[CrossProjectConfigurationReportingGradle](https://github.com/gradle/gradle/blob/57d2fbbac2b2e9efe0a02a8b593d34086f6accdf/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt#L60)

I don't think `IsolatedProjects` prefix will make it easier to find, because now, when you grep for that prefix you get a ton of the tests and test fixtures, not production code.

At the same time I agree with the point `and the Gradle part clearly achors the purpose of the class to wrap specifically the extensions of the Gradle instance, and not an arbitrary extension container`. 

Although, given that we decided to workaround DV issue, an report `getExtensions` unconditionally, we don't need a wrapper at all and having an agreement on the name of the class doesn't make much sense

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingExtensionsContainer.kt
6hundreds: Makes sense

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingExtensionsContainer.kt
alllex: But we do need a wrapper for the `ExtensionContainer` though. Otherwise, how can you avoid issueing a violation, when `StateAccess` is requested via `getByType()`? I might be missing something

What we don't need so far is the wrapper for the extra properties container

6hundreds: > But we do need a wrapper for the ExtensionContainer though.

Hm, yes, I thought we had some additional information, like the requesting task name or so, as we have for other Workarounds

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
6hundreds: The idea was not to add additional dependency(`ipProblems`) to the enclosing class, but given that it will be added anyways later, I agree to inline the instantiation

6hundreds:
@bot-gradle test this

6hundreds:
@alllex I've added a workaround for the DV use case. There are still some CI failures related to DV tests, this PR should be on the top of https://github.com/gradle/gradle/pull/38051 I believe. 
Also there is a test `GradleEnterprisePluginBackgroundJobExecutorsIntegrationTest` which setup is not IP compatible with this PR, I wonder should we disable it for IP, altogether with all other tests that doing putting/getting `serviceRef` to Gradle extensions.

subprojects/core/src/main/java/org/gradle/api/internal/project/CrossProjectModelAccess.java
```diff
@@ -136,4 +137,6 @@ public interface CrossProjectModelAccess {
      */
     @Nullable
     HierarchicalDynamicObject parentProjectDynamicInheritedScope(ProjectState referrer);
+
+    ExtensionContainer getExtensionContainerForProject(ProjectIdentity referrer, ExtensionContainer extensionContainer);
```
alllex: ❓  We no longer need this method, right?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradleExtensionsContainer.kt
```diff
@@ -0,0 +1,144 @@
...
+        onMutableStateAccess("access extra properties")
+        return delegate.extraProperties
+    }
+
+    override fun <T : Any> add(publicType: Class<T>, name: String, extension: T) {
+        add(TypeOf.typeOf(publicType), name, extension)
+    }
+
+    override fun <T : Any> add(publicType: TypeOf<T>, name: String, extension: T) {
+        onMutableStateAccess("add extension `$name` with public type `${publicType.simpleName}`")
```
alllex: 💭  I think we can make all violation messages the same, saying "you can't access Gradle.extensions". That's our intention anyway. It's only due to the workaround that we have to delay the violation until somebody calls practically any method on the container.

alllex:
> There are still some CI failures related to DV tests, this PR should be on the top of https://github.com/gradle/gradle/pull/38051 I believe.

yes

> Also there is a test GradleEnterprisePluginBackgroundJobExecutorsIntegrationTest which setup is not IP compatible with this PR, I wonder should we disable it for IP, altogether with all other tests that doing putting/getting serviceRef to Gradle extensions.

I'd rather not disable tests if we can avoid it. Can you check if we can update the fixture to get/put this value via a shared build service instead?

subprojects/core/src/main/java/org/gradle/api/internal/project/CrossProjectModelAccess.java
6hundreds: We don't, thanks

6hundreds:
@bot-gradle test prf

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradleExtensionsContainer.kt
6hundreds: Makes sense, such thorough reporting may give a feeling that some methods are allowed

6hundreds:
@alllex See the commit fixed tests
https://github.com/gradle/gradle/pull/37999/changes/63afe24be5eac92c8f293ea429a70b62fc712d43

platforms/enterprise/enterprise/src/integTest/groovy/org/gradle/internal/enterprise/DevelocityPluginConfigurationCachingIntegrationTest.groovy
```diff
@@ -212,7 +212,7 @@ class DevelocityPluginConfigurationCachingIntegrationTest extends AbstractIntegr
 
     private String taskPrintBuildInvocationId(String taskName) {
         """
-            def serviceRef = gradle.extensions.serviceRef
+            def serviceRef = services.get(${GradleEnterprisePluginServiceRef.name})
```
alllex: ❔  Since we are touching this, and pretty much use an internal API to get a special service, can we have a dedicated method / fixture for this, so that next time it's easy to locate all such places?

It might not be worth it, but I'd like us to explicitly consider that path

platforms/enterprise/enterprise/src/integTest/groovy/org/gradle/internal/enterprise/DevelocityPluginConfigurationCachingIntegrationTest.groovy
```diff
@@ -223,15 +223,7 @@ class DevelocityPluginConfigurationCachingIntegrationTest extends AbstractIntegr
...
-            def rootServiceRef() {
-                def rootGradle = gradle
-                while (rootGradle.parent != null) {
-                    rootGradle = gradle.parent
-                }
-                rootGradle.extensions.serviceRef
-            }
-
-            def serviceRef = rootServiceRef()
+            def serviceRef = services.get(${GradleEnterprisePluginServiceRef.name})
```
alllex: ❓  Isn't this changing the semantics, as previously we've always accessed the services on the root build of the build tree, and now not anymore?

6hundreds: I noticed this as well, but the test is green, what make me think the test doesn't exercise this behavior enough

6hundreds:
@bot-gradle test prf

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
```diff
@@ -0,0 +1,74 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl.isolated
+
+class IsolatedProjectsCrossProjectGradleAccessIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
```
alllex, line 19: 💭  We should probably add a test that does only `getExtensions()` and there is no violation, as this is a behavior that we might change in the future.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
```diff
@@ -0,0 +1,74 @@
...
+            "findByName('ext')",
+
+            "configure(ExtraPropertiesExtension) {}",
+            "configure(TypeOf.typeOf(ExtraPropertiesExtension)) {}",
+            "configure('ext') {}",
+
+            "extraProperties",
+
+            // Groovy dynamic access
+            "ext",
```
alllex, line 70: ❓  Is it allowed to do something like `gradle.ext`, implicitly going via extensions? Should we test that as well?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/Workarounds.kt
```diff
@@ -67,6 +68,11 @@ object Workarounds {
             }
         }
 
+    fun canAccessGradleExtensionFromProjectScope(extensionType: TypeOf<*>) =
```
alllex: 🤔  I'd rather we separate CC workaround from IP workarounds.

💅  I think having explicit return types makes the signature easier to read, without having to read the body

6hundreds: > I think having explicit return types makes the signature easier to read, without having to read the body

I agree, but the rest of the code in the class ignores it, wanted to keep it consistent

6hundreds: > I'd rather we separate CC workaround from IP workarounds.

But I'd rather not introduce a new separate file for this, as I see it beneficial to keep workarounds tied together. Would a new nested `object IsolatedProjects` work for you? Then the call sites become `Workarounds.IsolatedProjects.canAccessGradleExtensionFromProjectScope` ?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
6hundreds: Not for wrapped `Gradle`. `DefaultGradle` instance has this runtime decoration because of it instantiated indirectly via `Instantiator`. `CrossProjectReportingGradle` is being instantiated plainly via constructor, so it doesn't support this pattern. That's always been the case, not something this PR brings

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
6hundreds: IDK, we're about to introduce a test(I mean in follow up PRs on project-to-build reporting ground) that will exercise all violations on `Gradle`, when the time will come `getExtensions` will be added there. 

But at the moment - what's the point to test allowed APIs? Imagine we have such a test - why only `getExtensions` deserves be tested there? Then we should add all the allowed APIs. I don't see much value in such coverage, but maybe I miss something

alllex: I don't insist, but it's best to do related things together. The point of such a "positive" test is the same as when we test that a deprecation is not issued in some cases that are relevant. We'd add a test only for extensions, because we _are_ adding violations related to extensions, and not other things.

In general, we should prefer to ship more complete things, because the follow-ups do not always happen or might not happen immediately anyway.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
alllex: Do you mean that `gradle.ext.foo` works in Vintage, and throws some exception when running with IP? (assuming the property actually exist)

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/Workarounds.kt
alllex: > as I see it beneficial to keep workarounds tied together

How do you see this workaround being tied to others in this class?

> Would a new nested object IsolatedProjects work for you?

My personal preference is to have a separate class, but something like `Workarounds.IsolatedProjects` is the next best thing.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
6hundreds: That's what I'm saying, yes

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/Workarounds.kt
6hundreds: > How do you see this workaround being tied to others in this class?

For instance, reuse the ability to disable some workarounds areas by using `withWorkaroundsFor`. I don't know if that feature even used by anyone, but I don't see how IP workarounds are different from CC in this regard. 

Because of entire cohesion in between CC and IP, I'd imaging it's good to see all the workarounds listed in file, than stumble upon `Workarounds` class, consider it holistic enough, and never find, or find later `IPWorkarounds`.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
6hundreds: It's going to be 
`> No such property: ext for class: org.gradle.internal.cc.impl.CrossProjectConfigurationReportingGradle`

alllex: This seems like a bug that we introduced together with the wrapping. Can you create an issue for it?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/Workarounds.kt
alllex: > Because of entire cohesion in between CC and IP

However, our goal is the opposite -- it's to decouple CC and IP. 

> reuse the ability to disable some workarounds areas by using withWorkaroundsFor

This is a mechanical reason to keep the workarounds together, not a semantic one.

As long as there is a clear "isolated projects" in the name, I'm okay with that. It can be a nested class or directly in the method name; that doesn't matter.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
6hundreds: Do we want to have an issue, if it's going to be a violation anyways?

alllex:
```
> No such property: ext for class: org.gradle.internal.cc.impl.CrossProjectConfigurationReportingGradle
```

Ideally, that's not the experience we are providing. And this error is not going to turn into an actionable violation, even if we made `getExtensions()` a violation without workarounds.

6hundreds: Fair, even when it become a violation, the failure mode will still be the same due to the wrong code path currently. 
The issue is here https://github.com/gradle/gradle/issues/38076

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/Workarounds.kt
6hundreds: Added a nested `IsolatedProjects` object

alllex (approved):
LGTM!

There is this comment that didn't get an answer yet: https://github.com/gradle/gradle/pull/37999#discussion_r3340188991
Just making sure it didn't get lost, it's not required to address it for merge

platforms/enterprise/enterprise/src/integTest/groovy/org/gradle/internal/enterprise/DevelocityPluginConfigurationCachingIntegrationTest.groovy
6hundreds: Added a fixture

PR #37832, opened 2026-05-07
Add NO_IMPLICIT_LOOKUP_IN_PARENT_PROJECTS feature preview

reinsch82:
## Summary

Adds a feature preview `NO_IMPLICIT_LOOKUP_IN_PARENT_PROJECTS` that opts a Vintage build out of implicit property and method lookup through the project hierarchy.

When enabled in `settings.gradle(.kts)`, child projects no longer fall back to the parent's dynamic scope. References that would have walked up the hierarchy fail locally with `MissingPropertyException` / `MissingMethodException`, matching the eventual Gradle 10 behavior. After addressing the related deprecations from #37830, users can enable the preview to confirm their migration is complete and prevent accidental reintroduction of implicit hierarchy lookups.

The preview is Vintage-only by intent: under Isolated Projects the implicit hierarchy lookup is already disabled (per #37830), so the preview is a no-op there.

Also includes:
- An integration test verifying the preview disables the parent walk in Vintage
- A release-notes entry under "Build authoring improvements"
- An entry in the `deprecated_implicit_project_hierarchy_lookup` section of the 9.6 upgrade guide, pointing migrators to the preview as a verification step

## Stack

Builds directly on #37830 (merged). #37831 (caller-context detail in deprecation messages) builds on this PR.

The plugin-ecosystem smoke validation that was previously bundled in here has been split out into a separate independent PR — it's not behaviorally tied to the feature preview, it validates #37830's deprecation across community plugins.

## Test plan

- [x] \`./gradlew :core:forkingIntegTest --tests "*ParentProjectPropertyLookup*"\`
- [x] \`./gradlew :core:test --tests "*DefaultProjectSpec*" --tests "*DefaultProjectTest*"\`
- [x] \`./gradlew :docs:checkDeadInternalLinks\`
- [x] \`./gradlew sanityCheck\`

PR #38003, opened 2026-05-26
Validate parent-walk deprecation across community plugins

reinsch82:
## Summary

Validates the parent-property-lookup deprecation merged in #37830 across the community-plugin ecosystem.

Extends `AbstractPluginValidatingSmokeTest` with a new `configures plugin ... from subproject` iteration that applies each smoke-tested plugin at the root project and references its DSL extension from a child subproject. The child reference walks the parent chain, which triggers the deprecation. Per-plugin opt-in via two new hooks:

- `getSubprojectExtensionAccess(testedPluginId, version)` — the DSL snippet placed in `child/build.gradle` (typically a reference to the plugin's project extension, e.g. `spotless { }`). Returning `null` (the default) skips the iteration for that plugin/version.
- `getSubprojectExtensionDeprecations(testedPluginId, version)` — deprecation messages the iteration expects, built with the `parentMethodInvocationDeprecation` / `parentPropertyAccessDeprecation` helpers.

Plugins that don't opt in return `null`, and the iteration is reported as **skipped** (via `Assume`, so the data provider stays non-empty for classes where nothing opts in) rather than passing as a no-op. The interesting subset (Asciidoctor, Develocity 3.x, Kotlin, Kotlin Multiplatform, Detekt, Micronaut, etc.) opts in and declares expected deprecations, documenting which plugin ecosystems trip the deprecation through normal usage.

## Isolated Projects

The subproject test is guarded with `@Requires(TestExecutionPreconditions.NotIsolatedProjects)`: under Isolated Projects #37830 disables the parent walk entirely, so the plugin's extension is not reachable from a subproject and the scenario this test exercises is Vintage-specific.

## Deprecation message alignment

The expected parent-walk warning text and upgrade-guide anchor were realigned with the message as reworded on `master` (`Implicit lookup of methods/properties in parent projects` / `deprecated_implicit_lookup_in_parent_projects`), so the opted-in iterations match the actual runtime output.

## Stack

Independent of the open PRs in the parent-property-lookup series. Builds directly on #37830 (merged). Not in the stack with #37831 (caller-context detail) or #37832 (feature preview).

## Test plan

- [x] `./gradlew :smoke-test:smokeTest --tests "*configures plugin*from subproject*"` (Vintage) — verified opt-in (Spotless) passes and non-opt-in (Nohttp) is skipped
- [ ] `./gradlew :smoke-test:isolatedProjectsSmokeTest --tests "*configures plugin*from subproject*"` — iterations excluded via `@Requires`
- [ ] `./gradlew :smoke-test:configCacheSmokeTest --tests "*configures plugin*from subproject*"`
- [ ] `./gradlew sanityCheck`

### DIFF
--- a/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractPluginValidatingSmokeTest.groovy
+++ b/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractPluginValidatingSmokeTest.groovy
@@ -17,6 +17,8 @@
 package org.gradle.smoketests
 
 import groovy.transform.SelfType
+import org.gradle.test.precondition.Requires
+import org.gradle.test.preconditions.TestExecutionPreconditions
 
 @SelfType(AbstractSmokeTest)
 abstract class AbstractPluginValidatingSmokeTest extends AbstractSmokeTest implements WithPluginValidation {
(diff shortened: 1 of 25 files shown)


testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractPluginValidatingSmokeTest.groovy
```diff
@@ -35,6 +35,50 @@ abstract class AbstractPluginValidatingSmokeTest extends AbstractSmokeTest imple
...
+     * Override to opt a plugin into the "configures plugin from subproject" smoke test. Return the
+     * DSL snippet that should go into {@code child/build.gradle} — typically a reference to the
+     * plugin's project extension (e.g. {@code spotless { }}). Return {@code null} (default) to
+     * skip the test for this plugin/version combination.
+     *
+     * <p>The purpose is to catch plugins that resolve their DSL extension through the deprecated
+     * implicit parent-project property/method lookup when configured from a subproject — the same
+     * failure mode as the Develocity 3.x case.
+     */
+    String getChildProjectConfiguration(String testedPluginId, String version) {
```
6hundreds: 🤔 I think the method name is too common. I think it either should state _what_ we need this configuration for(we use it in the only test below) or be more precise in terms what we're returning(a plugin extension).

So my options would be:
- `getPluginExtensionBlock()` - this may be reused somewhere in future, but I wouldn't bet on it
- `configurePluginInSubproject` - more precise as we say what type of configuration user need to return

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractPluginValidatingSmokeTest.groovy
```diff
@@ -35,6 +35,50 @@ abstract class AbstractPluginValidatingSmokeTest extends AbstractSmokeTest imple
         ""
     }
 
+    /**
+     * Override to opt a plugin into the "configures plugin from subproject" smoke test. Return the
+     * DSL snippet that should go into {@code child/build.gradle} — typically a reference to the
+     * plugin's project extension (e.g. {@code spotless { }}). Return {@code null} (default) to
+     * skip the test for this plugin/version combination.
```
6hundreds: ❓ 
> skip the test for this plugin/version combination

But the test isn't skipped right now, as I see? See the comment below on the test

testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/AbstractPluginValidatingSmokeTest.groovy
```diff
@@ -73,6 +117,40 @@ abstract class AbstractPluginValidatingSmokeTest extends AbstractSmokeTest imple
         (id, version) = iterations
     }
 
+    def "configures plugin #id version #version from subproject"(String id, String version) {
```
6hundreds, line 124: 🤔 I think this test isn't skipped when `getChildProjectConfiguration` returns `null`. We can't use `@Requires` here since the condition must be static for that, but maybe we can filter iterations somehow? 

Or do you think it doesn't worth it?

6hundreds (approved):
LGTM!

reinsch82:
@bot-gradle merge

PR #38005, opened 2026-05-26
Drop import layout that matches IDEA defaults

alllex:
Since we've migrated to use IDEA 2026.1.1 or later,
it seems that the new defaults for the import layout cover our needs and after sync IDEA removes such values from the `Project.xml`

`Editor > Code Style > Java`
image

---

Otherwise, after every sync, the `Project.xml` is being edited by IDEA to remove those values.


octylFractal (approved):
LGTM!

Issue #38006, opened 2026-05-27
Address Dependency Resolution IP test failures


jvandort:
### Issue type

Other

### Problem description

After enabling IP integration tests for the majority of `:dependency-management`, a few remaining tests continue to fail for unknown reasons. These tests have been marked disabled for IP test execution. We should investigate why these tests fail with IP enabled. The disabled tests do not fail with some trivial IP violation, and potentially point to bugs or behavior changes with IP. 

### Context (optional)

The disabled tests:
* https://github.com/gradle/gradle/blob/e5f7b81dfe3399d58d545db6882acc9b8500e72f/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/suppliers/DynamicRevisionRemoteResolveWithMetadataSupplierIntegrationTest.groovy#L1011
* https://github.com/gradle/gradle/blob/7b1d738675f5a4d29798bde93744af81af67736d/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/CacheResolveIntegrationTest.groovy#L74
* https://github.com/gradle/gradle/blob/fbdaf914cd0d2c0801c24172541e8c652ae28156/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/CachedMissingModulesIntegrationTest.groovy#L553

PR #38008, opened 2026-05-27
Enable IP tests for dependency management

jvandort:
Some tests require mechnical refactoring for IP compatibility Many artifact transform tests remain disabled due to the way they are written, requiring non-trivial refactorings
- Caching artifact transform tests require different Projects to share the same artifact transform Class instance, meaning we cannot duplicate the class in each project. We can potentially use included builds or buildsrc here, but these cause very slow integ test execution durations and therefore are not ideal from a DX perspective
- Some tests are marked incompatible since they specifically test IP incompatible behavior
- Four attribute compatbility tests are marked incompatible since they need to share a Class instance to detect registered rules. This is 1. a limitation of dependency management and 2. can potentially be solved with included builds, but again these are not ideal for DX
- Three tests fail during IP for unknown reasons. They are tracked by https://github.com/gradle/gradle/issues/38006

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependencyResolveIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependencyResolveIntegrationTest.groovy
@@ -17,6 +17,7 @@ package org.gradle.integtests.resolve
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.integtests.fixtures.StableConfigurationCacheDeprecations
+import org.gradle.integtests.fixtures.ToBeFixedForIsolatedProjects
 import org.gradle.integtests.fixtures.UnsupportedWithConfigurationCache
 import org.gradle.integtests.fixtures.extensions.FluidDependenciesResolveTest
 import org.gradle.integtests.fixtures.resolve.ResolveTestFixture
@@ -618,6 +619,7 @@ class ProjectDependencyResolveIntegrationTest extends AbstractIntegrationSpec im
         file("b/build/copied/a-1.0.zip").exists()
     }
 
+    @ToBeFixedForIsolatedProjects(because = "Explicitly tests IP incompatible behavior")
     def "resolving configuration with project dependency marks dependency's configuration as observed"() {
         settingsFile << """
             include 'api'
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependencyResolveIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependencyResolveIntegrationTest.groovy
@@ -17,10 +17,11 @@ package org.gradle.integtests.resolve
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.integtests.fixtures.StableConfigurationCacheDeprecations
-import org.gradle.integtests.fixtures.ToBeFixedForIsolatedProjects
 import org.gradle.integtests.fixtures.UnsupportedWithConfigurationCache
 import org.gradle.integtests.fixtures.extensions.FluidDependenciesResolveTest
 import org.gradle.integtests.fixtures.resolve.ResolveTestFixture
+import org.gradle.test.precondition.Requires
+import org.gradle.test.preconditions.TestExecutionPreconditions
 import spock.lang.Issue
 
 @FluidDependenciesResolveTest
@@ -619,7 +620,7 @@ class ProjectDependencyResolveIntegrationTest extends AbstractIntegrationSpec im
         file("b/build/copied/a-1.0.zip").exists()
     }
 
-    @ToBeFixedForIsolatedProjects(because = "Explicitly tests IP incompatible behavior")
+    @Requires(value = [TestExecutionPreconditions.NotIsolatedProjects], reason = "Explicitly tests IP incompatible behavior")
     def "resolving configuration with project dependency marks dependency's configuration as observed"() {
         settingsFile << """
             include 'api'
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/RepositoriesDeclaredInSettingsIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/RepositoriesDeclaredInSettingsIntegrationTest.groovy
@@ -18,6 +18,7 @@ package org.gradle.integtests.resolve
 
 import org.gradle.integtests.fixtures.GradleMetadataResolveRunner
 import org.gradle.integtests.fixtures.RequiredFeature
+import org.gradle.integtests.fixtures.ToBeFixedForIsolatedProjects
 import org.gradle.test.fixtures.plugin.PluginBuilder
 import org.gradle.test.fixtures.server.http.MavenHttpPluginRepository
 import org.gradle.test.precondition.Requires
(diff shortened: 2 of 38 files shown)


jvandort:
@bot-gradle help

jvandort:
@bot-gradle test ipt

jvandort:
@bot-gradle test ipt

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/api/ConfigurationRoleUsageIntegrationTest.groovy
```diff
@@ -202,6 +203,7 @@ class ConfigurationRoleUsageIntegrationTest extends AbstractIntegrationSpec impl
         'buildSrc:implementation'   | 'setCanBeDeclared'   | 'Dependency Scope' | false
     }
 
+    @ToBeFixedForIsolatedProjects(because = "Explicitly tests IP incompatible behavior")
```
octylFractal: :x: Use `@Requires` with `NotIsolatedProjects` if we never intend to fix this.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/api/ExtendingConfigurationsIntegrationTest.groovy
```diff
@@ -178,6 +179,7 @@ task checkResolveParentThenChild {
         succeeds("resolve")
     }
 
+    @ToBeFixedForIsolatedProjects(because = "Explicitly tests IP incompatible behavior")
```
octylFractal: :x: Also use `@Requires` here.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/attributes/ProjectVariantResolutionIntegrationTest.groovy
```diff
@@ -22,57 +22,76 @@ import org.gradle.integtests.resolve.VariantAwareDependencyResolutionTestFixture
 
 class ProjectVariantResolutionIntegrationTest extends AbstractIntegrationSpec implements VariantAwareDependencyResolutionTestFixture, TasksWithInputsAndOutputs {
     def "does not realize tasks that produce outgoing artifacts that are not required"() {
-        createDirs("a", "b")
         settingsFile << """
             include 'a', 'b'
         """
-        setupBuildWithColorVariants()
-        taskTypeWithOutputFileProperty()
```
octylFractal, line 30: 🤔 Why not keep some of these helpers?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
```diff
@@ -90,6 +91,7 @@ class ArtifactTransformIntegrationTest extends AbstractHttpDependencyResolutionT
         """
     }
 
+    @ToBeFixedForIsolatedProjects(because = "ArtifactTransformTestFixture is not IP compatible")
```
octylFractal, line 94: 💭 Should this potentially be on the class here? I didn't do a check if all methods are affected, but the name of the test class makes it feel like it belongs there.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/DisambiguateArtifactTransformIntegrationTest.groovy
```diff
@@ -43,6 +44,7 @@ class DisambiguateArtifactTransformIntegrationTest extends AbstractHttpDependenc
         """
     }
 
+    @ToBeFixedForIsolatedProjects(because = "Transforms registered and configured in root project")
```
octylFractal: ❓ What does this mean? This doesn't really tell me why this is incompatible.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ProjectDependencyResolveIntegrationTest.groovy
```diff
@@ -618,6 +619,7 @@ class ProjectDependencyResolveIntegrationTest extends AbstractIntegrationSpec im
         file("b/build/copied/a-1.0.zip").exists()
     }
 
+    @ToBeFixedForIsolatedProjects(because = "Explicitly tests IP incompatible behavior")
```
octylFractal: :x: Also should be `@Requires`.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/RepositoriesDeclaredInSettingsIntegrationTest.groovy
```diff
@@ -217,38 +218,39 @@ class RepositoriesDeclaredInSettingsIntegrationTest extends AbstractModuleDepend
         failure.assertHasCause("Build was configured to prefer settings repositories over project repositories but repository 'maven' was added by build file 'build.gradle'")
     }
 
+    @ToBeFixedForIsolatedProjects(because = "Explicitly tests IP incompatible behavior")
```
octylFractal: :x: Also should be `@Requires`.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/UnsafeConfigurationResolutionDeprecationIntegrationTest.groovy
```diff
@@ -17,11 +17,13 @@
 package org.gradle.integtests.resolve
 
 import org.gradle.integtests.fixtures.AbstractDependencyResolutionTest
+import org.gradle.integtests.fixtures.ToBeFixedForIsolatedProjects
 import org.gradle.integtests.fixtures.executer.GradleContextualExecuter
 import org.gradle.util.GradleVersion
 import org.spockframework.lang.Wildcard
 
 class UnsafeConfigurationResolutionDeprecationIntegrationTest extends AbstractDependencyResolutionTest {
+    @ToBeFixedForIsolatedProjects(because = "Explicitly tests IP incompatible behavior")
```
octylFractal: :x: Also should be `@Requires`.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
jvandort: Yeah I had originally put it on the class. 

But, a test marked with `@ToBeFixedForIsolatedProjects` will fail if the underlying test succeeds (as that would mean that something that was previously broken with IP now works)

But, this class has some tests already marked with `ToBeFixedForConfigurationCache`, meaning those tests "succeed" from the POV of the test interceptor in charge of processing ToBeFixedForIsolatedProjects, meaning if we annotate the whole class as ToBeFixedForIsolatedProjects, the test marked with ToBeFixedForConfigurationCache start to fail, since they succeed (since they fail)

It would be nice to mark the whole class. This applies to some of the other classes too, but I wasn't really ready to dig into test infrastructure to try to make these two annotations work nicely together.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/attributes/ProjectVariantResolutionIntegrationTest.groovy
jvandort: These helpers are used by the artifact transform tests. The helpers use `allprojects`, which the AT tests depend on. Since we can't really change the helpers to not use `allprojects` atm, we can't use these helpers in this test.

jvandort: Plus `taskTypeWithOutputFileProperty` now delegates to `fileProducerTask`, which this test still uses.

octylFractal: Alright. I suppose it's not too bad repetition.

octylFractal (approved):

PR #38010, opened 2026-05-27
Prepare release notes for Gradle 9.6.0RC1

lkasso:


platforms/core-runtime/launcher/src/main/resources/release-features.txt
alllex, line 1: ❓ Is this an intentional change?

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -52,11 +52,35 @@ For Java, Groovy, Kotlin, and Android compatibility, see the [full compatibility
 
 ## New features and usability improvements
 
-### Configuration Cache improvements
-Gradle provides a [Configuration Cache](userguide/configuration_cache.html) that improves build time by caching the result of the configuration phase and reusing it for subsequent builds.
+### Isolated Projects improvements
```
alllex: ❌  We have decided in the past that we aren't mentioning Isolated Projects in the release notes until the feature is incubating. Same as we (don't) do for Declarative.

The IP updates go into a dedicated section on its documentation page: https://docs.gradle.org/nightly/userguide/isolated_projects.html#sec:changelog

alllex (dismissed):

PR #38011, opened 2026-05-27
Fail-fast on IP violations by default

### LINKED ISSUE #37936: Deferring IP violations under parallelism can cause undesirable effects
When IP is enabled, the default build configuration strategy is _optimistic parallelism_. Gradle _assumes_ the IP constraints are honored and runs project configuration in parallel. In case a violation of such constraints is detected, it is recorded and the build outcome will ultimately be a failure.

However, the current implementation relies on the CC-flavored handling of problems -- upon discovery the problem/violation is recorded, but crucially, the execution continues, and the continuation of the work does not explicitly know that the data or state it works with can't be considered reliable or correct.

To guarantee reliability, Gradle should not execute user code against data/state that wasn't acquired reliably. Instead, the build configuration strategy should be adjusted to **fail fast** upon the discovery of problems.

The fail-fast strategy makes the migration experience harder, because the violations are then reported one at a time. That's why we also need a complementary mode dedicated to helping migration/troubleshooting scenarios:
- https://github.com/gradle/gradle/issues/34799

octylFractal:
Fixes #37936, replaces https://github.com/gradle/gradle/pull/37941.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsAccessFromGroovyDslIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsAccessFromGroovyDslIntegrationTest.groovy
@@ -26,7 +26,7 @@ import spock.lang.Issue
  * For DSL-agnostic tests prefer {@link IsolatedProjectsAccessIntegrationTest}.
  */
 class IsolatedProjectsAccessFromGroovyDslIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
-    def "reports problem when build script uses #block block to apply plugins to another project"() {
+    def "reports problem when build script uses #block block to apply plugins to another project [#mode]"() {
         createDirs("a", "b")
         settingsFile << """
             include("a")
@@ -39,10 +39,10 @@ class IsolatedProjectsAccessFromGroovyDslIntegrationTest extends AbstractIsolate
         """
 
         when:
-        isolatedProjectsDiagnosticsFails("assemble")
+        isolatedProjectsFailsUsing(mode, "assemble")
 
         then:
-        fixture.assertStateStoredAndDiscarded {
+        fixture.assertIsolatedProjectsProblems(mode) {
             projectsConfigured(":", ":a", ":b")
             problem("Build file 'build.gradle': line 3: Project ':' cannot access 'Project.plugins' functionality on $message", 2)
         }
@@ -51,31 +51,33 @@ class IsolatedProjectsAccessFromGroovyDslIntegrationTest extends AbstractIsolate
         block         | message
         "allprojects" | "subprojects via 'allprojects'"
         "subprojects" | "subprojects"
+
+        combined:
+        mode << ALL_MODES
     }
 
-    def "reports problem when build script uses #block block to access dynamically added elements"() {
+    def "reports problem when build script uses #block block to access dynamically added elements via '#statement' [#mode]"() {
         createDirs("a", "b")
         settingsFile << """
             include("a")
             include("b")
+            gradle.lifecycle.beforeProject {
+                plugins.apply('java-library')
+            }
         """
         buildFile << """
             $block {
-                plugins.apply('java-library')
-                java { }
-                java.sourceCompatibility
+                $statement
             }
         """
 
         when:
-        isolatedProjectsDiagnosticsFails("assemble")
+        isolatedProjectsFailsUsing(mode, "assemble")
 
         then:
-        fixture.assertStateStoredAndDiscarded {
+        fixture.assertIsolatedProjectsProblems(mode) {
             projectsConfigured(":", ":a", ":b")
-            problem("Build file 'build.gradle': line 3: Project ':' cannot access 'Project.plugins' functionality on $message", 2)
-            problem("Build file 'build.gradle': line 4: Project ':' cannot access 'java' extension on $message", 2)
-            problem("Build file 'build.gradle': line 5: Project ':' cannot access 'java' extension on $message", 2)
+            problem("Build file 'build.gradle': line 3: Project ':' cannot access 'java' extension on $message", 2)
         }
 
         where:
@@ -83,9 +85,15 @@ class IsolatedProjectsAccessFromGroovyDslIntegrationTest extends AbstractIsolate
         "allprojects"                       | "subprojects via 'allprojects'"
         "subprojects"                       | "subprojects"
         "configure(childProjects.values())" | "child projects"
+
+        combined:
+        statement << ["java { }", "java.sourceCompatibility"]
+
+        combined:
+        mode << ALL_MODES
     }
 
-    def "reports problem when build script uses #property property to apply plugins to another project"() {
+    def "reports problem when build script uses #property property to apply plugins to another project [#mode]"() {
         createDirs("a", "b")
         settingsFile << """
             include("a")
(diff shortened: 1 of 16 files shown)


platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsAccessFromKotlinDslIntegrationTest.groovy
```diff
@@ -207,7 +207,7 @@ class IsolatedProjectsAccessFromKotlinDslIntegrationTest extends AbstractIsolate
         """
 
         when:
-        isolatedProjectsFails("help")
+        isolatedProjectsDiagnosticsFails("help")
```
alllex: 💅 I think it's more appropriate to test with default IP here, given the failure is not a reported violation. Same as if it wasn't an IP-specific test.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsAccessFromGroovyDslIntegrationTest.groovy
```diff
@@ -51,9 +51,12 @@ class IsolatedProjectsAccessFromGroovyDslIntegrationTest extends AbstractIsolate
         block         | message
         "allprojects" | "subprojects via 'allprojects'"
         "subprojects" | "subprojects"
+
+        combined:
```
alllex, line 55: 🤩 Without `combined` the same tests would have been much more cumbersome to write. Great timing from Spock!

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsAccessFromGroovyDslIntegrationTest.groovy
```diff
@@ -51,9 +51,12 @@ class IsolatedProjectsAccessFromGroovyDslIntegrationTest extends AbstractIsolate
...
+        mode << ALL_MODES
```
alllex, line 56: 💭 It seems IntelliJ does not yet know about `combined`, so it doesn't understand that `mode` is a variable same as the ones in the `where`. Should we check if there is already a request to JB to support this? If there isn't one, maybe we should submit it ourselves

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsAccessFromGroovyDslIntegrationTest.groovy
```diff
@@ -68,10 +71,10 @@ class IsolatedProjectsAccessFromGroovyDslIntegrationTest extends AbstractIsolate
         """
 
         when:
-        isolatedProjectsDiagnosticsFails("assemble")
+        isolatedProjectsFailsUsing(mode, "assemble")
```
alllex, line 75: 🤔 This is one of those tests where I doubt it's useful to run it in both modes. There is no parallelism with respect to the violations -- all of them are reported from the root project configuration.

However, judging by the way this test is written, it's poor man's `combined:` that was relying on the leniency of the IP problem reporting.

If we were to lift the different ways of accessing the extensions into its own `combined:` clause, then running in both IP modes makes more sense.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsAccessFromGroovyDslIntegrationTest.groovy
```diff
@@ -68,10 +71,10 @@ class IsolatedProjectsAccessFromGroovyDslIntegrationTest extends AbstractIsolate
...
 
         then:
-        fixture.assertStateStoredAndDiscarded {
+        fixture.assertIpViolation(mode) {
             projectsConfigured(":", ":a", ":b")
             problem("Build file 'build.gradle': line 3: Project ':' cannot access 'Project.plugins' functionality on $message", 2)
```
alllex: 💅 This problem is superficial and blurs the purpose of the test, because the test is really verifying the access to extensions. We should avoid having this problem in the test, e.g. by applying the `java-library` plugin via `gradle.lifecycle.beforeProject {}` in the settings script.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsFixture.groovy
```diff
@@ -117,6 +118,55 @@ class IsolatedProjectsFixture {
...
+     *
+     * <p>{@link IsolatedProjectsMode#DIAGNOSTICS} forwards to {@link #assertStateStoredAndDiscarded}:
+     * all violations are collected, the cache entry is stored and then discarded.
+     *
+     * <p>{@link IsolatedProjectsMode#FAIL_FAST} only asserts that the cache was not produced
+     * and that at least one declared violation surfaced. Parallel configuration means we cannot
+     * assert which projects configured or that every declared violation fired before the
+     * interrupt propagated.
+     */
+    void assertIpViolation(IsolatedProjectsMode mode, @DelegatesTo(StateDiscardedWithProblemsDetails) Closure closure) {
```
alllex: 💅 Given this allow asserting on many IP problems, not just a single violation, I'd adjust the naming to something like `assertIsolatedProjectsProblems`

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsFixture.groovy
```diff
@@ -117,6 +118,55 @@ class IsolatedProjectsFixture {
...
+        def details = new StateDiscardedWithProblemsDetails()
+        closure.delegate = details
+        closure.resolveStrategy = Closure.DELEGATE_FIRST
+        closure()
+
+        assert spec.failed: "expected build to fail in fail-fast IP mode"
+
+        // Fail-fast may interrupt before or after the store operation starts; either way
+        // the cache entry must not be reusable.
+        def store = buildOperations.singleOrNone("Store configuration cache state")
```
alllex: 💅 It would be better to use the `ConfigurationCacheBuildOperationsFixture` instead

alllex (approved):
Looks good overall! I left some suggestions

alllex:
@bot-gradle test without pts

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsAccessFromGroovyDslIntegrationTest.groovy
octylFractal: Let's vote for https://youtrack.jetbrains.com/issue/IDEA-387234/.

octylFractal:
@bot-gradle test without pts

octylFractal:
@bot-gradle test this

PR #24833, opened 2023-04-20
6hundreds/24289/cc capturing task

6hundreds:
Fixes https://github.com/gradle/gradle/issues/24289

PR #38020, opened 2026-05-27
Fix CME when using TaskCollection<Task> as task input with Configuration Cache

### LINKED ISSUE #33318: ConcurrentModificationException when using TaskCollection<Task> as task input with Configuration Cache
### Current Behavior

Here is a simple reproducer that breaks with a CME one gradle 8.14:

```kotlin
plugins {
  base
}

val task1 by tasks.registering(Exec::class)
val task2 by tasks.registering(Exec::class)

val task3 by tasks.registering(DefaultTask::class) {
    val tasks = tasks.withType<Exec>()
    dependsOn(tasks)
    inputs.files(tasks)
} 
```

Output:

```
* What went wrong:
Configuration cache state could not be cached: field `pendingMap` of `org.gradle.api.internal.DefaultNamedDomainObjectCollection$UnfilteredIndex` bean found in field `delegate` of `org.gradle.api.internal.DefaultNamedDomainObjectCollection$FilteredIndex` bean found in field `index` of `org.gradle.api.internal.tasks.DefaultTaskCollection_Decorated` bean found in field `delegate` of `org.gradle.api.internal.tasks.DefaultRealizableTaskCollection_Decorated` bean found in input property `$1` of task `:task3` of type `org.gradle.api.DefaultTask`: error writing value of type 'java.util.LinkedHashMap'
> java.util.ConcurrentModificationException (no error message)
```

### Expected Behavior

Since this works without configuration caching, I'd also expect it to work with configuration caching.

### Context (optional)

Trying to get a project configuration cache compatible.

```kotlin
plugins {
  base
}

val task1 by tasks.registering(Exec::class)
val task2 by tasks.registering(Exec::class)

val task3 by tasks.registering(DefaultTask::class) {
    val tasks = tasks.withType<Exec>()
    dependsOn(tasks)
    inputs.files(tasks)
} 
```

### Gradle version

8.14

### Build scan URL (optional)

_No response_

Linux, Temurin JDK 23

ov7a:
Fixes
* https://github.com/gradle/gradle/issues/33318

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
@@ -550,6 +550,40 @@ class ConfigurationCacheTaskSerializationIntegrationTest extends AbstractConfigu
         outputContains("dir contents = [child.txt]")
     }
 
+    @Issue("https://github.com/gradle/gradle/issues/33318")
+    def "can use filtered custom NamedDomainObjectContainer as ad hoc task output"() {
+        buildFile """
+            class Artifact implements Named, java.util.concurrent.Callable<File> {
+                final String name
+                File file
+                Artifact(String name) { this.name = name }
+                String getName() { name }
+                File call() { file }
+            }
+
+            def artifacts = objects.domainObjectContainer(Artifact) { name -> new Artifact(name) }
+            artifacts.register("foo") { it.file = file("\${buildDir}/foo.txt") }
+            artifacts.register("bar") { it.file = file("\${buildDir}/bar.txt") }
+
+            tasks.register("aggregate") {
+                def selected = artifacts.matching { it.name in ["foo", "bar"] }
+                // `selected` exposes DefaultNamedDomainObjectCollection internals (pendingMap) — without the
+                // codec-level wrap this hits ConcurrentModificationException at CC store time.
+                outputs.files(selected)
+                def outFiles = outputs.files
+                doLast {
+                    println "outputs = " + outFiles.files*.name.sort()
+                }
+            }
+        """
+
+        when:
+        configurationCacheRun "aggregate"
+
+        then:
+        outputContains("outputs = [bar.txt, foo.txt]")
+    }
+
     @Issue("https://github.com/gradle/gradle/issues/35721")
     def "gives proper attribution to serialization failures in task closures"() {
         buildKotlinFile """
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
@@ -41,8 +41,10 @@ class ConfigurationCacheTaskSerializationIntegrationTest extends AbstractConfigu
 
             tasks.register("reader") {
                 inputs.files($tasksInput)
+                def inputFiles = inputs.files
                 doLast {
-                    println inputs.files.files*.name
+                    println "names = " + inputFiles.asFileTree.files*.name.sort()
+                    println "contents = " + inputFiles.asFileTree.files.collect { it.text }.sort()
                 }
             }
         """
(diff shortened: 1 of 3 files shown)


ov7a:
@bot-gradle test this

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -417,6 +419,17 @@ suspend fun WriteContext.writeRegisteredPropertiesOf(task: Task) {
...
 }
 
+private
+fun WriteContext.adaptInputFileValueForSerialization(value: Any?, filePropertyType: InputFilePropertyType): Any? {
+    if (filePropertyType != InputFilePropertyType.FILES) {
+        return value
+    }
+    if (value is TaskCollection<*>) {
+        return isolate.owner.serviceOf<FileCollectionFactory>().resolvingLeniently(value)
+    }
```
mlopatkin: ❓ I'm not a big fan of such a special case handling. Why only `TaskCollection`? I came up with another (similar, but not exactly) failure that can be solved by similar wrapping:
```groovy
    @Issue("https://github.com/gradle/gradle/issues/33318")
    def "can use DomainObjectCollection of Configuration as ad hoc task input"() {
        file("foo.txt").text = "foo"
        file("bar.txt").text = "bar"

        buildFile """
            configurations.register("foo") {
                canBeResolved = true
                canBeConsumed = false
            }
            configurations.register("bar") {
                canBeResolved = true
                canBeConsumed = false
            }

            dependencies {
                foo files("foo.txt")
                bar files("bar.txt")
            }

            tasks.register("reader") {
                def myConfigs = configurations.matching { it.name in ["foo", "bar"] }
                inputs.files(files(myConfigs))
                def inputFiles = inputs.files
                doLast {
                    println "inputs = " + inputFiles.files*.name.sort()
                }
            }
        """

        when:
        configurationCacheRun "reader"

        then:
        outputContains("inputs = [bar.txt, foo.txt]")
    }
``` 

In the end, I think its better to wrap the `value` unconditionally. This is something `TaskInputs` consumers are going to do later anyway.

💅 Maybe it is worth to add a comment why we're wrapping, and why we're wrapping here and not in `TaskInputs.files` (for which the reason is likely memory usage).

mlopatkin (changes requested):
I think we should go a bit further and fix other "underwrapped" things that work without CC today.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
ov7a: Thanks, agreed. I've updated the PR.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
```diff
@@ -61,7 +61,8 @@ class ConfigurationCacheTaskSerializationIntegrationTest extends AbstractConfigu
         "Collection(Task, TaskProvider)" | "[copy1.get(), copy2]" | "[copy1, copy2]"
     }
 
-    def "using a tasks from another project as 'files(#type)' input is prohibited"() {
+    def "using a task from another project as 'files' input is allowed"() {
```
ov7a, line 68: 🤔 This test is passing now. Is it ok? Note: the test was introduced in 
* https://github.com/gradle/gradle/pull/24833

mlopatkin: > 🤔 This test is passing now. Is it ok? Note: the test was introduced in
> 
> * [CC task serialization attempt when use TaskInputs.files(TaskProvider) #24833](https://github.com/gradle/gradle/pull/24833)

I think the problem with the previous solution was that we could not reliably resolve cross-project task references upon restoring. As we're now using file collection as a serialization proxy, we no longer care. 

❗ I would enrich the test (and its single-project neighbor) to verify that task dependencies are actually working (i.e. produce and consume actual files), but having it passing is fine, IMO.

🤔 It looks like we no longer need the workaround from the original PR, for any sensible build at least.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -417,6 +418,27 @@ suspend fun WriteContext.writeRegisteredPropertiesOf(task: Task) {
...
+ * The raw value registered with `TaskInputs.files(...)` can be anything `FileCollectionFactory.resolvingLeniently(...)`
+ * accepts. Wrapping the value in a resolving [FileCollection] up-front routes serialization through [FileCollectionCodec],
+ * which only serializes the resolved set of files and the task dependencies — never the source collection's
+ * mutable internals. This is semantically equivalent to what consumers of `TaskInputs` do at execution time
+ * via `FileParameterUtils.resolveInputFileValue`.
+ *
+ * Only applied to [InputFilePropertyType.FILES] — [InputFilePropertyType.FILE] and [InputFilePropertyType.DIRECTORY]
+ * expect a single path-like value on read, so wrapping in a [FileCollection] would break `inputs.file(...)` /
+ * `inputs.dir(...)`.
+ */
```
mlopatkin, line 433: 👍

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
mlopatkin: ❓ I wonder if we need to apply the same treatment to output properties. You can technically "steal" the output location from another task, though the feasibility of that is dubious.

Could you please look into a it too? From my experiments, it faces the same problem and the same workaround helps.

mlopatkin (changes requested):
I have a couple comments, and addressing the outputs case feels important enough - so the issue can be closed once and for all.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
ov7a: Great point, added

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
mlopatkin: > 🤔 It looks like we no longer need the workaround from the original PR, for any sensible build at least.

Do you foresee any troubles if we remove `withTaskReferencesAllowed` and the rest of the code that supports it? AFAICT, it only serves to work around the same problem we're fixing by introducing the `FileCollection` wrapper. Moreover, it is superseded by the wrapper - codec will never see the task again.

ov7a: I was thinking about doing that in a separate PR (sorry for not being explicit about that). I'll try creating one on top of it

ov7a: * https://github.com/gradle/gradle/pull/38070

mlopatkin (approved):

ov7a:
@bot-gradle test APT

PR #38021, opened 2026-05-27
Expose variant name directly on SourceVariantData

### LINKED ISSUE #31503: Expose variant name in SourceVariantData directly
### Issue type

Polishing

### Problem description

We have to unpack the name from the description provided to the constructor.  We should supply the name directly and format in this class.  We want to remove the need for this:

```
public String getFormattedVariantName() {
        int variantIdx = variantName.indexOf(" variant ");
        if (variantIdx == -1) {
            return variantName;
        } else {
            return variantName.substring(0, variantIdx + 9) + "'" + variantName.substring(variantIdx + 9) + "'";
        }
    }
``` 

### Context (optional)

Continues work from https://github.com/gradle/gradle/pull/30768, suggested in https://github.com/gradle/gradle/pull/30768#pullrequestreview-2476241769.

tresat:
SourceVariantData used to receive a single composed display name and
reverse-parse the variant-name token out of it via indexOf(" variant ").
That made the class brittle and dependent on the structure of an upstream
DisplayName.

Pass the variant name and its owner display separately so the formatter
can compose the user-facing string locally with no string surgery:

`ownerDisplay == null ? variantName : ownerDisplay + " variant '" + variantName + "'"`

Fixes: 
- #31503

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/model/VariantResolveMetadata.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/model/VariantResolveMetadata.java
@@ -17,6 +17,7 @@
 package org.gradle.internal.component.model;
 
 import com.google.common.collect.ImmutableList;
+import org.gradle.api.Describable;
 import org.gradle.api.internal.attributes.ImmutableAttributes;
 import org.gradle.internal.DisplayName;
 import org.gradle.internal.component.external.model.ImmutableCapabilities;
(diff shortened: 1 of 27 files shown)


tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

platforms/software/dependency-management/src/main/java/org/gradle/internal/component/model/VariantResolveMetadata.java
```diff
@@ -42,6 +42,22 @@ public interface VariantResolveMetadata {
...
+     * May be {@code null} when this variant has no meaningful owner separation (e.g. adhoc artifact sets or
+     * configurations that present themselves without a distinct variant suffix).
+     * <p>
+     * Defaults to {@code null} (no owner); implementations that have a meaningful
+     * parent must override to expose it.
+     *
+     * @return the owner display name, or {@code null} when no separation exists
+     */
+    @Nullable
+    default DisplayName getOwnerDisplayName() {
```
jvandort: It doesn't seem right for a variant to expose the display name of its owner. If you want the display name of your owner you should have a reference to the owner

tresat: This is a much bigger change.  The owner could be several different types of things.  The owner's display name was already (implicitly) present in the data here.  The goal is just to avoid needing to use string parsing to interpret and extract that name.

jvandort: This seems like something we should discuss. I don't think this solution as it stands is the direction we want to go

tresat: OK, let's discuss next week.

PR #38030, opened 2026-05-28
Restore :properties task under Isolated Projects

### LINKED ISSUE #35797: `gradle :child-project:properties`  fails with isolated projects
### Current Behavior

`gradle :child-project:properties`  fails with isolated projects enabled.

See also:

https://github.com/gradle/gradle/issues/35132

Stacktrace from CC report:

```
- [error] Project `:core-api` cannot dynamically look up a property in the parent project `:`
    - Gradle runtime
        - Exception at `kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)`
org.gradle.api.InvalidUserCodeException: Project ':core-api' cannot dynamically look up a property in the parent project ':'
        ...
	at org.gradle.internal.cc.impl.CrossProjectModelAccessTrackingParentDynamicObject.maybeReportProjectIsolationViolation(CrossProjectModelAccessTrackingParentDynamicObject.kt:142)
        ...
	at org.gradle.internal.cc.impl.TrackingDynamicLookupRoutine.getProperties$lambda$0(TrackingDynamicLookupRoutine.kt:56)
	at org.gradle.internal.cc.impl.TrackingDynamicLookupRoutine.withDynamicCall(TrackingDynamicLookupRoutine.kt:74)
	at org.gradle.internal.cc.impl.TrackingDynamicLookupRoutine.getProperties(TrackingDynamicLookupRoutine.kt:55)
	at org.gradle.api.internal.project.DefaultProject.getProperties(DefaultProject.java:1172)
	at org.gradle.api.tasks.diagnostics.PropertyReportTask.computePropertyReportModel(PropertyReportTask.java:81)
	at org.gradle.api.tasks.diagnostics.PropertyReportTask.calculateReportModelFor(PropertyReportTask.java:76)
	at org.gradle.api.tasks.diagnostics.PropertyReportTask.calculateReportModelFor(PropertyReportTask.java:43)
	at org.gradle.api.tasks.diagnostics.AbstractProjectBasedReportTask.calculateReportModel(AbstractProjectBasedReportTask.java:46)
	at org.gradle.internal.Try.ofFailable(Try.java:46)
	at org.gradle.internal.serialization.Cached$Deferred.lambda$tryComputation$0(Cached.java:80)
	at org.gradle.internal.evaluation.EvaluationContext.evaluate(EvaluationContext.java:91)
	at org.gradle.internal.serialization.Cached$Deferred.tryComputation(Cached.java:80)
	at org.gradle.internal.serialization.Cached$Deferred.result(Cached.java:71)
	at org.gradle.internal.serialization.Cached$Deferred.writeReplace(Cached.java:84)
        ...
	at org.gradle.internal.serialize.codecs.core.jos.JavaObjectSerializationCodec$WriteReplaceEncoding.encode(JavaObjectSerializationCodec.kt:167)
```

### Expected Behavior

Gradle built-in tasks are either IP-compatible, gracefully degrade, or are deprecated.

### Context (optional)

_No response_

Any project with subprojects will do. For instance, `gradle/gradle`:

`./gradlew -Dorg.gradle.unsafe.isolated-projects=true :core-api:properties`

fails with:

```

* What went wrong:
Configuration cache problems found in this build.

- Unknown location: Project ':core-api' cannot dynamically look up a property in the parent project ':'

See the complete report at file:///Users/rafael/sources/gradle/trees/master-gradle/build/reports/configuration-cache/50mndhdaz2lb62mo5huhfcx6y/5uqbrr0lsa7xe7s7guca5cf6s/configuration-cache-report.html
> Project ':core-api' cannot dynamically look up a property in the parent project ':'
```

### Gradle version

9.3.0-milestone-1

### Build scan URL (optional)

_No response_

_No response_

reinsch82:
## Summary

Closes #35797.

`:properties` fails under Isolated Projects with *use of `Project.getProperties()` is not allowed with Isolated Projects* (regression from #37782). The violation was justified when `Project.getProperties()` walked the parent projects properties, but #37830 disabled that walk under IP, making the violation superficial for internal callers like `PropertyReportTask`.

## Approach

Introduce internal-only `ProjectInternal.collectProperties()` that returns the same map without firing the deprecation or the IP violation. `PropertyReportTask` switches to it. The user-facing `Project.getProperties()` keeps the Vintage deprecation **and** the IP violation — calling `project.properties` from a build script still fails under IP exactly as before.

The method is named `collectProperties()` rather than `getPropertiesInternal()` so it isn't treated as a Groovy bean property — otherwise `BeanDynamicObject.getProperties()` would enumerate the getter and recurse into itself.

## Test plan

- [x] `:base-diagnostics:isolatedProjectsIntegTest --tests PropertyReportTaskIntegrationTest` — class-level `@ToBeFixedForIsolatedProjects` removed; new IP-only test verifies parent-only properties do **not** leak into a child's `:properties` report.
- [x] `:build-init:isolatedProjectsIntegTest --tests "BuildInitPluginIntegrationTest.creates a simple project*"` — stale `@ToBeFixedForIsolatedProjects` annotations removed (these tests also call `:properties`).
- [x] `:configuration-cache:forkingIntegTest --tests "IsolatedProjectsAccessFromGroovyDslIntegrationTest.reports problem when build script uses *"` — user-code regression guard unchanged; `println(properties)` and `println(project.properties)` still fire the violation.

Targeting `release` for RC2 per [Slack discussion](https://gradle.slack.com/archives/C05Q92AUM8D/p1779956515867689?thread_ts=1778161808.514089&cid=C05Q92AUM8D).

### DIFF
--- a/platforms/core-configuration/base-diagnostics/src/integTest/groovy/org/gradle/api/tasks/diagnostics/PropertyReportTaskIntegrationTest.groovy
+++ b/platforms/core-configuration/base-diagnostics/src/integTest/groovy/org/gradle/api/tasks/diagnostics/PropertyReportTaskIntegrationTest.groovy
@@ -17,9 +17,10 @@
 package org.gradle.api.tasks.diagnostics
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
-import org.gradle.integtests.fixtures.ToBeFixedForIsolatedProjects
+import org.gradle.test.precondition.Requires
+import org.gradle.test.preconditions.TestExecutionPreconditions
+import spock.lang.Issue
 
-@ToBeFixedForIsolatedProjects(because = "The `properties` task calls Project.getProperties() internally, which is a hard violation under Isolated Projects")
 class PropertyReportTaskIntegrationTest extends AbstractIntegrationSpec {
 
     def setup() {
@@ -52,5 +53,27 @@ class PropertyReportTaskIntegrationTest extends AbstractIntegrationSpec {
         run "properties", "-q", "--property=nonexistent"
         then:
         outputContains 'nonexistent: null'
+    }
+
+    @Issue("https://github.com/gradle/gradle/issues/35797")
+    @Requires(value = TestExecutionPreconditions.IsolatedProjects, reason = "Under Isolated Projects the parent walk is disabled, so parent-only properties must not appear in a child's properties report")
+    def "subproject properties report does not include parent-only properties under Isolated Projects"() {
+        given:
+        settingsFile << """
+            include 'sub'
+        """
+        buildFile << """
+            ext.parentOnly = 'from-root'
+        """
+        file("sub/build.gradle") << """
+            ext.childOnly = 'from-sub'
+        """
+
+        when:
+        run ":sub:properties", "-q"
+
+        then:
+        outputContains 'childOnly: from-sub'
+        outputDoesNotContain 'parentOnly'
     }
 }
--- a/platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
@@ -440,6 +440,17 @@ It is recommended to enable this flag in builds that are migrating to or already
 enableFeaturePreview("NO_IMPLICIT_LOOKUP_IN_PARENT_PROJECTS")
 ----
 
+[[sec:properties_report_no_inherited]]
+=== Properties report does not include properties inherited from parent projects
+
+Under Isolated Projects, the <<command_line_interface.adoc#sec:listing_properties,properties report>>
+of a child project lists only the properties defined on that project itself.
+Properties inherited from parent projects are not present in the report.
+This is a consequence of the <<sec:no_implicit_parents_lookup,no implicit parent lookup>> behavior.
+
+Direct uses of link:{javadocPath}/org/gradle/api/Project.html#getProperties--[`Project.getProperties()`]
+continue to be reported as an Isolated Projects violation, as that API is being phased out.
+
 [[sec:migration]]
 == Migration
 
--- a/subprojects/core/src/main/java/org/gradle/api/internal/project/MutableStateAccessAwareProject.java
+++ b/subprojects/core/src/main/java/org/gradle/api/internal/project/MutableStateAccessAwareProject.java
@@ -471,6 +471,11 @@ public abstract class MutableStateAccessAwareProject implements ProjectInternal,
         return delegate.getProperties();
     }
 
+    @Override
+    public Map<String, ? extends @Nullable Object> collectPropertiesInternal() {
+        return delegate.collectPropertiesInternal();
+    }
+
     @Nullable
     @Override
     public Object property(String propertyName) throws MissingPropertyException {
--- a/subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectInternal.java
+++ b/subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectInternal.java
@@ -232,6 +232,20 @@ public interface ProjectInternal extends Project, ProjectIdentifier, HasScriptSe
     Object getLifecycleActionsState();
 
     /**
+     * Returns the project's properties without firing the deprecation of {@link Project#getProperties()}
+     * or the Isolated Projects violation it triggers.
+     *
+     * <p>Intended for internal callers (e.g., diagnostic tasks) that have a legitimate need to enumerate
+     * the project's properties and accept responsibility for the call.
+     *
+     * @apiNote <p><b>Deliberately named {@code collect..}</b>, a {@code get*} name would make
+     * this a Groovy bean property on {@code Project}. {@code BeanDynamicObject.getProperties()}
+     * enumerates bean properties by invoking each getter — which would recurse straight back into
+     * this method via {@code extensibleDynamicObject.getProperties()}.
+     */
+    Map<String, ? extends @Nullable Object> collectPropertiesInternal();
+
+    /**
      * Two {@link ProjectInternal} instances are considered equal if their {@link #getProjectIdentity() identity} is equal.
      *
      * @param obj the object to compare with this project
--- a/platforms/core-configuration/base-diagnostics/src/main/java/org/gradle/api/tasks/diagnostics/PropertyReportTask.java
+++ b/platforms/core-configuration/base-diagnostics/src/main/java/org/gradle/api/tasks/diagnostics/PropertyReportTask.java
@@ -17,6 +17,7 @@ package org.gradle.api.tasks.diagnostics;
 
 import org.gradle.api.Incubating;
 import org.gradle.api.Project;
+import org.gradle.api.internal.project.ProjectInternal;
 import org.gradle.api.provider.Property;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Internal;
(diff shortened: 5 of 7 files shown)


subprojects/core/src/main/java/org/gradle/api/internal/project/MutableStateAccessAwareProject.java
```diff
@@ -471,6 +471,16 @@ public boolean hasProperty(String propertyName) {
         return delegate.getProperties();
     }
 
+    @Override
+    public Map<String, ? extends @Nullable Object> collectProperties() {
+        // collectProperties() is for internal callers operating on their own project (e.g. diagnostic
```
alllex: 🤔  I don't think it's a useful comment, or it can be condensed to a single short sentence. The fact that it's easy to create a long comment shouldn't be the reason to make humans read more than necessary.

For this case, the main/all information should be in the javadoc of the method

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectInternal.java
```diff
@@ -231,6 +231,22 @@ public interface ProjectInternal extends Project, ProjectIdentifier, HasScriptSe
...
+     * the project's properties and accept responsibility for the call. The user-facing
+     * {@link Project#getProperties()} continues to deprecate and raise its IP violation.
+     *
+     * <p><b>Deliberately not named {@code getPropertiesInternal()}:</b> a {@code get*} name would make
+     * this a Groovy bean property on {@code Project}. {@code BeanDynamicObject.getProperties()}
+     * enumerates bean properties by invoking each getter — which would recurse straight back into
+     * this method via {@code extensibleDynamicObject.getProperties()}. A non-{@code get} name keeps
+     * it out of bean-property enumeration.
+     */
+    Map<String, ? extends @Nullable Object> collectProperties();
```
alllex: 🤔  I'd like us to have a more telling name of the method, rather than a long javadoc. Adding a suffix or similar, like `internal`, `untracked`, etc. For example, see `StartParameterInternal.getProjectPropertiesUntracked()`.

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectInternal.java
```diff
@@ -231,6 +231,22 @@ public interface ProjectInternal extends Project, ProjectIdentifier, HasScriptSe
     @Nullable
     Object getLifecycleActionsState();
 
+    /**
+     * Returns the project's properties without firing the deprecation of {@link Project#getProperties()}
+     * or the Isolated Projects violation it triggers.
+     *
+     * <p>Intended for internal callers (e.g. diagnostic tasks) that have a legitimate need to enumerate
+     * the project's properties and accept responsibility for the call. The user-facing
+     * {@link Project#getProperties()} continues to deprecate and raise its IP violation.
```
alllex: 🤔  I don't think line belongs here as a comment. It might be something for a commit message at best

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectInternal.java
```diff
@@ -231,6 +231,22 @@ public interface ProjectInternal extends Project, ProjectIdentifier, HasScriptSe
...
+     *
+     * <p><b>Deliberately not named {@code getPropertiesInternal()}:</b> a {@code get*} name would make
+     * this a Groovy bean property on {@code Project}. {@code BeanDynamicObject.getProperties()}
+     * enumerates bean properties by invoking each getter — which would recurse straight back into
+     * this method via {@code extensibleDynamicObject.getProperties()}. A non-{@code get} name keeps
+     * it out of bean-property enumeration.
```
alllex: 🤔  `not named {@code getPropertiesInternal()}` is a bit of an over-specification. The essential part is that the name does not start with `get*` and that's all we should say as an intro.

👍  I think having a short explanation of why is useful, but I feel like it can be written more concisely.

💭  Consider an `@apiNote` section of the javadoc for this comment.

alllex (changes requested):

platforms/core-configuration/base-diagnostics/src/integTest/groovy/org/gradle/api/tasks/diagnostics/PropertyReportTaskIntegrationTest.groovy
```diff
@@ -53,4 +54,26 @@ class PropertyReportTaskIntegrationTest extends AbstractIntegrationSpec {
         then:
         outputContains 'nonexistent: null'
     }
+
+    @Issue("https://github.com/gradle/gradle/issues/35797")
+    @Requires(value = TestExecutionPreconditions.IsolatedProjects, reason = "Under Isolated Projects the parent walk is disabled, so parent-only properties must not appear in a child's properties report")
+    def "subproject properties report does not include parent-only properties under Isolated Projects"() {
```
alllex, line 60: ❌  We must also add a mention of this in the Behavior change section of the IP userguide

tresat (commented):
Software platform change is fine.

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectInternal.java
reinsch82: I shortened this a bit, the `@apiNote` thing is a nice touch, TIL :).

I still thing what is there is worth having.

gradle/wrapper/gradle-wrapper.properties
alllex, line 3: ❓  Does this change belong to the PR?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -440,6 +440,15 @@ It is recommended to enable this flag in builds that are migrating to or already
 enableFeaturePreview("NO_IMPLICIT_LOOKUP_IN_PARENT_PROJECTS")
 ----
 
+[[sec:properties_task_no_inherited]]
+=== `:properties` task does not list properties inherited from parent projects
```
alllex: 💅  `:properties` is a task in the root project, where this behavior change is not applicable. We mention it in the body of the section as well

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -440,6 +440,15 @@ It is recommended to enable this flag in builds that are migrating to or already
...
+
+Under Isolated Projects, the `:properties` task lists only the properties defined on the target project itself.
+Properties inherited from parent projects (e.g., `ext.foo = "..."` set on the root project) are not present in the report.
```
alllex: ❔  Instead of or in addition to the example, can we link to some other part of the documentation that already explains this? We might not have it...

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -440,6 +440,15 @@ It is recommended to enable this flag in builds that are migrating to or already
...
+This is a consequence of the <<sec:no_implicit_parents_lookup,no implicit parent lookup>> behavior.
+
+Direct uses of `Project.getProperties()` from build scripts (e.g., `project.properties` or bare `properties`) continue to be reported as an Isolated Projects violation, as that API is being phased out.
```
alllex: 💅  `Project.getProperties()` -- we should link to the javadoc

💅  `from build scripts` -- this is not entirely accurate as it doesn't matter where the method is called.

alllex (changes requested):

alllex (approved):

PR #38033, opened 2026-05-28
Provide more details for lambdas in CC-report

### LINKED ISSUE #22920: Configuration cache problems report should display more diagnostic information about `SerializedLambda`
### Expected Behavior
Gradle Configuration Cache report displays something like this:
```
lambda of type java.lang.invoke.SerializedLambda (impl: pkg.SomeClass.method())
```

Where:
* `pkg.SomeClass` - the full class name that is stored in `SerializedLambda.implClass` fields
* `method` - the value of `SerializedLambda.implMethodName`
* `()` - the method's signature, the value of `SerializedLambda.implMethodSignature`

### Current Behavior
Gradle Configuration Cache report displays messages like this:
```
bean of type java.lang.invoke.SerializedLambda
```

### Context
Functional tests of my plugin show some issues with Configuration Cache. I'm trying to understand the root cause, but it's not trivial now, as I get reports like this:
```
cannot serialize object of type org.gradle.api.internal.project.DefaultProject, a subtype of org.gradle.api.Project, as these are not supported with the configuration cache.
  fieldcapturedArgs of java.lang.invoke.SerializedLambda
    bean of type java.lang.invoke.SerializedLambda
      fieldval$action of org.gradle.configuration.internal.DefaultUserCodeApplicationContext$CurrentApplication$1
        bean of type org.gradle.configuration.internal.DefaultUserCodeApplicationContext$CurrentApplication$1
          fielddelegate of org.gradle.api.internal.DefaultCollectionCallbackActionDecorator$BuildOperationEmittingAction
            bean of type org.gradle.api.internal.DefaultCollectionCallbackActionDecorator$BuildOperationEmittingAction
              fieldactions of org.gradle.internal.ImmutableActionSet$SetWithFewActions
                bean of type org.gradle.internal.ImmutableActionSet$SetWithFewActions
                  fieldaddActions of org.gradle.api.internal.collections.DefaultCollectionEventRegister
                    bean of type org.gradle.api.internal.collections.DefaultCollectionEventRegister
                      fielddelegate of org.gradle.api.internal.collections.DefaultCollectionEventRegister$FilteredEventRegister
                        bean of type org.gradle.api.internal.collections.DefaultCollectionEventRegister$FilteredEventRegister
                          fieldeventRegister of org.gradle.api.internal.DefaultNamedDomainObjectSet
                            bean of type org.gradle.api.internal.DefaultNamedDomainObjectSet
                              fieldenabled of org.gradle.internal.jacoco.JacocoReportsContainerImpl
                                bean of type org.gradle.internal.jacoco.JacocoReportsContainerImpl
                                  fieldreports of org.gradle.testing.jacoco.tasks.JacocoReport
                                    task:jacocoTestAdditionalReport of type org.gradle.testing.jacoco.tasks.JacocoReport
```

It'd be really useful if we had more context in such reports. For example, what exactly lambda causes the issue.

The issue is not about `JacocoReport` only. I got similar problems with tasks of other types.

ov7a:
Fixes
* https://github.com/gradle/gradle/issues/22920

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheLambdaIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheLambdaIntegrationTest.groovy
@@ -382,4 +382,55 @@ class ConfigurationCacheLambdaIntegrationTest extends AbstractConfigurationCache
         then:
         outputContains("some value")
     }
+
+    @Issue("https://github.com/gradle/gradle/issues/22920")
+    def "report identifies lambda by impl class and method when it appears in property trace"() {
+        given:
+        file("buildSrc/src/main/java/my/LambdaTask.java").tap {
+            parentFile.mkdirs()
+            text = """
+                package my;
+
+                import java.util.function.Supplier;
+                import org.gradle.api.*;
+                import org.gradle.api.tasks.*;
+
+                public class LambdaTask extends DefaultTask {
+                    private Supplier<String> supplier;
+
+                    public void setSupplier(Supplier<String> supplier) {
+                        this.supplier = supplier;
+                    }
+
+                    public void captureProject() {
+                        Project p = getProject();
+                        setSupplier(() -> "project name is " + p.getName());
+                    }
+
+                    @TaskAction
+                    void printValue() {
+                        System.out.println("supplier -> " + supplier.get());
+                    }
+                }
+            """
+        }
+
+        buildFile << """
+            task ok(type: my.LambdaTask) {
+                captureProject()
+            }
+        """
+
+        when:
+        configurationCacheFails("ok")
+
+        then:
+        problems.htmlReport(failure.error).assertContents {
+            problemsWithStackTraceCount = 0
+            withProblem("cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.") {
+                at(":ok").at("supplier").at("lambda my.LambdaTask.lambda\$captureProject\$")
+            }
+            ignoringUnexpectedInputs()
+        }
+    }
 }
--- a/platforms/core-configuration/configuration-cache/src/test/kotlin/org/gradle/internal/cc/impl/problems/PropertyTraceTest.kt
+++ b/platforms/core-configuration/configuration-cache/src/test/kotlin/org/gradle/internal/cc/impl/problems/PropertyTraceTest.kt
@@ -19,11 +19,8 @@ package org.gradle.internal.cc.impl.problems
 import org.gradle.api.Task
 import org.gradle.internal.configuration.problems.PropertyKind
 import org.gradle.internal.configuration.problems.PropertyTrace
-
 import org.hamcrest.CoreMatchers.equalTo
-import org.hamcrest.CoreMatchers.not
 import org.hamcrest.MatcherAssert.assertThat
-
 import org.junit.Test
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertNotEquals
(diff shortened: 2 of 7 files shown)


tresat (commented):
This looks very helpful.  I think it's worth pointing out this improvement in the release notes and docs as well.

Ideally, it would be good to also see an example in the docs like the test in `ConfigurationCacheLambdaIntegrationTest` as a snippet, demonstrating the exact message when you accidentally capture something like `Project`.  If this was present then newcomers to the CC could work backwards from the error message to the information about what's permitted.

Here's the message from that test in the CC problems report:

Image

Do you think we could remove the `field capturedArgs` bit?  There isn't a field named this, which is slightly confusing.

ov7a:
> I think it's worth pointing out this improvement in the release notes

I think if we manage to provide a bunch of similar improvements, then yes, otherwise this single change is not release note worthy IMO.

>  it would be good to also see an example in the docs

Sorry, I don't think we should do that. The report should be explanatory enough. This may be part of a general doc for dealing with CC-problems, but I think it should be handled separately.

> Do you think we could remove the field capturedArgs bit? There isn't a field named this, which is slightly confusing.

I think that we should not hide this completely, since it carries information that the violating class is part of the lambda captured args. And `capturedArgs` is a field in SerializedLambda. But we could rephrase that so it's more user-facing. However, changing it only for the report turned out harder than I thought, and I don't think it's worth the effort for now.

tresat:
> > I think it's worth pointing out this improvement in the release notes
> 
> I think if we manage to provide a bunch of similar improvements, then yes, otherwise this single change is not release note worthy IMO.

If we don't advertise _somehow_ that we're putting effort into improving the CC experience, no one will appreciate the work we do. Small usability gains are still worth promoting with a sentence or two, IMO.

> Sorry, I don't think we should do that. The report should be explanatory enough. 

Maybe with the release note entry?  How do we expect people to Google the error message if we don't show an example somewhere? 

> I think that we should not hide this completely, since it carries information that the violating class is part of the lambda captured args. And `capturedArgs` is a field in SerializedLambda. But we could rephrase that so it's more user-facing. However, changing it only for the report turned out harder than I thought, and I don't think it's worth the effort for now.

If it's too much effort, than I guess it gets put in the later bin. The user-facing impact is the important part, I think. `SerializedLambda` is an internal class our users should ever have to know about.

ov7a:
> If we don't advertise somehow that we're putting effort into improving the CC experience, no one will appreciate the work we do. Small usability gains are still worth promoting with a sentence or two, IMO.

I'm not against that, but I'd prefer to postpone this until we have more to cover. cc @lkasso

> Maybe with the release note entry? How do we expect people to Google the error message if we don't show an example somewhere?

Yeah, this might be an option

> If it's too much effort, then I guess it gets put in the later bin

👍 

Given that, I'll merge what's already there and follow-up with release notes later

mlopatkin:
Is there a corresponding change to CC report? When adding new kind of a problem it is necessary to add some conversion logic here, similar to what is already there:
* https://github.com/gradle/configuration-cache-report/blob/9cb423bcbb3b4d8e6031a2d87e05f07271deadd1/src/jsMain/kotlin/configurationCache/InitConfigCacheModel.kt#L236-L285

Otherwise it doesn't render properly, AFAICT.

ov7a:
@mlopatkin My bad, sorry. I thought the improvement was good enough, even with the current shape, but I didn't realize that CC-report rendering could be tweaked further. I didn't pay attention to `Gradle Runtime` that hides SerializedLambda trace.

Now, after experimenting with it a bit, I think I can also address Tom's comment about removing an extra line, so the report can be something like:
```
- cannot serialize...
   - captured argument of lambda with parameters X,Y returning Z
      - field supplier
```  
I'll create a follow-up PR.

ov7a:
See https://github.com/gradle/gradle/pull/38114

Issue #31886, opened 2024-12-26
Expose Property.getExplicitValue to Public API


ov7a:
### Expected Behavior

Sometimes, it is useful to get the explicit value provided by the user (and/or detect that there is none). This can be done to set a complex convention based on multiple properties.

```
 // absent then there is no value set. Does not apply convention
Provider<T> configuredValue = someProperty.getConfiguredValue() 
```
Note that Property already has [a method for that](https://github.com/gradle/gradle/blob/93844251a39b463e0d03b2b2853ae3acbca8bf50/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/AbstractProperty.java#L307) but it's protected.

### Context

This request arose during provider API migration for source/target compatibility. The JavaPluginExtension sets a convention for targetCompatibility based on explicit values provided by the user.

mlopatkin:
@ov7a could you please clarify the use case here for those not deeply familiar with the internals of the java plugin, and why it cannot be modelled with the existing capabilities?

ov7a:
@mlopatkin Please see
* https://github.com/gradle/gradle/pull/31887#discussion_r1897858975

There are multiple sources of potential `targetCompatibility` - `JavaCompile..getOptions().getRelease()`, `JavaPluginExtension.targetCompatibility` (configured value without convention), `JavaCompile.sourceCompatibility` (configured value without convention) and finally `JavaCompile.javaCompiler.metadata.languageVersion`. Note that `JavaBasePlugin.configureCompileDefaults` configures both `sourceCompatibility` and `targetCompatibility` conventions. And for the `targetCompatibility` convention we need raw, configured `sourceCompatibility` value without conventions applied.

That being said, this is needed if we want to preserve the logic as is.

mlopatkin:
Ok, now I see it better.

We want to fallback to `JPE.targetCompatibility`, but only if it was explicitly set. I suppose, if we abandon backward compatibility, we can go with omitting conventions on the task and extension properties, and expose `getEffectiveTargetCompatibility` on the task, which would apply fallbacks in proper order. This would be annoying for build scripts that wire these task/extension properties to something else.

Issue #16242, opened 2021-02-22
API to determine if a property was set or if a value is obtained because of the convention


ljacomet:
When combining property values in a task and having possible conventions on these properties, it might be necessary to determine if a property was explicitly set or has a value through its convention.
The lack of this API is the reason why it is illegal to use both `*Compatibility` and toolchain in the `java` extension.

See #14160 and linked PRs.

ljacomet:
Closing as duplicate of #14768

Issue #12222, opened 2020-02-13
Do ListProperty, SetProperty and MapProperty differentiate between "EMPTY" and "UNSET" ?


martinbonnin:
Following my comment on a previous issue (https://github.com/gradle/gradle/issues/7485#issuecomment-585289792), I'm opening a new bug as I believe that even if not a bug, the behavior should be clearly documented.

I need to be able to distinguish between EMPTY and UNSET MapProperty because a user might override a previously set property with an empty one. I'm doing `mapProperty.set(null as Map<String, String>?)` to unset the value from the initial empty value.

But somehow it is possible to `.put()` values into this empty MapProperty:

### Expected Behavior

```
mapProperty.set(null as Map<String, String>?)
mapProperty.put("key", "value")
mapProperty.isPresent // this should be true
```

### Current Behavior
```
mapProperty.set(null as Map<String, String>?)
mapProperty.put("key", "value")
mapProperty.isPresent // this is currently false
```

### Context

See https://github.com/apollographql/apollo-android/issues/1987 for some context. The user is able to add to the MapProperty but his/her configuration is not picked up.

tkrullmann:
I contributed the original `MapProperty` in 5.1, so I'll try to explain the concepts as I understand them. Some additions have been made since then, but I believe the fundamentals are still true.

Collection and map properties use the concept of "collectors" to allow for lazy evaluation. Each collector can contribute a number of elements to the collection or entries to the map (possibly zero). When the property is evaluated, it iterates through all of the registered collectors to determine the actual value.

All such properties are initialized with a single collector that just contributes an empty collection or map. This wasn't always the case, but makes them much more usable and predictable, because starting with an empty collection/map is what you'd expect in most cases. It is also one of the big differences to the "simple" properties which start without a value.

When `add`ing to a collection property or `put`ting into a map property, you're really just registering another collector that will provide the given values when it is evaluated. If the collection/map starts empty, this leads to what you would expect. When you `set` the property to an explicit value, you discard all the existing collectors and replace them with one providing the given value.

However, when explicitly un-setting the property (by setting it to `null`), you replace the initial "empty" collector with a special collector that says "no value at all", and will short-circuit any evaluation of the property to "no value". It is still possible to add further collectors (hence the call to `put` is allowed) because of the lazy evaluation, but they will have no effect.

[...]

PR #38034, opened 2026-05-28
ADR Proposal: API observability standard

ljacomet:
Every public setter on a Gradle configuration API must come with a way to read the currently configured value.

Uses ResolutionStrategy as the worked counter-example and positions this ADR as the observability complement to ADR-0006.

Status:
* PROPOSED on 2026-05-28.
* ACCEPTED on 2026-06-25

### DIFF
--- a/architecture/standards/0011-api-observability.md
+++ b/architecture/standards/0011-api-observability.md
@@ -7,7 +7,7 @@
 ## Context
 
 Gradle's public API exposes a large number of *configuration APIs*: types and methods whose purpose is to let a build author or plugin tell Gradle how to behave.
-A recurring shape problem affects many of them: they ship a setter (a `set*` method, a `configure*` builder, or an action-based mutator), but no symmetric way to read the resulting state, and no way to put it back the way it was.
+A recurring shape problem affects many of them: they ship a setter (a `set*` method or a fluent verb-style mutator), but no symmetric way to read the resulting state, and no way to put it back the way it was.
 
 This makes builds:
 
@@ -19,28 +19,22 @@ This makes builds:
   There is no way to introspect that final state, and no way to recover an earlier one.
 
 The problem is not new — ADR-0006 ([Use of Provider APIs in Gradle](0006-use-of-provider-apis-in-gradle.md)) already mandates lazy `Property<T>` / `Provider<T>` properties for new configuration on tasks, extensions and domain objects.
-When that ADR is followed, observability is essentially free: a `Property<T>` getter is both the setter and the observer, the convention machinery models "default in effect", and `unset()` restores the default.
-This ADR is the observability complement to ADR-0006: it generalizes the same expectation to *every* public configuration API, including those where `Property<T>` is not the natural shape (action-based mutators, semantic verbs like `failOnVersionConflict()`, etc.).
+When that ADR is followed, observability is essentially free: a `Property<T>` getter is both the setter and the observer, and `unset()` restores the convention.
+This ADR is the observability complement to ADR-0006: it generalizes the same expectation to *every* public configuration API, including those where `Property<T>` is not the natural shape (semantic verbs like `failOnVersionConflict()`, etc.).
 
 ### Worked counter-example: `ResolutionStrategy`
 
 The public interface `org.gradle.api.artifacts.ResolutionStrategy` (`subprojects/core-api/src/main/java/org/gradle/api/artifacts/ResolutionStrategy.java`) is the clearest illustration of the cost of *not* having this standard.
 It is a configuration API with broad reach — every `Configuration` has one — and almost none of its mutators are observable.
 
-Setters with no observer and no way to restore the default:
+Setters with no observer and no way to restore the convention:
 
 * `preferProjectModules()` at `ResolutionStrategy.java:147` flips a boolean-shaped behavior; there is no `getPreferProjectModules()` and no way to unset it.
 * `cacheDynamicVersionsFor(int, String)` at `ResolutionStrategy.java:289` and `cacheDynamicVersionsFor(int, TimeUnit)` at `ResolutionStrategy.java:302` set a duration that nothing can query.
   `cacheChangingModulesFor(int, String)` at `ResolutionStrategy.java:314` and `cacheChangingModulesFor(int, TimeUnit)` at `ResolutionStrategy.java:327` are the same pattern.
-  Once any of these has been called, neither the build nor another plugin can recover the configured duration, the unit, or the fact that the default is no longer in effect.
+  Once any of these has been called, neither the build nor another plugin can recover the configured duration, the unit, or the fact that the convention is no longer in effect.
 * `sortArtifacts(SortOrder)` at `ResolutionStrategy.java:405` writes the sort order; the matching `getSortArtifacts()` / `getArtifactSortOrder()` does not exist.
 * Verb-style mutators in the same file follow the same shape: `failOnVersionConflict()`, `failOnDynamicVersions()`, `failOnChangingVersions()`, `failOnNonReproducibleResolution()`, `activateDependencyLocking()`, `deactivateDependencyLocking()`, `enableDependencyVerification()`, `disableDependencyVerification()` — all write-only, all silent about the resulting state.
-
-Action-based mutators where registered behavior is not introspectable:
-
-* `eachDependency(Action<? super DependencyResolveDetails>)` at `ResolutionStrategy.java:277` — each registered action is opaque to every later caller.
-* `dependencySubstitution(Action<? super DependencySubstitutions>)` at `ResolutionStrategy.java:375` — the rules added inside the action vanish behind a `DependencySubstitutions` whose registered rules are not enumerable.
-* `componentSelection(Action<? super ComponentSelectionRules>)` at `ResolutionStrategy.java:344` and `capabilitiesResolution(Action<? super CapabilitiesResolution>)` at `ResolutionStrategy.java:414` — same shape.
 
 Acknowledging the partial prior art on this same interface:
 
@@ -49,7 +43,10 @@ Acknowledging the partial prior art on this same interface:
 * `getUseGlobalDependencySubstitutionRules()` at `ResolutionStrategy.java:389` is a `Property<Boolean>`, and reads exactly the way ADR-0006 prescribes — observable and restorable through the convention machinery.
   The contrast with `preferProjectModules()` on the same interface is striking: both are boolean-shaped, but only one is fit to compose with.
 
-The remainder of this ADR sets a standard that would have prevented this shape, and gives `ResolutionStrategy` an explicit path forward.
+The same interface also carries action-based mutators — `eachDependency` at `ResolutionStrategy.java:277`, `dependencySubstitution` at `ResolutionStrategy.java:375`, `componentSelection` at `ResolutionStrategy.java:344`, and `capabilitiesResolution` at `ResolutionStrategy.java:414` — where behavior is either registered to run later or used to configure a container.
+Whether (and how) such APIs need to be observable raises a different set of questions and is left to a follow-up ADR; this ADR focuses on plain setters.
+
+The remainder of this ADR sets a standard that would have prevented the setter shape above, and gives `ResolutionStrategy` an explicit path forward.
 
 ## Decision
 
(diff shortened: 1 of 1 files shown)


architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,121 @@
...
+
+This ADR applies to every public configuration API in Gradle: a public type, or public method on a public type, whose purpose is to let users configure how some Gradle behavior runs.
+It does not apply to internal types, to behavior-free data carriers, or to identity (see ADR-0006 on identity information).
+
+The keywords MUST, MUST NOT, SHOULD, SHOULD NOT and MAY are used as defined in RFC 2119.
+
+1. **Every public setter MUST have an observer.**
+   For every public setter — whether spelled `setX(...)`, `x(...)`, a fluent verb like `failOnVersionConflict()`, or a `configure*` block — the same API MUST expose a way to read the currently configured value.
+   The observer MAY take any of these shapes: a `Property<T>` / `Provider<T>` getter (preferred; see ADR-0006), a plain getter, or a documented query method that returns the effective configuration as data.
+   The observer MUST also let the caller distinguish "the default is in effect" from "the default value happens to have been configured explicitly".
```
mlopatkin: ❓ Convention vs. explicit value is not something `Property` gives you today. Do you imply we need this? Another tangentially related topic - should the convention value (default) be available separately or just `unset` is enough?

architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,121 @@
...
+   With `Property<T>`, this falls out of the convention machinery; with other shapes it MUST be documented.
+
+2. **Public configuration APIs SHOULD let the user restore the default.**
+   For every behavior that has a default, the API SHOULD expose a way to put it back: `Property.unset()` on a lazy property, an explicit `resetX()` / `clearX()` method, a nullable setter documented to mean "use the default", or an equivalent.
+   Where restoration is fundamentally impossible (the side effect cannot be undone), the API MUST document this explicitly.
+   "Restoration is not implemented yet" is not the same thing as "restoration is impossible" and MUST NOT be papered over.
+
+3. **Action-based mutators MUST be introspectable, or MUST be re-modelled as data.**
+   When a public API accepts an `Action<...>` / `Closure` that registers behavior (rules, callbacks, configuration blocks), one of the following MUST hold:
```
mlopatkin: 💅 I think immediately executed configuration blocks should be explicitly excluded there. E.g.
`configuration blocks to be executed later`. OTOH, it is somewhat covered by the first form applied to the object being configured.

architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,121 @@
...
+  Once any of these has been called, neither the build nor another plugin can recover the configured duration, the unit, or the fact that the default is no longer in effect.
+* `sortArtifacts(SortOrder)` at `ResolutionStrategy.java:405` writes the sort order; the matching `getSortArtifacts()` / `getArtifactSortOrder()` does not exist.
+* Verb-style mutators in the same file follow the same shape: `failOnVersionConflict()`, `failOnDynamicVersions()`, `failOnChangingVersions()`, `failOnNonReproducibleResolution()`, `activateDependencyLocking()`, `deactivateDependencyLocking()`, `enableDependencyVerification()`, `disableDependencyVerification()` — all write-only, all silent about the resulting state.
+
+Action-based mutators where registered behavior is not introspectable:
+
+* `eachDependency(Action<? super DependencyResolveDetails>)` at `ResolutionStrategy.java:277` — each registered action is opaque to every later caller.
+* `dependencySubstitution(Action<? super DependencySubstitutions>)` at `ResolutionStrategy.java:375` — the rules added inside the action vanish behind a `DependencySubstitutions` whose registered rules are not enumerable.
```
mlopatkin: :thinking: This is mostly about `DependencySubstitutions` not following the rules? The action is just a DSL convenience, and it runs immediately, so there is nothing to observe about it.

architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,121 @@
...
+   Migration MAY be incremental — adding an observer in one release, restoration in the next, action introspection later — and SHOULD be tracked alongside the normal deprecation pipeline.
+   Where adding a getter to a public interface would break binary compatibility for implementors outside Gradle, the migration SHOULD use a default method or a companion query type rather than skipping the observer altogether.
+
+### Terminology
+
+* **Configuration API** — a public type or public method on a public type whose purpose is to let users tell Gradle how to behave.
+  Tasks, extensions, domain objects, and types like `ResolutionStrategy` all qualify.
+  Pure data carriers and identity getters do not.
+* **Observer** — a public read API that returns the currently configured value of some piece of behavior, and that lets the caller distinguish "the default is in effect" from "an explicit value has been configured".
+  A `Provider<T>` whose presence reflects whether a convention is in effect counts; a plain getter that returns the convention without disclosing that fact does not.
```
mlopatkin: :question: 
> a plain getter that returns the convention without disclosing that fact does not.

This is all what `Property` gives you today, though.

architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,121 @@
...
+   Where restoration is fundamentally impossible (the side effect cannot be undone), the API MUST document this explicitly.
+   "Restoration is not implemented yet" is not the same thing as "restoration is impossible" and MUST NOT be papered over.
+
+3. **Action-based mutators MUST be introspectable, or MUST be re-modelled as data.**
+   When a public API accepts an `Action<...>` / `Closure` that registers behavior (rules, callbacks, configuration blocks), one of the following MUST hold:
+
+   * the API exposes an observer over the *result* of all registered actions (e.g. the set of rules they produced, the count and provenance of registered actions, or an equivalent introspection surface), or
+   * the API is re-modelled so that the configuration is data, not opaque behavior — typically a managed domain object container, a `NamedDomainObjectContainer`, or a `Property<T>` over a value type.
+     The container MUST let callers enumerate what has been configured (entries, names, identities); it does NOT have to expose recursive observability into the internal state of each entry.
```
mlopatkin: :question: These were the most controversial during our discussion in AC

The usefulness of this as a blanket rule is questionable. For example, for tasks it is reasonable to be able to query task names of registered tasks. However, there are other opaque callbacks, like `lifecycle.beforeProject`. Is there any value in looking at those? What should you be able to do with them then?

There is also a case of lazy callbacks that we discard after applying. So the introspector may query the number of registered task configuration callbacks and see e.g. 3, but then see only 0 after the task has been created and configured.

Probably, this entry can benefit from a more elaborate description and more concrete examples.

architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,121 @@
...
+  Without an observer, a plugin cannot inspect what an earlier plugin (or the user) has already configured, and cannot decide intelligently whether to override, merge, or step aside.
+* **Order-dependent and lossy.** With several plugins all calling write-only setters, the final state is whatever happened to be applied last.
+  There is no way to introspect that final state, and no way to recover an earlier one.
+
+The problem is not new — ADR-0006 ([Use of Provider APIs in Gradle](0006-use-of-provider-apis-in-gradle.md)) already mandates lazy `Property<T>` / `Provider<T>` properties for new configuration on tasks, extensions and domain objects.
+When that ADR is followed, observability is essentially free: a `Property<T>` getter is both the setter and the observer, the convention machinery models "default in effect", and `unset()` restores the default.
+This ADR is the observability complement to ADR-0006: it generalizes the same expectation to *every* public configuration API, including those where `Property<T>` is not the natural shape (action-based mutators, semantic verbs like `failOnVersionConflict()`, etc.).
```
mlopatkin: :nail_care: It might be better to call "default" a "convention" to be consistent with ADR-0006, though I don't feel too strongly about it.

mlopatkin (commented):
Thank you for raising this!
We went through the proposal during the AC session. Undoubtedly, there is value in unifying our API and reducing the number of "verb-style" mutators 👍 

We're still uncertain about requiring mandatory observability of registered callbacks, etc. It is not clear what value the users can get from them. Sharing more concrete use cases may help. Alternatively, to move forward, we may extract that requirement into a separate ADR and land the rest.

Comments below are mine, though most are based on the discussion.

architecture/standards/0011-api-observability.md
ljacomet: Softened to `SHOULD` and made out of scope to resolve the fact that `Property` does not give it.

However, I really believe it should (see https://github.com/gradle/gradle/issues/16242, which probably should not be closed as duplicate of the other one, not sure what I did then, it has overlap, but not a full one 🤔 )

architecture/standards/0011-api-observability.md
ljacomet: Agreed, removed. Let's focus on non container APIs in this ADR

ljacomet:
Thanks @mlopatkin. Integrated the feedback in df3591c84ed:

* Action-based mutators are out of this ADR. Dropped rule 3 entirely and took the AC's suggestion to land the setter-focused parts first. Lifecycle hooks, lazy-discarded callbacks, and the use-case questions are substantial enough to warrant their own ADR.
* Convention-vs-explicit softened to SHOULD, scoped to cases where the API's behavior depends on it (e.g. deprecating a convention value), with an explicit note that Property<T> doesn't surface this distinction today and that closing the gap is out of scope here.
* "default" → "convention" throughout, for consistency with ADR-0006.

architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,108 @@
...
+Whether (and how) such APIs need to be observable raises a different set of questions and is left to a follow-up ADR; this ADR focuses on plain setters.
+
+The remainder of this ADR sets a standard that would have prevented the setter shape above, and gives `ResolutionStrategy` an explicit path forward.
+
+## Decision
+
+This ADR applies to every public configuration API in Gradle: a public type, or public method on a public type, whose purpose is to let users configure how some Gradle behavior runs.
+It does not apply to internal types, to behavior-free data carriers, or to identity (see ADR-0006 on identity information).
+
+The keywords MUST, MUST NOT, SHOULD, SHOULD NOT and MAY are used as defined in RFC 2119.
```
big-guy:
```suggestion
The keywords MUST, MUST NOT, SHOULD, SHOULD NOT and MAY are used as defined in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
```

architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,108 @@
...
+
+1. **Every public setter MUST have an observer.**
+   For every public setter — whether spelled `setX(...)`, `x(...)`, or a fluent verb like `failOnVersionConflict()` — the same API MUST expose a way to read the currently configured value.
+   The observer MAY take any of these shapes: a `Property<T>` / `Provider<T>` getter (preferred; see ADR-0006), a plain getter, or a documented query method that returns the effective configuration as data.
```
big-guy, line 59:
```suggestion
   The observer MAY take any of these shapes: a `Property<T>` / `Provider<T>` getter (preferred; see [ADR-0006](0006-use-of-provider-apis-in-gradle.md)), a plain getter, or a documented query method that returns the effective configuration as data.
```

architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,108 @@
...
+   Where the API's behavior depends on whether the convention is in effect — most commonly, deprecating or changing a convention value — the observer SHOULD also let the caller distinguish "the convention is in effect" from "an explicit value happens to equal the convention".
+   `Property<T>` does not surface this distinction today; closing that gap is out of scope for this ADR.
```
big-guy: I don't think this is justified above. I know this is something you've advocated for, but since we don't have it for Providers and Provider API is the preferred migration path, this is telling people they should do something extra on top of Provider API and I don't think that's the intention.

Do you have some example where this would make sense today if we didn't introduce Provider API alternatives? Maybe there are APIs that should never be Provider API and need this?

architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,108 @@
...
+   Where adding a getter to a public interface would break binary compatibility for implementors outside Gradle, the migration SHOULD use a Java `default` method or a companion query type rather than skipping the observer altogether.
+
+This ADR deliberately does not cover action-based mutators — APIs that accept an `Action<...>` / `Closure` registering behavior to run later, or used to configure a container.
+That question — whether and how registered callbacks should be observable, and where lifecycle hooks fit — is left to a follow-up ADR.
+
+### Terminology
+
+* **Configuration API** — a public type or public method on a public type whose purpose is to let users tell Gradle how to behave.
+  Tasks, extensions, domain objects, and types like `ResolutionStrategy` all qualify.
+  Pure data carriers and identity getters do not.
```
big-guy: What's a "pure data carrier"?  Is an "identity getter" just a name/String?

architecture/standards/0011-api-observability.md
big-guy: Ah, I see you mention this ADR above too, so feel free to move this to the first occurrence only

architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,108 @@
...
+* `getUseGlobalDependencySubstitutionRules()` at `ResolutionStrategy.java:389` is a `Property<Boolean>`, and reads exactly the way ADR-0006 prescribes — observable and restorable through the convention machinery.
+  The contrast with `preferProjectModules()` on the same interface is striking: both are boolean-shaped, but only one is fit to compose with.
+
+The same interface also carries action-based mutators — `eachDependency` at `ResolutionStrategy.java:277`, `dependencySubstitution` at `ResolutionStrategy.java:375`, `componentSelection` at `ResolutionStrategy.java:344`, and `capabilitiesResolution` at `ResolutionStrategy.java:414` — where behavior is either registered to run later or used to configure a container.
+Whether (and how) such APIs need to be observable raises a different set of questions and is left to a follow-up ADR; this ADR focuses on plain setters.
+
+The remainder of this ADR sets a standard that would have prevented the setter shape above, and gives `ResolutionStrategy` an explicit path forward.
+
+## Decision
```
big-guy, line 51: Could you add a preferred/acceptable/unacceptable example kind of like ADR-0006?

Maybe start with an existing API (it could be made up) and then show how each recommendation changes it.

big-guy (changes requested):

architecture/standards/0011-api-observability.md
mlopatkin: >Do you have some example where this would make sense today if we didn't introduce Provider API alternatives? Maybe there are APIs that should never be Provider API and need this?

We have some use cases for getting the explicitly set value, but one can argue this is just an unfortunate legacy shape, which shouldn't have happened in the first place:
* https://github.com/gradle/gradle/issues/31886

There were also some discussions around this with regard to `MapProperty` vs `Property<Map>`, but they only touch the topic:
* https://github.com/gradle/gradle/issues/12222#issuecomment-1936385836

architecture/standards/0011-api-observability.md
ljacomet: Very first occurrence has a link, but I still added one link per occurrence per section.

architecture/standards/0011-api-observability.md
ljacomet: Removed

architecture/standards/0011-api-observability.md
ljacomet: These referred to effectively types that don't have setters, removing mentions since this ADR is about data you can change, and thus should be able to observe. Mentioning data you cannot change is irrelevant.

architecture/standards/0011-api-observability.md
ljacomet: I've added something, about the first rule applied to an existing API.

big-guy (approved):

architecture/standards/0011-api-observability.md
```diff
@@ -0,0 +1,127 @@
+# ADR-0011 - API observability
+
+## Status
+
+- PROPOSED on 2026-05-28
```
big-guy: Change this to ACCEPTED and announce it in #bt

PR #38045, opened 2026-05-29
Update Documentation: address naming pattern confusion with CC

### LINKED ISSUE #22122: Document `name punning` pattern in the configuration cache docs
And review documentation samples/snippets to add an explaining comment to each occurrence.

lkasso:
### Details

### Context
- Fix https://github.com/gradle/gradle/issues/22122

### Documentation Checklist


platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -268,6 +268,38 @@ Use a task input or output property or a script variable to capture the result o
...
+== Reducing Captured References in Ad-Hoc Tasks
+
+When using ad-hoc tasks in build scripts, task actions (such as `doLast {}`) can inadvertently capture references to the enclosing build script scope.
+In Kotlin DSL scripts, referencing a script-level property inside a `doLast {}` lambda causes the lambda to capture the build script object, which is not serializable.
+
+The following ad-hoc task directly references a script-level property inside its `doLast {}` action:
+
+====
+include::sample[dir="snippets/optimizing-builds/configuration-cache/capturedReferences/kotlin",files="build.gradle.kts[tags=dont]"]
+include::sample[dir="snippets/optimizing-builds/configuration-cache/capturedReferences/groovy",files="build.gradle[tags=dont]"]
```
mlopatkin: ❓ The whole "name punning" is only relevant to Kotlin. For Groovy it only becomes problematic when you're accessing something actually defined in `Project`, like `version` or `name`.

I suggest we drop Groovy example for top-level variable capture. 

However we can instead go further and cover some stuff that is universally applicable. We reshuffle things a bit and highlight the accidental implicit project capture first:
```gradle
tasks.register("dont") {
  doLast { println(version) }  // Requires script/project because version is resolved there.
}

tasks.register("do") {
  def projectVersion = version // This is necessary in both Groovy and Kotlin
  doLast { println(projectVersion) }
}
```

Then we proceed and say that in Kotlin top-level variables are also problematic and show with the current recipe (Kotlin-only, no Groovy).

Then there is another angle of the problem, which is related and likely belongs here. Sometimes in Kotlin you also need to widen the type of the variable:
```kotlin
tasks.register("dont") {
  val runtimeClasspath = configurations.runtimeClasspath  // Not enough
  doLast { println(runtimeClasspath.get().files) }
}

tasks.register("do") {
  val runtimeClasspath: Provider<out FileCollection> = configurations.runtimeClasspath  // Note the explicit type
  doLast { println(runtimeClasspath.get().files) }
}
```
I think covering that here too makes sense.

platforms/documentation/docs/src/snippets/optimizing-builds/configuration-cache/capturedReferences/tests/capturedReferences.sample.conf
```diff
@@ -0,0 +1,4 @@
+commands: [{
```
mlopatkin, line 1: 💅 Maybe also add a test for `produce` but run it with `--no-configuration-cache`. That way we'll be sure the snippet actually works without CC.

mlopatkin (changes requested):
Looks great overall, but its applicability, or rather lack thereof, to Groovy bugs me. I'd rather leave Groovy out for that particular case, though see inline comments. I left some suggestions on including some closely related topics that, I believe, aren't yet covered specifically in CC docs too.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
lkasso: Thanks for clearing this up. I wasn't too sure.
I updated the Kotlin snippet and removed the Groovy one.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -272,31 +272,62 @@ Use a task input or output property or a script variable to capture the result o
...
+
+====
+include::sample[dir="snippets/optimizing-builds/configuration-cache/capturedReferences/kotlin",files="build.gradle.kts[tags=widen-dont]"]
+====
+<1> The inferred type is `NamedDomainObjectProvider<Configuration>`, which carries a project reference.
+
+====
+include::sample[dir="snippets/optimizing-builds/configuration-cache/capturedReferences/kotlin",files="build.gradle.kts[tags=widen-do]"]
+====
+<1> The explicit `Provider<out FileCollection>` type drops the container reference.
```
mlopatkin: 💅 I'd rephrase it "The explicit Provider type, dropping both Container reference and unsupported Configuration type, allows CC to simplify the internal representation of the instance and ignore the unsupported state" to be closer to what actually happens.

mlopatkin (approved):
Looks good, with one suggestion.

PR #38047, opened 2026-05-29
Add lazy destinationDirectory property to Copy and Sync

### LINKED ISSUE #13275: Introduce directory property to represent output of Copy and Sync tasks
These task should get a `DirectoryProperty` called `destinationDirectory` to replace the simple `getDestinationDir() : File` property.

### LINKED ISSUE #25824: Add a lazy way to set Copy#destinationDir
### Expected Behavior

Add a lazy way to set Copy#destinationDir

Currently, the setting Copy task destination means calling setDestinationDir that takes a file. If you have a provider, you have no way to use it for set up.

### Context

As Project.buildDir is deprecated, the suggested path is to use Project.layout.buildDirectory.file("foo"), but you are unable to use that on a Copy task in a lazy way.

asodja:
Adds a new `DirectoryProperty getDestinationDirectory()` to the Copy and Sync tasks so the copy destination can be configured lazily with a Provider (e.g. `layout.buildDirectory.dir(...)`) and wired as a task output to downstream tasks.

The change is purely additive and non-breaking: the existing `into(...)`, `setDestinationDir(File)` and `getDestinationDir()` API keep working unchanged. They share a single source of truth with the new property, which lives in DestinationRootCopySpec (the destination-aware root spec shared by Copy and Sync). `into(Object)` funnels every legacy notation into the property and `getDestinationDir()` reads back from it, so all entry points stay consistent.

The legacy `getDestinationDir()` getter moves from @OutputDirectory to @ReplacedBy("destinationDirectory") (the new property is now the registered output) and is marked @NotToBeReplacedByLazyProperty. The new getter is @Incubating.

DestinationRootCopySpec gains a FilePropertyFactory dependency, threaded through FileCopier, DefaultFileOperations and the configuration cache codec (which continues to serialize the resolved destination File).

Fixes #25824
Fixes https://github.com/gradle/gradle/issues/13275


asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test ACT

platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/copy/DestinationRootCopySpec.java
```diff
@@ -41,14 +50,36 @@ protected CopySpecInternal getDelegateCopySpec() {
...
+    public CopySpec into(@Nullable Object destinationDir) {
+        if (destinationDir == null) {
+            destinationDirectory.set((Directory) null);
+        } else if (destinationDir instanceof DirectoryProperty) {
+            destinationDirectory.set((DirectoryProperty) destinationDir);
+        } else if (destinationDir instanceof Directory) {
+            destinationDirectory.set((Directory) destinationDir);
+        } else if (destinationDir instanceof Provider) {
+            destinationDirectory.fileProvider(((Provider<?>) destinationDir).map(transformer(value ->
+                value instanceof FileSystemLocation
```
abstratt, line 65: ❓ 

Is this to support a `RegularFileLocation`?

asodja: It's mostly to support passing `Provider<Directory>`. We handle `DirectoryProperty` explicitly, but we can't `Provider<Directory>`. Additionally it also handles `Provider<RegularFile>` and `RegularFileProperty`

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/DestinationRootCopySpecCodec.kt
abstratt, line 35: ❓ 

TBH, I don't know why we have a custom codec for `DestinationRootCopySpec`, but shouldn't we change this codec to encode/decode the lazy property instead of the eager getter?

abstratt: Actually, I think we may not need this codec at all anymore.

subprojects/core/src/main/java/org/gradle/api/tasks/Copy.java
```diff
@@ -91,13 +93,28 @@ public DestinationRootCopySpec getRootSpec() {
...
+     * <p>
+     * Setting this property is equivalent to calling {@link #into(Object)} on this task, and reading it reflects
+     * the destination configured through {@link #into(Object)} or {@link #setDestinationDir(File)}.
+     *
+     * @return the destination directory property
+     * @since 9.7.0
+     */
+    @Incubating
+    @OutputDirectory
+    public DirectoryProperty getDestinationDirectory() {
```
big-guy, line 108: I think you'll need to add this to  `platforms/documentation/docs/src/docs/dsl/org.gradle.api.tasks.Copy.xml` and Sync's.

big-guy (approved):

asodja:
Thanks for a review! 

Even though it's accepted, I decided to move this to 9.8 since I don't see it urgent for 9.7 and I will be off for 9.7 release. And with that I won't be able to fix any possible issues with this (and since it's a complex Copy task there may be some).

Goooler:
This should also close #13275.

asodja:
Thanks @Goooler, linked

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/api/tasks/Copy.java
asodja: Good catch, fixed

asodja:
@bot-gradle test this

Issue #37922, opened 2026-05-18
Deprecate implicit property and method lookup in project hierarchy


alllex:
Context:
- https://github.com/gradle/gradle/issues/37921

---

### Why deprecate now

The linked feature-change issue covers the motivation for removing implicit parent-walking in Gradle 10: hidden coupling between projects, error-prone silent resolution, and fundamental incompatibility with Isolated Projects. This deprecation is the 9.x notice that the behavior is going away, and it doubles as a pre-flight check for IP migration.

Two user populations benefit:

- **Vintage users planning to stay on Vintage through the 9.x cycle.** They need warning that the resolution semantics will change in Gradle 10, with enough lead time to migrate without a flag-day rewrite.
- **Vintage users moving to Isolated Projects in 9.x.** Today the parent walk produces IP violations that users perceive as mechanical, because lookups that were never intended to reach a parent still trigger them. Addressing the deprecation in Vintage removes the underlying parent reliance, which removes those IP violations in turn.

Users who resolve all of the resulting deprecations during 9.x experience no breaking change from this concern when upgrading to 10.

### Priorities for the design

The detailed design is being worked through in the implementation PR. The priorities that should drive design decisions:

- **Warn only when the behavior is actually relied on.** The deprecation should fire when a child project's lookup *successfully* resolves from a parent — i.e., when removing the behavior would change the outcome. Lookups that resolve locally, or that return no result, should not produce noise.
- **The message should point at the right thing.** Users invoke this lookup in several different ways, and the deprecation should be accurate to how they actually triggered it, so the migration step is obvious from the message alone.
- **Offer an early opt-in into the Gradle 10 end-state.** Users who want to pre-migrate — particularly those preparing for IP — should be able to opt into the post-removal semantics in 9.x without waiting for Gradle 10. The opt-in should apply uniformly to both Vintage and IP so users see the same behavior they will see in 10.
- **Make the silent failure modes auditable.** Some shapes of the lookup will silently return no result after removal rather than failing loudly. The deprecation, the upgrade guide, and the early opt-in documentation should give users a concrete way to find these call sites before they switch.

### Out of scope

- Changing the resolution semantics by default in 9.x. The behavior change happens in 10. The 9.x cycle is warning + opt-in only.
- Designing replacements for inheritance patterns. The upgrade guide covers the recommended migrations (declared properties, convention plugins, explicit ancestor access as a transitional fix); this issue is about the deprecation itself.

---

Relevant issues and docs
[...]

PR #37742, opened 2026-04-29
Deprecate Develocity plugin versions before 4.0

reinsch82:
## Summary

- Emit a build-level deprecation warning at Develocity plugin check-in for any plugin version below 4.0, advising users to upgrade ahead of the parent-property-lookup removal in Gradle 10.
- Add a self-contained `[[deprecated_develocity_plugin_pre_4_0]]` section in the v9 upgrade guide explaining the deprecation and the upcoming change it pre-announces.
- Add a shared `expectParentPropertyLookupDeprecation(executer, version)` helper to `BaseBuildScanPluginCheckInFixture` so the affected enterprise tests can register the expected warning in one line; wire it into the existing tests that explicitly use plugin versions below 4.0 (auto-apply matrix, IP-supported branch, classpath transitive plugin).

This change lands ahead of the larger parent-property-lookup deprecation work (gradle/gradle#37205) so users on Develocity 3.x get advance notice that an upgrade will be required for Gradle 10 compatibility.

## Why

Develocity plugin 4.0 is the first version that registers its project extension on every project — earlier versions rely on Gradle's implicit parent-project property lookup to resolve `develocity { ... }` from subprojects. Gradle 10 removes that implicit lookup, after which subprojects on Develocity 3.x will no longer be able to configure scan tags or values. Rather than have those users discover this when they upgrade Gradle, we surface it now via a deprecation tied to the plugin-check-in machinery.

## Test plan

- [x] New `emits deprecation when Develocity plugin version #pluginVersion relies on parent-property lookup` test covers 3.17, 3.19.2 (deprecated) and 4.0, 4.4.1 (silent).
- [x] Existing tests broken by the new deprecation updated:
    - `BuildScanAutoApplyIntegrationTest` — `deprecated` column added to the three "uses #sequence version of plugin when ..." data tables.
    - `BuildScanAutoApplyClasspathIntegrationTest` (Develocity + GradleEnterprise variants) — both tests using `MINIMUM_SUPPORTED_PLUGIN_VERSION_NUMBER` for the transitive plugin.
    - `DevelocityPluginCheckInIntegrationTest` — IP `[3.15, supported: true]` row.
- [x] All 94 affected tests pass locally on `:enterprise:embeddedIntegTest`.
- [x] `:docs:checkDeadInternalLinks` passes (new upgrade-guide anchor is valid).

PR #38051, opened 2026-05-29
Require DV plugin 4.0+ with IP

alllex:
Gradle 9.6.0 changes Isolated Projects behavior by removing the implicit parent-project property lookup. This produces the same effect on the Develocity plugin as upgrading past `FIRST_PLUGIN_VERSION_WITHOUT_PARENT_PROPERTY_LOOKUP`: plugin versions older than 4.0 silently miss properties they used to inherit from parent projects, leading to potentially incomplete build scans under IP.

That's why we are raising the plugin version requirement to `4.0` for IP in the same Gradle release.

---

Additionally, this PR also removes a misleading part of the deprecation message for implicit lookup in parent projects: it was saying that the plugin application will be ignored, while in reality the plugin was still applied and just a deprecation was issued.

---

Related:
- https://github.com/gradle/gradle/issues/37922
- https://github.com/gradle/gradle/pull/37742


alllex:

reinsch82 (approved):

alllex:
@bot-gradle test this

wolfs (approved):

lkasso (approved):

Issue #38049, opened 2026-05-29
`ConcurrentModificationException` during dependency graph resolution (9.6.0-rc-1)


alllex:
Starting with **Gradle 9.6.0-rc-1**, resolving a configuration in a specific multi-project setup fails with a `java.util.ConcurrentModificationException` thrown deep inside the dependency graph builder. The same build resolves successfully with **Gradle 9.5.1**.

The failure is deterministic and reproducible. It surfaces both when resolving the configuration as part of a normal build and when simply inspecting it via the `dependencies` report task.

## Affected / unaffected versions

| Gradle version            | Result                            |
| ------------------------- | --------------------------------- |
| 9.5.1                     | ✅ works                          |
| 9.6.0-rc-1                | ❌ `ConcurrentModificationException` |
| 9.7.0 nightly             | ❌ `ConcurrentModificationException` |

The regression was introduced somewhere between 9.5.1 and 9.6.0-rc-1.

## Stack trace

```
Caused by: java.util.ConcurrentModificationException
    at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.ModuleResolveState.clearIncomingUnattachedConstraints(ModuleResolveState.java:417)
    at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.ModuleResolveState.disconnectIncomingEdge(ModuleResolveState.java:400)
    at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.NodeState.disconnectOutgoingEdge(NodeState.java:1229)
    at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.NodeState.removeOutgoingEdges(NodeState.java:1214)
    at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.DependencyGraphBuilder.traverseGraph(DependencyGraphBuilder.java:182)
    at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.builder.DependencyGraphBuilder.resolve(DependencyGraphBuilder.java:160)
    at org.gradle.api.internal.artifacts.ivyservice.resolveengine.DependencyGraphResolver.resolve(DependencyGraphResolver.java:110)
    at org.gradle.api.internal.artifacts.ivyservice.ResolutionExecutor.doResolve(ResolutionExecutor.java:424)
[...]

PR #37330, opened 2026-03-29
Separate capability conflict handling from module conflict handling

jvandort:
Previously, capability conflicts were resolved by routing through module-level APIs (clearSelection/changeSelection), which evicted entire components even when only one variant was involved. This caused a bug where an unrelated variant of a component could not be selected if a sibling variant lost a capability conflict.

This change makes the two conflict types independent:
- Capability conflicts now track and resolve at the node (variant) level via NodeState, using a per-node conflict counter
- Module conflicts use an explicit inConflict flag on ModuleResolveState instead of clearing selection
- ComponentState replaces the ComponentSelectionState enum with a simple boolean evicted flag

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #38055, opened 2026-05-29
Fix ConcurrentModificationException during dependency graph teardown of pending modules

jvandort:
When a module loses all hard incoming edges during graph traversal, it becomes "pending" and all remaining constraint edges targeting it are removed in `clearIncomingAttachedConstraints` and
`clearIncomingUnattachedConstraints`. Previously, removing each constraint edge called `clearSelector`, which called `SelectorState.release()`, which called
`ModuleResolveState.removeSelector()`. `removeSelector` eagerly called `maybeUpdateSelection()` if the module still had remaining selectors and a selected component.

This eager reselection was problematic during the pending teardown. When iterating over unattached constraint edges to remove them, clearing the selector of one constraint could trigger `maybeUpdateSelection` using selectors from a *subsequent* unattached constraint that had not yet been removed. This reselection could call `changeSelection`, which calls `attachUnattachedEdges`, which on attach failure adds edges back to the `unattachedEdges` list — the same list currently being iterated — causing a ConcurrentModificationException.

This reentrant behavior has always existed but was latent. It became a regression after 5e020e42d01 (PR #37330), which added an early return in `EdgeState.getTargetComponent()` for modules in conflict:

```
    if (targetModule.isInModuleConflict()) {
        return null;
    }
```

Previously, edges would still attach to modules in conflict. After this change, `attachToTargetNodes` returns early for modules in conflict, leaving the edge unattached and adding it back to `unattachedEdges`, which triggers the CME during the pending teardown iteration.

The root cause is that `maybeUpdateSelection` should never run on a module that is being deconstructed because it has lost all hard edges. The module is leaving the graph — reselecting based on selectors from constraint edges that are about to be removed is both wasteful and dangerous.

The fix lifts the `maybeUpdateSelection` call out of `removeSelector` and into the caller (`disconnectIncomingEdge`), where we have the context to know whether the module is going back to pending. The key changes:

- `SelectorState.release()` now returns a boolean indicating whether the target module needs reselection, instead of triggering it immediately.
- `ModuleResolveState.removeSelector()` no longer calls `maybeUpdateSelection`. It only updates selector bookkeeping.
- `EdgeState.clearSelector()` returns the boolean from `release()`.
- `ModuleResolveState.disconnectIncomingEdge()` captures the return value and only calls `maybeUpdateSelection` if the module is not going back to pending.

This is safe for all other `clearSelector` call sites:
- `EdgeState.computeSelector`: the edge is unconditionally added to `discoveredEdges`, which triggers selection via `performSelectionSerially`.
- `NodeState.removeOutgoingEdge` (via `disconnectIncomingConstraint`): only called when the module is pending, so selection is unnecessary.

Fixes https://github.com/gradle/gradle/issues/38049

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
@@ -740,15 +740,19 @@ public class NodeState implements DependencyGraphNode {
     }
 
     /**
-     * Determine if this node should be processed when it is dequeued during traversal, or if it
-     * instead be removed from the graph.
+     * Determine if this node should be processed when it is dequeued during traversal, or if its
+     * subgraph should be removed from the graph.
      * <p>
-     * False if this node has no incoming edges, or is in conflict and should temporarily not
-     * contribute to the graph. We need special handling for root since it does not yet have
-     * its own module, but should never be considered a conflict participant.
+     * True if this node has incoming edges and is not in a conflict. We temporarily delay building
+     * subgraphs of nodes in conflict while the conflict has not yet been resolved to avoid subgraphs
+     * of losing nodes from affecting the graph shape.
      */
-    boolean contributesToGraph() {
-        return !isSelected() || isInCapabilityConflict() || (getComponent().getModule().isInModuleConflict() && !isRoot());
+    boolean shouldBuildSubgraph() {
+        return isSelected() &&
+            !isInCapabilityConflict() &&
+            // We need special handling for root since it does not yet have
+            // its own module, but should never be considered a conflict participant.
+            !(getComponent().getModule().isInModuleConflict() && !isRoot());
     }
 
     @Override
--- a/platforms/software/dependency-management/README.md
+++ b/platforms/software/dependency-management/README.md
@@ -219,11 +219,11 @@ continues. Only one conflict is resolved at a time before returning to queue pro
 
 ### Module Conflicts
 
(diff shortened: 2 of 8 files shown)


jvandort:
@bot-gradle test rfr

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
```diff
@@ -747,7 +747,7 @@ void clearTransitiveExclusionsAndEnqueue() {
      * contribute to the graph. We need special handling for root since it does not yet have
      * its own module, but should never be considered a conflict participant.
      */
-    boolean contributesToGraph() {
+    boolean doesNotContributeToGraph() {
```
big-guy: I think the javadoc here is now backwards

big-guy (dismissed):
LGTM

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java
jvandort: Nah it was backwards before this fixed it

jvandort: Ahh the javadoc. Will fix that

ljacomet (approved):

PR #38061, opened 2026-06-01
Add lazy ProcessForkOptions.workingDirectory

asodja:
A second part of https://github.com/gradle/gradle/pull/37801

Done through a different PR due to complexity


asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
```diff
@@ -224,7 +226,16 @@ private Provider<JavaLauncher> createJavaLauncherConvention() {
...
+    public DirectoryProperty getWorkingDirectory() {
+        return forkOptions.getWorkingDirectory();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @Internal
+    @NotToBeReplacedByLazyProperty(because = "Bridge for backward compatibility, use getWorkingDirectory() instead", willBeDeprecated = true)
```
abstratt, line 239: ❓ 

Why are we preserving this getter (while removing most/all others)? I am not opposed to it, just trying to understand if there is a pattern for this exception.

subprojects/core-api/src/main/java/org/gradle/process/ProcessForkOptions.java
```diff
@@ -88,6 +94,16 @@ public interface ProcessForkOptions {
      */
     ProcessForkOptions workingDir(Object dir);
 
+    /**
+     * Returns the working directory for the process as a {@link DirectoryProperty}. Defaults to the project directory.
+     *
+     * @return The working directory property. Must not be unset.
+     * @since 9.7.0
```
abstratt, line 101: ❓ 

Isn't it `@since 10.0`?

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultExecAction.java
```diff
@@ -34,194 +37,227 @@
...
+        if (execSpec.getStandardOutput() != null) {
+            execHandleBuilder.setStandardOutput(execSpec.getStandardOutput());
+        }
+        if (execSpec.getErrorOutput() != null) {
+            execHandleBuilder.setErrorOutput(execSpec.getErrorOutput());
+        }
+        if (execSpec.getExecutable() != null) {
+            execHandleBuilder.setExecutable(execSpec.getExecutable());
+        }
+        if (execSpec.getWorkingDir() != null) {
```
abstratt: ❓ 

Why is do we use the eager (to be deprecated) property here?

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
```diff
@@ -16,21 +16,54 @@
...
     private Map<String, Object> environment;
 
+    /**
+     * Kept for backward compatibility with some external plugins (e.g. KGP). Use {@link #DefaultProcessForkOptions(ObjectFactory, PathToFileResolver)} instead.
+     *
+     * @deprecated Use {@link #DefaultProcessForkOptions(ObjectFactory, PathToFileResolver)} instead.
+     */
+    @Deprecated
     public DefaultProcessForkOptions(PathToFileResolver resolver) {
+        this(resolver, SimplePropertyFactory.directoryProperty((FileResolver) resolver), SimplePropertyFactory.directoryProperty((FileResolver) resolver));
```
abstratt, line 57: ❓ 

how often do we expect this type to be instantiated via KGP (since we are instantiating new factories for every invocation of `SimplePropertyFactory.directoryProperty()`?

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
```diff
@@ -55,26 +88,43 @@ public ProcessForkOptions executable(Object executable) {
...
     }
 
     @Override
     public ProcessForkOptions workingDir(Object dir) {
-        setWorkingDir(dir);
+        if (dir instanceof DirectoryProperty) {
+            getWorkingDirectory().set((DirectoryProperty) dir);
+        } else if (dir instanceof Provider) {
+            getWorkingDirectory().fileProvider(((Provider<?>) dir).map((SerializableTransformer<File, Object>) value -> {
+                if (value instanceof File) {
```
abstratt: ❓ 

Why do we need three branches when we have a `Provider`, but one branch is enough for the non-`Provider` case?

subprojects/core-api/src/main/java/org/gradle/process/ProcessForkOptions.java
```diff
@@ -88,6 +94,16 @@ public interface ProcessForkOptions {
      */
     ProcessForkOptions workingDir(Object dir);
 
+    /**
+     * Returns the working directory for the process as a {@link DirectoryProperty}. Defaults to the project directory.
+     *
+     * @return The working directory property. Must not be unset.
```
abstratt, line 100: ❓ 

Why "must not be unset"? At least `DefaultForkOptions` configures a proper convention, so calling `unset` would revert to the convention value.

subprojects/core-api/src/main/java/org/gradle/process/ProcessForkOptions.java
asodja: We can already merge this for 9.7, no need for wait for Gradle 10, since it's not a breaking change

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultExecAction.java
asodja: Good point, I can fix this

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
asodja: I don't expect they create a lot of new instances. One per javascro[t task that uses this

subprojects/core-api/src/main/java/org/gradle/process/ProcessForkOptions.java
asodja: I used unset because original getter has `Must not be null`. What do you think would make it clearer?

platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java
asodja: This is part of mitigation strategy to rename `File` properties that has short names, e.g. in this case we rename `workingDir` to `workingDirectory`. In https://github.com/gradle/gradle/pull/37801 we rename some other properties.

The idea is, that upgrade `File -> RegularFileProperty/DirectoryProperty` has higher probability that it will break users scripts compared to other upgrades `String -> Property<String>`. So if a property has short name, we decided to rather rename them and they can still use old getters for a short period of time

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
asodja: Simplified

subprojects/core-api/src/main/java/org/gradle/process/ProcessForkOptions.java
asodja: Changed to "Value must be present"

asodja:
@bot-gradle test this

platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/DefaultWorkerExecutor.java
```diff
@@ -132,14 +132,17 @@ public WorkQueue classLoaderIsolation(Action<? super ClassLoaderWorkerSpec> acti
     @Override
     public WorkQueue processIsolation(Action<? super ProcessWorkerSpec> action) {
         DefaultProcessWorkerSpec spec = instantiator.newInstance(DefaultProcessWorkerSpec.class, forkOptionsFactory.newDecoratedJavaForkOptions());
-        File defaultWorkingDir = spec.getForkOptions().getWorkingDir();
+        File defaultWorkingDir = spec.getForkOptions().getWorkingDirectory().get().getAsFile();
+        File specWorkingDir = spec.getForkOptions().getWorkingDirectory().isPresent()
```
big-guy: You'll never see `getWorkingDirectory().isPresent()` false here because the line above it throws if its unset. 

I think what we're trying to achieve here is that only we can set the working directory.

This is what the code was:
```
File defaultWorkingDir = spec.getForkOptions().getWorkingDir();
File workingDirectory = workerDirectoryProvider.getWorkingDirectory();
action.execute(spec);

if (!defaultWorkingDir.equals(spec.getForkOptions().getWorkingDir())) {
    throw new IllegalArgumentException("Setting the working directory of a worker is not supported.");
} else {
    spec.getForkOptions().setWorkingDir(workingDirectory);
}
```
IIUC, this was saying something like "if spec.getForkOptions().getWorkingDir() is changed by the user, blow up, but always use workingDirectory as the working directory". Putting the setWorkingDir in the else makes this look more complicated.

I don't see why we can't do something like this
```
DefaultProcessWorkerSpec spec = instantiator.newInstance(DefaultProcessWorkerSpec.class, forkOptionsFactory.newDecoratedJavaForkOptions());
File workingDirectory = workerDirectoryProvider.getWorkingDirectory();
spec.getForkOptions().getWorkingDirectory().convention(workingDirectory);
spec.getForkOptions().getWorkingDirectory().disallowChanges();
action.execute(spec);
return instantiator.newInstance(DefaultWorkQueue.class, this, spec, daemonWorkerFactory);
```

We lose the nicer message, but I think it's the same behavior.

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultExecSpec.java
```diff
@@ -34,8 +35,8 @@ public class DefaultExecSpec extends DefaultProcessForkOptions implements ExecSp
     private final ProcessArgumentsSpec argumentsSpec = new ProcessArgumentsSpec(this);
 
     @Inject
-    public DefaultExecSpec(PathToFileResolver resolver) {
-        super(resolver);
+    public DefaultExecSpec(ObjectFactory objectFactory, PathToFileResolver resolver) {
```
big-guy, line 39: Does method injection work for these classes? Instead of having to wire new services everywhere, could you just add a protected abstract getter?

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
```diff
@@ -55,26 +89,43 @@ public ProcessForkOptions executable(Object executable) {
...
+            getWorkingDirectory().fileProvider(((Provider<?>) dir).map((SerializableTransformer<File, Object>) value -> {
+                if (value instanceof FileSystemLocation) {
+                    return resolver.resolve(((FileSystemLocation) value).getAsFile());
+                } else {
+                    return resolver.resolve(value);
+                }
+            }));
+        } else {
+            getWorkingDirectory().set(resolver.resolve(dir));
+        }
```
big-guy, line 128: Slightly less gross might be to cast to PropertyInternal and call setFromAnyValue(dir)

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
```diff
@@ -55,26 +89,43 @@ public ProcessForkOptions executable(Object executable) {
...
+    @Override
+    public File getWorkingDir() {
+        return getWorkingDirectory().get().getAsFile();
+    }
+
     @Override
     public void setWorkingDir(File dir) {
-        this.workingDir = resolver.resolve(dir);
+        // We call resolver to resolve "." in some scopes
+        getWorkingDirectory().set(resolver.resolve(dir));
```
big-guy, line 104: fileValue doesn't work here?

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
big-guy: Can you create an issue against Kotlin and a reminder for us to properly deprecate with nagging?

subprojects/core-api/src/main/java/org/gradle/process/ProcessForkOptions.java
```diff
@@ -88,6 +94,16 @@ public interface ProcessForkOptions {
...
+     * @since 9.7.0
+     */
+    @Incubating
+    @Internal
+    DirectoryProperty getWorkingDirectory();
```
big-guy, line 105: I'm not sure if we have XML docs for this type that need to be updated, but please check.

big-guy (changes requested):

asodja:
@bot-gradle test this

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
asodja: It seems that `PropertyInternal. setFromAnyValue()` doesn't handle all cases as resolver.resolve does. It handles just String, File and Provider

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultProcessForkOptions.java
asodja: We need to manually resolve, otherwise in some cases we get:
```
  java.lang.UnsupportedOperationException: Cannot convert relative path . to an absolute file.
      at org.gradle.api.internal.file.IdentityFileResolver.doResolve(IdentityFileResolver.java:30)
      ...
 ```
 
 shown by our Gradleception tests

subprojects/core-api/src/main/java/org/gradle/process/ProcessForkOptions.java
asodja: Good catch. Added to xml

platforms/core-runtime/process-services/src/main/java/org/gradle/process/internal/DefaultExecSpec.java
asodja: In Global/Worker scope these classes are not decorated, so we can't use abstract getter until we fix that there. But that is a bigger change I wouldn't like to do as part of this PR

platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/DefaultWorkerExecutor.java
asodja: Makes sense, applied suggested changes.

Note that there is no `convention(File)` method, so I just used set with disallowChanges. Result should be equal in the end

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

big-guy (approved):

asodja:
@bot-gradle test this

PR #38063, opened 2026-06-01
Add dangerously-ignore-problems for IP evaluation

### LINKED ISSUE #38040: Allow evaluating expected performance of IP easily
Making a build IP-compatible is not trivial. We would like for users to be able to easily run a build or sync in parallel and get a rough estimate of the speedup they can expect, without worrying about fixing all of the correctness issues first.

To accomplish this, we should add a third mode for IP, "unsafe"/"evaluation"/??? mode. This mode must allow the build to succeed (at least under TAPI/sync) to allow evaluating IDE sync speeds properly, but it should be EXTREMELY LOUD about this not being a production mode that will produce correct outputs, and it may even crash unexpectedly.

The other modes:
- Diagnostics: https://github.com/gradle/gradle/issues/34799
- Fail-fast: https://github.com/gradle/gradle/issues/37936

octylFractal:
Fixes https://github.com/gradle/gradle/issues/38040

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIgnoreProblemsIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIgnoreProblemsIntegrationTest.groovy
@@ -34,11 +34,11 @@ class IsolatedProjectsIgnoreProblemsIntegrationTest extends AbstractIsolatedProj
     def "cross-project access succeeds and shows the banner instead of failing"() {
         given:
         createDirs("a", "b")
-        settingsFile << """
+        settingsFile """
             include("a")
             include("b")
         """
-        buildFile << """
+        buildFile """
             allprojects {
                 plugins.apply('java-library')
             }
@@ -59,11 +59,11 @@ class IsolatedProjectsIgnoreProblemsIntegrationTest extends AbstractIsolatedProj
     def "cross-project access in the Kotlin DSL succeeds and shows the banner"() {
         given:
         createDirs("a", "b")
-        settingsKotlinFile << """
+        settingsKotlinFile """
             include("a")
             include("b")
         """
-        buildKotlinFile << """
+        buildKotlinFile """
             allprojects {
                 plugins.apply("java-library")
             }
@@ -84,11 +84,11 @@ class IsolatedProjectsIgnoreProblemsIntegrationTest extends AbstractIsolatedProj
     def "discarded entry is not reused: a second run reconfigures and re-reports"() {
         given:
         createDirs("a", "b")
-        settingsFile << """
+        settingsFile """
             include("a")
             include("b")
         """
-        buildFile << """
+        buildFile """
             allprojects {
                 plugins.apply('java-library')
             }
@@ -119,11 +119,11 @@ class IsolatedProjectsIgnoreProblemsIntegrationTest extends AbstractIsolatedProj
     def "is orthogonal to diagnostics: combined run still succeeds"() {
         given:
         createDirs("a", "b")
-        settingsFile << """
+        settingsFile """
             include("a")
             include("b")
         """
-        buildFile << """
+        buildFile """
             allprojects {
                 plugins.apply('java-library')
             }
@@ -143,8 +143,8 @@ class IsolatedProjectsIgnoreProblemsIntegrationTest extends AbstractIsolatedProj
 
     def "a clean build shows the banner but stores and reuses a normal cache entry"() {
         given:
-        settingsFile << "rootProject.name = 'root'"
-        buildFile << "task ok {}"
+        settingsFile "rootProject.name = 'root'"
+        buildFile "task ok {}"
 
         when:
         isolatedProjectsSucceedsIgnoringProblems("ok")
@@ -193,11 +193,11 @@ class IsolatedProjectsIgnoreProblemsIntegrationTest extends AbstractIsolatedProj
     def "a genuine task failure still fails the build while the IP violation is only reported"() {
         given:
         createDirs("a", "b")
-        settingsFile << """
+        settingsFile """
             include("a")
             include("b")
         """
-        buildFile << """
+        buildFile """
             allprojects {
                 plugins.apply('java-library')
             }
@@ -226,8 +226,8 @@ class IsolatedProjectsIgnoreProblemsIntegrationTest extends AbstractIsolatedProj
 
     def "the flag is inert when Isolated Projects is not enabled"() {
         given:
-        settingsFile << "rootProject.name = 'root'"
-        buildFile << "task ok { doLast {} }"
+        settingsFile "rootProject.name = 'root'"
+        buildFile "task ok { doLast {} }"
 
         when:
         run("-D${StartParameterBuildOptions.IsolatedProjectsDangerouslyIgnoreProblemsOption.PROPERTY_NAME}=true", "ok")
@@ -238,14 +238,14 @@ class IsolatedProjectsIgnoreProblemsIntegrationTest extends AbstractIsolatedProj
         postBuildOutputDoesNotContain(DANGEROUSLY_IGNORE_PROBLEMS_BANNER)
     }
 
-    def "combining the flag with problems=warn still discards the entry"() {
+    def "combining the flag with CC warn-mode still discards the entry"() {
         given:
         createDirs("a", "b")
-        settingsFile << """
+        settingsFile """
             include("a")
             include("b")
         """
-        buildFile << """
+        buildFile """
             allprojects {
                 plugins.apply('java-library')
             }
@@ -267,14 +267,42 @@ class IsolatedProjectsIgnoreProblemsIntegrationTest extends AbstractIsolatedProj
         }
     }
 
-    def "a configuration cache problem still fails the build while IP violations are ignored"() {
+    def "IP violations still fail the build in CC warn-mode when the flag is not enabled [#mode]"() {
         given:
         createDirs("a", "b")
-        settingsFile << """
+        settingsFile """
             include("a")
             include("b")
         """
-        buildFile << """
+        buildFile """
+            allprojects {
+                plugins.apply('java-library')
+            }
+        """
+
+        when: "CC warn-mode must not suppress IP violations without dangerously-ignore-problems"
+        withIsolatedProjectsUsing(mode, WARN_PROBLEMS_CLI_OPT)
+        fails("help")
+
+        then:
+        outputDoesNotContain(DANGEROUSLY_IGNORE_PROBLEMS_BANNER)
+        fixture.assertIsolatedProjectsProblems(mode) {
+            projectsConfigured(":", ":a", ":b")
+            problem(GROOVY_VIOLATION, 2)
+        }
+
+        where:
+        mode << ALL_MODES
+    }
+
+    def "a configuration cache problem still fails the build while IP violations are ignored"() {
+        given:
+        createDirs("a", "b")
+        settingsFile """
+            include("a")
+            include("b")
+        """
+        buildFile """
             allprojects {
                 plugins.apply('java-library')
             }
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIgnoreProblemsIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIgnoreProblemsIntegrationTest.groovy
@@ -1,0 +1,268 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl.isolated
+
+import org.gradle.initialization.StartParameterBuildOptions
+import org.gradle.util.internal.ToBeImplemented
+
+/**
+ * Tests for the "dangerously ignore problems" IP mode, where IP violations are reported but do not
+ * fail the build, so a parallel build or sync can be timed before the violations are fixed.
+ *
+ * <p>Unlike the fail-fast and diagnostics modes (see {@link IsolatedProjectsAccessIntegrationTest}),
+ * the build is expected to <em>succeed</em> here. The entry is still stored and then discarded
+ * because problems were reported, so it is never reused by a later build.
+ */
+class IsolatedProjectsIgnoreProblemsIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
+
+    private static final String GROOVY_VIOLATION = "Build file 'build.gradle': line 3: Project ':' cannot access 'Project.plugins' functionality on subprojects via 'allprojects'"
+    private static final String KOTLIN_VIOLATION = "Build file 'build.gradle.kts': Project ':' cannot access 'Project.plugins' functionality on subprojects via 'allprojects'"
+
+    def "cross-project access succeeds and shows the banner instead of failing"() {
+        given:
+        createDirs("a", "b")
+        settingsFile << """
+            include("a")
+            include("b")
+        """
+        buildFile << """
+            allprojects {
+                plugins.apply('java-library')
+            }
+        """
+
+        when:
+        isolatedProjectsSucceedsIgnoringProblems("help")
+
+        then:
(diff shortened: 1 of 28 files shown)


platforms/core-runtime/base-services/src/main/java/org/gradle/internal/buildtree/BuildModelParameters.java
```diff
@@ -99,6 +99,18 @@ default boolean isVintage() {
      */
     boolean isIsolatedProjectsDiagnostics();
 
+    /**
+     * Whether the IP "dangerously ignore problems" mode is enabled.
+     * <p>
+     * In this mode, IP violations are reported but do not fail the build, so a parallel build or
+     * sync can be timed to estimate the speedup before the violations are fixed. Outputs may be
+     * incorrect and the build may crash. It can be set alongside IP Diagnostics mode, which may
```
alllex: 💭 "parallel build" trips me up every time I read this, because I think of another build running next to the current one. I think we use the same terminology for some user-facing messages, and I think we should either change the wording or simplify it.

I think we won't lose anything if we simply say `so it can be timved to estimate the speedup`.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiInvocationValidationIntegrationTest.groovy
```diff
@@ -17,6 +17,27 @@
 package org.gradle.internal.cc.impl.isolated
 
 class IsolatedProjectsToolingApiInvocationValidationIntegrationTest extends AbstractIsolatedProjectsToolingApiIntegrationTest {
+    def "dangerously ignoring problems lets a sync with cross-project access succeed"() {
```
alllex: 💅 I'd reserve "sync" wording for actual sync tests. For these tests we have been using "fetching tooling model" terminology, which is more accurate.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/ConfigurationCacheStartParameter.kt
```diff
@@ -191,6 +191,18 @@ class ConfigurationCacheStartParameter internal constructor(
     val isIsolatedProjectsDiagnostics: Boolean
         get() = modelParameters.isIsolatedProjectsDiagnostics
 
+    /**
+     * Determines whether Isolated Projects "dangerously ignore problems" option is enabled.
+     *
+     * With this option, IP violations are reported but do not fail the build, so a parallel build or sync
```
alllex: 💅 "parallel build"

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
```diff
@@ -303,7 +320,8 @@ class ConfigurationCacheProblems(
                 trace(problem.trace.containingUserCode)
             }
         }.also {
-            if (severity == ProblemSeverity.Interrupting || (severity == ProblemSeverity.Deferred && !isWarningMode)) {
+            // IP-deferred problems always report as errors; CC-deferred problems only when not in warn mode.
```
alllex: 🤔 The intention of the `reportError` is that errors actually fail the build. However, in the IP danger-more we are not failing the build. So I wonder if we should be reporting it as a regular non-error problem instead.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
```diff
@@ -66,6 +66,19 @@ import java.io.File
...
+val DANGEROUSLY_IGNORE_PROBLEMS_BANNER = """
+
+==============================================================================
+Isolated Projects: dangerously-ignore-problems is ENABLED. Isolated Projects
+violations are being ignored, so build outputs may be incorrect and the build
+may crash unexpectedly. Use this only to estimate parallel/sync speedup, never
+to produce artifacts.
+==============================================================================
+
+""".trimIndent()
```
alllex: ❔ Do we need `trimIndent()`? If we remove it, we can make the property `const`

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
```diff
@@ -66,6 +66,19 @@ import java.io.File
 import java.io.IOException
 
 
+private
+val DANGEROUSLY_IGNORE_PROBLEMS_BANNER = """
+
+==============================================================================
+Isolated Projects: dangerously-ignore-problems is ENABLED. Isolated Projects
+violations are being ignored, so build outputs may be incorrect and the build
+may crash unexpectedly. Use this only to estimate parallel/sync speedup, never
```
alllex: 💭 I don't think it will be clear to users what "parallel speedup" means. We'll also continue evolving IP, so the parallel part might become one of multiple ways the performance is better.

I'd suggest going a bit more neutral: `Use this only to evaluate performance.`

I think it would also be a nice touch if the last line is a clear sentence -- `Do not use this to produce artifacts.` My assumption is that once people have seen the banner (even just once), they won't be reading it anymore in full ever, so the last line in the message will be the closest to the end of the build. It would be nicer if it gives a self-contained advice.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
```diff
@@ -255,26 +271,27 @@ class ConfigurationCacheProblems(
...
 
     override fun onProblem(problem: PropertyProblem) {
         onProblem(problem, ProblemSeverity.Deferred)
     }
 
     private
-    fun onProblem(problem: PropertyProblem, severity: ProblemSeverity) {
-        if (summarizer.onProblem(problem, severity)) {
-            problemsService.onProblem(problem, severity)
+    fun onProblem(problem: PropertyProblem, severity: ProblemSeverity, isolated: Boolean = false) {
```
alllex: 💅 I think `isolatedProjects` would be clearer

platforms/core-configuration/configuration-cache/src/test/groovy/org/gradle/internal/buildtree/BuildModelParametersProviderTest.groovy
```diff
@@ -40,6 +40,7 @@ class BuildModelParametersProviderTest extends Specification {
 
             isolatedProjects: false,
             isolatedProjectsDiagnostics: false,
+            isolatedProjectsDangerouslyIgnoreProblems: false,
```
alllex, line 43: ❌ We should have some tests that assert enabling this option

platforms/core-configuration/configuration-cache/src/test/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheKeyTest.kt
```diff
@@ -124,6 +124,37 @@ class ConfigurationCacheKeyTest {
         )
     }
 
+    @Test
+    fun `cache key honours dangerously ignore problems option`() {
```
alllex: 💅 clarify it's about IP

alllex (changes requested):
Looks good overall. Also great to cut the ties with CC warning mode!

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
octylFractal: I am not quite sure what you mean by this? In danger-mode we use `Suppressed`, so it is not an "IP-deferred" problem, and therefore is already sent to `report`, not `reportError`. I can word this comment better.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
octylFractal: I went with `forIsolatedProjects`, to be even clearer that it's not about wider enablement.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
octylFractal: I reworded it.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiInvocationValidationIntegrationTest.groovy
```diff
@@ -17,6 +17,27 @@
...
+        """
+
+        when:
+        withIsolatedProjectsDangerouslyIgnoreProblems()
+        def model = fetchModel()
+
+        then:
+        model != null
+        outputContains(DANGEROUSLY_IGNORE_PROBLEMS_BANNER)
+        outputContains("Project ':' cannot access 'Project.plugins' functionality on subprojects via 'allprojects'")
```
alllex: 💅 We should assert problems via the fixture

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
alllex: I missed the step where we changed up the severity for IP. I think the rewording helps too 👍

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
```diff
@@ -66,6 +66,19 @@ import java.io.File
 import java.io.IOException
 
 
+private
+const val DANGEROUSLY_IGNORE_PROBLEMS_BANNER = """
```
alllex: ❔ Should the property name be clear that it is IP-specific?

Would it be useful to link to something like `BuildModelParameters` in the doc string?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
```diff
@@ -66,6 +66,19 @@ import java.io.File
...
+
+==============================================================================
+Isolated Projects: dangerously-ignore-problems is ENABLED. Isolated Projects
+violations are being ignored, so build outputs may be incorrect and the build
+may crash unexpectedly. Use this only to evaluate performance.
+Do not use this to produce artifacts.
+==============================================================================
```
alllex, line 83: 💭 Would it be useful to link to the userguide section about the danger mode?

platforms/core-runtime/start-parameter/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java
```diff
@@ -704,6 +705,19 @@ public void applyTo(boolean value, StartParameterInternal settings, Origin origi
         }
     }
 
+    public static class IsolatedProjectsDangerouslyIgnoreProblemsOption extends BooleanBuildOption<StartParameterInternal> {
+        public static final String PROPERTY_NAME = "org.gradle.unsafe.isolated-projects.dangerously-ignore-problems";
```
alllex, line 709: ❌ We are missing a userguide section on this

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIgnoreProblemsIntegrationTest.groovy
```diff
@@ -0,0 +1,295 @@
...
+ */
+class IsolatedProjectsIgnoreProblemsIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
+
+    private static final String GROOVY_VIOLATION = "Build file 'build.gradle': line 3: Project ':' cannot access 'Project.plugins' functionality on subprojects via 'allprojects'"
+    private static final String KOTLIN_VIOLATION = "Build file 'build.gradle.kts': Project ':' cannot access 'Project.plugins' functionality on subprojects via 'allprojects'"
+
+    def "cross-project access succeeds and shows the banner instead of failing"() {
+        given:
+        createDirs("a", "b")
+        settingsFile << """
```
alllex: 💅 Drop `<<` to get syntax highlight in all such blocks

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIgnoreProblemsIntegrationTest.groovy
```diff
@@ -0,0 +1,295 @@
...
+        when:
+        run("-D${StartParameterBuildOptions.IsolatedProjectsDangerouslyIgnoreProblemsOption.PROPERTY_NAME}=true", "ok")
+
+        then:
+        result.assertTaskExecuted(":ok")
+        outputDoesNotContain(DANGEROUSLY_IGNORE_PROBLEMS_BANNER)
+        postBuildOutputDoesNotContain(DANGEROUSLY_IGNORE_PROBLEMS_BANNER)
+    }
+
+    def "combining the flag with problems=warn still discards the entry"() {
```
alllex:
```suggestion
    def "combining the flag with CC warn-mode still discards the entry"() {
```

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIgnoreProblemsIntegrationTest.groovy
```diff
@@ -0,0 +1,295 @@
...
+            include("b")
+        """
+        buildFile << """
+            allprojects {
+                plugins.apply('java-library')
+            }
+        """
+
+        when:
+        run(ENABLE_CLI, ENABLE_DANGEROUSLY_IGNORE_PROBLEMS, WARN_PROBLEMS_CLI_OPT, "help")
```
alllex, line 255: ❓ Do we have a test that demostrates IP problems now fail the build in CC warm-mode, when IP danger mode is NOT enabled?

alllex (changes requested):

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIgnoreProblemsIntegrationTest.groovy
octylFractal: We had one using diagnostics mode, but I improved it in the next batch.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -557,42 +557,49 @@ A build that is clean under Isolated Projects is also clean under _Diagnostics_
...
-
-[source,bash]
-----
-$ ./gradlew --configuration-cache-problems=warn
-----
-
 [WARNING]
 ====
-Warning mode is a migration and troubleshooting aid and not intended as a persistent way of ignoring incompatibilities.
+This mode is a migration and troubleshooting aid and not intended as a persistent way of ignoring incompatibilities.
```
alllex: 💭 Very minor nit: Having two callouts one after another might be too concentrated. I wonder if we should move the Warning to the top of the section to set the expectations immediately?

Alternatively, we could move the note right after the `gradle.properties` snippet.

alllex (approved):
LGTM! Great work 👍

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
lkasso: +1 Two callouts, one after the other, is not ideal

lkasso (approved):
Docs LGTM. I added some changes to a few .adoc files and started the release notes. Please feel free to update.

blindpirate:
This is consistently failing: https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_SmokeTest_GradleBuildSmokeTests_virtual_Batch_2_1/113723510

octylFractal:
@bot-gradle test AST without PTS

PR #2213, opened 2017-05-31
Fix Invalid cookie header messages

wolfs:
When downloading from some repositories like the oracle maven repository
httpclient is not able to parse the expires attribute in the set cookie
header.
This looks like using the `standard` cookie spec instead of the default
one fixes the problem. It is also necessary to upgrade httpclient to
the latest version to make this work for
https://issues.apache.org/jira/browse/HTTPCLIENT-1077.

This would fix #1596.

PR #38068, opened 2026-06-02
Fix flaky IsolatedProjectsToolingApiParallelConfigurationIntegrationTest

blindpirate:
## Problem

`IsolatedProjectsToolingApiParallelConfigurationIntegrationTest` > _"projects are configured and models created in parallel when project scoped model is queried concurrently"_ fails intermittently, observed on macOS Aarch64 (Java 25, configuration-cache):

```
org.gradle.tooling.BuildActionFailureException: The supplied build action failed with an exception.
...
Failed to handle GET /model-root due to a timeout waiting for other requests.
Waiting for 1 further requests, received [GET /model-root, GET /configure-b],
released [], not yet received [GET /configure-a]
```

### Root cause

The test uses `BlockingHttpServer` with `expectConcurrent("model-root", "configure-a", "configure-b")`, which blocks each received request until **all three** have arrived (120s timeout). Each `callFromBuild` request occupies a Gradle worker thread for the duration of the block, so the build needs **at least 3 worker threads** to satisfy the expectation.

Unlike the standard `ParallelForkingGradleExecuter` (which injects `--max-workers=4`), this test runs via `ToolingApiBackedGradleExecuter`, which applies **no** `--max-workers` default. Worker count therefore falls back to `Runtime.getRuntime().availableProcessors()`. On a CI agent that reports only 2 usable processors, only 2 of the 3 requests get a thread; they block waiting for the third, which can never be scheduled → 120s timeout → flaky failure.

The sibling `IsolatedProjectsParallelConfigurationIntegrationTest` already documents and handles this: _"this blocks the worker thread, hence max-workers is based on # of projects"_ and _"max-workers must be >= # of projects or else we run out of workers"_.

## Fix

Pin `--max-workers=4` on each `withIsolatedProjects(...)` invocation in the affected test (peak concurrency is 3, +1 headroom; matches the standard integ-test default). This removes the dependency on the agent's reported core count.

## Evidence

- Failing build: macOS Aarch64, Java 25, configuration-cache (`Gradle_Master_Check_Platform_36_bucket2`), failing since build #2213
- Representative error: `not yet received [GET /configure-a]` after a 120s concurrent-request timeout

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiParallelConfigurationIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiParallelConfigurationIntegrationTest.groovy
@@ -27,6 +27,38 @@ class IsolatedProjectsToolingApiParallelConfigurationIntegrationTest extends Abs
     @Rule
     BlockingHttpServer server = new BlockingHttpServer()
 
+    // DEBUG (flaky investigation): init script that starts a daemon-side watchdog which periodically
+    // dumps all Gradle thread stacks together with lock ownership, so we can see what the :a/:b
+    // configuration threads are blocked on while the model-root request is held. Every line is prefixed
+    // with @@FLAKYDUMP@@ so it can be grepped out of the captured build output. Remove before merging.
+    private static final String WATCHDOG_INIT_SCRIPT = $/
+import java.lang.management.ManagementFactory
+
+Thread watchdog = new Thread({
+    def bean = ManagementFactory.threadMXBean
+    for (int i = 0; i < 14; i++) {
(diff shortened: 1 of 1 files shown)


platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiParallelConfigurationIntegrationTest.groovy
```diff
@@ -27,6 +27,38 @@ class IsolatedProjectsToolingApiParallelConfigurationIntegrationTest extends Abs
...
+        System.out.println(sb.toString())
+        System.out.flush()
+    }
+})
+watchdog.setName("flaky-debug-watchdog")
+watchdog.setDaemon(true)
+watchdog.start()
+/$
+
     def setup() {
```
6hundreds, line 62: ❌ I don't think this is intended to be a part of the PR

PR #38077, opened 2026-06-02
Add feature preview for enhanced graph orderings

### LINKED ISSUE #29275: Classpath order is unexpected
### Current Behavior

I'm reporting this as bug in case it is a code issue, but it could also be that just documentation should make it clear.

Version constraints influence the classpath order, as do the concrete configurations dependencies are added to.
To me this comes unexpected.
From the past I was under the impressions that the order of classpaths should be controlled by order of declaration in the build script.

If it "works as intended", I'd at least want to ask for clarifying the situation explicitly in the documentation. At least I did not find anything about it.

If you for example have
```kts
dependencies {
    runtimeOnly("saxon:saxon:6.5.3")
    implementation("net.sourceforge.nekohtml:nekohtml:1.9.22")
}
```
and do
```java
System.out.println(SAXParserFactory.newInstance().newSAXParser());
```
You get Aelfred which is shipped with Saxon 6.
But if you do
```kts
dependencies {
    constraints {
        implementation("xerces:xercesImpl:2.12.2")
    }
    runtimeOnly("saxon:saxon:6.5.3")
    implementation("net.sourceforge.nekohtml:nekohtml:1.9.22")
}
```
you "suddenly" get Xerces.

Actually using
```kts
val foo by tasks.registering {
    doLast {
        configurations.runtimeClasspath.get().forEach { println(it) }
    }
}
```
to output the runtime classpath also shows that the order is not in declaration order anyway.
Without the constraint you
- first have neko,
- then saxon,
- then the neko dependency xerces,
- then the xerces dependency xml-apis

With the constraint you
- first have neko,
- then xerces,
- then saxon,
- then xml-apis

So it seems constraint is enough so that a dependency comes first, and direct `implementation` dependencies come before direct `runtimeOnly` dependencies and transitive dependencies for which no constraint exists come after direct ones.

Of course relying on classpath order is bad in most situations and it is usually better to make sure classpaths do not contain classes in multiple jars, or select services explicitly if possible, but sometimes this is just not feasible and manual classpath sorting is also not always possible. In my concrete case the alternative would probably be an artifact transform that removes the service description from the Saxon 6 dependency, but that would imho be quite hacky. :-D

I need the Xerxes parser for parsing, I need Saxon 6 for docbook xsl 1 processing. Saxon has Aelfred parser shipped in the same jar. Manual ordering is not possible as it is in `buildSrc` dependencies. Setting system property in `gradle.properties` to explicitly select the parser is also not possible as then the settings script fails as it also parses an XML file and there the built-in parser needs to be used.
In the end I just moved the dependencies around so that the resulting order works as I need it in this case.

### Expected Behavior

More intuitive classpath ordering, derived from declaration order but independent from actual configuration the dependencies are declared on and version constraints not influencing order.

Alternatively some documentation that clearly documents how classpaths are ordered, or what influences it and so on.

Well, documentation should be added in either case imho.

### Context (optional)

_No response_

### Steps to Reproduce

see above

### Gradle version

8.7

### Build scan URL (optional)

_No response_

_No response_

jvandort:
Replace the two-phase graph visitor contract (`visitNode` + `visitEdges`) with a single `visitNode` pass, and change the traversal to a BFS-biased topological sort using Kahn's algorithm.

BFS ordering prioritizes nodes closer to the root, placing them earlier on the classpath. When this graph represents a JVM classpath, these nodes are likely used earlier and more often, so ordering them first may improve classloading performance.

Constraint edges are excluded from in-degree computation and traversal, so constraints no longer affect the shape of the visited graph. This was the main motivation for the change. Cycles are broken by falling back to the earliest discovered unvisited node, preserving a BFS-like order in cyclic subgraphs.

The `visitEdges` method is removed from `DependencyGraphVisitor` since all visitor implementations now only require a single pass. The `VisitState` enum and associated mutable state on `ComponentState` are also removed.

Fixes https://github.com/gradle/gradle/issues/29275

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/DependencyGraphVisitor.java
```diff
@@ -22,24 +22,20 @@
...
-     * Visits a node of the graph. Includes the root. This method is called for all nodes before {@link #visitEdges(DependencyGraphNode)} is called.
+     * Visits a node of the graph, including the root. Nodes are visited in consumer-first order
      */
     default void visitNode(DependencyGraphNode node) {}
 
-    /**
-     * Visits edges to/from a node of the graph. Includes the root. This method is called for all nodes after {@link #visitNode(DependencyGraphNode)} has been called for all nodes.
-     * Nodes are visited in consumer-first order.
-     */
-    default void visitEdges(DependencyGraphNode node) {}
```
jvandort, line 39: Previous `DependencyGraphVisitor`s required a two-pass visit. Recent changes have updated visitors to no longer require two passes, meaning we can remove this second method. Both `visitNode` and `visitEdges` visited each node in the graph, but in different orders. This is no longer necessary.

jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

lkasso (approved):
Docs LGTM

jvandort:
TODO:
- [x] Update upgrade guide
- [x] Add tests for new sorting with feature flag enabled
- [ ] Consider other names for feature flag

lkasso (approved):
Docs LGTM.

subprojects/core/src/main/java/org/gradle/api/internal/FeaturePreviews.java
```diff
@@ -49,7 +49,17 @@ public enum Feature implements FeatureFlag {
         /**
          * This exists to test inactive feature previews
          */
-        ALWAYS_INACTIVE(false, null);
+        ALWAYS_INACTIVE(false, null),
+        /**
+         * Whether to use improved dependency graph sort ordering. With this feature enabled:
```
big-guy, line 54: Add a link to the upgrade guide section too

big-guy (changes requested):
one small nit

big-guy (approved):

Issue #38073, opened 2026-06-02
Gradle 9.6.0 RC1 incompatible with Android Gradle Plugin 8.13.2


dglynch:
### Current Behavior

In Gradle 9.6.0 RC1, when I try to build a project using Android Gradle Plugin 8.13.2, every task fails with the following error:

```
FAILURE: Build failed with an exception.

* Where:
Build file 'C:\dev\Agp8132Gradle960Rc1\app\build.gradle.kts' line: 1

* What went wrong:
An exception occurred applying plugin request [id: 'com.android.application', version: '8.13.2']
> Failed to apply plugin 'com.android.internal.application'.
   > Failed to create service 'com.android.build.gradle.internal.errors.AndroidProblemReporterProvider_d67adb98-0a4b-40ca-ac42-1413f05d389c'.
      > Could not create an instance of type com.android.build.gradle.internal.errors.AndroidProblemReporterProvider.
         > Could not generate a decorated class for type AndroidProblemReporterProvider.
            > org/gradle/api/problems/internal/InternalProblems

```

### Expected Behavior

In Gradle 9.5.1 and earlier, the Android Gradle Plugin 8.13.2 works as expected.

### Context (optional)

Android Gradle Plugin 8.13.2 is not the latest version, but it is very recent and many projects are still using it. The next major version (9.x) introduces many breaking changes, so the old version should still be supported by Gradle for a much longer time.

### Self-contained Reproducer Project

https://github.com/dglynch/Agp8132Gradle960Rc1

### Gradle version

9.6.0 RC1

### Gradle version that used to work

9.5.1

ov7a:
This issue needs a decision from the team responsible for that area. They have been informed. Response time may vary.

***

Note: AGP uses an internal API, hence the unexpected failure
> org/gradle/api/problems/internal/InternalProblems

If we decide to not to fix it, we should still update the upgrade guide to highlight the incompatibility.

eskatos:
AGP 8.x itself only supports being used with Gradle 8.x.
We kept supporting it until 9.5 as a best effort but we need to move on and the only reason it breaks is because AGP was wrongly using our internal APIs.
You need to stick with Gradle 9.5 or upgrade to AGP 9.x.

This is actually in the upgrade guide at https://docs.gradle.org/9.6.0-rc-1/userguide/upgrading_version_9.html#agp_8x_incompatible

eskatos:
[...]

PR #38078, opened 2026-06-02
Report actionable error for removed Problems API

eskatos:
The internal type org.gradle.api.problems.internal.InternalProblems was removed in 9.6.0. Plugins that bound to it directly, most notably AGP 8.x, then fail with an opaque class loading error that explains neither the cause nor the remedy. Surface a clear failure instead: name the plugin at fault and point to the upgrade guide so users know to update the plugin or remain on Gradle 9.5.

* Fixes https://github.com/gradle/gradle/issues/38073

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/ide/problems/src/main/java/org/gradle/problems/internal/services/ProblemsBuildTreeServices.java
+++ b/platforms/ide/problems/src/main/java/org/gradle/problems/internal/services/ProblemsBuildTreeServices.java
@@ -17,6 +17,8 @@
 package org.gradle.problems.internal.services;
 
 import com.google.common.collect.ImmutableList;
+import org.gradle.api.GradleException;
+import org.gradle.api.internal.DocumentationRegistry;
 import org.gradle.api.internal.StartParameterInternal;
 import org.gradle.api.internal.file.temp.TemporaryFileProvider;
 import org.gradle.api.problems.internal.DefaultProblems;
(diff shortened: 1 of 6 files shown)


mlopatkin (approved):
Nice trick

platforms/ide/problems/src/main/java/org/gradle/problems/internal/services/ProblemsBuildTreeServices.java
```diff
@@ -123,4 +126,27 @@ ProblemReportCreator createProblemsReportCreator(
...
+    @Deprecated
+    @Provides
+    org.gradle.api.problems.internal.InternalProblems createInternalProblems(UserCodeApplicationContext userCodeContext) {
+        UserCodeApplicationContext.Application current = userCodeContext.current();
+        String culprit;
+        if (current == null) {
+            culprit = "A plugin you are using";
+        } else {
+            String displayName = current.getSource().getDisplayName().getDisplayName();
+            culprit = Character.toUpperCase(displayName.charAt(0)) + displayName.substring(1);
```
reinsch82, line 144: 💅 

```suggestion
            culprit = StringUtils.capitalize(displayName);
```

reinsch82 (commented):
LGTM

PR #38079, opened 2026-06-02
Emit build operation events to JFR

hegyibalint:
### Context

Build operations are Gradle's own unit of "what the build is doing," but there was no first-class way to see them on a JFR timeline. This emits each build operation as a JFR event into whatever recording is active, so you can profile a build with `-XX:StartFlightRecording` and see Gradle's operations in JDK Mission Control (or `jfr print`).

**What it adds**

- `BuildOperationJfrEmitter`, a sibling of `BuildOperationTrace` in `:build-operations-trace`, enabled with `-Dorg.gradle.internal.operations.jfr=true` (mirroring the file-based `org.gradle.internal.operations.trace`). It emits an `org.gradle.BuildOperation` event per build operation while *any* recording is active, with near-zero cost otherwise (no listener is registered when the option is off).
- Each event carries the operation id, parent id, display name, failure info, and **Gradle's own clock** at start and finish. That lets JFR's JVM-clock `startTime`/`duration` be compared against Gradle's clock to diagnose timing drift. An integration test cross-checks every shared field against the `BuildOperationTrace`.
- The emitter is a `CrossBuildSession` (daemon) service, so its JFR code (Java 11+) is never loaded on the Java 8 client or worker, even though `:build-operations-trace` also targets them.

This PR only **emits** events into an existing recording. Gradle starting and managing its own recording (`org.gradle.internal.tracing.jfr.managed`) is a separate, stacked follow-up.

### DIFF
--- a/platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationJfrEmitterIntegrationTest.groovy
+++ b/platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationJfrEmitterIntegrationTest.groovy
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.operations.trace
+
+import jdk.jfr.consumer.RecordedEvent
+import jdk.jfr.consumer.RecordingFile
+import org.gradle.integtests.fixtures.daemon.DaemonIntegrationSpec
+
+/**
+ * Tests that build operations are emitted into an active (user-managed) JFR recording.
+ */
+class BuildOperationJfrEmitterIntegrationTest extends DaemonIntegrationSpec {
+
+    def "emits build operation events into an active JFR recording"() {
+        given:
+        def jfr = startJfrRecording()
+        enableBuildOperationEmission()
+
+        when:
+        succeeds("help")
+        stopDaemonAndWait()
+
+        then:
+        def buildOps = buildOperationEvents(jfr)
+        buildOps.any { it.getString("displayName") == "Run build" }
+        buildOps.every { it.getLong("operationId") != 0L }
+        buildOps.every { it.getInstant("gradleStartTime").toEpochMilli() > 0 }
+        buildOps.every { it.getInstant("gradleEndTime").toEpochMilli() > 0 }
+    }
+
+    def "JFR build operation events agree with the BuildOperationTrace"() {
+        given:
+        def operations = newBuildOperationsFixture()
+        def jfr = startJfrRecording()
+        enableBuildOperationEmission()
+
+        when:
+        succeeds("help")
+        stopDaemonAndWait()
+
+        then:
+        def jfrEventsById = buildOperationEvents(jfr).collectEntries { [(it.getLong("operationId")): it] }
+        def recordsById = operations.records.collectEntries { [(it.id): it] }
+        !jfrEventsById.isEmpty()
+
+        and: "JFR never records an operation that the BuildOperationTrace did not"
+        // JFR only captures operations that ran while a recording was active, so it is a subset of the trace.
+        (jfrEventsById.keySet() - recordsById.keySet()).isEmpty()
+
+        and: "JFR and the BuildOperationTrace agree on every shared field of every operation they both recorded"
+        jfrEventsById.each { id, event ->
+            def record = recordsById[id]
+            assert event.getString("displayName") == record.displayName
+            assert event.getLong("parentId") == (record.parentId == null ? 0L : record.parentId)
+            assert event.getInstant("gradleStartTime").toEpochMilli() == record.startTime
+            assert event.getInstant("gradleEndTime").toEpochMilli() == record.endTime
+        }
+    }
+
+    def "emits nothing when the feature is not enabled"() {
+        given:
+        // An active recording, but the emission flag is left off.
+        def jfr = startJfrRecording()
+
+        when:
+        succeeds("help")
+        stopDaemonAndWait()
+
+        then:
+        buildOperationEvents(jfr).isEmpty()
+    }
+
+    private void enableBuildOperationEmission() {
+        executer.withArgument("-D${BuildOperationJfrEmitter.SYSPROP}=true")
+    }
+
+    private File startJfrRecording() {
+        def jfrFile = file("build.jfr")
+        // Ask the daemon JVM to dump the recording to a known path on exit.
+        // withBuildJvmOpts appends to the args already set by GRADLE_OPTS so the
+        // flag actually reaches the daemon JVM (not just the client).
+        executer.withBuildJvmOpts("-XX:StartFlightRecording=filename=${jfrFile.absolutePath},dumponexit=true")
+        jfrFile
+    }
+
+    private static List<RecordedEvent> buildOperationEvents(File jfrFile) {
+        RecordingFile.readAllEvents(jfrFile.toPath()).findAll { RecordedEvent event ->
+            event.eventType.name == "org.gradle.BuildOperation"
+        }
+    }
+}
--- a/platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationJfrEmitterIntegrationTest.groovy
+++ b/platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationJfrEmitterIntegrationTest.groovy
@@ -27,7 +27,7 @@ class BuildOperationJfrEmitterIntegrationTest extends DaemonIntegrationSpec {
 
     def "emits build operation events into an active JFR recording"() {
         given:
-        def jfr = startJfrRecording()
+        def jfr = withJfrRecording()
         enableBuildOperationEmission()
 
         when:
@@ -45,7 +45,7 @@ class BuildOperationJfrEmitterIntegrationTest extends DaemonIntegrationSpec {
     def "JFR build operation events agree with the BuildOperationTrace"() {
         given:
         def operations = newBuildOperationsFixture()
-        def jfr = startJfrRecording()
+        def jfr = withJfrRecording()
         enableBuildOperationEmission()
 
         when:
@@ -74,7 +74,7 @@ class BuildOperationJfrEmitterIntegrationTest extends DaemonIntegrationSpec {
     def "emits nothing when the feature is not enabled"() {
         given:
         // An active recording, but the emission flag is left off.
-        def jfr = startJfrRecording()
+        def jfr = withJfrRecording()
 
         when:
         succeeds("help")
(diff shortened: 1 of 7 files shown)


platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEvent.java
```diff
@@ -0,0 +1,89 @@
...
+import jdk.jfr.Timestamp;
+
+/**
+ * JFR event describing a single Gradle build operation.
+ */
+@Name(BuildOperationJfrEvent.NAME)
+@Label("Build Operation")
+@Category({"Gradle", "Build Operations"})
+@StackTrace(false)
+class BuildOperationJfrEvent extends Event {
```
alllex, line 35: ❔ The JFR APIs are marked as errors when opening the file in the IDE. The compilation and tests work, because we eventually use the right release flags for the compiler, I presume. Is there a way to communicate the same to the IDE?

I don't think this is a blocker, just something that would be nice to figure eventually. Or, if we can't, at least leave a comment that this is expected IDE behavior for given reasons.

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEvent.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * The name of the JFR event type.
+     */
+    static final String NAME = "org.gradle.BuildOperation";
```
alllex: ❔ Should we make it explicit in the even type that these are internals and not the public contract?

We do require an internal option to enable the JFR event production, so I wonder if we should be consistent here. Then I'd go for `org.gradle.internal.operations.BuildOperation` to be precise.

Can the length of this name have a performance impact? Does it affect ergomonics of exploring or looking at the JFR (long names ≈ annoying)?

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEvent.java
```diff
@@ -0,0 +1,89 @@
...
+    long gradleEndTime;
+
+    /**
+     * {@code true} if the operation threw an exception, {@code false} otherwise
+     */
+    @Label("Failed")
+    boolean failed;
+
+    /**
+     * Message from the exception, or null
```
alllex, line 82: ❔ Should this be annotated with `@Nullable`? This also applies to the `failureType` field

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEmitter.java
```diff
@@ -0,0 +1,127 @@
...
+    private final BuildOperationListenerManager manager;
+    private final @Nullable BuildOperationListener listener;
+
+    /**
+     * In-flight events keyed by operation id.
+     *
+     * <p>A build operation may start on one thread and finish on another.
+     * The {@link BuildOperationListener} contract only guarantees the two signals are serialized, not that they share a thread.
+     * A per-thread map would silently drop such operations, so we key across threads.
+     */
```
alllex: 💭 The comment about different start/finish threads for the _build operations_ is very valuable. However, I think the rest is questionable. There is a difference between threads used for running build operations and notifying about them. In theory, `BuildOperationListener` contract doesn't even guarantee that the notification is synchronous, just that it is serialized.

The last line discards one among many potential optimizations that is rather obvious given the first line. I don't think it's useful.

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEvent.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+    @Label("Gradle End Time")
+    @Timestamp
+    long gradleEndTime;
+
+    /**
+     * {@code true} if the operation threw an exception, {@code false} otherwise
+     */
+    @Label("Failed")
+    boolean failed;
```
alllex: ❔ It seems at least the `failureType` is guaranteed to be non-null if the operation has failed. I can imagine how the message is optional or can become optional. Do we need the additional `failed` field for ergonomics or some other reasons?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/daemon/DaemonIntegrationSpec.groovy
```diff
@@ -33,6 +35,27 @@ abstract class DaemonIntegrationSpec extends AbstractIntegrationSpec {
...
     }
 
+    /**
+     * Stops the daemon and waits for its JVM to actually exit, so that shutdown hooks have run.
+     * Useful when a test relies on side effects that only happen on a graceful JVM shutdown, such
+     * as a JFR recording dumped via {@code -XX:StartFlightRecording=...,dumponexit=true}.
+     */
+    void stopDaemonAndWait() {
+        def daemonPid = daemons.daemon.context.pid
+        stopDaemonsNow()
```
alllex, line 52: 💅 If that's the case, can we add a doc to that method being explicit that it requests the daemons to stop, but doesn't actually wait for it? I don't think it's obvious and it's useful context for future readers

platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationJfrEmitterIntegrationTest.groovy
```diff
@@ -0,0 +1,105 @@
...
+
+        then:
+        buildOperationEvents(jfr).isEmpty()
+    }
+
+    private void enableBuildOperationEmission() {
+        executer.withArgument("-D${BuildOperationJfrEmitter.SYSPROP}=true")
+    }
+
+    private File startJfrRecording() {
```
alllex: 💅 `withJfrRecording`, since the method doesn't start anything and only instructs the build JVM to do so, when it will start (eventually)

alllex (approved):
LGTM!

I left a couple of suggestions and questions, but none of them are blocking. It would be nice to have this as part of 9.7!

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEvent.java
hegyibalint: Hmm, you are right. This might require a bit of a fine touch. 

IIRC, the original implementation was located in its own project with `workers=false` (i.e., Java 17).
This submodule has `workers=true`, but `usesFutureStdlib`, making the compilation successful, but causing exactly this problem.

WDYT? I think the implementation located in `build-operations-trace` has some merit (all the buildop tracing is colocated), but if we want to be correct, this should be spun out.

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEvent.java
hegyibalint: > Should we make it explicit in the even type that these are internals and not the public contract?

That's not a bad idea; even if the event class itself is in an internal package, we can reflect this in the name.

> Can the length of this name have a performance impact? Does it affect ergomonics of exploring or looking at the JFR (long names ≈ annoying)?

[This](https://www.morling.dev/blog/jdk-flight-recorder-file-format/) article claims that:

- A constant pool is used to store repeatedly referenced objects, such as String literals, stack traces, class and method names, etc.

> Then I'd go for org.gradle.internal.operations.BuildOperation to be precise.

Let's do this!

platforms/core-runtime/build-operations-trace/src/integTest/groovy/org/gradle/internal/operations/trace/BuildOperationJfrEmitterIntegrationTest.groovy
hegyibalint: Good point!

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEvent.java
hegyibalint: Yeah, this feels redundant!

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEmitter.java
hegyibalint: I've compacted this down

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/daemon/DaemonIntegrationSpec.groovy
hegyibalint: Added!

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEvent.java
```diff
@@ -0,0 +1,86 @@
...
+    @Nullable
+    @Label("Failure Message")
+    String failureMessage;
+
+    /**
+     * FQN of the exception class, or null
+     */
+    @Nullable
+    @Label("Failure Type")
+    String failureType;
```
alllex: 💅 I'd move it up, because there can be a failure message without a failure type (at least that's how I see the contract), so there is a semantic precedence. I'd also maybe spell out the contract (at the event-level javadoc), since there is no longer a dedicated `failed` field

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEvent.java
alllex: I think it's okay as is for now, but I'd leave a comment that this is expected. We might address in the future, when we add more JFR-related stuff

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEmitter.java
```diff
@@ -0,0 +1,124 @@
...
+
+    private final BuildOperationListenerManager manager;
+    private final @Nullable BuildOperationListener listener;
+
+    /**
+     * In-flight events keyed by operation id.
+     *
+     * <p> A build operation may start on one thread and finish on another, therefore, we need a synchronized map
+     */
+    private final Map<Long, BuildOperationJfrEvent> inFlightEvents = new ConcurrentHashMap<>();
```
alllex: 💅  It's usually good to keep the field type that carries the thread-safety aspect, i.e. change the type to `ConcurrentMap` instead of just `Map`

alllex (approved):

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEvent.java
hegyibalint: Good idea! Implemented.

platforms/core-runtime/build-operations-trace/src/main/java/org/gradle/internal/operations/trace/BuildOperationJfrEmitter.java
hegyibalint: This is something I never thought about. Thanks for sharing. 
It has merit. I changed the type.

PR #38083, opened 2026-06-02
Allow hits for project default task request in Configuration Cache

tresat:
Store configuration cache entries when a project's default task is invoked implicitly (e.g., `gradle`) and allow matching against explicit invocations (e.g., `gradle build`), and vice-versa.

This improves cache hit rates for default task invocations at the small cost of storing additional index data when the default task is invoked (implicitly or explicitly).

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheDefaultTaskIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheDefaultTaskIntegrationTest.groovy
@@ -1,0 +1,359 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl
+
+class ConfigurationCacheDefaultTaskIntegrationTest extends AbstractConfigurationCacheIntegrationTest {
+
+    def "running build twice with no task arguments hits the configuration cache"() {
+        given:
+        def configurationCache = newConfigurationCacheFixture()
+        buildFile("""
+            tasks.register("echo") {
+                doLast { println("echo executed") }
+            }
+            defaultTasks "echo"
+        """)
+
+        when:
+        configurationCacheRun()
+
+        then:
+        configurationCache.assertStateStored()
+        outputContains("echo executed")
+
+        when:
+        configurationCacheRun()
+
+        then:
+        configurationCache.assertStateLoaded()
+        outputContains("echo executed")
+    }
+
+    def "running build with no arguments then explicit default task name hits the configuration cache"() {
+        given:
+        def configurationCache = newConfigurationCacheFixture()
+        buildFile("""
+            tasks.register("echo") {
+                doLast { println("echo executed") }
+            }
+            defaultTasks "echo"
+        """)
+
+        when:
+        configurationCacheRun()
+
+        then:
+        configurationCache.assertStateStored()
+        outputContains("echo executed")
+
+        when:
+        configurationCacheRun("echo")
+
+        then:
+        configurationCache.assertStateLoaded()
+        outputContains("echo executed")
+    }
+
+    def "running build with explicit default task name then no arguments hits the configuration cache"() {
+        given:
+        def configurationCache = newConfigurationCacheFixture()
+        buildFile("""
+            tasks.register("echo") {
+                doLast { println("echo executed") }
+            }
+            defaultTasks "echo"
+        """)
+
+        when:
+        configurationCacheRun("echo")
+
+        then:
+        configurationCache.assertStateStored()
+        outputContains("echo executed")
+
+        when:
+        configurationCacheRun()
+
+        then:
+        configurationCache.assertStateLoaded()
+        outputContains("echo executed")
+    }
+
+    def "running unrelated explicit task after no-args does not hit the configuration cache"() {
+        given:
+        def configurationCache = newConfigurationCacheFixture()
+        buildFile("""
+            tasks.register("echo") {
+                doLast { println("echo executed") }
+            }
+            tasks.register("build") {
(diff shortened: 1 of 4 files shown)


tresat:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheDefaultTaskIntegrationTest.groovy
```diff
@@ -0,0 +1,369 @@
...
+        when:
+        configurationCacheRun("help")
+
+        then:
+        configurationCache.assertStateStored()
+    }
+
+    def "aliasing respects the excluded tasks list"() {
+        given:
+        def configurationCache = newConfigurationCacheFixture()
```
6hundreds, line 294: 💅 I'd say `def configurationCache = newConfigurationCacheFixture()` might be in `setup` for deduplication, as every test contains it

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/ConfigurationCacheStartParameter.kt
```diff
@@ -147,6 +150,16 @@ class ConfigurationCacheStartParameter internal constructor(
...
     }
 
+    /**
+     * Whether this build was invoked without explicit task names (i.e. with [RunDefaultTasksExecutionRequest]).
+     *
+     * Captured eagerly at construction time because [DefaultTasksBuildTaskScheduler] replaces
+     * [StartParameter.taskRequests] with a [DefaultTaskExecutionRequest] once it resolves the
+     * default tasks, after which the original signal is unrecoverable.
+     */
+    val wasDefaultTaskRequest: Boolean =
```
6hundreds, line 160: 💅 According to the rest of the fields in this class, wouldn't it be more natural to call it with something prefixed by `is`? Like `isDefaultTaskRequest`?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheKey.kt
```diff
@@ -61,10 +74,8 @@ class ConfigurationCacheKey(
 
         buildActionRequirements.appendKeyTo(this)
 
-        // TODO:bamboo review with Adam
-//        require(buildActionRequirements.isRunsTasks || startParameter.requestedTaskNames.isEmpty())
```
6hundreds, line 65: ❓ Why do we remove it? Does this PR make it obsolete?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheDefaultTaskIntegrationTest.groovy
```diff
@@ -0,0 +1,369 @@
...
+ * Verifies that the configuration cache treats a no-argument invocation and an
+ * explicit invocation of the project's {@code defaultTasks} as interchangeable:
+ * either form on a subsequent build hits the cache entry stored by the other.
+ *
+ * Negative tests pin down that unrelated tasks, partial subsets, reordered task
+ * names, qualified paths, and {@code -x} exclusions are <em>not</em> incorrectly
+ * aliased; an additional test confirms that a shared entry is still invalidated
+ * by the normal fingerprint check when a build input changes between runs.
+ */
+class ConfigurationCacheDefaultTaskIntegrationTest extends AbstractConfigurationCacheIntegrationTest {
```
6hundreds, line 29: ❓ I see the implementation touches CC multiple entries logic, should we have tests with multiple entries enabled then?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultConfigurationCache.kt
```diff
@@ -409,8 +410,85 @@ class DefaultConfigurationCache internal constructor(
...
+    }
+
+    /**
+     * True when the invocation's effective tasks match the project's resolved
+     * default tasks, so an alias index would make the just-committed entry
+     * reachable via the "other form" of the invocation (e.g. `gradle` vs
+     * `gradle <defaultTaskName>`) on a future build.
+     */
+    private
+    fun needToWriteSecondaryEntry(): Boolean {
```
6hundreds, line 425: 💅 This method should have more precise name. We don't have secondary entry concept in general, this is something that only required by this default tasks support. So I'd propose `needToWriteDefaultTaskAliasEntry`.

Also, the method is guarded by this check duplicate same calls for `computeAliasTaskNames` (I think it's cheap) and `cacheKey.stringForRequestedTaskNames(aliasTaskNames)`(not sure how cheap is it). 

As an option we can re-shape the api to smth like:
```
writeDefaultTaskAliasEntryIfNeeded(write: (aliasKey : String) -> Unit)
```

WDYT?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultConfigurationCache.kt
```diff
@@ -409,8 +410,85 @@ class DefaultConfigurationCache internal constructor(
...
+     * Returns the task-names list to use for the alias key, or null if no alias
+     * should be written. The list is the "opposite" view of this invocation:
+     * empty when invocation was explicit-default; the resolved defaults when
+     * invocation was no-args.
+     */
+    private
+    fun computeAliasTaskNames(): List<String>? {
+        val resolvedDefaults = resolvedDefaultTasks()
+        if (resolvedDefaults.isEmpty()) return null
+        val primaryTaskNames = if (startParameter.wasDefaultTaskRequest) {
```
6hundreds, line 469: 💅 this is `requestedTaskNames` I believe.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultConfigurationCache.kt
```diff
@@ -575,6 +653,18 @@ class DefaultConfigurationCache internal constructor(
...
     private
     fun updateCandidateEntries(update: List<CandidateEntry>.() -> List<CandidateEntry>) = store.useForStore {
+        applyCandidateEntriesUpdate(update)
+    }
+
+    /**
+     * Shared read/update/write/schedule-eviction body for candidate-entry index updates.
+     * Reads the current index, applies [update], and if the result differs writes the
+     * new index and schedules now-evicted entries for collection.
+     */
```
6hundreds, line 663: 💅 🤖 I don't think this comment makes much sense

6hundreds (changes requested):
Some comment applied, PTAL!
I also think that majority of comments are redundant.

PR #38086, opened 2026-06-03
Fix Windows start script not exiting when failing to locate Java

### LINKED ISSUE #38082: Windows start script doesn't exit cleanly when no JVM is found
### Current Behavior

The rest of the script continues to exit, producing some garbage output. It still reports a `1` exit code though.

### Expected Behavior

It should fail cleanly after printing the appropriate message.

### Context (optional)

Originally reported in https://github.com/gradle/gradle/pull/37965

`gradle init` reproduces this as the wrapper uses the buggy script.

### Gradle version

9.5.1

### Build scan URL (optional)

_No response_

_No response_

octylFractal:
Fixes #38082

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/jvm/plugins-application/src/integTest/groovy/org/gradle/api/plugins/ApplicationPluginIntegrationTest.groovy
+++ b/platforms/jvm/plugins-application/src/integTest/groovy/org/gradle/api/plugins/ApplicationPluginIntegrationTest.groovy
@@ -238,6 +238,92 @@ System.out.println("CLASSPATH: " + System.getenv("CLASSPATH"));
         outputContains('CLASSPATH: \n')
     }
 
+    @Requires(OsTestPreconditions.Windows)
+    @Issue("https://github.com/gradle/gradle/issues/38082")
+    def "Windows start script fails cleanly when JAVA_HOME is invalid"() {
+        given:
+        def startScriptDir = file('build/install/sample/bin')
+        def outputFile = file('script-output.txt')
+        def invalidJavaHome = file('invalid-java-home')
+
+        when:
+        succeeds('installDist')
+        runViaWindowsStartScript(startScriptDir, [JAVA_OPTS: '', JAVA_HOME: invalidJavaHome.canonicalPath], outputFile)
+
+        then:
+        def scriptOutput = outputFile.text
+        scriptOutput.contains('ERROR: JAVA_HOME is set to an invalid directory')
+        // The script must stop after the error rather than falling through to :execute and
+        // trying to launch the missing java.exe, which prints "cannot find the path/file
+        // specified".
+        !scriptOutput.contains('cannot find the')
+    }
+
+    @Requires(OsTestPreconditions.Windows)
+    @Issue("https://github.com/gradle/gradle/issues/38082")
+    def "Windows start script fails cleanly when no 'java' command can be found"() {
+        given:
+        def startScriptDir = file('build/install/sample/bin')
+        def outputFile = file('script-output.txt')
+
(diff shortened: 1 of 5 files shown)


octylFractal:
@bot-gradle test ACTR

platforms/jvm/plugins-application/src/integTest/groovy/org/gradle/api/plugins/ApplicationPluginIntegrationTest.groovy
```diff
@@ -238,6 +238,47 @@ System.out.println("CLASSPATH: " + System.getenv("CLASSPATH"));
...
+        def outputFile = file('script-output.txt')
+        def invalidJavaHome = file('invalid-java-home')
+
+        when:
+        succeeds('installDist')
+        runViaWindowsStartScript(startScriptDir, [JAVA_OPTS: '', JAVA_HOME: invalidJavaHome.canonicalPath], outputFile)
+
+        then:
+        def scriptOutput = outputFile.text
+        scriptOutput.contains('ERROR: JAVA_HOME is set to an invalid directory')
```
tresat: 🤔 Should this also print the invalid directory?

tresat (approved):

octylFractal:
This has issues with upgrading older wrapper scripts, so we're deferring it until we figure out exactly how we want to move forward with that.

platforms/jvm/plugins-application/src/integTest/groovy/org/gradle/api/plugins/ApplicationPluginIntegrationTest.groovy
octylFractal: It does if you look at the script. I could try to assert it if you want.

octylFractal:
@bot-gradle test ACVT

octylFractal:
@bot-gradle test ACT

platforms/jvm/plugins-application/src/integTest/groovy/org/gradle/api/plugins/ApplicationPluginIntegrationTest.groovy
tresat: I think asserting is worth doing - it's a more useful error message if it shows what directory is invalid, so that's worth explicitly testing.

octylFractal: Done

tresat (approved):

cobexer:
Can't be merged because `org.gradle.api.plugins.ApplicationPluginIntegrationTest.can generate start scripts with module path` hangs

Should see the stacktraces there eventually: https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_Quick_2_bucket29/116258620?buildTab=overview

octylFractal:
I can't find any reason for the hang or reproduce it on my Windows machine, it seems to be hanging before it even starts the subprocess, which should be unrelated to my changes. If it keeps happening I can investigate more, but I think it's likely unrelated.

octylFractal:
@bot-gradle test QF

octylFractal:
@bot-gradle test QF without PTS

octylFractal:
The `ComSpec` read was causing a deadlock that I'm fixing in #38888. I've changed the code so the test should work now.

PR #38089, opened 2026-06-03
Name the declaring project in the parent-walk deprecation message

reinsch82:
### Context

The parent-hierarchy lookup deprecation says `Property 'foo' was not declared in project ':a:b' and was resolved from <project>` — but the named project was always the **immediate parent**, not the project that actually declares the member.

The walk queried each ancestor through its inherited view (`snapshotInheritable()`), which embeds that ancestor's own parent as its last composite delegate. The first level therefore transitively searched the entire ancestor chain: the loop always matched on its first iteration and attributed the member to the immediate parent, even when it is declared further up (e.g. on the root project).

### Change

- Add `HierarchicalDynamicObject.withoutParent()`, exposing a single level's own members (for `InheritedDynamicObject`: extra properties, script members, extensions — without the parent delegate).
- Walk the hierarchy one level at a time via `withoutParent()` + `getParent()`, so the loop's position on a match identifies the exact declaring project, which the deprecation (and the strict `fail-on-parent-property-lookup` error) message then names.

Resolution semantics are unchanged: the recursive composite search linearizes to exactly the level-by-level sequence the loop now walks explicitly (closest ancestor wins; per-level order extras → script members → extensions). The message format is also unchanged — only the attribution becomes precise — so existing two-level test expectations and the Develocity deprecation type are unaffected.

### Verification

New integration tests in `ParentProjectPropertyLookupIntegrationTest` using a three-level hierarchy (`root` → `:a` → `:a:b`):
- property declared only on root → message names `root project 'root'` (previously `project ':a'`)
- property declared on both root and `:a` → names `project ':a'` and resolves the closer value (precedence preserved)
- method declared only on root → names `root project 'root'`

Full `ParentProjectPropertyLookupIntegrationTest` suite passes (23/23). `ExtensibleDynamicObjectTest` shows the same 3 pre-existing failures as vanilla `release` (deprecation nag in uninitialized unit-test environment) and no new ones.

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
@@ -177,16 +177,25 @@ public class ExtensibleDynamicObject extends AbstractDynamicObject implements Hi
     }
 
     private DynamicObject snapshotInheritable() {
+        return snapshotInheritable(true);
+    }
+
+    private DynamicObject snapshotInheritable(boolean includeParent) {
         List<DynamicObject> delegates = new ArrayList<>(4);
         delegates.add(extraPropertiesDynamicObject);
         if (beforeConvention != null) {
             delegates.add(beforeConvention);
         }
         delegates.add(extensionContainer.getExtensionsAsDynamicObject());
-        if (parent != null) {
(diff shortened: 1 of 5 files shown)


reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
Superseded by https://github.com/gradle/gradle/pull/38094

PR #33846, opened 2025-06-12
Compile Kotlin DSL scripts against the public ABI jar from the distribution

jbartok:
* Fix https://github.com/gradle/gradle/issues/26484
* Relates to #15886
----

This PR does a few main things:
* build a new Public API ABI jar (`gradle-public-api-legacy`), with dependencies equivalent to the current generated API jar
* embed that Public API ABI jar in the distribution (~6MiB)
* let the Kotlin DSL script compilation use the ABI JAR instead of requiring the Gradle API jar generation
* let the TAPI model builders use the ABI JAR instead of requiring the Gradle API jar generation

It doesn't change:
* the fact that the ABI jar contains internals
* the fact that plugin projects with a dependency on `gradleKotlinDsl()` (or applying `kotlin-dsl` plugin) use the generated API JAR

The first few commits are opportunistic refinements.

A side effect of this PR is that our integration tests that rely on `:distribution-full` now don't need to re-generate the API jar on every production code change when running tests that use Kotlin DSL scripts. They get faster to start. It is not the case for integ tests using the other smaller testing distributions as they don't get a ABI jar.

Potential breaking changes:
- scripts importing from relocated dependencies `org.gradle.internal.impldep.*` will not compile anymore

Potential follow ups:
- let testing distributions (basics, core, jvm ...) embed an adequate ABI jar in order to remove the fallback to the generated API jar (simplifying production code and making more of our integ tests faster)

----

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #38092, opened 2026-06-03
Embed a public API ABI jar in all integ-testing distributions

eskatos:
Kotlin DSL script compilation only avoided generating the large gradle-api.jar in :distributions-full, the sole distribution that embedded the public API ABI jar. Every other distribution fell back to generating it, defeating the benefit for those distributions.

Each distribution now produces its own ABI jar scoped to its runtime, so the runtime fallback can be removed and script compilation resolves the embedded jar in all distributions.

Split gradlebuild.public-api-jar so jar production is reusable by the distributions, while publication stays specific to :public-api.

* Follow up for https://github.com/gradle/gradle/pull/33846

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/model/GradleDistribution.kt
+++ b/build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/model/GradleDistribution.kt
@@ -75,6 +75,12 @@ abstract class GradleDistribution {
             include("lib/plugins/*.jar")
         }
 
+    @get:Classpath
+    val apiJars: FileCollection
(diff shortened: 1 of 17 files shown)


eskatos:
@bot-gradle test this

eskatos:
@bot-gradle test this

eskatos:
@bot-gradle test this

eskatos:

jbartok (approved):

PR #38094, opened 2026-06-03
Name the declaring project in the parent-walk deprecation message

reinsch82:
### Context

The parent-hierarchy lookup deprecation names the project a member was resolved from — but the attribution was always the **immediate parent**, not the project that actually declares the member. Each ancestor's inherited view (`snapshotInheritable()`) embedded its own parent as its last composite delegate, so the first queried level transitively searched the entire ancestor chain: the walk loop matched on its first iteration and could only ever blame the immediate parent.

### Changes

**1. Attribute the lookup to the declaring project** (`don't use parent in inheritable list`)

Drop the parent from the inherited view, so `getInheritable()` exposes only a single level's own inheritable members (extra properties, script members, extensions). The existing walk loops in `ExtensibleDynamicObject` then advance level by level via `getParent()` — the loop's position on a match identifies the exact declaring project, which the deprecation (and strict `fail-on-parent-property-lookup` error) message then names.

**2. Adapt the inherited-scope unit tests to the level-local contract**

`ExtensibleDynamicObjectTest.parentPropertiesAreInherited` and `DefaultProjectTest.inheritedPropertiesAreInheritable` asserted the old transitive contract of the inherited view; they now assert the level-local contract (ancestors reached via `getParent()`). The parent-resolution unit tests also wrap their lookups in `DeprecationLogger.whileDisabled()` so they pass deterministically regardless of whether another test initialized the `DeprecationLogger` in the same JVM.

**3. Reword the deprecation message**

The action is renamed from `Implicitly resolving {properties|methods} in the project hierarchy` to `Implicit lookup of {properties|methods} in parent projects`, and the upgrade guide anchor from `deprecated_implicit_project_hierarchy_lookup` to `deprecated_implicit_lookup_in_parent_projects`. Since the problem id derives from the action, it becomes `deprecation:implicit-lookup-of-{properties|methods}-in-parent-projects`. All test expectations, `KnownProblemIds`, the upgrade guide section and its cross-references, and the release notes are updated accordingly.

Note: this changes the deprecation type relative to what 9.6 RC1/RC2 emit (a one-time stable rename — the action still does not vary per call), and RC-emitted doc links point to the renamed anchor.

### Verification

New integration tests in `ParentProjectPropertyLookupIntegrationTest` using a three-level hierarchy (`root` → `:a` → `:a:b`):
- property declared only on root → message names `root project 'root'` (previously `project ':a'`)
- property declared on both root and `:a` → names `project ':a'` and resolves the closer value (precedence preserved)
- method declared only on root → names `root project 'root'`

`ExtraPropertiesIntegrationTest` had encoded the old immediate-parent attribution for a root-declared property in a grandchild lookup; it now expects the root project.

Replaces #38089, which achieved the same attribution fix by introducing a `withoutParent()` API instead of changing the inherited view's contract.


reinsch82:
@bot-gradle test on linux without PTS

reinsch82:
@bot-gradle test on linux without PTS

reinsch82:
@bot-gradle test on linux without PTS

reinsch82:
@bot-gradle test RFN

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/extensibility/ExtensibleDynamicObject.java
```diff
@@ -183,9 +183,6 @@ private DynamicObject snapshotInheritable() {
             delegates.add(beforeConvention);
         }
         delegates.add(extensionContainer.getExtensionsAsDynamicObject());
-        if (parent != null) {
```
reinsch82, line 186: The actual fix, the rest was already in place

lkasso (approved):
Docs LGTM

alllex (approved):
💯

PR #36174, opened 2026-01-01
Allow ProjectComponentIdentifiers to be used as task inputs

jvandort:
Previously, we only allowed ModuleComponentIdentifiers to be used as task inputs. Plenty of dependency graphs include project components, so it is very useful to be able to include these as task inputs

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #38097, opened 2026-06-03
Add class dependency analysis task for gradle projects

ghale:
Adds a new analyzeClassDependencies Gradle task to the gradlebuild.refactoring build-logic plugin. The task computes the transitive project/class dependency closure starting from --class and/or --package inputs and emits a JSON report optimized for consumption by agents driving an effort to extract those classes from `:core` or `:core-api`.

The task is a sibling of splitProject (introduced in #37822): it reuses the same DependencyAnalyzer.

## Motivation

Agent workflows that extract classes need a way to ask "what does this class/package transitively depend on within the project, and which other Gradle subprojects does it reach?"  JSON output makes the answer ergonomic to consume from an agent.

## What's added

New task: analyzeDependencies (registered on every project that applies gradlebuild.refactoring):

```
./gradlew :<subproject>:analyzeClassDependencies \
    --class=org.gradle.api.plugins.catalog.VersionCatalogPlugin \
    --output=/tmp/analysis.json

./gradlew :<subproject>:analyzeClassDependencies \
    --package=org.gradle.api.plugins.catalog \
    --output=/tmp/analysis.json
```

- --class / --package accept comma-separated values; at least one is required; both may be combined; --package matches recursively into subpackages.
- --output is a required path to the JSON destination file.
- Unknown classes, empty packages, and a missing --output all fail with a clear GradleException.

### Output shape
```
{
  "projectPath": ":plugins-version-catalog",
  "roots": ["org.gradle.api.plugins.catalog.VersionCatalogPlugin"],
  "subprojectDependencies": [":base-services", ":core", ":core-api", ":dependency-management", ...],
  "projectClasses": [
    {
      "fqcn": "org.gradle.api.plugins.catalog.VersionCatalogPlugin",
      "sourceFile": "src/main/java/org/gradle/api/plugins/catalog/VersionCatalogPlugin.java",
      "dependencies": [
        { "fqcn": "org.gradle.api.Plugin", "kind": "subproject", "subproject": ":core-api" },
        { "fqcn": "org.gradle.api.plugins.catalog.internal.CatalogExtensionInternal", "kind": "project" }
      ]
    }
  ]
}
```

- projectClasses enumerates every project class in the transitive closure, with its direct one-hop dependencies tagged as either kind: "project" (same subproject) or kind: "subproject" (another Gradle subproject in this build). Third-party / external dependencies are intentionally omitted.
- subprojectDependencies is the sorted aggregate of reached subproject paths.
- All arrays are sorted alphabetically for deterministic, diff-friendly output.

## Notes
This PR is stacked on top of (and includes the changes from) #37822,  which brought in SplitProjectTask and DependencyAnalyzer; this PR builds on this to provide `AnalyzeClassDependenciesTask`.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


build-logic/refactoring/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
+plugins {
+    id("gradlebuild.build-logic.kotlin-dsl-gradle-plugin")
+}
+
+description = "Provides plugins to help refactor Gradle projects"
```
jvandort: When I had originally written this I did not realize we had a `:build-logic:build-update-utils` project. 

It might make sense to put this plugin there instead of creating a new project

build-logic/refactoring/src/main/kotlin/gradlebuild/refactoring/DependencyAnalyzer.kt
```diff
@@ -0,0 +1,531 @@
...
+import java.io.File
+import java.util.zip.ZipFile
+
+/**
+ * Result of the dependency analysis.
+ *
+ * @property projectClasses The set of classes that belong to the project and are part of the transitive dependency closure.
+ * @property apiDependencyRoots The set of external JARs or directories that contain API dependencies of the [projectClasses].
+ * @property implementationDependencyRoots The set of external JARs or directories that contain Implementation dependencies of the [projectClasses].
+ */
```
jvandort: This javadoc seems misplaced. Seems like it should be on `AnalysisResult` and is missing some properties.

build-logic/refactoring/src/main/kotlin/gradlebuild/refactoring/SplitProjectTask.kt
```diff
@@ -0,0 +1,527 @@
...
+        }
+
+        println("Successfully generated project ${toProject.get()} at ${targetDir.asClickableFileUrl()}")
+        println()
+        println("The following manual steps must still be performed:")
+        println("- Add 'subproject(\"${toProject.get()}\")' to ${rootProjectDirectory.file("settings.gradle.kts").get().asFile.asClickableFileUrl()}")
+        println("- Update the the current project to depend on `projects.${toProjectAccessorName(":${toProject.get()}")}`: ${currentProjectDirectory.file("build.gradle.kts").get().asFile.asClickableFileUrl()}")
+        println("- Run `./gradlew :generateSubprojectsInfo`")
+        println("- Run `./gradlew projectHealth --continue` and resolve any reported dependency declaration issues")
+        println("- Sync the build and commit the idea.xml file changes")
```
jvandort: This step is no longer necessary as we've removed `idea.xml` from git

build-logic/build-update-utils/src/main/kotlin/gradlebuild/refactoring/tasks/AnalyzeClassDependenciesTask.kt
```diff
@@ -0,0 +1,319 @@
...
+/**
+ * A task that computes the transitive project-class dependency closure for a given set of root classes
+ * (by FQCN and/or by package) and writes a JSON file optimized for agent consumption.
+ *
+ * The task reuses [DependencyAnalyzer] to perform the bytecode-driven traversal, then emits an entry
+ * for each reachable project class with its direct one-hop dependencies, tagged as either belonging
+ * to this project ([DependencyKind.PROJECT]) or to another Gradle subproject ([DependencyKind.SUBPROJECT]).
+ * Third-party / external dependencies are intentionally omitted from the output.
+ */
+@DisableCachingByDefault(because = "Unable to snapshot ComponentIdentifier") // Same root cause as SplitProjectTask: gradle/gradle#36174
```
jvandort: This should be fixed now https://github.com/gradle/gradle/pull/36174
This was merged into 9.4.0, so the build already runs on a Gradle version which can snapshot project component IDs

build-logic/refactoring/src/main/kotlin/gradlebuild/refactoring/AnalyzeClassDependenciesTask.kt
```diff
@@ -0,0 +1,319 @@
...
+    @get:Input
+    @get:Optional
+    @get:Option(option = "package", description = "Comma-separated list of packages to analyze (recursive)")
+    abstract val rootPackages: Property<String>
+
+    /**
+     * Path to the JSON output file. Required.
+     */
+    @get:OutputFile
+    @get:Optional
```
jvandort: @Optional, but also `Required.`? 

Is the Gradle behavior if this is not specified insufficient?

Can we make this print to console if not specified? Or is that not useful for agentic workflows?

build-logic/refactoring/src/main/kotlin/gradlebuild/refactoring/AnalyzeClassDependenciesTask.kt
```diff
@@ -0,0 +1,319 @@
...
+
+    @get:InputFiles
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    abstract val groovySourceDirectories: ConfigurableFileCollection
+
+    /**
+     * The Gradle path of the subproject being analyzed (e.g., `:core`).
+     */
+    @get:Internal
+    abstract val projectPath: Property<String>
```
jvandort: Curious how/why this is used in your workflow

build-logic/build-update-utils/src/main/kotlin/gradlebuild/refactoring/tasks/AnalyzeClassDependenciesTask.kt
```diff
@@ -0,0 +1,319 @@
...
+    private fun toRelativeSourceFilePath(fqcn: String): String {
+        val topLevel = if (fqcn.contains('$')) fqcn.substringBefore('$') else fqcn
+        return topLevel.replace('.', '/')
+    }
+
+    private fun String?.toCommaSet(): Set<String> =
+        this.orEmpty().split(',').map { it.trim() }.filter { it.isNotEmpty() }.toSet()
+
+    private fun File.asClickableFileUrl(): String =
+        URI("file", "", toURI().path, null, null).toASCIIString()
```
jvandort, line 294: Might be worth moving this to a common location

`CheckTargetRuntimes` also uses this

jvandort (approved):

build-logic/build-update-utils/src/main/kotlin/gradlebuild/refactoring/tasks/AnalyzeClassDependenciesTask.kt
ghale: Removed

build-logic/refactoring/src/main/kotlin/gradlebuild/refactoring/SplitProjectTask.kt
ghale: Removed

build-logic/refactoring/src/main/kotlin/gradlebuild/refactoring/DependencyAnalyzer.kt
ghale: Moved.

build-logic/refactoring/build.gradle.kts
ghale: Ah, yes.  I've moved it to the existing project.

build-logic/refactoring/src/main/kotlin/gradlebuild/refactoring/AnalyzeClassDependenciesTask.kt
ghale: It was originally optional, and then changed to required, but the annotation wasn't removed.

The main reason I use json here is that it doesn't need to be read by the agent in its entirety.  The agent can use `jq` to parse for the subsets it needs when it needs them without reading the entire thing into context.  So, even if it printed to the console, a smart agent would just write the output to a file anyways.  I'm not hostile to the idea of having it default to this when this isn't set, but it seems to have limited usefulness in the way I'm currently using it.  I'm happy to revisit this if we find a use case where this is useful.

build-logic/refactoring/src/main/kotlin/gradlebuild/refactoring/AnalyzeClassDependenciesTask.kt
ghale: It's not.  This is AI slop that got added after I reviewed the initial round of work.  I should have caught this. 

Removed.

PR #38099, opened 2026-06-04
Stop debug integration tests hanging on the wrong network interface

blindpirate:
### Context

`JavaExecDebugIntegrationTest` regularly hung for ~50 minutes and timed out the whole build on the **Flaky Test Quarantine – ConfigCache** pipeline (one iteration of the `server = true` tests consumed the entire 60-minute execution timeout, so the build failed with `Execution timeout` and the remaining tests never ran). `CommandLineIntegrationSpec` has the same bug.

#### Root cause: the debugger attaches over the wrong network interface

`nonLoopbackAddress()` returned the **first** non-loopback IPv4 it enumerated. On the CI agents that is `172.17.0.1` — the **`docker0` bridge** — and on dev machines it is often a **VPN tunnel** (`utunN`). Attaching a JDWP debugger across such an interface stalls intermittently, and the tests had no fast-failure path.

Reproduced locally:

```
OLD (first non-loopback) -> 172.16.0.2    # utun1, a VPN tunnel (wrong)
NEW (default route)      -> 192.168.8.126 # en0, the real interface (right)
```

Neither `isVirtual()` nor `isPointToPoint()` excludes both `docker0` and `utunN`, so the fix selects the interface that carries the **default route** (a UDP "connect" route lookup that sends no packets). The logic is shared via `JDWPUtil` rather than duplicated in each test.

### Fix

1. **Root cause** — `nonLoopbackAddress()` (now `JDWPUtil.nonLoopbackAddress()`) returns the default-route interface address, falling back to enumeration only if that can't be determined. Applied to both `JavaExecDebugIntegrationTest` and `CommandLineIntegrationSpec`.
2. **Defense in depth**, so a future stall can never again take down a whole build:
   - **`JDWPUtil`**: bound `attach()` with a **watchdog thread**. The JDWP handshake is a blocking, *uninterruptible* socket read — I verified the connector's own `timeout` argument bounds only the TCP connect, **not** the handshake (attach stayed blocked >12s with `timeout=2000`), and `Thread.interrupt()`/`@Timeout` can't unblock a plain socket read either. So the attach runs on a thread we abandon after a fixed timeout, failing fast.
   - **`JavaExecDebugIntegrationTest`**: `@Timeout` on the two `server = true` tests backstops the *other* wait — `handle.waitForFinish()` — which **is** interruptible (`DefaultExecHandle.waitForFinish()` aborts the process on interrupt).

### Verification

- Reproduced the wrong-interface selection locally and confirmed the default-route approach picks the real NIC.
- Empirically confirmed the connector `timeout` arg does not bound the handshake, and that the watchdog thread does (gives up at the bound instead of hanging).
- `:internal-distribution-testing:compileGroovy`, `:language-java:compileIntegTestGroovy`, `:launcher:compileIntegTestGroovy` all compile.
- The CI hang itself is a Docker-bridge race that does not reproduce locally, so no new regression test is added.

### Contributor Checklist
- [x] [Review Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md)
- [x] Commits are [signed off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff) (DCO)
- [x] Code is distributable under the Apache License 2.0
- [ ] "Allow edit from maintainers" — N/A, branch lives on `gradle/gradle`
- [x] Integration tests — fixes the existing integration tests / their shared fixture; the hang is a CI-only race that cannot be reliably reproduced in a test
- [ ] Unit tests — N/A (test infrastructure)
- [ ] User Guide / DSL Reference / Javadoc — N/A (no public-facing change)
- [ ] `./gradlew sanityCheck` — not run; verified the changed source sets compile

### DIFF
--- a/platforms/core-runtime/launcher/src/integTest/groovy/org/gradle/launcher/CommandLineIntegrationSpec.groovy
+++ b/platforms/core-runtime/launcher/src/integTest/groovy/org/gradle/launcher/CommandLineIntegrationSpec.groovy
@@ -116,7 +116,7 @@ class CommandLineIntegrationSpec extends AbstractIntegrationSpec {
         given:
         JDWPUtil jdwpClient = new JDWPUtil()
 
-        def jdwpHost = nonLoopbackAddress()
+        def jdwpHost = JDWPUtil.nonLoopbackAddress()
         Assume.assumeNotNull(jdwpHost)
         jdwpClient.host = jdwpHost
 
@@ -152,7 +152,7 @@ class CommandLineIntegrationSpec extends AbstractIntegrationSpec {
         given:
         JDWPUtil jdwpClient = new JDWPUtil()
 
-        def address = nonLoopbackAddress()
+        def address = JDWPUtil.nonLoopbackAddress()
         Assume.assumeNotNull(address)
         jdwpClient.host = address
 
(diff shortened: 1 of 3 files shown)


ghale (approved):
LGTM

PR #38104, opened 2026-06-04
Standardize variant name quoting in display messages

### LINKED ISSUE #31503: Expose variant name in SourceVariantData directly

tresat:
Ensure variant names are consistently quoted in user-facing messages, such as error reports. This improves readability and reduces ambiguity by clearly delineating variant names from surrounding text.

Replaces:
- https://github.com/gradle/gradle/pull/38021

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/resolution/failure/transform/SourceVariantData.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/resolution/failure/transform/SourceVariantData.java
@@ -47,12 +47,7 @@ public final class SourceVariantData {
     }
 
     public String getFormattedVariantName() {
-        int variantIdx = variantName.indexOf(" variant ");
-        if (variantIdx == -1) {
-            return variantName;
-        } else {
-            return variantName.substring(0, variantIdx + 9) + "'" + variantName.substring(variantIdx + 9) + "'";
-        }
+        return variantName;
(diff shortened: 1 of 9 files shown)


platforms/software/dependency-management/src/main/java/org/gradle/internal/component/resolution/failure/transform/SourceVariantData.java
```diff
@@ -47,12 +47,7 @@ public ImmutableAttributes getAttributes() {
     }
 
     public String getFormattedVariantName() {
```
jvandort: The return value of this method is no more formatted than any other display name. I think we should rename this to getDisplayName or getVariantDisplayName

jvandort (approved):

PR #38105, opened 2026-06-04
Deprecate undeclared project artifact transform dependencies

### LINKED ISSUE #37219: Undeclared artifact transform dependencies should be disallowed rather than silently accepted
When a task queries a `FileCollection` containing the output of an artifact transform on project artifacts without declaring this as a task dependency, Gradle currently allows it. 

With Configuration Cache, this causes confusing "project not found" errors during cache replay because the transform nodes are not serialized.                                                                                                  
                                                                                                                                                 
Rather than improving the error message under Configuration Cache, Gradle should probably disallow this usage entirely: if a task consumes the output of an artifact transform, the dependency must be declared. The current undesired behavior is documented by the tests in `UndeclaredDependencyResolutionIntegrationTest`.                                                                                                   
                                                                                                                                                 
See `UndeclaredDependencyResolutionIntegrationTest` for examples of the affected scenarios:                                                        
- Querying transform output of project artifacts without declaring input
- Querying transform output at task graph calculation time                                                                                       
- Chained transforms of project artifacts without declared dependencies                                                                        
                                                                                                                                                   
Related:                                                                                                                                         
- https://github.com/gradle/gradle/issues/24273                                                                                                  
- https://github.com/gradle/gradle/issues/37202

tresat:
Fixes:
- https://github.com/gradle/gradle/issues/37219


testing/architecture-test/src/changes/archunit-store/platform-coupling.txt
```diff
@@ -153,6 +153,7 @@ Class 'org.gradle.internal.serialize.codecs.dm.ResolveArtifactNodeCodec' from pl
 Class 'org.gradle.internal.serialize.codecs.dm.ResolvedComponentResultCodec' from platform 'core' imports class 'org.gradle.api.internal.artifacts.result.ResolvedComponentResultInternal' from a forbidden platform 'software'
 Class 'org.gradle.internal.serialize.codecs.dm.ResolvedComponentResultCodec' from platform 'core' imports class 'org.gradle.api.internal.artifacts.result.ResolvedGraphResult' from a forbidden platform 'software'
 Class 'org.gradle.internal.serialize.codecs.dm.ResolvedGraphResultCodec' from platform 'core' imports class 'org.gradle.api.internal.artifacts.result.ResolvedGraphResult' from a forbidden platform 'software'
+Class 'org.gradle.internal.serialize.codecs.dm.transform.AbstractTransformStepNodeCodec' from platform 'core' imports class 'org.gradle.api.internal.artifacts.transform.TransformStepNode' from a forbidden platform 'software'
```
tresat, line 156: This violation was already present, I just refactored some functionality into the parent codec to deduplicate.

platforms/core-configuration/dependency-management-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/dm/transform/AbstractTransformStepNodeCodec.kt
```diff
@@ -16,26 +16,71 @@
...
 import org.gradle.internal.serialize.graph.Codec
 import org.gradle.internal.serialize.graph.ReadContext
 import org.gradle.internal.serialize.graph.WriteContext
 import org.gradle.internal.serialize.graph.decodePreservingSharedIdentity
 import org.gradle.internal.serialize.graph.encodePreservingSharedIdentityOf
+import org.gradle.internal.serialize.graph.readNonNull
 
 
-abstract class AbstractTransformStepNodeCodec<T : TransformStepNode> : Codec<T> {
+abstract class AbstractTransformStepNodeCodec<T : TransformStepNode>(
```
tresat, line 33: Many of the changes in this class and the other Chained and Initial codecs are just moving common functionality down to the base class.

The addition of `writeBoolean(value.wasScheduledViaTaskDependency())` and the reader for it is the only new functionality here.

lkasso (approved):
Docs LGTM. Added a cross-link.

platforms/core-runtime/logging/src/integTest/groovy/org/gradle/internal/logging/LoggingIntegrationTest.groovy
```diff
@@ -181,6 +182,11 @@ class LoggingIntegrationTest extends AbstractIntegrationSpec {
         List<String> allArgs = logLevel.args + ['-I', initScript]
 
         when:
+        // The nestedBuildLog GradleBuild task triggers the undeclared-transform deprecation
+        // at lifecycle level; not visible at quiet level
+        if (level != 'quiet') {
+            expectUndeclaredArtifactTransformInputDeprecation()
```
big-guy: Does this mean every GradleBuild task triggers undeclared transforms?

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -616,6 +616,65 @@ tasks.register("okay") {
...
 
+[[undeclared_artifact_transform_input]]
+==== Deprecation of querying an undeclared artifact transform output from a task action
+
+Reading the files of a `FileCollection` whose artifacts are produced by an artifact transform of a project artifact, from inside a task action that does not declare those files as a task input, is now deprecated and will become an error in Gradle 10.
+
+Without the Configuration Cache, this pattern silently runs the transform inline on the calling thread.
+With the Configuration Cache, the producer project's state is not reachable from a task action that did not declare the transform as an input, and the first build fails with `project <name> not found`.
+
+To fix this, declare the artifact view as a task input — either via `inputs.files(...)` on the `FileCollection` exposed by `ArtifactView.getFiles()`, or, when consuming an `ArtifactCollection`, via `inputs.files(...)` on `ArtifactCollection.getArtifactFiles()`. Both wire the transform steps into the execution plan. You can also expose the collection through a typed `@InputFiles` property on a custom task type.
```
big-guy: I would reverse these. We don't want to make it seem like the runtime API (`input.files(...)`) is the better path.

Outside of our tests, tasks that are using ArtifactCollections/ArtifactViews/artifact transforms are already in the advance category, so we should be encouraging people to write correct tasks vs adhoc things.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/api/ResolvedArtifactsApiIntegrationTest.groovy
```diff
@@ -611,10 +611,10 @@ class ResolvedArtifactsApiIntegrationTest extends AbstractHttpDependencyResoluti
             }
 
             task show {
-                inputs.files configurations.compile
                 def artifacts = configurations.compile.incoming.artifactView {
                     attributes({it.attribute(usage, 'transformed')})
                 }.artifacts
+                inputs.files artifacts.artifactFiles
```
big-guy, line 617: Would `inputs.files(artifacts)` not work?

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformBuildOperationIntegrationTest.groovy
```diff
@@ -1328,6 +1329,10 @@ class ArtifactTransformBuildOperationIntegrationTest extends AbstractIntegration
         """
 
         when:
+        // The view is wired via files.from(view) without inputs.files(view), so the chained
+        // transform steps are not scheduled in the plan. Reading files.files in the task action
+        // runs them inline — exactly the case the new undeclared-transform deprecation flags.
+        expectUndeclaredArtifactTransformInputDeprecation()
```
big-guy: The other way of saying this is that this test relies on undeclared artifact transforms.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
```diff
@@ -1661,6 +1662,10 @@ Found the following transformation chains:
         outputContains("files: [jar1.jar, jar2.jar]")
 
         when:
+        // queryView's doLast iterates configView without declaring inputs.files(configView),
+        // so the project-artifact transforms run inline at execution time — triggers the
+        // new undeclared-transform deprecation.
+        expectUndeclaredArtifactTransformInputDeprecation()
```
big-guy: I don't think this test is intentionally relying on running these transforms this way. we can probably just fix the tasks to declare the dependency.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/transform/TransformedProjectArtifactSet.java
```diff
@@ -98,6 +107,36 @@ public void visit(ArtifactVisitor visitor) {
         visitor.endVisitCollection(this);
     }
 
+    private void nagIfUndeclared(TransformStepNode node) {
+        if (node.wasScheduledViaTaskDependency() || !workExecutionTracker.isExecutingTaskOrTransformAction()) {
```
big-guy: when would `workExecutionTracker.isExecutingTaskOrTransformAction()` be false?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/transform/TransformedProjectArtifactSet.java
```diff
@@ -98,6 +107,36 @@ public void visit(ArtifactVisitor visitor) {
...
+            return;
+        }
+
+        //noinspection OptionalGetWithoutIsPresent as isExecutingTaskOrTransformAction() guarantees a current task is on the stack.
+        String taskPath = workExecutionTracker.getCurrentTask().get().getPath();
```
big-guy, line 119: Could a chain of transforms get here without current task?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/transform/TransformedProjectArtifactSet.java
```diff
@@ -84,6 +92,7 @@ public void startFinalization(BuildOperationQueue<RunnableBuildOperation> action
     public void visit(ArtifactVisitor visitor) {
         DisplayName artifactSetName = Describables.of(targetVariant.getComponentId());
         for (TransformStepNode node : transformedArtifacts) {
+            nagIfUndeclared(node);
             node.executeIfNotAlready();
```
big-guy, line 97: In the cases that we've declared the input upfront, wouldn't the node already be executed at this point?

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildDeclaredSubstitutionsIntegrationTest.groovy
```diff
@@ -389,6 +390,7 @@ class CompositeBuildDeclaredSubstitutionsIntegrationTest extends AbstractComposi
         buildB.file("buildSrc/build.gradle").touch()
 
         then:
+        expectUndeclaredArtifactTransformInputDeprecation()
```
big-guy: What's the undeclared dependency here?

platforms/core-runtime/logging/src/integTest/groovy/org/gradle/internal/logging/LoggingIntegrationTest.groovy
tresat: It did.  I tried to solve this by avoiding this during internal ArtifactTransform access.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/api/ResolvedArtifactsApiIntegrationTest.groovy
tresat: No.  `artifacts` is not a file collection, it's an `ArtifactCollection`.

```
> Cannot convert the provided notation to a File: test-lib.jar.
  The following types/formats are supported:
    - A String or CharSequence path, for example 'src/main/java' or '/usr/include'.
    - A String or CharSequence URI, for example 'file:/usr/include'.
    - A File instance.
    - A Path instance.
    - A Directory instance.
    - A RegularFile instance.
    - A URI or URL instance of file.
    - A TextResource instance.
```

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformBuildOperationIntegrationTest.groovy
tresat: Yeah, that's wordy.  Updated.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/transform/TransformedProjectArtifactSet.java
tresat: isExecutingTaskOrTransformAction() returns false (we skip the deprecation) when the visit happens outside any task or transform action stack. Several real scenarios:

1. Configuration-time iteration
configurations.foo.incoming.artifactView { ... }.files.each { println it }   // top-level in build script
Any access during project evaluation — top-level statements, afterEvaluate { }, plugin apply blocks — runs without a current task on the build-op ancestry.

2. Task graph callbacks before execution starts
gradle.taskGraph.whenReady { graph ->
    configurations.foo.files   // resolves now, not inside a task action
}
The graph-ready callback runs after the task graph is computed but before any task action runs. Same with gradle.buildFinished { }, gradle.projectsEvaluated { }, etc.

3. Tooling-API queries
External tools (IDE imports, build scans, etc.) sometimes query configurations.x.files via TAPI. These can run during model building, which is outside task execution.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/transform/TransformedProjectArtifactSet.java
tresat: Going to try a more comprehensive fix in another branch.

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildDeclaredSubstitutionsIntegrationTest.groovy
tresat: This was a false positive due to `buildSrc`, corrected with the latest additional check.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/transform/TransformedProjectArtifactSet.java
tresat: There is a test for chained transforms.  I don't think this case is significantly different - we check at each transform step.

tresat:
Superceded by: 
- https://github.com/gradle/gradle/pull/38173

PR #38108, opened 2026-06-04
Remove lock check in Gradle.getRootProject implementations

octylFractal:
This can cause issues in IP due to parallel CC executing user code that tries to get the root project lock while holding that lock in another thread. Sometimes the holder of the root project lock is blocking on some other lock which is held by the user code, causing a deadlock.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStoreDeadlockIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStoreDeadlockIntegrationTest.groovy
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl.isolated
+
+import org.gradle.api.provider.ValueSource
+import org.gradle.api.provider.ValueSourceParameters
+import org.gradle.test.fixtures.server.http.BlockingHttpServer
+import org.junit.Rule
+import spock.lang.Issue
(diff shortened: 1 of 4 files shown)


platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStoreDeadlockIntegrationTest.groovy
```diff
@@ -0,0 +1,112 @@
...
+        // At store time the captured project resolves to the real DefaultGradle, so the read hits
+        // DefaultGradle.getRootProject(): removing the lock it took there breaks the cycle.
+        //
+        // We force the interleaving deterministically with a shared latch in buildSrc. Writing a shared
+        // object only needs the owning project's lock, so the subproject writer can reach the monitor
+        // and signal even while the root writer parks on the latch holding the root lock.
+        given:
+        def configurationCache = newConfigurationCacheFixture()
+
+        file("buildSrc/src/main/java/Coordinator.java") << """
```
jvandort: Instead of using `static` state to coordinate different threads, can we use a `BlockingHttpServer`?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStoreDeadlockIntegrationTest.groovy
```diff
@@ -0,0 +1,112 @@
...
+                    Property<String> getRootName()
+                }
+                @Override String obtain() { parameters.rootName.get() }
+            }
+
+            def source = providers.of(RootNameSource) {
+                // Computed while this writer holds the monitor (the value source is a shared object).
+                parameters.rootName = provider {
+                    Coordinator.SUBPROJECT_HOLDS_MONITOR.countDown()
+                    Thread.sleep(2000)
```
jvandort: How come this sleep is necessary?

Can we use two latches to make sure both locks are acquired before trying to access the root project?

jvandort (approved):
LGTM other than potentially using a `BlockingHttpServer`

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStoreDeadlockIntegrationTest.groovy
octylFractal: Yeah, I keep forgetting about that, despite how often it causes me troubles when it introduces strange errors...

PR #38114, opened 2026-06-05
Polish lambda display in CC-report

ov7a:
Follow-up to
* https://github.com/gradle/gradle/pull/38033

Frontend:
* https://github.com/gradle/configuration-cache-report/pull/56

Here's how it looks:
image

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/SerializedLambdaParametersCheckingCodec.kt
mlopatkin: 🤔 AFAICT, `SerializedLambda` exposes all its state through getters, so we can replace the bean call with explicit teardown and rebuild it through constructor on the other side. What it gives you is that you can use `PropertyTrace.VirtualProperty` for `capturedArgs` instead of doing fragile conditions on field names.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheLambdaIntegrationTest.groovy
```diff
@@ -428,7 +428,7 @@ class ConfigurationCacheLambdaIntegrationTest extends AbstractConfigurationCache
         problems.htmlReport(failure.error).assertContents {
             problemsWithStackTraceCount = 0
             withProblem("cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.") {
-                at(":ok").at("supplier").at("lambda my.LambdaTask.lambda\$captureProject\$")
+                at(":ok").at("supplier").at("lambda of type java.util.function.Supplier returning java.lang.String").at("captured arguments of my.LambdaTask.captureProject")
```
mlopatkin: 🤔 Maybe also add a test for a bound method reference? E.g. `project::getName` for a supplier. AFAIR, these have a different layout.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/SerializedLambdaParametersCheckingCodec.kt
```diff
@@ -94,4 +95,12 @@ object SerializedLambdaParametersCheckingCodec : Codec<SerializedLambda> {
...
     val unsupportedTypes: Map<Type, KClass<*>> =
         unsupportedFieldDeclaredTypes.associateBy { Type.getType(it.java) }
+
+
+    private
+    fun sourceMethodName(implMethodName: String): String =
+        SYNTHETIC_LAMBDA_METHOD.matchEntire(implMethodName)?.groupValues?.get(1) ?: implMethodName
+
+    private
+    val SYNTHETIC_LAMBDA_METHOD = Regex("""lambda\$(.+)\$\d+""")
```
mlopatkin: ❓ It looks like you're relying on specific mangling used for the lambda method names. It doesn't feel robust, TBH. At the very least, lets document what's going on here.

mlopatkin (changes requested):
Nice improvement, but I'd vote for the `VirtualProperty` approach.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/SerializedLambdaParametersCheckingCodec.kt
ov7a: 🤔 Does `VirtualProperty` have support for structured data? IIUC, it's always rendered as 
```
  text("${node.name} of ")
  ref(node.owner)
```
so it wouldn't be possible to provide copyable/highlighted details about method, lambda type, etc?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheLambdaIntegrationTest.groovy
ov7a: Good catch, will do

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/SerializedLambdaParametersCheckingCodec.kt
ov7a: 🤔 But that's standard Java mangling, right? It will fall back to the plain method name. What do you want to cover with docs?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheLambdaIntegrationTest.groovy
ov7a: image

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/SerializedLambdaParametersCheckingCodec.kt
mlopatkin: > 🤔 But that's standard Java mangling, right? It will fall back to the plain method name. What do you want to cover with docs?

I'm pretty sure the names are up to implementation. By documenting I meant, add a comment "lambda name for something defined in Foo.bar() looks like this, as of Java 8-25. This regex extracts the enclosing method name back". Otherwise It is hard to trace the regex back to the idea behind it.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/SerializedLambdaParametersCheckingCodec.kt
mlopatkin: > 🤔 Does `VirtualProperty` have support for structured data? IIUC, it's always rendered as
> 
> ```
>   text("${node.name} of ")
>   ref(node.owner)
> ```
> 
> so it wouldn't be possible to provide copyable/highlighted details about method, lambda type, etc?

Yes, you're right, it is too simplistic. I forgot that it doesn't unfold the owner completely.

mlopatkin: > > 🤔 Does `VirtualProperty` have support for structured data? IIUC, it's always rendered as
> > ```
> >   text("${node.name} of ")
> >   ref(node.owner)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so it wouldn't be possible to provide copyable/highlighted details about method, lambda type, etc?
> 
> Yes, you're right, it is too simplistic. I forgot that it doesn't unfold the owner completely.

Nevertheless, it should be possible to move the special `kind: CapturedArguments` logic into this Codec and not have it in the `writePropertyTrace`. Also, the logic about extracting the lambda name doesn't belong to json-writing methods.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/SerializedLambdaParametersCheckingCodec.kt
```diff
@@ -48,17 +52,24 @@ object SerializedLambdaParametersCheckingCodec : Codec<SerializedLambda> {
...
+            implMethodName = value.implMethodName,
+            functionalInterfaceClass = value.functionalInterfaceClass.replace('/', '.'),
+            instantiatedReturnType = Type.getReturnType(value.instantiatedMethodType).className,
+            trace = trace
+        )
+        withPropertyTrace(lambdaTrace) {
             writeClass(SerializedLambda::class.java)
-            beanStateWriterFor(SerializedLambda::class.java).run {
-                writeStateOf(value)
+            for (field in SERIALIZED_LAMBDA_FIELDS) {
```
mlopatkin: ❓ TBH, I had something like
```kotlin
// SerializedLambda exposes the capturing class only as a string getter, so the actual Class<?> required
// by the constructor is read via the private field — preserving the writeClass/readClass classloader scope.
writeClass(capturingClassField.get(value) as Class<*>)
writeString(value.functionalInterfaceClass)
writeString(value.functionalInterfaceMethodName)
writeString(value.functionalInterfaceMethodSignature)
writeSmallInt(value.implMethodKind)
writeString(value.implClass)
writeString(value.implMethodName)
writeString(value.implMethodSignature)
writeString(value.instantiatedMethodType)
withPropertyTrace(lambdaTrace.toCapturedArguments()) {
    val capturedArgCount = value.capturedArgCount
    writeSmallInt(capturedArgCount)
    for (i in 0 until capturedArgCount) {
        write(value.getCapturedArg(i))
    }
}
```
in mind. Of course, we have to resort to reflection to get the capturing class, but everything else is readily available.

ov7a: 🤔 Previously, it was a bean codec. What's the benefit of unfolding all the fields? It feels a bit fragile in case SerializedLambda changes. SERIALIZED_LAMBDA_FIELDS is fragile too, to our own code changes, but we have more control over them.

mlopatkin: >What's the benefit of unfolding all the fields?

I'm a bit averse towards the reflective lookup, and we're also duplicating the logic of selecting reflective fields here. And we cannot use bean codec because we need to handle captured args separately without leaking this outside of the lambda codec. If that doesn't convince you, I'm happy to close this a "matter of taste" and move forward with your implementation.

>It feels a bit fragile in case SerializedLambda changes

The class and its serialized form are part of the JDK compatibility contract, so I wouldn't expect any major changes there. Additionally, any changes there are likely to reflect on the constructor, so we'll get a heads up and plenty of time to react.

ov7a: I don't have a strong opinion here and am fine with both outcomes. I'll update the code according to your suggestion

mlopatkin (approved):
This will need new version of the report integrated, but can be merged afterwards.

Issue #31138, opened 2024-11-05
Better error reporting for unsupported collections subtypes in configuration cache


mlopatkin:
Configuration cache serialization has special treatment for some collection implementations:
* `List`s: `ArrayList`, `LinkedList`, `CopyOnWriteArrayList`, guava's `ImmutableList`
* `Set`s: `HashSet`, `TreeSet`, `CopyOnWriteArraySet`, guava's `ImmutableSet`
* `Map`s: `LinkedHashMap`, `HashMap`, `TreeMap`, `ConcurrentHashMap`, `Properties`, `Hashtable`, guava's `ImmutableMap`
* `Queue`: `ArrayDeque`

When a custom descendant of one of these types is serialized, it is restored as a base type, which might be surprising.

We don't intend to support custom descendants of these collection types, so a serialization error should be emitted.

mlopatkin:
Note: it might need to go through a short deprecation cycle first. 

Also, Guava types aren't extensible by users, so they should be safe as they are.

PR #38118, opened 2026-06-05
Report custom collection types during CC serialization

ov7a:
Creates a deprecation for
* https://github.com/gradle/gradle/issues/31138

The issue can be fixed in 10

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheSupportedTypesIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheSupportedTypesIntegrationTest.groovy
@@ -131,7 +131,7 @@ class ConfigurationCacheSupportedTypesIntegrationTest extends AbstractConfigurat
         "LocalDateTime"                      | "LocalDateTime.of(2024, 1, 1, 1, 1)"      | "2024-01-01T01:01"
     }
 
-    def "reports a problem and restores as a standard type a custom descendant #concreteType of supported type #baseType"() {
+    def "serializing a custom descendant #concreteType of supported type #baseType is deprecated but the cache is still stored and reused"() {
         buildFile << """
             class $concreteType extends $baseType {}
 
@@ -148,22 +148,18 @@ class ConfigurationCacheSupportedTypesIntegrationTest extends AbstractConfigurat
         """
 
         when:
-        configurationCacheRunLenient "ok"
+        executer.expectDocumentedDeprecationWarning(
+            "Serializing a custom $kind type '$concreteType', a subtype of '$baseType', which will be restored as a standard $kind, losing any custom behavior. " +
+                "This behavior has been deprecated. This will fail with an error in Gradle 10. " +
+                "For more information, please refer to https://docs.gradle.org/current/userguide/configuration_cache_requirements.html#config_cache:requirements:custom_collection_types in the Gradle documentation."
+        )
+        configurationCacheRun "ok"
 
         then:
-        problems.assertResultHasProblems(result) {
-            totalProblemsCount = 1
-            withUniqueProblems(
-                "Task `:ok` of type `SomeTask`: cannot fully support serializing a custom $kind type '$concreteType', a subtype of '$baseType': it will be restored as a standard $kind, losing any custom behavior. This will become an error in a future Gradle version"
-            )
-            problemsWithStackTraceCount = 0
-        }
-
-        and: "the cache is stored and then loaded, so the task already runs against the restored value"
         outputContains("value type = $restoredType")
 
-        when: "the entry is reused, the value is restored as a standard type and nothing is serialized"
-        configurationCacheRunLenient "ok"
+        when:
+        configurationCacheRun "ok"
 
         then:
         outputContains("value type = $restoredType")
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheSupportedTypesIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheSupportedTypesIntegrationTest.groovy
@@ -26,7 +26,6 @@ import org.gradle.internal.event.ListenerManager
 import org.gradle.process.ExecOperations
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.JdkVersionTestPreconditions
-
 import org.gradle.tooling.provider.model.ToolingModelBuilderRegistry
 import org.gradle.workers.WorkerExecutor
 import org.slf4j.Logger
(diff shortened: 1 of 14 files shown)


platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/Combinators.kt
```diff
@@ -366,3 +377,25 @@ fun WriteContext.reportCollectionWriteFailure(collectionKind: String, size: Int,
...
+    if (actualType in supportedTypes) {
+        return
+    }
+    val baseType = supportedTypes.firstOrNull() ?: return
+    logPropertyProblem("serialize", DocumentationSection.RequirementsDisallowedTypes) {
+        text("cannot fully support serializing a custom $kind type ")
+        reference(actualType)
+        text(", a subtype of ")
+        reference(baseType)
+        text(": it will be restored as a standard $kind, losing any custom behavior. This will become an error in a future Gradle version")
```
ov7a: 🤔 
1. What's the best way to convey that this behavior is deprecated?
2. Should new problems be mentioned in the upgrade guide?

ov7a: One idea by @mlopatkin is to use `Suppressed`, but  it's currently hardwired
https://github.com/gradle/gradle/blob/300d36e7c704406baa9d56e4e01ed9e86984cfbe/platforms/core-configuration/configuration-problems-base/src/main/kotlin/org/gradle/internal/cc/impl/problems/ProblemSeverity.kt#L42-L45 to `notCompatibleWithConfigurationCache`

ov7a: I ended up creating a custom severity level in order to not to fail the build, and have a problem reported exactly once (by deprecation logger) in the console

lkasso (approved):
Docs LGTM.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheSupportedTypesIntegrationTest.groovy
```diff
@@ -132,6 +131,56 @@ class ConfigurationCacheSupportedTypesIntegrationTest extends AbstractConfigurat
...
+                    println "value type = " + value.getClass().name
+                }
+            }
+
+            task ok(type: SomeTask)
+        """
+
+        when:
+        executer.expectDocumentedDeprecationWarning(
+            "Serializing a custom $kind type '$concreteType', a subtype of '$baseType', which will be restored as a standard $kind, losing any custom behavior. " +
```
tresat: 🤔 Maybe we should say "data and behavior" explicitly, since any new fields in the sub-type are also lost.

platforms/core-configuration/configuration-problems-base/src/main/kotlin/org/gradle/internal/cc/impl/problems/ProblemSeverity.kt
```diff
@@ -30,6 +30,14 @@ enum class ProblemSeverity {
      */
     Deferred,
 
+    /**
+     * A problem reporting a deprecated behavior.
+     *
+     * Does not prevent CC from being reused.
+     * It is additionally surfaced as a Gradle deprecation warning.
+     */
+    DeprecatedBehavior,
```
tresat: ❗ I don't think we should be routing these deprecations through Problems like this.  We should probably continue to keep deprecation reporting and problem reporting separate  The TIDE team should consider if this is something that should be added here, as it seems like a larger change.  Until then I think we should avoid doing this.

platforms/core-configuration/graph-serialization/src/main/kotlin/org/gradle/internal/serialize/graph/Combinators.kt
```diff
@@ -366,3 +377,34 @@ fun WriteContext.reportCollectionWriteFailure(collectionKind: String, size: Int,
...
+
+
+/**
+ * Reports a configuration cache problem when [actualType] is a custom descendant of one of the [supportedTypes]
+ * a codec serializes specially. The codec is only reached because dispatch matched [actualType] to one of these
+ * base types, so anything that is not exactly one of them is a custom descendant that will be restored as a
+ * standard [kind] (collection/map) — losing its custom behavior. The first supported type is used as the base
+ * type in the message (e.g. `HashSet` for `HashSet`/`LinkedHashSet`).
+ */
+fun IsolateContext.warnAboutCustomImplementation(actualType: Class<*>, kind: String, supportedTypes: Array<out Class<*>>) {
```
tresat: It would be good to figure out what we're doing with https://github.com/gradle/gradle/pull/37786 at the same time, as these are related considerations and should probably be addressed together in a unified fashion.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheSupportedTypesIntegrationTest.groovy
```diff
@@ -221,6 +270,44 @@ class ConfigurationCacheSupportedTypesIntegrationTest extends AbstractConfigurat
...
         outputContains("bean.value = ${output}")
 
+        where:
+        type          | reference                         | output
+        ImmutableList | "ImmutableList.of('a', 'b', 'c')" | "[a, b, c]"
+        ImmutableSet  | "ImmutableSet.of('a', 'b', 'c')"  | "[a, b, c]"
+        ImmutableMap | "ImmutableMap.of(1, 'a', 2, 'b')" | "[1:a, 2:b]"
+    }
+
+    def "serializing a Guava #type.simpleName does not report a problem"() {
```
tresat, line 282: 💭 It might be nice to check a different Guava type or types in the parameterized failure case test above and ensure they also fail.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -129,6 +129,24 @@ Sometimes, a disallowed type is indirectly referenced through another type.
...
+The Configuration Cache serializes a fixed set of standard collection, set, map, and queue implementations directly:
+
+* `List`: `ArrayList`, `LinkedList`, `CopyOnWriteArrayList`
+* `Set`: `HashSet`, `LinkedHashSet`, `TreeSet`, `CopyOnWriteArraySet`
+* `Map`: `HashMap`, `LinkedHashMap`, `TreeMap`, `ConcurrentHashMap`, `Properties`, `Hashtable`
+* `Queue`: `ArrayDeque`
+
+These types, and Guava's immutable collections (`ImmutableList`, `ImmutableSet`, `ImmutableMap`), are stored and restored faithfully.
+
+A _custom_ subtype of one of the standard types above (for example, `class MyList extends ArrayList<String> {}`) cannot be restored as its own type: it is restored as the nearest standard type instead.
```
tresat, line 159: This change is all about extensions of already concrete types.

What about new implementations of interfaces (`class MyList extends List`)?

tresat (dismissed):
I'm not sure we should be routing these deprecations through the Problems API.  Other than that, only minor suggestions.

platforms/core-configuration/configuration-problems-base/src/main/kotlin/org/gradle/internal/cc/impl/problems/ProblemSeverity.kt
```diff
@@ -30,6 +30,14 @@ enum class ProblemSeverity {
...
+
```
abstratt: @ov7a I like the idea of what you are proposing, but I agree with Tom this should be tackled separately (at the expense of having those CC-related issues not appear on the report for now).

@mlopatkin is going to open an issue about CC-related deprecations vs CC report. We could use your approach here to address that, at a later time though.

ov7a: Cool, thanks. I'll wait for @mlopatkin PR then, and once the infra is ready, I'll just rebase this PR on top of it (or will open a new one, if you have a preference)

abstratt (changes requested):
I like the approach, and we should do something like that at some point, but for now, I propose we stick with vanilla deprecations for now

platforms/core-configuration/configuration-problems-base/src/main/kotlin/org/gradle/internal/cc/impl/problems/ProblemSeverity.kt
abstratt: Sorry, I meant to say Mike is opening an issue to be tackled later, while the core of your changes (i.e. minus deprecations as CC report problems) here could hit master now

ov7a: Ah, okay, got you. Sorry for the confusion.

abstratt (approved):
Looks good!

PR #38122, opened 2026-06-05
Pull out project lock from being used for all state mangement

octylFractal:
This results in some unsatisfactory reductions in concurrency due to children needing to check parents are configured.

Fixes https://github.com/gradle/gradle-private/issues/5248

Note that I also adjusted `ensureTasksDiscovered` as I realized it was also unprotected.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectState.java
```diff
@@ -242,19 +248,56 @@ public void applyToMutableState(Consumer<? super ProjectInternal> action) {
...
+            } else {
+                throw new IllegalStateException("Current thread holds more than one project lock. It should hold only one project lock at any given time.");
+            }
+        } else {
+            return workerLeaseService.withReplacedLocks(currentLocks, projectLock, action::get);
+        }
     }
 
     @Override
     public <S> S forceAccessToMutableState(Function<? super ProjectInternal, ? extends S> factory) {
```
jvandort, line 281: Obviously not to address here, but we should figure out what the heck this method is for

jvandort (approved):

PR #38127, opened 2026-06-06
Infer locations for more problems past the stack capture cap

### LINKED ISSUE #32362: Allow capturing stacktraces for problems when problem count is very large
### Expected Behavior

As a user of Gradle with a large project (e.g. 100 subprojects) I am able to view stacktraces for all the problems shown in problems HTML report and Develocity "deprecated" tab.

It could require a flag `--capture-all-deprecation-stack-traces` or `--capture-deprecation-stack-traces-for-issue-id "Space-assignment syntax in Groovy DSL has been deprecated."`

Gradle has a built in limit of 50 stacktraces that will be captured and any additional ones have no location information.

Example scan: https://ge.androidx.dev/s/s7ulkwj7mbtqg/deprecations#11
Example problems report https://ci.android.com/builds/submitted/13025144/androidx/latest/problem-reports%2Fproblems-report.html

This means that if you want to try to fix these issues, it is extremely difficult to find where to start. E.g. 

Image

you have no indication on where in your large build to even get started on.

### Context

We were affected by https://github.com/gradle/gradle/issues/31963 / https://youtrack.jetbrains.com/issue/KT-73968/ that produces 4 thousand problems in our build making it very difficult to fix other problems in our list.

eskatos:
### Why
The Problems API caps how many full stack traces it captures (for cost), so any problem reported past that cap without an explicit location carried no inferred source location. On large builds that emit thousands of problems, most lost their file and line, degrading the console output, the HTML report, and the Tooling API problem model.

* Fixes gradle/gradle#32362

### What changes
Problems reported past the cap now resolve to their originating file and line, up to a bounded budget per build. This raises location coverage well past the existing capture cap (50, or 5000 under Isolated Projects) by a further 2000 problems, at negligible cost. Best effort, console/report/TAPI only: no new public API, and the behaviour within the cap is unchanged.

The trace captured past the cap is reduced: it keeps the build logic that raised the problem, down to the calling script, and drops the Gradle runtime frames below it. So a problem that originates in shared build logic shows both where it surfaced (the script) and the build logic that caused it, while the deep runtime frames are trimmed. It resolves the correct location and keeps the calling context, but it is not a complete trace.

Beyond the budget, problems carry no inferred location, as before; `--warning-mode=all` lifts the limit entirely.

### Implementation
Past the cap, rather than capturing nothing, a new service walks a cheap bounded stack prefix with the Java 9+ `StackWalker` API, enough for the existing location analyser to infer the location. The walk keeps the frames from the top down to the script the location resolves to, seeing through compiled helpers (buildSrc or plugin code) to the
calling script, and stops there, so the runtime frames below the script are never walked or kept.

Each problem is resolved independently from its own live stack, so two problems that share a helper frame but originate in different scripts cannot be attributed to the same location. The capturer and the analyser share a small registry of script sources so they agree on which frames are scripts. A budget caps how many of these captures a build performs, so the total cost stays fixed regardless of build size.

### Performance
Bounded and negligible. The budget matters because an IDE sync builds project models repeatedly and re-fires the same deprecations tens of thousands of times; uncapped, each one walks the stack again and added ~15% to configuration time on the `largeAndroidBuild` sync, a real regression. Capping the captures at 2000 clears it while keeping a location for every problem within the budget, which covers a full `largeAndroidBuild` configuration (~1.4k problems). The residual cost is a fixed ceiling of ~2000 bounded stack walks, negligible against any large build's sync and flat regardless of size.

* https://github.com/gradle/gradle-private/issues/5253

### Documentation
A release note, plus the `--warning-mode` CLI reference and the problems-report userguide page, document the location limit and that `--warning-mode=all` lifts it.

### How to review
Best read commit by commit. The core is the bounded capture: a new module and the past-cap path in `DefaultProblemDiagnosticsFactory` that, beyond the cap, walks a reduced caller stack instead of capturing nothing. The reduced stack keeps the build logic above the resolved script and drops the runtime below it; the location still comes from the existing analyser, via a script registry shared with the capturer. A later change adds
the budget that bounds the cost (the response to the performance review). `ProblemLocationPastCapIntegrationTest` exercises the bounded path end to end (a capping warning mode, distinct problem ids, a compiled build logic helper) and pins that problems past the cap resolve to the calling script with the build logic kept above it.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


eskatos:
@bot-gradle test RFN APT

eskatos:
@bot-gradle test RFN APT

eskatos:
@bot-gradle test RFN

eskatos:
@bot-gradle test RFN

testing/performance/src/templates/generateDiverseDeprecationWarnings/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+
+int numSites = 50
+int perSite = 200
+int baseDepth = 30
+
+File sitesDir = new File(projectDir, "sites")
+sitesDir.mkdirs()
+
+for (int n = 0; n < numSites; n++) {
+    int depth = baseDepth + (n % 16)
```
hegyibalint, line 36: 🤔 What is the reason for fluctuating the `depth` as we iterate?

testing/performance/src/templates/generateDiverseDeprecationWarnings/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+    if (depth > 1) {
+        fire(depth - 1, message)
+        return
+    }
+    org.gradle.internal.deprecation.DeprecationLogger.deprecate(message).willBecomeAnErrorInGradle10().withUserManual("feature_lifecycle", "sec:deprecated").nagUser()
+}
+
+for (int i = 0; i < ${perSite}; i++) {
+    fire(${depth}, "Some repetitive deprecation from site ${n}")
+    if (i % 10 == 0) {
```
hegyibalint: 🤔 I think this `10` is an important hyperparameter, as IIUC this will test how well the deduplication would work. I would consider moving this up to the top of the file.

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -76,6 +76,11 @@ Gradle provides a [set of features and abstractions](userguide/java_testing.html
 ### CLI, logging, and problem reporting
 Gradle provides an intuitive [command-line interface](userguide/command_line_interface.html), detailed [logs](userguide/logging.html), and a structured [problems report](userguide/reporting_problems.html#sec:generated_html_report) that helps developers quickly identify and resolve build issues.
 
+#### Source locations for more problems
+
+To keep stack trace capture affordable, Gradle used to stop attaching a source location to problems past a low limit, so builds that report many problems (deprecations especially) left most of them without a file and line.
+It now infers a location for problems past that limit too, so the console, the [problems report](userguide/reporting_problems.html#sec:generated_html_report), and the Tooling API carry a source location for many more of them.
+
```
lkasso, line 158: Would it be possible to add a screenshot of this?

lkasso (approved):
Docs LGTM, but release notes would be higher impact with a screenshot/image.

platforms/documentation/docs/src/docs/release/notes.md
eskatos: Good idea, screenshot added

image

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
```diff
@@ -0,0 +1,126 @@
...
+ * location, so the resolved prefix is cached per site. The expensive resolution is then paid once per
+ * site rather than once per problem, which is what keeps capturing locations past the cap affordable.</p>
+ *
+ * <p>The cache is keyed on the located frame (the first user frame with a line) because that is the
+ * frame the location is taken from. Keying on the nearest user frame instead would conflate distinct
+ * sites that happen to share a line-less synthetic nearest frame, attributing one to the other.</p>
+ */
+public class DefaultBoundedCallerStackCapturer implements BoundedCallerStackCapturer {
+
+    // Walk cap when no user frame. Above measured depth-to-user (~13-37).
```
hegyibalint: > Above measured depth-to-user (~13-37).

🤔 Where does this number come from?

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
```diff
@@ -0,0 +1,126 @@
...
+            if (site == null && element.getLineNumber() >= 0 && !InternalStackTraceClassifier.isInternal(className)) {
+                // Located frame: where the location comes from. Key on it; hit reuses its prefix.
+                site = siteKey(element);
+                StackTraceElement[] cached = prefixBySite.getIfPresent(site);
+                if (cached != null) {
+                    return cached;
+                }
+            }
+            if (InternalStackTraceClassifier.isGradleCall(className)) {
+                break; // include Gradle boundary, stop
```
hegyibalint: 💄 

```suggestion
                break; // reached Gradle infrastructure; include it, and stop
```

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
```diff
@@ -0,0 +1,126 @@
...
+    public Throwable captureCallerStack() {
+        StackTraceElement[] prefix = WALKER.walk(this::cachedBoundedPrefix);
+        return prefix == null ? null : new BoundedStackHolder(prefix);
+    }
+
+    @VisibleForTesting
+    StackTraceElement @Nullable [] cachedBoundedPrefix(Stream<StackFrame> frames) {
+        Iterator<StackFrame> iterator = frames.iterator();
+        List<StackTraceElement> prefix = null; // from first user frame
+        String site = null;                    // located-frame key (first user frame with a line)
```
hegyibalint: 💄 I am just ideating how the command could be a bit more descriptive

```suggestion
        List<StackTraceElement> prefix = null; // stack frames starting from the user call-site
        String site = null;                    // string representation of the first user callsite with a line number
```

platforms/ide/problems-impl/src/test/groovy/org/gradle/internal/problems/impl/DefaultProblemLocationAnalyzerTest.groovy
```diff
@@ -54,4 +55,38 @@ class DefaultProblemLocationAnalyzerTest extends Specification {
...
+        def scriptFrame = new StackTraceElement("Build_gradle", "run", "build.gradle", 7)
+        def fullStack = [
+            new StackTraceElement("org.gradle.internal.problems.impl.DefaultBoundedCallerStackCapturer", "captureCallerStack", "DefaultBoundedCallerStackCapturer.java", 40),
+            new StackTraceElement("org.gradle.internal.deprecation.DeprecationLogger", "nagUserWith", "DeprecationLogger.java", 300),
+            scriptFrame,
+            new StackTraceElement("org.gradle.api.internal.project.DefaultProject", "evaluate", "DefaultProject.java", 100),
+            new StackTraceElement("org.gradle.launcher.Main", "main", "Main.java", 50)
+        ] as StackTraceElement[]
+        // What the bounded walk keeps: from the top down to and including the Gradle boundary after the first user frame.
+        def boundedPrefix = (fullStack[0..3]) as StackTraceElement[]
```
hegyibalint: ❓  Huh. This is strange; based on the code, I expected [2] and [3].
IIUC the algorithm we:
 - Skip the leading machinery, so [0] and [1] drop out.
 - We meet with [2], who is going to start the `prefix`.
 - We will keep the [3]; IIUC, we keep the first internal frame exiting the user-portion of the stack.

What am I missing here?

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Nullable
+    @Override
+    public Throwable captureCallerStack() {
+        StackTraceElement[] prefix = WALKER.walk(this::cachedBoundedPrefix);
+        return prefix == null ? null : new BoundedStackHolder(prefix);
+    }
+
+    @VisibleForTesting
+    StackTraceElement @Nullable [] cachedBoundedPrefix(Stream<StackFrame> frames) {
```
hegyibalint: 💭 This method is really hard to read. There is quite the interplay between `site` and `prefix`, and other predicates. Also the control flow is quite jarred: sometimes we continue, sometimes we break, sometimes we fall over.

I think even if it's a bit overengineering here, I would implement this as a finite state machine; it would make the predicates and the state transitions much clearer, even if the implementation needs an extra enum. I think `site` and `prefix` could stay as is; we would just use the FSM to drive the control flow.

---

Or as an alternative, I would order and harmonize the control flow so the inner loop is easier to read, even if it would be a bit more wordy.

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
```diff
@@ -0,0 +1,126 @@
...
+            }
+            StackTraceElement element = frame.toStackTraceElement();
+            if (prefix == null) {
+                prefix = new ArrayList<>(MAX_DEPTH);
+            }
+            prefix.add(element);
+            if (site == null && element.getLineNumber() >= 0 && !InternalStackTraceClassifier.isInternal(className)) {
+                // Located frame: where the location comes from. Key on it; hit reuses its prefix.
+                site = siteKey(element);
+                StackTraceElement[] cached = prefixBySite.getIfPresent(site);
```
hegyibalint: ❓I'm not sure if this early return based on `site` will be correct in all use-cases.

Let's take an imaginary `com.example.Checks.validate`, that is on the script classpath (i.e., a buildSrc or plugin utility) that would report a problem with a stacktrace. 

My theory is that `A`'s cache-line will shadow `B`'s use-site because of the caching.

---

**Invocation A — from `build.gradle:5`**

- `[0] …DefaultBoundedCallerStackCapturer.captureCallerStack`
  - internal → skipped
- `[1] …DefaultProblemStream.forCurrentCaller`
  - internal → skipped
- `[2] …DefaultProblemReporter.report`
  - internal → skipped
- `[3] com.example.Checks.validate (Checks.groovy:12)`
  - user, line ≥ 0 → **site** + prefix start
- `[4] Build_gradle.run (build.gradle:5)`
  - user
- `[5] org.gradle.api.internal.project.DefaultProject.evaluate`
  - `isGradleCall` → add, break

We will add a cache line:
  - `com.example.Checks#validate@12` → `[Checks.groovy:12, build.gradle:5, DefaultProject.evaluate]`

---

**Invocation B — from `other.gradle:9`**

- `[0..2]` (same)
- `[3] com.example.Checks.validate (Checks.groovy:12)`
  - user
  - ❗ cache hit, immediate return with the cached `[Checks.groovy:12, build.gradle:5, DefaultProject.evaluate]`
- `[4] Other_gradle.run (other.gradle:9)`
  - ❌  never walked

---

This can cause problems when we try to put a script location on a problem.
 `Checks.groovy` is a compiled helper (could come from buildSrc or a plugin), not a registered build script, so [`tryGetLocation`](https://github.com/gradle/gradle/blob/28bbc8c28b5aee49fd92391db2282c7c354b2b46/platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultProblemLocationAnalyzer.java#L121) can't resolve the originating frame `[3]` and walks down to the first registered script in the prefix. 

So B's problem is reported at A's cached `build.gradle:5` instead of its own `other.gradle:9`.

hegyibalint (changes requested):
Nice work! I like the overall implementation.
I have some concerns about the caching and potential problems it might cause.

testing/performance/src/templates/generateDiverseDeprecationWarnings/build.gradle
eskatos: It varies the capture depth across sites (staying under the 50-frame cap) so capture cost is exercised across a range of depths rather than one fixed value. Added a comment to that effect.

Note that this is deterministic and doesn't create flaky runs.

testing/performance/src/templates/generateDiverseDeprecationWarnings/build.gradle
eskatos: Good call, lifted it to a top-level `uniqueEvery` constant with a comment on what it tunes.

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
eskatos: It's from measuring the depth-to-user (the leading internal frames before the first user frame) on real builds, `gradle/gradle` and `gradle/nowinandroid`; 50 sits comfortably above the observed max. Added the source builds to the comment.

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
eskatos: Good catch, this was a real bug. The cache keyed on the nearest user frame, but the location resolves by seeing through frames that aren't scripts to the calling script, so two sites sharing such a frame (your `Checks.validate`) collided and the second got the first's `build.gradle:5`.

Fixed by keying the cache (and trimming the reduced stack) on the frame the location actually comes from: the first registered script at or below the located frame. Your A/B example now keys on `build.gradle:5` vs `other.gradle:9` and stays distinct; the walk still sees through to the calling script. The capturer learns which files are scripts from a small `RegisteredScripts` service, now shared with the analyzer.

Added your exact scenario as a unit test (red before the fix, green after) plus an integration test.

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
eskatos: Agreed. Reworked it: the walk state now lives in a small `BoundedWalk` object, so `cachedReducedStack` reads as a plain protocol (accept a frame, maybe a cache hit, maybe the boundary, then the result), and the dense predicates moved into guarded helper methods. The `site`/`prefix` interplay is gone.

I went with harmonizing the control flow rather than a literal enum FSM: the states here only move forward (skip machinery, then anchor, then past the boundary) with no real transitions, so an enum and switch would mostly rename the booleans and add bookkeeping.

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
eskatos: The refactor for fixing the correctness issue and readability made this suggestion moot. PTAL

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
eskatos: The refactor for fixing the correctness issue and readability made this suggestion moot. PTAL

platforms/ide/problems-impl/src/test/groovy/org/gradle/internal/problems/impl/DefaultProblemLocationAnalyzerTest.groovy
eskatos: You're right, the slice was misleading. The bounded walk skips the leading machinery (0, 1) and keeps from the first user frame (the script) down to the Gradle boundary, so the realistic bounded prefix is `[2, 3]`, not `[0..3]`. Fixed the test to use `fullStack[2..3]` and corrected the comment. The location resolves the same either way, which is what the test asserts.

eskatos:
@bot-gradle test RFN APT

eskatos:
@bot-gradle test RFN

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
```diff
@@ -54,52 +59,100 @@ public class DefaultBoundedCallerStackCapturer implements BoundedCallerStackCapt
...
+
+        private StackTraceElement @Nullable [] tryAnchor(StackTraceElement element, String className) {
+            if (anchoredOnScript || element.getLineNumber() < 0 || InternalStackTraceClassifier.isInternal(className)) {
+                return null; // anchor is final, or this is not a user frame with a line
             }
-            prefix.add(element);
-            if (site == null && element.getLineNumber() >= 0 && !InternalStackTraceClassifier.isInternal(className)) {
-                // Located frame: where the location comes from. Key on it; hit reuses its prefix.
-                site = siteKey(element);
-                StackTraceElement[] cached = prefixBySite.getIfPresent(site);
```
hegyibalint: :x: I think this is still broken.

Take the following test as the proof:

```groovy
    def "a call-site with no script below it does not shadow a later call-site that resolves to a script"() {
        given:
        // The same compiled helper first fires with no script frame below it (e.g. from a plugin's apply),
        // and later from a build script. The first, scriptless capture must not shadow the second one.
        def helper = frame('com.example.Checks', 'validate', 1, ste('com.example.Checks', 'validate', 'Checks.groovy', 12))
        def boundary = frame('org.gradle.api.internal.project.DefaultProject', 'evaluate', 1, ste('org.gradle.api.internal.project.DefaultProject', 'evaluate', 'DefaultProject.java', 100))
        def fromPlugin = [helper, boundary]
        def fromScript = [helper, frame('build_gradle', 'run', 5, ste('build_gradle', 'run', 'build.gradle', 5)), boundary]

        when:
        capturer.cachedReducedStack(fromPlugin.stream())
        def script = capturer.cachedReducedStack(fromScript.stream())

        then:
        script*.fileName.contains('build.gradle')
    }
```

---

### 1st invocation: from a plugin, no script on the stack

- `[0] com.example.Checks.validate (Checks.groovy:12)`
  - user, line ≥ 0, not a registered script → tentative **anchor** (`anchoredOnScript = false`), cache miss
- `[1] org.gradle.api.internal.project.DefaultProject.evaluate`
  - `isGradleCall` → add, stop
  - ❗ no script frame ever appeared → the anchor is never upgraded

After the walk,`reducedStack()` will add a cache line, keyed on a frame the location does **not** come from: 
 - `com.example.Checks#validate@12 → [Checks.groovy:12, DefaultProject.evaluate]` 

### 2nd invocation: same helper, from `build.gradle:5`

- `[0] com.example.Checks.validate (Checks.groovy:12)`
  - user, line ≥ 0 → probe cache
  - ❗ cache hit → immediate return of the 1st invocation's stack
- `[1] build_gradle.run (build.gradle:5)`
  - ❌ never walked
- `[2] org.gradle.api.internal.project.DefaultProject.evaluate`
  - ❌ never walked

Here we would expect `build_gradle:5` to be the location, but we get _nothing_, as the cache stack-trace didn't have any frames that led to a script location.

hegyibalint: I think a potential—and simple solution in terms of this implementation—is to _only cache if `anchoredOnScript == true`. 

This, as we don't populate the cache with the non-script entries, would sidestep this problem.

hegyibalint (dismissed):
I think we still have a bug! 🐛

platforms/ide/problems-impl/src/main/java/org/gradle/internal/problems/impl/DefaultBoundedCallerStackCapturer.java
eskatos: Confirmed and fixed. Added your test (red before). The cache now only stores and reuses entries for a definitive script anchor; a non-script frame is used to trim the stack but never cached, since its location depends on the frames below it, so a scriptless capture can't shadow a later one that resolves to a script. Sites that never reach a script just aren't cached (they re-walk), which doesn't affect the script-fired deprecation hot path; perf is flat.

eskatos:
@bot-gradle test RFN

eskatos:
@bot-gradle test RFN

eskatos:
We're back with performance regressions.
I need to take a step back and think, moving back to draft.

eskatos:
@bot-gradle test RFN

eskatos:
@bot-gradle test APT

eskatos:
Heads-up: I've narrowed the scope and rewritten the title and description, worth a re-read.

Profiling the `largeAndroidBuild` IDE sync showed that inferring a location for *every* problem past the cap regressed configuration time ~15% (the model build re-fires the same deprecations tens of thousands of times, and each capture walks the stack). So locations past the cap are now bounded by a budget (2000) rather than unlimited; `--warning-mode=all` still lifts the limit entirely.

The capture/dedup design you reviewed is unchanged. The only new piece is the budget ("Cap bounded location captures past the limit"). Full rationale and the perf evidence are in the updated PR description.

.teamcity/subprojects.json
```diff
@@ -1280,6 +1280,13 @@
     "functionalTests": true,
     "crossVersionTests": false
   },
+  {
+    "name": "problems-impl",
+    "path": "platforms/ide/problems-impl",
```
donat, line 1285: Why do we need a new subproject? We already have the `problems` subproject for the implementation details and `problems-api` for the internal bits.

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -70,52 +70,21 @@ vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv -->
...
 ### Test reporting and execution
 Gradle provides a [set of features and abstractions](userguide/java_testing.html) for testing JVM code, along with test reports to display results.
 
 ### CLI, logging, and problem reporting
 Gradle provides an intuitive [command-line interface](userguide/command_line_interface.html), detailed [logs](userguide/logging.html), and a structured [problems report](userguide/reporting_problems.html#sec:generated_html_report) that helps developers quickly identify and resolve build issues.
 
+#### Source locations for more problems
+
+To keep stack trace capture affordable, Gradle attached a source location to only the first 50 problems per build, so builds that report many problems (deprecations especially) left most of them without a file and line.
+It now captures those locations far more cheaply and raises that limit from 50 to 2000, so the console, the [problems report](userguide/reporting_problems.html#sec:generated_html_report), and the Tooling API show a source location for many more problems than before.
```
donat: Why 2000 specifically? 
Personally, I'm fine with a high upper limit. At the same time, I remember arguments from previous discussions that there should not be a default upper cap. Instead, the build author should be able to configure a limit if necessary.

lkasso (approved):
Docs LGTM.

testing/performance/src/templates/generateDiverseDeprecationWarnings/settings.gradle
donat, line 1: This probably should not be part of the changeset.

testing/performance/src/templates/generateDiverseDeprecationWarnings/gradle.properties
```diff
@@ -0,0 +1,3 @@
+org.gradle.jvmargs=-Xms1G -Xmx1G
+org.gradle.parallel=true
```
donat, line 2: Is the sequential variant interesting in any way?

testing/performance/src/performanceTest/groovy/org/gradle/performance/regression/corefeature/DeprecationCreationPerformanceTest.groovy
```diff
@@ -37,4 +37,16 @@ class DeprecationCreationPerformanceTest extends AbstractCrossVersionPerformance
         then:
         result.assertCurrentVersionHasNotRegressed()
     }
+
+    def "create many deprecation warnings with all warnings shown"() {
+        given:
+        runner.tasksToRun = ['help']
+        runner.minimumBaseVersion = '6.3'
+        runner.args = ['--warning-mode=all']
```
donat, line 45: ❓ is the `--warning-mode=summary` an interesting case?

platforms/ide/problems-api/src/main/java/org/gradle/internal/problems/DefaultProblemDiagnosticsFactory.java
```diff
@@ -156,7 +169,24 @@ public ProblemDiagnostics forCurrentCaller(Supplier<? extends Throwable> excepti
...
-            return locationFromStackTrace(getImplicitThrowable(EXCEPTION_FACTORY), false, false, transformer);
+            return locationFromStackTrace(getImplicitCallerThrowable(), false, false, transformer);
+        }
+
+        /**
+         * Within the limit, captures the full caller stack as before. Past it, captures a cheap bounded
+         * prefix for location inference, up to a further budget, then stops: the bounded walk is cheap per
+         * problem but not free, and a build firing tens of thousands of them (an IDE sync) would otherwise
+         * pay it on every one, while ordinary builds stay well within the budget and keep all locations.
+         */
```
donat: 💅 For complex code I don't mind a generated summary, but here we have more text than actual (easy-to--read) code. I think this is a case of overdocumentation.
```suggestion
```

platforms/ide/problems-api/src/main/java/org/gradle/internal/problems/DefaultProblemDiagnosticsFactory.java
```diff
@@ -156,7 +169,24 @@ public ProblemDiagnostics forCurrentCaller(Supplier<? extends Throwable> excepti
...
+        @Nullable
+        private Throwable getImplicitCallerThrowable() {
+            if (remainingStackTraces.getAndDecrement() > 0) {
```
donat, line 177: ❓ Would it make sense to refactor this logic to a generic `StacktraceCapturer` service? It's the capturer's responsibility to decide what strategy and limit to use.

platforms/ide/problems-impl/build.gradle.kts
```diff
@@ -0,0 +1,45 @@
...
+    implementation(projects.logging)
+    implementation(projects.problemsApi)
+}
+
+gradleModule {
+    requiredRuntimes {
+        daemon = true
+    }
+    computedRuntimes {
+        // Auto-generated by `:checkTargetRuntimes --fix`
```
donat, line 42:
```suggestion
```

donat (approved):

platforms/documentation/docs/src/docs/release/notes.md
donat: Outdated comment!
(The github UI didn't let me delete without restarting the review)

platforms/ide/problems-impl/build.gradle.kts
eskatos: This is actually added there by our own build logic and kept in every project script.

testing/performance/src/performanceTest/groovy/org/gradle/performance/regression/corefeature/DeprecationCreationPerformanceTest.groovy
eskatos: It is covered by the scenario just above that existed already.

testing/performance/src/templates/generateDiverseDeprecationWarnings/settings.gradle
eskatos: It is intentional and matches the other sibling templates.

testing/performance/src/templates/generateDiverseDeprecationWarnings/gradle.properties
eskatos: Not really, it would add very little. Plus, running with parallel on exercises the concurrency aware implementation

platforms/ide/problems-api/src/main/java/org/gradle/internal/problems/DefaultProblemDiagnosticsFactory.java
eskatos: Agreed, good idea. I'd like to do it together with making `--warning-mode=all` use the bounded capture instead of full stacks (so `all` scales on very large builds): at that point a `StacktraceCapturer` that owns the strategy and the limit is the natural place to choose full vs bounded vs none.

One thing to design through when we do it: the limits are per-stream state (`remainingStackTraces`/`remainingBoundedCaptures` reset on `newStream()`), so the capturer can't simply be a singleton service, it'll need per-stream state or be held by the stream.

.teamcity/subprojects.json
eskatos: We can't really put this into `:problems` because we need to use `StackWalker` which is Java 9+. `:problems` is a client module.

Another difference here is that `:problems-impl` is a runtime dependency only in the distro. No other module depends on it. `:problems` conflates too many things and I took this work as an opportunity to add a module where we can evolve things and reduce the tangling.

donat: I understand the reasoning, but we should keep the list of problems-api subproject list clean.  Could you please create an issue about that?

platforms/ide/problems-api/src/main/java/org/gradle/internal/problems/DefaultProblemDiagnosticsFactory.java
eskatos: Done in this PR, PTAL
* https://github.com/gradle/gradle/pull/38251

PR #38133, opened 2026-06-08
Remove implicit property/method lookup in parent projects

### LINKED ISSUE #37922: Deprecate implicit property and method lookup in project hierarchy

reinsch82:
## Summary

Removes the implicit parent-hierarchy lookup that was deprecated in Gradle 9.6. A child project no longer resolves properties or methods from its parent projects — an unresolved reference now fails at the lookup site (`MissingPropertyException` / `MissingMethodException`).

This makes permanent the behavior that was previously gated behind the `NO_IMPLICIT_LOOKUP_IN_PARENT_PROJECTS` feature preview, and matches what Isolated Projects already did. The deprecation warning and the `org.gradle.internal.fail-on-parent-property-lookup` enforcement flag are gone.

## Production changes

- **`ExtensibleDynamicObject`** no longer walks the parent hierarchy: removed the `parent` field, `getParent`/`setParent`, `getInheritable`/`InheritedDynamicObject`/`snapshotInheritable`, the parent walk in `hasProperty`/`tryGetProperty`/`hasMethod`/`tryInvokeMethod` and `getProperties`, the `failOnParentAccess` flag, and the deprecation nag. It no longer implements `HierarchicalDynamicObject`.
- **`DefaultProject`** no longer wires the parent inherited scope; removed `FAIL_ON_PARENT_PROPERTY_LOOKUP` and `getInheritedScope`.
- Removed the now-dead inheritance plumbing: `CrossProjectModelAccess.parentProjectDynamicInheritedScope` (+ `DefaultCrossProjectModelAccess` / `ProblemReportingCrossProjectModelAccess` impls), `ProjectInternal.getInheritedScope`, the `MutableStateAccessAwareProject` override, and the `NO_IMPLICIT_LOOKUP_IN_PARENT_PROJECTS` feature preview.

## Test & doc changes

- Inheritance tests flipped to assert that parent-declared members are **not** visible (`ExtensibleDynamicObjectTest`, `DefaultProjectTest`, `ParentProjectPropertyLookupIntegrationTest`, `ExtraPropertiesIntegrationTest`, `ConfigurationOnDemandIntegrationTest`, `BuildScriptVisibilityIntegrationTest`, `SoftwareModelTaskAndBuildScriptIntegrationTest`, `DynamicObjectIntegrationTest`).
- Deleted `ParentPropertyLookupFailOnAccessIntegrationTest` (the fail-on-parent flag is gone).
- Removed the two `deprecation:implicit-lookup-of-*-in-parent-projects` entries from `KnownProblemIds`.
- Removed the feature-preview opt-in section from the v9 upgrade guide and updated the Isolated Projects behavior note (the lookup is now removed in all modes).

Net: ~1200 lines removed.

## Verification

Compiles clean (production + all touched test source sets). Locally green so far:
- `ExtensibleDynamicObjectTest` (unit)
- `ParentProjectPropertyLookupIntegrationTest` (11), `ExtraPropertiesIntegrationTest` (2)

The remaining touched integ suites (`DynamicObjectIntegrationTest`, `ConfigurationOnDemandIntegrationTest`, `BuildScriptVisibilityIntegrationTest`, `ScriptPluginClassLoadingIntegrationTest`) and the cross-module/composite dependency-resolution tests are left to CI; the behavioral assertions (exact `MissingProperty`/`MissingMethod` messages) may need minor follow-up tweaks based on CI.


reinsch82:
@bot-gradle test this

lkasso (approved):
Docs LGTM. I added another change in this PR to build_environment.adoc

reinsch82:
@bot-gradle test this

PR #38139, opened 2026-06-08
Add AGP version-update PRs to the TIDE board

hegyibalint:
Fixes https://github.com/gradle/gradle-private/issues/5254

The weekly **Update tested AGP versions** automation (`update-agp-versions.yml`) opens a PR with labels and a `gradle/bt-tide` reviewer, but never placed it on the TIDE team board — so these PRs were easy to miss during triage.

### DIFF
--- a/.github/workflows/update-agp-versions.yml
+++ b/.github/workflows/update-agp-versions.yml
@@ -65,10 +65,12 @@ jobs:
           git push origin --delete tide/update-agp 2>/dev/null || true
           git push origin tide/update-agp
 
-          gh pr create \
+          pr_url=$(gh pr create \
             --base master \
             --title "Update tested AGP versions" \
             --body "Update tested AGP versions" \
             --reviewer "gradle/bt-tide" \
-            --label "a:chore,in:performance-test,in:smoke-test,re:android"
-          gh pr comment tide/update-agp --body '@bot-gradle test and merge'
+            --label "a:chore,in:performance-test,in:smoke-test,re:android")
+          # Add the PR to the TIDE team's board (project 147)
+          gh project item-add 147 --owner gradle --url "$pr_url"
+          gh pr comment "$pr_url" --body '@bot-gradle test and merge'


.github/workflows/update-agp-versions.yml
```diff
@@ -65,10 +65,12 @@ jobs:
...
-          gh pr create \
+          pr_url=$(gh pr create \
             --base master \
             --title "Update tested AGP versions" \
             --body "Update tested AGP versions" \
             --reviewer "gradle/bt-tide" \
-            --label "a:chore,in:performance-test,in:smoke-test,re:android"
-          gh pr comment tide/update-agp --body '@bot-gradle test and merge'
+            --label "a:chore,in:performance-test,in:smoke-test,re:android")
+          # Add the PR to the TIDE team's board (project 147)
```
ov7a, line 74: wouldn't it be easier to update TIDE board settings to add a workflow matching these issues? E.g. by label(s) or title prefix?

hegyibalint: Good point!
Hmm, normally we don't necessarily label and automate PRs. IIUC the devprod folks do this by having a label `@dev-productivity` that creates a strong signal.

WDYT, what would fit this case better? Adding `@tide` or trying to match on a label?
My vote would be on a `@tide`.

eskatos: Labels won't match for these chores.

A workflow by title prefix would work given these PRs always have the same title. We wouldn't need to add anything then.

ov7a: I'm against `@tide`, and originally was thinking about a combination of labels (e.g., PR with re:android and in:smoke-test). 

+1 to Paul

hegyibalint: I see, indeed we could do this:
image

hegyibalint:
Closing, as there is a workflow-based way to do this.

PR #37954, opened 2026-05-20
Fix isolated projects violation when using -x task exclusion

joshfriend:
Use `ProjectState` API instead of `Project.getAllprojects()` in `TaskNameResolver.tryFindUnqualifiedTaskCheaply()` to avoid cross-project access violations with isolated projects enabled.

Closes #29154

### Contributor Checklist
- [x] [Review Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md).
- [x] Make sure that all commits are [signed off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff) to indicate that you agree to the terms of [Developer Certificate of Origin](https://developercertificate.org/).
- [x] Make sure all contributed code can be distributed under the terms of the [Apache License 2.0](https://github.com/gradle/gradle/blob/master/LICENSE), e.g. the code was written by yourself or the original code is licensed under [a license compatible to Apache License 2.0](https://apache.org/legal/resolved.html).
- [x] Check ["Allow edit from maintainers" option](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) in pull request so that additional changes can be pushed by Gradle team.
- [x] Provide integration tests (under `<subproject>/src/integTest`) to verify changes from a user perspective.
- [x] Provide unit tests (under `<subproject>/src/test`) to verify logic.
- [x] Update User Guide, DSL Reference, and Javadoc for public-facing changes.
- [x] Ensure that tests pass sanity check: `./gradlew sanityCheck`.
- [x] Ensure that tests pass locally: `./gradlew <changed-subproject>:quickTest`.

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #38142, opened 2026-06-08
Unify test gating for Configuration Cache and Isolated Projects

alllex:
## Summary

Unifies the four Gradle-mode test-gating annotations (`@ToBeFixedForConfigurationCache`, `@UnsupportedWithConfigurationCache`, `@ToBeFixedForIsolatedProjects`, and the new `@UnsupportedWithIsolatedProjects`) onto one framework-agnostic infrastructure so Spock and JUnit tests gate identically and stay in lockstep going forward. As a consequence, `@UnsupportedWithIsolatedProjects` now exists too, giving IP-incompatible features the same treatment as `@UnsupportedWithConfigurationCache`.

### Architecture

- **`GradleModeTestingPolicy`** is the single source of truth for the verdict (`RUN` / `SKIP` / `EXPECT_FAILURE`). `GradleModeTestingExtensions` (Spock) and `GradleModeTestingRules` (JUnit) are thin adapters over it.
- **`@GradleModeTestingIntent`** is a meta-annotation declaring `mode` + `kind` (`TO_BE_FIXED` vs `WONT_SUPPORT`). Adding a new gated annotation only requires meta-annotating it — no new extension/rule plumbing.
- **`GradleModeTestingIntentValidator`** runs at Spock discovery and rejects conflicting placements that previously silently double-wrapped or produced order-dependent gating: class+method, two via hierarchy, or two on one method. Per-class verdicts cached via `ClassValue`; per-method checks intentionally uncached so an in-progress method does not poison its siblings.

### User-visible additions

- All four annotations now take `issue()` linking to the tracking issue for the incompatibility itself — distinct from `@spock.lang.Issue`, which links the test to its own tracking issue.
- `ToBeFixedForIsolatedProjects` rewritten in Java so Kotlin tests can use it.

### Migration notes

- `AbstractIntegrationTest` now wires four `GradleModeTestingRules.*` instances. The eight legacy `*Rule.groovy` / `*Extension.groovy` files are deleted.
- The new validator caught real cases of redundant annotations through inheritance — see commit `Remove redundant ToBeFixed annotations`. Downstream tests with the same shape will fail at discovery until cleaned up.
- One smoke test (`ImportJUnitXmlReports`) converted from `@ToBeFixedForIsolatedProjects` to `@UnsupportedWithIsolatedProjects` — its IP violation is intentional behavior of a legacy API, not a bug to fix.

## Follow-ups (not in this PR)

- Define how Gradle-mode gating composes with `@NotYetImplemented` / `@ToBeImplemented`.


alllex:

alllex:

subprojects/core/src/integTest/groovy/org/gradle/api/ConfigurationOnDemandIntegrationTest.groovy
```diff
@@ -507,8 +507,10 @@ allprojects {
         fixture.assertProjectsConfigured(":", ":a", ":b")
     }
 
-    @Issue("https://github.com/gradle/gradle/issues/29154")
-    @ToBeFixedForIsolatedProjects(because = "-x is not IP compatible")
+    @ToBeFixedForIsolatedProjects(
+        because = "-x is not IP compatible",
```
6hundreds, line 511: 🤔 We have [contributed PR](https://github.com/gradle/gradle/pull/37954) that makes -x IP compatible, AFAIR. Why is it still to be fixed?

alllex: I noticed this as well. And indeed we fixed `-x`. However, we didn't fix (and might never) the integration between IP and Configure-on-demand, which is what this test class is about.

I wanted to avoid moving around annotations in this PR, and only have the absolute minimum to demostrate the new annotations work.

For this specific test class, I think the right call is to have a single annotation at the top of the class that says "IP does not support CoD behavior today" or something similar.

Do you feel strongly that we should fix it in this PR?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingExtension.groovy
```diff
@@ -0,0 +1,115 @@
...
+        // that carry their own intent annotation.
+        FeatureInfo anyFeature = spec.bottomSpec.allFeatures.find()
+        if (anyFeature != null) {
+            GradleModeTestingIntentValidator.validateFeature(
+                spec.bottomSpec.reflection,
+                anyFeature.featureMethod.reflection
+            )
+        }
+
+        apply(annotation, spec, spec.bottomSpec.name)
```
6hundreds: 💭 💅 May be considered for expressiveness:
```
anyFeature 
    ? applyAtFeature(annotation, anyFeature)
    : apply(annotation, spec, spec.bottomSpec.name)
```

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingExtensions.groovy
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Spock extensions that gate Spock specs and features by the active {@link GradleModeTesting}.
+ */
+final class GradleModeTestingExtensions {
+
+    private GradleModeTestingExtensions() {}
+
+    static class ToBeFixedForCC implements IAnnotationDrivenExtension<ToBeFixedForConfigurationCache> {
+
+        private final GradleModeTestingExtension<ToBeFixedForConfigurationCache> adapter = new GradleModeTestingExtension<>(
```
6hundreds: 💅 `adapter` -> `extension` ?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingExtensions.groovy
```diff
@@ -0,0 +1,94 @@
...
+            new GradleModeTestingPolicy.ToBeFixedForCC(),
+            new ToBeFixedSpecInterceptor(GradleModeTesting.CONFIGURATION_CACHE.displayName())
+        )
+
+        @Override
+        void visitFeatureAnnotation(ToBeFixedForConfigurationCache annotation, FeatureInfo feature) {
```
6hundreds: ❓ Why this guy doesn't have `visitSpecAnnotation`?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingExtension.groovy
alllex: The first part of the method only did the validation of the class to highlight class-level annotation conflicts early. But looking at the code I realized it's better to have an explicit helper that does the class-level validation rather than simulating this with a check of a random feature of the spec.

I updated the code and now it simply calls `GradleModeTestingIntentValidator.validateSpec`

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingExtensions.groovy
alllex: When I wanted to rename it, I realized the implementation would be more concise if we used inheritance. Now there is no need for this field.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingExtensions.groovy
alllex: I didn't realize this myself before you asked, but appearently it has been this way all this time. Maybe there was not need for it in the past. But given we are doing an overhaul anyway it's cleaner to keep things symmetric. I updated it, and now it can work on a class too. Good catch!

alllex:

alllex:
@6hundreds, thanks for the feedback! I addressed some of the things directly, and others sparked ideas on how to improve the code. PTAL!

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/ToBeFixedForIsolatedProjects.java
```diff
@@ -0,0 +1,98 @@
...
+ * <p>
+ * Use this annotation when the intention is to fix either the test itself or the underlying feature,
+ * making it compatible with Isolated Projects. If the intention is to not support the tested feature
+ * with Isolated Projects, use {@link UnsupportedWithIsolatedProjects} instead.
+ * <p>
+ * The expectation of failure essentially flips the test result.
+ * A specific failure is not verified, and we only confirm that the test is not passing.
+ * <p>
+ * Instead of expecting failure, you can skip the test in case the test doesn't fail consistently
+ * or has other undersirable effects, such as timeouts.
```
octylFractal:
```suggestion
 * or has other undesirable effects, such as timeouts.
```

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingPolicy.java
```diff
@@ -0,0 +1,289 @@
...
+     * Decide what should happen given the {@code annotation}, the {@code bottomSpecName}
+     * (simple Spock spec name, or fully-qualified JUnit class name), and the
+     * {@code iterationName} (may be {@code null} at spec-level visits, before any
+     * iteration is known).
+     */
+    public final Verdict decide(A annotation, String bottomSpecName, @Nullable String iterationName) {
+        if (!mode.isActive()) {
+            return Verdict.RUN;
+        }
+        if (!matchesAll(bottomSpecs(annotation), iterationMatchers(annotation), bottomSpecName, iterationName)) {
```
octylFractal: ❓ This occurs before the `skip` check, so if you have `FLAKY` and `iterationMatchers`, it will still all run iterations. Should we handle skipping in `InnerIterationInterceptor`?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingIntentValidator.java
```diff
@@ -0,0 +1,210 @@
...
+ * so the hierarchy walk runs once per spec class. Per-method checks run on every
+ * feature visit and are intentionally not cached: a temporary inconsistency on one
+ * method should not poison other, correctly annotated methods of the same spec.
+ *
+ * <p>Note: an abstract base spec that is never instantiated by a concrete subclass
+ * is never validated. Concrete subclasses trigger validation on first feature visit,
+ * and inherited conflicts will surface there.
+ */
+@NullMarked
+final class GradleModeTestingIntentValidator {
```
6hundreds, line 54: ❌I don't think this complexity is justified.

The description says that the validation aims to prevent annotation conflicts or redundancy, which is harmless behaviorally (both extensions fire, same verdict). I assume same verdict because of `GradleModeTestingPolicy` now is a single source of truth for it.

I see potentially broken case for `ToBeFixed` + `Unsupported` on the same target, but this can be solved targetly, without introducing validation framework. But again, how harmful is it? The natural failure mode when two interceptors registered is "test behaves weirdly during local development" — which gets caught by the author within minutes I believe.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingPolicy.java
```diff
@@ -0,0 +1,289 @@
...
+import java.lang.annotation.Annotation;
+
+/**
+ * Decides whether a test should run, be skipped, or expect a failure under the active
+ * {@link GradleModeTesting}, so that Spock-based and JUnit-based tests gate identically.
+ *
+ * @param <A> the annotation type this policy decides for
+ */
+@NullMarked
+public abstract class GradleModeTestingPolicy<A extends Annotation> {
```
6hundreds: ❌Hierarchy is too heavy for what this class expresses. Four concrete subclasses, each ~40 lines, almost all of which is pass-through to the annotation's getters.

I'd consider collapsing to one class parameterized by a few method references / lambdas — `GradleModeTestingPolicy.of(mode, ann::bottomSpecs, ann::iterationMatchers, ...)` to reduce the ceremony

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingIntent.java
```diff
@@ -0,0 +1,49 @@
...
+ * Declarative metadata for the Gradle-mode test-gating annotations.
+ */
+@NullMarked
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.ANNOTATION_TYPE)
+public @interface GradleModeTestingIntent {
+
+    GradleModeTesting mode();
+
+    Kind kind();
```
6hundreds, line 36: ❓How kind is gonna be used? I didn't find its usages in the PR

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingRules.groovy
```diff
@@ -0,0 +1,95 @@
...
+                case GradleModeTestingPolicy.Verdict.EXPECT_FAILURE:
+                    return new ExpectingFailureRuleStatement(base, policy.gradleMode())
+                case GradleModeTestingPolicy.Verdict.RUN:
+                    return base
+            }
+            throw new IllegalStateException("Unhandled verdict: $verdict")
+        }
+    }
+
+    static class ToBeFixedForCC extends BaseRule<ToBeFixedForConfigurationCache> {
```
6hundreds: ❓Same question about hierarchy - do you see it's justified here? Can't we use, say, static factories for instantiation?

6hundreds (changes requested):
@alllex Sorry for this two-stages review, I wasn't able to finish it yesterday.

I see the intention to unify gating is quite beneficial. Having single source for the verdicts is a great thing. However, I think we should assess how useful the validation aspect of the implementation. My feeling is that we can either completely get rid of it or to implement targeted check(s) for harmful scenario(s).

Also, using hierarchies is not quite justified in some places and just bloating LOC without clear benefits.

subprojects/core/src/integTest/groovy/org/gradle/api/ConfigurationOnDemandIntegrationTest.groovy
6hundreds: > For this specific test class, I think the right call is to have a single annotation at the top of the class that says "IP does not support CoD behavior today" or something similar.

This sounds good to me. I'd fix it as we're touching the file

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingIntentValidator.java
alllex: What do you see as complexity?

Gradle's testing infra is already complicated, and we are adding (well, it's already there) a non-trivial layer on top of it -- reverting (!) the expectations of tests, when running with one mode or the other, or skipping tests entirely. 

All the validator does is perform sanity checks that ensure only a single annotation affects the behavior of a given test -- making it simpler to reason about for Gradle developers. I'd argue, if we talk about the user-facing complexity (Gradle dev is a user here), then this change actually reduces it.

It does require some code: ~200 LOC, half of which are javadocs and composition of human-friendly error messages.

You have a point that the potential impact of having multiple applicable annotations might be minimal. However, I can think of at least one nasty case: if multiple annotations are allowed and exist, then by removing an annotation closest to the test, a Gradle dev (or an agent) might think that now the test will be running in a given Gradle mode, but it won't be the case, because there is another annotation higher up that still prevents it from running.

You can argue that humans (and agents) should always verify that the test actually runs, and I'd definitely agree with you. But then we are in the territory of the human (and agent) error. And since it's an error that can be avoided via automated checks, I see no reason to avoid preventing it.

So far, I don't find myself agreeing that the validator adds unnecessary complexity. At least in the sense that I described.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingPolicy.java
alllex: I refactored the code to use fewer hierarchies. I personally do not think it became more readable. With inheritance, it was longer, but it was boring/trivial, which is a good thing. Now it has become "fancy".

What do you think about the updated implementation?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingIntent.java
alllex: It's redundant indeed. I initially introduced it, because I thought it might be used to understand the precendence of the annotations (method vs class vs parent class), but then I moved on to the design, where we explicitly verify only one annotation can be at play. I'll remove this.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingRules.groovy
alllex: Updated to not use inheritance. Not sure it is giving us much, though. Previously, there was a clear symmetry with the Spock extensions, where the classes are required. Now the code is more different.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingPolicy.java
alllex: This is a good catch! It makes me think the current implementation with the `policy.decide` and `policy.requiresPerIterationCheck` is not very robust.

I think a cleaner design might be to rework the `ToBeFixedSpecInterceptor` to become more flexible and handle failure expectations as well as skipping, as you suggest. I'll give this a try -- not yet addressed.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingIntentValidator.java
6hundreds: > What do you see as complexity?

Using all the JDK arsenal: `record`, `sealed interface`, `EnumMap`, `ClassValue`, a caching layer and a test for it to enforce a contract, that wasn't being violated before OR have a small chances to be violated OR which violation is harmless. For the better understanding - does this PR make violation of that contract possible? Or it was possible before?

> It does require some code: ~200 LOC

And ~200 LOC for its test. To clarify - I don't think fewer LOC is always better.

Talking about actions, what I'd consider is to inline validation to extension with dropping IntentSite / ClassSite / MethodSite hierarchy, use just simple formatter and a stacktrace. Not sure how necessary a caching layer there also. But I see your point `if we talk about the user-facing complexity (Gradle dev is a user here), then this change actually reduces it.` and somewhat also agree with this.

If these arguments aren't strong enough, and you decide to keep this validator as a separate class of the presented complexity - I can live with it for sure

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/GradleModeTestingPolicy.java
6hundreds: > What do you think about the updated implementation?

Not dramatically better, agree, but a bit more concise

alllex:
Superseded by
- https://github.com/gradle/gradle/pull/38248

PR #38152, opened 2026-06-09
Add violations when mutating StartParameter after settings evaluation

### LINKED ISSUE #37944: Handling `StartParameter` state with Isolated Projects
`StartParameter` is a problematic API, it's not an interface, it must be potentially immutable while it's not. Here is umbrella issue on this matter https://github.com/gradle/gradle/issues/37410.

Whereas `StartParameterInternal` exposes a lot of setters, which must be IP reported, some getters there are not safe as well due to returning mutable beans or collections(see `projectProperties`, `systemPropertiesArgs`,  `loggingConfiguration` and etc.)

Another thing to be considered is how safe cross-build access to `StartParameter`, and what is reporting mode there.

jvandort:
IP violations are not emitted when mutating StartParameter after settings for a build has completed evaluation. This is the same instant where we prevent some state on settings from being mutated.

Perhaps we should do this earlier, after init scripts are evaluated, but this is a reasonable instant for now.

We achieve this by having StartParameterInternal store a listener to be called whenever its state is mutated. We set this listener after settings are evaluated to a callback which emits violations.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
@@ -121,4 +121,98 @@ class IsolatedProjectsStartParameterIntegrationTest extends AbstractIsolatedProj
         }
     }
 
+    def "mutating StartParameter from init script is allowed"() {
+        file("init.gradle") << """
+            gradle.startParameter.setContinueOnFailure(true)
+        """
+        buildFile("""
+            tasks.register("ok")
+        """)
+
+        when:
+        isolatedProjectsRun("ok", "-I", "init.gradle")
+
+        then:
+        postBuildOutputContains("Configuration cache entry stored.")
+    }
+
+    def "reading StartParameter from a build script after settings evaluation is not a violation"() {
+        // Only mutation is reported; reads stay silent, including reads that go through the
+        // mutation-notifying collection-view wrappers (Set/Map/List and their keySet/values/entrySet
+        // views, iterators and entries). These reads run at configuration time, which must complete for
+        // the build to reach projectsConfigured(":"), so asserting their values in the script is
+        // reliable: a wrong value fails the build, and the trailing marker proves every assert ran.
+        // The decisive check is still that none of these reads is reported: assertStateStored succeeds.
+        buildFile("""
+            def sp = gradle.startParameter
+            assert sp.offline == false
+            assert sp.excludedTaskNames.contains('x') == false
+            assert sp.excludedTaskNames.collect { it } == []
+            assert sp.projectProperties['seed'] == 'value'
+            assert sp.projectProperties.keySet().contains('seed')
+            assert sp.projectProperties.values().contains('value')
+            assert sp.projectProperties.entrySet().find { it.key == 'seed' }?.value == 'value'
+            assert sp.taskRequests.any { it.args.contains('ok') }
+            println("all reads verified")
+            tasks.register("ok")
+        """)
+
+        when:
+        isolatedProjectsRun("ok", "-Pseed=value")
+
+        then:
+        outputContains("all reads verified")
+        fixture.assertStateStored {
+            projectsConfigured(":")
+        }
+    }
+
+    def "replacing the requested tasks from a build script is currently allowed"() {
+        // setTaskNames and setTaskRequests are exempt from violation reporting for now: tooling model
+        // builders, e.g. during IDE sync, still legitimately replace the tasks to run while the build
+        // is already running. This pins the exemption until a better pattern exists for them.
+        buildFile("""
+            tasks.register('a')
+            tasks.register('b') { doLast { println 'b executed' } }
+            gradle.startParameter.setTaskNames(['b'])
+        """)
+
+        when:
+        isolatedProjectsRun("a")
+
+        then:
+        outputContains("b executed")
+        fixture.assertStateStored {
+            projectsConfigured(":")
+        }
+    }
+
+    def "mutating a StartParameter collection view is reported with the access path (#invocation)"() {
+        buildFile("""
+            gradle.startParameter.$invocation
+        """)
+
+        when:
+        // Fail-fast only: the violation is thrown before the delegate collection is touched, making
+        // the outcome independent of the delegate's mutability. In diagnostics mode the mutation
+        // proceeds into the delegate, and some backing collections (e.g. the project properties
+        // populated by the CLI converter) are immutable in real distributions, so the build would
+        // additionally fail with an UnsupportedOperationException there.
+        isolatedProjectsFailsUsing(IsolatedProjectsMode.FAIL_FAST, "help", "-Pseed=value")
+
+        then:
+        fixture.assertIsolatedProjectsProblems(IsolatedProjectsMode.FAIL_FAST) {
+            projectsConfigured(":")
+            problem("Build file 'build.gradle': line 2: Cannot call '${signature}' on StartParameter after settings have been evaluated when Isolated Projects is enabled.")
+        }
+
+        where:
+        invocation                                                                        | signature
+        "projectProperties.put('p', 'v')"                                                 | "getProjectProperties().put(Object, Object)"
+        "projectProperties.keySet().remove('seed')"                                       | "getProjectProperties().keySet().remove(Object)"
+        "projectProperties.keySet().with { def i = it.iterator(); i.next(); i.remove() }" | "getProjectProperties().keySet().iterator().remove()"
+        "projectProperties.entrySet().find { it.key == 'seed' }.setValue('x')"            | "getProjectProperties().entrySet().Entry.setValue(Object)"
+        "excludedTaskNames.removeIf { it == 'x' }"                                        | "getExcludedTaskNames().removeIf(Predicate)"
+    }
+
 }
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl.isolated
+
+class IsolatedProjectsStartParameterIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
+
+    def "mutating StartParameter after settings evaluation is a violation (#location)"() {
+        // The violation pipeline (onMutableCall -> listener -> IP problem) is the same for every setter
+        // and only the call location varies, so one representative setter exercised from each location
+        // covers it; that every individual setter is instrumented is checked by the fast reflective unit
+        // test StartParameterMutationInstrumentationTest. The root case has no included build on purpose:
+        // adding one would make its configured-project set depend on the mode (in fail-fast the root
+        // throws before the included build is configured, in diagnostics everything is configured).
+        if (scriptPath.startsWith("included/")) {
+            settingsFile("""
+                includeBuild("included")
+            """)
+        }
+        file(scriptPath) << """
+            ${apiExpr}.setDryRun(true)
+        """
+
+        when:
+        isolatedProjectsFailsUsing(mode, "help")
+
+        then:
+        fixture.assertIsolatedProjectsProblems(mode) {
+            projectsConfigured(*configured)
+            problem("Build file '${scriptPath}': line 2: Cannot call 'setDryRun(boolean)' on StartParameter after settings have been evaluated when Isolated Projects is enabled.")
+        }
+
+        where:
+        location           | scriptPath              | apiExpr                        | configured
+        "root project"     | "build.gradle"          | "gradle.startParameter"        | [":"]
+        "included project" | "included/build.gradle" | "gradle.startParameter"        | [":", ":included"]
+        "parent build"     | "included/build.gradle" | "gradle.parent.startParameter" | [":", ":included"]
+
+        combined:
+        mode << ALL_MODES
+    }
+
+    def "mutating StartParameter from root build settings script is allowed"() {
+        settingsFile("""
+            gradle.startParameter.setDryRun(true)
+        """)
+
+        when:
+        isolatedProjectsRun("help")
+
+        then:
+        fixture.assertStateStored {
+            projectsConfigured(":")
+        }
+    }
+
+    def "mutating StartParameter from included build settings script is allowed"() {
+        settingsFile("""
+            includeBuild("included")
+        """)
+        file("included/settings.gradle") << """
+            gradle.startParameter.setDryRun(true)
+        """
+
+        when:
+        isolatedProjectsRun("help")
+
+        then:
+        fixture.assertStateStored {
+            projectsConfigured(":", ":included")
+        }
+    }
+
+    def "configuring the StartParameter of a GradleBuild task is not a violation"() {
+        // The task's StartParameter is a copy created to define the nested build; mutating it
+        // configures a build that has not started yet, which is the window where mutation is legal.
+        // The mutation listener only guards a running build's own StartParameter.
+        settingsFile("""
+            rootProject.name = 'outer'
+        """)
+        buildFile("""
+            tasks.register('nested', GradleBuild) {
+                dir = file('other')
+                tasks = ['hello']
+                startParameter.setOffline(true)
+                startParameter.projectProperties.put('greeting', 'hi from outer')
+            }
+        """)
+        file("other/settings.gradle") << """
+            rootProject.name = 'inner'
+        """
+        file("other/build.gradle") << """
+            def greeting = providers.gradleProperty('greeting')
+            def offline = gradle.startParameter.offline
+            tasks.register('hello') {
+                doLast { println("inner says: \${greeting.get()}, offline=\$offline") }
+            }
+        """
+
+        when:
+        isolatedProjectsRun("nested")
+
+        then:
+        outputContains("inner says: hi from outer, offline=true")
+        fixture.assertStateStored {
+            // ":" is the outer build; ":other" is the nested build run by the task
+            projectsConfigured(":", ":other")
+        }
+    }
+
+}
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
@@ -83,7 +83,6 @@ import org.gradle.internal.serialize.graph.readCollection
 import org.gradle.internal.serialize.graph.readEnum
 import org.gradle.internal.serialize.graph.readList
 import org.gradle.internal.serialize.graph.readNonNull
-import org.gradle.internal.serialize.graph.readStrings
 import org.gradle.internal.serialize.graph.readStringsSet
 import org.gradle.internal.serialize.graph.runWriteOperation
 import org.gradle.internal.serialize.graph.withDebugFrame
@@ -693,7 +692,6 @@ class ConfigurationCacheState(
     fun WriteContext.writeGradleState(gradle: GradleInternal) {
         withGradleIsolate(gradle, userTypesCodec) {
             // per build
-            writeStartParameterOf(gradle)
             writeChildBuilds(gradle)
         }
     }
@@ -705,23 +703,8 @@ class ConfigurationCacheState(
         val gradle = build.gradle
         return withGradleIsolate(gradle, userTypesCodec) {
             // per build
-            readStartParameterOf(gradle)
             readChildBuilds()
         }
-    }
-
-    private
-    fun WriteContext.writeStartParameterOf(gradle: GradleInternal) {
-        val startParameterTaskNames = gradle.startParameter.taskNames
-        writeStrings(startParameterTaskNames)
-    }
-
-    private
-    fun ReadContext.readStartParameterOf(gradle: GradleInternal) {
-        // Restore startParameter.taskNames to enable `gradle.startParameter.setTaskNames(...)` idiom in included build scripts
-        // See org/gradle/caching/configuration/internal/BuildCacheCompositeConfigurationIntegrationTest.groovy:134
-        val startParameterTaskNames = readStrings()
-        gradle.startParameter.setTaskNames(startParameterTaskNames)
     }
 
     private
--- a/platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingCollection.java
+++ b/platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingCollection.java
@@ -1,0 +1,217 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.collect;
+
+import org.jspecify.annotations.Nullable;
+
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+/**
+ * A generic decorator for a {@link Collection} that reports every mutating operation to an
+ * {@link Interceptor}, including mutations reached through {@link #iterator()}. Reads pass straight
+ * through. It owns the tricky mechanics (forwarding, wrapping the iterator, applying an element view,
+ * serializing as the delegate) so callers only supply the notification.
+ *
+ * @param <E> the element type
+ * @param <C> the delegate's concrete collection type, so subtypes can reach its specific methods
+ */
+public class InterceptingCollection<E, C extends Collection<E>> implements Collection<E>, Serializable {
+
+    /**
+     * Notified whenever a mutating method is called, with the source-level {@code methodSignature} of
+     * that method (e.g. {@code "add(Object)"}). A view prepends its path, e.g. {@code "keySet().clear()"}.
+     */
+    public interface Interceptor {
+        void onMutate(String methodSignature);
+    }
+
+    protected final C delegate;
+    final transient Interceptor interceptor;
+    private final transient Function<E, E> elementView;
+    private final transient boolean wrapsElements;
+
+    public InterceptingCollection(C delegate, Interceptor interceptor) {
+        this.delegate = delegate;
+        this.interceptor = interceptor;
+        this.elementView = Function.identity();
+        this.wrapsElements = false;
+    }
+
+    protected InterceptingCollection(C delegate, Interceptor interceptor, Function<E, E> elementView) {
+        this.delegate = delegate;
+        this.interceptor = interceptor;
+        this.elementView = elementView;
+        this.wrapsElements = true;
+    }
+
+    private E view(E element) {
+        return elementView.apply(element);
+    }
+
+    // Serialize as the delegate: the interceptor is runtime-only wiring and instances may be captured
+    // in task state and stored to the configuration cache.
+    protected Object writeReplace() {
+        return delegate;
+    }
+
+    // region reads
+    @Override
+    public int size() {
+        return delegate.size();
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return delegate.isEmpty();
+    }
+
+    @Override
+    public boolean contains(Object o) {
+        return delegate.contains(o);
+    }
+
+    @Override
+    public boolean containsAll(Collection<?> c) {
+        return delegate.containsAll(c);
+    }
+
+    @Override
+    public Iterator<E> iterator() {
+        Iterator<E> it = delegate.iterator();
+        return new Iterator<E>() {
+            @Override
+            public boolean hasNext() {
+                return it.hasNext();
+            }
+
+            @Override
+            public E next() {
+                return view(it.next());
+            }
+
+            @Override
+            public void remove() {
+                interceptor.onMutate("iterator().remove()");
+                it.remove();
+            }
+        };
+    }
+
+    @Override
+    public Object[] toArray() {
+        return viewAll(delegate.toArray());
+    }
+
+    @Override
+    @SuppressWarnings("SuspiciousToArrayCall")
+    public <T> T[] toArray(T[] a) {
+        return viewAll(delegate.toArray(a));
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T> T[] viewAll(T[] array) {
+        if (wrapsElements) {
+            for (int i = 0; i < array.length; i++) {
+                array[i] = (T) view((E) array[i]);
+            }
+        }
+        return array;
+    }
+
+    @Override
+    public Spliterator<E> spliterator() {
+        if (!wrapsElements) {
+            return delegate.spliterator();
+        }
+        // Elements are wrapped (e.g. entries whose setValue is intercepted), so stream/spliterator must
+        // hand out wrapped elements too. Route through the wrapping iterator at the cost of split quality.
+        return Spliterators.spliterator(iterator(), delegate.size(), Spliterator.DISTINCT);
+    }
+
+    @Override
+    public void forEach(Consumer<? super E> action) {
+        delegate.forEach(e -> action.accept(view(e)));
+    }
+
+    @Override
+    @SuppressWarnings("UndefinedEquals") // We're fine with having weak contract of Iterable/Collection.equals.
+    public boolean equals(@Nullable Object o) {
+        return delegate.equals(o);
+    }
+
+    @Override
+    public int hashCode() {
+        return delegate.hashCode();
+    }
+
+    @Override
+    public String toString() {
+        return delegate.toString();
+    }
+    // endregion
+
+    // region mutations
+    @Override
+    public boolean add(E e) {
+        interceptor.onMutate("add(Object)");
+        return delegate.add(e);
+    }
+
+    @Override
+    public boolean addAll(Collection<? extends E> c) {
+        interceptor.onMutate("addAll(Collection)");
+        return delegate.addAll(c);
+    }
+
+    @Override
+    public boolean remove(@Nullable Object o) {
+        interceptor.onMutate("remove(Object)");
+        return delegate.remove(o);
+    }
+
+    @Override
+    public boolean removeAll(Collection<?> c) {
+        interceptor.onMutate("removeAll(Collection)");
+        return delegate.removeAll(c);
+    }
+
+    @Override
+    public boolean retainAll(Collection<?> c) {
+        interceptor.onMutate("retainAll(Collection)");
+        return delegate.retainAll(c);
+    }
+
+    @Override
+    public boolean removeIf(Predicate<? super E> filter) {
+        interceptor.onMutate("removeIf(Predicate)");
+        return delegate.removeIf(filter);
+    }
+
+    @Override
+    public void clear() {
+        interceptor.onMutate("clear()");
+        delegate.clear();
+    }
+    // endregion
+}
--- a/platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/RootBuildLifecycleBuildActionExecutor.java
+++ b/platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/RootBuildLifecycleBuildActionExecutor.java
@@ -16,9 +16,9 @@
 
 package org.gradle.launcher.exec;
 
-import org.gradle.StartParameter;
 import org.gradle.api.JavaVersion;
 import org.gradle.api.internal.BuildDefinition;
+import org.gradle.api.internal.StartParameterInternal;
 import org.gradle.api.logging.configuration.ShowStacktrace;
 import org.gradle.api.problems.internal.ProblemsInternal;
 import org.gradle.internal.build.BuildStateRegistry;
@@ -49,7 +49,7 @@ public class RootBuildLifecycleBuildActionExecutor {
     private final BuildTreeLifecycleListener lifecycleListener;
     private final ProblemsInternal problemsService;
     private final BuildOperationProgressEventEmitter eventEmitter;
-    private final StartParameter startParameter;
+    private final StartParameterInternal startParameter;
     private final ProblemStream problemsStream;
     private final BuildActionRunner buildActionRunner;
     private final BuildStateRegistry buildStateRegistry;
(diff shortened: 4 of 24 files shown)


jvandort:
@bot-gradle help

jvandort:
@bot-gradle test ipt

jvandort:
Looks like `setTaskRequests(Iterable)` and `setTaskNames(Iterable)` are being called in a number of places after settings have been evaluated. We have a couple of solution to this:
1. Find the code that calls these methods late and find out how to call them earlier
2. Synchronize these two methods so they are thread safe, remove the violations, and push this to a later problem
3. Find another API the internal callers of these method can use. We need to determine the use case to see if/why third-party plugins will call these methods "late".

The other outstanding issue `sanityCheck`. This change relies on adding a `protected void onMutableCall(String methodSignature)`, which `StartParameterInternal` overrides to trigger violations. This comes up as a binary breaking change, missing a `@since` and an `@Incubating`. Before adding these annotations I wanted to validate whether we think they are 1. necessary 2. okay. I believe this approach is much less cumbersome than a "wrapping" approach like we are doing with Gradle, Project, as StartParameter has a ton of fields and would be tedious to wrap.

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/MutationNotifyingList.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Override
+    public int lastIndexOf(Object o) {
+        return delegate.lastIndexOf(o);
+    }
+
+    @Override
+    public Iterator<E> iterator() {
+        return delegate.iterator();
```
reinsch82: 🤔 
The collections do not provide a mutation safe Iterator. Users might be tempted to use these as an escape hatch for the other restrictions.

jvandort:
@bot-gradle test ipt

reinsch82:
@bot-gradle test ipt

jvandort:
@bot-gradle test ipt

reinsch82:
@bot-gradle test ipt

alllex:
@bot-gradle test this without pts

alllex:
@jvandort, what is the strategy of handling the `GradleBuild` task that uses the start parameters as the task input, where it's valid to mutate it?

lukebemish:
In regards to the use of `setTaskRequests`/`setTaskNames`; I (and others with similar plugins) use it primarily to register tasks during IDE sync that need to be executed for IDE sync to correctly complete. For instance, we have several plugins that generate dependencies. The dependencies in question require a bunch of work to set up, and are best modeled with tasks whose execution creates the relevant dependency jar (which can then be depended on as normal), but that requires us to properly trigger those tasks to run during IDE sync. The most reliable way I've found to do so (without internal APIs or anything) is something like:
```gradle
// If IJ IDE sync...
var startParameter = getProject().getGradle().getStartParameter();
var taskRequests = new ArrayList<>(startParameter.getTaskRequests());

startParameter.setTaskNames(List.of("whateverTheTaskThatMustRunIs"));
taskRequests.addAll(startParameter.getTaskRequests());
startParameter.setTaskRequests(taskRequests);
```
which is _technically_ not IP safe as if multiple projects configure in parallel this could end up as a race condition. That said, I don't think limiting this further makes sense unless there is a better API being provided for this type of use case; that said, since what is being done here is purely additive, it should be perfectly possible to design an API for _adding_ tasks to be executed that is IP-safe?

For examples from plugins where this API is used this way, see:
* https://github.com/FabricMC/fabric-loom/blob/24b41d2cd29edda066379d9d655ba81ea56ce66d/src/main/java/net/fabricmc/loom/configuration/ide/idea/IdeaConfiguration.java#L44-L62
* https://github.com/neoforged/ModDevGradle/blob/846b2d70f99519640efd6620e7d6f9034fe38285/src/main/java/net/neoforged/moddevgradle/internal/IntelliJIntegration.java#L67-L74
* https://github.com/terrarium-earth/cloche/blob/287d132aab6424712d870862ee9bc92ebbef94ba/src/main/kotlin/earth/terrarium/cloche/ClochePlugin.kt#L36-L41

This is currently *the* only reliable way of triggering dependency generation (or other necessary-at-sync tasks) during IDE sync; other approaches (such as the use of the `idea-ext` plugin to trigger a post-sync extra Gradle invocation) have some major disadvantages and are even less "safe" (far less IP safe *and* you're relying on weird undocumented JSON serialization "APIs").

reinsch82:
@bot-gradle test this without pts

reinsch82:
> what is the strategy of handling the GradleBuild task that uses the start parameters as the task input, where it's valid to mutate it?

@alllex this is fine as `GradleBuild` doesn't interact with the global instance, it uses it's own copy

platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
```diff
@@ -71,7 +77,7 @@ public class StartParameter implements LoggingConfiguration, ParallelismConfigur
 
     private final DefaultLoggingConfiguration loggingConfiguration = new DefaultLoggingConfiguration();
     private final DefaultParallelismConfiguration parallelismConfiguration = new DefaultParallelismConfiguration();
-    private List<TaskExecutionRequest> taskRequests = new ArrayList<>();
+    private volatile List<TaskExecutionRequest> taskRequests = new CopyOnWriteArrayList<>();
```
alllex, line 83: 💭  This would benefit from a field-targeting comment on why its volatile and copy-on-write

subprojects/core/src/main/java/org/gradle/initialization/SettingsEvaluatedCallbackFiringSettingsProcessor.java
```diff
@@ -35,6 +40,18 @@ public SettingsState process(GradleInternal gradle, SettingsLocation settingsLoc
         SettingsInternal settings = state.getSettings();
         gradle.getBuildListenerBroadcaster().settingsEvaluated(settings);
         settings.preventFromFurtherMutation();
+        if (startParameter instanceof StartParameterInternal) {
```
alllex: ❓ Can it be anything else?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
```diff
@@ -705,25 +703,10 @@ class ConfigurationCacheState(
...
-    }
-
-    private
-    fun ReadContext.readStartParameterOf(gradle: GradleInternal) {
-        // Restore startParameter.taskNames to enable `gradle.startParameter.setTaskNames(...)` idiom in included build scripts
-        // See org/gradle/caching/configuration/internal/BuildCacheCompositeConfigurationIntegrationTest.groovy:134
-        val startParameterTaskNames = readStrings()
-        gradle.startParameter.setTaskNames(startParameterTaskNames)
-    }
-
```
mlopatkin, line 726: ❓ Where did this go? It doesn't look like mutating task names is now a hard error with CC?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl.isolated
+
+class IsolatedProjectsStartParameterIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
```
alllex, line 19: ❓ After a cursory look, it seems like we are missing test coverage for positive cases that show what's allowed. It's useful to have these tests to prove those things actually keep working

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+
+        then:
+        fixture.assertIsolatedProjectsProblems(mode) {
+            projectsConfigured(":")
+            problem("Build file 'build.gradle': line 2: Cannot call '${method}' on StartParameter after settings have been evaluated when Isolated Projects is enabled.")
+        }
+
+        where:
+        method                              | call
+        "setDryRun(boolean)"                | 'setDryRun(true)'
```
alllex: ❓ Any reason not to extend this coverage to all forbidden methods to ensure they work as expected and potential future changes in behavior are caught?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+
+        then:
+        fixture.assertIsolatedProjectsProblems(mode) {
+            projectsConfigured(":")
+            problem("Build file 'build.gradle': line 2: Cannot call '${method}' on StartParameter after settings have been evaluated when Isolated Projects is enabled.")
+        }
+
+        where:
+        method                                      | call
+        "setConfigurationCacheDebug(boolean)"       | 'setConfigurationCacheDebug(true)'
```
alllex: ❓ Any reason not to extend this coverage to all forbidden methods to ensure they work as expected and potential future changes in behavior are caught?

It's not as important here, because it's internal API, but I think it is useful for completeness, since we actually have those violations.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+        fixture.assertIsolatedProjectsProblems(mode) {
+            projectsConfigured(":", ":included")
+            problem("Build file 'included/build.gradle': line 2: Cannot call 'setOffline(boolean)' on StartParameter after settings have been evaluated when Isolated Projects is enabled.")
+        }
+
+        where:
+        mode << ALL_MODES
+    }
+
+    def "mutating StartParameter from root build settings script is allowed"() {
```
alllex, line 56: 💭 Should we test the init script case as well?

platforms/core-runtime/collections/src/test/groovy/org/gradle/internal/collect/MutationNotifyingCollectionsTest.groovy
```diff
@@ -0,0 +1,368 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.collect
+
+import spock.lang.Specification
+
+class MutationNotifyingCollectionsTest extends Specification {
```
alllex: 🤔 We have these generic tests for the collections, but we are missing integration tests that show how these are materialized for the user-facing scenarios, in case of the access to the deeply tracked structures.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
reinsch82: The restored task names are not read anywhere.

Since we want a way out of it anyway, we decided to drop it already.

subprojects/core/src/main/java/org/gradle/initialization/SettingsEvaluatedCallbackFiringSettingsProcessor.java
reinsch82: No — it's always a `StartParameterInternal` in practice; the parameter was just typed as the public class. Tightened `SettingsProcessor` (internal API) to take `StartParameterInternal` directly, so the check and cast are gone entirely — and a violation of the assumption now fails to compile rather than silently skipping the listener installation.

platforms/core-runtime/collections/src/test/groovy/org/gradle/internal/collect/MutationNotifyingCollectionsTest.groovy
reinsch82: created integration test

https://github.com/gradle/gradle/blob/35b6c8a829e85745c18958f424f0976ff6e68cec/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy#L360

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
reinsch82: we do have
- https://github.com/gradle/gradle/blob/35b6c8a829e85745c18958f424f0976ff6e68cec/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy#L178
- https://github.com/gradle/gradle/blob/35b6c8a829e85745c18958f424f0976ff6e68cec/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy#L89
- https://github.com/gradle/gradle/blob/35b6c8a829e85745c18958f424f0976ff6e68cec/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy#L103

alllex: Yes, but what about non-mutating cases?

reinsch82:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,395 @@
...
+        method                              | call
+        "setDryRun(boolean)"                | 'setDryRun(true)'
+        "setBuildCacheEnabled(boolean)"     | 'setBuildCacheEnabled(true)'
+        "setOffline(boolean)"               | 'setOffline(true)'
+
+        combined:
+        mode << ALL_MODES
+    }
+
+    def "mutating StartParameterInternal method from root project build script is a violation"() {
```
6hundreds: 🤔  Do we care about `StartParameterInternal` at all?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,395 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl.isolated
+
+class IsolatedProjectsStartParameterIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
```
6hundreds, line 19: 💭 Some tests in this class looks very similar, but it's hard to parametrize it with api call location(build script/included project script/etc) because of build layout. What if to use one build layout, with included build in place, and then have a single test that will be parametrized with api call and location?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,395 @@
...
+            println("taskNames: \${sp.taskNames}")
+            println("excludedTaskNames: \${sp.excludedTaskNames}")
+            println("excluded contains: \${sp.excludedTaskNames.contains('x')}")
+            println("excluded iterated: \${sp.excludedTaskNames.collect { it }}")
+            println("projectProperties: \${sp.projectProperties}")
+            println("projectProperties lookup: \${sp.projectProperties['seed']}")
+            println("projectProperties keySet: \${sp.projectProperties.keySet().toList()}")
+            println("projectProperties values: \${sp.projectProperties.values().toList()}")
+            println("projectProperties entries: \${sp.projectProperties.entrySet().collect { "\${it.key}=\${it.value}" }}")
+            println("taskRequests: \${sp.taskRequests.collect { it.args }}")
```
6hundreds: 🤔 Not a blocker at all, but we print a lot, but verifying only `projectProperties lookup`. Also I'd use test parametrization for this, but maybe it doesn't worth as it spawns a test for each call, what can be an overkill for this scenario

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,395 @@
...
+        fixture.assertStateStored {
+            projectsConfigured(":")
+        }
+    }
+
+    // Every notifying public mutator of StartParameter, with an invocation whose value matches the
+    // current/default state: the build completes configuration in diagnostics mode, so later build
+    // phases read the mutated values. setTaskNames/setTaskRequests are deliberately absent: they are
+    // exempt from violation reporting for now (see above).
+    private static final List<List> PUBLIC_MUTATORS = [
```
6hundreds: ❓ Not sure I get it why it's list of lists?

6hundreds: I mean, isn't it 1-1 `signature`-`invocation`? What are the instances on 1-Many?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,395 @@
...
+        ["setNoBuildScan(false)", ["setNoBuildScan(boolean)"]],
+        ["setWriteDependencyLocks(false)", ["setWriteDependencyLocks(boolean)"]],
+        ["setLockedDependenciesToUpdate([])", ["setLockedDependenciesToUpdate(List)"]],
+        ["setWriteDependencyVerifications([])", ["setWriteDependencyVerifications(List)"]],
+        ["setDependencyVerificationMode(org.gradle.api.artifacts.verification.DependencyVerificationMode.STRICT)", ["setDependencyVerificationMode(DependencyVerificationMode)"]],
+        ["setRefreshKeys(false)", ["setRefreshKeys(boolean)"]],
+        ["setExportKeys(false)", ["setExportKeys(boolean)"]],
+    ]
+
+    def "every mutating method of StartParameter is reported (chunk starting with #first)"() {
```
6hundreds: 🤔 How this test is different from, say, `mutating StartParameter method from root project build script is a violation`. Asserting on all mutation calls reported? I have feeling it exercises IP reporting in general then, rather then something `StartParameter` related. No?

platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
```diff
@@ -106,6 +115,9 @@ public class StartParameter implements LoggingConfiguration, ParallelismConfigur
     private boolean exportKeys;
     private WelcomeMessageConfiguration welcomeMessageConfiguration = new WelcomeMessageConfiguration(WelcomeMessageDisplayMode.ONCE);
 
+    protected void onMutableCall(String methodSignature) {
+    }
```
6hundreds, line 128: ❓ Do nothing?

reinsch82: That is the base for the behavior we introduce in `StartParameterInternal`.

It's intended we do nothing here.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
reinsch82: there used to be a few cases, that got dropped that where 1 to many.
fixed it.

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/MutationNotifying.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Collection;
+import java.util.function.Consumer;
+
+/**
+ * The mutation-notifying policy for the {@code Intercepting*} decorators: reacts to mutating
+ * operations only (reads stay silent) and reports the source-level method signature. Every mutating
+ * callback funnels to the same {@code onMutation}; this uniform treatment is what made the single-hook
+ * design a one-liner — with per-operation hooks it has to override each mutator explicitly.
+ */
+final class MutationNotifying {
```
jvandort: `CollectionInterceptors`

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/MutationNotifying.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private MutationNotifying() {
+    }
+
+    static Interceptor interceptor(Consumer<String> onMutation) {
```
jvandort: `onMutation`

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingCollection.java
```diff
@@ -0,0 +1,259 @@
...
+/**
+ * A generic decorator for a {@link Collection} that reports every interface operation to an
+ * {@link Interceptor}, including operations reached through {@link #iterator()}. It owns the tricky
+ * mechanics (forwarding, wrapping the iterator, applying an element view, serializing as the delegate)
+ * so callers only supply the per-operation policy.
+ *
+ * @param <E> the element type
+ * @param <C> the delegate's concrete collection type, so subtypes can reach its specific methods
+ */
+public class InterceptingCollection<E, C extends Collection<E>> implements Collection<E>, Serializable {
```
jvandort, line 39: - Lets revert CC `input-tracking` changes
- Remove fine-grained access tracking methods from `Interceptor`, leave only `onMutate`
- Delete `MutationNotifying` extension classes

platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
```diff
@@ -252,7 +253,10 @@ public File getProjectCacheDir() {
 
     /**
      * Creates a {@code StartParameter} with default values. This is roughly equivalent to running Gradle on the command-line with no arguments.
+     *
+     * @deprecated A {@code StartParameter} is provided by the build (e.g. {@code gradle.startParameter}) and should not be created directly. Derive a copy from that one with {@link #newInstance} or {@link #newBuild} if you need to.
      */
+    @Deprecated
```
jvandort: Lets address this deprecation in another PR

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
jvandort: > Yes, but what about non-mutating cases?

Wouldn't every other test in Gradle that touches start parameter be the non-mutating tests? We did just merge (or are about to merge) a PR that runs all of our integ test suite with IP.

If we are missing that test coverage elsewhere then I think thats more of a separate issue.

platforms/core-runtime/start-parameter/src/main/java/org/gradle/api/internal/StartParameterInternal.java
```diff
@@ -70,6 +74,25 @@ protected StartParameterInternal(BuildLayoutParameters layoutParameters) {
         super(layoutParameters);
     }
 
+    public void setMutationListener(Consumer<String> mutationListener) {
```
alllex, line 77: 💅 The method would benefit from a javadoc concisely explaining why it exists

subprojects/core/src/main/java/org/gradle/initialization/SettingsEvaluatedCallbackFiringSettingsProcessor.java
```diff
@@ -16,25 +16,39 @@
...
     }
 
     @Override
-    public SettingsState process(GradleInternal gradle, SettingsLocation settingsLocation, ClassLoaderScope buildRootClassLoaderScope, StartParameter startParameter) {
+    public SettingsState process(GradleInternal gradle, SettingsLocation settingsLocation, ClassLoaderScope buildRootClassLoaderScope, StartParameterInternal startParameter) {
         SettingsState state = delegate.process(gradle, settingsLocation, buildRootClassLoaderScope, startParameter);
         SettingsInternal settings = state.getSettings();
         gradle.getBuildListenerBroadcaster().settingsEvaluated(settings);
         settings.preventFromFurtherMutation();
+        startParameter.setMutationListener(methodSignature ->
```
alllex, line 42: ❓ Is there need for having the mutation listeners when IP is not enabled?

platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/RootBuildLifecycleBuildActionExecutor.java
```diff
@@ -98,6 +98,8 @@ public BuildActionRunner.Result execute(BuildAction action) {
                 RootBuildState rootBuild = buildStateRegistry.createRootBuild(BuildDefinition.fromStartParameter(action.getStartParameter(), null));
                 return rootBuild.run(buildController -> buildActionRunner.run(action, buildController));
             } finally {
+                // Since continuous builds reuse the same StartParameter for multiple build trees.
+                startParameter.clearMutationListener();
```
alllex, line 102: ❓ We install mutation listeners on settings in included builds as well. Are they not required to be cleaned?

Do we actually ever install the mutation listener on the build-tree-scoped start parameter? 

Another concern is that this call is very detached from a place where we actually install the listener.

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingCollection.java
```diff
@@ -0,0 +1,217 @@
...
+        this.wrapsElements = true;
+    }
+
+    private E view(E element) {
+        return elementView.apply(element);
+    }
+
+    // Serialize as the delegate: the interceptor is runtime-only wiring and instances may be captured
+    // in task state and stored to the configuration cache.
+    protected Object writeReplace() {
```
alllex, line 74: 💭 This is too specific of a comment for a generic mutation tracking collection. Maybe we could express this as a general serialization contract for the intercepting collections without being specific about CC?

It would also be good to make sure we have similar comments on both Collection and Map, because the Map is lacking one.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
alllex: We might have additional coverage, but I don't think it will be exhaustive. We could hand-wave at this and assume this is tested. I still think it will be much more explicit having a test here, which allows to have the fully expressed contract -- what's forbidden is indeed forbidden, and what's allowed is indeed allowed.

Having one additional data-driven test is not expensive neither in terms of added test time nor test code.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
alllex: @reinsch82 , can you explain how this is not a change in behavior for CC?
If there is code (in an included build) that was reading / printing the task names, would it keep printing the same thing on a CC hit after this change?

Can you also clarify why the inline comment about the test no longer applies?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,218 @@
...
+        // the outcome independent of the delegate's mutability. In diagnostics mode the mutation
+        // proceeds into the delegate, and some backing collections (e.g. the project properties
+        // populated by the CLI converter) are immutable in real distributions, so the build would
+        // additionally fail with an UnsupportedOperationException there.
+        isolatedProjectsFailsUsing(IsolatedProjectsMode.FAIL_FAST, "help", "-Pseed=value")
+
+        then:
+        fixture.assertIsolatedProjectsProblems(IsolatedProjectsMode.FAIL_FAST) {
+            projectsConfigured(":")
+            problem("Build file 'build.gradle': line 2: Cannot call '${signature}' on StartParameter after settings have been evaluated when Isolated Projects is enabled.")
```
alllex, line 206: 🤔 The violation message is not very kind in explaining that there is a larger category of things that is not allowed, and not a very specific case. We have this problem with IP violations generally, and should address it, but I feel like here it's a bit more pronounced, because the reporting is very mechanical.

Imagine the user reading `Cannot call 'getProjectProperties().entrySet().Entry.setValue(Object)' on StartParameter after settings have been evaluated`. _Why_ can't it be called? What _can_ be called? Why is this so _specific_?

We should provide users with the mental model: _mutating_ the state of the StartParameter after settings is not allowed. What they are observing is a concrete instance of this, that we help locating/explaining by providing _supplimentary_ information on how we got here.

subprojects/core/src/main/java/org/gradle/initialization/SettingsEvaluatedCallbackFiringSettingsProcessor.java
```diff
@@ -16,25 +16,39 @@
...
+            problems.report(factory ->
+                factory.problem(null, messageBuilder -> {
+                    messageBuilder.text(
+                        "Cannot call '" + methodSignature + "' on StartParameter after settings have been evaluated when Isolated Projects is enabled."
```
alllex, line 46: 💅  Use `messageBuilder.reference(..)` for code-like parts

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
reinsch82: I think you are right, it would be a behavior change.
I didn't follow that part through well enough.
I think we should reintroduce it.

platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/RootBuildLifecycleBuildActionExecutor.java
reinsch82: I'd leave this alone, included Builds always work on a fresh instance of their own StartParameter clone.
So there is no need to clear it.

alllex: I don't understand. If what you are saying is correct, that means we never clear listeners on those instances, but here we are removing a listener on something that never has one in the first place 🤔

I think we should investigate this deeper. I'm not saying that leaving that call that clears the mutator is incorrect. But at this point I am not certain about it. Can you explain a bit more on how the start parameter gets reused and how you know this?

Also, I just noticed that there is a `startParameter` field in this class, which we are clearing. But we actually create the root build with the `action.getStartParameter()`. Are those guaranteed to be always the same?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,218 @@
...
+class IsolatedProjectsStartParameterIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
+
+    def "mutating StartParameter after settings evaluation is a violation (#location)"() {
+        // The violation pipeline (onMutableCall -> listener -> IP problem) is the same for every setter
+        // and only the call location varies, so one representative setter exercised from each location
+        // covers it; that every individual setter is instrumented is checked by the fast reflective unit
+        // test StartParameterMutationInstrumentationTest. The root case has no included build on purpose:
+        // adding one would make its configured-project set depend on the mode (in fail-fast the root
+        // throws before the included build is configured, in diagnostics everything is configured).
+        if (scriptPath.startsWith("included/")) {
```
alllex, line 28: 🤔This makes assumptions about the implementation in an integration test. It does propose an alternative. However, I personally don't trust it much, because of the complexities of Gradle's implementation and wiring. I think we should have an integration test for all mutation methods. It doesn't seem like much of a cost in test time.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,218 @@
...
+        isolatedProjectsRun("a")
+
+        then:
+        outputContains("b executed")
+        fixture.assertStateStored {
+            projectsConfigured(":")
+        }
+    }
+
+    def "mutating a StartParameter collection view is reported with the access path (#invocation)"() {
```
alllex, line 190: ❓ This doesn't test `taskRequests` collection mutation. Do we miss anything else?

platforms/core-runtime/start-parameter/src/test/groovy/org/gradle/api/internal/StartParameterMutationInstrumentationTest.groovy
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private static List<Method> mutatingMethods() {
+        (StartParameter.methods.toList() + StartParameterInternal.methods.toList())
+            .findAll { isCandidateMutator(it) }
+            .unique { signature(it) }
+            .findAll { !NOT_NOTIFYING.contains(signature(it)) }
+            .sort { signature(it) }
+    }
+
```
alllex, line 80: 🤔 I think this is too fancy, which is a smell for a test implementation.

Our goal is to have a clear picture of which methods are notifying about the mutations, and which aren't. But we also want to avoid missing test coverage, if we modify the set of methods on the start parameter. The current test implemenatation achieves this, but does so rather implicitly.

We already have examples of methods that are mutating but don't start with `set` or `add`. If another one appears, the test coverage will degrade.

A better approach is to be fully explicit and have all methods of the `StartParameter[Internal]` specified as plain lists in the test. There are those that don't notify on mutation, those that do, and the rest are intercepting collections. All of those deserve to be tested.

The reflective check is still necessary, but it's much simpler -- check whether the number of methods in the hardcoded lists matches the number of methods on the class. No being clever is a good thing here.

Then, if the set of methods change in any way, one of the tests will fail, and we'll be forced to modify the test to account for a new method.

reinsch82: tests are still fancy, but use exact lists, and we'll detect any change.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
reinsch82: testing it now

https://github.com/gradle/gradle/blob/e0a52fb01075e35a356c3bf51934b76a74f4eebc/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy#L155

platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/RootBuildLifecycleBuildActionExecutor.java
reinsch82: So my understanding is that since this is the [RootBuildLifecycleBuildActionExecutor.java](https://github.com/gradle/gradle/pull/38152/files/b96bd9d7e50933421f0f8710f6ccbfb019cc2ac4#diff-839978af727132034639830ef90615d41f4f04afe3bb31aa389af7eb9f8e925f) it only applies to the root build actions. And it only matters for the root build action because we only reuse the root build action in the continues build.
In https://github.com/gradle/gradle/blob/8ca95ed4142f98b4c8e45487de61af3e12439e06/platforms/core-runtime/launcher/src/main/java/org/gradle/tooling/internal/provider/continuous/ContinuousBuildActionExecutor.java#L177 you can see that with a new run in a continues build, no new BuildAction is used, hence it uses the same one including the startParameter.
I dropped the StartParameter member in favor of the action startParameter, it is obvious now they are the same.
Also the services for the BuildSessionRegistry are instantiated from `action.getStartParameter()`.
@jvandort do you have anything to add or correct?

subprojects/core/src/main/java/org/gradle/initialization/SettingsEvaluatedCallbackFiringSettingsProcessor.java
jvandort: Probably not in this PR as-is

But, I think we will also want to soon emit deprecations in vintage/CC when IP is not enabled

platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/RootBuildLifecycleBuildActionExecutor.java
jvandort: The point here is that we reuse the same instance of `StartParameter` across different root build state instances during a continuous build. For included builds, they get a fresh copy so no clearing is necessary for them.

IMO we should probably also give the root build a "fresh" isolated SP instance when it is initialized, so we don't need to clear here.

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingCollection.java
jvandort: I'm not convinced these `writeReplace` methods are necessary at the moment. 

And if we do find that CC is serializing these maps (you use a collection from SP as a task input), the deserialized version will be mutable again. I suppose this is okay, but I want to first confirm this behavior if supported.

And, if it is supported, I think marking the `interceptor` as `transient` would be a cleaner solution than `writeReplace`. Notice we already have `transient` fields in `InterceptingCollection`. 

I don't recall writing this code but my name is associated with this code -- must have been a rebase of one of my commits plus some of your changes reinhold. I can't attest to why we have transient already. Can you speak to this @reinsch82 ?

reinsch82: 1. Yes I think your name attachment here ist due to refactoring and rebases.
2. if we don't use `writeReplace` it will deserialize as `Intercepting*` and the `interceptor` will be `null` therefore potentially causing NPEs.
So I think it is a reasonable choice to allow the `writeReplace` route.

PR #38153, opened 2026-06-09
Compare canonical paths in FileLock.isLockFile

cobexer:
## Summary

`DefaultFileLockManager$DefaultFileLock.isLockFile` compared paths with
`File.equals`, a structural path-string comparison. When the cache
directory is reachable via a path that differs from the stored lock
file path but resolves to the same physical location — symlinks,
junctions, or a Windows `subst`-aliased drive — two `File` instances
for the same lock file are not equal.

`DefaultPersistentDirectoryCache$Initializer.initialize` then failed to
skip the still-held lock file during re-initialization cleanup and
tried to delete it:

    java.io.IOException: Cannot delete file: ...lock

## Fix

Fall back to a canonical-path comparison when the fast `File.equals`
check fails, with a filename short-circuit so the canonicalization I/O
only runs for entries that could plausibly be the lock file. Both
checks only run during re-initialization, never on the hot read path.

### DIFF
--- a/platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFileLockManager.java
+++ b/platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/DefaultFileLockManager.java
@@ -229,7 +229,22 @@ public class DefaultFileLockManager implements FileLockManager {
 
         @Override
         public boolean isLockFile(File file) {
-            return file.equals(lockFile);
+            if (file.equals(lockFile)) {
+                return true;
+            }
+            // The same physical lock file may be reachable via different path strings
+            // (symlinks, junctions, Windows `subst`-aliased drives). Fall back to a
+            // canonical-path comparison so the lock-file identity check survives that.
+            // Short-circuit on filename first to avoid filesystem I/O on entries that
+            // obviously cannot be the lock file.
+            if (!file.getName().equals(lockFile.getName())) {
+                return false;
+            }
+            try {
+                return file.getCanonicalFile().equals(lockFile.getCanonicalFile());
+            } catch (IOException e) {
+                return false;
+            }
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


asodja:
Could you give a bit more context, or a scenario how can this happen? 

As far as I know `file.getCanonicalFile()` can be pretty expensive operation on some file systems, so I wonder if there is some performance penalty doing this.

Maybe we should atleast avoid getCanonicalFile() comparison for files that have a different name than this lock (I would assume the name of a lock file doesn't change) -> so we would return early `if (file.name != lockFile.name)`

cobexer:
Good point — pushed 3257d021c42 with the filename short-circuit:

```java
if (!file.getName().equals(lockFile.getName())) {
    return false;
}
```

So `getCanonicalFile()` only runs for entries that share the lock file's exact name (at most one or two per cache dir during re-init cleanup, which is itself a rare path).

I considered caching `lockFile.getCanonicalFile()` in the constructor too, but `determineLockTargetFile(target)` just appends `.lock` to the supplied target — it doesn't canonicalize — so that would add I/O to every cache open (a hot path) to save one canonicalization on the rare re-init path. The name short-circuit alone seemed like the better trade-off; happy to revisit if you'd prefer the cached form.

asodja:
How can this happen in the beginning, do we relocate cache once we acquire lock?

cobexer:
@asodja see for example https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_PerformanceTest6_Trigger/113964556?buildTab=overview

we set this up here: https://github.com/gradle/gradle/blob/0c0bac44f9ff9804af46ddf9dfbbd19dea82cf57/.teamcity/src/main/kotlin/common/PerformanceTestExtensions.kt#L79-L93

cobexer:
@bot-gradle test RfR

cobexer:
@bot-gradle test RfR

asodja (approved):
LGTM

cobexer:
This is not enough to run Gradle from a path that is [`subst`ed](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/subst) on Windows

PR #38156, opened 2026-06-09
Do not drop project lock in BuildOperationQueue when disallowing lock changes

octylFractal:
This leads to deadlocks when a lock for a resource is held above.

The behavior is kept for allowAccessToProjectState as there is potential for a different kind of deadlock to occur if the work being done needs the project lock in order to progress. In the future, we should work to ensure this is not necessary.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/internal/operations/DefaultBuildOperationQueue.java
+++ b/subprojects/core/src/main/java/org/gradle/internal/operations/DefaultBuildOperationQueue.java
@@ -16,6 +16,8 @@
 
 package org.gradle.internal.operations;
 
+import org.gradle.internal.Factories;
+import org.gradle.internal.Factory;
 import org.gradle.internal.UncheckedException;
 import org.gradle.internal.time.ExponentialBackoff;
 import org.gradle.internal.work.DefaultWorkerLimits;
@@ -173,7 +175,8 @@ class DefaultBuildOperationQueue<T extends BuildOperation> implements BuildOpera
         }
 
         // Need to wait for work to complete, so release worker lease while waiting
-        workerLeases.blocking(() -> {
+        // We purposefully only drop the project lock if the work might need it (allowAccessToProjectState)
+        withProjectLockChangePolicy(Factories.toFactory(() -> workerLeases.blocking(() -> {
             lock.lock();
             try {
                 // Wait for any work still running in other threads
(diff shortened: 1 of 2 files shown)


ljacomet:
@bot-gradle test this

subprojects/core/src/test/groovy/org/gradle/internal/operations/DefaultBuildOperationQueueTest.groovy
```diff
@@ -375,6 +380,249 @@ class DefaultBuildOperationQueueTest extends Specification {
...
+        allowAccessToProjectState | lockChangesDisallowed
+        false                     | true
+        true                      | false
+        policyDesc = lockChangesDisallowed ? "disallows" : "allows"
+    }
+
+    @Timeout(value = 60, unit = TimeUnit.SECONDS)
+    def "runAll does not lend project lock while waiting, avoiding deadlock against a resource held across the queue"() {
+        given:
+        // Mirrors the CI hang: the waiter owns some resource (there, cache ownership) plus a
```
alllex, line 489: 💅 "Mirrors the CI hang" will be out of context for future readers. Should we drop it?

alllex:
@bot-gradle test this

subprojects/core/src/test/groovy/org/gradle/internal/operations/DefaultBuildOperationQueueTest.groovy
```diff
@@ -375,6 +380,249 @@ class DefaultBuildOperationQueueTest extends Specification {
...
+
+        where:
+        allowAccessToProjectState | lockChangesDisallowed
+        false                     | true
+        true                      | false
+        policyDesc = lockChangesDisallowed ? "disallows" : "allows"
+    }
+
+    @Timeout(value = 60, unit = TimeUnit.SECONDS)
+    def "runAll does not lend project lock while waiting, avoiding deadlock against a resource held across the queue"() {
```
alllex, line 487: 🤔 It seems that it's not testing `runAll` but its approximation. I'd adjust the test name and/or add a comment explaining this.

subprojects/core/src/test/groovy/org/gradle/internal/operations/DefaultBuildOperationQueueTest.groovy
```diff
@@ -375,6 +380,249 @@ class DefaultBuildOperationQueueTest extends Specification {
...
+
+        cleanup:
+        releaseOperation.countDown()
+        other.interrupt()
+        waiter.join(10_000)
+        other.join(10_000)
+    }
+
+    @Timeout(value = 60, unit = TimeUnit.SECONDS)
+    def "runAllWithAccessToProjectState lends project lock while waiting, deadlocking against a resource held across the queue"() {
```
alllex, line 527: 🤔 Also here we are approximating what `runAllWithAccessToProjectState` does. Worth a comment or test rename

subprojects/core/src/main/java/org/gradle/internal/operations/DefaultBuildOperationQueue.java
```diff
@@ -173,7 +175,8 @@ private void waitForWorkToComplete() {
         }
 
         // Need to wait for work to complete, so release worker lease while waiting
-        workerLeases.blocking(() -> {
+        // We purposefully only drop the project lock if the work might need it (allowAccessToProjectState)
```
alllex, line 178: 🤔 The comment is confusing or I am reading it wrong: "only drop [it] if the work might need it"

I think the doc string on `withProjectLockChangePolicy` explains it better. We could have a shorter version of that here, maybe.

alllex (approved):
The change makes sense. I only have suggestions for comments and test naming.

It is somewhat concerning how complex and brittle-_looking_ the tests are. However, it is mostly a consequence of the existing design. It is better to have these tests than not.

PR #38157, opened 2026-06-09
Enhance error for unscheduled project transforms

### LINKED ISSUE #13567: Improve error reporting when configuration cache is enabled and task references the output of an artifact transform without declaring this as an input
The user can receive a cryptic internal exception when a task references the output of an artifact transform without declaring this as an input and the configuration cache is enabled.

This case is also broken for vintage builds, and fails in arbitrary ways based on the state of the file system, the tasks that are currently running and the transform implementation.

This is essentially a special case of the more general situation, where a task references the output of some other task or transform (or any work) without declaring that as an input.

### Expected Behavior

Inform the user that this situation will not work. There are many options for implementing this, ranging from just improving the error message for this specific case (transform outputs + configuration cache) through to deprecating all cases where a task uses the output of some other task or work without declaring this as an input for both vintage and configuration cache builds.

### Current Behavior

Writing to the configuration cache fails with a cryptic internal exception, when configuration cache is enabled and fails in different ways when not enabled.

tresat:
When a task resolves a project artifact via an artifact transform without declaring the producing project as a task input, the Configuration Cache prevents the project from being scheduled. This previously resulted in a vague 'project not found' error.

Introduce `UnknownProjectStateException` to provide a clear, actionable message that guides users to declare their task inputs and references the upgrading guide.

Fixes #13567

Related to #38105 (which adds a deprecation for a similar, but not identical, scenario).

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
@@ -3005,7 +3005,7 @@ Found the following transformation chains:
         if (configurationCache) {
             failure.assertHasCause("Could not resolve all files for configuration ':sizes'.")
             failure.assertThatCause(matchesRegexp("Failed to transform lib\\.jar \\(project ':lib'\\) to match attributes \\{.*\\}\\."))
-            failure.assertHasCause("Could not access project :lib. No task declared :lib as part of an input, so it was not scheduled. Properly declare all task inputs (including the result of any dependency resolutions) to ensure this project is scheduled for execution.")
+            failure.assertHasCause("Could not access project ':lib'. No task declared this project as part of an input, so it was not scheduled. Properly declare all task inputs (including the result of any dependency resolutions) to ensure this project is scheduled for execution.")
             failure.assertHasResolution("Declare the files or artifacts produced by the configuration using the transform as a task input to properly wire it into the execution plan.")
             failure.assertHasResolution("Consult the upgrading guide for further information: " + new DocumentationRegistry().getDocumentationFor("upgrading_version_9", "undeclared_artifact_transform_input"))
         } else {
@@ -3086,7 +3086,7 @@ Found the following transformation chains:
 
         then:
         failure.assertHasCause("Could not resolve all files for configuration ':sizes'.")
-        failure.assertHasCause("Could not access project :lib-build:producer. No task declared :lib-build:producer as part of an input, so it was not scheduled. Properly declare all task inputs (including the result of any dependency resolutions) to ensure this project is scheduled for execution.")
+        failure.assertHasCause("Could not access project ':lib-build:producer'. No task declared this project as part of an input, so it was not scheduled. Properly declare all task inputs (including the result of any dependency resolutions) to ensure this project is scheduled for execution.")
         failure.assertHasResolution("Declare the files or artifacts produced by the configuration using the transform as a task input to properly wire it into the execution plan.")
         failure.assertHasResolution("Consult the upgrading guide for further information: " + new DocumentationRegistry().getDocumentationFor("upgrading_version_9", "undeclared_artifact_transform_input"))
     }
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
@@ -16,10 +16,12 @@
 
 package org.gradle.integtests.resolve.transform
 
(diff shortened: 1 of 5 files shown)


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
```diff
@@ -2921,6 +2923,168 @@ Found the following transformation chains:
...
+            succeeds "resolve"
+        }
+
+        then:
+        if (configurationCache) {
+            failure.assertHasCause("Could not resolve all files for configuration ':sizes'.")
+            failure.assertThatCause(matchesRegexp("Failed to transform lib\\.jar \\(project ':lib'\\) to match attributes \\{.*\\}\\."))
+            failure.assertHasCause("Could not access project :lib. No task declared :lib as part of an input, so it was not scheduled. Properly declare all task inputs (including the result of any dependency resolutions) to ensure this project is scheduled for execution.")
+            failure.assertHasResolution("Declare the files or artifacts produced by the configuration using the transform as a task input to properly wire it into the execution plan.")
+            failure.assertHasResolution("Consult the upgrading guide for further information: " + new DocumentationRegistry().getDocumentationFor("upgrading_version_9", "undeclared_artifact_transform_input"))
```
big-guy: You can use a regular link here and it'll be normalized

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
```diff
@@ -2921,6 +2923,168 @@ Found the following transformation chains:
...
+        if (configurationCache) {
+            fails "resolve"
+        } else {
+            succeeds "resolve"
+        }
+
+        then:
+        if (configurationCache) {
+            failure.assertHasCause("Could not resolve all files for configuration ':sizes'.")
+            failure.assertThatCause(matchesRegexp("Failed to transform lib\\.jar \\(project ':lib'\\) to match attributes \\{.*\\}\\."))
```
big-guy: Why does this need to be a regex?

subprojects/core/src/main/java/org/gradle/api/internal/project/UnknownProjectStateException.java
```diff
@@ -0,0 +1,65 @@
...
+        this.identifier = identifier;
+    }
+
+    public ProjectComponentIdentifier getIdentifier() {
+        return identifier;
+    }
+
+    private static String buildMessage(ProjectComponentIdentifier identifier) {
+        String path = identifier.getBuildTreePath();
+        return "Could not access project " + path + ". "
```
big-guy: We shouldn't build the display name here (project + path).  Does ProjectComponentIdentifier have a display name we can use?

big-guy (changes requested):

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
tresat: I think this was Claude trying to do some sort of minimal verification.  Replaced with fixed output assertion.

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
```diff
@@ -2927,6 +2929,168 @@ Found the following transformation chains:
...
+        executer.withArgument("--configuration-cache")
+
+        when:
+        fails "resolve"
+
+        then:
+        failure.assertHasCause("Could not resolve all files for configuration ':sizes'.")
+        failure.assertHasCause("Could not access project ':lib-build:producer'. No task declared this project as part of an input, so it was not scheduled. Properly declare all task inputs (including the result of any dependency resolutions) to ensure this project is scheduled for execution.")
+        failure.assertHasResolution("Declare the files or artifacts produced by the configuration using the transform as a task input to properly wire it into the execution plan.")
+        failure.assertHasResolution("Consult the upgrading guide for further information: " + new DocumentationRegistry().getDocumentationFor("upgrading_version_9", "undeclared_artifact_transform_input"))
```
big-guy: This one too

big-guy (approved):

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/transform/ArtifactTransformIntegrationTest.groovy
tresat: OK, found the reason on CI.  The attribute `org.gradle.jvm.version` differs when run in the different CI test builds, so the regexp is needed for that.

PR #22312, opened 2022-10-04
verification-metadata Trust.Reasons and Checksum.Reasons

mathjeff:
Fixes #20260

### Context
This should allow us to specify a Reason for Trust entries and sha256 entries.
We're already specifying these values via commands and/or nonstandard attributes, and having Gradle recognize them allows us to observe fewer differences when asking Gradle to regenerate the dependency-verification.xml

### Contributor Checklist
- [X] [Review Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md)
- [X] Make sure that all commits are [signed off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff) to indicate that you agree to the terms of [Developer Certificate of Origin](https://developercertificate.org/).
- [X] Make sure all contributed code can be distributed under the terms of the [Apache License 2.0](https://github.com/gradle/gradle/blob/master/LICENSE), e.g. the code was written by yourself or the original code is licensed under [a license compatible to Apache License 2.0](https://apache.org/legal/resolved.html).
- [X] Check ["Allow edit from maintainers" option](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) in pull request so that additional changes can be pushed by Gradle team
- [X] Provide integration tests (under `<subproject>/src/integTest`) to verify changes from a user perspective
- [ ] Provide unit tests (under `<subproject>/src/test`) to verify logic
- [ ] Update User Guide, DSL Reference, and Javadoc for public-facing changes
- [x] Ensure that tests pass sanity check: `./gradlew sanityCheck`
- [x] Ensure that tests pass locally: `./gradlew <changed-subproject>:quickTest`

### Gradle Core Team Checklist
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation
- [ ] Recognize contributor in release notes

PR #38163, opened 2026-06-10
Record origin/reason on trusted-key and pgp verification entries

### LINKED ISSUE #37538: Verification-metadata: Add support to record the URL where key or checksum was verified
### Expected Behavior

In the verification-metadata.xml file, after verifying either a public key ID or checksum, I want to be able to record the URL that was used to do the verification. Please add a "source" field in either the <trusted-key>, <pgp>, <sha512>, etc XML elements that can be used to record the URL for future reference.

Currently there is no way to record the URL that was used to validate either the public key ID or the checksum.

### Context

Currently once a public key ID or checksum is verified, there is no good way to record the URL that was used to validate. There is a field to record whether verification was done, but we need a way to store the URL so in the future it can be easily referenced again. For example if a new version of a dependency is being used, the same URL probably has the new checksums, and if it's stored in the verification-metadata.xml file, it can be easily referenced again.

ljacomet:
Dependency-verification checksums already carry optional `origin` and `reason` attributes documenting where a checksum was obtained and why it is trusted. This extends the same mechanism to the `<trusted-key>` (global) and `<pgp>` (artifact-specific) elements so the provenance of a public key (e.g. the URL where it was verified) can be recorded alongside the key.

- Add dependency-verification-1.4.xsd with origin/reason on trusted-keyType and pgpType; the writer now emits the 1.4 schema.
- Add a TrustedPgpKey model (value + origin + reason) and thread it through the reader, writer, builder, configuration and verifier.
- Carry origin/reason on the global TrustedKey configuration entry.
- Preserve full backwards compatibility: pre-1.4 files keep reading, with null origin/reason; existing builder APIs keep working via overloads.

The attributes are informational only and are never used during verification.

Fixes #37538


lkasso (approved):
Docs LGTM!

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/verification/DependencyVerificationSignatureCheckIntegTest.groovy
```diff
@@ -1338,6 +1338,75 @@ If the artifacts are trustworthy, you will need to update the gradle/verificatio
...
+        serveValidKey()
+
+        then:
+        succeeds ":compileJava"
+
+        and: "origin and reason are preserved in the metadata file"
+        def metadata = file("gradle/verification-metadata.xml").text
+        metadata.contains('dependency-verification-1.4.xsd')
+        metadata.contains('origin="https://example.com/key.asc"')
+        metadata.contains('reason="verified manually"')
```
ov7a: 💅 It should be contained twice - would be nice to validate that. Or even parse the metadata properly

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/serializer/DependencyVerificationsXmlReader.java
```diff
@@ -260,6 +262,8 @@ private void addTrustedArtifact(Attributes attributes) {
 
         private void addTrustedKey(Attributes attributes) {
             currentTrustedKey = getAttribute(attributes, ID);
+            currentTrustedKeyOrigin = getNullableAttribute(attributes, ORIGIN);
```
ov7a, line 282: ❌  I'm confused by this change - here, global props (current*) are set just to be read in the next call - can we avoid having this pattern? What's the reason for it?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/verifier/DependencyVerifierBuilder.java
```diff
@@ -244,7 +257,11 @@ List<Checksum> buildChecksums() {
         }
 
         public void addTrustedKey(String key) {
```
ov7a: 🤔 What's the benefit of keeping this and similar methods compared to passing null explicitly? There's no external consumers IIUC

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/verifier/DependencyVerificationConfiguration.java
```diff
@@ -273,13 +284,17 @@ public boolean equals(@Nullable Object o) {
 
             TrustedKey that = (TrustedKey) o;
 
-            return keyId.equals(that.keyId);
+            if (!keyId.equals(that.keyId)) {
+                return false;
+            }
+            return Objects.equals(origin, that.origin);
```
ov7a: 🤔 Why does origin matter here?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/serializer/DependencyVerificationsXmlWriter.java
```diff
@@ -150,10 +151,14 @@ private void writeGloballyTrustedKeys(DependencyVerificationConfiguration config
...
         writeAttribute(ID, keyId);
+        // origin and reason describe the key itself, so they are shared by all the grouped entries
+        DependencyVerificationConfiguration.TrustedKey representative = trustedKeys.get(0);
+        writeNullableAttribute(ORIGIN, representative.getOrigin());
+        writeNullableAttribute(REASON, representative.getReason());
         trustedKeys.stream().sorted().forEach(trustedKey -> {
             try {
                 writer.startElement(TRUSTING);
-                writeTrustCoordinates(trustedKey);
+                writeCoordinates(trustedKey);
```
ov7a, line 166: ❓  Previously, you could have a reason per trusted coordinates, right? 
Note the semantic difference: reason to trust the key vs reason to trust particular coordinates.

Isn't it a breaking change?

ov7a (changes requested):

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/verification/DependencyVerificationSignatureCheckIntegTest.groovy
ljacomet: Sorry, on second read, those tests are pointless. They test the fixtures ...

Replacing by a test confirming that running generation preserves those added comments.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/serializer/DependencyVerificationsXmlWriter.java
ljacomet: No, you could not have anything - look at the comparison between 1.3 and 1.4 of the XSD.

So this PR adds fields to the `pgp` and `trusted-key` nodes. Maybe it should also add it to the `trusting` nodes.

I'll look into that.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/serializer/DependencyVerificationsXmlReader.java
ljacomet: This was a consequence of the difference between the internal model and the XML.
Now that I have added origin and reason on the `trusting` level, those are the defaults when they are not overridden at the lower level.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/verifier/DependencyVerificationConfiguration.java
ljacomet: It should not, `reason` also was made to matter when it should not have in #22312, I pulled both out.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/verifier/DependencyVerifierBuilder.java
ljacomet: Right, removed

ljacomet:
Hey @ov7a ,

I addressed all your comments, PTAL

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/serializer/DependencyVerificationsXmlReader.java
```diff
@@ -94,7 +96,22 @@ public static void readFromXml(InputStream in, DependencyVerifierBuilder builder
...
+        // Only nag when reading the file for verification. The write path reads the file through the
+        // two-argument overload with its own builder and is expected to rewrite (and so collapse) it.
+        nagAboutDuplicateTrustEntries(builder.getDroppedDuplicateTrustEntries());
+        return verifier;
+    }
+
+    private static void nagAboutDuplicateTrustEntries(List<String> droppedDuplicateTrustEntries) {
+        for (String entry : droppedDuplicateTrustEntries) {
+            DeprecationLogger.deprecateBehaviour("The dependency verification metadata declares a duplicate " + entry + " that differs only by its origin or reason.")
+                .withContext("Trust entries are identified by their coordinates, so such duplicates are redundant: only the first one is kept and the duplicate, including its origin and reason, is removed the next time the metadata file is written (for example with --write-verification-metadata).")
```
ov7a, line 109: ❌ Please, cover this with a test

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/serializer/DependencyVerificationsXmlWriter.java
ov7a: Ah, ok, the old code was just confusing.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/verifier/DependencyVerifierBuilder.java
```diff
@@ -243,8 +296,8 @@ List<Checksum> buildChecksums() {
                 .collect(Collectors.toList());
         }
 
-        public void addTrustedKey(String key) {
-            pgpKeys.add(key.toUpperCase(Locale.ROOT));
+        public void addTrustedKey(String key, @Nullable String origin, @Nullable String reason) {
+            pgpKeys.add(new TrustedPgpKey(key, origin, reason));
```
ov7a: 🤔 If we want to report duplicate entries, we should do it here as well (because pgpKeys is a set)

ov7a (changes requested):

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/serializer/DependencyVerificationsXmlReader.java
ljacomet: Done already:
* https://github.com/gradle/gradle/pull/38163/changes#diff-3bb03b5815dda5515f92b34e87d7f797febe6e6ba48552d7cf02d5170bc326ffR382
* https://github.com/gradle/gradle/pull/38163/changes#diff-3bb03b5815dda5515f92b34e87d7f797febe6e6ba48552d7cf02d5170bc326ffR408

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/verifier/DependencyVerifierBuilder.java
ljacomet: Yes, adding that

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/serializer/DependencyVerificationsXmlReader.java
ov7a: Indeed, the deprecation line is covered.

I searched by context string ("Trust entries are identified...") and found nothing - can it also be covered?

ov7a (approved):

PR #38165, opened 2026-06-10
Report project lifecycle related callbacks on `Gradle`

6hundreds:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


6hundreds:
@bot-gradle test prf on linux

6hundreds:
@bot-gradle test prf

blindpirate:
@bot-gradle test prf

6hundreds:
@bot-gradle test prf on linux

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -214,28 +172,45 @@ class CrossProjectConfigurationReportingGradle(
...
-            delegate.beforeEvaluate(crossProjectModelAccess.access(referrerProject, project as ProjectInternal))
-        }
 
-        override fun afterEvaluate(project: Project, state: ProjectState) {
-            delegate.afterEvaluate(crossProjectModelAccess.access(referrerProject, project as ProjectInternal), state)
-        }
+    override fun addListener(listener: Any) {
+        onMutableStateAccess("addListener")
+        delegate.addListener(listener)
+    }
```
6hundreds, line 187: ❗ This is a behavior change actually. Before - we're allowing `ProjectEvaluationListener` passing, and delegate otherwise(see `maybeWrapListener` above). On the delegate side there is a logic [in regard of buildSrc builds and other supported listeners ](https://github.com/gradle/gradle/blob/9562b8d308231eea1e429db863fbf26953e030f2/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/initialization/ConfigurationCacheProblemsListener.kt#L158). Now we report this unconditionally.

CI shows smoke tests are green. @gradle/bt-isolated-projects-project WDYT?

alllex:
@bot-gradle test on linux

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
alllex: We usually classify a behavior change as something that doesn't change the outcome of the build. E.g. removing parent lookup was a behavior change, because the build might succeed before and after (same outcome), but the results might be different (`findProperty()` returned null instead of a value).

In this case, it's an additional constraint -- the build that was previously succeeding will start failing. It's okay to add new constraints for IP. We can introduce more excessive constraints if necessary, if this significantly simplifes the implementation, as long as we don't cut out valid use cases (accidentally).

I didn't dive deep, but it seems that `addListener()` is the only way in the public API to register `DependencyResolutionListener`. Can you double-check? If that's true, we'll be saying that with IP you can use that listener at all (from the project scope at least). It's build-scoped, and looking at its API surface, it's not immediately clear to me it's safe for parallel access to the data it exposes. cc @jvandort

Apart from that, I believe Gradle should move away from untyped APIs. If IP provides such an opportunity and it is somewhat justified, I think we should take it. At least we should try and see what feedback we get.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -214,28 +172,45 @@ class CrossProjectConfigurationReportingGradle(
...
+        return delegate.addProjectEvaluationListener(listener)
+    }
+
+    override fun removeProjectEvaluationListener(listener: ProjectEvaluationListener) {
+        onMutableStateAccess("removeProjectEvaluationListener")
+        delegate.removeProjectEvaluationListener(listener)
+    }
+
+    override fun beforeProject(closure: Closure<*>) {
+        onMutableStateAccess("beforeProject")
```
alllex, line 209: 💭  We are not explaining it to the users yet, but we should at least be explicit ourselves about exactly why we are prohibiting this. I think a comment in the code would help. This really applies to all violations we introduce, and we haven't been explicit about this before, unfortunately.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
jvandort: `DependencyResolutionListener` gives access to `ResolvableDependencies`, which is mutable state of a `Configuration`.

6hundreds:
@bot-gradle test this on linux

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
alllex: @6hundreds, I think we should add an inline comment explaining these nuances, at least with regards to the listeners we forbid this way, which were allowed with CC

6hundreds:
@bot-gradle test this on linux

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
6hundreds: Added a comment

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
6hundreds: Added an explanation

6hundreds: > I didn't dive deep, but it seems that addListener() is the only way in the public API to register DependencyResolutionListener

Yeah, I didn't find any other option to register it on the project scope

6hundreds:
@alllex Thanks for the review, I've added some clarifications. PTAL!

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -214,28 +172,66 @@ class CrossProjectConfigurationReportingGradle(
...
-    private
-    class CrossProjectModelAccessProjectEvaluationListener(
-        private val delegate: ProjectEvaluationListener,
-        private val referrerProject: ProjectIdentity,
-        private val crossProjectModelAccess: CrossProjectModelAccess
-    ) : ProjectEvaluationListener {
-        override fun beforeEvaluate(project: Project) {
-            delegate.beforeEvaluate(crossProjectModelAccess.access(referrerProject, project as ProjectInternal))
+    override fun addListener(listener: Any) {
+        if (isSupportedListener(listener)) {
```
alllex: ❌ I don't think the IP and CC problems are mutually exclusive. They are there for different reasons. Plus, it's brittle to simulate the CC check here: it already differs, because we don't have the buildSrc check here. I think this should be unconditional IP violation reporting.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
alllex: ❌ The comment still misses the explanation about the `DependencyResolutionListener`

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -214,28 +172,66 @@ class CrossProjectConfigurationReportingGradle(
...
+    override fun removeListener(listener: Any) {
+        onMutableStateAccess("removeListener")
+        delegate.removeListener(listener)
+    }
 
-        override fun equals(other: Any?): Boolean =
-            javaClass == (other as? CrossProjectModelAccessProjectEvaluationListener)?.javaClass &&
-                other.delegate == delegate &&
-                other.referrerProject == referrerProject
+    // This let a project tie its mutable state to the lifecycle of other projects
```
alllex: 💅 `Prevent tying project's mutable state to the lifecycle of other projects`

I'd also move it directly before the corresponding `onMutableStateAccess` call, because the comment is about the violation, not about the method.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -214,28 +172,66 @@ class CrossProjectConfigurationReportingGradle(
...
+    // and receiving notifications in an unspecified order.
+    override fun addProjectEvaluationListener(listener: ProjectEvaluationListener): ProjectEvaluationListener {
+        onMutableStateAccess("addProjectEvaluationListener")
+        return delegate.addProjectEvaluationListener(listener)
+    }
 
-        override fun hashCode(): Int = Objects.hash(delegate, referrerProject)
+    // This let a project tie its mutable state to the lifecycle of other projects
+    // and receiving notifications in an unspecified order.
+    override fun removeProjectEvaluationListener(listener: ProjectEvaluationListener) {
```
alllex: 💅 I don't think this duplication is necessary and the comment is not very accurate. The removal doesn't really do anything dangerous. But we have a violation for symmetry with the `addProjectEvaluationListener`. That's would be the comment I put next to the mutable access call.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -214,28 +172,66 @@ class CrossProjectConfigurationReportingGradle(
...
+        onMutableStateAccess("removeProjectEvaluationListener")
+        delegate.removeProjectEvaluationListener(listener)
+    }
+
+    // This let a project tie its mutable state to the lifecycle of other projects
```
alllex: 💅 I'd avoid duplicating the reasoning, and instead put `See addProjectEvaluationListener` in the rest of the callbacks. This removes noise and indicates a single _shared_ reason for why all of these are forbidden.

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/DeferredTaskConfigurationIntegrationTest.groovy
```diff
@@ -219,6 +219,10 @@ class DeferredTaskConfigurationIntegrationTest extends AbstractDeferredTaskDefin
         succeeds 'assertActionExecutionOrder'
     }
 
+    @Requires(
+        value = TestExecutionPreconditions.NotIsolatedProjects,
+        reason = "before/afterProject is prohibited under IP. Replace with @UnsupportedWithIsolatedProjects with iteration matcher"
```
alllex: ❓ Can we replace the annotation now?

Applies to a couple of tests

subprojects/core/src/integTest/groovy/org/gradle/configuration/ExecuteUserLifecycleListenerBuildOperationIntegrationTest.groovy
```diff
@@ -335,7 +336,7 @@ class ExecuteUserLifecycleListenerBuildOperationIntegrationTest extends Abstract
         verifyHasNoChildren(subBeforeEvaluated, subOtherScriptAppId)
     }
 
-    @ToBeFixedForIsolatedProjects(because = "Access to root from projects")
+    @UnsupportedWithIsolatedProjects(because = "Gradle.(afterProject/addListener/addProjectEvaluationListener) cannot be used at project scope with Isolated Projects")
```
alllex: ❓ Technically, this loses useful coverage. The init script and settings script cases should continue working under IP, IIUC. Ideally, we split the test or make it data-driven to make the testing more precise.

Though, I wouldn't necessarily do it in the same PR. I think a good compromise is to make this as "to be fixed", and create an issue to split out the tests later.

Same might apply to other tests in the class

alllex (changes requested):

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
6hundreds: >  I think this should be unconditional IP violation reporting.

It leads to double reporting then for unsupported listeners. Once we report for IP, and then delegate reports for CC. It may be worked around by introducing `BuildScopeListenerRegistrationListener` instance for IP, that will be no-op for all buildListeners registrations, or add an IP mode check to the existing implementation of `BuildScopeListenerRegistrationListener`

subprojects/core/src/integTest/groovy/org/gradle/api/tasks/DeferredTaskConfigurationIntegrationTest.groovy
6hundreds: Yes, missed it

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
alllex: I am saying it's not double reporting, because we report for different reasons. Can you show what the double reporting looks like, if we have it?

6hundreds:
```
2 problems were found storing the configuration cache.
- Build file 'a/build.gradle': line 2: Project ':a' cannot access Gradle.addListener
- Build file 'a/build.gradle': line 2: registration of listener on 'Gradle.addListener' is unsupported
```

alllex: This is unfortunate wording and an additional complication that IP problems are clumped together with CC problems.

First of all, the double reporting will only happen in the Diagnostic IP, right? If so, I would worry even less about the double reporting.

Secondly, we can do better with the IP message and clarify that the violation occurs because of the constraint that project-scoped logic must not modify mutable Gradle state, which the added listeners allow to do. Though I don't think we should spend a lot of time now to refine this message, and instead we should separately look at how we phrase IP problems and improve them.

I'd also consider that we might pursue having IP-without-CC, and it is better to have the requirement sets independent.

My position stays unchanged -- we should avoid checking for unsupported visitors and have a blanket ban for IP

@gradle/bt-isolated-projects-project, what do others think?

subprojects/core/src/integTest/groovy/org/gradle/configuration/ExecuteUserLifecycleListenerBuildOperationIntegrationTest.groovy
6hundreds: Makes sense!

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
6hundreds: > First of all, the double reporting will only happen in the Diagnostic IP, right? If so, I would worry even less about the double reporting.

Testing this might be a bit cumbersome. It will be different set of problems for fail-fast and diagnostics. 

Current solution with simulation CC check is chosen because it keeps all the reporting logic in `CrossProjectReportingGradle`. Introducing of `BuildScopeListenerRegistrationListener` would eliminate your concern, but make reporting logic spread across several classes.

alllex: > Testing this might be a bit cumbersome. It will be different set of problems for fail-fast and diagnostics.

So, an extra if inside the problem checking closure?

> Current solution with simulation CC check is chosen because it keeps all the reporting logic in CrossProjectReportingGradle. 

I think it only appear that way. The logic is still split between IP and CC, and for some reason we are trying to compesate for the CC part in the IP part. That's what I don't like.

> Entroducing of BuildScopeListenerRegistrationListener

This complicates things further, I would not do that

6hundreds:
@alllex Made `addListener` report unconditionally and add a test coverage for that, fixed comments. PTAL!

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -214,28 +172,59 @@ class CrossProjectConfigurationReportingGradle(
...
-    class CrossProjectModelAccessProjectEvaluationListener(
-        private val delegate: ProjectEvaluationListener,
-        private val referrerProject: ProjectIdentity,
-        private val crossProjectModelAccess: CrossProjectModelAccess
-    ) : ProjectEvaluationListener {
-        override fun beforeEvaluate(project: Project) {
-            delegate.beforeEvaluate(crossProjectModelAccess.access(referrerProject, project as ProjectInternal))
-        }
+    override fun addListener(listener: Any) {
+        // IP prohibits project scope registration even for supported listeners:
```
alllex: 💅 `even for listeners supported by CC`

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -214,28 +172,59 @@ class CrossProjectConfigurationReportingGradle(
...
+        // - ProjectEvaluationListener, see `addProjectEvaluationListener` method
+        // - TaskExecutionGraphListener, as it gives access mutable state of TaskExecutionGraph
+        // - DependencyResolutionListener, as it gives access to ResolvableDependencies, which is mutable state of a Configuration
+        //
+        // Moreover, vintage considering any listener as supported on buildSrc build,
```
alllex: ❔ Is it vintage or is it CC?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -214,28 +172,59 @@ class CrossProjectConfigurationReportingGradle(
...
-        private val referrerProject: ProjectIdentity,
-        private val crossProjectModelAccess: CrossProjectModelAccess
-    ) : ProjectEvaluationListener {
-        override fun beforeEvaluate(project: Project) {
-            delegate.beforeEvaluate(crossProjectModelAccess.access(referrerProject, project as ProjectInternal))
-        }
+    override fun addListener(listener: Any) {
+        // IP prohibits project scope registration even for supported listeners:
+        // - ProjectEvaluationListener, see `addProjectEvaluationListener` method
+        // - TaskExecutionGraphListener, as it gives access mutable state of TaskExecutionGraph
```
alllex: 💭 While this is true, the mutable state of the execution graph is protected by violations, right? At least the one you get via `getTaskGraph().whenReady {}` is protected. Does the same wrapping apply here (if we were to allow `TaskExecutionGraphListener` as an argument) ?

So the motivation here is more that it's pretty much the only listener that we could support in for IP here. But given the `taskGraph.whenReady` option, I am tempted to be more restrictive, at least for now.

Do you agree?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
```diff
@@ -93,14 +93,88 @@ class IsolatedProjectsCrossProjectGradleAccessIntegrationTest extends AbstractIs
...
+        when:
+        isolatedProjectsFailsUsing mode, "help"
+
+        then:
+        fixture.assertIsolatedProjectsProblems(mode) {
+            projectsConfigured(*configuredProjects)
+            expectedProblems(mode, accessor)
+                .each {
+                    problem("Build file '$buildFileName': line 2: $it")
+                }
```
alllex, line 133: 💭  I had something like this in mind:
```suggestion
            problem("Build file '$buildFileName': line 2: Project '$accessor' cannot access Gradle.addListener")
            if (mode != IsolatedProjectsMode.FAIL_FAST && !alwaysSupportedByCC && accessor != ":buildSrc") {
                problem("Build file '$buildFileName': line 2: registration of listener on 'Gradle.addListener' is unsupported")
            }
```

This would still benefit from a one-line comment, but I think this would make the data tables easier.

Also, `combined:` 🤩

alllex (approved):
LGTM!

With some suggestions to consider

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
6hundreds: CC for sure

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
6hundreds: > Does the same wrapping apply here (if we were to allow TaskExecutionGraphListener as an argument) ?

At the moment? It doesn't.

> Do you agree?

In the essence yes, wanted to be a bit more concise here as amount of prose to be written to report a single method started to tire me out

alllex: > At the moment? It doesn't.

I think it's an important detail, because we were actually missing reporting violations as we do for `gradle.taskGraph.whenReady { <actions> }`.

> amount of prose to be written to report a single method started to tire me out

It's important prose! It will save time for our future self to avoid reverse-engineering the reasons and decisions

We should still be concise with the comment, but not skip over valuable details

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsCrossProjectGradleAccessIntegrationTest.groovy
6hundreds: I don't see how it's better than the current version. Also, `if` condition should be corrected to 
```
if (mode != IsolatedProjectsMode.FAIL_FAST &&
    !alwaysSupportedByCC && 
    accessor != ":buildSrc" && 
    invocation.startsWith("addListener"))
```
Right?

alllex: It's simpler to read the test this way, in my subjective opinion, because there are fewer computations you need to look at to understand the expectation. The expectations become more local.
It's just a suggestion: take or leave it

I think the condition shouldn't be adjusted, but we'd need a data column telling us which _method_ is forbidden. It will be either `addListener` or `removeListener`. The `removeListener` cases will always be supported by CC, hence the need for that boolean column.

PR #38166, opened 2026-06-10
Rework allIncubationReports into a rich single-page report

ljacomet:
The aggregated incubating-API report was a bare list of names grouped by version with no links or metadata. Replace it with a self-contained HTML page: a sortable/searchable table (API/Kind/Since/Project/Age), a summary block (totals, per-since distribution, top projects), GitHub source links, age-in-releases highlighting for long-incubating APIs, and an on-page CSV download. All client-side behaviour is vanilla inline JS/CSS over a single JSON data block; no CDN libraries.

The per-since distribution doubles as a filter: clicking a version (or the "unknown since" bucket) sets up the matching filter and scrolls to the table.

To support this:
- emit the API kind (class/interface/.../method/property/...) into the intermediate txt report (new IncubatingApiKind enum), parsed by the aggregator; the name field is split robustly so semicolons are safe
- thread version.txt and released-versions.json into the aggregator and compute each API's age in minor releases (shared ReleasedVersions helper, reused by the per-project producer)
- lift the GitHub blob-URL builder into a helper shared by CSV and HTML

The CSV output format is unchanged. The task name, output paths, and configuration-cache compatibility are preserved.

### DIFF
--- a/build-logic/buildquality/src/main/kotlin/gradlebuild/incubation/action/ReleasedVersions.kt
+++ b/build-logic/buildquality/src/main/kotlin/gradlebuild/incubation/action/ReleasedVersions.kt
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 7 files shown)


build-logic/buildquality/src/main/kotlin/gradlebuild/incubation/action/ReleasedVersions.kt
```diff
@@ -0,0 +1,69 @@
...
+ */
+object ReleasedVersions {
+
+    data class ReleasedVersion(val version: String, val date: String)
+
+    /**
+     * All released versions in the order they appear in the file (newest first), each with its
+     * release date formatted as `yyyy-MM-dd`.
+     */
+    fun parse(releasedVersionsFile: File): List<ReleasedVersion> {
```
blindpirate, line 36: Mind adding a few test for it? Looks like very easy for unit tests.

blindpirate (approved):
LGTM with a tiny comment

build-logic/buildquality/src/main/kotlin/gradlebuild/incubation/action/ReleasedVersions.kt
ljacomet: Done

PR #38173, opened 2026-06-10
Deprecate undeclared transform inputs 

### LINKED ISSUE #37219: Undeclared artifact transform dependencies should be disallowed rather than silently accepted

tresat:
This branch deprecates a long-standing footgun in Gradle's artifact-transform machinery: reading the files of a `FileCollection` whose artifacts are produced by an artifact transform of a **project** artifact, from inside a task action that did not declare those files as a task input. The pattern silently runs the transform inline on the calling thread (non-CC) and outright fails with a confusing `project <name> not found` error under Configuration Cache if the project containing the task necessary to produce the files was not scheduled.

The detection mechanism is built around per-task declarer attribution computed from the fully-resolved work graph, which catches the pattern reliably across chained transforms, `@InputArtifactDependencies` cross-set relationships, `componentFilter`-restricted artifact views, and eager configuration-time resolution scenarios — none of which the original prototype handled correctly.

The deprecation is emitted **once per (undeclared task, transform-node, build)**. A single task with multiple query sites in the same action sees one warning, not one per call site; this is by design and dedup-tracked on the node.

## How undeclared resolution is detected

The mechanism is a **post-graph walk** in `DefaultExecutionPlan.determineExecutionPlan()`:

1. A new marker interface `TaskDeclarationAware` (in `subprojects/core/.../execution/plan/`) is implemented by `TransformStepNode`. Marking lives in the dependency-management layer; the interface lives in core so the execution-plan code can see it without violating the module dependency direction (`:dependency-management` → `:core`, never the reverse).

2. After the execution plan finishes resolving all dependency relationships, `propagateTaskDeclarationsToTransitiveSuccessors()` runs a BFS from every `LocalTaskNode`, walking the dependency-successor edges of the live work graph. For each reachable `TaskDeclarationAware` node (i.e. every `TransformStepNode`), it calls `markDeclaredBy(task)`. The walk **stops at any other `LocalTaskNode` boundary** — a task `B` that `dependsOn` task `A` does *not* inherit A's declarations, preserving per-task semantics.

3. At task execution time, `TransformStepNode.executeIfNotAlready()` consults `WorkExecutionTracker.getCurrentTask()`. If the current task's path is in the node's `declaringTaskPaths` set, no warning. Otherwise — and gated by a per-(node, task) `naggedTaskPaths` set to dedupe multi-site warnings — the deprecation is emitted with contextual information about the offending task and configuration.

The walk-based approach is path-agnostic. As long as the dependency edge exists in the final work graph, the BFS finds it. This captures relationships that the per-node `TransformStepNodeDependencyResolver` path alone cannot, including:

- Chained transforms (a previous `TransformStepNode` reachable through the chain).
- `@InputArtifactDependencies`-fed transforms (one transform's upstream is another transform's output).
- `componentFilter`-restricted artifact views in dependency-substitution scenarios.
- Eager `configurations.foo.each { … }` resolution at configuration time, which poisons the `visitDependencies` path the resolver uses but leaves the work graph itself intact.

## What is NOT warned

- **External artifact transforms** (`TransformedExternalArtifactSet`). The nag path runs only from `TransformedProjectArtifactSet.visit()`. By design — the CC `project not found` failure that motivates this deprecation is specifically a project-artifact concern.
- **Gradle-internal code paths** wrapped in `SafeInternalArtifactTransformAccess.runWithSuppression(...)`. The script classpath resolver is wired to use this so plugin classpath resolution doesn't trigger spurious warnings. Plugins and user code do not enter this scope.
- **Tasks that themselves only `dependsOn` a declaring task without declaring directly**. If task B `dependsOn` task A and A declares the view, B does not inherit A's declaration; B's own undeclared query still warns. This is intentional per-task semantics.

## Configuration Cache behavior

`declaringTaskPaths` is serialized through the `AbstractTransformStepNodeCodec` so CC reloads preserve declarer attribution without needing to re-run the resolver. `naggedTaskPaths` is intentionally *not* serialized — it's UX dedup state, not correctness state, and a CC reload of an unfixed build correctly re-warns the user on first query.

The existing CC `project not found` failure under undeclared resolution is left unchanged (separately scoped); the deprecation correctly fires before the failure to surface the actionable cause.

## Test coverage

`UndeclaredDependencyResolutionIntegrationTest` (new) exercises 20 scenarios across:

- single-task undeclared (both task A and task B forms).
- task A undeclared, task B declares: with and without `dependsOn` linkage, across both invocation orders when both are requested on the command line.
- per-task semantics ("task B is still undeclared when triggered as a downstream of declaring task A").
- single-task multiple query sites (verifies dedup).
- task-graph-calculation-time `dependsOn { … view.files; [] }` queries.
- CC `project not found` characterization.
- `@InputArtifactDependencies` upstream chains.

Pre-existing transform tests that exercise legitimate undeclared usage (`ArtifactTransformBuildOperationIntegrationTest.no build operation for planned steps executed non-planned`) explicitly assert the deprecation via `expectUndeclaredArtifactTransformInputDeprecation()` with `when: "this test relies on undeclared artifact transforms"` to flag the intent for future maintainers.

## Key files

| File | Role |
|---|---|
| `TaskDeclarationAware.java` | New marker interface bridging the layering between `:core` and `:dependency-management` |
| `DefaultExecutionPlan.java` | New `propagateTaskDeclarationsToTransitiveSuccessors()` BFS run from `determineExecutionPlan()` |
| `transform/TransformStepNode.java` | Implements `TaskDeclarationAware`; per-node `declaringTaskPaths` and `naggedTaskPaths`; `nagAboutUndeclaredResolution()` |
| `TransformStepNodeDependencyResolver.java` | Null-guarded `markDeclaredBy` call from the existing resolver path (defense in depth alongside the post-graph walk) |
| `AbstractTransformStepNodeCodec.kt` | CC serialization of `declaringTaskPaths` |
| `SafeInternalArtifactTransformAccess.java` | Internal-code exemption scope; wired into `DefaultScriptClassPathResolver` |
| `UndeclaredArtifactTransformInputDeprecation.groovy` | Spock trait used by tests to expect the deprecation message |
| `upgrading_version_9.adoc` | User-facing migration guide with Kotlin DSL examples |


tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

lkasso (approved):
Docs LGTM.

tresat:
@bot-gradle test this

big-guy (approved):

PR #38196, opened 2026-06-12
Allow claiming abstract-getter properties with concrete forwarder setters

hegyibalint:
### Context

Back-port of the standalone class-generation commit from #37783 onto `master`, ahead of the Provider API migration (gradle/gradle#37342)

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
@@ -641,6 +641,10 @@ abstract class AbstractClassGenerator implements ClassGenerator {
         public Type getGenericReturnType() {
             return returnType;
         }
+
+        public boolean isAnnotationPresent(Class<? extends Annotation> annotationType) {
+            return method.isAnnotationPresent(annotationType);
+        }
     }
 
     protected static class PropertyMetadata {
@@ -1103,7 +1107,15 @@ abstract class AbstractClassGenerator implements ClassGenerator {
                     return false;
                 }
             }
+            boolean allAbstract = allGettersAreAbstract(property);
             for (Method setter : property.setters) {
+                // When every getter is abstract, the user cannot be managing a backing field
+                // themselves (an abstract getter has no body to read it from). Any concrete
+                // setter must then be a forwarder (e.g. setX(value) { getX().set(value); }),
+                // so it shouldn't prevent the property from being claimed and given a managed body.
+                if (allAbstract) {
+                    continue;
+                }
                 if (!Modifier.isAbstract(setter.getModifiers())) {
                     return false;
                 }
@@ -1122,6 +1134,18 @@ abstract class AbstractClassGenerator implements ClassGenerator {
                 // Read only but unrecognized type
                 return false;
             }
+        }
+
+        private boolean allGettersAreAbstract(PropertyMetadata property) {
+            if (property.getters.isEmpty()) {
+                return false;
+            }
+            for (MethodMetadata getter : property.getters) {
+                if (!getter.isAbstract()) {
+                    return false;
+                }
+            }
+            return true;
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


hegyibalint:
@bot-gradle test this

hegyibalint:
@bot-gradle test this

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
```diff
@@ -1103,7 +1107,15 @@ boolean claimPropertyImplementation(PropertyMetadata property) {
...
             }
+            boolean allAbstract = allGettersAreAbstract(property);
             for (Method setter : property.setters) {
+                // When every getter is abstract, the user cannot be managing a backing field
+                // themselves (an abstract getter has no body to read it from). Any concrete
+                // setter must then be a forwarder (e.g. setX(value) { getX().set(value); }),
+                // so it shouldn't prevent the property from being claimed and given a managed body.
+                if (allAbstract) {
+                    continue;
+                }
```
mlopatkin: ❓ If `allAbstract` is true, then all iterations become no-op. Can we hoist this check out of the loop?

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
```diff
@@ -1103,7 +1107,15 @@ boolean claimPropertyImplementation(PropertyMetadata property) {
                     return false;
                 }
             }
+            boolean allAbstract = allGettersAreAbstract(property);
```
mlopatkin: ❓ By this point we know that all getters are either abstract or bridge methods. Is presence of a bridge method important to negate the optimization below? AFAIK, it would just delegate to the original method.

mlopatkin (commented):
I have some questions about the logic in the generator

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
hegyibalint: Good point, see the new implementation that skips when `!hasAbstractGetter`

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
hegyibalint: The point is that we don't want to care about bridge methods; they only delegate.
What we want to catch is if there are any custom implementations of a getter, because it has effects, when we come to the setters, if we can claim.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
```diff
@@ -1097,19 +1101,32 @@ void visitProperty(PropertyMetadata property) {
...
+                        //  - The property has no getter at all (write-only, e.g. Task.setOnlyIf): a concrete getter would have returned above, an abstract one would have set the flag.
+                        //  - This setter is concrete, and with no getter to work through, it can only be managing its own state.
+                        // So the property is not ours to claim.
+                        return false;
+                    }
                 }
             }
 
-            // Property is readable and all getters and setters are abstract
+            // Property has no user-managed state
```
mlopatkin, line 1129: 🤔 So here we have either:
1. Property has only abstract getters (and maybe setters, who knows)
2. Property has no getters at all, and only abstract setters, if any.

I guess, without getters and setters, there is no property, so we can be sure at least one of those is present.

Now we assume, that if the getter is abstract, then the concrete setter only manages the state through the getter. It is probably correct for the PAPI migration use cases, but I'm not so sure about the general case. A setter can e.g. modify other fields.

I'm not sure what consequences a false positive here would bring, but this logic doesn't look robust.

❓ Can we provide a more precise way to detect "safe" getters? E.g. marking a legacy delegating setter with an annotation?

mlopatkin (commented):
Now that I understand the approach better, I have some further reservations.

asodja:
Will continue in https://github.com/gradle/gradle/pull/38829

Issue #38185, opened 2026-06-11
Unify Gradle mode testing annotation logic for CC and IP


alllex:
Our testing infrastructure provides annotations like `@ToBeFixedForConfigurationCache` and `@ToBeFixedForIsolatedProjects`. However, the capabilities provided by these annotations is not the same, and they are also not handled uniformly, making them harder to reason about.

The IP-specific annotations lack capabilities that are needed to annotate the remaining in compatible tests.

alllex:
Fixed in https://github.com/gradle/gradle/pull/38248

PR #38197, opened 2026-06-12
Bring IP testing annotations to parity with CC

alllex:
Part of #38185.

Gradle integration tests run in three modes: Vintage (baseline), Configuration Cache (CC), and Isolated Projects (IP). Per-mode expectations are expressed via annotations: `@ToBeFixedFor*` (expect failure, or skip with a reason) and `@UnsupportedWith*` (skip). CC has a richer surface than IP today — this PR closes the gap.

These changes intentionally introduce duplication to unblock using IP annotations. The unification and a refactor will follow later.

## What changes

**`Skip` enum → free-form `String skipBecause()`** on `@ToBeFixedForConfigurationCache` and `@ToBeFixedForIsolatedProjects`. Empty = expect failure (unchanged default), non-empty = skip with that reason. Placed after `because` so call sites read naturally: `(because = "…", skipBecause = "…")`. Of the enum's possible reasons only `FLAKY` was ever used in practice (~5 call sites); the rest are dropped along with the enum, and the `FLAKY` sites become `skipBecause = "flaky"`.

**IP gains CC's capability set:**
- `@ToBeFixedForIsolatedProjects` gains `bottomSpecs` and `iterationMatchers` — narrowing a base-class annotation to specific concrete subclasses / specific Spock iterations.
- New `@UnsupportedWithIsolatedProjects` (annotation + Spock extension + JUnit rule), modeled on `@UnsupportedWithConfigurationCache`. Previously IP could express "expect failure" but not "skip".
- IP Spock extension and JUnit rule aligned with their CC counterparts (spec- and feature-level dispatch, bottom-spec gating, iteration matching).
- `@ToBeFixedForConfigurationCache`'s `@Target` extended from `METHOD` to `{METHOD, TYPE}` so both families allow class-level placement.

The IP Spock extension / JUnit rule / `@UnsupportedWithIsolatedProjects` are intentional near-mechanical mirrors of their CC counterparts. The duplication is temporary — a unified dispatch layer will collapse it once both families have the same shape.

Additive at call sites — existing usages keep working.

### DIFF
--- a/platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/KotlinApplicationInitIntegrationTest.groovy
+++ b/platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/KotlinApplicationInitIntegrationTest.groovy
@@ -18,7 +18,6 @@ package org.gradle.buildinit.plugins
 
 import org.gradle.api.JavaVersion
 import org.gradle.buildinit.plugins.fixtures.ScriptDslFixture
-import org.gradle.integtests.fixtures.modes.ToBeFixedForIsolatedProjects
 import org.gradle.test.fixtures.file.LeaksFileHandles
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.JdkVersionTestPreconditions
@@ -199,7 +198,6 @@ class KotlinApplicationInitIntegrationTest extends AbstractJvmLibraryInitIntegra
     }
 
     @Requires(value = JdkVersionTestPreconditions.KotlinSupportedJdk.class)
-    @ToBeFixedForIsolatedProjects(because = "KGP modifies service parameter properties concurrently", skipBecause = "flaky")
     def "initializes Kotlin application with JUnit Jupiter test framework with --split-project"() {
         when:
         run('init', '--type', 'kotlin-application', '--test-framework', 'junit-jupiter', "--split-project", '--java-version', JavaVersion.current().majorVersion)
--- a/platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/KotlinApplicationInitIntegrationTest.groovy
+++ b/platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/KotlinApplicationInitIntegrationTest.groovy
@@ -26,7 +26,6 @@ import org.gradle.test.preconditions.JdkVersionTestPreconditions
 import spock.lang.Issue
 
 import static org.gradle.buildinit.plugins.internal.modifiers.BuildInitDsl.KOTLIN
-import static org.gradle.integtests.fixtures.modes.ToBeFixedForIsolatedProjects.Skip.FLAKY
 import static org.hamcrest.CoreMatchers.allOf
 import static org.hamcrest.CoreMatchers.containsString
 import static org.hamcrest.CoreMatchers.not
@@ -200,7 +199,7 @@ class KotlinApplicationInitIntegrationTest extends AbstractJvmLibraryInitIntegra
     }
 
     @Requires(value = JdkVersionTestPreconditions.KotlinSupportedJdk.class)
-    @ToBeFixedForIsolatedProjects(skip = FLAKY, because = "KGP modifies service parameter properties concurrently")
+    @ToBeFixedForIsolatedProjects(because = "KGP modifies service parameter properties concurrently", skipBecause = "flaky")
     def "initializes Kotlin application with JUnit Jupiter test framework with --split-project"() {
         when:
         run('init', '--type', 'kotlin-application', '--test-framework', 'junit-jupiter', "--split-project", '--java-version', JavaVersion.current().majorVersion)
--- a/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/modes/ToBeFixedForConfigurationCache.java
+++ b/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/modes/ToBeFixedForConfigurationCache.java
@@ -36,19 +36,20 @@ import java.lang.annotation.Target;
  * <p>
  * Instead of expecting failure, you can skip the test in case the test doesn't fail consistently
  * or has other undesirable effects, such as timeouts.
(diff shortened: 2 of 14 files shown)


alllex:

platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/KotlinApplicationInitIntegrationTest.groovy
```diff
@@ -200,7 +198,6 @@ class KotlinApplicationInitIntegrationTest extends AbstractJvmLibraryInitIntegra
     }
 
     @Requires(value = JdkVersionTestPreconditions.KotlinSupportedJdk.class)
-    @ToBeFixedForIsolatedProjects(skip = FLAKY, because = "KGP modifies service parameter properties concurrently")
```
alllex, line 203: ℹ️ These and other similar are dropped, because KGP should have fixed this issue. In my local runs and on CI this seems to pass, so let's include this in the testing unless the flakiness shows up again

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/modes/ToBeFixedForConfigurationCache.java
```diff
@@ -67,37 +68,4 @@
      * Defaults to an empty array, meaning this annotation applies to all iterations of the annotated feature.
      */
     String[] iterationMatchers() default {};
-
-    /**
-     * Reason for skipping a test with configuration cache.
-     */
-    enum Skip {
```
alllex, line 74: ℹ️  Apart from the default, only the FLAKY one was used, and only in a couple places. Since we don't really need to be strict with the skip reason, the string-based approach is cleaner imo

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/modes/ToBeFixedForConfigurationCacheExtension.groovy
```diff
@@ -19,38 +19,50 @@ package org.gradle.integtests.fixtures.modes
...
 
-import static org.gradle.integtests.fixtures.modes.ToBeFixedForConfigurationCache.Skip.DO_NOT_SKIP
-
 class ToBeFixedForConfigurationCacheExtension implements IAnnotationDrivenExtension<ToBeFixedForConfigurationCache> {
 
     private final ToBeFixedSpecInterceptor toBeFixedSpecInterceptor = new ToBeFixedSpecInterceptor("Configuration Cache")
 
     @Override
-    void visitFeatureAnnotation(ToBeFixedForConfigurationCache annotation, FeatureInfo feature) {
+    void visitSpecAnnotation(ToBeFixedForConfigurationCache annotation, SpecInfo spec) {
```
6hundreds, line 31: ❓ Can't `visitSpecAnnotation` and `visitFeatureAnnotation` share some implementation?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/modes/ToBeFixedForIsolatedProjectsExtension.groovy
```diff
@@ -19,35 +19,47 @@ package org.gradle.integtests.fixtures.modes
...
 
-import static org.gradle.integtests.fixtures.modes.ToBeFixedForIsolatedProjects.Skip.DO_NOT_SKIP
+import static ToBeFixedForConfigurationCacheExtension.isEnabledBottomSpec
 
 class ToBeFixedForIsolatedProjectsExtension implements IAnnotationDrivenExtension<ToBeFixedForIsolatedProjects> {
 
     private final ToBeFixedSpecInterceptor toBeFixedSpecInterceptor = new ToBeFixedSpecInterceptor("Isolated Projects")
 
     @Override
     void visitSpecAnnotation(ToBeFixedForIsolatedProjects annotation, SpecInfo spec) {
```
6hundreds, line 31: ❓ Can't `visitSpecAnnotation` and `visitFeatureAnnotation` share some implementation?

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/modes/UnsupportedWithIsolatedProjectsExtension.groovy
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    void visitFeatureAnnotation(UnsupportedWithIsolatedProjects annotation, FeatureInfo feature) {
+        if (GradleContextualExecuter.isIsolatedProjects()) {
+            if (isAllIterations(annotation.iterationMatchers()) && isEnabledBottomSpec(annotation.bottomSpecs(), { feature.parent.bottomSpec.name == it })) {
+                feature.skipped = true
+            } else {
+                feature.getFeatureMethod().addInterceptor(new IterationMatchingMethodInterceptor(annotation.iterationMatchers()))
+            }
+        }
+    }
```
6hundreds, line 55: ❓ Same question here, about sharing some bits, but I see that CC siblings of these methods don't share.

6hundreds (approved):
Functional changes are LGTM, I have some questions about logic duplication in annotation handlers, but I've noticed that CC siblings do the same. I'm okay with keeping this as is if sharing logic there is not straightforward.

testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/modes/ToBeFixedForConfigurationCacheExtension.groovy
alllex: They can, but it would be a bit tedious. I don't think it's worth doing it in the same PR, given that we want to unify this.

alllex:
I agree that duplication is not perfect and there are obvious and not so obvious ways to unify the implementations. This PR was created specifically to avoid dealing with the unification just yet. It's going to happen when we fully address the issue in the follow-ups:
- https://github.com/gradle/gradle/issues/38185

PR #38198, opened 2026-06-12
Improve error message when signing task has no signatory 

### LINKED ISSUE #19417: Improve message for Cannot perform signing task ':...' because it has no configured signatory

psoni674:
Fixes #19417

Context

The current error thrown when a signing task runs without a signatory — "Cannot perform signing task ':X' because it has no configured signatory" — gives users no indication of what a "signatory" is or how to resolve the problem. Users who hit it (often just by running a build that signs artifacts) are left with a dead end, and "no configured signatory" is a common, unresolved search online.

In practice the fix is simply to configure a signatory in the signing {} block, but the message never says so.

This change extends the message to explain that a signatory holds the PGP key used to sign artifacts and must be configured in the signing {} block, and adds a link to the Signing Plugin documentation. The original leading sentence is left unchanged, so the existing NoSigningCredentialsIntegrationSpec assertion (a substring match on the cause) remains compatible.

### DIFF
--- a/platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/GradleCoreProblemGroup.java
+++ b/platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/GradleCoreProblemGroup.java
@@ -30,6 +30,7 @@ public abstract class GradleCoreProblemGroup {
     private static final ProblemGroup CONFIGURATION_USAGE_PROBLEM_GROUP = ProblemGroup.create("configuration-usage", "Configuration usage");
     private static final DaemonToolchainProblemGroup DAEMON_TOOLCHAIN_PROBLEM_GROUP = new DefaultDaemonToolchainProblemGroup();
     private static final ProblemGroup SCRIPTS_PROBLEM_GROUP = ProblemGroup.create("scripts", "Scripts");
+    private static final DefaultPackagingProblemGroup PACKAGING_PROBLEM_GROUP = new DefaultPackagingProblemGroup();
 
     public static CompilationProblemGroup compilation() {
         return COMPILATION_PROBLEM_GROUP;
@@ -71,6 +72,10 @@ public abstract class GradleCoreProblemGroup {
         return SCRIPTS_PROBLEM_GROUP;
     }
 
+    public static PackagingProblemGroup packaging() {
+        return PACKAGING_PROBLEM_GROUP;
+    }
+
     public interface CompilationProblemGroup {
         ProblemGroup thisGroup();
         ProblemGroup java();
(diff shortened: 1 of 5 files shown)


alllex:

Please add tests to cover your changes.  

***

Documentation links must use the `DocumentationRegistry`, please look at the examples of its usage in the code.

psoni674:
> Thank you for your contribution!
> 
> Please add tests to cover your changes.
> 
> Documentation links must use the `DocumentationRegistry`, please look at the examples of its usage in the code.

Thanks @alllex ! I have addressed the review feedback and updated the implementation accordingly — the doc link now uses DocumentationRegistry, and I extended NoSigningCredentialsIntegrationSpec to assert the full message, including the documented link, using assertHasDocumentedCause.

alllex:
@bot-gradle test this

alllex:

big-guy:
@psoni674 This looks reasonable to me, so I think we can get to this soon. Instead of adding this message to the exception, we want to move to using the Problems API which will do something similar + capture this misconfiguration in a structured way. 

@octylFractal can you provide an example of how to use Problems API to do this?

octylFractal:
Providing a proper example proved a little bit difficult, I ended up just writing the full change since I think it works as an example the best. If you'd like to learn about how the Problems API should be used in Gradle itself, I recommend looking at the commit I pushed up. Note that I added a new group following our in-progress spec: https://docs.google.com/document/d/1uyAKVISEwhy-AZppQhMST-XankI-9gtIelxhaQttqRI/edit?usp=sharing

If you want to use Problems API in your own plugins, most of it still applies aside from the changes to `KnownProblemIds` and `GradleCoreProblemGroup`.

octylFractal (approved):

psoni674:
> Providing a proper example proved a little bit difficult, I ended up just writing the full change since I think it works as an example the best. If you'd like to learn about how the Problems API should be used in Gradle itself, I recommend looking at the commit I pushed up. Note that I added a new group following our in-progress spec: https://docs.google.com/document/d/1uyAKVISEwhy-AZppQhMST-XankI-9gtIelxhaQttqRI/edit?usp=sharing
> 
> If you want to use Problems API in your own plugins, most of it still applies aside from the changes to `KnownProblemIds` and `GradleCoreProblemGroup`.

Thanks @octylFractal for the detailed example. I'll go through the commit and the documentation.

eskatos (approved):

Issue #37352, opened 2026-03-30
Guard Project to Build mutable state access with IP


6hundreds:
We do wrapping https://github.com/gradle/gradle/blob/e0fbb224640e9fa71398fe351136091cf791c74c/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt#L225-L241 but don't protect mutable state access with problems reporting.

Some APIs doesn't make sense to be called on the project level, like `Gradle.settingsEvaluated`, we may also want to consider to prohibit such calls

PR #38200, opened 2026-06-12
Report Isolated Projects problem when a project plugin registers shared model defaults

### LINKED ISSUE #38028: Injectable services from build-scope should be safe with IP
Project plugins can inject a number of public services and use them for their operations. Some of these services live in the build scope or parent scopes, which means they become shared and potentially mutable resources. Access to these services can be unsafe under parallel project configuration that happens when Isolated Projects is enabled.

Examples of such services:
- `StartParameter`
- ... more tbd ...

The `Gradle` instance can also be injected as a service, but it plays a special role, as it carries a number of injectable services on itself. There is a dedicated issue to deal with it:
- https://github.com/gradle/gradle/issues/37352

Steps
- [x] Identify all public services that come from the build-scope or its parent scopes
- [x] Assess the safety of these services under IP
- [ ] Make the services safe or forbid their usage under IP

reinsch82:
Part of #38028.

### What

`SharedModelDefaults` is a build-scoped public service that is conceptually settings-only (the `defaults { }` block), but it is injectable into project plugins. Calling `add()` from a project today fails with a late `IllegalStateException` ("Cannot add shared model defaults after processing.") that gives the plugin author no hint about what they did wrong, and under Isolated Projects diagnostics mode it hard-stops configuration instead of being collected as a problem.

With this change, when Isolated Projects is enabled, project scope resolves a problem-reporting view of the service:

```
> Configure project :
Problem found: Project ':' cannot register shared model defaults. Shared model defaults
can only be registered in a settings script, using the 'defaults' block.
    Location: build file 'build.gradle' line 10
```

- **FAIL_FAST** (default IP): the problem is thrown immediately with build script location.
- **DIAGNOSTICS**: the problem is recorded in the configuration cache report and configuration continues; the registration is skipped (it could never take effect, since registrations are processed at the end of settings evaluation).
- **Without Isolated Projects**: the decorator is not installed; projects resolve the build-scoped service directly and behavior is unchanged (verified by test).

### How

- `ProjectScopeServices.decorateSharedModelDefaults()` shadows the build-scoped service for project-scope lookups under IP (same pattern as the tooling model builder registry and the build services registry IP guard).
- `ProblemReportingSharedModelDefaults` reports via `IsolatedProjectsProblemsReporter` on `add()` and on the Groovy dynamic-method registration path, and delegates the internal lifecycle (`getLayout`, `withProjectLayout`, `processRegistrations`) untouched — the declarative DSL machinery resolves the service at project scope and relies on these.
- The `@ServiceScope` of `SharedModelDefaults` is extended to `{Build, Project}` to permit the project-scope registration. Annotation-value changes are not binary-incompatible.

This also removes a dependence on a theoretical data race: the existing protection relies on a non-volatile `processed` flag read from parallel project configuration threads, while the new view never touches the shared mutable state.

### Test coverage

`IsolatedProjectsSharedModelDefaultsIntegrationTest`:
- violation reported in both DIAGNOSTICS and FAIL_FAST modes
- non-IP baseline asserting the pre-existing `IllegalStateException` is preserved

Regression: `ProjectTypeModelDefaultsIntegrationTest` (62 tests) passes, covering legitimate settings-side `defaults { }` usage.

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsSharedModelDefaultsIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsSharedModelDefaultsIntegrationTest.groovy
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl.isolated
+
+import spock.lang.Issue
+
+class IsolatedProjectsSharedModelDefaultsIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
+
+    @Issue("https://github.com/gradle/gradle/issues/38028")
+    def "reports problem when a project plugin registers shared model defaults"() {
+        given:
+        buildFile """
+            import org.gradle.api.initialization.SharedModelDefaults
+            import javax.inject.Inject
+
+            abstract class DefaultsRegisteringPlugin implements Plugin<Project> {
+                @Inject
+                abstract SharedModelDefaults getSharedModelDefaults()
+
+                void apply(Project project) {
+                    sharedModelDefaults.add("myProjectType", String, {})
+                }
+            }
+            apply plugin: DefaultsRegisteringPlugin
+        """
+
+        when:
+        isolatedProjectsFailsUsing(mode, "help")
+
+        then:
+        fixture.assertIsolatedProjectsProblems(mode) {
+            projectsConfigured(":")
+            problem("Build file 'build.gradle': line 10: Project ':' cannot register shared model defaults. Shared model defaults can only be registered in a settings script, using the 'defaults' block.", 1)
+        }
+
+        where:
+        mode << ALL_MODES
+    }
+
+    def "without isolated projects, a project plugin registering shared model defaults fails with the late IllegalStateException"() {
+        given:
+        buildFile """
+            import org.gradle.api.initialization.SharedModelDefaults
+            import javax.inject.Inject
+
+            abstract class DefaultsRegisteringPlugin implements Plugin<Project> {
+                @Inject
+                abstract SharedModelDefaults getSharedModelDefaults()
+
+                void apply(Project project) {
+                    sharedModelDefaults.add("myProjectType", String, {})
+                }
+            }
+            apply plugin: DefaultsRegisteringPlugin
+        """
+
+        when:
+        fails("help")
+
+        then:
+        failure.assertHasCause("Cannot add shared model defaults after processing.")
+    }
+}
--- a/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
+++ b/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
@@ -36,6 +36,7 @@ import java.util.Collections;
 import java.util.Formatter;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
 import java.util.TreeSet;
@@ -65,7 +66,7 @@ import static org.gradle.util.internal.CollectionUtils.join;
  * be registered as a listener of that type. Alternatively, service implementations can be annotated with {@link org.gradle.internal.service.scopes.ListenerService} to indicate that the should be
  * registered as a listener.</p>
  */
-public class DefaultServiceRegistry extends AbstractServiceRegistry implements CloseableServiceRegistry {
+public class DefaultServiceRegistry extends AbstractServiceRegistry implements CloseableServiceRegistry, ServiceRegistryIntrospection {
     private enum State {INIT, STARTED, CLOSED}
 
     private final static ServiceRegistry[] NO_PARENTS = new ServiceRegistry[0];
@@ -137,6 +138,15 @@ public class DefaultServiceRegistry extends AbstractServiceRegistry implements C
     @Override
     ServiceProvider asServiceProvider() {
         return thisAsServiceProvider;
+    }
+
+    @Override
+    public Set<Class<?>> getAllServiceTypes() {
+        // thisAsServiceProvider composes this registry's own services with the parent chain,
+        // so this walks every reachable scope. Reads type metadata only; no service is realized.
+        Set<Class<?>> types = new HashSet<Class<?>>();
+        thisAsServiceProvider.collectServiceTypes(types);
+        return types;
     }
 
     private static ServiceProvider toParentServices(ServiceRegistry serviceRegistry) {
@@ -461,6 +471,17 @@ public class DefaultServiceRegistry extends AbstractServiceRegistry implements C
                 visitor = serviceProvider.getAll(serviceType, token, visitor);
             }
             return visitor;
+        }
+
+        @Override
+        public void collectServiceTypes(Set<Class<?>> dest) {
+            // providersByType is keyed by every type each registered service can be resolved as
+            // (declared types plus their supertypes/interfaces). Read the current snapshot once.
+            @SuppressWarnings("NullAway") // TODO(https://github.com/uber/NullAway/issues/681) Can't infer that AtomicReference holds non-nullable type
+            PersistentMap<Class<?>, PersistentArray<ServiceProvider>> providersByType = services.get().providersByType;
+            for (Map.Entry<Class<?>, PersistentArray<ServiceProvider>> entry : providersByType) {
+                dest.add(entry.getKey());
+            }
         }
 
         @Override
@@ -1076,6 +1097,13 @@ public class DefaultServiceRegistry extends AbstractServiceRegistry implements C
         }
 
         @Override
+        public void collectServiceTypes(Set<Class<?>> dest) {
+            for (ServiceProvider serviceProvider : serviceProviders) {
+                serviceProvider.collectServiceTypes(dest);
+            }
+        }
+
+        @Override
         public void stop() {
             try {
                 CompositeStoppable.stoppable(Arrays.asList(serviceProviders)).stop();
@@ -1108,6 +1136,11 @@ public class DefaultServiceRegistry extends AbstractServiceRegistry implements C
         @Override
         public Visitor getAll(Class<?> serviceType, @Nullable ServiceAccessToken token, Visitor visitor) {
             return parent.getAll(serviceType, token, visitor);
+        }
+
+        @Override
+        public void collectServiceTypes(Set<Class<?>> dest) {
+            parent.collectServiceTypes(dest);
         }
 
         @Override
--- a/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
+++ b/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/DefaultServiceRegistry.java
@@ -149,6 +149,15 @@ public class DefaultServiceRegistry extends AbstractServiceRegistry implements C
         return types;
     }
 
+    @Override
+    public Set<Class<?>> getOwnServiceTypes() {
+        // Only this registry's own services - no parent traversal. For a project-scoped registry this
+        // is exactly the project-scoped services. Reads type metadata only; no service is realized.
+        Set<Class<?>> types = new HashSet<Class<?>>();
+        ownServices.collectServiceTypes(types);
+        return types;
+    }
+
     private static ServiceProvider toParentServices(ServiceRegistry serviceRegistry) {
         if (serviceRegistry instanceof AbstractServiceRegistry) {
             return new ParentServices(((AbstractServiceRegistry) serviceRegistry).asServiceProvider());
--- a/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/ServiceProvider.java
+++ b/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/ServiceProvider.java
@@ -20,6 +20,7 @@ import org.gradle.internal.concurrent.Stoppable;
 import org.jspecify.annotations.Nullable;
 
 import java.lang.reflect.Type;
+import java.util.Set;
 
 /**
  * Provides a set of zero or more services. The get-methods may be called concurrently. {@link #stop()} is guaranteed to be only called once,
@@ -38,6 +39,15 @@ interface ServiceProvider extends Stoppable {
      * @return A visitor that should be used for all subsequent services.
      */
     Visitor getAll(Class<?> serviceType, @Nullable ServiceAccessToken token, Visitor visitor);
+
+    /**
+     * Adds to {@code dest} every type by which a service from this provider can be resolved, for introspection
+     * (see {@link ServiceRegistryIntrospection}). Implemented by the structural providers that hold the registry
+     * contents and the parent chain; per-service providers contribute their types via the owning registry's
+     * type index, so the default is a no-op. Must not realize any service.
+     */
+    default void collectServiceTypes(Set<Class<?>> dest) {
+    }
 
     interface Visitor {
         void visit(Service service);
--- a/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/ServiceRegistryIntrospection.java
+++ b/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/ServiceRegistryIntrospection.java
@@ -39,4 +39,15 @@ public interface ServiceRegistryIntrospection {
      * Only type metadata is read; no service instances are realized.
      */
     Set<Class<?>> getAllServiceTypes();
+
+    /**
+     * Returns the types provided by <em>this</em> registry's own services only, excluding anything inherited
+     * from ancestor registries. For the per-project registry returned by {@code ProjectInternal.getServices()},
+     * this is the set of <em>project-scoped</em> services (including services declared at multiple scopes that
+     * have a project-scoped registration, e.g. {@code {Build, Project}} ones); {@link #getAllServiceTypes()}
+     * minus this set is what is inherited from build/global scope.
+     * <p>
+     * Only type metadata is read; no service instances are realized.
+     */
+    Set<Class<?>> getOwnServiceTypes();
 }
--- a/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/ServiceRegistryIntrospection.java
+++ b/platforms/core-runtime/service-registry-impl/src/main/java/org/gradle/internal/service/ServiceRegistryIntrospection.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.service;
+
+import java.util.Set;
+
+/**
+ * Introspection over a {@link ServiceRegistry}, allowing the set of resolvable service types to be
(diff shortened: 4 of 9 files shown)


reinsch82:
# Injectable services from build scope and above: IP safety findings

Working notes for [#38028](https://github.com/gradle/gradle/issues/38028): project plugins can inject
services that live in build scope or parent scopes. Under Isolated Projects (IP), projects configure in
parallel, so these services become shared resources across concurrently-configuring projects. This
document records which services were assessed, the verdict, and the reasoning.

`Gradle` itself is out of scope here — tracked separately in
[#37352](https://github.com/gradle/gradle/issues/37352) (wrapped by `CrossProjectConfigurationReportingGradle`,
mutable-state problem reporting still missing).

## What "IP-safe" means here (the criterion)

**Thread-safety is NOT the criterion.** A fully synchronized service can still be IP-unsafe, and an
unsynchronized one can be IP-safe. What Isolated Projects requires is that **configuring one project
cannot observe or perturb shared (build+-scope) state that another project's configuration depends on** —
so that each project's configuration is reproducible from its own inputs and independent of which other
projects were configured (or served from cache).

Operational rule (sufficient condition): a build+-scope service is safe to inject into project
configuration if its shared mutable state is **frozen at the settings-evaluation boundary** and read-only
during project configuration. This is the `preventFromFurtherMutation()` / `processRegistrations()`
pattern — `JvmToolchainManagement` and `SharedModelDefaults` are the model.

Two cases technically "change state" during configuration but remain safe because they don't feed back
into any project's configuration:
- **Write-only side-band sinks** not read back into config — problem reporting, logging.
- **Mutation confined to the project's own scope** — e.g. writing to a per-project child registry, not
  shared state.

And one read-time case is safe: **memoized deterministic reads** of build-level inputs (e.g. value
sources, `GradleProperties` immutable snapshot) — the result is a pure function of fixed inputs and is
recorded in the configuration-cache fingerprint.

So "safe" verdicts below are justified by *this* criterion (frozen / per-project / write-only / immutable
read), not by synchronization. Where a service lets project configuration mutate shared state that isn't
frozen and isn't a write-only sink, it is a concern regardless of how thread-safe that mutation is.

## How injection actually works (and why the inventory below is only a slice)

Plain task / plugin / extension / `ObjectFactory.newInstance` injection is **unrestricted**.
`DependencyInjectingInstantiator.addServicesToParameters()`
(`platforms/core-configuration/model-core/.../generator/DependencyInjectingInstantiator.java:180`)
resolves each `@Inject` parameter with a bare `services.find(serviceType)` against the project's full
`ServiceRegistry`, which chains through **every** ancestor scope (Project → Build → BuildTree →
BuildSession → CrossBuildSession → UserHome → Global) via `CompositeServiceProvider`
(`DefaultServiceRegistry`). There is **no filtering at injection time**: `@ServiceScope` is validated
only at *registration* time by `ServiceScopeValidator`, and plays no role in deciding what user code may
inject.

Consequences for this investigation:
- The truly injectable surface is *every service resolvable from the project registry*, not just the
  annotated public-API types in the table below. The documented contract in `service_injection.adoc`
  ("avoid injecting types not listed") is advisory, not enforced.
- Services registered under an **internal type with a public supertype** are injectable via the public
  type even with no `@ServiceScope` — this is exactly how `StartParameter` (registered as
  `StartParameterInternal`) hid from the annotation sweep. Our `git grep @ServiceScope` inventory is
  therefore a *lower bound*; instance/internal-typed registrations can add more.

**Existing precedent for restricting injection** (directly relevant to the issue's "forbid usage"
option — the mechanism already exists, twice):
- Worker actions: `IsolationScheme.ServicesForIsolatedObject.find()`
  (`model-core/.../isolated/IsolationScheme.java:122-168`) hard-whitelists ~8 services
  (`ExecOperations`, `FileSystemOperations`, `ArchiveOperations`, `ObjectFactory`, `ProviderFactory`,
  `BuildServiceRegistry`, `ProblemsInternal`, `ManagedObjectRegistry`) + caller-supplied extras;
  everything else throws `UnknownServiceException`.
- Project-feature apply actions: `DefaultProjectFeatureApplicator.SAFE_APPLY_ACTION_SERVICES`
  (`plugin-use/.../binding/DefaultProjectFeatureApplicator.java:384-391`) whitelists
  `ObjectFactory`, `ProviderFactory`, `DependencyFactory`, `TaskRegistrar`, `ProjectFeatureLayout`,
  `ConfigurationRegistrar`. This is effectively Gradle's *already-curated* answer to "what is safe to
  inject during isolated configuration" — a strong starting point for any project-scope allowlist.

## Inventory

Every *public API* type annotated `@ServiceScope` at `Build` scope or above (so reachable from a
project's service registry). NB: per the section above this is a lower bound, not the full injectable
surface:

| Service | Scope | Verdict |
|---|---|---|
| `org.gradle.api.invocation.Gradle` | Build | see #37352 |
| `org.gradle.StartParameter` | BuildSession (instance registration, no `@ServiceScope`) | **Unsafe — handled on separate branch** |
| `org.gradle.api.initialization.SharedModelDefaults` | Build | **Guarded — PR [#38200](https://github.com/gradle/gradle/pull/38200)** |
| `org.gradle.api.provider.ProviderFactory` | Build | Safe |
| `org.gradle.api.flow.FlowScope` / `FlowProviders` | Build | Safe |
| `org.gradle.tooling.provider.model.ToolingModelBuilderRegistry` | Build | Safe — project config writes only its own per-project child (see below) |
| `org.gradle.jvm.toolchain.JvmToolchainManagement` | Build | Already guarded (one small gap) |
| `org.gradle.api.problems.Problems` | BuildTree | Safe |
| `org.gradle.api.tasks.testing.TestEventReporterFactory` | Build | Safe |
| `org.gradle.api.artifacts.dsl.DependencyConstraintFactory` | Build | Safe |
| `org.gradle.api.invocation.BuildInvocationDetails` | Build | Safe |
| `org.gradle.api.configuration.BuildFeatures` | BuildTree | Safe |
| `org.gradle.api.component.SoftwareComponentFactory` | Global | Safe |
| `org.gradle.api.model.ObjectFactory` | Global/BuildTree/Project | Safe |
| `org.gradle.api.file.FileSystemOperations`, `ArchiveOperations`, `org.gradle.process.ExecOperations` | {Build, Project} | Safe |
| `org.gradle.api.artifacts.dsl.DependencyFactory` | {Build, Project} (on `DependencyFactoryInternal`) | Safe — scope declared on the registered internal type |
| `org.gradle.api.file.BuildLayout` | Settings | Not reachable from project scope |

## No threat, and why

### ProviderFactory (Build) — full collaborator audit (worked example)
`DefaultProviderFactory` is one shared build-scoped instance (`BuildScopeServices.java:459`). All 19
public methods were traced to their return types and delegates. Against the IP criterion (no
project-config mutation of non-frozen shared state): every method is either a pure factory, a memoized
deterministic read of build-level inputs, or a write that doesn't feed back into configuration. Four
build-scoped collaborators are reachable, none of which lets project configuration mutate shared
config-state:

1. **Pure provider construction** — `provider(Callable)` → `new DefaultProvider<>` (wraps user callable);
   `zip(a,b,fn)` → `a.zip(b,fn)` (delegates to the provider). No factory state touched.
2. **Value-source path** — `of(...)`, `environmentVariable*`, `systemProperty*`,
   `environmentVariablesPrefixedBy*`, `systemPropertiesPrefixedBy*`, `fileContents`, `exec`, `javaexec`
   all funnel through `ValueSourceProviderFactory.createProviderOf` / `ProcessOutputProviderFactory`
   (both Build scope, all `final` fields). `createProviderOf` mutates **no** shared field — it builds a
   fresh parameters object and returns a fresh `ValueSourceProvider`. Value computation is per-provider
   `CalculatedValue` (compute-once, thread-safe) with an `AtomicReference` exactly-once handoff
   (`DefaultValueSourceProviderFactory.java:268-311`). The only shared write at obtain-time is
   `valueListener.valueObtained(...)` → `ConfigurationCacheFingerprintWriter` (per-thread input-tracking
   state + synchronized sinks; already invoked concurrently from parallel task execution today).
   `fileContents`/`exec` return per-call objects (anon `FileContents`, `DefaultExecOutput`) with no state.
3. **Gradle properties path** — `gradleProperty*`, `gradlePropertiesPrefixedBy*` read the build-scoped
   `GradleProperties`, whose impl `DefaultGradleProperties` holds an **`ImmutableMap`**
   (`DefaultGradleProperties.java:27`); `find`/`getPropertiesWithPrefix` are pure reads. Each call returns
   a fresh `memoizing`/`changing`/`BiProvider` (per-project instance). Semantic note: all projects see the
   same build-level immutable snapshot — consistent under IP, no per-project mutable state.
4. **Credentials path** — `CredentialsProviderFactory` (one per build, registered as a
   `TaskExecutionGraphListener`). Caches are `ConcurrentHashMap` populated via `computeIfAbsent`
   (`provide()` line 66 — thread-safe lazy creation; same identity across projects returns the same
   read-only provider), `call()` is `synchronized`, and `missingProviderErrors` is a concurrent set
   drained at `graphPopulated` (post-configuration).

**Verdict: SAFE** (by the IP criterion): the value-source and credentials caches are memoizations of
deterministic functions of fixed inputs (recorded by the CC fingerprint), not config-state a project can
change for another's benefit; `GradleProperties` is an immutable build-level snapshot; the rest are pure
factories. No project configuration mutates non-frozen shared state. (The `ConcurrentHashMap`/`synchronized`
details are incidental robustness, not the basis for the verdict.)

### Problems (BuildTree) — fully traced
Single method `getReporter()`; each call returns a fresh `DefaultProblemReporter`. All reporters share
one `ProblemSummarizer` → `SummarizerStrategy`.

**Verdict: SAFE** by the IP criterion: problem reporting is a **write-only side-band sink** — a reported
problem is aggregated for the problems report, never read back into any project's configuration, so it
cannot couple projects or affect a cached configuration result. (Worth noting only because two separate
subagents mis-flagged a `ProblemSummaryInfo.count` data race here; that was both the wrong axis *and*
factually wrong — `count++` and all count reads are confined to the two `synchronized` methods of
`SummarizerStrategy`, `SummarizerStrategy.java:44-50,37`, with `ProblemSummaryInfo`/`seenProblemsWithCounts`
referenced nowhere else. Do not re-raise.)

### FlowScope / FlowProviders (Build) — the one to scrutinize under the criterion
`FlowProviders.getBuildWorkResult()` is safe: it returns a single provider whose value is written once
post-configuration (`buildFinished`) and read only when flow actions run — projects only read it, never
mutate it.

`FlowScope.always()` is the interesting case. It **mutates shared build-scoped state during project
configuration** — it appends a registration to `BuildFlowScope`'s pending-actions list. That is precisely
"changing state after settings evaluation," so the synchronization on that list (`BuildFlowScope.kt:70-74`)
is *not* what makes it safe and isn't the question. It is *not* classic cross-project coupling because the
list is a **write-only sink of deferred actions** — no project reads another project's registrations during
configuration. The real IP question is whether each project's registrations are correctly attributed to
*that project's* configuration-cache fragment (so invalidating/!configuring one project keeps its flow
actions correct). **Verdict: SAFE for the data-race lens, but the per-project CC attribution under IP needs
a verification test** — this is the genuine open item, not thread-safety.

### ToolingModelBuilderRegistry (Build) — IP-SAFE (earlier "downgrade" was the wrong axis)
Two methods: `register()`, `getBuilder()`. Against the IP criterion:
- **Project configuration mutates only its own scope.** `ProjectScopeServices.java:199` gives each project
  its own child registry via `createChild()`, and a project plugin's `register()` appends to the *child's*
  list — never to shared state another project reads. So project configuration cannot perturb another
  project through this service. **IP-SAFE.**
- The build-scoped parent list is populated by `BuildScopeToolingModelBuilderRegistryAction`s at build
  setup; reads (`find()` → `parent.find()`) happen at tooling-API model-build / task-execution time,
  post-configuration (`getBuilder(...).buildAll(...)` — confirmed in-ecosystem: AGP Lint, IntelliJ init
  script, vscode-gradle, etc.). Returned builders wrap immutable builder refs.

The unsynchronized parent `ArrayList` (`DefaultToolingModelBuilderRegistry.java:149-167`) is a **generic
thread-safety nit, not an IP concern** — I previously conflated the two and over-flagged it. It is not on
the parallel-configuration path. If anyone wants to close it for unrelated robustness, `CopyOnWriteArrayList`
is the one-line fix, but it is **not** part of the IP story. **Verdict: IP-SAFE.**

### TestEventReporterFactory (Build) — fully traced
Stateless factory (all fields `final` service refs). Each `createTestEventReporter()` call builds a fully
independent object tree — own anonymous `ListenerBroadcast`, fresh `GenericHtmlTestReportGenerator`, fresh
`SerializableTestResultStore.Writer`, fresh `DefaultRootTestEventReporter`. Per-reporter state
(`failureMessage`, child set, `TestResultState` counters which are `AtomicLong`) is never shared across
reporters. Concurrent calls from parallel projects contend on nothing. **Verdict: SAFE.**

### Stateless factories and read-only values — fully traced
All confirmed by reading impls (all fields `final`, methods create fresh objects or read immutable state):
- `DependencyConstraintFactory` — 4 `create(...)` methods, fresh `DependencyConstraint` each. SAFE.
- `SoftwareComponentFactory` — `adhoc(name)` → fresh component via `objectFactory.newInstance`. SAFE.
- `BuildFeatures` (BuildTree) — `getConfigurationCache`/`getIsolatedProjects` return `Lazy`-cached
  immutable `DefaultBuildFeature` (Provider refs only). SAFE.
- `BuildInvocationDetails` — `getBuildStartedTime()` is a `long` read. Backing `BuildStartedTime` is
  BuildSession-scoped and mutable (`volatile long`, `reset()`), but set once at build init before any
  project configures. SAFE.
- `ObjectFactory` (Global/BuildTree/Project) — **a project plugin gets the Project-scoped instance**
  (`ObjectFactoryFactory` is `@ServiceScope(Scope.Project.class)`), so per-project, not shared. The impl
  is all-`final`-field stateless delegation across its ~17 methods. SAFE.
- `org.gradle.initialization.layout.BuildLayout` (internal, but elasticsearch injects it): immutable
  value holder. SAFE.

### DependencyFactory — Safe (earlier "annotation gap" was wrong)
The public `DependencyFactory` interface has no `@ServiceScope`, but the **registered** type does:
`DependencyFactoryInternal extends DependencyFactory` is annotated `@ServiceScope({Scope.Build.class,
Scope.Project.class})` (`DependencyFactoryInternal.java:29`), and both `@Provides` methods declare that
internal type (`DependencyManagementBuildScopeServices.java:189`,
`DependencyManagementProjectScopeServices.java:55`). So:
- Registration passes strict scope validation (`ProjectScopeServices` uses `.scopeStrictly(...)`;
  `ServiceScopeValidator` only requires the annotation on the *declared* type, which is the annotated
  internal one — declaring the bare public type would throw "missing scope").
- A project injecting `DependencyFactory` resolves the project-scoped instance by assignable-type lookup
  (`@ServiceScope` plays no role at resolution). The 8 factory methods are stateless.
The public interface lacking the annotation is **cosmetic only** (siblings annotate the public type; here
it's on the internal subtype) — no functional gap, no action required. Earlier NEEDS-ATTENTION was a
mis-call inherited from a sub-agent.

### Dual-scoped operations services
`FileSystemOperations`, `ArchiveOperations`, `ExecOperations` are registered at both Build and Project
scope — a project plugin gets the project-scoped instance. **Verified by reading the implementations**
(not just inferred from scope):
- `DefaultFileSystemOperations` / `DefaultArchiveOperations` / `DefaultExecOperations` are thin facades
  holding only `final` collaborators (`FileOperations`, `Instantiator`, `ProcessOperations`); they
  delegate everything.
- `DefaultFileOperations` (the shared `FileOperations` impl) is itself stateless beyond `final`
  collaborators — every operation method (`copy`, `delete`, `zipTree`, `tarTree`, `copySpec`…) builds
  fresh per-call objects rather than mutating instance fields.
- The one genuinely shared, build-scoped, *stateful* collaborator reachable through the API —
  `DefaultDecompressionCoordinator` (used by `zipTree`/`tarTree`) — holds no configuration state: it is a
  filesystem decompression cache whose entries are a deterministic function of the archive (not anything
  a project's configuration sets). A project can't perturb another project's configuration through it.
  (It is also internally concurrency-safe — file lock + per-dir `ProducerGuard`,
  `DefaultDecompressionCoordinator.java:50-61` — but that robustness is incidental, not the IP basis.)

## Potential threat, and why

### StartParameter (BuildSession) — the issue's named example
Mutable: **40 public setters**, no `@ServiceScope` on the type (nor on `StartParameterInternal`;
registered as an instance in `BuildSessionScopeServices.java:58`). One instance shared across the whole
build session: a project plugin calling e.g. `setTaskNames(...)` during parallel configuration is a data
race and a cross-project visibility leak; reads become order-dependent. **Being handled on a separate
branch.**

### SharedModelDefaults (Build) — guarded in PR #38200
Semantically settings-only (the `defaults { }` block), but injectable into project plugins. Internals:
unsynchronized `ArrayList` + non-volatile `processed` flag (`DefaultSharedModelDefaults.java:40-41`).
Verified (code trace + test): `processRegistrations()` runs in the settings-processor chain
(`BuildScopeServices.java:633`) strictly before project configuration, so project-side `add()` was
already *functionally* prevented by a late `IllegalStateException` — but that protection rested on a
non-volatile flag read from parallel threads (theoretical stale-read → list corruption), and the ISE
gave no location and hard-stopped IP diagnostics mode. The fix installs a project-scoped
problem-reporting view under IP (no wrapping when IP is off).

### JvmToolchainManagement (Build) — the model of the criterion (frozen at settings eval), one gap
This is the canonical "IP-safe because frozen at settings evaluation" case.
`DefaultSettings.preventFromFurtherMutation()` (`DefaultSettings.java:394`) finalizes it right after
settings evaluation, before any project configures; later mutation throws
`InvalidUserCodeException("Mutation of toolchain repositories declared in settings is only allowed during settings evaluation")`.
Finalization also eagerly realizes the repository list, so the lazy realize path never runs concurrently.
**Gap:** `DefaultJavaToolchainResolverRegistry.register()` (lines 73-80) mutates a plain `HashMap` with
no finalization guard and no synchronization — normally called by settings-applied resolver plugins, but
nothing prevents a late/parallel call. One-line `assertMutable()`-style guard would close it.
Could also be aligned with IP problem reporting for consistency.

## Methodology limitation — transitive collaborator state

Several "Safe" verdicts below were originally reached from *scope* alone ("project gets a project-scoped
instance, so no sharing"). That reasoning is insufficient: a service can be a **stateless facade over
shared mutable state**. The real attack surface is the *transitive graph* reachable through a service's
API — the collaborators it delegates to, and the objects its methods return — not just the fields of the
service class. A service object can be immutable while handing a project plugin a live, build-scoped,
mutable collaborator.

What has actually been done per service varies and should not be overstated:
- **Fully traced** (every public method's return type + all shared collaborators audited):
  `ProviderFactory` (worked example — 4 collaborators), `FlowScope`/`FlowProviders`,
  `ToolingModelBuilderRegistry`, `Problems`, `TestEventReporterFactory`, `DependencyConstraintFactory`,
  `BuildInvocationDetails`, `BuildFeatures`, `SoftwareComponentFactory`, `ObjectFactory`,
  `DependencyFactory`, and the `FileSystemOperations`/`ArchiveOperations`/`ExecOperations` family
  (incl. `DecompressionCoordinator`).
- **Implementation-verified** (service impl + key shared collaborators read): `SharedModelDefaults`,
  `JvmToolchainManagement`.
- **Outcome of the full pass** (re-judged against the IP criterion, not thread-safety): all services are
  IP-SAFE *except* `StartParameter` (unsafe, mutable & not frozen — handled #38152) and
  `SharedModelDefaults` (was effectively frozen but now explicitly guarded — #38200). The single genuine
  open item is `FlowScope.always()`, which mutates shared build state during configuration as a write-only
  deferred-action sink — needs a per-project CC-attribution verification test, not a thread-safety fix.
  Corrections made along the way: the `ToolingModelBuilderRegistry` "unsynchronized parent list" was a
  thread-safety nit wrongly elevated to an IP concern (project config only writes its own child → IP-safe);
  the `Problems` `count++` "race" was refuted twice and is anyway the wrong axis (write-only sink); the
  `DependencyFactory` "annotation gap" was withdrawn (scope is on `DependencyFactoryInternal`).

For rigor a follow-up would trace, per allowed service, every public method's return type and delegate
for build/global-scoped mutable state — or sidestep the whole graph by switching to the allowlist
approach (only a curated, individually-audited set is injectable; see below).

## Out of scope / context

- **Non-public-API build-scoped services** (in non-`internal` packages like `org.gradle.execution.*`,
  `org.gradle.initialization.*`): ~30 types, all technically injectable. GitHub code search
  (import-statement queries, vendored Gradle source copies filtered out) shows **near-zero real-world
  usage** — calibration: public `FlowScope` had 56 genuine hits per 100 sampled (JetBrains/kotlin etc.),
  while the non-public types had 0–5, almost all junk/dataset repos. Exceptions: `BuildLayout`
  (elasticsearch build-tools; immutable, safe), `DependenciesAccessors` (version-catalog workaround
  snippet in build-logic build scripts + IntelliJ tooling — not project-configuration-time),
  `TaskExecutionGraphInternal`/`TaskNodeFactory` (single digits; remal-gradle-plugins/toolkit).
  → Deny-by-default for non-documented services would have near-zero ecosystem blast radius.
- **ProfileService** (Build): zero-method `@ListenerService` marker — nothing to call when injected; not
  public API; zero GitHub usage. The state behind it (`BuildProfile`, BuildTree scope, plain
  `LinkedHashMap`s, multiple independent listener broadcasters writing) has a plausible pre-existing race
  under `--profile`, but the attack surface is listener dispatch, not injection — belongs to the IP
  listener-safety workstream.
- There is **no runtime enforcement** of the public/internal API boundary at the injection point (see
  "How injection actually works"), and no annotation meaning "safe for parallel project configuration"
  (`@StatefulListener`/`@ParallelListener` exist for listeners only). Users cannot reliably recognize
  non-`internal`-package internal types. Any "forbid" outcome of #38028 must therefore be mechanical at
  service lookup/injection — and the mechanism already exists in two places (`IsolationScheme` for
  workers, `SAFE_APPLY_ACTION_SERVICES` for project features). The cleanest path is likely a project-scope
  injection allowlist seeded from the project-feature safe set + the documented `service_injection.adoc`
  services, with everything else either forbidden or routed through an IP problem report.

PR #38203, opened 2026-06-13
Enable remaining IP integ tests

alllex:
Part of https://github.com/gradle/gradle/issues/34266

Integration tests in all projects are now running with IP executor as well. The only exception for now is `configuration-cache` project where tests control CC/IP directly.

Many tests are not compatible though, so we mark test cases and classes with a to-be-fixed annotation to tackle them later.

The changeset is large, but trivial:
1. We remove the disablement of the IP executor from the build scripts
2. We add `@ToBeFixedForIsolatedProjects` or `@UnsupportedWithIsolatedProjects` annotations to the tests that failed.

FTR, the `because` of the annotations might not be 100% accurate, since it was suggested by an agent. We accept this trade-off, which is better than having an empty reason.

---

Better online review experience: https://diffshub.com/gradle/gradle/pull/38203/changes

### DIFF
--- a/platforms/core-configuration/kotlin-dsl-integ-tests/src/integTest/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
+++ b/platforms/core-configuration/kotlin-dsl-integ-tests/src/integTest/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
@@ -44,6 +44,7 @@ import org.hamcrest.MatcherAssert.assertThat
 import org.junit.Assert.assertNotEquals
 import org.junit.Test
 import spock.lang.Issue
+import org.gradle.integtests.fixtures.modes.ToBeFixedForIsolatedProjects
 
 
 class GradleKotlinDslIntegrationTest : AbstractKotlinIntegrationTest() {
@@ -817,6 +818,7 @@ class GradleKotlinDslIntegrationTest : AbstractKotlinIntegrationTest() {
         )
     }
 
+    @ToBeFixedForIsolatedProjects(because = "Kotlin DSL cross-project configuration")
     @Test
     fun `can cross configure buildscript`() {
 
@@ -965,6 +967,7 @@ class GradleKotlinDslIntegrationTest : AbstractKotlinIntegrationTest() {
         build("help")
     }
 
+    @ToBeFixedForIsolatedProjects(because = "Kotlin DSL cross-project configuration")
     @Test
     fun `can use kotlin java8 inline-only methods`() {
 
--- a/platforms/core-configuration/declarative-dsl-core/build.gradle.kts
+++ b/platforms/core-configuration/declarative-dsl-core/build.gradle.kts
@@ -40,6 +40,4 @@ gradleModule {
     }
 }
 
-tasks.isolatedProjectsIntegTest {
-    enabled = false
-}
+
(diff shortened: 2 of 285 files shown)


alllex:

alllex:
@bot-gradle test RfR

eskatos (approved):
TIDE scope LGTM

platforms/core-configuration/kotlin-dsl-integ-tests/src/integTest/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
```diff
@@ -817,6 +818,7 @@ class GradleKotlinDslIntegrationTest : AbstractKotlinIntegrationTest() {
         )
     }
 
+    @ToBeFixedForIsolatedProjects(because = "Kotlin DSL cross-project configuration")
```
6hundreds, line 821: ❓ I don't think it's `to be fixed`, the scenario is unsupported under IP by the definition. It's too many places in this PR to verify the correctness of `to be fixed` application tho, so I'll stop on the first one. 
I think it's important to use proper annotation for the further analysis of `to be fixed` scenarios. WDYT?

alllex: > I think it's important to use proper annotation for the further analysis of to be fixed scenarios. WDYT?

I agree it might be better to annotate everything perfectly, but I don't think it's feasible at the scale of the incompatible tests that we have. Either way, the goal is to eventually check every incompatible test and assess what is the problem and what is actually being tested.

In that sense, I think the `unsupported` is a stronger statement than `to be fixed`. Nothing prevents us from changing the annotations later as we look at every individual test.

I see these annotations as non-prescriptive. None of them signal our commitment about the IP support for certain features. Do you see it differently?

I do see a lot of value in not having blanket ignores for running test with IP at the project level. So I don't want to block this PR on investigations of individual tests.

As a compromise, we can add some prefix to the reason of the annotation, such as `requires investigation to confirm:` or just `INVESTIGATE: `. This way it becomes more clear that some annotations are less trustworth than others.

6hundreds: > I do see a lot of value in not having blanket ignores for running test with IP at the project level. 

💯 

> Nothing prevents us from changing the annotations later as we look at every individual test.

For sure.

> None of them signal our commitment about the IP support for certain features.

Well, partially. I just thinking about the confusion of low-context readers(team members at the very least) when see this explicitly cross-project scenario marked by `ToBeFixedWithIP`. Maybe I overestimate it tho.

6hundreds (approved):
Benefits of removing blanket IP disablement overweights some incorrectly applied annotations, I think. LGTM

cobexer:
looks like there is a test failure: https://ge.gradle.org/s/nkhoeobjw3eqa/tests/task/:kotlin-dsl:isolatedProjectsIntegTest/details/org.gradle.kotlin.dsl.integration.KotlinDslParentProjectPropertyLookupIntegrationTest/val%20by%20project%20resolved%20from%20parent%20is%20deprecated%20with%20the%20Kotlin%20delegation%20caller%20context?top-execution=1

alllex:
This PR is going to be tricky to merge 🫤

PR #38210, opened 2026-06-15
Disable Kotlin DSL accessor build caching by default

### LINKED ISSUE #37278: Investigate if build cache is worth it for Kotlin DSL script compilation
## Context

Kotlin DSL script compilation encompasses multiple separately cached units of work:

1. **Accessor generation** (`GenerateProjectAccessors`, `GeneratePluginSpecBuilderAccessors`, `GenerateVersionCatalogAccessors`)
2. **Script compilation** (`KotlinScriptCompilationAndInstrumentation` for stage1/CLASSPATH and stage2/BODY)

These units of work are cached via the execution engine's build cache support. However, analysis of real-world traces (build with ~270 projects after a build-logic ABI change) shows that the caching overhead for accessors can significantly exceed the actual execution time, especially on cache miss.

### What we know so far

From a Perfetto trace of a build-logic ABI change (all cache misses, no IP):

- **Accessor generation**: 271 work units took 21.6s to execute but incurred 59.4s of caching overhead (2.75x ratio). The overhead is dominated by remote cache lookup (33.0s), output snapshotting (17.7s), and packing (8.7s). All work runs on a single thread, so the overhead directly adds to wall-clock configuration time.
- **Script compilation** had caching disabled via internal property in this trace, so no comparable data is available yet for the cache miss overhead of .kts compilation.

Accessor results are cached in the GUH (`ImmutableUnitOfWork`) which acts as a local cache keyed by inputs hash. This means that even without build cache, results are shared across projects within the same build and reusable when e.g. switching branches. The build cache adds remote lookup, output snapshotting, packing, and store operations on top.

### Precedent

Groovy DSL script compilation had the same problem. Build cache was enabled in 8.7 and then disabled in 8.8 after users reported 87% regression on cache miss with 50ms remote cache latency (#28844).

### Phase 1: Measure cache miss overhead

Quantify the overhead of build cache support on cache miss for each unit of work type.

**For accessor generation:**
- [x] Data already collected from Perfetto trace analysis (see context above)
- [ ] Validate with additional traces from different builds/projects

**For .kts script compilation (stage1 and stage2):**
- [ ] Collect a Perfetto trace from a build-logic ABI change scenario with script caching enabled (i.e. without `kotlin-script-caching-disabled`)
- [ ] Measure: execution time vs. remote cache lookup, output snapshotting, packing, and store per work unit
- [ ] Compare the overhead ratio to the accessor generation data

### Phase 2: Measure cache hit savings

Quantify the benefit of build cache support on cache hit.

- [ ] Collect a Perfetto trace from an ephemeral CI scenario (cold local cache, warm remote cache) where all Kotlin DSL cache entries are hits
- [ ] For each work unit type, compare: time to fetch and unpack from cache vs. time to execute from scratch
- [ ] Compute the net savings (or cost) of build cache support on cache hit, accounting for output snapshotting overhead even on hits

### Phase 3: Decision

Based on the data:

- [ ] If cache miss overhead significantly exceeds cache hit savings (as was the case for Groovy), disable build cache by default for the affected units of work. The GUH already provides local caching.
- [ ] If some unit of work types benefit from build cache (e.g. expensive script compilations) while others do not (e.g. fast accessor generation), consider per-type defaults.

## Related

- #28844 -- Disable build cache for Groovy DSL script compilation (same problem, resolved by disabling)
- #37276 -- Honor `kotlin-script-caching-disabled` in accessor generation

hegyibalint:
Implementation of the decision made in https://github.com/gradle/gradle/issues/37278

The analysis of the issue showed that the payoff, especially for accessor generation artifacts, is only applicable in a thin zone.

For accessor generation, we established that a 19ms RTT[^1] to the cache server already zeroes out the benefits.
Also, we identified that the cache-related operations (snapshotting, packing, unpacking) is significant vs. the script compilation artifacts. Therefore, we decided to disable accessor generation caching.

[^1]: This is measured on one specific machine; numbers might vary depending on how fast a given machine is. Slower machines push this number up, faster ones down.

### DIFF
--- a/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/Stage1BlocksAccessorClassPath.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/Stage1BlocksAccessorClassPath.kt
@@ -72,8 +72,7 @@ class Stage1BlocksAccessorClassPathGenerator @Inject internal constructor(
     internalOptions: InternalOptions,
 ) {
 
-    private val cachingDisabled: Boolean =
-        internalOptions.getBoolean(KotlinDslInternalOptions.CACHING_DISABLED_PROPERTY)
+    private val scriptDisabledReason = KotlinDslInternalOptions.accessorCachingDisabledReason(internalOptions)
 
     private
     val stage1BlocksAccessorClassPath by lazy {
@@ -123,7 +122,7 @@ class Stage1BlocksAccessorClassPathGenerator @Inject internal constructor(
                     fileCollectionFactory,
                     inputFingerprinter,
                     workspaceProvider,
-                    cachingDisabled,
+                    scriptDisabledReason
                 )
                 executionEngine.createRequest(work)
                     .execute()
@@ -148,7 +147,7 @@ class Stage1BlocksAccessorClassPathGenerator @Inject internal constructor(
             fileCollectionFactory,
             inputFingerprinter,
             workspaceProvider,
-            cachingDisabled,
+            scriptDisabledReason
         )
         return executionEngine.createRequest(work)
             .execute()
@@ -166,7 +165,7 @@ abstract class AbstractStage1BlockAccessorsUnitOfWork(
     private val fileCollectionFactory: FileCollectionFactory,
     private val inputFingerprinter: InputFingerprinter,
     private val workspaceProvider: KotlinDslWorkspaceProvider,
-    private val cachingDisabled: Boolean,
+    private val cachingDisabledReason: CachingDisabledReason?
 ) : ImmutableUnitOfWork {
 
     companion object {
@@ -175,12 +174,7 @@ abstract class AbstractStage1BlockAccessorsUnitOfWork(
         const val CLASSES_OUTPUT_PROPERTY = "classes"
     }
 
-    override fun shouldDisableCaching(detectedOverlappingOutputs: OverlappingOutputs?): Optional<CachingDisabledReason> {
-        if (cachingDisabled) {
-            return Optional.of(KotlinDslInternalOptions.CACHING_DISABLED_REASON)
-        }
-        return super.shouldDisableCaching(detectedOverlappingOutputs)
-    }
+    override fun shouldDisableCaching(detectedOverlappingOutputs: OverlappingOutputs?): Optional<CachingDisabledReason> = Optional.ofNullable(cachingDisabledReason)
 
     override fun identify(scalarInputs: MutableMap<String, ValueSnapshot>, fileInputs: MutableMap<String, CurrentFileCollectionFingerprint>) =
         Identity { "$classLoaderHash-$identitySuffix" }
--- a/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinDslInternalOptions.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinDslInternalOptions.kt
@@ -22,8 +22,30 @@ import org.gradle.internal.execution.caching.CachingDisabledReason
 import org.gradle.internal.execution.caching.CachingDisabledReasonCategory
 
 internal object KotlinDslInternalOptions {
-    val CACHING_DISABLED_PROPERTY: InternalOption<Boolean> =
+    // Disables build caching for both Kotlin script compilation AND accessor generation
+    private val CACHING_DISABLED_PROPERTY: InternalOption<Boolean> =
         InternalOptions.ofBoolean("org.gradle.internal.kotlin-script-caching-disabled", false)
-    val CACHING_DISABLED_REASON: CachingDisabledReason =
+    private val CACHING_DISABLED_REASON: CachingDisabledReason =
         CachingDisabledReason(CachingDisabledReasonCategory.NOT_CACHEABLE, "Caching of Kotlin script compilation and Kotlin DSL accessors generation disabled by property")
+
+    // Disables build caching only for accessor generation (overridden by CACHING_DISABLED_PROPERTY)
+    private val ACCESSOR_CACHING_DISABLED_PROPERTY: InternalOption<Boolean> =
+        InternalOptions.ofBoolean("org.gradle.internal.kotlin-script-accessors-caching-disabled", true)
+    private val ACCESSOR_CACHING_DISABLED_REASON: CachingDisabledReason =
+        CachingDisabledReason(CachingDisabledReasonCategory.NOT_CACHEABLE, "Build caching of Kotlin script accessor generation disabled by property")
+
+    /**
+     * Returns a reason why the script and accessor generation caching is disabled, or `null` if there is none
+     */
+    fun cachingDisabledReason(internalOptions: InternalOptions): CachingDisabledReason? =
+        if (internalOptions.getBoolean(CACHING_DISABLED_PROPERTY)) CACHING_DISABLED_REASON else null
+
+    /**
+     * Returns a reason why specifically the accessor generation is disabled, or `null` if there is none
+     */
+    fun accessorCachingDisabledReason(internalOptions: InternalOptions): CachingDisabledReason? = when {
+        internalOptions.getBoolean(CACHING_DISABLED_PROPERTY) -> CACHING_DISABLED_REASON
+        internalOptions.getBoolean(ACCESSOR_CACHING_DISABLED_PROPERTY) -> ACCESSOR_CACHING_DISABLED_REASON
+        else -> null
+    }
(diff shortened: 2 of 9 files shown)


lkasso (approved):
Docs LGTM.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinDslInternalOptions.kt
```diff
@@ -26,4 +26,11 @@ internal object KotlinDslInternalOptions {
         InternalOptions.ofBoolean("org.gradle.internal.kotlin-script-caching-disabled", false)
     val CACHING_DISABLED_REASON: CachingDisabledReason =
         CachingDisabledReason(CachingDisabledReasonCategory.NOT_CACHEABLE, "Caching of Kotlin script compilation and Kotlin DSL accessors generation disabled by property")
+
+    // Build-cache control for Kotlin DSL accessor generation (gradle/gradle#37278).
+    // Unlike CACHING_DISABLED_PROPERTY, this targets only the remote/local build cache:
```
eskatos: ❌ This is wrong, `CACHING_DISABLED_PROPERTY` also only affects the build-cache but for both accessors generation and script compilation.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinDslInternalOptions.kt
```diff
@@ -26,4 +26,11 @@ internal object KotlinDslInternalOptions {
...
+    val ACCESSOR_BUILD_CACHE_DISABLED_PROPERTY: InternalOption<Boolean> =
+        InternalOptions.ofBoolean("org.gradle.internal.kotlin-accessor-build-cache-disabled", true)
```
eskatos: ❌ Let's make this property name similar to the existing one above, for uniformity. For example `org.gradle.internal.kotlin-script-accessors-caching-disabled`.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinDslInternalOptions.kt
```diff
@@ -26,4 +26,11 @@ internal object KotlinDslInternalOptions {
...
+    val ACCESSOR_BUILD_CACHE_DISABLED_REASON: CachingDisabledReason =
+        CachingDisabledReason(CachingDisabledReasonCategory.NOT_CACHEABLE, "Build cache for Kotlin DSL accessor generation is disabled by default (gradle/gradle#37278)")
```
eskatos: ❌ Let's not reference issues with a GitHub notation in caching disabled reasons. We do that nowhere else. Make the reason similar to the other one, for uniformity. For example `Caching of Kotlin script accessors generation disabled by property`.

eskatos (changes requested):

eskatos:
@bot-gradle test RFN ACT without PTS

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinDslInternalOptions.kt
hegyibalint: Right. Also this comment doesn't have a lot of value, I think I will just remove it

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinDslInternalOptions.kt
hegyibalint: How about deepening the structure?

```yaml
"org.gradle.internal.kotlin-script":
  - "caching": (default false)
  - "accessor-build-caching": (default false)
```

I.e. we would have here:

`InternalOptions.ofBoolean("org.gradle.internal.kotlin-script.accessor-build-caching", false)`

And above:
`InternalOptions.ofBoolean("org.gradle.internal.kotlin-script.caching", false)`

This IMO would make the options a bit more understandable.

hegyibalint: Maybe an even better alternative would be to completely detach the two from each other.

The current implementation makes `kotlin-script-caching-disabled` a global option (also disabled accessor generation). This is not a straightforward design, and somebody needs to see e.g. `Stage1BlocksAccessorClassPath` to understand how the flags interact.

```yaml
"org.gradle.internal.kotlin-script":
  - "script-caching": (default false) // Only controls script compilation.
  - "accessor-build-caching": (default false) // Only controls the accessor generation
```

WDYT?

eskatos: Today we have one property that encompasses build-cache for compilation and accessors generation:
```
org.gradle.internal.kotlin-script-caching-disabled
```

If we would add two properties to control them independently I would keep it simple and uniform:
```
org.gradle.internal.kotlin-script-accessors-caching-disabled
org.gradle.internal.kotlin-script-compilation-caching-disabled
```

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/Stage1BlocksAccessorClassPath.kt
```diff
@@ -179,6 +174,9 @@ abstract class AbstractStage1BlockAccessorsUnitOfWork(
         if (cachingDisabled) {
             return Optional.of(KotlinDslInternalOptions.CACHING_DISABLED_REASON)
         }
+        if (buildCachingDisabled) {
```
eskatos: ❌ The other `cachingDisabled` field is already about the build cache. The name of that new field is confusing at least. Given we are in the UoW already, I would resolve the interaction between the two system properties outside of the UoW to make things more readable.

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinDslInternalOptions.kt
eskatos: Don't get me wrong, I'm not saying "let's add both in this PR".

hegyibalint: > If we would add two properties to control them independently

I assume replacing the all-encompassing `org.gradle.internal.kotlin-script-caching-disabled`. Can you confirm?

eskatos: No, I never mentioned removing it.
Also, I explicitly wrote above that I'm not saying we should add two properties now.
The only reason to remove the existing property would be if we introduce the other two.
Even if is internal, some folks are unfortunately relying on the existing one. If we remove it right away we will cause pain.

Can you clarify what is it that you have in mind?

hegyibalint: The fact that some people rely on this changes my approach.
Otherwise I could've argued that the separation might be worthwhile to be done now, as I believe the behavior would be easier to understand. 

Let's not change the approach here then, I will just align the name here.

hegyibalint:
@bot-gradle test RFN ACT without PTS

hegyibalint:
@bot-gradle cancel

hegyibalint:
@bot-gradle test RFN ACT without PTS

hegyibalint:
@eskatos thanks for the comments, I've realized that the design has some holes.

I've stepped back, put Claude aside a bit, and came up with a modified design that makes the decision more centralized, and a type change that allows us to give the unit of work the precise reason why the caching is not happening.

(Tests are still running at the time of commenting)

hegyibalint:
@bot-gradle test RFN ACT without PTS

hegyibalint:
Looks like flaky tests
I'm re-running the config to see if it repairs itself: https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_IsolatedProjects_21_bucket6/114093517

hegyibalint:
Build ran successfully, let me retrigger the whole pipeline

hegyibalint:
@bot-gradle test RFN ACT without PTS

eskatos (approved):

PR #37370, opened 2026-03-31
Package accessor generation output into JARs

reinsch82:
## Summary

- Instead of writing ~500 class files and ~500 source files per project, write directly into two JAR files (`classes.jar` and `sources.jar`)
- Reduces execution engine output snapshotting from ~1000 files to 2 files, eliminating a cost that was 82% as long as the actual generation
- Adds `OUTPUT_FORMAT_VERSION` to the identity hash to invalidate old directory-based workspaces

----

* Closes #37279

## Test plan

- [x] Unit tests: `ProjectAccessorsClassPathTest` — verifies JAR structure (classes, sources, module metadata) and runtime behavior
- [ ] Integration tests: `:kotlin-dsl:forkingIntegTest`
- [ ] Config cache tests: `:kotlin-dsl:configCacheIntegTest`
- [ ] Sanity check: `./gradlew sanityCheck`

PR #38211, opened 2026-06-15
Add `accessorHeavyJavaMultiProjectKotlinDsl` performance testbed

hegyibalint:
Secondary results of implementing https://github.com/gradle/gradle/issues/37278

---

Adds a large multi-project Kotlin DSL performance testbed, `largeAccessorHeavyJavaMultiProjectKotlinDsl`, in which each subproject has a **distinct** accessor schema.

The normal `largeJavaMultiProjectKotlinDsl` scenario is too uniform, and only generates 2 units of work, regardless of how many projects are generated.

## What it does

- New `largeAccessorHeavyJavaMultiProjectKotlinDsl` test project: 500 Kotlin DSL subprojects, each applying a buildSrc precompiled script plugin that registers a uniquely-named extension (`ext_<projectName>`). This will make the schema unique and will trigger a whole accessor generation pipeline.
- The `accessorHeavy` flag is plumbed through the generator config/builder so the buildSrc is upgraded to a `kotlin-dsl` project only for this testbed; all other testbeds are unchanged.

### DIFF
--- a/build-logic/performance-testing/src/main/groovy/gradlebuild.performance-templates.gradle
+++ b/build-logic/performance-testing/src/main/groovy/gradlebuild.performance-templates.gradle
@@ -56,9 +56,24 @@ performanceTest.registerTestProject("largeEmptyMultiProjectDeclarativeDsl", Java
 
 
 // === Java ===
-['largeMonolithicJavaProject', 'largeJavaMultiProject', 'smallJavaMultiProjectManyExternalDependencies', 'largeJavaMultiProjectHierarchy', 'largeMonolithicGroovyProject', 'largeGroovyMultiProject', 'largeJavaMultiProjectNoBuildSrc', 'largeJavaMultiProjectKotlinDsl',
- 'mediumMonolithicJavaProject', 'mediumJavaMultiProject', 'mediumJavaMultiProjectWithTestNG', 'mediumJavaCompositeBuild', 'mediumJavaPredefinedCompositeBuild',
- 'smallJavaMultiProject', 'smallJavaMultiProjectNoBuildSrc'].each { template ->
+[
+    'smallJavaMultiProject',
+    'smallJavaMultiProjectManyExternalDependencies',
+    'smallJavaMultiProjectNoBuildSrc',
+    'mediumJavaCompositeBuild',
+    'mediumJavaMultiProject',
+    'mediumJavaMultiProjectWithTestNG',
+    'mediumJavaPredefinedCompositeBuild',
+    'mediumMonolithicJavaProject',
+    'largeAccessorHeavyJavaMultiProjectKotlinDsl',
+    'largeGroovyMultiProject',
+    'largeJavaMultiProject',
+    'largeJavaMultiProjectHierarchy',
+    'largeJavaMultiProjectKotlinDsl',
+    'largeJavaMultiProjectNoBuildSrc',
+    'largeMonolithicGroovyProject',
+    'largeMonolithicJavaProject'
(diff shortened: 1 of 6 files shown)


build-logic/performance-testing/src/main/groovy/gradlebuild.performance-templates.gradle
```diff
@@ -56,9 +56,24 @@ performanceTest.registerTestProject("largeEmptyMultiProjectDeclarativeDsl", Java
...
+[
+    'smallJavaMultiProject',
+    'smallJavaMultiProjectManyExternalDependencies',
+    'smallJavaMultiProjectNoBuildSrc',
+    'mediumJavaCompositeBuild',
+    'mediumJavaMultiProject',
+    'mediumJavaMultiProjectWithTestNG',
+    'mediumJavaPredefinedCompositeBuild',
+    'mediumMonolithicJavaProject',
+    'largeAccessorHeavyJavaMultiProjectKotlinDsl',
```
hegyibalint, line 68: ℹ️ I only added the large case here (500 projects). I am torn if the small or medium case would bring any benefit; accessor generation is a per-project task, so running it 500 times:
 1. provides enough datapoints that the scenario smooths itself out
 2. if the computation becomes more expensive, it will show up just as well in the 500 case as in a 10 or 100 one.

A case for the smaller tests could be exactly the opposite of point 1. above: maybe we should try to measure cases where the optimization is cold or lukewarm (i.e., JVM didn't yet optimize out all paths). With large numbers, their optimized timing might overshadow bad startup behavior.

I'm interested in thoughts about this.

testing/internal-performance-testing/src/main/groovy/org/gradle/performance/generator/JavaTestProjectGenerator.groovy
```diff
@@ -94,6 +94,17 @@ enum JavaTestProjectGenerator {
         .withDsl(KOTLIN)
         .create()),
 
+    LARGE_ACCESSOR_HEAVY_JAVA_MULTI_PROJECT_KOTLIN_DSL(new TestProjectGeneratorConfigurationBuilder("largeAccessorHeavyJavaMultiProjectKotlinDsl", "largeJavaMultiProject")
+        .withSourceFiles(100)
```
hegyibalint, line 98: 🤔 I'm just thinking to myself: I don't think sources add anything to this benchmark, we won't even need them.

h0tk3y:
https://github.com/gradle/gradle/pull/37370 also adds an accessor-heavy test project for performance tests, although without multiple subprojects. Should this testbed be used for that performance test as well?

hegyibalint:
First of all, forgot to post my own comments on the PR...

> https://github.com/gradle/gradle/pull/37370 also adds an accessor-heavy test project for performance tests

Ah, you also did a performance test!

> although without multiple subprojects. Should this testbed be used for that performance test as well?

I think the number of projects doesn't add anything here; heck, maybe worsen the situation if I think about it (accessor generation is getting conflated with other heavy things like configuration performance).

---

Looking at your PR, I think your version is neater than this.
Maybe what I would consider is [my comment](https://github.com/gradle/gradle/pull/38211#discussion_r3414789034) about having smaller accessor generations than 500, maybe surfacing some interesting cold/lukewarm performance behavior.

If you don't have anything to add, I think I can close this.

eskatos:
+1 to close this, the other test is better targeted

hegyibalint:
Closing; comments don't have any effect on this PR

PR #38212, opened 2026-06-15
Open JDK modules on demand for CC serialization

### LINKED ISSUE #20089: Configuration Cache: Unable to make field (...) accessible. module java.desktop does not "opens java.awt" to unnamed module
### Expected Behavior
I can use Configuration Cache with Serializable inputs on Java 17. Or at least have more meaningful error message why it's not supported.
I've read https://docs.gradle.org/7.4/userguide/configuration_cache.html#config_cache:requirements:disallowed_types but I don't see a category `java.awt.Color` would fall into.
Even if such type is disallowed as Task input, or If it's because of https://github.com/gradle/gradle/issues/13588 then I'd expect see a link to the documentation or see the error in configuration-cache report

### Current Behavior
Execution fails with
```
* What went wrong:
Configuration cache state could not be cached: field 'filters' from type 'GreetingTask': error writing value of type 'org.gradle.api.internal.provider.DefaultListProperty'
> Configuration cache state could not be cached: field 'color' from type 'SomeClass': error writing value of type 'java.awt.Color'
   > Unable to make field private java.awt.color.ColorSpace java.awt.Color.cs accessible: module java.desktop does not "opens java.awt" to unnamed module @4e0eec1b
```

### Context
I'm a https://github.com/usefulness/easylauncher-gradle-plugin plugin author. I want to let plugin consumers configure the plugin according to their needs. In this specific case, to configure ribbon colour, which the plugin then use to draw on Canvas. As a workaround I could use String and ask for `#RRGGBB` representation, but it's less convenient to use, especially when invoked from Kotlin scripts

### Steps to Reproduce 
Have a Task with `java.awt.Color` as an input:
```kotlin
@Input val filters = objectFactory.property(java.awt.Color::class.java)
```
Running `./gradlew repro` works, inputs are defined properly. Adding `--configuration-cache` fails the build. 

Full repro: https://github.com/mateuszkwiecinski/configuration_cache_doesnt_work

### Your Environment
JDK 17 Temurin  
Gradle 7.4  
[Build scan URL:https://gradle.com/s/sqor3j7w6j3ym](https://scans.gradle.com/s/ijgegj6d2ymsy)

mlopatkin:
Fixes #20089.

### Context

Configuration Cache serialization uses deep reflection (`Field.setAccessible(true)`, reflective constructor/method invocation) to write and read user task state. Under Java 9+ module rules, deep reflection over JDK types requires the target package to be `open` to the caller's module. Today we mitigate this with `--add-opens=...=ALL-UNNAMED` flags on the daemon JVM command line. The list is never complete — any serializable graph touching a JDK type whose declaring package isn't pre-opened (most recently `java.awt.Color` from `java.desktop`) surfaces as `InaccessibleObjectException` deep inside `BeanSchema`, with no ergonomic recovery path.

We already bundle an instrumentation agent that provides an `Instrumentation` reference, and on Java 17 the `redefineModule(...)` lets us grant that access at runtime. This change routes the cached schema/encoding lookups through a new `ObjectOpener` service that opens the declaring package on first encounter, with the cost bounded to one `redefineModule` call per `(module, package)` per daemon lifetime.

When the agent isn't loaded, the service degrades to a `FallbackObjectOpener` selected at registration time that just calls `setAccessible(true)` and lets the original `InaccessibleObjectException` propagate — matching pre-agent behavior. Graceful diagnostics for the agent-absent case are a follow-up; existing `--add-opens` lines stay on the daemon command line, since other subsystems may still depend on them.

### What changed

- **New module** `platforms/core-runtime/reflection-services` hosts the `ObjectOpener` service. Both the JDK module-package open *and* the subsequent `setAccessible(true)` happen inside this module — same module the agent opens the package to — so caller-side module identity never enters the picture.
- **`Agent.getInstrumentation()`**, a reflective `AgentControl.getInstrumentation()` hop, and `AgentStatus.getInstrumentation()` expose the agent's `Instrumentation` instance.
- **Boot-layer precondition**: `DefaultObjectOpener` throws `IllegalArgumentException` if a class outside `ModuleLayer.boot()` reaches it. CC isn't expected to serialize classes from custom module layers (those require custom classloaders that CC rejects upstream), so this is a defensive guarantee; dynamic-module classes (e.g., `Proxy` synthetic modules) are returned silently — they're handled by a dedicated codec.
- **Cache** is a `Set<String>` of already-opened packages. `redefineModule` is idempotent so the first-encounter race is benign. As boot layer enforces "no split packages" rule, we don't need to track Module separately.
- **Consumers updated**: `BeanSchema`/`DefaultBeanState{Reader,Writer}Lookup`, `JavaSerializationEncodingLookup`/`ReadResolveCache`/`JavaObjectSerializationCodec`, `JavaRecordCodec`+`baseTypes(...)`, plus `ConfigurationCacheCodecs` and `IsolatedActionCodecsFactory`. The previous `Field.makeAccessibleVia(opener)` extension in `BeanSchema.kt` is gone — every reflective site now calls `opener.makeAccessible(ao)`.
- **`SerializedLambdaParametersCheckingCodec`** is deferred (still relies on existing `--add-opens=java.base/java.lang.invoke`).

### Tests

- `ConfigurationCacheJdkModuleAccessIntegrationTest` (in `:configuration-cache`):
  - Smoke: a task field of type `java.awt.Color`. Red before the fix; green after.
  - Cold-daemon variant (`@Requires(NotEmbeddedExecutor)`): stores under one daemon, kills it via `DaemonsFixture.killAll()`, loads under a fresh JVM — exercises the load-path reflection from a daemon that didn't warm any bean-schema caches.
- `ObjectOpenerIntegrationTest` (in `:reflection-services`): exercises the boot-layer precondition. Synthesizes a JPMS JAR via `ModuleJarFixture`, builds a custom `ModuleLayer` from the build script, and calls `ObjectOpener.makeAccessible(...)` on a constructor from the custom layer. Verifies the exception with the expected message.


mlopatkin:
@bot-gradle test ACC

mlopatkin:
@bot-gradle test this

mlopatkin:
@coderabbitai review

mlopatkin:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheJdkModuleAccessIntegrationTest.groovy
```diff
@@ -0,0 +1,74 @@
...
+
+@Issue("https://github.com/gradle/gradle/issues/20089")
+class ConfigurationCacheJdkModuleAccessIntegrationTest extends AbstractConfigurationCacheIntegrationTest {
+    def "restores task field referencing JDK type from non-open module"() {
+        buildFile << colorTask()
+
+        expect:
+        configurationCacheRun "ok"
+        outputContains("color = java.awt.Color[r=1,g=2,b=3]")
+    }
```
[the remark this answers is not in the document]
mlopatkin: Not really applicable.

mlopatkin:
@bot-gradle cancel

mlopatkin:
@bot-gradle test this

mlopatkin:
The failing tests are flakes

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -63,6 +63,16 @@ The isolation also applies to objects used in `synchronized` blocks, so they can
 
 To share or coordinate state between tasks, use <<build_services.adoc#build_services, shared build services>>.
 
+[[config_cache:requirements:user_defined_modules]]
+==== Classes from User-defined Modules
+
+Types loaded into a user-defined `ModuleLayer` (for example, one created at runtime via `ModuleLayer.defineModulesWith*Loader(...)`) must not be referenced by task state.
+The Configuration Cache ignores module information when saving and cannot restore such types properly, so they're rejected at store time.
+This restriction only applies to build logic (including plugins and dependencies of plugins). Code built with Gradle isn't affected.
```
lkasso:
```suggestion
This restriction only applies to build logic (including plugins and dependencies of plugins).
Code built with Gradle isn't affected.
```

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -63,6 +63,16 @@ The isolation also applies to objects used in `synchronized` blocks, so they can
...
+
+Standard JDK types are unaffected: every module shipped with the JVM lives in the boot `ModuleLayer` and is fully supported, including types from non-`java.base` modules such as `java.desktop` (`java.awt.Color`, `javax.swing.*`) or `java.xml`.
+In environments where the agent is available, Gradle automatically opens any JDK packages that need deep reflective access during serialization, so user code does not need to add `--add-opens` for them.
```
lkasso: Nit pick on coderabbit suggestion:
```suggestion
When the Gradle Java agent is enabled (the default), Gradle automatically opens the JDK packages it needs during serialization, so user code does not need to add `--add-opens` for them.
```

lkasso (approved):
Docs LGTM.

platforms/core-configuration/bean-serialization-services/src/main/kotlin/org/gradle/internal/serialize/beans/services/BeanSchema.kt
```diff
@@ -37,29 +38,34 @@ val unsupportedFieldDeclaredTypes = listOf(
...
+        else -> relevantFieldsOf(beanType, objectOpener)
     }
 
 
 private
-fun relevantFieldsOf(beanType: Class<*>) =
+fun relevantFieldsOf(beanType: Class<*>, objectOpener: ObjectOpener) =
     relevantTypeHierarchyOf(beanType)
         .flatMap(Class<*>::relevantFields)
-        .onEach(Field::makeAccessible)
```
abstratt, line 51: 💭 

It seems this extension may have no remaining usages, if so, maybe delete it?

platforms/core-runtime/reflection-services/src/main/java/org/gradle/internal/reflection/access/ObjectOpener.java
```diff
@@ -0,0 +1,50 @@
...
+    /// @throws IllegalArgumentException if the module doesn't come from [ModuleLayer#boot()]
+    /// @throws InaccessibleObjectException if the agent is not available and making the entity accessible fails
+    void makeAccessible(AccessibleObject accessibleObject) throws InaccessibleObjectException;
+
+    /// Returns an [ObjectOpener] that does not open any JDK module packages - equivalent to the
+    /// implementation used in production when the instrumentation agent is unavailable. Suitable
+    /// for tests and fixtures that don't exercise module access; still calls `setAccessible(true)`
+    /// on the given object.
+    static ObjectOpener agentless() {
+        return new FallbackObjectOpener();
```
abstratt, line 48: 💭 

Maybe use a singleton for `FallbackObjectOpener`, unclear how many instances we are talking about

abstratt (approved):
Pretty cool, pointed out a couple of minor things

platforms/core-runtime/reflection-services/src/main/java/org/gradle/internal/reflection/access/ObjectOpener.java
mlopatkin: > 💭
> 
> Maybe use a singleton for `FallbackObjectOpener`, unclear how many instances we are talking about

It is a single `@Scope(Global)` instance, so I don't think optimizing this is worth it.

PR #35252, opened 2025-10-08
Update tested Kotlin versions

eskatos:
add 2.2.21-RC and 2.3.0-Beta1

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #38217, opened 2026-06-15
Automate tested Kotlin version updates

hegyibalint:
### Context

Tested-Kotlin-version bumps have been done manually (e.g. #35363, #35252, #34480), whereas the equivalent AGP bumps are automated by a scheduled workflow. This adds the same automation for Kotlin and factors out the shared plumbing both share.

The flow is very similar to the previous AGP automation:
 - run the `updateKotlinVersions` task
 - commit / push / create a PR
 - send the PR to the merge queue

An added interesting thing is a new abstraction, the `open-pr` action that takes care of doing the (quite involved) process of managing git and GitHub.

### DIFF
--- a/.github/actions/open-pr/action.yml
+++ b/.github/actions/open-pr/action.yml
@@ -1,0 +1,82 @@
+name: "Open a bot PR"
+description: "Commit the current working-tree changes onto a branch, push it, and open a PR as bot-gradle (optionally asking the merge bot to test and merge)."
+
+inputs:
+  branch:
+    description: "Branch to create the changes on and open the PR from"
+    required: true
+  title:
+    description: "PR title (also used as the PR body)"
+    required: true
+  message:
+    description: "Commit message"
+    required: true
+  reviewer:
+    description: "Reviewer to request on the PR"
+    required: true
+  labels:
+    description: "Comma-separated labels to apply to the PR"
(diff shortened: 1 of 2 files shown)


.github/actions/open-pr/action.yml
hegyibalint, line 1: ℹ️ This was originally the core of `update-agp-versions.yml`, but since both pipelines need this mechanism, got abstracted into a GitHub action.

.github/workflows/update-agp-versions.yml
```diff
@@ -21,54 +21,18 @@ jobs:
...
-            --label "a:chore,in:performance-test,in:smoke-test,re:android"
-          gh pr comment tide/update-agp --body '@bot-gradle test and merge'
+      - name: Open bot PR
+        uses: ./.github/actions/open-pr
+        with:
+          branch: tide/update-agp
+          title: Update tested AGP versions
+          message: Update AGP versions
+          reviewer: gradle/bt-tide
+          labels: a:chore,re:android,in:performance-test,in:smoke-test,to-triage
```
hegyibalint: ☝️ Apply extra scrutiny to the labels! I think I have the right set, but I'm never sure

eskatos (approved):

PR #38218, opened 2026-06-15
Junit 4 and TestNG test logging logs fatal initialization events

### LINKED ISSUE #26177: Test logging should log fatal events regardless of granularity
See https://github.com/gradle/gradle/issues/25857.

Test logging currently has the concept of "granularity" which allows users to make the events that are logged to the console as verbose or as sparse as they would like.  Test events are assigned a granularity level based on what they are associated with.  Level 0 are events associated with the test task execution, 1 are events related to the forked JVM process(es) and framework initialization, 2 are events related to the test class/suite, 3 is the test method and higher levels correspond to iterations of the test method (for instance parameterized tests).  The user can set the minimum and maximum granularity to log to the console.  For example min=0 and max=2 corresponds to all events related to the test task, the spawned test processes and each test class, but not each test method.  On the other hand, min=3 and max=3 corresponds to only test method events, but not task/process/class events and not test method iterations (i.e. events higher than 3).

The default granularity is -1 which means the highest granularity events, whatever those are - in other words, level 3 (test method) or higher depending on if the tests are parameterized or not.  In practice, this means excluding events that are associated with test descriptor that is a "composite" (i.e. test suites, classes, parent methods that are parameterized, etc).

When the default is in place, and there is a failure while initializing the framework, the exception is not logged to the console because it is associated with a composite descriptor (i.e. test initialization) and the event is filtered out.  However, in such a case, it is unreasonable to exclude this test event because it actually is telling the user why higher level events will not arrive.

We should always log these initialization failure events and distinguish them from other test suite failure events.

tresat:
When a test framework fails to initialize — TestNG test class constructor
throws, JUnit Platform `@BeforeAll` throws, JUnit 4 custom Runner throws —
the failure is reported against a composite descriptor (worker root suite,
test class, engine container). Under the default `testLogging` granularity
(`minGranularity = -1`, leaves only) the `TestEventLogger` silently filters
the event out: the user sees `> There were failing tests` with no exception,
and has to dig through the XML report to find out *why* no tests ran.

## Key points

- The semantic distinction is "this failure is from a specific test method" vs "this failure is from framework lifecycle / initialization."
- The opt-in factory `fromTestMethodFailure` makes this explicit at every emitter site.
- Mapper-level wiring uses `isFailureDuringTest` for the short-circuit + fallback hint.
- All three frameworks (JUnit 4, JUnit Platform, TestNG) have symmetric positive + negative integ test coverage.
- The `DefaultThrowableToTestFailureMapper` change is the heart of this that
Fixes:
- https://github.com/gradle/gradle/issues/26177

## The fix

- **API**: new `TestFailureDetails.isFrameworkFailure()` predicate
  (`@Incubating`, `@since 9.7.0`) implemented by a new internal
  `TestFrameworkFailureDetails` subclass. Public factory
  `TestFailure.fromTestMethodFailure(Throwable)` is the opt-in for ordinary
  per-method failures (non-bypass-eligible).
- **Mapper plumbing**: `ThrowableToTestFailureMapper.createFailure(Throwable, boolean isFailureDuringTest)`
  takes a context hint. The default mapper short-circuits when the hint is
  `false` — skips the mapper chain entirely and emits
  `fromTestFrameworkFailure`. This closes the mapper-claim hazard for
  container-level assertion failures (e.g. `@BeforeAll` using
  `Assertions.fail(...)`).


tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

lkasso (approved):
Docs LGTM. Made minor edits.

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -113,30 +113,43 @@ In all modes, the severity of Isolated Projects violations is now independent of
...
+ExampleTest > initializationError FAILED
+    framework-startup org.testng.TestNGException: Cannot instantiate class ExampleTest
+        at org.testng.internal.ObjectFactoryImpl.newInstance(...)
+        ...
 ```
 
-See the [Timestamp for files inside archives](userguide/working_with_files.html#sec:reproducible_timestamp) section in the Gradle User Manual for more details.
+The `testLogging.events` predicate still applies, explicitly silencing `FAILED` events is honored.
+
+The new `TestFailureDetails.isFrameworkStartupFailure()` predicate exposes this distinction to Tooling-API and Build-Scan consumers, who may render framework-startup failures differently from ordinary test failures.
```
ghale: Didn't this get changed to `isFrameworkFailure()`?
```suggestion
The new `TestFailureDetails.isFrameworkFailure()` predicate exposes this distinction to Tooling-API and Build-Scan consumers, who may render framework-startup failures differently from ordinary test failures.
```

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/logging/TestEventLogger.java
```diff
@@ -105,13 +106,34 @@ private void before(TestDescriptor descriptor) {
 
     private void after(TestDescriptor descriptor, TestResult result) {
         TestLogEvent event = getEvent(result);
-
-        if (shouldLogEvent(descriptor, event)) {
+        boolean bypassGranularity = hasFrameworkFailure(result) || hasOwnFailureOnComposite(descriptor, result);
```
ghale, line 109: ❌  So, `TestFailureDetails.isFrameworkFailure()` states that a "framework failure" is one that doesn't occur `inside of a test method`.  However, because of the way the `DefaultThrowableToTestFailureMapper` works, it tries to map test failures to known `TestFailureMapper` instances.  If it doesn't match any, then it falls through and calls `TestFailure.fromTestFrameworkFailure(throwable)`.  This means that anything that comes out of a test method that doesn't match a `TestFailureMapper` is considered a "test framework failure".  This includes things like NPEs and other runtime exceptions which are probably somewhat common.  So these errors will bypass granularity and get sent to the console.   

I don't see this as a problem, per se, since I don't think that's a terrible way to "fail" (or at least it is better than before) and I feel like this granularity setting is garbage that should go away in favor of a better solution, anyways, but it does contradict what is in the javadoc for `TestFailureDetails.isFrameworkFailure()`.  i.e. a "framework failure" is not somthing that happens outside of a test method, it's any failure that isn't mapped to a known failure type.  I think we should at least update that javadoc to reflect this.  We might also consider changing the method name to be something other than "framework failure" to something like "unmapped failure" or "runtime failure" or something like that (although I'm not sure either of those is better).

ghale (changes requested):
This mostly looks good.  I left a comment about how the javadoc for `isFrameworkFailure()` doesn't actually match the real behavior.  I think we should resolve this in some way before merging.

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/junit/platform/JUnitPlatformInitFailureConsoleLoggingIntegrationTest.groovy
```diff
@@ -109,4 +110,106 @@ class JUnitPlatformInitFailureConsoleLoggingIntegrationTest extends AbstractInte
...
+
+        expect:
+        fails("test")
+
+        outputContains("FooTest > initializationError FAILED")
+        outputContains("org.opentest4j.AssertionFailedError")
+    }
+
+    @Issue("https://github.com/gradle/gradle/issues/26177")
+    static class JUnitPlatformOrdinaryFailuresArentMisclassifiedIntegrationTest extends AbstractIntegrationSpec {
```
ghale, line 159: 💭 This is kind of odd - why an inner class for this rather than just additional test methods?  Doesn't need to be fixed, I'm just curious if this was intentional and if so, why?

ghale (approved):
LGTM

platforms/jvm/testing-jvm/src/integTest/groovy/org/gradle/testing/junit/platform/JUnitPlatformInitFailureConsoleLoggingIntegrationTest.groovy
tresat: Just where I pasted it, to keep it separate from the other test methods.

PR #38222, opened 2026-06-16
Use shared intTestHome dir for smoke tests and leverage artifact cache to speed up smoke tests

blindpirate:
## Why

Closes https://github.com/gradle/gradle-private/issues/5263

We noticed slow smoke tests, [example](https://ge.gradle.org/s/hbdpp7m2cyyno/tests/overview), each test case may take as slow as 10m. Checking out the builds inside the smoke test, we can find that it spend most time on downloading deps: https://ge.gradle.org/s/65cokdhnqwuti/performance/network-activity

This is because:

1. Some smoke tests use dedicated gradle user home instead of `intTestHomeDir`. Every time the test needs to download ~1G deps.
2. We don't share intTestHomeDir on EC2.

## What

**1. Share the integration-test Gradle user home in the heavy Android smoke tests**
`AbstractAndroidProjectSmokeTest` now uses the shared `intTestHomeDir` home (like the rest of the smoke/integ tests) instead of a fresh per-test `test-kit-home`. Dependencies are downloaded once and reused across methods. These tests run serially (`maxParallelForks = 1`), so sharing the home — and reusing its daemons — is safe; the daemon cleanup moves to `cleanupSpec`.

**2. Artifact-cache hooks for the integration-test home**
New repository hooks under `.teamcity/hooks/` (run by the [teamcity-hooks-plugin](https://github.com/gradle/teamcity-hooks-plugin)) restore (pre-build) and store (post-build) the integration-test Gradle user homes `intTestHomeDir/distributions-*`. Each distribution home maps to its own per-base-branch image. Restore/store failures never fail the build.

The hooks know nothing about how the cache works — no CLI jar, no Java version, no agent paths. They go through the single, stable contract the CI agent provides and publishes as `GRADLE_UNIVERSAL_CACHE_CLI_PATH` (added in gradle/dev-infrastructure#3163):

```
"$GRADLE_UNIVERSAL_CACHE_CLI_PATH" (--store|--restore) <image-name> <gradle-home>
```

The repo side decides only **what** to cache:
- `pre-build`/`post-build` are thin shims (those file names are required by the plugin) over a single `hooks.sh (pre|post)`;
- `hooks.sh` gates on `BUILD_TYPE_ID` containing `SmokeTest`, enumerates the distribution homes, and computes the per-base-branch image name;
- **if `GRADLE_UNIVERSAL_CACHE_CLI_PATH` is set and executable it is referenced for store/restore; otherwise the hook does nothing** (local builds, non-EC2 agents). Its presence also implies an EC2 agent with the CLI baked in, so no separate EC2 probe is needed.

> Companion PR: gradle/dev-infrastructure#3163 introduces `universal-cache.sh` and publishes `GRADLE_UNIVERSAL_CACHE_CLI_PATH`. The two land together.

## Verification (live, on real EC2 agents)

Before:

image

image

After:

image

image

### DIFF
--- a/.teamcity/hooks/hooks.sh
+++ b/.teamcity/hooks/hooks.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+#
+# Copyright 2026 the original author or authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Repository build hook for the integration-test Gradle user home (intTestHomeDir).
+# Restores (pre) or stores (post) each distribution home via the artifact-cache contract
+# the CI agent provides. The agent publishes that contract's path as
+# GRADLE_UNIVERSAL_CACHE_CLI_PATH; this hook knows nothing about the CLI itself. When the
+# variable is absent (local builds, non-EC2 agents) the hook does nothing.
+#
+# Usage: hooks.sh (pre|post)   -- invoked by the thin pre-build/post-build shims.
+
+set -euo pipefail
+
+PHASE="${1:-}"
+case "${PHASE}" in
+  pre|post) ;;
+  *) echo "Usage: $(basename "$0") (pre|post)" >&2; exit 2 ;;
+esac
+
+# Only smoke tests share the integration-test Gradle user home, so only they benefit here.
+if [[ "${BUILD_TYPE_ID:-}" != *SmokeTest* ]]; then
+  echo "Skipping artifact cache: BUILD_TYPE_ID='${BUILD_TYPE_ID:-}' is not a SmokeTest build."
+  exit 0
+fi
+
+# The CI agent provides the artifact-cache contract; without it (local/non-EC2) do nothing.
+if [[ ! -x "${GRADLE_UNIVERSAL_CACHE_CLI_PATH:-}" ]]; then
+  echo "Skipping artifact cache: GRADLE_UNIVERSAL_CACHE_CLI_PATH is not set or not executable."
+  exit 0
+fi
+
+INT_TEST_HOME_DIR="$(pwd)/intTestHomeDir"
+# Homes to restore; on store we only pick up the ones that were populated.
+INT_TEST_DISTRIBUTIONS="${INT_TEST_DISTRIBUTIONS:-full jvm basics native}"
+
+# Image name, stable across builds of the same base branch (release vs master).
+if [[ "${BUILD_TYPE_ID:-}" == Gradle_Release* ]]; then
+  baseBranch="release"
+else
+  baseBranch="master"
+fi
+
+for dist in ${INT_TEST_DISTRIBUTIONS}; do
+  gradleHome="${INT_TEST_HOME_DIR}/distributions-${dist}"
+  imageName="gradle-${BUILD_TYPE_ID}-${baseBranch}-distributions-${dist}"
+
+  if [[ "${PHASE}" == "pre" ]]; then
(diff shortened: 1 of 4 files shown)


blindpirate:
@bot-gradle squash

blindpirate:
@bot-gradle test

.teamcity/hooks/common.sh
```diff
@@ -0,0 +1,83 @@
...
+# with cwd set to the checkout directory.
+
+# shellcheck disable=SC2034  # used by the sourcing pre-build/post-build hooks
+INT_TEST_HOME_DIR="$(pwd)/intTestHomeDir"
+
+# Homes to restore; on store we only pick up the ones that exist.
+INT_TEST_DISTRIBUTIONS="${INT_TEST_DISTRIBUTIONS:-full jvm basics native}"
+
+# The agent-wide hooks directory baked by dev-infrastructure.
+AGENT_HOOKS_DIR="${HOME}/agent/hooks"
```
cobexer: ❌  must be passed from the hooks plugin / agent

.teamcity/hooks/common.sh
```diff
@@ -0,0 +1,83 @@
...
+  if [[ "${BUILD_TYPE_ID:-}" == Gradle_Release* ]]; then
+    echo "release"
+  else
+    echo "master"
+  fi
+}
+
+# Image name for a distribution home, stable across builds of the same base branch.
+imageNameFor() {
+  echo "gradle-integtest-$(baseBranchName)-distributions-$1"
```
cobexer: ❓we should likely tie this to the `BUILD_TYPE_ID` maybe `${BUILD_TYPE_ID}-intTestHomeDir-$1`

.teamcity/hooks/common.sh
```diff
@@ -0,0 +1,83 @@
...
+    echo "Skipping artifact cache: not running on an EC2 instance."
+    return 1
+  fi
+  if [ ! -d "${AGENT_HOOKS_DIR}" ]; then
+    echo "Skipping artifact cache: agent hooks directory ${AGENT_HOOKS_DIR} not found."
+    return 1
+  fi
+  # Reuse the agent-wide common.sh (assertDevelocityServerUrl, ...).
+  # shellcheck source=/dev/null
+  source "${AGENT_HOOKS_DIR}/common.sh"
```
cobexer: ❌ let's not do that

cobexer (dismissed):
We need to abstract the interface between CI machines and repositories into something very small and much simpler to maintain.

A couple of things:
- there is far too much knowledge about AC/SC and the scripts from the agent in here.
  - make the interface something like:
  - `...some path.../universal-cache.sh --gradle [--store|--restore] <image> <gradle-home>`
- how the AC_CLI is executed (location, filenames, Java version, ...) must not be part of g/g as it's lifecycle is managed with the AMI, not here
- `post-build` and `pre-build` as separate files looks simple, but looks like it forces us to have ~3 scripts in any repo using that feature
  - should be improved to a single file `.teamcity/hooks/hooks.sh [pre|post]` -> removes need for `common.sh`

.teamcity/hooks/common.sh
blindpirate: I don't want to do that, because the real `BUILD_TYPE_ID` is `Gradle_Master_Check_SmokeTest_ConfigCacheAndroidProjectSmokeTests_virtual_Batch_1_1`. I'd keep it to `SmokeTest` for now.

cobexer: Is there an issue with a long image name?

blindpirate: Nope, but I don't like the `virtual_Batch_1_1` part. The batch might change over time.

cobexer: Let's confirm with DV: I'm pretty sure we should have a 1:1 relationship between what a shard does and the shard and not use one image across different build configurations (shards)

blindpirate:
@bot-gradle test

blindpirate:
@bot-gradle test

.teamcity/hooks/common.sh
blindpirate: I read the source code of universal cache implementation in DV repo, and decided to keep the same behavior -  "gradle-${BUILD_TYPE_ID}-$(baseBranchName)-distributions-$1"

blindpirate:
@bot-gradle squash and merge

Issue #38184, opened 2026-06-11
Failed IDE sync spends minutes in working state before giving feedback to the user


asodja:
### Summary

When an IDE sync (Tooling API `fetch` API) encounters a build failure, the daemon can spend a very long time (minutes of 100% CPU on a single thread) converting the failure into `InternalFailure` objects before the client even sees the error. The conversion is quadratic in the depth/size of the failure's cause tree, and with Isolated Projects the same conversion is repeated once per project per model, multiplying the cost further.

The user-visible symptom is a sync that appears to hang after something failed, with a thread dump showing the `Daemon worker` busy in `FailurePrinter`:

```
"Daemon worker" RUNNABLE
    at org.gradle.internal.problems.failure.FailurePrinter$Job.appendFrame(FailurePrinter.java:122)
    at org.gradle.internal.problems.failure.FailurePrinter$Job.appendFrames(FailurePrinter.java:105)
    at org.gradle.internal.problems.failure.FailurePrinter$Job.printRecursively(FailurePrinter.java:72)
    at org.gradle.internal.problems.failure.FailurePrinter.printToString(FailurePrinter.java:37)
    at org.gradle.internal.build.event.types.DefaultFailure.fromFailure(DefaultFailure.java:84)
    at org.gradle.internal.build.event.types.DefaultFailure.lambda$convertCausesToFailures$2(DefaultFailure.java:103)
    ...
    at org.gradle.internal.build.event.types.DefaultFailure.convertCausesToFailures(DefaultFailure.java:104)
    at org.gradle.internal.build.event.types.DefaultFailure.fromFailure(DefaultFailure.java:85)
    at org.gradle.tooling.internal.provider.runner.DefaultBuildController.toInternalFailures(DefaultBuildController.java:188)
    at org.gradle.tooling.internal.provider.runner.DefaultBuildController.fetch(DefaultBuildController.java:176)
    at com.intellij.gradle.toolingExtension.impl.modelAction.GradleModelControllerImpl.fetchModelOrNull(...)
```

### Root cause

There are two compounding problems.

**1. `DefaultFailure.fromFailure` re-prints the entire failure subtree at every recursion level — O(N²)**

[...]

PR #38224, opened 2026-06-16
Convert Tooling API model fetch failures in linear time and reuse shared conversions

eskatos:
Addresses the daemon side stall reported in 
* #38184.

## Why
An IDE that builds models through the resilient `fetch` API (`BuildController.fetch`, which returns a
model or failures result instead of throwing) makes the daemon convert every returned failure eagerly,
in `DefaultBuildController.fetch`. Converting one failure rendered its entire cause subtree again at
every node (quadratic in the cause chain depth), so even a single deep failure was expensive per fetch.
Under Isolated Projects this turns pathological: the IDE fetches a model per project, and with eager
configuration (configure-on-demand off) a single configuration failure is aggregated and attached to
every project's result, so that same failure is reconverted once per project. On a reproduction of the
reported scenario (100 failing projects) the sync pegged a daemon CPU core for about three minutes before
surfacing any error.

## What changes
The daemon side conversion of a source failure to the Tooling API `Failure` is now linear, and converted
failures are cached by the identity of the original throwable so a failure shared across per project
fetches is converted once. Each node now carries only its own text and the full `getDescription()` is
reassembled lazily, so the serialized payload per failure shrinks from O(N²) to O(N). The displayed text
is unchanged (`getDescription()`, `getMessage()`, and `getCauses()` behave exactly as before) and no
public API is added: the own text accessor lives only on the internal protocol.

## Implementation
`DefaultFailure.fromFailure` builds the failure tree in one pass, each node carrying only its own
description (`FailurePrinter.printNodeToString`); `getDescription()` reconstructs the full subtree text
lazily and memoized via `FailureDescriptionReconstructor`, which reapplies the parent relative stack
trace tail elision and preserves each line's original terminator. A build tree scoped `FetchFailureConverter`
holds a `Throwable` identity cache consulted at every recursion step, threaded through `DefaultBuildController`.
The own text accessor is added only to the protocol `InternalFailure.getOwnDescription()`, so the daemon
ships per node text instead of a full subtree per node; the consumer reconstructs over the concrete node
type and falls back to `getDescription()` on `NoSuchMethodError`/`AbstractMethodError` for an old daemon.
The public `org.gradle.tooling.Failure` is untouched.

## Performance
Per fetch conversion goes from O(N²) to O(N) in the cause chain depth N, the identity cache removes the
per project reconversion, and the per failure payload shrinks from O(N²) to O(N). In an end to end
reproduction against the parent commit: converting a deep failure across 100 independent per project
fetches drops from about 26 s to about 0.5 s at depth 160, and a 100 project failing Isolated Projects
sync drops from about three minutes to about 10 seconds. This removes the daemon side cost the thread
dump shows; the IDEs checked (IntelliJ, Buildship) render failures from per node message and stack plus
the cause structure rather than `getDescription()`, and IntelliJ's deep sync failures are currently
`Throwable` driven. One scalability limit is left untouched and out of scope here: under eager Isolated
Projects the configuration failure is aggregated and attached to every project, so the node count stays
quadratic in the project count, which configure-on-demand on already avoids.

## How to review
The three commits are tracer bullets and read well commit by commit: (1) linear conversion, (2) shared
failure dedup, (3) end to end cross version coverage. Suggested file order: `FailurePrinter` (single node
print), `DefaultFailure` (single pass build, lazy `getDescription`), `FailureDescriptionReconstructor`
(lazy full text rebuild), then `InternalFailure.getOwnDescription()` with the `BuildProgressListenerAdapter`
fallback, `FailureCache`/`FetchFailureConverter` with the `DefaultBuildController` wiring, and finally
`r970/ResilientFetchFailureCrossVersionSpec`. This touches no public API but does change the cross
version protocol, so validate with `./gradlew :tooling-api:forkingCrossVersionTest` alongside the
`:logging`, `:daemon-messaging`, and `:tooling-api` unit tests.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/daemon-messaging/src/main/java/org/gradle/internal/build/event/types/DefaultFailure.java
+++ b/platforms/core-runtime/daemon-messaging/src/main/java/org/gradle/internal/build/event/types/DefaultFailure.java
@@ -112,28 +112,42 @@ public class DefaultFailure implements Serializable, InternalFailure {
     }
 
     public static InternalFailure fromFailure(Failure buildFailure, Function<ProblemInternal, InternalBasicProblemDetailsVersion3> mapper) {
+        return fromFailure(buildFailure, mapper, FailureCache.NONE);
+    }
+
+    public static InternalFailure fromFailure(Failure buildFailure, Function<ProblemInternal, InternalBasicProblemDetailsVersion3> mapper, FailureCache cache) {
+        // Reuse the conversion of a throwable already seen, e.g. a configuration failure shared across per-project fetches.
+        InternalFailure cached = cache.get(buildFailure.getOriginal());
+        if (cached != null) {
+            return cached;
+        }
         // Build the failure tree in a single pass: each node carries only its own description. The full description (the
-        // whole cause subtree) is reconstructed lazily from these on demand, so no node prints its subtree here.
+        // whole cause subtree) is reconstructed lazily from these on demand, so no node prints its subtree here. Children
+        // are interned before the parent, so the cache is never re-entered while a single mapping is being computed.
         String ownDescription = FailurePrinter.printNodeToString(buildFailure);
-        List<InternalFailure> causeFailures = convertCausesToFailures(buildFailure.getCauses(), mapper);
+        List<InternalFailure> causeFailures = convertCausesToFailures(buildFailure.getCauses(), mapper, cache);
         List<InternalBasicProblemDetailsVersion3> problemDetails = buildFailure.getProblems().stream()
             .map(mapper)
             .collect(toList());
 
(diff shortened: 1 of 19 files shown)


eskatos:
@bot-gradle test RFN

eskatos:
Green APT https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_AllPerformanceTests_Trigger/114022761?expandBuildChangesSection=false

donat (approved):

PR #38231, opened 2026-06-17
Deprecate the public StartParameter constructor

reinsch82:
Stacked on top of #38152 (base branch `jvandort/ip/add-start-parameter-violations`).

## What

A `StartParameter` is always provided by the running build, which operates on a `StartParameterInternal`. A directly constructed `StartParameter` is a detached value object that no build ever sees, so offering a public constructor is misleading.

This deprecates the public no-arg `StartParameter()` constructor:

- **Compile-time**: `@Deprecated` + `@deprecated` Javadoc, pointing to `newInstance()` / `newBuild()`.
- **Runtime nag**: emits a deprecation warning when called directly by third parties.
- **Internal usage does not nag**: Gradle's own duplication (`newInstance()`/`newBuild()`, and `StartParameterInternal`) routes through the protected `StartParameter(BuildLayoutParameters)` constructor, so a real build never warns.

## Test migration

Tests that constructed `new StartParameter()` directly (which now nags, and throws in unit tests where `DeprecationLogger` is uninitialized) were updated:

- Collaborator usages switched to `StartParameterInternal` (the real runtime type).
- `StartParameterTest` keeps exercising the class itself, constructing via a `DeprecationLogger.whileDisabled { ... }` helper so it still validates the deprecated constructor's behaviour.

### DIFF
--- a/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
+++ b/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
@@ -64,7 +64,7 @@ import static java.util.Collections.emptyList;
  * <p>{@code StartParameter} defines the configuration used by a Gradle instance to execute a build. The properties of {@code StartParameter} generally correspond to the command-line options of
  * Gradle.
  *
- * <p>You can obtain an instance of a {@code StartParameter} by duplicating an existing one using {@link #newInstance} or {@link #newBuild}.</p>
+ * <p>The {@code StartParameter} for a build is provided by Gradle, e.g. via {@link org.gradle.api.invocation.Gradle#getStartParameter()}. It is not intended to be created or copied directly.</p>
  */
 @HasInternalProtocol
 public class StartParameter implements LoggingConfiguration, ParallelismConfiguration, Serializable {
@@ -254,14 +254,22 @@ public class StartParameter implements LoggingConfiguration, ParallelismConfigur
     /**
      * Creates a {@code StartParameter} with default values. This is roughly equivalent to running Gradle on the command-line with no arguments.
      *
-     * @deprecated A {@code StartParameter} is provided by the build (e.g. {@code gradle.startParameter}) and should not be created directly. Derive a copy from that one with {@link #newInstance} or {@link #newBuild} if you need to.
+     * @deprecated The {@code StartParameter} for a build is provided by Gradle (e.g. {@code gradle.startParameter}) and is not intended to be created directly.
      */
     @Deprecated
     public StartParameter() {
-        this(new BuildLayoutParameters());
-        // Only direct construction reaches here; Gradle's own duplication goes through the protected
-        // constructor, so this nags third-party callers without nagging internal usage.
+        this((Void) null);
+        // Only direct construction reaches here; Gradle's own construction goes through the protected
+        // constructors, so this nags third-party callers without nagging internal usage.
         StartParameterDeprecations.nagOnStartParameterConstructor();
+    }
+
+    /**
+     * Creates a {@code StartParameter} with default values, for internal use. The {@code Void} parameter
+     * exists only to distinguish this from the deprecated public no-arg constructor; pass {@code null}.
+     */
+    protected StartParameter(@SuppressWarnings("unused") Void marker) {
+        this(new BuildLayoutParameters());
     }
 
     /**
@@ -282,9 +290,12 @@ public class StartParameter implements LoggingConfiguration, ParallelismConfigur
      * Duplicates this {@code StartParameter} instance.
      *
      * @return the new parameters.
+     * @deprecated The {@code StartParameter} for a build is provided by Gradle and is not intended to be copied.
      */
+    @Deprecated
     public StartParameter newInstance() {
-        return prepareNewInstance(new StartParameter(new BuildLayoutParameters()));
+        StartParameterDeprecations.nagOnStartParameterCopy("newInstance()");
+        return prepareNewInstance(new StartParameter((Void) null));
     }
 
     protected StartParameter prepareNewInstance(StartParameter p) {
@@ -310,9 +321,12 @@ public class StartParameter implements LoggingConfiguration, ParallelismConfigur
      * build specific properties (eg task names).</p>
      *
      * @return The new parameters.
+     * @deprecated The {@code StartParameter} for a build is provided by Gradle and is not intended to be copied.
      */
+    @Deprecated
     public StartParameter newBuild() {
-        return prepareNewBuild(new StartParameter(new BuildLayoutParameters()));
+        StartParameterDeprecations.nagOnStartParameterCopy("newBuild()");
+        return prepareNewBuild(new StartParameter((Void) null));
     }
 
     protected StartParameter prepareNewBuild(StartParameter p) {
--- a/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
+++ b/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
@@ -64,7 +64,7 @@ import static java.util.Collections.emptyList;
  * <p>{@code StartParameter} defines the configuration used by a Gradle instance to execute a build. The properties of {@code StartParameter} generally correspond to the command-line options of
  * Gradle.
  *
- * <p>You can obtain an instance of a {@code StartParameter} by either creating a new one, or duplicating an existing one using {@link #newInstance} or {@link #newBuild}.</p>
+ * <p>You can obtain an instance of a {@code StartParameter} by duplicating an existing one using {@link #newInstance} or {@link #newBuild}.</p>
  */
 @HasInternalProtocol
 public class StartParameter implements LoggingConfiguration, ParallelismConfiguration, Serializable {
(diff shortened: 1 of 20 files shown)


platforms/core-runtime/start-parameter/src/main/java/org/gradle/internal/deprecation/StartParameterDeprecations.java
```diff
@@ -20,6 +20,14 @@
 
 public class StartParameterDeprecations {
 
+    public static void nagOnStartParameterConstructor() {
+        DeprecationLogger.deprecate("Creating a StartParameter instance directly")
+            .withAdvice("A StartParameter is provided by the build; obtain it (e.g. 'gradle.startParameter') and derive a copy with newInstance() or newBuild() if you need one.")
+            .willBeRemovedInGradle10()
+            .undocumented()
```
alllex: ❌ Requires an upgrade guide entry

platforms/core-runtime/start-parameter/src/main/java/org/gradle/internal/deprecation/StartParameterDeprecations.java
```diff
@@ -20,6 +20,14 @@
 
 public class StartParameterDeprecations {
 
+    public static void nagOnStartParameterConstructor() {
+        DeprecationLogger.deprecate("Creating a StartParameter instance directly")
+            .withAdvice("A StartParameter is provided by the build; obtain it (e.g. 'gradle.startParameter') and derive a copy with newInstance() or newBuild() if you need one.")
```
alllex: ❔ Do we know of a valid use case where it's useful to derive a copy of the StartParameter?

Unless we know of one, I'd not suggest a replacement.

It seems like we repeat this advice in multiple javadocs as well. If we decide to update one place, we should align others

platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
```diff
@@ -304,7 +311,7 @@ protected StartParameter prepareNewInstance(StartParameter p) {
      * @return The new parameters.
      */
     public StartParameter newBuild() {
```
alllex, line 323: ❓ If the intention is to close down some methods that unintentionally became part of the public API, shouldn't we also deprecate the `newInstance()` and `newBuild()` ?  For our internal purposes, we can have similar methods on `StartParameterInternal`.

platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
```diff
@@ -276,7 +283,7 @@ protected StartParameter(BuildLayoutParameters layoutParameters) {
      * @return the new parameters.
      */
     public StartParameter newInstance() {
-        return prepareNewInstance(new StartParameter());
+        return prepareNewInstance(new StartParameter(new BuildLayoutParameters()));
```
alllex: 🤔 We are now duplicating the layout parameter creation in many places. We can't use a different name, because it's a constructor, but we can create a protected constructor with a sentinel argument that does nothing, e.g. `protected StartParameter(Void marker)`. That constructor will keep centralized the logic of creating a "parameterless" start parameter.

lkasso (approved):
Docs LGTM.

platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
```diff
@@ -64,7 +64,7 @@
  * <p>{@code StartParameter} defines the configuration used by a Gradle instance to execute a build. The properties of {@code StartParameter} generally correspond to the command-line options of
  * Gradle.
  *
- * <p>You can obtain an instance of a {@code StartParameter} by either creating a new one, or duplicating an existing one using {@link #newInstance} or {@link #newBuild}.</p>
+ * <p>The {@code StartParameter} for a build is provided by Gradle, e.g. via {@link org.gradle.api.invocation.Gradle#getStartParameter()}. It is not intended to be created or copied directly.</p>
```
jvandort, line 67:
```suggestion
 * <p>The {@code StartParameter} for a build can be acquired via {@link org.gradle.api.invocation.Gradle#getStartParameter()}. It is not intended to be created or copied directly.</p>
```

platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
```diff
@@ -252,8 +253,22 @@ public File getProjectCacheDir() {
 
     /**
      * Creates a {@code StartParameter} with default values. This is roughly equivalent to running Gradle on the command-line with no arguments.
+     *
+     * @deprecated The {@code StartParameter} for a build is provided by Gradle (e.g. {@code gradle.startParameter}) and is not intended to be created directly.
```
jvandort, line 257:
```suggestion
     * @deprecated This constructor will be removed in Gradle 10. Please use {@link org.gradle.api.invocation.Gradle#getStartParameter()} instead.
```

platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
```diff
@@ -252,8 +253,22 @@ public File getProjectCacheDir() {
...
+        // Only direct construction reaches here; Gradle's own construction goes through the protected
+        // constructors, so this nags third-party callers without nagging internal usage.
+        StartParameterDeprecations.nagOnStartParameterConstructor();
+    }
+
+    /**
+     * Creates a {@code StartParameter} with default values, for internal use. The {@code Void} parameter
+     * exists only to distinguish this from the deprecated public no-arg constructor; pass {@code null}.
+     */
+    protected StartParameter(@SuppressWarnings("unused") Void marker) {
```
jvandort: Do we need to add this constructor? `StartParmaterInternal` already calls the existing `StartParameter(BuildLayoutParameters)` constructor. Can't we do that too in the deprecated no-arg constructor? 

Alternatively, if we leave this constructor in, we should call this void constructor in StartParameterInternal's no-arg constructor.

platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
```diff
@@ -275,9 +290,12 @@ protected StartParameter(BuildLayoutParameters layoutParameters) {
      * Duplicates this {@code StartParameter} instance.
      *
      * @return the new parameters.
+     * @deprecated The {@code StartParameter} for a build is provided by Gradle and is not intended to be copied.
```
jvandort, line 285:
```suggestion
     * @deprecated This method will be removed in Gradle 10. Please use {@link org.gradle.api.invocation.Gradle#getStartParameter()} instead to get an existing instance.
```

platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
```diff
@@ -303,9 +321,12 @@ protected StartParameter prepareNewInstance(StartParameter p) {
      * build specific properties (eg task names).</p>
      *
      * @return The new parameters.
+     * @deprecated The {@code StartParameter} for a build is provided by Gradle and is not intended to be copied.
```
jvandort, line 320:
```suggestion
     * @deprecated This method will be removed in Gradle 10. Please use {@link org.gradle.api.invocation.Gradle#getStartParameter()} instead to get an existing instance.
```

platforms/core-runtime/start-parameter/src/test/groovy/org/gradle/StartParameterTest.groovy
```diff
@@ -313,12 +334,26 @@ class StartParameterTest extends Specification {
         assert parameter.taskRequests.size() == 1 && parameter.taskRequests[0] instanceof RunDefaultTasksExecutionRequest
     }
 
+    // The public constructor and the copy methods are deprecated; this test class exercises
```
jvandort: I don't think there is much value at all in this test running against the non-internal StartParameter. I think we should update this test to execute against `StartParemeterInternal`, as that is the actual type that we use at runtime. 

In other words, I think its much more valuable that these tests execute against `StartParameterInternal` rather than the public `StartParameter`. 

Note, thats not to say we can't sprinkle in a couple StartParameter tests that still verify the deprecations for the constructor and newBuild/newInstance (the ones we have at the top of the file)

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -195,6 +195,20 @@ To fix it, use a standard collection or map type instead of a custom subtype, an
...
+The `StartParameter` for a build is provided by Gradle and is accessed through `Gradle.getStartParameter()`.
+It was never intended to be instantiated or copied by build logic, but the public no-argument constructor and the `newInstance()` / `newBuild()` methods made that possible.
+
+These are now deprecated and will be removed in Gradle 10:
+
+- the public `StartParameter()` constructor;
+- `StartParameter.newInstance()`;
+- `StartParameter.newBuild()`.
+
+Obtain the start parameter from the build (for example via `gradle.startParameter`) instead of constructing or copying one.
```
```suggestion

Starting in Gradle 10, constructing new instances of `StartParameter` will no longer be permitted.

The following have been deprecated for removal:

- The `StartParameter()` constructor
- `StartParameter.newInstance()`
- `StartParameter.newBuild()`

The existing `StartParameter` instance for a build may be obtained via the `Gradle.getStartParamter()` API.
```

jvandort: I think some links to the javadoc should also be included in addition to just wrapping class/method names in ``

jvandort (approved):
LGTM mostly some wording suggestions and thoughts on testing.

PR #38237, opened 2026-06-17
Address review feedback on StartParameter Isolated Projects violations

### LINKED ISSUE #37944: Handling `StartParameter` state with Isolated Projects

reinsch82:
Follow-up to #38152 (merged), addressing @alllex's review comments.

## Changes (one concern per commit)

- **Scope start-parameter mutation reporting to Isolated Projects** — the listener that detects post-settings mutation was installed for every build regardless of IP. Moved it out of the general `SettingsEvaluatedCallbackFiringSettingsProcessor` (restoring its single responsibility) into a dedicated `StartParameterMutationReportingSettingsProcessor` added to the settings-processing chain only when IP is enabled. Also documents `setMutationListener`/`clearMutationListener`, and generalises the `Intercepting*` serialization comment (and adds the missing one to `InterceptingMap`).
- **Add an explicit allowed/forbidden contract test** — one data-driven test stating the contract for accessing `StartParameter` from a build script after settings evaluation: reads and the exempt task-name replacement are allowed; every mutation (setter or collection view) is reported.
- **Clarify the violation message** — leads with the rule ("the start parameter cannot be mutated after settings have been evaluated…") so it conveys the mental model rather than a mechanical per-call report; the offending call is supplementary and emitted via `StructuredMessage.reference()`.
- **Restore start-parameter task-name serialization for CC-hit parity** — the cached work graph drives what executes on a cache hit, so this isn't needed to run the build; it keeps `gradle.startParameter.taskNames` consistent on a hit for internal consumers (build operations, problem reports, build scans), since scheduling doesn't run on a hit.

## Review comments addressed
`StartParameterInternal` listener Javadoc; install listener only under IP; generalise serialization comment; explicit allowed/forbidden contract test; mutation-message mental model + `reference()`; restore CC task-name serialization (with rationale, replacing the stale inline comment).

## Not included
- The listener-cleanup lifecycle redesign (root vs included-build clear / cross-module coupling) — there is no functional bug (the root clear targets the correct instance and is needed for continuous builds; included builds are recreated fresh per iteration), so it's tracked separately as a clarity/robustness change.
- Broader "IP violation messages are too mechanical" UX, which @alllex noted is a general concern for a separate effort.


platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -18,6 +18,42 @@ package org.gradle.internal.cc.impl.isolated
 
 class IsolatedProjectsStartParameterIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
 
+    def "StartParameter access from a build script after settings evaluation honors the IP contract (#description)"() {
```
alllex: 🤔 This test seems like 3 tests in a trench coat. In my previous comment, I didn't actually mean we have to have a single test to express the contract. Even though, I can see how it can be interpreted that way. Sorry about that.

Also see https://github.com/gradle/gradle/pull/38152#discussion_r3435914319

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
```diff
@@ -703,10 +705,26 @@ class ConfigurationCacheState(
...
+    private
+    fun WriteContext.writeStartParameterOf(gradle: GradleInternal) {
+        writeStrings(gradle.startParameter.taskNames)
+    }
+
+    private
+    fun ReadContext.readStartParameterOf(gradle: GradleInternal) {
+        // Restore the requested task names captured during configuration. The cached work graph already
+        // drives what executes on a cache hit, so this is not needed to run the build. It keeps
+        // gradle.startParameter.taskNames consistent on a hit for internal consumers that read it (build
```
alllex, line 722: 💭 We should add a test for this to prove the behavior is actually observable. The lack of such a test allowed us to remove the code without the feedback from the test suite

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -40,7 +76,7 @@ class IsolatedProjectsStartParameterIntegrationTest extends AbstractIsolatedProj
         then:
         fixture.assertIsolatedProjectsProblems(mode) {
             projectsConfigured(*configured)
-            problem("Build file '${scriptPath}': line 2: Cannot call 'setDryRun(boolean)' on StartParameter after settings have been evaluated when Isolated Projects is enabled.")
+            problem("Build file '${scriptPath}': line 2: The start parameter cannot be mutated after settings have been evaluated when Isolated Projects is enabled. This happened when calling 'setDryRun(boolean)'.")
```
alllex: 💅 `The mutation occurred via 'setDryRun(boolean)' call`

alllex (changes requested):

subprojects/core/src/main/java/org/gradle/initialization/internal/settings/StartParameterMutationReportingSettingsProcessor.java
```diff
@@ -0,0 +1,62 @@
...
+ * is enabled, since that is the only mode in which such mutations are reported.
+ */
+@NullMarked
+public class StartParameterMutationReportingSettingsProcessor implements SettingsProcessor {
+
+    private final SettingsProcessor delegate;
+    private final IsolatedProjectsProblemsReporter problems;
+
+    public StartParameterMutationReportingSettingsProcessor(SettingsProcessor delegate, IsolatedProjectsProblemsReporter problems) {
+        this.delegate = delegate;
```
jvandort, line 41: We have similar delegate chains elsewhere and I am not much of a fan. I don't really see the value in this change and I think it makes things harder to follow. Was there a specific reason we went this route?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
jvandort: We got a CC cache hit though, right? Shouldn't the SP task names already be correct? 

There is also something strange about this code. In vintage, we never set the task names for included builds. Included builds (which have their own start parameter instances) always have empty task names (unless the user set them). I find it strange that we are setting the task names in CC but not in vintage. 

The original commit that added this to CC was here:
https://github.com/gradle/gradle/commit/828bccf6a8e8fa2372b26269b394df5944255572

It seems like it was specifically added to ensure we restore user-made calls to `setTaskNames` for included builds. The test which originally required this sort of behavior had the `setTaskNames` removed here: https://github.com/gradle/gradle/commit/8354ba5f230fe7aab0979ef1221e9c65c5647f73#diff-76e1cc030f95aaff23753af9eb65b9bdb1c98e674107178b4a32fa9f039f922b

setTaskNames was part of the original issue where we added this test:
https://github.com/gradle/gradle/commit/9f1a2b6634ed6cf1695e2f81a92b07f74c8851b9

Given this, I would say the comment on this code in this PR is incorrect. We are calling `setTaskNames` in order to restore user-configured task names on included builds. Now, I am not sure where these values are ever read, and the fact that we need to restore these values are suspicious.

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingCollection.java
```diff
@@ -69,8 +69,8 @@ private E view(E element) {
         return elementView.apply(element);
     }
 
-    // Serialize as the delegate: the interceptor is runtime-only wiring and instances may be captured
-    // in task state and stored to the configuration cache.
+    // These decorators serialize as their plain delegate: the interceptor is transient runtime wiring,
+    // so a decorated collection captured in serializable state round-trips as the underlying collection.
     protected Object writeReplace() {
```
jvandort, line 57: Why did we add these in the first place?

Was it because the `AccessTracking` sets did? If we are no longer using this collection to back those `AccessTracking` collections, I would prefer to leave this serialization code out until the use case requires it again.

subprojects/core/src/main/java/org/gradle/initialization/internal/settings/StartParameterMutationReportingSettingsProcessor.java
reinsch82: It was the response to 
https://github.com/gradle/gradle/pull/38152#discussion_r3426594610

I'd rather have structural decisions than runtime decisions, but I guess you could argue, this decision doesn't happen very often, so "who cares?"

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingCollection.java
reinsch82: We want to serialize the StartParameter object but without the callbacks.

writeReplace is the supported way to serialize just the delegate and not the actual object.
https://github.com/gradle/gradle/blob/ed7ac521cd4ece3464c2b1bc7d17dcb35a3ead20/platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/jos/JavaSerializationEncodingLookup.kt#L82

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
reinsch82: Here is the test
https://github.com/gradle/gradle/blob/f02fa15afb10829e9c1f96c9b3f9d79c7e83167f/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheStartParameterIntegrationTest.groovy#L25

this would fail if we would not serialize it.
We can argue about whether this is good or bad, but it is a behavior change.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
reinsch82: enough coverage now elsewhere, removed.

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingMap.java
```diff
@@ -115,9 +121,78 @@ public Collection<V> values() {
...
+                }
+
+                @Override
+                public Entry<K, V> next() {
+                    return wrap(base.next());
+                }
+
+                @Override
+                public void remove() {
+                    base.remove();
```
alllex, line 174: ❓ Does this report the mutation of the underlying map?

Would it be useful to create an `InterceptingIterator` abstraction? There is overlap with the anonymous iterator class in the `InterceptingCollection`.

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingMap.java
```diff
@@ -115,9 +121,78 @@ public Collection<V> values() {
...
+        }
+
+        @Override
+        @SuppressWarnings("SuspiciousToArrayCall")
+        public <T> T[] toArray(T[] a) {
+            return wrapAll(super.toArray(a));
+        }
+
+        @SuppressWarnings("unchecked")
+        private <T> T[] wrapAll(T[] array) {
```
alllex: 💭 The implementation is straight-forward, but the method name doesn't convey that the argument array is modified in-place.

Should we move this helper closer to the `wrap` method?

Also, can an entry ever be null in this array?

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingCollection.java
```diff
@@ -48,29 +46,14 @@ public interface Interceptor {
 
     protected final C delegate;
     final transient Interceptor interceptor;
-    private final transient Function<E, E> elementView;
-    private final transient boolean wrapsElements;
```
alllex, line 49: 👍 This is a good reduction of complexity for a generic collection

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingCollection.java
alllex: I wonder if the decision to throw away the interceptors should live in the CC codecs, and not be co-located with the intercepting collections themselves. Even writing the comment on this method creates the tension between where the decision lives.

subprojects/core/src/main/java/org/gradle/initialization/internal/settings/StartParameterMutationReportingSettingsProcessor.java
```diff
@@ -0,0 +1,62 @@
...
+        this.problems = problems;
+    }
+
+    @Override
+    public SettingsState process(GradleInternal gradle, SettingsLocation settingsLocation, ClassLoaderScope buildRootClassLoaderScope, StartParameterInternal startParameter) {
+        SettingsState state = delegate.process(gradle, settingsLocation, buildRootClassLoaderScope, startParameter);
+        startParameter.setMutationListener(methodSignature ->
```
alllex, line 51: ❓ Is it important that the mutation reporter is registered _after_ the settings state is created? If so, should we explain why not _before_ in a short comment?

subprojects/core/src/main/java/org/gradle/internal/service/scopes/BuildScopeServices.java
```diff
@@ -627,26 +628,30 @@ protected SettingsProcessor createSettingsProcessor(
...
+                ),
+                gradleProperties,
+                textFileResourceLoader,
+                buildIncludeListener
+            )
+        );
+        if (buildModelParameters.isIsolatedProjects()) {
+            // Mutating the start parameter after settings evaluation is only reported under Isolated
+            // Projects, so the reporting layer is added to the chain only then.
+            settingsProcessor = new StartParameterMutationReportingSettingsProcessor(settingsProcessor, isolatedProjectsProblemsReporter);
```
alllex: ❔ Does this comment add value, given the `if` just above?

platforms/core-runtime/start-parameter/src/main/java/org/gradle/api/internal/StartParameterInternal.java
```diff
@@ -74,13 +74,23 @@ protected StartParameterInternal(BuildLayoutParameters layoutParameters) {
         super(layoutParameters);
     }
 
+    /**
+     * Arms the listener invoked on every subsequent mutating call (see {@link #onMutableCall}). The
+     * build tree installs it once settings have been evaluated, so that later mutations of the running
+     * build's start parameter can be reported as Isolated Projects violations. Throws if a listener is
+     * already set, to catch a stale one surviving from an earlier build into a reused start parameter.
+     */
     public void setMutationListener(Consumer<String> mutationListener) {
```
alllex, line 83: 💭 `Arms` is an interesting choice of word. Not sure what I think about it, but it is surprising. FTR, the "arms" wording is used in multiple places, so if we decide to change it, we should be consistent.

> build tree installs

Is it actually build-tree? Aren't we doing that per included build?

> Throws if a listener is
> already set, to catch a stale one surviving from an earlier build into a reused start parameter.

This is weirdly specific. I can't put my finger on it, but it sounds strange. Can we instead state the contract? Something like "There can be only one listener"

alllex (commented):
I haven't reviewed everything yet, but here are some questions and suggestions

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingMap.java
reinsch82: A common base class would implement only `hasNext`, I don't see this as enough benefit currently to actually do this.

this calls 
`InterceptingCollection.iterator()` which already handles the interceptor call of the underlying `Set`

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingMap.java
reinsch82: > Should we move this helper closer to the wrap method?

can do.

> Also, can an entry ever be null in this array?

with the code as it is yes, but I found a slightly nicer way, that solves this.

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingCollection.java
reinsch82: I think the delegating collection can and should decide what exactly to serialize.
It knows that it delegates stuff to an actual data containing class and it knows it has runtime state we don't want to serialize. I don't see a really strong argument to put this into the serialization. this way it can also be serialized from other serializing mechanism correctly. It has to have some knowledge on it's serialization properties either way e.g. by using transient and being `Serializable`.
this is also common practice in `guava` and the JDK itself.

platforms/core-runtime/collections/src/main/java/org/gradle/internal/collect/InterceptingMap.java
alllex: You are right, I missed the bit where we call the super iterator

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,60 @@
...
+
+import org.gradle.integtests.fixtures.configurationcache.ConfigurationCacheFixture
+
+class ConfigurationCacheStartParameterIntegrationTest extends AbstractConfigurationCacheIntegrationTest {
+
+    ConfigurationCacheFixture fixture = new ConfigurationCacheFixture(this)
+
+    def "resolved default task names are restored on the build start parameter for a configuration cache hit"() {
+        given:
+        // On the store run the scheduler resolves the configured default tasks into
```
alllex: 🤔 What is the "scheduler" in this context and why do we have to mention it here, in an integration test?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,60 @@
...
+        // startParameter.taskNames. On a cache hit the scheduler does not run, so without restoring the
+        // captured names an execution-time read of the build's start parameter would observe an empty
+        // list. The task reads the live build start parameter (injected, not via project access).
+        buildFile """
+            import javax.inject.Inject
+            import org.gradle.StartParameter
+
```
alllex: 💅 Are the imports necessary?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheStartParameterIntegrationTest.groovy
```diff
@@ -0,0 +1,60 @@
...
+            abstract class PrintRequestedTasks extends DefaultTask {
+                @Inject abstract StartParameter getStartParameter()
+                @TaskAction void printIt() {
+                    println("REQUESTED=" + getStartParameter().taskNames)
+                }
+            }
+
+            defaultTasks 'printRequestedTasks'
```
alllex, line 35: ❔ Is this detail important for the test?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -213,6 +213,117 @@ class IsolatedProjectsStartParameterIntegrationTest extends AbstractIsolatedProj
         "projectProperties.keySet().with { def i = it.iterator(); i.next(); i.remove() }" | "getProjectProperties().keySet().iterator().remove()"
         "projectProperties.entrySet().find { it.key == 'seed' }.setValue('x')"            | "getProjectProperties().entrySet().Entry.setValue(Object)"
         "excludedTaskNames.removeIf { it == 'x' }"                                        | "getExcludedTaskNames().removeIf(Predicate)"
+        "systemPropertiesArgs.put('p', 'v')"                                              | "getSystemPropertiesArgs().put(Object, Object)"
+        "writeDependencyVerifications.add('x')"                                           | "getWriteDependencyVerifications().add(Object)"
+        "lockedDependenciesToUpdate.add('x')"                                             | "getLockedDependenciesToUpdate().add(Object)"
+        // setTaskNames/setTaskRequests are exempt because tooling legitimately replaces the requested
```
alllex: 🤔 It's unclear what "tooling" refers to here

Another confusing part is that the test checks the mutation on the obtained collections, while the comment refers to setters

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -213,6 +213,117 @@ class IsolatedProjectsStartParameterIntegrationTest extends AbstractIsolatedProj
...
     }
 
+    // Exhaustive end-to-end coverage: every notifying mutator of StartParameter and
+    // StartParameterInternal is exercised through a real build and asserted to be reported, so a change
+    // in behavior for any single method is caught here, not only by the instrumentation unit test. One
+    // build per method keeps every case independent and the assertion trivial (a single problem).
+    // setTaskNames/setTaskRequests are intentionally absent: they are exempt (see "replacing the
+    // requested tasks ..."). Each invocation uses a value matching the current/default state so the
+    // build still completes configuration in diagnostics mode.
+    def "every mutating method of StartParameter is reported (#signature)"() {
```
alllex: 🤔 This reads like an excerpt from an LLM chat. _A lot_ of it is implied by the fact that this is an integ test, and by the way the test is written. So a lot of the comment is redundant. 

The mention of `setTaskNames/setTaskRequests` might be useful, but is too far away from where the lists are declared. The test itself doesn't mention them. Maybe worth moving this bit of into into the list of methods.

> Each invocation uses a value matching the current/default state so the build still completes configuration in diagnostics mode.

I don't understand this part. Why is diagnostics mode important?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsStartParameterIntegrationTest.groovy
```diff
@@ -213,6 +213,117 @@ class IsolatedProjectsStartParameterIntegrationTest extends AbstractIsolatedProj
...
+        buildFile("""
+            gradle.startParameter.$invocation
+        """)
+
+        when:
+        isolatedProjectsDiagnosticsFails("help")
```
alllex: ❔ Why is this diagnostic and not default mode?

platforms/core-runtime/start-parameter/src/test/groovy/org/gradle/api/internal/StartParameterInternalMutationInstrumentationTest.groovy
```diff
@@ -0,0 +1,318 @@
...
+
+    def "exempt mutators do not notify"() {
+        given:
+        def notified = []
+        def parameter = new StartParameterInternal()
+        parameter.setMutationListener { notified << it }
+        def bySignature = declaredMethods().collectEntries { [signature(it), it] }
+
+        expect:
+        ["setTaskNames(Iterable)", "setTaskRequests(Iterable)"].each { sig ->
```
alllex: 💅 To avoid potential drift, it's better to refer to one of the method _lists_ here, since we already have them

platforms/core-runtime/start-parameter/src/test/groovy/org/gradle/api/internal/StartParameterInternalMutationInstrumentationTest.groovy
```diff
@@ -0,0 +1,318 @@
...
+
+    // Mutators that deliberately do not notify, plus the listener-wiring infrastructure.
+    private static final List<String> NOT_NOTIFYING = [
+        // Tooling model builders legitimately replace the requested tasks while the build runs.
+        "setTaskNames(Iterable)",
+        "setTaskRequests(Iterable)",
+        // Listener wiring, not tracked state.
+        "setMutationListener(Consumer)",
+        "clearMutationListener()",
+    ]
```
alllex, line 126: 💅 These methods are obviously special, so maybe they can live directly inside the `every method of StartParameter(Internal) is classified exactly once` test?

platforms/core-runtime/start-parameter/src/test/groovy/org/gradle/api/internal/StartParameterInternalMutationInstrumentationTest.groovy
```diff
@@ -0,0 +1,318 @@
...
+
+    // Getters returning intercepting collections; mutating the returned view notifies through the view.
+    private static final List<String> INTERCEPTING_COLLECTIONS = [
```
alllex, line 129: 💅 Any reason not to add a test that sanity checks this as well, e.g. with `.clear()` for instance?

platforms/core-runtime/start-parameter/src/test/groovy/org/gradle/api/internal/StartParameterInternalMutationInstrumentationTest.groovy
```diff
@@ -0,0 +1,318 @@
...
+        "getExcludedTaskNames()",
+        "getProjectProperties()",
+        "getSystemPropertiesArgs()",
+        "getWriteDependencyVerifications()",
+        "getLockedDependenciesToUpdate()",
+    ]
+
+    // Reads and other non-mutating methods (queries, copies, Object overrides).
+    private static final List<String> READS = [
+        "equals(Object)",
```
alllex, line 140: 💅 Any reason not to add a test that sanity checks this as well?

alllex (approved):
LGTM! The test coverage is much better now.

There are a lot of small things that I think can be improved. I trust they can be addressed without another review round.

One thing that makes me sad is the rise of the wall-of-text comments. A lot of them don't have a good return on interest, when it comes to reading them. It's very easy to produce these comments these days, but we should really strive to keep them minimal.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheStartParameterIntegrationTest.groovy
reinsch82: it's a minor optimization to no have to add the task name parameter to the `configurationCacheRun()` calls

alllex: In that case, I'd rather we have an explicit task in the run calls, as we do for the majority of our tests. Here it's especially non-trivial, given that we are asserting what ends up being in the startParameter.tasknames

reinsch82: I double checked the defaulttask test, it fails when we remove the taskname deserialization.

if I don't use the defaulttask it succeeds either way.
So this is the guarantee to not remove the deserialization.

alllex: That is... very obscure. I think this deserves a comment then.

subprojects/core/src/main/java/org/gradle/initialization/internal/settings/StartParameterMutationReportingSettingsProcessor.java
jvandort: Yeah this happens hardly ever. So who cares?

This code is definitely more complicated than the other. 

I definitely prefer the older code

PR #38251, opened 2026-06-19
Use bounded location captures under `--warning-mode=all` and `fail`

eskatos:
### Why
Follow-up to #38127, implementing the bounded-capture variant proposed in #32362. `--warning-mode=all` and `fail` captured a full stack trace for every problem so each gets a source location. On a very large build, particularly an IDE sync that re-fires the same deprecations tens of thousands of times, that does not scale.

### What changes
`all` and `fail` now keep the full-stack cap (50, or 5000 under Isolated Projects) and give every problem past it a cheap bounded location with no upper limit, rather than a full stack for all. Every problem still resolves to a location; only the first 50 carry a complete trace. The capped modes (`summary`, `none`) are unchanged, and there is no public API change.

### Implementation
The capture policy and its per-stream budgets, previously inlined in the diagnostics factory, move into a per-stream `StackTraceCapturer`; `all` and `fail` differ from the default stream only by lifting the bounded budget. The capturer offers a caller-derived capture and a supplied-throwable capture as distinct operations, so the bounded fallback can never substitute for a supplied throwable.

### Performance
Bounded and negligible. Lifting only the bounded budget, not the full-stack cap, means each problem past the cap pays the cheap bounded walk from #38127 instead of a full stack, so cost stays flat as build size grows. Covered by the existing `--warning-mode=all` scenario in `DeprecationCreationPerformanceTest`.

Before that, using `--warning-mode=all` on a build with a large number of projects emitting problems was very costly, crawling to a halt in extreme cases.

### Documentation
The #38127 release note already presents `--warning-mode=all` as giving every problem a location through the cheaper capture; this adds one sentence clarifying that past the cap the trace is reduced to the originating build logic and calling script, not the full call chain.

### How to review
Best read commit by commit: the behavior-preserving capturer extraction, the budget change that makes `all` and `fail` use bounded captures past the cap, the release-note clarification, and the split into caller and supplied captures. `ProblemLocationPastCapIntegrationTest` (over `summary` and `all`) and the `all`/`fail` case in `DeprecatedUsageBuildOperationProgressIntegrationTest` pin the bounded path end to end.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/logging/src/integTest/groovy/org/gradle/DeprecatedUsageBuildOperationProgressIntegrationTest.groovy
+++ b/platforms/core-runtime/logging/src/integTest/groovy/org/gradle/DeprecatedUsageBuildOperationProgressIntegrationTest.groovy
@@ -349,7 +349,8 @@ class DeprecatedUsageBuildOperationProgressIntegrationTest extends AbstractInteg
         mode << [WarningMode.None, WarningMode.Summary]
     }
 
-    def "collects stack traces for deprecation usages without limit for warning mode #mode"() {
+    @ToBeFixedForIsolatedProjects(skipBecause = "Isolated Projects raises the cap to 5000, so all 100 stay full and never reach the bounded path; IP also emits extra deprecation events that make the count unstable")
+    def "collects full stack traces up to the cap then bounded captures past it for warning mode #mode"() {
         file('settings.gradle') << "rootProject.name = 'root'"
 
         100.times {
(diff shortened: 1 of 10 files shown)


eskatos:
@bot-gradle test RFN

lkasso (approved):
I updated the existing docs to reflect this change.

eskatos:
@bot-gradle test RFN

eskatos:
@bot-gradle test RFN

donat (approved):

PR #38252, opened 2026-06-19
Report deprecated API on Gradle

6hundreds:
Part of https://github.com/gradle/gradle/issues/37352
Report a deprecated region of `Gradle` API, as technically it's same as add a listener, which started to be reported just recently

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiInvocationValidationIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiInvocationValidationIntegrationTest.groovy
@@ -124,26 +124,24 @@ class IsolatedProjectsToolingApiInvocationValidationIntegrationTest extends Abst
         """
 
         when:
-        withIsolatedProjectsDiagnostics()
+        withIsolatedProjects()
         fetchModelFails()
 
         then:
         fixture.assertModelStoredAndDiscarded {
             projectConfigured(":buildSrc")
             modelsCreated(":")
-            problem("Build file 'build.gradle': line 3: Project ':' cannot access Gradle.buildFinished")
             problem("Build file 'build.gradle': line 3: registration of listener on 'Gradle.buildFinished' is unsupported")
         }
 
         when:
-        withIsolatedProjectsDiagnostics()
+        withIsolatedProjects()
         fetchModelFails()
 
         then:
         fixture.assertModelStoredAndDiscarded {
             projectConfigured(":buildSrc")
             modelsCreated(":")
-            problem("Build file 'build.gradle': line 3: Project ':' cannot access Gradle.buildFinished")
             problem("Build file 'build.gradle': line 3: registration of listener on 'Gradle.buildFinished' is unsupported")
         }
     }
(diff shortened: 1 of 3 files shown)


6hundreds:
@bot-gradle test this on linux

6hundreds:
@bot-gradle test this on linux

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiInvocationValidationIntegrationTest.groovy
```diff
@@ -124,24 +124,26 @@ class IsolatedProjectsToolingApiInvocationValidationIntegrationTest extends Abst
         """
 
         when:
-        withIsolatedProjects()
+        withIsolatedProjectsDiagnostics()
```
alllex: 💅 Let's test both default IP and diagnostics to capture the expected UX in both cases

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -253,6 +253,34 @@ class CrossProjectConfigurationReportingGradle(
         return delegate.pluginManager
     }
 
+    @Suppress("OVERRIDE_DEPRECATION", "DEPRECATION")
+    override fun buildFinished(action: Action<in BuildResult>) {
+        // see `addListener`
+        onMutableStateAccess("buildFinished")
+        delegate.buildFinished(action)
```
alllex: ❔ Is it actually forbidden for the same reason as `addListener` ? Is `onMutableStateAccess` the right way to report it?

The reporting provides the same message for all, but even if we don't adjust the message, maybe we can be more precise with internal semantics of why we don't allow this?

6hundreds: > ❔ Is it actually forbidden for the same reason as addListener 

Yes

> Is onMutableStateAccess the right way to report it?

Assuming we doing right for `addListener` - yes. What are the other ways to report we can use either?

> The reporting provides the same message for all

Not sure I got you, messages for user aren't same for `addListener` and `buildFinished`

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingApiInvocationValidationIntegrationTest.groovy
6hundreds: Hm, I can't find a way to trigger CC problem in this test, as all CC-problematic-methods are violations now

alllex: What I meant is to use the data-driven test that runs with both the default and diagnostic IP. The corresponding fixture can help with validating the problems

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
alllex: What are the full messages? I'd be surprised if they differ on anything rather than the mention of the fobidden method name

But my question is basically whether saying that this is "mutable state access" problem is the right thing to say. It's clear that is the case for `gradle.extensions`. It's less clear, but we clarified it for the `addListener(DependencyResolutionListener)`. I am asking whether the `buildFinished` also falls into the category of actual mutable access of something on the Gradle instance that is unsafe to access.

What I had in mind, is maybe using a different intermediate function name to clarify the intent (if it differs, depending on the answers to the above questions). E.g. it could be even `onIsolatedProjectsUnsupported` call, that for now calls the same reporter underneath. But then we can differentiate between the two helpers and make the violation message more precise when it is actually the mutable access. I think right now the violation message is entirely generic, without any further justification.

6hundreds: Now I think that we shouldn't report `buildListener` and `addBuildListener`, as the callback runs at build finish after configuration. Yes, `BuildResult` exposes unwrapped `Gradle`, but at execution time it shouldn't be a problem. WDYT?

alllex (approved):
Really insightful comments which should save us some design reverse-engineering in the future 👍

PR #38253, opened 2026-06-19
Preserve laziness of `ProcessForkOptions.executable(Object)`

abstratt:
If a non-provider lazy object is used as `executable` (possibly an object wrapping a provider), we would eagerly evaluate its `toString()` representation. This change makes it as lazy as when a proper `Provider` is used. 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


asodja (approved):
LGTM

abstratt:
@bot-gradle test this

PR #38256, opened 2026-06-19
Fix corrupt test tree for class-level descriptions in JUnit 4

### LINKED ISSUE #38193: ClassCastException in gradle test reporter code when using specs2 test library with BeforeAfterAll
### Current Behavior

See the attached repro. Run
```
./gradlew test --tests com.example.FailureCaseSpec
```

FailureCaseSpec implements a test spec with BeforeAfterAll which means it runs the beforeAll() method before all the tests and afterAll() after the tests have completed.  

If there is an error in the beforeAll() method, there is a failure as follows
```
jreilly@host gradle_test_reporter_repro (main) $ ./gradlew test --tests com.example.FailureCaseSpec

> Task :test FAILED
No reporter found for test descriptor: Test classMethod(com.example.FailureCaseSpec)
java.lang.IllegalStateException: No reporter found for test descriptor: Test classMethod(com.example.FailureCaseSpec)
        at org.gradle.api.internal.tasks.testing.junit.result.TestEventReporterAsListener.completed(TestEventReporterAsListener.java:76)
        at org.gradle.api.internal.tasks.testing.results.SynchronizedTestListenerInternal.completed(SynchronizedTestListenerInternal.java:49)
        at org.gradle.api.internal.tasks.testing.results.StateTrackingTestResultProcessor.completed(StateTrackingTestResultProcessor.java:98)
        at org.gradle.api.internal.tasks.testing.results.AttachParentTestResultProcessor.completed(AttachParentTestResultProcessor.java:61)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at org.gradle.internal.dispatch.MethodInvocation.invokeOn(MethodInvocation.java:77)
        at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:28)
        at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:19)
        at org.gradle.internal.dispatch.FailureHandlingDispatch.dispatch(FailureHandlingDispatch.java:30)
        at org.gradle.internal.dispatch.AsyncDispatch.dispatchMessages(AsyncDispatch.java:87)
        at org.gradle.internal.dispatch.AsyncDispatch.access$000(AsyncDispatch.java:36)
        at org.gradle.internal.dispatch.AsyncDispatch$1.run(AsyncDispatch.java:71)
        at org.gradle.internal.concurrent.InterruptibleRunnable.run(InterruptibleRunnable.java:44)
        at org.gradle.internal.operations.CurrentBuildOperationRef.with(CurrentBuildOperationRef.java:84)
        at org.gradle.internal.operations.CurrentBuildOperationPreservingRunnable.run(CurrentBuildOperationPreservingRunnable.java:51)
        at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:64)
        at org.gradle.internal.concurrent.AbstractManagedExecutor$1.run(AbstractManagedExecutor.java:47)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1583)

com.example.FailureCaseSpec.classMethod FAILED

3 tests completed, 1 failed, 2 skipped

* What went wrong:
Execution failed for task ':test' (registered by plugin 'org.gradle.jvm-test-suite').
> Test process encountered an unexpected problem.
   > class org.gradle.api.internal.tasks.testing.LifecycleTrackingTestEventReporter cannot be cast to class org.gradle.api.internal.tasks.testing.GroupTestEventReporterInternal (org.gradle.api.internal.tasks.testing.LifecycleTrackingTestEventReporter and org.gradle.api.internal.tasks.testing.GroupTestEventReporterInternal are in unnamed module of loader org.gradle.initialization.MixInLegacyTypesClassLoader @194bcebf)

* Try:
> Check common problems https://docs.gradle.org/9.5.0/userguide/java_testing.html#sec:java_testing_troubleshooting.
> Run with --stacktrace option to get the stack trace.
> Run with --scan to get full insights from a Build Scan (powered by Develocity).

BUILD FAILED in 733ms
2 actionable tasks: 1 executed, 1 up-to-date
jreilly@host gradle_test_reporter_repro (main) $
```

Notice in particular 
```
> Test process encountered an unexpected problem.
   > class org.gradle.api.internal.tasks.testing.LifecycleTrackingTestEventReporter cannot be cast to class org.gradle.api.internal.tasks.testing.GroupTestEventReporterInternal (org.gradle.api.internal.tasks.testing.LifecycleTrackingTestEventReporter and org.gradle.api.internal.tasks.testing.GroupTestEventReporterInternal are in unnamed module of loader org.gradle.initialization.MixInLegacyTypesClassLoader @194bcebf)
```

I have also tried with the latest gradle in master and see the same issue. 

It seems that the issue is that https://github.com/gradle/gradle/blob/master/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/TestEventReporterAsListener.java#L57-L58 assumes that `testDescriptor.getParent()` will always return a GroupTestEventReporterInternal but this is not always the case.

### Expected Behavior

The test reporter should not through an exception and should output the error that happened in the code in the test.  

If I revert back to using gradle 9.2.1 I see the following as expected
```
jreilly@host gradle_test_reporter_repro (main) $ ./gradlew test --tests com.example.FailureCaseSpec
Fetching distribution.
Downloading https://services.gradle.org/distributions/gradle-9.2.1-bin.zip
............10%.............20%.............30%.............40%.............50%.............60%.............70%.............80%.............90%.............100%
Starting a Gradle Daemon, 2 incompatible Daemons could not be reused, use --status for details

> Task :test FAILED

FailureCaseSpec > classMethod FAILED
    org.specs2.specification.core.FatalExecution
        Caused by: java.lang.RuntimeException at FailureCaseSpec.scala:20

3 tests completed, 1 failed, 2 skipped

* What went wrong:
Execution failed for task ':test'.
> There were failing tests. See the report at: file:///Users/jreilly/checkout/gradle_test_reporter_repro/build/reports/tests/test/index.html

* Try:
> Run with --scan to generate a Build Scan (powered by Develocity).

BUILD FAILED in 13s
2 actionable tasks: 2 executed
jreilly@host gradle_test_reporter_repro (main) $ 
```

### Context (optional)

_No response_

[gradle_test_reporter_repro.tgz](https://github.com/user-attachments/files/28856951/gradle_test_reporter_repro.tgz)

### Gradle version

9.5.0

### Build scan URL (optional)

_No response_

Apple MacBookPro M4 w/ MacOS 26.5.1 with Java 21

openjdk version "21.0.11" 2026-04-21 LTS
OpenJDK Runtime Environment Zulu21.50+19-CA (build 21.0.11+10-LTS)
OpenJDK 64-Bit Server VM Zulu21.50+19-CA (build 21.0.11+10-LTS, mixed mode, sharing)

tresat:
## The Problem

`JUnitTestEventAdapter`'s tree-walk during `testRunStarted` could
register a class-level `Description` as a non-composite leaf when
its `isSuite()` returned false because the runner hadn't yet
populated children. The user-reported case is Specs2's JUnitRunner
reporting a spec class whose nested `should` blocks become
children only after the runner's discovery phase. A subsequent
synthetic `classMethod` failure (e.g., from `BeforeAfterAll`'s
`beforeAll()` throwing) attached to that leaf parent caused
`TestEventReporterAsListener`'s unconditional cast at lines 57-58
to throw `ClassCastException`, surfacing as "Test process
encountered an unexpected problem" along with the stack trace
in the user's bug report.

## The Fix

Detect class-level `Description`s in `addDescriptorAndChildren`
via the new `isClassLevelDescription` helper:

```
(methodName == null && (testClass != null || description.toString() == className(description)))
```

and route them to `RegistrationMode.SUITE`
rather than `DETECTED`. Class nodes now register as
`DefaultTestClassDescriptor` (composite) from the start; every
downstream consumer of the event tree sees a consistent shape.

## Limitations and Compatibility with Older JUnit Versions

The fix is scoped to `addDescriptorAndChildren` only.
`createNode`'s existing predicate, `processIgnoredClass`'s child
loop, and `testIgnored` continue to use `RegistrationMode.DETECTED`
unchanged, so JUnit 4 `@Ignore` and `Enclosed` runner behavior is
not affected.

The predicate's fallback branch
`(description.toString() == className)` covers JUnit <= 4.5 and
custom-runner-produced synthetic suites
`(Description.createSuiteDescription(String))`. Routing those to
the composite branch is benign because `createNode` already
sends null-testClass cases to `DefaultTestSuiteDescriptor` rather
than `DefaultTestClassDescriptor`.

Fixes:
- https://github.com/gradle/gradle/issues/38193

### DIFF
--- a/platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/junit/result/TestEventReporterAsListenerTest.groovy
+++ b/platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/junit/result/TestEventReporterAsListenerTest.groovy
@@ -27,19 +27,14 @@ import java.util.function.Function
 
 @Issue("https://github.com/gradle/gradle/issues/38193")
 class TestEventReporterAsListenerTest extends Specification {
-
     /*
-     * Negative regression test: pins the consumer-side invariant that the producer-side
-     * fix in JUnitTestEventAdapter.addDescriptorAndChildren now relies on.
-     *
-     * The cast at TestEventReporterAsListener.java:57-58 is intentionally unguarded.
-     * The producer side guarantees that any descriptor which will acquire children
+     * Negative regression test: pins the consumer-side invariant that any descriptor which will acquire children
      * is registered as composite from the start, so the parent-slot reporter is always
      * a GroupTestEventReporterInternal. If that producer-side invariant is ever broken
      * (a future change emits a non-composite parent slot) the cast will throw
      * ClassCastException — and this test goes red as the canary.
      */
-    def "throws ClassCastException when fed a non-composite parent — pins the consumer-side invariant the producer-side fix relies on"() {
+    def "throws ClassCastException when fed a non-composite parent"() {
         given:
         def rootGroupReporter = Mock(GroupTestEventReporterInternal)
         def specClassReporter = Mock(TestEventReporter)
--- a/platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/junit/result/TestEventReporterAsListenerTest.groovy
+++ b/platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/junit/result/TestEventReporterAsListenerTest.groovy
@@ -32,9 +32,9 @@ class TestEventReporterAsListenerTest extends Specification {
      * is registered as composite from the start, so the parent-slot reporter is always
      * a GroupTestEventReporterInternal. If that producer-side invariant is ever broken
      * (a future change emits a non-composite parent slot) the cast will throw
-     * ClassCastException — and this test goes red as the canary.
+     * IllegalStateException — and this test goes red as the canary.
      */
-    def "throws ClassCastException when fed a non-composite parent"() {
+    def "throws IllegalStateException when fed a non-composite parent"() {
         given:
         def rootGroupReporter = Mock(GroupTestEventReporterInternal)
         def specClassReporter = Mock(TestEventReporter)
@@ -67,8 +67,11 @@ class TestEventReporterAsListenerTest extends Specification {
         when: "a synthetic child descriptor is started under the (now-leaf) spec-class"
         listener.started(classMethod, new TestStartEvent(0L))
 
-        then: "the unguarded cast throws ClassCastException naming GroupTestEventReporterInternal"
-        def e = thrown(ClassCastException)
-        e.message.contains("GroupTestEventReporterInternal")
+        then: "the guarded cast throws IllegalStateException with a ClassCastException cause naming GroupTestEventReporterInternal"
+        def e = thrown(IllegalStateException)
+        e.message == "An internal test structure error occurred. Please file a new issue on https://github.com/gradle/gradle/issues, and if possible include the test class causing this failure."
+        def cause = e.cause
+        cause instanceof ClassCastException
+        cause.message.contains("GroupTestEventReporterInternal")
     }
 }
--- a/platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/junit/result/TestEventReporterAsListenerTest.groovy
+++ b/platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/junit/result/TestEventReporterAsListenerTest.groovy
@@ -28,7 +28,18 @@ import java.util.function.Function
 @Issue("https://github.com/gradle/gradle/issues/38193")
 class TestEventReporterAsListenerTest extends Specification {
 
-    def "does not throw ClassCastException when a child descriptor's parent was reported as a non-composite leaf"() {
+    /*
+     * Negative regression test: pins the consumer-side invariant that the producer-side
+     * fix in JUnitTestEventAdapter.addDescriptorAndChildren now relies on.
+     *
+     * The cast at TestEventReporterAsListener.java:57-58 is intentionally unguarded.
+     * The producer side guarantees that any descriptor which will acquire children
+     * is registered as composite from the start, so the parent-slot reporter is always
+     * a GroupTestEventReporterInternal. If that producer-side invariant is ever broken
+     * (a future change emits a non-composite parent slot) the cast will throw
+     * ClassCastException — and this test goes red as the canary.
+     */
+    def "throws ClassCastException when fed a non-composite parent — pins the consumer-side invariant the producer-side fix relies on"() {
         given:
         def rootGroupReporter = Mock(GroupTestEventReporterInternal)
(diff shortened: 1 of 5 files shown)


tresat:
@bot-gradle test this

tresat:
Failing test seems like a CI issue.

platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/junit/result/TestEventReporterAsListenerTest.groovy
```diff
@@ -0,0 +1,74 @@
...
+        and: "the root and the spec-class start normally"
+        listener.started(root, new TestStartEvent(0L))
+        listener.started(specClass, new TestStartEvent(0L))
+
+        when: "a synthetic child descriptor is started under the (now-leaf) spec-class"
+        listener.started(classMethod, new TestStartEvent(0L))
+
+        then: "the unguarded cast throws ClassCastException naming GroupTestEventReporterInternal"
+        def e = thrown(ClassCastException)
+        e.message.contains("GroupTestEventReporterInternal")
```
octylFractal: :x: IMO we should get a better error than a CCE. Since this is something we can reasonably expect might occur with misuse of the `TestEventReporter` API, I think we should change our cast to check first and report a proper error if it's wrong.

octylFractal (commented):
LGTM aside from also updating the exception.

platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/junit/result/TestEventReporterAsListenerTest.groovy
tresat: Sounds reasonable.  What do you think the new error should say?  I don't think there's much a user could do other than restructure test suites/nested tests/other test containers, so I don't think making a custom exception type that would carry resolution suggestions would help.

I tried wrapping the exception in the latest commit, what do you think?

octylFractal: Ah, I didn't realize it was in the internal adapter at first. What you put is good enough.

octylFractal (approved):

PR #38258, opened 2026-06-20
Return Dependency from Kotlin dependency helper

### LINKED ISSUE #12172: DependencyHandler.kotlin returns Any instead of String and doesn't work with implementation taking dependencyConfiguration
This is a very small issue, similar to https://github.com/gradle/gradle/issues/11956.

The `DependencyHandler.kotlin` helper returns `Any`, which works usually, because, for example, `DependencyHandler.implementation` also accepts `Any`. However, I just noticed, when you add `dependencyConfiguration` to the `dependency`, it won't work anymore, because the latter only accepts a `String`. It seems to me that `DependencyHandler.kotlin` could simply return a `String` instead of `Any` to make this work.

### Expected Behavior

The following should work:
```
    implementation(kotlin("reflect")) {
        because("reason")
    }
If `kotlin` would return a `String` it would work.
```

### Current Behavior
A type cast is needed:
```
    implementation(kotlin("reflect") as String) {
        because("reason")
    }
```

### Your Environment

Gradle 6.1.

hamzigh:
Fixes #12172

Context

The kotlin() dependency helper returns dependency notation as Any, which prevents it from being used with a dependency configuration block:

```groovy
implementation(kotlin("reflect")) {
    because("reason")
}
```

Create the dependency through create() and return it as a Dependency, allowing the dependency to be configured without a cast.

An integration test covers this case.

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic/kotlin-dsl/src/main/kotlin/gradlebuild/kotlindsl/generator/tasks/GenerateKotlinDependencyExtensions.kt
+++ b/build-logic/kotlin-dsl/src/main/kotlin/gradlebuild/kotlindsl/generator/tasks/GenerateKotlinDependencyExtensions.kt
@@ -63,6 +63,7 @@ val expectedKotlinDslPluginsVersion: String
 
 package org.gradle.kotlin.dsl
 
+import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.dsl.DependencyHandler
 
 import org.gradle.plugin.use.PluginDependenciesSpec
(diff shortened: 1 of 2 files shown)


ov7a:

***

Note that line 83 also contains the same problem.

https://github.com/gradle/gradle/pull/38259 fixes the same issue; the team will decide which approach to keep.

jbartok:
Seems like we have a parallel PR, doing the same fix (https://github.com/gradle/gradle/pull/38259) and it has more tests, so we are going to close this one in favor of that. 
Please review that PR, if you can.

PR #38259, opened 2026-06-20
Issue #12172 DependencyHandler.kotlin returns Any instead of String

### LINKED ISSUE #12172: DependencyHandler.kotlin returns Any instead of String and doesn't work with implementation taking dependencyConfiguration

sarosh-engineer:
Fixes #12172 

### Context

Resolves Issue #12172 

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic/kotlin-dsl/src/main/kotlin/gradlebuild/kotlindsl/generator/tasks/GenerateKotlinDependencyExtensions.kt
+++ b/build-logic/kotlin-dsl/src/main/kotlin/gradlebuild/kotlindsl/generator/tasks/GenerateKotlinDependencyExtensions.kt
@@ -80,7 +80,7 @@ val embeddedKotlinVersion = "$embeddedKotlinVersion"
  *
  * @param module simple name of the Kotlin module, for example "reflect".
  */
-fun DependencyHandler.embeddedKotlin(module: String): Any =
+fun DependencyHandler.embeddedKotlin(module: String): String =
     kotlin(module, embeddedKotlinVersion)
 
 
--- a/build-logic/kotlin-dsl/src/main/kotlin/gradlebuild/kotlindsl/generator/tasks/GenerateKotlinDependencyExtensions.kt
+++ b/build-logic/kotlin-dsl/src/main/kotlin/gradlebuild/kotlindsl/generator/tasks/GenerateKotlinDependencyExtensions.kt
@@ -63,6 +63,7 @@ val expectedKotlinDslPluginsVersion: String
 
 package org.gradle.kotlin.dsl
 
+import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.dsl.DependencyHandler
 
 import org.gradle.plugin.use.PluginDependenciesSpec
@@ -80,7 +81,7 @@ val embeddedKotlinVersion = "$embeddedKotlinVersion"
  *
  * @param module simple name of the Kotlin module, for example "reflect".
  */
-fun DependencyHandler.embeddedKotlin(module: String): String =
+fun DependencyHandler.embeddedKotlin(module: String): Dependency =
     kotlin(module, embeddedKotlinVersion)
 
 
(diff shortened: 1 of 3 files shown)


ov7a:

***

Note that line 83 also contains the same problem.

https://github.com/gradle/gradle/pull/38258 fixes the same issue; the team will decide which approach to keep.

sarosh-engineer:
> Thank you for your proposed contribution!
> 
> This PR has a valid DCO and tests. The relevant team for this area will confirm the implementation choices.
> 
> ***
> 
> Note that line 83 also contains the same problem.
> 
> https://github.com/gradle/gradle/pull/38258 fixes the same issue; the team will decide which approach to keep.

Hi , Can I push more changes on the PR?

ov7a:
> Hi , Can I push more changes on the PR?

Sure

sarosh-engineer:
> > Hi , Can I push more changes on the PR?
> 
> Sure

I see the issue prefers Dependency as the return type. Should I change my implementation to return a dependency instead ?

jbartok:
> > > Hi , Can I push more changes on the PR?
> > 
> > 
> > Sure
> 
> I see the issue prefers Dependency as the return type. Should I change my implementation to return a dependency instead ?

Yes, please do.

sarosh-engineer:
> > > > Hi , Can I push more changes on the PR?
> > > 
> > > 
> > > Sure
> > 
> > 
> > I see the issue prefers Dependency as the return type. Should I change my implementation to return a dependency instead ?
> 
> Yes, please do.

Changes have been implemented & sanity test has been run.

jbartok:

This PR looks good, we will delay merging it until the next major version (Gradle 10) because it contains breaking changes.

sarosh-engineer:
> Thank you for your proposed contribution!
> 
> This PR looks good, we will delay merging it until the next major version (Gradle 10) because it contains breaking changes.

Thank you!

build-logic/kotlin-dsl/src/main/kotlin/gradlebuild/kotlindsl/generator/tasks/GenerateKotlinDependencyExtensions.kt
```diff
@@ -80,7 +81,7 @@ val embeddedKotlinVersion = "$embeddedKotlinVersion"
  *
  * @param module simple name of the Kotlin module, for example "reflect".
  */
-fun DependencyHandler.embeddedKotlin(module: String): Any =
+fun DependencyHandler.embeddedKotlin(module: String): Dependency =
```
hfhbd, line 84: Why `Dependency` and not `ExternalModuleDependency`?

sarosh-engineer: Hi @hfhbd ,
I'm new to gradle. The issue mentioned to return a Dependency. I understand we can keep it as ExternalModuleDependency but it would be specific instead of a generic Dependency. Are there use cases where we would need to cast this Dependency as an ExternalModuleDependency to make it work?

jbartok: Some conclusions after discussing it internally:
* we want to stick with `Dependency`; `ExternalModuleDependency` exposes an API surface we don't want to expose here
* there might be a backwards compatibility problem with changing the return type here; if you build a plugin that calls `embeddedKotlin(...)` with the old signature and somebody tries to use it in a build with the new signature, the binary plugin might have a reference to the old signature and not work; something to check
* additionally, if people figured out the workaround `(embeddedKotlin("...") as String)`, this also breaks them because now we're definitely not returning a String
* we should explore why this method is generated instead of being defined in `DependencyHandlerExtensions`
* we might want to remove these Kotlin-specific methods completely at some point, there should be no special treatment for any DSL in the long run

PR #38241, opened 2026-06-18
Annotate ToBeFixedForConfigurationCache with issues using the new param

mlopatkin:
### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #22160, opened 2022-09-23
Upgrade `release` wrapper to latest nightly

eskatos:
Upgrade `release` wrapper to latest nightly

PR #22158, opened 2022-09-23
Upgrade the spock version used in samples/snippets

DPUkyle:
Also upgrade the version used by the init plugin.

Also limit TestNG to < 7.6, as it requires JDK 11

Extracted from #20038

PR #38265, opened 2026-06-22
Don't fail the performance report on results restored from the build cache

blindpirate:
## Problem

The `…PerformanceTestTestLinux_Trigger` builds fail consecutively (e.g. [#22158–#22160 on PR #38241](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_PerformanceTestTestLinux_Trigger?branch=gh-readonly-queue%2Fmaster%2Fpr-38241-2651110360f0b273e08c5d9bd4b0714310045f47)) with a scenario failure that references a `buildId` from a **different PR**.

Root cause: the `PerformanceTest` bucket task is `@CacheableTask`, and it baked build-specific data — the producing build's `teamCityBuildId`, the `webUrl` derived from it, and the pass/fail `status`/`testFailure` — into its result JSON, while `buildId` was `@Internal` (excluded from the cache key). On a Gradle **build-cache hit** the bucket is not re-executed: the restored `perf-results-*.json` replays the *original* build's identity and its (often noise-level, e.g. 0.59% @ 99% confidence) failure. The aggregate report then attributes that stale failure to the current pipeline. In the failing case all 50 buckets were cache hits from the previous merge-queue entry.

## Fix: put no build-specific data in the cacheable output; derive it downstream

**The cacheable bucket output is now identity-only.** `ScenarioBuildResultData` / `PerformanceTestExecutionResult` carry only the scenario name/class/project — no build id, URL, status or failure text. So a cache hit can no longer replay a previous build's identity or verdict.

**The report derives everything else at aggregation time:**
- **Provenance** — the Trigger passes this pipeline's authoritative bucket build IDs from the TeamCity dependency graph (`%dep.<id>.env.BUILD_ID%`) as `-Porg.gradle.performance.dependencyBuildIds`. The TeamCity build links and the history-graph build filter are built from that system property.
- **Verdict** — computed from the **performance database**, not the JSON status. `PerformanceReportScenario` identifies this pipeline's "current" executions by matching each DB row's own `teamCityBuildId` (written accurately by the run that measured it; a cache hit writes no DB row at all) against those authoritative IDs — falling back to the commit under test for local runs. A scenario fails only if those current-pipeline measurements show a confident regression. All scenario classifications (regressed / improved / successful / unknown) are DB-based; report generators, comparators and flakiness detection were updated to match.

Net: a build-cache hit contributes no DB row and no build-specific JSON, so it is simply "no current measurement" (`unknown`) and cannot fail the build; genuine in-pipeline regressions still do.

## Notes

- Targets the **`release`** branch.
- Not covered here: caching a stochastic benchmark is itself questionable (a cached result is a replay, not a measurement) — left as a separate discussion.

## Verification

- `DefaultPerformanceExecutionDataProviderTest` green — covers: CI fresh run → regression gated from the DB; **build-cache hit → no current execution → not gated**; local run → commit fallback; and regressed-first / unknown-last ordering.
- `.teamcity` config-generation tests green.
- Compilation verified across `build-logic` and the report module (every consumer of the removed fields updated). The generated HTML report was not render-checked locally — worth a glance when CI publishes the report artifact.

### DIFF
--- a/.teamcity/src/main/kotlin/configurations/PerformanceTestsPass.kt
+++ b/.teamcity/src/main/kotlin/configurations/PerformanceTestsPass.kt
@@ -82,9 +82,13 @@ testing/$performanceProjectName/build/performance-test-results.zip
             )
             val bucketedPerformanceTests = performanceTestProject.performanceTests.filter { it.testProjects.isNotEmpty() }
             if (bucketedPerformanceTests.isNotEmpty()) {
-                // Baselines come from this build's own parameter; per-bucket TeamCity build IDs are derived inside the
-                // report task from the surviving perf-results-*.json files. The expected bucket count is passed so the
-                // verify task can fail the Trigger when fewer buckets reported than configured.
+                // Authoritative per-bucket TeamCity build IDs from the dependency graph. The report needs these to tell
+                // results this pipeline produced apart from results restored from the Gradle build cache: a cached
+                // bucket's perf-results-*.json carries the `teamCityBuildId` of the build that originally produced it,
+                // so the JSON cannot be trusted to identify this pipeline's buckets.
+                val dependencyBuildIds = bucketedPerformanceTests.joinToString(",") { "%dep.${it.id}.env.BUILD_ID%" }
+                // Baselines come from this build's own parameter. The expected bucket count is passed so the verify
+                // task can fail the Trigger when fewer buckets reported than configured.
                 val verifyTaskName = "verify${taskName.replaceFirstChar { it.titlecase() }}Buckets"
                 gradleRunnerStep(
                     model,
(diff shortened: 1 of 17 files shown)


.teamcity/src/main/kotlin/configurations/PerformanceTestsPass.kt
```diff
@@ -82,9 +82,13 @@ testing/$performanceProjectName/build/performance-test-results.zip
             )
             val bucketedPerformanceTests = performanceTestProject.performanceTests.filter { it.testProjects.isNotEmpty() }
             if (bucketedPerformanceTests.isNotEmpty()) {
-                // Baselines come from this build's own parameter; per-bucket TeamCity build IDs are derived inside the
-                // report task from the surviving perf-results-*.json files. The expected bucket count is passed so the
-                // verify task can fail the Trigger when fewer buckets reported than configured.
+                // Authoritative per-bucket TeamCity build IDs from the dependency graph. The report needs these to tell
+                // results this pipeline produced apart from results restored from the Gradle build cache: a cached
+                // bucket's perf-results-*.json carries the `teamCityBuildId` of the build that originally produced it,
```
cobexer, line 87: ❌ If the producing build stores the `teamCityBuildId` in the output JSON - and if we really need that ID, that ID needs to be an input of the task that produces the JSON and then we don't need to workaround any problem here.

cobexer (dismissed):

PR #38266, opened 2026-06-22
Fix a bug with chained ValueSource deserialization

### LINKED ISSUE #38255: Configuration Cache fails with chained ValueSource providers
### Current Behavior

When running a task that uses a chained ValueSource setup (a ValueSource parameter referencing another Provider created via providers.of(ValueSource::class)), the build fails when the configuration cache is enabled.

```
tasks.register("myTask") {
    var defaultValue = providers.of(MyValueSource::class) {
        parameters.value.set("hello world")
    }
    var value = providers.of(MyValueSource::class) {
        parameters.defaultValue.set(defaultValue)
    }
    doLast {
        println(value.get())
    }
}
```

```
* What went wrong:
Could not load the value of field `element` of `kotlin.jvm.internal.Ref$ObjectRef` bean found in field `$value` of `Build_gradle$$$result$1$1` bean found in field `action` of `org.gradle.api.internal.AbstractTask$TaskActionWrapper` bean found in actions of task `:myTask` of type `org.gradle.api.DefaultTask`.
> Failed to decode shared value
```

This is reproducible only with:

--configuration-cache

and only in newer Gradle 9.7 snapshot builds.
This is a regression between Gradle snapshot builds 20260608005045 (works) and 20260610005640 (fails).

### Expected Behavior

The build succeeds without errors.

### Context (optional)

_No response_

https://github.com/desoletti/chained-value-source-issue

### Gradle version

9.7.0-20260610005640

9.7.0-20260608005045

### Build scan URL (optional)

_No response_

_No response_

ov7a:
Fixes #38255

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/ProviderCodecs.kt
+++ b/platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/ProviderCodecs.kt
@@ -69,8 +69,8 @@ import org.gradle.internal.serialize.graph.readClassOf
 import org.gradle.internal.serialize.graph.readNonNull
 import org.gradle.internal.serialize.graph.runReadOperation
 import org.gradle.internal.serialize.graph.serviceOf
+import org.gradle.internal.serialize.graph.withCodec
 import org.gradle.internal.serialize.graph.withDebugFrame
-import org.gradle.internal.serialize.graph.withImmediateMode
 import org.gradle.internal.serialize.graph.withIsolate
 import org.gradle.internal.serialize.graph.withPropertyTrace
 
@@ -281,8 +281,14 @@ object BuildServiceParameterCodec : Codec<BuildServiceParameters> {
         }.uncheckedCast()
 }
 
-
-object ValueSourceProviderCodec : Codec<ValueSourceProvider<*, *>> {
+/**
+ * @param newUserTypeCodecs Creates an independent set of user-type codecs to decode ValueSource params
+ * in a nested decoding session.
+ * Decoding the parameters must not end up suspending, because it is triggered from synchronous Java code.
+ */
+class ValueSourceProviderCodec(
+    private val newUserTypeCodecs: () -> Codec<Any?>
+) : Codec<ValueSourceProvider<*, *>> {
 
     override suspend fun WriteContext.encode(value: ValueSourceProvider<*, *>) {
         writeSharedObject(value) {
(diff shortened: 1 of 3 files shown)


ov7a:
@bot-gradle test this

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/ProviderCodecs.kt
```diff
@@ -336,7 +337,9 @@ object ValueSourceProviderCodec : Codec<ValueSourceProvider<*, *>> {
             ) { providerInstance ->
                 readContext.runReadOperation {
                     sharedIdentities.putInstance(id, providerInstance)
-                    read()!!.uncheckedCast()
+                    withImmediateMode {
```
mlopatkin: ❓ I'm not sure immediate mode is the proper solution here. It exposes us to StackOverflow that coroutines should shield us from. Maybe @bamboo should also take a look as the original author of the serialization pipeline.

Conceptually, we need to make sure that the `reentrant` that wraps  the `BeanCodec` unfolds everything it stacks within `runReadOperation` and resumes itself. For now it doesn't do that because there are some pending frames from the outside. I think a cleaner solution would be to just start a nested set of codecs, though this would require some plumbing to deliver them:

```suggestion
                    withCodec(newUserTypeCodecs()) { // newUserTypeCodecs() reaches into the enclosing DefaultConfigurationCacheCodecs instance and creates the proper set for fingerprint or entry.
```

mlopatkin (changes requested):

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/ProviderCodecs.kt
```diff
@@ -281,7 +282,18 @@ object BuildServiceParameterCodec : Codec<BuildServiceParameters> {
...
-object ValueSourceProviderCodec : Codec<ValueSourceProvider<*, *>> {
+/**
+ * @param newUserTypeCodecs creates a fresh, independent set of user-type codecs. A new set is needed
+ * when decoding the parameters of a value source: the parameters are read inside a nested
+ * [runReadOperation], which starts a fresh coroutine that cannot suspend. The shared codec set wraps
+ * [BeanCodec] in a stateful `reentrant` codec, and that wrapper suspends on a nested decode if an
+ * outer decode is still in progress (as is the case here, when one value source's parameters
+ * reference another value source). Decoding through a fresh codec set gives the `reentrant` wrapper
+ * clean state so it can unfold its trampoline to completion within the nested coroutine.
+ */
```
mlopatkin, line 288: 💅 All this doesn't belong to the api docs, but to an implementation comment. 

I'd put  there something along the lines of
> creates an independent set of user-type codecs to decode ValueSource params in a nested decoding session. Decoding the parameters must not end up suspending, because it is triggered from synchronous Java code.

And the rest of the comment to `withCodec` - to explain why we need the new codec.

mlopatkin (approved):
Small comment nitpick, but overall good to go.

PR #38269, opened 2026-06-22
TeamCity: switch weekly Gradleception run to use the latest version of Java supported

cobexer:
- Change the weekly Gradleception CI run to build with the maximum supported JDK instead of the maximum LTS JDK
- Rename the `GradleceptionWithMaxLtsJdk` build type to `GradleceptionWithMaxJdk` and switch from `JvmCategory.MAX_LTS_VERSION` to `JvmCategory.MAX_VERSION`


cobexer:
@bot-gradle test this

testing/architecture-test/src/test/java/org/gradle/architecture/test/XmlFactoriesTest.java
```diff
@@ -62,9 +62,15 @@ public class XmlFactoriesTest {
...
             .should()
-            .callMethodWhere(describe("static XML factories", methodCall ->
-                xmlFactoryClasses.stream().anyMatch(clazz -> methodCall.getTarget().getOwner().isAssignableFrom(clazz)) &&
-                    methodCall.getTarget().resolveMember().get().getModifiers().contains(JavaModifier.STATIC)
-            ))
+            .callMethodWhere(describe("static XML factories", methodCall -> {
+                var resolveMember = methodCall.getTarget().resolveMember();
+                if (resolveMember.isEmpty()) {
+                    return false;
+                }
```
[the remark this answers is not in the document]
cobexer: We hit this in places we have no control over, so we must allow places where we cant check for this

platforms/core-runtime/io/src/main/java/org/gradle/internal/io/StreamByteBuffer.java
```diff
@@ -266,6 +267,7 @@ protected int allocateSpace() {
         return spaceLeft;
     }
 
+    @SuppressWarnings("ReferenceEquality")
     protected int prepareRead() {
```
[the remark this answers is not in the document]
cobexer: not sure there is a lot of value TBH

platforms/core-runtime/io/src/main/java/org/gradle/internal/io/StreamByteBuffer.java
```diff
@@ -242,6 +242,7 @@ private boolean hasRemaining(ByteBuffer nextBuf) {
         return nextBuf != null && nextBuf.hasRemaining();
     }
 
+    @SuppressWarnings("ReferenceEquality")
     public int totalBytesUnread() {
```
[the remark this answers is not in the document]
cobexer: not sure there is a lot of value TBH

gradle/dependency-management/shared-versions.properties
```diff
@@ -17,7 +17,7 @@
 # platforms/core-runtime/base-asm/src/main/java/org/gradle/model/internal/asm/AsmConstants.java
 # when upgrading ASM.
 asm=9.9!!
-errorProne=2.42.0!!
+errorProne=2.50.0!!
 groovy=4.0.32!!
```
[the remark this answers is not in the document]
cobexer: prerequisites...

testing/architecture-test/src/test/java/org/gradle/architecture/test/XmlFactoriesTest.java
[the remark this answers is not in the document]
cobexer: is there a way to freeze allowed places we don't verify we could add easily?

.teamcity/src/main/kotlin/configurations/Gradleception.kt
```diff
@@ -98,10 +98,9 @@ class Gradleception(
         }
 
         if (buildJvm.version != BuildToolBuildJvm.version) {
-            steps.gradleWrapper {
+            steps.script {
```
cobexer, line 101: 🎉 this saves a couple of minutes because we avoid starting Gradle on a Java version we don't use later any more just to change one number...

.teamcity/src/main/kotlin/configurations/Gradleception.kt
```diff
@@ -98,10 +98,9 @@ class Gradleception(
...
                 name = "UPDATE_DAEMON_JVM_CRITERIA_FILE"
-                tasks = "updateDaemonJvm --jvm-version=${buildJvm.version.major}"
-                gradleParams = defaultParameters
+                scriptContent = "echo 'toolchainVersion=${buildJvm.version.major}' > gradle/gradle-daemon-jvm.properties"
```
cobexer, line 103: 💭 we already provision JDKs in CI and disable all auto provisioning, so we only need the major version number

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/work/ConditionalExecutionQueue.java
```diff
@@ -25,7 +25,7 @@
 public interface ConditionalExecutionQueue<T> extends Stoppable {
     /**
      * Submit a new conditional execution to the queue.  The execution will occur asynchronously when the provided
-     * resource lock (see {@link ConditionalExecution#getResourceLock()}) can be acquired.  On completion,
```
cobexer, line 28: 🤔 This may be no longer accurate as of https://github.com/gradle/gradle/commit/b88d14b9715bbb8fa1a7dd8405b612d4e56c4a01

testing/architecture-test/src/test/java/org/gradle/architecture/test/XmlFactoriesTest.java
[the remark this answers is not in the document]
alllex: > is there a way to freeze allowed places we don't verify we could add easily?

You could have a neighbor rule next to this one that is freezable and only checks for the list of unresolved members. Then, if it changes, it becomes part of the changeset.

Alternatively, just have a list hardcoded in this class itself and fail if an unresolved member is not known. Though this will not catch scenarios when the list shrinks.

gradle/dependency-management/build.versions.toml
```diff
@@ -62,7 +62,7 @@ kotlinMetadata = { module = "org.jetbrains.kotlin:kotlin-metadata-jvm"}
 kotlinSamWithReceiver = { group = "org.jetbrains.kotlin", name = "kotlin-sam-with-receiver"}
 log4jCore = { group = "org.apache.logging.log4j", name = "log4j-core", version.ref = "log4j" }
 mockitoKotlin = { group = "org.mockito.kotlin", name = "mockito-kotlin", version.ref = "mockitoKotlin" }
-nullaway = { module = "com.uber.nullaway:nullaway", version = "0.12.10" }
+nullaway = { module = "com.uber.nullaway:nullaway", version = "0.13.7" }
```
mlopatkin, line 66: 🎉

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/ExecutorPolicy.java
```diff
@@ -80,6 +80,7 @@ public <T> T onExecute(Callable<T> command) throws Exception {
             }
         }
 
+        @SuppressWarnings("NullAway") // AtomicReference.compareAndSet permits a null update value
```
mlopatkin: :x: The comment is incorrect, the type should be fixed to
```
AtomicReference<@Nullable Throwable> failure
```

As it can lead to a cascade of other changes, I'll be okay with marking those as `TODO: fix nullability` next to the suppression to unblock the PR. Let's just not pretend it is a legit suppression.

platforms/core-runtime/io/src/main/java/org/gradle/internal/io/LinePerThreadBufferingOutputStream.java
```diff
@@ -86,12 +86,14 @@ public void flush() {
     }
 
     @Override
+    @SuppressWarnings("AnnotateFormatMethod") // delegating override of java.io.PrintStream; cannot change the signature
     public PrintStream format(String format, Object... args) {
```
mlopatkin: ❓ I don't buy the explanation. Annotations aren't part of the signature:

```suggestion
@FormatMethod
public PrintStream format(@FormatString String format, Object... args) {
```

platforms/core-runtime/logging/src/main/java/org/gradle/api/internal/DocumentationRegistry.java
```diff
@@ -89,6 +89,7 @@ public String getSampleIndex() {
         return BASE_URL + "/samples";
     }
 
+    @SuppressWarnings("FormatStringShouldUsePlaceholders") // format string prefix is computed at runtime; concatenation kept to avoid behavior change
     public String getSampleFor(String id) {
         return String.format(getSampleIndex() + "/sample_%s.html", id);
```
mlopatkin: :x: This should be trivially fixed instead. I suppose Claude overly cautiously assumed there can be placeholders, but if there are some, we're doomed anyway.
```suggestion
    public String getSampleFor(String id) {
        return String.format("%s/sample_%s.html", getSampleIndex(), id);
```

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/ExecutorPolicy.java
```diff
@@ -88,6 +89,7 @@ public void onFailure(String message, Throwable throwable) {
         }
 
         @Override
+        @SuppressWarnings("NullAway") // AtomicReference.getAndSet permits a null update value
```
mlopatkin: :x: ditto, justification should be turned into TODO

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/ProjectConfigurationTracker.java
```diff
@@ -95,6 +95,7 @@ void increment(PluginApplication pluginApplication, long duration) {
                 .increment(duration);
         }
 
+        @SuppressWarnings("BoxingComparator") // intentional comparison of boxed values
         List<InternalPluginApplicationResult> toInternalPluginApplicationResults() {
             return pluginApplicationResults.values().stream()
                 .sorted(comparing(PluginApplicationResult::getFirstApplicationId))
```
mlopatkin: :x: Trivially fixable
```suggestion
        List<InternalPluginApplicationResult> toInternalPluginApplicationResults() {
            return pluginApplicationResults.values().stream()
                .sorted(Comparator.comparingLong(PluginApplicationResult::getFirstApplicationId))
```

platforms/native/plugins-model-native/src/main/java/org/gradle/language/base/plugins/ComponentModelBasePlugin.java
```diff
@@ -99,6 +99,7 @@
 public abstract class ComponentModelBasePlugin implements Plugin<Project> {
 
     @Override
+    @SuppressWarnings("ReferenceEquality") // intentional identity comparison of Project instances
```
mlopatkin: :question: I think this may be a potential bug, even though self-returning methods do not typically decorate.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/verification/report/DependencyVerificationReportWriter.java
```diff
@@ -42,7 +42,9 @@
 public class DependencyVerificationReportWriter {
     private static final Logger LOGGER = Logging.getLogger(DependencyVerificationReportWriter.class);
 
+    @SuppressWarnings("BoxingComparator") // intentional comparison of boxed values
     private static final Comparator<Map.Entry<ModuleComponentArtifactIdentifier, Collection<RepositoryAwareVerificationFailure>>> DELETED_LAST = Comparator.comparing(e -> e.getValue().stream().anyMatch(f -> f.getFailure() instanceof DeletedArtifact) ? 1 : 0);
+    @SuppressWarnings("BoxingComparator") // intentional comparison of boxed values
     private static final Comparator<Map.Entry<ModuleComponentArtifactIdentifier, Collection<RepositoryAwareVerificationFailure>>> MISSING_LAST = Comparator.comparing(e -> e.getValue().stream().anyMatch(f -> f.getFailure() instanceof MissingChecksums) ? 1 : 0);
```
mlopatkin: :x: Obviously not intentional boxing, should be changed to `comparingInt`.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/verification/writer/PgpKeyGrouper.java
```diff
@@ -149,7 +149,7 @@ private void groupUsingRegex(Map.Entry<String, Collection<PgpEntry>> e, List<Str
 
     // Tries to find the common super-group for a list of groups
     // For example given ["org.foo", "org.foo.bar", "org.foo.baz"] it will group using "org.foo.*"
-    @SuppressWarnings("MixedMutabilityReturnType")
+    @SuppressWarnings({"MixedMutabilityReturnType", "BoxingComparator"}) // intentional comparison of boxed values
```
mlopatkin: :x: ditto, should use `Comparator.comparingInt` there.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/verification/writer/VerificationEntry.java
mlopatkin: :x: `thenComparingInt`

platforms/software/security/src/main/java/org/gradle/security/internal/KeyringStripper.java
```diff
@@ -59,7 +59,7 @@ public static PGPPublicKeyRing strip(PGPPublicKeyRing keyring, KeyFingerPrintCal
         return new PGPPublicKeyRing(strippedKeys);
     }
 
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"unchecked", "BoxingComparator"}) // comparing String length; boxing is acceptable here
```
mlopatkin: ❓ I wonder why "boxing is acceptable". Just using `comparingInt` should be fine.

subprojects/core/src/main/java/org/gradle/api/internal/catalog/parser/TomlCatalogFileParser.java
mlopatkin, line 1: :x: not intentional boxing

mlopatkin (changes requested):
Great to see that coming! There are some issues that I'd like to see addressed before it can be merged. I'm not sure if we're aiming at "no production code changes"
- `ReferenceEquality` check is being further refined right now, and EP 2.50 has it arguably too strict. I suggest we disable the check globally and revert all individual suppressions until we can get a refined version. So far, it has too many false positives to be useful.
- `BoxingComparator` are all legitimate and should be fixed. If we don't want to touch production code here, I'm fine with leaving suppressions + TODO and doing a follow-up PR. If you don't have time to do it right now, please file an issue.
- `NullAway` with AtomicReference - these are also legitimate, but leaving a TODO is fine by me
- FormatString suppressions I'd rather have fixed, as the remedy is trivial and unlikely to cause issues.
- One suspicious `ReferenceEquality` suppression in `ComponentModelBasePlugin.java‎` is worth fixing, though these plugins are on their way out anyway - so I'm fine with postponing this (maybe file an issue and add a TODO with an issue link next to it)

testing/architecture-test/src/test/java/org/gradle/architecture/test/XmlFactoriesTest.java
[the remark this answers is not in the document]
cobexer: I tried freezing but that caught 0 instances, but I discovered that I can avoid the problem by reordering the checks in the XML test so I did that

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/ExecutorPolicy.java
cobexer: Added as separate commit, if it needs discussion I'd move it to a separate PR, PTAL

platforms/core-runtime/io/src/main/java/org/gradle/internal/io/LinePerThreadBufferingOutputStream.java
cobexer: Added as separate commit, if it needs discussion I'd move it to a separate PR, PTAL

platforms/core-runtime/logging/src/main/java/org/gradle/api/internal/DocumentationRegistry.java
cobexer: Added as separate commit, if it needs discussion I'd move it to a separate PR, PTAL

platforms/core-runtime/concurrent/src/main/java/org/gradle/internal/concurrent/ExecutorPolicy.java
cobexer: Added as separate commit, if it needs discussion I'd move it to a separate PR, PTAL

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/ProjectConfigurationTracker.java
cobexer: Added as separate commit, if it needs discussion I'd move it to a separate PR, PTAL

platforms/native/plugins-model-native/src/main/java/org/gradle/language/base/plugins/ComponentModelBasePlugin.java
cobexer: Added as separate commit, if it needs discussion I'd move it to a separate PR, PTAL

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/verification/report/DependencyVerificationReportWriter.java
cobexer: Added as separate commit, if it needs discussion I'd move it to a separate PR, PTAL

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/verification/writer/PgpKeyGrouper.java
cobexer: Added as separate commit, if it needs discussion I'd move it to a separate PR, PTAL

platforms/software/security/src/main/java/org/gradle/security/internal/KeyringStripper.java
cobexer: Added as separate commit, if it needs discussion I'd move it to a separate PR, PTAL

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/ivyresolve/verification/writer/VerificationEntry.java
cobexer: Added as separate commit, if it needs discussion I'd move it to a separate PR, PTAL

subprojects/core/src/main/java/org/gradle/api/internal/catalog/parser/TomlCatalogFileParser.java
cobexer: Added as separate commit, if it needs discussion I'd move it to a separate PR, PTAL

mlopatkin (approved):

cobexer:
@bot-gradle test RfR

PR #38272, opened 2026-06-22
Keep op queue main thread working even if there are extra workers

octylFractal:
If we don't, then the invariant that all queued work is processed by the main thread before `waitForWorkToComplete` can be violated.

Fixes https://github.com/gradle/gradle-private/issues/5272
Fixes https://github.com/gradle/gradle/issues/38291

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/internal/operations/DefaultBuildOperationQueue.java
+++ b/subprojects/core/src/main/java/org/gradle/internal/operations/DefaultBuildOperationQueue.java
@@ -271,7 +271,14 @@ class DefaultBuildOperationQueue<T extends BuildOperation> implements BuildOpera
                 } catch (Throwable t) {
                     addFailure(t);
                 } finally {
-                    invalidateIfNeeded();
+                    if (token != null) {
+                        lock.lock();
+                        try {
+                            token.invalidateIfNeeded();
+                        } finally {
+                            lock.unlock();
+                        }
+                    }
                 }
             });
         }
@@ -310,10 +317,16 @@ class DefaultBuildOperationQueue<T extends BuildOperation> implements BuildOpera
             if (queueState != QueueState.Working && helper.isQueueEmpty()) {
                 return true;
             }
-            if (helper.isExtraWorker()) {
-                if (token != null) {
-                    token.invalidateIfNeeded();
-                }
+            return tryRetireIfExtraWorker();
+        }
+
+        @GuardedBy("lock")
+        private boolean tryRetireIfExtraWorker() {
+            // The main thread (token == null) never retires: it's the only worker guaranteed to run,
+            // so it must keep draining in case no spawned worker ever starts.
+            if (token != null && helper.isExtraWorker()) {
+                // Invalidate now so the worker count drops immediately, not when runOperations() ends.
+                token.invalidateIfNeeded();
                 return true;
             }
             return false;
@@ -379,9 +392,7 @@ class DefaultBuildOperationQueue<T extends BuildOperation> implements BuildOpera
                 T operation;
                 lock.lock();
                 try {
-                    if (helper.isExtraWorker()) {
-                        // We should exit, immediately invalidate our token to ensure the count goes down now.
-                        invalidateIfNeeded();
+                    if (tryRetireIfExtraWorker()) {
                         break;
                     }
                     operation = helper.pollWork();
@@ -408,15 +419,5 @@ class DefaultBuildOperationQueue<T extends BuildOperation> implements BuildOpera
             }
         }
 
-        private void invalidateIfNeeded() {
-            lock.lock();
-            try {
-                if (token != null) {
-                    token.invalidateIfNeeded();
-                }
-            } finally {
-                lock.unlock();
-            }
-        }
     }
 }
--- a/subprojects/core/src/test/groovy/org/gradle/internal/operations/DefaultBuildOperationExecutorParallelExecutionTest.groovy
+++ b/subprojects/core/src/test/groovy/org/gradle/internal/operations/DefaultBuildOperationExecutorParallelExecutionTest.groovy
@@ -17,6 +17,11 @@
 package org.gradle.internal.operations
 
 import org.gradle.api.GradleException
+import org.gradle.internal.concurrent.DefaultExecutorFactory
+import org.gradle.internal.concurrent.ExecutorFactory
+import org.gradle.internal.concurrent.ExecutorPolicy
+import org.gradle.internal.concurrent.ManagedExecutor
+import org.gradle.internal.concurrent.ManagedExecutorImpl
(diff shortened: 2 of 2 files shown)


subprojects/core/src/test/groovy/org/gradle/internal/operations/DefaultBuildOperationQueueTest.groovy
```diff
@@ -339,6 +339,48 @@ class DefaultBuildOperationQueueTest extends Specification {
...
+        } as BuildOperationQueue.QueueWorker<TestBuildOperation>
+
+        coordinationService = new DefaultResourceLockCoordinationService()
+        workerRegistry = new DefaultWorkerLeaseService(coordinationService, new DefaultWorkerLimits(1), ResourceLockStatistics.NO_OP)
+        workerRegistry.startProjectExecution(true)
+        // Main thread holds the only lease, so it is the only thread that can drain the queue.
+        lease = workerRegistry.startWorker()
+        // Drops spawned pool workers instead of running them, so workerCount can be driven
+        // above the limit without the queue being drained on another thread.
+        def discardingExecutor = new ManagedExecutorImpl(Executors.newSingleThreadExecutor(), new ExecutorPolicy.CatchAndRecordFailures()) {
```
jvandort: Can this just be a mock?

subprojects/core/src/main/java/org/gradle/internal/operations/DefaultBuildOperationQueue.java
```diff
@@ -310,10 +317,15 @@ private boolean shouldExitWithExtraWorkerInvalidation() {
...
-            if (helper.isExtraWorker()) {
-                if (token != null) {
-                    token.invalidateIfNeeded();
-                }
+            return tryRetireIfExtraWorker();
+        }
+
+        @GuardedBy("lock")
+        private boolean tryRetireIfExtraWorker() {
+            // If token is null, we are the main thread and shouldn't retire ourselves as we don't have a lease to give up.
```
jvandort: I'm not sure I'm a fan of the reasoning this comment states

yeah if token == null we are the main thread. And yeah we shouldn't retire ourselves. But I don't think its cause of worker leases. We might have a worker lease, we might not, but I wouldn't say returning early here is "giving it up"

I think the reasoning here is more like the main thread is guaranteed to execute after all work is complete (or as it is getting added if we leak the queue)

The reason we're not returning here is cause we can't guarantee any other workers will ever run, and given that, we cannot have the main thread exit early if no other workers may start

jvandort: Not sure how to encode this in a comment. I just am not really sure whats here captures the context of why we are adding the null check

subprojects/core/src/test/groovy/org/gradle/internal/operations/DefaultBuildOperationQueueTest.groovy
```diff
@@ -339,6 +339,48 @@ class DefaultBuildOperationQueueTest extends Specification {
...
+                // intentionally not run
+            }
+        }
+        def executionContext = new BuildOperationExecutionContext(discardingExecutor, 1, true)
+        operationQueue = new DefaultBuildOperationQueue(false, workerRegistry, executionContext, recordingWorker, null)
+
+        when:
+        operationQueue.add(new Success())           // workerCount -> 1
+        operationQueue.notifyBlockingWorkStarting()  // spawns extra worker, workerCount -> 2
+        operationQueue.notifyBlockingWorkFinished()  // max drops back to 1 => isExtraWorker() == true
```
jvandort: This reproducer is significantly different than what it would "look like" when using "public apis" (like `BuildOperationExecutor`, and `blocking`)

I worry as we rewrite this infrastructure, the subtlety of this test will not exactly reproduce the scenario we see here. 

In your existing rewrite PR, does this `notifyBlockingWorkStarting` concept still exist?

subprojects/core/src/test/groovy/org/gradle/internal/operations/DefaultBuildOperationQueueTest.groovy
octylFractal: Sure, I personally try to reduce mock usage but it is somewhat cleaner.

jvandort: Thats fine if you want to leave it as is

subprojects/core/src/test/groovy/org/gradle/internal/operations/DefaultBuildOperationQueueTest.groovy
octylFractal: Yes, we keep all the concepts. Though because of how submissions work in the refactor, this issue doesn't actually exist at all because we don't re-use the worker class to run the main thread bits. So this issue is kinda moot.

I can rewrite it to try and work it explicitly in terms of `runAll` only, with no internal API.

octylFractal: I did this, I think it's a lot better.

subprojects/core/src/test/groovy/org/gradle/internal/operations/DefaultBuildOperationExecutorParallelExecutionTest.groovy
```diff
@@ -349,6 +359,110 @@ class DefaultBuildOperationExecutorParallelExecutionTest extends ConcurrentSpec
...
+            BuildOperationDescriptor.Builder description() {
+                return BuildOperationDescriptor.displayName(name)
+            }
+        }
+    }
+
+    /**
+     * Creates executors that start only the first worker they are given and drop the rest.
+     */
+    static class RunFirstWorkerOnlyExecutorFactory implements ExecutorFactory {
```
jvandort: Can we use a normal executor if we submit work that indefinitely blocks until some condition?

jvandort (approved):

subprojects/core/src/test/groovy/org/gradle/internal/operations/DefaultBuildOperationExecutorParallelExecutionTest.groovy
octylFractal: I feel like that's more convoluted than this, but it is closer to the behavior we want to reproduce.

octylFractal: It actually wasn't more complex, so I did it

PR #38273, opened 2026-06-23
Make ResolutionResult CC compatible

### LINKED ISSUE #26897: Configuration Cache compatible api for `ResolutionResult#getAllComponents`
### Expected Behavior

Quite a lot of our tasks end up getting all the components from a `Configuration` and then doing something. However, the previous way we ended up doing this, which was to pass in a `Configuration` object as an input to the task, then call `configuration.getIncoming().getResolutionResult().getAllComponents()` inside the task no longer works. Instead, [you need to pass in a `Provider<ResolvedComponentResult>` as an input](https://docs.gradle.org/8.4/userguide/configuration_cache.html#config_cache:requirements:~:text=Referencing%20dependency%20resolution,invoking%20ResolutionResult.getRootComponent()) which you get from `configuration.getIncoming().getResolutionResult().`[`getRootComponent()`](https://docs.gradle.org/8.4/javadoc/org/gradle/api/artifacts/result/ResolutionResult.html#getRootComponent--).

The problem is that this just gives you the root - you need to do a search of the graph yourself to get all the components. This is somewhat tricky (especially when constraint platforms are involved) and quite a lot of code - we've had to extract it out to a library to avoid copy/pasting it everywhere. Ideally, there would be some method or way to iterate over all components from the root provided by Gradle. Or a `Provider<Set< ResolvedComponentResult>>` on `ResolutionResult` that can be passed to a task.

_No response_

### Context

As I upgrade our many Gradle plugins to Gradle 8, I'm trying to make them configuration cache compatible where it's very easy. This is a very common pattern for us that would ideally be easy to do in Gradle in configuration cache supported way.

jvandort:
Allow ResolutionResult to be serialized by the Configuration Cache. Its components, ResolvedComponentResult and AttributeContainer were already compatible, so making its also compatible is relatively trivial and gives users access to the utility methods on ResolutionResult during CC.

We also ensure it is properly snapshottable, meaning you can mark it as a task input

Fixes https://github.com/gradle/gradle/issues/26897

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheUnsupportedTypesIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheUnsupportedTypesIntegrationTest.groovy
@@ -40,7 +40,6 @@ import org.gradle.api.artifacts.dsl.DependencyLockingHandler
 import org.gradle.api.artifacts.dsl.RepositoryHandler
 import org.gradle.api.artifacts.query.ArtifactResolutionQuery
 import org.gradle.api.artifacts.repositories.ArtifactRepository
-import org.gradle.api.artifacts.result.ResolutionResult
 import org.gradle.api.artifacts.type.ArtifactTypeContainer
 import org.gradle.api.attributes.AttributeMatchingStrategy
 import org.gradle.api.attributes.AttributesSchema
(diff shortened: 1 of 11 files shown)


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test apt

big-guy (approved):

PR #38274, opened 2026-06-23
Isolate docsTest sample Gradle user home per worker to fix cache-lock deadlock

blindpirate:
## Problem

`:docs:docsTest` intermittently hangs until the build execution-timeout (e.g. the 90-minute cap on `Gradle_Master_Check_DocsTest_Linux`).

**Root cause (reproduced on `master`):** the snippet samples run in parallel across docsTest test workers, each forking a Gradle daemon, and they all share a single Gradle user home (`intTestHomeDir/distributions-full`). The parallel daemons deadlock on the cross-process file lock of that home's shared dependency/build caches:

```
org.gradle.cache.internal.DefaultFileLockManager$AwaitableFileLockReleasedSignal.await
  <- DefaultFileLockManager.acquireFileLock
  <- LockOnDemandCrossProcessCacheAccess.acquireFileLock
```

It only surfaces when the full sample suite actually executes — e.g. on branches that change `core-runtime` and bust the build cache. Cache-hit `master` CI builds are up-to-date and skip the samples, so the hang can look branch-specific when it isn't.

## Fix

Give each docsTest worker its own writable Gradle user home, so parallel workers never share a cache. A worker runs its samples sequentially, so for the rest of the run it still reuses its own daemon and now-warm caches (cheaper than per-sample isolation). When a shared dependency cache is already populated, it is served **read-only** so the isolated workers don't all re-resolve — guarded by an existence check so the RO cache is never pointed at a missing directory (Gradle otherwise disables it with a warning that pollutes the samples' asserted output).

This extends the per-sample isolation that already exists for `--build-cache` samples (for the build cache) to the dependency cache for the remaining samples.

### Lifecycle of the per-worker homes

They are siblings of the shared home, named `distributions-full-sample-worker-<test worker id>`:

- Named after the Gradle test worker id (`org.gradle.test.worker`) rather than the JVM pid, so the set of directory names is small and stable across builds.
- Deleted by `:docs:docsTest` before each run, so nothing accumulates in `intTestHomeDir` (which `clean` deliberately does not wipe) — including leftovers of an earlier crashed run.
- The `distributions-` name prefix keeps them inside the existing `CachesCleaner` sweep of `intTestHomeDir/distributions-*` as a backstop.
- Daemons are unaffected: the daemon registry is `build/daemon`, independent of the Gradle user home, so `DaemonTracker`/`KillLeakingJavaProcesses` still track and kill them as before.

## Verification

- Reproduced the deadlock locally on `master`: thread dumps show the cross-process artifact-cache file-lock waiters among parallel sample daemons sharing the one user home.
- With this change `:docs:docsTest` completes (~14.5 min, 2408 tests) instead of hanging.

**CI validation still needed:** local runs can't fully replicate CI (dependency mirror, worker count), so throughput with per-worker dependency resolution should be confirmed by CI here. The few `declaring-configurations-{android,kmp}` samples depend on `androidHomeWarmup`, which runs on CI but was excluded in the local runs.

### DIFF
--- a/platforms/documentation/docs/src/docsTest/java/org/gradle/docs/samples/IntegrationTestSamplesExecutor.java
+++ b/platforms/documentation/docs/src/docsTest/java/org/gradle/docs/samples/IntegrationTestSamplesExecutor.java
@@ -33,6 +33,7 @@ import org.gradle.test.fixtures.file.TestNameTestDirectoryProvider;
 import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.lang.management.ManagementFactory;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -106,6 +107,23 @@ class IntegrationTestSamplesExecutor extends CommandExecutor {
             //    Our operation:
             //    Lock file: /mnt/tcagent1/work/b6cfc23ab10332e6/intTestHomeDir/distributions-full/caches/build-cache-1/build-cache-1.lock
             executer.withGradleUserHomeDir(new File(workingDir, "user-home"));
+        } else {
+            // Snippet samples run in parallel across docsTest workers, each forking its own Gradle daemon.
+            // When those daemons share a single Gradle user home they contend on - and can deadlock on - the
+            // cross-process file lock of its shared dependency/build caches (LockOnDemandCrossProcessCacheAccess),
+            // which hangs :docs:docsTest until the build times out. Give each worker its own writable user home so
+            // parallel workers never share a cache. A worker runs its samples sequentially, so it still reuses that
+            // worker's daemon and warm caches.
+            File sharedUserHome = IntegrationTestBuildContext.INSTANCE.getGradleUserHomeDir();
+            String workerKey = ManagementFactory.getRuntimeMXBean().getName().replaceAll("[^A-Za-z0-9]", "_");
+            executer.withGradleUserHomeDir(new File(sharedUserHome.getParentFile(), sharedUserHome.getName() + "-sample-worker-" + workerKey));
+            // If a shared dependency cache is already populated, serve it read-only so the isolated workers don't
+            // each re-resolve everything. Only set it when the modules cache actually exists, otherwise Gradle
+            // disables the read-only cache and prints a warning that pollutes the sample's asserted output.
+            File sharedCaches = new File(sharedUserHome, "caches");
+            if (new File(sharedCaches, "modules-2").isDirectory()) {
+                executer.withReadOnlyCacheDir(sharedCaches);
+            }
         }
 
         if (flags.stream().anyMatch(NO_STACKTRACE_CHECK::equals)) {
--- a/platforms/documentation/docs/src/docsTest/java/org/gradle/docs/samples/IntegrationTestSamplesExecutor.java
+++ b/platforms/documentation/docs/src/docsTest/java/org/gradle/docs/samples/IntegrationTestSamplesExecutor.java
@@ -49,6 +49,18 @@ class IntegrationTestSamplesExecutor extends CommandExecutor {
 
     private static final String SAMPLE_ENV_PREFIX = "-Dorg.gradle.sampletest.env.";
 
+    /**
+     * The id of the Gradle test worker this JVM is, as set by {@code org.gradle.api.internal.tasks.testing.worker.TestWorker}.
+     * Declared here rather than referenced, as that class is not on the docsTest classpath.
+     */
+    private static final String WORKER_ID_SYS_PROPERTY = "org.gradle.test.worker";
+
+    /**
+     * Infix of the per-worker Gradle user home directory names. Keep in sync with the cleanup in the {@code docsTest}
+     * task in {@code platforms/documentation/docs/build.gradle.kts}, which deletes these directories before each run.
+     */
+    private static final String SAMPLE_WORKER_HOME_INFIX = "-sample-worker-";
+
     private final File workingDir;
     private final boolean expectFailure;
     private final GradleExecuter gradle;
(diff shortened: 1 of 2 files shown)


lkasso:
Thanks Bo! This is great.

blindpirate:
## CI performance validation

Ran the docs test configurations on this branch via manually triggered TeamCity builds, compared against master builds that actually executed the full sample suite (most master builds are build-cache hits that skip the samples, so only full executions are comparable).

| Configuration | This PR | Master baseline | Verdict |
|---|---|---|---|
| DocsTest Windows | [25.0 min, 2703 tests](https://builds.gradle.org/build/116848980) | [27.5 min, 2703 tests](https://builds.gradle.org/build/116766295) | ~2.5 min faster |
| DocsTest macOS | [11.4 min, 2618 tests](https://builds.gradle.org/build/116848990) | 10–15 min typical, 2618 tests | neutral |
| DocsTest Linux | [7.4 min, 2703 tests](https://builds.gradle.org/build/116890151) | [9.2 min, 2703 tests](https://builds.gradle.org/build/116890153) | ~1.8 min faster |
| DocsTest Linux (config cache) | [8.5 min, 2523 tests](https://builds.gradle.org/build/116890155) | [11.2 min, 2523 tests](https://builds.gradle.org/build/116890157) | ~2.7 min faster |

**Windows** is the configuration affected by the shared-home deadlock (gradle/gradle-private#5315), and it came in slightly *faster* than baseline with the full suite passing — the removal of cross-worker cache-lock contention appears to offset the per-worker cache warmup cost.

Conclusion: no measurable slowdown on any platform; the deadlock-prone configurations are slightly faster.

### Linux validation (Sep 1)

Linux was previously blocked: every attempt failed ~3 min in with `The Develocity server (gbt-td.grdev.net) denied the request to use Test Distribution`. Test Distribution is serving again, so both Linux variants were re-run.

Method — the two earlier attempts show why it matters:

1. **Predictive Test Selection.** A plain manual trigger on a PR branch enables PTS, which narrowed the run to 438 tests in 5.8 min — not comparable to the 2703-test master baselines. All runs below therefore use `-PrerunAllTests`, which both disables PTS and sets `doNotTrackState` on the test tasks, forcing the full suite to re-execute.
2. **Test Distribution fallback.** The first `-PrerunAllTests` attempt on this branch ([116888684](https://builds.gradle.org/build/116888684), 21.3 min) hit `gbt-td.grdev.net` returning 503 twenty-seven times and logged `Falling back to local execution due to connection timeout with Develocity`, so all 50 test classes ran locally on the agent. Its master counterpart six minutes later ([116888688](https://builds.gradle.org/build/116888688), 11.0 min) distributed 46 classes remotely. That 21.3-vs-11.0 gap is a TD outage artifact, not a property of this branch.

The table's Linux rows are from a paired re-run: PR (`ad984ec`) and master (`438758a`) triggered together on each configuration, all four with `-PrerunAllTests`, and each verified to have actually distributed (45–46 of 50 test classes remote) before being used.

Isolating `:docs:docsTest` from build overhead, using its task wall-clock time from the build scans:

| Configuration | This PR | Master | Δ |
|---|---|---|---|
| DocsTest Linux | 4.9 min ([scan](https://ge.gradle.org/s/6z3jmupce3z5u)) | 6.9 min ([scan](https://ge.gradle.org/s/m7dugqv5wtsys)) | −2.0 min |
| DocsTest Linux (config cache) | 5.7 min ([scan](https://ge.gradle.org/s/jf7b7v6lipiqi)) | 8.3 min ([scan](https://ge.gradle.org/s/3mtqlskjatups)) | −2.6 min |

Serial (total across all workers) test time is also lower on this branch — 115.1 min vs 127.8 min for DocsTest Linux, 113.3 min vs 120.0 min for the config-cache variant — so the per-worker Gradle user homes are not paying a re-resolution penalty; the read-only shared dependency cache appears to be doing its job.

Caveat: one paired run per configuration, and remote Test Distribution wall-clock varies with pool contention, so treat the exact deltas as indicative. The direction is consistent across both Linux variants and matches the Windows result.

blindpirate:
@bot-gradle test this

blindpirate:
@bot-gradle test this

platforms/documentation/docs/src/docsTest/java/org/gradle/docs/samples/IntegrationTestSamplesExecutor.java
```diff
@@ -106,6 +107,23 @@ private GradleExecuter createExecuter(List<String> args, List<String> flags) {
...
+        } else {
+            // Snippet samples run in parallel across docsTest workers, each forking its own Gradle daemon.
+            // When those daemons share a single Gradle user home they contend on - and can deadlock on - the
+            // cross-process file lock of its shared dependency/build caches (LockOnDemandCrossProcessCacheAccess),
+            // which hangs :docs:docsTest until the build times out. Give each worker its own writable user home so
+            // parallel workers never share a cache. A worker runs its samples sequentially, so it still reuses that
+            // worker's daemon and warm caches.
+            File sharedUserHome = IntegrationTestBuildContext.INSTANCE.getGradleUserHomeDir();
+            String workerKey = ManagementFactory.getRuntimeMXBean().getName().replaceAll("[^A-Za-z0-9]", "_");
+            executer.withGradleUserHomeDir(new File(sharedUserHome.getParentFile(), sharedUserHome.getName() + "-sample-worker-" + workerKey));
```
lkasso, line 119: Do these get cleaned up correctly?

blindpirate: Directory-wise and daemon-wise yes, but you spotted a real gap — pushed a fix in ebeb9bb.

What already worked: the per-worker homes are siblings of `distributions-full` under `intTestHomeDir`, so they are covered by `CachesCleaner.cleanUpCaches()`, which sweeps everything matching `distributions-*` in that dir. And daemons are unaffected: the daemon registry is `build/daemon`, independent of the Gradle user home, so `DaemonTracker`/`KillLeakingJavaProcesses` still track and kill them as before.

What did not: the worker key was the JVM's `pid@host`, so the names differed on every build, and `CachesCleaner` only prunes *within* a `distributions-*` dir — it never removes a whole stale one. On a reused CI checkout they would have piled up indefinitely.

So now:
* the home is named after the Gradle test worker id (`org.gradle.test.worker`) instead of the pid, so the set of names is small and reused across builds;
* `:docs:docsTest` deletes `intTestHomeDir/*-sample-worker-*` before each run, which also clears leftovers from a crashed run;
* the `distributions-` prefix stays, so the `CachesCleaner` sweep remains a backstop.

lkasso (approved):

PR #38288, opened 2026-06-23
Report each project's own failure in resilient model fetching

### LINKED ISSUE #38184: Failed IDE sync spends minutes in working state before giving feedback to the user

hegyibalint:
Fixes: https://github.com/gradle/gradle/issues/38184
Follows up: https://github.com/gradle/gradle/pull/38224
Triggered by: https://github.com/gradle/gradle/issues/38184#issuecomment-4768320421

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r970/ResilientSyncPerProjectFailureCrossVersionSpec.groovy
```diff
@@ -0,0 +1,148 @@
...
+class ResilientSyncPerProjectFailureCrossVersionSpec extends ToolingApiSpecification {
+
+    private static final String IP_ENABLED = "-Dorg.gradle.isolated-projects=true"
+    private static final String CONFIGURE_ON_DEMAND_ENABLED = "-Dorg.gradle.internal.isolated-projects.configure-on-demand=true"
+
+    def setup() {
+        settingsFile << "rootProject.name = \"root\"\n"
+    }
+
+    def "resilient sync isolates each sibling project's configuration failure (#description, #mode)"() {
```
hegyibalint: 🤦 Never seen Groovy's syntax highlighting this broken on GH.

hegyibalint:
@bot-gradle test this

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r970/CustomResilientModelCrossVersionSpec.groovy
jbartok, line 50: Please don't use this style of specifying file contents. Indent them in sources, then trim indent on the string. I, as a reader of the code, would very much appreciate that.

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
```diff
@@ -97,17 +97,18 @@ public ToolingModelBuilderResultInternal getModel(ToolingModelRequestContext mod
...
         @Override
         ToolingModelBuilderLookup.Builder locateBuilder() throws UnknownModelException {
-            // Force configuration of the target project to ensure all builders have been registered
-            Try<Void> projectConfiguration = ownerBuildConfiguration.isSuccessful()
-                ? tryRunConfiguration(targetProject::ensureConfigured)
-                : ownerBuildConfiguration;
+            boolean canRunEvenIfProjectNotFullyConfigured = canRunEvenIfProjectNotFullyConfigured(modelName);
+
+            // Configuring the target project itself scopes the failure to that project rather than the whole-build aggregate, which is what we want for per-project models.
+            Try<Void> projectConfiguration = canRunEvenIfProjectNotFullyConfigured && !ownerBuildConfiguration.isSuccessful()
```
jbartok: The change only configures the target project. 

A failure that is genuinely build-scoped, the canonical case is `gradle.projectsEvaluated { throw ... }`, but also any settings/afterEvaluate-at-build-scope hook, is recorded only in ownerBuildConfiguration, because every individual project still reaches Configured cleanly. ResilientToolingModelBuilder.build() returns the model with no failures attached. 

Before the PR, every per-project model carried that failure. Traced through `DefaultBuildController.fetch` → `DefaultBuildTreeModelCreator.getModel`: there is no other channel that re-surfaces it (the build action doesn't throw, since fetch catches per-query), so the failure is lost for the entire build regardless of how many projects the client queries. The project-attributable cases (broken project, broken ancestor) are correctly relocated, this gap is specifically failures that belong to no single project. 

Was this a deliberate decision: is dropping build-scoped failures acceptable, or should they still be attached somewhere?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r970/ResilientSyncPerProjectFailureCrossVersionSpec.groovy
```diff
@@ -0,0 +1,115 @@
...
+    private TestFile createProject(String path) {
+        settingsFile << "include('${path.substring(1)}')\n"
+        def projectDir = file(path.substring(1).split(':'))
+        projectDir.createDir()
+        return projectDir
+    }
+
+    private static Set<String> failureSourcesOf(result, String projectPath) {
+        def messages = result.failureMessagesByProject[projectPath]
+        return result.failureMessagesByProject.keySet().findAll { source ->
```
jbartok: Returns only which known project paths appear as a `FAILURE(<path>)` substring. If a regression attaches a spurious failure to a clean project whose message lacks the `FAILURE(` marker (a config-cache problem, a generic "A problem occurred configuring project", a serialization error), the assertion passes. Adding `result.failureMessagesByProject[path].isEmpty()` for the clean cases would close the hole.

jbartok (changes requested):

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
```diff
@@ -78,6 +78,7 @@ class CustomPlugin implements Plugin<Project> {
 """
     }
 
+    @TargetGradleVersion('>=9.4.0 <9.7.0')
```
asodja, line 81: ❌ We should not change behaviour of resilient sync. We should just change what exception is reported

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
asodja: ❌ This change allows to query models for project even if parent failed. Which is not something we should allow in normal mode without IP + configure-on-demand

What we want is basically to return a "per project" error instead of whole hierarchy of errors. So we would need to somehow map project id to an error in the exception that we return out of build state

asodja (changes requested):

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
hegyibalint: I see, the `gradle.projectsEvaluated { throw ... }` is an excellent counterexample for why this breaks down. 
To answer the question, we cannot lose failure information; this needs to be included in some way.

I will have a deeper look how this reproduces, and what is encoded in `ownerBuildConfiguration`.

hegyibalint: I see what you mean. I wanted to cut how much data we already produce on the daemon end, but you are right that this changes assumptions about what models can be fetched.

I am redoing this idea into an error-mapping solution, and keeping this behavior.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r970/ResilientSyncPerProjectFailureCrossVersionSpec.groovy
hegyibalint: Thanks, that is a good idea!

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
hegyibalint: I've implemented a new approach, and now resilient sync behaves as before; only the reported exception changed, as you suggested. 

The version gate stays necessary because these r940 tests assert the pre-9.7 (whole-build) exception shape, so they can't run against >=9.7.

hegyibalint:
@bot-gradle help

hegyibalint:
@bot-gradle test ACT without PTS

subprojects/core/src/test/groovy/org/gradle/internal/build/ResilientBuildToolingModelControllerTest.groovy
```diff
@@ -0,0 +1,48 @@
...
+        where:
+        description                                  | failure                                                   || buildScoped
+        "a project configuration failure"            | projectFailure()                                          || false
+        "a failure not tied to a project"            | new RuntimeException("boom")                              || true
+        "an aggregate of only project failures"      | aggregate(projectFailure(), projectFailure())             || false
+        "an aggregate mixing project and non-project"| aggregate(projectFailure(), new RuntimeException("boom")) || true
+        "an aggregate of only non-project failures"  | aggregate(new RuntimeException("boom"))                   || true
+        "an empty aggregate"                         | aggregate()                                               || false
+        "a nested aggregate hiding a non-project one"| aggregate(aggregate(projectFailure(), new RuntimeException("boom"))) || true
+        "a nested aggregate of only project failures"| aggregate(aggregate(projectFailure()))                    || false
```
hegyibalint: ℹ️ Not all of these would be real-world scenarios (IIUC `ProjectConfigurationException` should not happen together with non-`ProjectConfigurationException` exceptions), but worth unit testing for completeness.

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
hegyibalint: I think this is addressed by attaching non-project-associated problems (see `ResilientBuildToolingModelController#hasBuildScopedFailure`). They can be surfaced once on the root project.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/CustomResilientModelCrossVersionSpec.groovy
asodja: I looked at r970 tests and I see that if b project is broken, you can still query custom model for root and a. But that should not be allowed

Or am I missing something?

hegyibalint:
To make the merging smoother, this PR got stacked on top of https://github.com/gradle/gradle/pull/38288

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
```diff
@@ -122,7 +127,10 @@ public ToolingModelScopeResult getModel(ToolingModelRequestContext modelRequestC
                 Object model = canRunEvenIfProjectNotFullyConfigured(modelName)
                     ? Try.ofFailable(() -> buildModelWithParameter(parameter)).getOrMapFailure(failure -> null)
                     : null;
-                return configurationFailureResult(failureFactory, projectConfiguration.getFailure().get(), model);
+                // The client sees this project's own recorded failure, not the whole-build aggregate that is otherwise
```
asodja: 💅 Does this comment add any extra value?

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
```diff
@@ -136,6 +144,18 @@ public ToolingModelScopeResult getModel(ToolingModelRequestContext modelRequestC
             });
         }
 
+        /**
+         * The target project's own recorded configuration failure, read from state without reconfiguring it.
+         * Falls back to a general failure when the project has none of its own: it configured cleanly, or was never reached because another project aborted configuration.
+         */
+        private Throwable projectOwnConfigurationFailure() {
+            Throwable ownFailure = targetProject.getMutableModelEvenAfterFailure().getState().getFailure();
```
asodja, line 154: ❓ does this contain just a project error or it can have longer chain of errors

hegyibalint: I'm quite certain. If you start to follow the `getFailure()`, you will end up in the method that sets this failure:

https://github.com/gradle/gradle/blob/63d4fa7cbaa87b7bb0ff0046624dd1727c8752d6/subprojects/core/src/main/java/org/gradle/configuration/project/LifecycleProjectEvaluator.java#L108-L115

I think this will try to configure this individual project, and if it fails, will only encode its own problem; but nothing more.

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
hegyibalint: Maybe it's too terse, but I think this might be worth a rephrase.

asodja (approved):
LGTM

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r970/CustomResilientModelCrossVersionSpec.groovy
hegyibalint: I've indented every snippet 👍

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.groovy
hegyibalint, line 1: ℹ️ The aim of this PR is to change the exception messages, but not the underlying behavior of the resilient sync. Because of that, I've found it dubious to replicate this cross-version test, as next to the messages, this test also establishes a lot of behavior that shouldn't have changed.

The test got a built-in way to handle exception messages by providing a (from, to) version range.

hegyibalint:
@bot-gradle test ACT without PTS

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.groovy
```diff
@@ -825,18 +849,27 @@ class ResilientKotlinDslScriptsModelBuilderCrossVersionSpec extends KotlinDslPlu
...
-        "buildSrc-included/build.gradle.kts"    | ["settings.gradle.kts", "buildSrc/settings.gradle.kts", "buildSrc-included/settings.gradle.kts", "buildSrc-included/build.gradle.kts"] | [Pair.of(".", ".*Build file.*build\\.gradle\\.kts.*Script compilation error.*"), Pair.of("buildSrc-included", ".*Build file.*build\\.gradle\\.kts.*Script compilation error.*")]                                                                                      | "with isolated projects"   | IP_FLAGS
+        fromVersion | untilVersion | brokenFile                              | scriptModels                                                                                                                           | expectedFailures                                                                                                                                               | mode                     | extraGradleProperties
```
jbartok, line 863: This feels inconsistent with how we normally do such things, I would expect the "from 9.7 until later" bits to be in the "r970" test class and the "from 9.4 to 9.7" part to be here.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.groovy
```diff
@@ -825,18 +849,27 @@ class ResilientKotlinDslScriptsModelBuilderCrossVersionSpec extends KotlinDslPlu
...
+        "9.7"       | null         | "buildSrc/build.gradle.kts"             | ["settings.gradle.kts", "buildSrc/settings.gradle.kts", "buildSrc/build.gradle.kts", "buildSrc-included/settings.gradle.kts"]          | [Pair.of(".", GENERAL_CONFIGURATION_FAILURE), Pair.of("buildSrc", BUILD_SCRIPT_COMPILE_ERROR)]                                                                 | ""                       | NO_EXTRA_PROPERTIES
+        "9.7"       | null         | "buildSrc-included/settings.gradle.kts" | ["settings.gradle.kts"]                                                                                                                | [Pair.of(".", GENERAL_CONFIGURATION_FAILURE), Pair.of("buildSrc", SETTINGS_SCRIPT_COMPILE_ERROR), Pair.of("buildSrc-included", GENERAL_CONFIGURATION_FAILURE)] | ""                       | NO_EXTRA_PROPERTIES
```
jbartok: Nitpick from Claude:

The GENERAL message has exactly one producer, projectOwnConfigurationFailure(), reachable only in the project-was-created branch. A settings-script compile failure means that build's root project is never created, so the fetch resolves through the PR-unchanged not-created paths (lines 79 and 119), which surface the actual SETTINGS_SCRIPT_COMPILE_ERROR. 

The row is internally self-contradictory — it expects the clean buildSrc key to report the real settings error but the broken buildSrc-included key to report GENERAL — and it disagrees with both its own 9.4 counterpart (line 858, expects SETTINGS) and the non-version-split sibling test at 694-696. 

Likely uncaught because the 9.7 cross-version leg hasn't run (CI shows only a "Ready for Release" trigger, not an executed tooling-api cross-version job). 

Fix: expect Pair.of("buildSrc-included", SETTINGS_SCRIPT_COMPILE_ERROR).

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
jbartok: If there is a failure in a subproject, that will be lost. Instead of saying "A problem occurred configuring project ':b'." now we'll just have a generic "could not be configured". Is that intended?

jbartok: Also, please test what happens if a failure originates in a `beforeProject` hook, rather than a script body. Likely `getFailure()` will not see it.

jbartok: Also, consider putting a `getConfigurationFailure()`/`hasConfigurationFailure()` on `ProjectState` and using that. The concept of "this project's own configuration failure" already exists one layer down, in `ProjectStateInternal`, just not exposed. Would be less brittle than navigating 3 layers to get to the failures.

subprojects/core/src/main/java/org/gradle/internal/build/ResilientBuildToolingModelController.java
```diff
@@ -139,6 +145,18 @@ public ToolingModelScopeResult getModel(ToolingModelRequestContext modelRequestC
...
+            if (ownFailure != null) {
+                return ownFailure;
+            }
+            return new GradleException("The build could not be configured; see the reported build failures for the underlying problems.");
```
jbartok: Claude says this is wasteful:

That exception is a plain GradleException(String), which chains to RuntimeException and captures a native stack trace via fillInStackTrace() at construction — there's no stackless constructor in play. Then the trace gets touched a second time downstream: configurationFailureResult hands the exception to failureFactory.create(...), and DefaultFailureFactory.create() calls getStackTrace() (which materializes the native backtrace into a StackTraceElement[]), copies it into an immutable list, and classifies every frame.

So for one message-only exception you pay three things:
1. fillInStackTrace() — native backtrace capture at construction,
2. getStackTrace() — decode that backtrace into a StackTraceElement[],
3. copy + per-frame classification of the whole array.

And none of it is ever useful: the captured frames point only into projectOwnConfigurationFailure → getModel (Gradle-internal code, not user code), the exception has no cause, and its stack is never shown to anyone — the real, useful failure travels separately as buildFailure. The client only ever needs the message text (which is exactly what the tests assert).

jbartok: - Make it stackless — a RuntimeException subclass constructed with super(msg, null, false, false) (disables writable stack trace), or overriding fillInStackTrace() to return this. That collapses all three costs to ~zero.

- Make it a shared instance — the value is invariant (constant message, no per-project data), so hoist it to a single private static final Throwable and return that same object instead of allocating a fresh one per project. This is safe: failureFactory.create() uses its own per-job IdentityHashMap "seen" set, so sharing one immutable throwable across projects doesn't collide.

jbartok (commented):
Looks better than before, see my comments.

hegyibalint:
@jbartok could you please follow the reviewing cheatsheet format? Otherwise, I cannot gauge the severity of your comments.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.groovy
hegyibalint: See my comment above the file about the reasoning why this was not moved into r970. I don't feel your comment addresses that.

Also, this is not without precedent. Please see 
- [`r940/ResilientGradleBuildBuilderCrossVersionSpec.groovy#L117`](https://github.com/gradle/gradle/blob/4e7604b8edd93542d46f5584e20913acd150e9c2/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientGradleBuildBuilderCrossVersionSpec.groovy#L117) — expected failure texts branch on `targetVersion >= 9.5` inside the r940 class (from #38240 itself); see also [`#L214`](https://github.com/gradle/gradle/blob/4e7604b8edd93542d46f5584e20913acd150e9c2/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientGradleBuildBuilderCrossVersionSpec.groovy#L214) (`expectedFailures.call(targetVersion)` from the data table) and [`#L257-L262`](https://github.com/gradle/gradle/blob/4e7604b8edd93542d46f5584e20913acd150e9c2/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientGradleBuildBuilderCrossVersionSpec.groovy#L257-L262) (message selected per 9.6 boundary)
- [`r73/DeprecationsCrossVersionSpec.groovy#L30-L36`](https://github.com/gradle/gradle/blob/4e7604b8edd93542d46f5584e20913acd150e9c2/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r73/DeprecationsCrossVersionSpec.groovy#L30-L36) — an r73 spec selecting deprecation wording across the 8.0 and 8.2 boundaries
- [`r930/FetchBuildActionCrossVersionSpec.groovy#L61`](https://github.com/gradle/gradle/blob/4e7604b8edd93542d46f5584e20913acd150e9c2/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/FetchBuildActionCrossVersionSpec.groovy#L61) — an r930 spec adjusting expected failure causes around the 8.7 boundary (also [`#L143`](https://github.com/gradle/gradle/blob/4e7604b8edd93542d46f5584e20913acd150e9c2/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/FetchBuildActionCrossVersionSpec.groovy#L143), [`#L246`](https://github.com/gradle/gradle/blob/4e7604b8edd93542d46f5584e20913acd150e9c2/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r930/FetchBuildActionCrossVersionSpec.groovy#L246))
- [`r88/TestDisplayNameJUnit4CrossVersionSpec.groovy#L133`](https://github.com/gradle/gradle/blob/4e7604b8edd93542d46f5584e20913acd150e9c2/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r88/TestDisplayNameJUnit4CrossVersionSpec.groovy#L133) — an r88 spec branching the expected test-event tree on the 9.3 boundary
- [`r51/TaskOriginCrossVersionSpec.groovy#L66`](https://github.com/gradle/gradle/blob/4e7604b8edd93542d46f5584e20913acd150e9c2/platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r51/TaskOriginCrossVersionSpec.groovy#L66) — an r51 spec branching the expected origin plugin on the 7.2 boundary

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r940/ResilientKotlinDslScriptsModelBuilderCrossVersionSpec.groovy
hegyibalint: Good catch. You were right about the expectation, and the test now asserts exactly that (`57e8c353df9`). The mechanism turned out to be different from what both of us assumed, though.

The GENERAL entry was never produced by `buildSrc-included`'s fetch, because that fetch never happened. `GradleBuildBuilder` does not synthesize a root project for a broken build, only for healthy builds that merely reference broken settings; for a broken build it sets just the build identifier ([`GradleBuildBuilder.java#L154-L181`](https://github.com/gradle/gradle/blob/57e8c353df916647927b03bf75272fde8550735f/platforms/ide/ide/src/main/java/org/gradle/plugins/ide/internal/tooling/GradleBuildBuilder.java#L154-L181)). Our test action's `build.getRootProject()` therefore returned null, and `controller.fetch(null, ...)` (used in [`KotlinModelAction.java#L87`](https://github.com/gradle/gradle/blob/57e8c353df916647927b03bf75272fde8550735f/platforms/ide/tooling-api/src/crossVersionTestModels/java/org/gradle/integtests/tooling/r940/KotlinModelAction.java#L87)) silently resolved to the default target, i.e., the root build's default project ([`DefaultBuildController.java#L136-L148`](https://github.com/gradle/gradle/blob/57e8c353df916647927b03bf75272fde8550735f/platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/DefaultBuildController.java#L136-L148)). The general failure we asserted was the root build's answer, recorded under the broken build's key by the action itself. The not-created path you pointed at works exactly as you described once the fetch is actually routed to the broken build.

The fix is therefore in our test model action, not in production code: when the `GradleBuild` model carries no root project, the action targets the build itself, and the existing not-created path reports the build's own settings failure ([`KotlinModelAction.java#L75-L84`](https://github.com/gradle/gradle/blob/57e8c353df916647927b03bf75272fde8550735f/platforms/ide/tooling-api/src/crossVersionTestModels/java/org/gradle/integtests/tooling/r940/KotlinModelAction.java#L75-L84)). In other words, it is up to the TAPI client to fall back to a build target for settings-broken builds; our action is the reference for that pattern.

This is not trivial(a nullable `getRootProject()` plus a silent default-target fallback on `fetch(null, ...)` means a naive client gets a wrong answer instead of an error). This might be worth some documentation.

hegyibalint:
There are some strange comments here @jbartok which I cannot reply to. Strange.

Let me address them here:

https://github.com/gradle/gradle/pull/38288#discussion_r3533987262
> If there is a failure in a subproject, that will be lost. Instead of saying "A problem occurred configuring project ':b'." now we'll just have a generic "could not be configured". Is that intended?

_If there is a failure in a subproject, that will be lost_ is not true. The fact is that the information is available in the query of that specific project, and in the build failure when the build finishes. This PR removed this redundant information from other projects, because in this way we can reduce the amount of data we generate when a resilient model query fails; this solution completely linearizes this equation, as we return $O(P)$ amount of project failures maximum.

My limitation was the deal we made with @asodja: we didn't want to change when resilient sync succeeds or fails, only what failure payload each query carries. 

---

https://github.com/gradle/gradle/pull/38288#discussion_r3533999663
> Also, please test what happens if a failure originates in a beforeProject hook, rather than a script body. Likely, getFailure() will not see it.

\+

https://github.com/gradle/gradle/pull/38288#discussion_r3534085738
> Also, consider putting a getConfigurationFailure()/hasConfigurationFailure() on ProjectState and using that. The concept of "this project's own configuration failure" already exists one layer down, in ProjectStateInternal, just not exposed. Would be less brittle than navigating 3 layers to get to the failures.

Good ideas, let me see if I can widen the test coverage. 
UPDATE: I think all of these are covered in https://github.com/gradle/gradle/pull/38288/changes/badee02d87ad7c5cf70c2f8665356a4ab94edf19

---

https://github.com/gradle/gradle/pull/38288#discussion_r3534123769
> Claude says this is wasteful

You gave a good point that the exception is not reaching the user boundary. Yeah, a stackless RuntimeException might be a good choice here.
UPDATE: Implemented as a new `GeneralConfigurationFailure`, though not a shared one, as the `StaticAssignmentOfThrowable` ErrorProne check blows up if I do that. I think this is more than enough.

jbartok (approved):
LGTM

PR #36809, opened 2026-02-19
Fix IvyModuleArtifactResolutionIntegrationTest for Configuration Cache

tresat:
`ARQ` support added here.

PR #38294, opened 2026-06-23
Add regression test for issue 27582

### LINKED ISSUE #27582: Clarify if ResolvedArtifactResult is serializable
### Issue type

### Problem description

The configuration cache documentation suggests that `ArtifactResult` may be referenced as long as a lazy provider is used but that doesn't seem to be the case:
```
- Task `:test:demo` of type `DemoTask`: cannot serialize object of type 'org.gradle.api.internal.artifacts.result.DefaultResolvedArtifactResult', a subtype of 'org.gradle.api.artifacts.result.ArtifactResult', as these are not supported with the configuration cache.
```

Even more confusing, when the configuration contains a dependency to a local project, serialization does *not* fail anymore.

### Context (optional)

```groovy
plugins {
    id 'java'
}

dependencies {
    implementation "commons-cli:commons-cli:1.6.0"
    // removing the following dependency makes the serialization fail
    implementation project(":some-other-project")
}


tasks.register("demo", DemoTask) {
    artifacts = configurations.runtimeClasspath.incoming.artifacts.resolvedArtifacts
}

abstract class DemoTask extends DefaultTask {
    @Internal
    abstract SetProperty<ResolvedArtifactResult> getArtifacts()

    @TaskAction
    void print() {
        artifacts.get().each {
            println it.file
        }
    }
}

```

### Page with the problem

https://docs.gradle.org/8.5/userguide/configuration_cache.html#config_cache:requirements:disallowed_types

tresat:
Issue #27582 reported that `ArtifactCollection.getResolvedArtifacts()`, when used as a `SetProperty<ResolvedArtifactResult>` task input, failed configuration-cache storage with "cannot serialize `DefaultResolvedArtifactResult`" — but only when the underlying configuration had no task dependencies. Adding a project dependency masked the failure because BuildableBackedProvider.calculate\ ExecutionTimeValue() takes the deferred changingValue(...) branch when hasDependencies() is true.

The literal failure stopped reproducing in `11e032ad1920` (PR #36809, March 2026), which added codecs for `ResolvedArtifactResult`, `ComponentArtifactsResult`, and `UnresolvedComponentResult` as a side effect of fixing `IvyModuleArtifactResolutionIntegrationTest`. The issue stayed open because no regression test was tied to its scenario.

This commits three integration tests as a regression net:
- the issue's snippet verbatim, with both external and project deps;
- the failing variant from the issue (external-only) that exercises the no-task-dependencies branch via the java plugin;
- a bare resolvable configuration with a single external dep, which is the most direct exercise of the no-task-dependencies code path (`BuildableBackedProvider.calculateExecutionTimeValue`'s `fixedValue(get())` branch) without java-plugin scaffolding.

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ResolvedArtifactsWithConfigurationCacheIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ResolvedArtifactsWithConfigurationCacheIntegrationTest.groovy
@@ -1,0 +1,224 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.integtests.resolve
+
+import org.gradle.integtests.fixtures.AbstractHttpDependencyResolutionTest
+import org.gradle.test.precondition.Requires
+import org.gradle.test.preconditions.TestExecutionPreconditions
+import spock.lang.Issue
+
+/**
+ * Tests configuration-cache serialization of the provider returned by
+ * {@code ArtifactCollection.getResolvedArtifacts()} when used as a task input
+ * of type {@code SetProperty<ResolvedArtifactResult>}.
+ * <p>
+ * The historical defect (issue 27582) was asymmetric: configurations with no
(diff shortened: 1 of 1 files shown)


platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ResolvedArtifactsWithConfigurationCacheIntegrationTest.groovy
```diff
@@ -0,0 +1,223 @@
...
+                id 'java'
+            }
+        """
+
+        when:
+        succeeds "demo"
+
+        then:
+        configurationCache.assertStateStored()
+        outputContains("commons-cli-1.6.0.jar")
```
big-guy, line 214: Should this check if the project artifact is available too?

big-guy (approved):

platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/ResolvedArtifactsWithConfigurationCacheIntegrationTest.groovy
tresat: Probably a good idea.

blindpirate:
@bot-gradle squash and merge

Issue #30505, opened 2024-09-11
Expose dependencies on `ResolvedVariantResult`


jvandort:
### Expected Behavior

Variant dependencies are exposed on `ResolvedVariantResult`

### Current Behavior (optional)

Dependencies are only accessible from `ResolvedComponentResult#getDependenciesForVariant(ResolvedVariantResult)`.

### Context

The existing component-based API is clunky, making it tedious to traverse a dependency graph, as both the component and the variant must be tracked.

We should expose the variant's dependencies on the variant itself so that we do not need to keep a reference to the component when traversing a graph.

There is some weirdness with `ResolvedArtifactResult`, is it also exposes a `ResolvedVariantResult`, but a nonstandard version of it -- one that describes the artifact. The data exposed by that field does _not_ describe a variant but rather an artifact of a variant. For that reason, we should deprecate `ResolvedArtifactResult#getVariant` and instead expose attributes and capabilities directly on `ResolvedArtifactResult`. In the meantime, `getDependencies` on `ResolvedVariantResult` should either return an empty collection or throw an exception for the variant instance returned by `ResolvedArtifactResult`.

PR #38295, opened 2026-06-23
Add getAttributes getCapabilities to ResolvedArtifactResult

jvandort:
Currently to get these for the artifact you need to go through getVariant which returns you a ResolvedVariantResult. However, this is a poor API. A ResolvedVariantResult represents a node in a dependency graph. The attributes and capabilities accessible via ResovledArtifactResult describe the artifacts _derived_ from a variant of a graph, which may have different attributes than the source node itself.

The existing API is confusing, as we expose attribute and capabilities as if they describe a variant when in fact they describe the artifact. The ResolvedVariantResult returned by getVariant is _not_ the variant that produced the artifact set. Even if we wanted to expose that variant on the artifact set, we could not just change the behavior of the existing getVariant method.

So, we expose the necessary information directly on ResolvedVariantResult -- getAttributes and getCapabilties. Eventually, we will deprecated getVariant from ResolvedArtifactResult, but we will need to give time for the community to start using these new methods before doing so

Related: https://github.com/gradle/gradle/issues/30505

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/ConfigurationCacheCodecs.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/ConfigurationCacheCodecs.kt
@@ -271,7 +271,7 @@ class DefaultConfigurationCacheCodecs(
             bind(WorkNodeActionCodec)
             bind(CapabilitySerializer())
             bind(DefaultComponentArtifactsResultCodec())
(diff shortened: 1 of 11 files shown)


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

lkasso (approved):
Docs LGTM.

big-guy (approved):

jvandort:
@bot-gradle test apt

PR #38301, opened 2026-06-24
Adjust CLI flag to configure non-interactive console

donat:
The --non-interactive flag shipped in 9.6.0 with no Gradle property. It was also a negative-only flag, inconsistent with Gradle's positive-flag + --no- convention.

Since the feature is incubating, reshape it: add --interactive and --no-interactive backed by the org.gradle.console.interactive property, and drop --non-interactive. Rename the incubating LoggingConfiguration and StartParameter API from isNonInteractive/setNonInteractive to isInteractive/setInteractive.

Fixes https://github.com/gradle/gradle/issues/38270

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #38303, opened 2026-06-24
Restore --non-interactive flag instead of --interactive/--no-interactive

donat:
PR #38301 introduced --interactive/--no-interactive to toggle interactive console prompting. Following review, restore the original --non-interactive flag (shipped in 9.6.0), now marked incubating, and drop the newly added --interactive/--no-interactive flags.

The org.gradle.console.interactive Gradle property (true/false) remains as the property-based control, and the incubating isInteractive/setInteractive API is unchanged.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


donat:
@bot-gradle test RFM

ljacomet (approved):

PR #38304, opened 2026-06-24
Generate a removal report

ljacomet:
* Report organized by removal methods (`willBeRemovedInGradle10`, `willBecomeAnErrorInGradle10`, ...)
* Ability to filter by project, or by team - using CODEOWNERS as the truth
* Smart source anchoring on the most recent "published" commit
* Designed to be generic - that is should not need modifications for Gradle 11


build-logic/buildquality/src/main/kotlin/gradlebuild/removal/action/CodeOwners.kt
```diff
@@ -0,0 +1,91 @@
...
+                Rule(patternToRegex(tokens.first()), teams)
+            }
+        )
+
+        private val WHITESPACE = Regex("\\s+")
+
+        /** Translates a CODEOWNERS pattern into a regex matching any repo-relative file path it owns. */
+        private fun patternToRegex(pattern: String): Regex {
+            var p = pattern
+            val anchored = p.startsWith("/")
```
blindpirate: Anchoring only triggers on a leading `/`, but in gitignore/CODEOWNERS any pattern with a **non-trailing** slash is root-anchored. So `platforms/jvm/` floats here and also matches a nested `subprojects/foo/platforms/jvm/Bar.java` — wrong team. Only trailing-slash-only patterns like `gradle/` should float.

```kotlin
val anchored = pattern.startsWith("/") || p.trim('/').contains("/")
```

blindpirate: We have such a line in owners:

> gradle/                                 @gradle/bt-developer-productivity

This will make it matching all paths including `gradle`, like `some-project/src/main/org/gradle/`.

blindpirate (approved):

PR #38311, opened 2026-06-25
Throw when default plugin version is set after projects are loaded

reinsch82:
Part of #38028

### Context

`DefaultPluginResolutionStrategy` is a build-scoped service that is injectable into project configuration. Under Isolated Projects, projects configure in parallel, and `applyTo()` reads the default-version map (`pluginVersions`) concurrently while that happens. The strategy is therefore meant to be *frozen* once projects are loaded.

`eachPlugin()` already enforces this freeze by throwing after `projectsLoaded`, but `setDefaultPluginVersion()` did not — it mutated the map unconditionally, leaving the freeze to rely purely on caller discipline. That is a late, shared-state mutation that can race the concurrent reads in `applyTo()`.

This PR:

- Throws `IllegalStateException` when `setDefaultPluginVersion()` is called after projects have been loaded, matching how `eachPlugin()` already enforces the same freeze — so a late write can never race the concurrent reads in `applyTo()`.
- Makes the `locked` flag `volatile` (it is set on one thread and read on configuration threads).

The intended caller is the settings-time `pluginManagement { plugins { ... } }` DSL, which always runs before the lock. The only way to reach the post-lock path is to capture the `pluginManagement` spec at settings time and call `version()` later (e.g. from a `projectsLoaded` callback) — previously a silent write that was unsafe under Isolated Projects, now an error in all modes. `setDefaultPluginVersion` backs an incubating feature, so failing loudly on this misuse is acceptable.

### Contributor Checklist
- [ ] Make sure all contributed code can be distributed under the terms of the [Apache License 2.0](https://github.com/gradle/gradle/blob/master/LICENSE).
- [ ] Update User Guide, DSL Reference, and Javadoc for public-facing changes. _(N/A: internal API.)_
- [x] Ensure that tests pass locally: `./gradlew :plugin-use:test :plugin-use:forkingIntegTest :plugin-use:configCacheIntegTest`.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/extensibility/plugin-use/build.gradle.kts
```diff
@@ -29,6 +30,7 @@ dependencies {
     implementation(projects.jvmServices)
     implementation(projects.startParameter)
 
+    implementation(libs.kotlinStdlib)
```
alllex: 🤔  Is `kotlin.Unit` the only thing that is causing this dependency?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsPluginResolutionStrategyIntegrationTest.groovy
```diff
@@ -0,0 +1,72 @@
...
+        def strategy = (gradle as org.gradle.api.internal.GradleInternal).services
+            .get(org.gradle.plugin.management.internal.PluginResolutionStrategyInternal)
+        strategy.setDefaultPluginVersion(org.gradle.plugin.use.internal.DefaultPluginId.of("com.example.foo"), "1.0")
+    '''
+
+    private static final String EXPECTED_VIOLATION = "Cannot set a default plugin version after projects have been loaded when Isolated Projects is enabled. The default version for plugin 'com.example.foo' was set too late."
+
+    def "reports a violation when a default plugin version is set after projects are loaded"() {
+        given:
+        buildFile SET_DEFAULT_VERSION_DURING_CONFIGURATION
```
6hundreds: 💅 Let's be consistent over definitions in tests - either we use parenthesis or not.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsPluginResolutionStrategyIntegrationTest.groovy
```diff
@@ -0,0 +1,72 @@
...
+
+        expect:
+        succeeds("help")
+    }
+
+    def "reports a violation when a captured plugin management spec sets a version after projects are loaded"() {
+        given:
+        settingsFile '''
```
6hundreds: 💅 Double quotes for consistency with the rest of the definitions

platforms/extensibility/plugin-use/build.gradle.kts
reinsch82: yes, I'd still keep it, it's ubiquitous anyway.

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/management/internal/DefaultPluginResolutionStrategy.java
```diff
@@ -65,6 +69,19 @@ public PluginRequestInternal applyTo(PluginRequestInternal pluginRequest) {
...
     @Override
     public void setDefaultPluginVersion(PluginId id, String version) {
+        if (locked) {
+            problems.report(factory ->
+                factory.problem(null, messageBuilder -> {
+                    messageBuilder
+                        .text("Cannot set a default plugin version after projects have been loaded ")
+                        .text("when Isolated Projects is enabled. The default version for plugin ")
+                        .reference(id.toString())
+                        .text(" was set too late.");
```
octylFractal: 🤔 This seems a little clunky, what about "Cannot set a default plugin version for plugin '<id>' after projects have been loaded when Isolated Projects is enabled."?

platforms/extensibility/plugin-use/src/main/java/org/gradle/plugin/management/internal/DefaultPluginResolutionStrategy.java
```diff
@@ -65,6 +69,18 @@ public PluginRequestInternal applyTo(PluginRequestInternal pluginRequest) {
 
     @Override
     public void setDefaultPluginVersion(PluginId id, String version) {
+        if (locked) {
+            problems.report(factory ->
```
alllex: ❓  Can we just throw an IllegalArgument here instead of reporting the IP problem?

PR description states:
>The post-lock path is unreachable today — the sole production caller is the settings-time pluginManagement { plugins { ... } } DSL, which always runs before the lock — so this is a defensive guard that makes the freeze enforced rather than assumed.

If we never call this after projects have loaded and there is not public API to do so, then we don't need to go the IP route.

octylFractal (approved):

Issue #37942, opened 2026-05-20
Gradle 9.5.0: Accessor classes are generated for source set configurations, but using the configurations without quotes fails the build


enbirr:
### Current Behavior

I found that Gradle 9.5.0 behaves differently from the prior versions.

Steps to reproduce:
- Create a source set: 
```
val customSourceSet by sourceSets.registering() {}
```
- Sync the project in IDEA
- Write any configuration for the source set:
```
dependencies {
    customSourceSetImplementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
}
```
- With Gradle 9.5.0, IDEA doesn't highlight any errors for `customSourceSetImplementation`. Moreover: 
  - `Command + Click` successfully navigates to the accessor class - to the same directory as for accessors for other configurations, like implementation. 
  - IDEA could complete `customSourceSet<caret>` to `customSourceSetImplementation`. 
  - BUT: syncing a project, or calling `./gradlew build` fails:
```
What went wrong:
Script compilation error:
  Line 16:     customSourceSetImplementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
               ^ Unresolved reference 'customSourceSetImplementation'.
```

### Expected Behavior

With Gradle 9.4.0, `customSourceSetImplementation` is immediately highlighted as unresolved - because there is no accessor class for the configuration. This is the expected behavior.

### Context (optional)

I am refining the configuration completion in IntelliJ IDEA, in the dependencies block of `build.gradle.kts`. For this purpose, it's important to determine if a dependency configuration could be resolved, before the completion inserts the configuration into a build script. I wanted to rely on the presence of accessor classes for this, and it works for prior versions - but not for Gradle 9.5.0.

### Self-contained Reproducer Project

It's a simple project without buildSrc, included builds and precompiled plugins.
[reproduce-unresolved-source-set-configuration.zip](https://github.com/user-attachments/files/28055961/reproduce-unresolved-source-set-configuration.zip)

### Gradle version

9.5.0

### Your Environment (optional)

IntelliJ IDEA 2026.2 EAP

enbirr:
Dear Gradle team,
[...]

PR #38317, opened 2026-06-25
Make Kotlin DSL accessor hash order-independent

### LINKED ISSUE #37719: KotlinDSL accessors hash is not stable under IP
IP produces different KotlinDslScriptsModel accessor paths due to non-deterministic project schema.

Under Isolated Projects, the `kotlin-dsl/accessors/<hash>/` directory hash differs from non-IP runs, and is also non-deterministic across IP runs themselves. This makes the accessor paths in `classPath/sourcePath` of `KotlinDslScriptModel` unstable.

The accessor directory hash is the identity of `GenerateProjectAccessors` unit of work, computed from:
  1. Project schema (extensions, tasks, configurations, container elements)
  2. DCL enabled flag
  3. Classpath fingerprint

  The project schema differs under IP for included build-logic projects in three ways:
  1. Non-deterministic task ordering — tasks are registered in different order across IP runs due to parallel project configuration. Two separate IP runs produce different task lists (e.g., test task
  moves position). Since `hashCodeFor(schema)` is order-dependent, this alone makes the hash unstable.
  2. Extra extensions — IP run registers `kotlinScripting` extension that non-IP doesn't (42 vs 44 extensions).
  3. Extra configurations — IP run produces `apiElements-published` and `runtimeElements-published` configurations that non-IP doesn't (38 vs 40 configurations).

Having said that, any test comparing IP vs non-IP KotlinDslScriptModel must normalize kotlin-dsl/accessors/<hash>/ paths. The models are structurally equivalent — same accessors are generated, they just live in different directories.

6hundreds:
## Problem

The Kotlin DSL accessor workspace identity (the `kotlin-dsl/accessors/<hash>/` directory) is `hashCodeFor(schema) + DCL flag + classpath fingerprint`. `hashCodeFor` hashed `tasks`/`extensions`/`containerElements`/`containerElementFactories`/`projectFeatures` **in schema-collection order**, while only `configurations` were sorted.

Task entries come from `DefaultTaskCollection.getCollectionSchema()` as `realized(sorted) ++ pending(registration order)` — **not** name-sorted. So the order depends on which tasks are realized vs still-lazy when the schema is read.

## Root cause (#37719), traced

Under Isolated Projects, CC project-metadata serialization realizes a task that non-IP leaves lazy:

```
ProjectMetadataController.write → writeVariantArtifactSets        (IP-only)
  → LocalVariantMetadata.getArtifacts()
  → DefaultIvyArtifactName.forPublishArtifact → DecoratingPublishArtifact.getName
  → LazyPublishArtifact.getName() → DefaultProperty → FlatMapProvider
  → TaskCreatingProvider.tryCreate → realizes `test`
```

Realizing `test` moves it from the registration-ordered pending tail into the sorted realized block, so the task list — and therefore the schema hash and the accessor directory — differs between IP and non-IP. (This is precisely the issue's "test task moves position".) Verified by instrumenting `GenerateProjectAccessors.identify()`: across modes the `classpathFp` and the *sorted* schema hash were identical; only the *list-order* schema hash differed, and the differing category was the task list.

## Fix

Sort each entry list before hashing (mirroring the existing `configurations` handling). The accessor identity is now independent of realization/collection order. With the hash stable, the `withNormalizedAccessorHash` workaround in `KotlinDslModelChecker` is removed; adds order-insensitivity unit tests for tasks, extensions, and container elements.

This is a one-time accessor-cache invalidation (hash values change), after which accessors regenerate normally.

## Why removing the normalization is safe — and why we still want the sort

The parity tests pass on master today only **because** `withNormalizedAccessorHash` masks the differing `<hash>` segment. The relevant finding is that they also pass **without** that normalization — but only since `b7844f3557d` (an unrelated #37942 fix), which added a per-`ClassLoaderScope` accessor cache that snapshots the schema on first computation, before the IP realization above happens. In other words, the underlying IP/non-IP divergence is no longer observable, so the normalization is now dead weight and can be dropped.

That masking is **incidental and timing-dependent**, though: `getCollectionSchema()` is still realized-then-pending, `hashCodeFor` still hashed in list order, and `test` is still force-realized under IP. If the first accessor computation ever moves after the metadata write, the divergence returns. So we sort the hash inputs to fix the identity directly (robust regardless of realization timing), and removing the normalizer restores real regression coverage — a future divergence now fails the parity tests instead of being silently normalized away.

## Verification

- `ProjectSchemaHashCodeTest` — 10/10 (incl. 3 new order-insensitivity tests)
- `kotlin-dsl-tooling-builders:test` — 11/11; `AccessorsClassPathModelCrossVersionSpec` — 9/9
- IP KotlinDsl parity suites with the normalizer removed — 24 tests, 0 failures (1 pre-existing skip): `…KotlinDslIntegrationTest`, `…CompositeBuildIntegrationTest`, `…BrokenScriptsIntegrationTest`
- Confirmed by bisect that **without** the normalization the composite suite fails at `09dfe47f47d` and passes from `b7844f3557d` onward, isolating the incidental masking commit.

Closes #37719.

### DIFF
--- a/platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/KotlinDslModelChecker.groovy
+++ b/platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/KotlinDslModelChecker.groovy
@@ -21,7 +21,6 @@ import java.nio.file.Paths
 import org.gradle.tooling.model.kotlin.dsl.EditorReport
 import org.gradle.tooling.model.kotlin.dsl.KotlinDslScriptModel
 import org.gradle.tooling.model.kotlin.dsl.KotlinDslScriptsModel
-import spock.lang.Issue
 
 import static org.gradle.integtests.tooling.fixture.ToolingApiModelChecker.checkModel
 
@@ -49,8 +48,8 @@ class KotlinDslModelChecker {
 
     static void checkKotlinDslScriptModel(KotlinDslScriptModel actual, KotlinDslScriptModel expected) {
         checkModel(actual, expected, [
-            { withNormalizedAccessorHash(it.classPath) },
-            { withNormalizedAccessorHash(it.sourcePath) },
+            { it.classPath },
+            { it.sourcePath },
             { it.implicitImports },
             [{ it.editorReports }, { a, e -> checkEditorReport(a, e) }],
             // Stack-trace strings legitimately differ between IP and non-IP runs
@@ -90,19 +89,5 @@ class KotlinDslModelChecker {
                 { it.column },
             ]]
         ])
-    }
-
-    // replaces `kotlin-dsl/accessors/*hash*/` with `kotlin-dsl/accessors/<hash>`
-    @Issue("https://github.com/gradle/gradle/issues/37719")
-    private static List<File> withNormalizedAccessorHash(List<File> paths) {
-        final String accessorPathPrefix = "kotlin-dsl" + File.separator + "accessors" + File.separator
-        paths.collect { File f ->
-            int idx = f.path.indexOf(accessorPathPrefix)
-            if (idx < 0) {
-                return f
-            }
-            int hashEnd = f.path.indexOf(File.separator, idx + accessorPathPrefix.length())
-            return new File(f.path.substring(0, idx + accessorPathPrefix.length()) + "<hash>" + f.path.substring(hashEnd))
-        }
     }
 }
--- a/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorsClassPath.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorsClassPath.kt
@@ -699,6 +699,17 @@ fun classLoaderScopeOf(scriptTarget: Any) = when (scriptTarget) {
 }
 
 
+/**
+ * Computes the hash that identifies [schema] for accessor caching.
+ *
+ * The hash is independent of the order in which entries appear within each schema category: every
+ * category is sorted before hashing. This is required because the schema is collected from live,
+ * mutable containers whose iteration order is not stable across configuration modes. For example,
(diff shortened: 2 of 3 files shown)


6hundreds:
@bot-gradle test apt

6hundreds:
@bot-gradle test this without pts on linux

platforms/core-configuration/kotlin-dsl-tooling-builders/src/testFixtures/groovy/org/gradle/kotlin/dsl/tooling/fixtures/KotlinDslModelChecker.groovy
```diff
@@ -49,8 +48,8 @@ class KotlinDslModelChecker {
 
     static void checkKotlinDslScriptModel(KotlinDslScriptModel actual, KotlinDslScriptModel expected) {
         checkModel(actual, expected, [
-            { withNormalizedAccessorHash(it.classPath) },
-            { withNormalizedAccessorHash(it.sourcePath) },
+            { it.classPath },
+            { it.sourcePath },
```
6hundreds, line 52: ℹ️ The main outcome of the PR

platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorsClassPath.kt
```diff
@@ -719,30 +719,40 @@ fun Hasher.putConfigurationEntries(configurations: List<ConfigurationEntry<Strin
 
 private
 fun Hasher.putAll(entries: List<ProjectSchemaEntry<SchemaType>>) {
+    // Sort before hashing so the identity is independent of schema collection order.
+    // Task entries come from DefaultTaskCollection.getCollectionSchema() as
+    // realized-then-pending, so a task realized under one mode but not another (e.g. `test`
+    // realized by IP's project-metadata serialization) reorders the list. See gradle/gradle#37719.
     putInt(entries.size)
```
alllex: 💭 I think the comment is valuable but maybe not in the best place? The method is generic, but tasks about tasks only. Also, other places do the sorting too. It feels like the comment should be moved up, maybe.

Maybe we can additionally clarify the contract of the `hashCodeFor` method by specifying that it hashes `TypedSprojectSchema` order-independently?

alllex (approved):
Nice 🎉

h0tk3y (approved):
👍 LGTM

PR #38319, opened 2026-06-25
Add write-side mark/commit/rollback to CC serialization

mlopatkin:
Introduces a rollback capability for Configuration Cache writes: mark a
position, then either commit everything written since (relayed verbatim)
or roll it back as if it never happened. The reader and on-disk format
are unchanged.

Mechanism (spool-and-relay, write-side only):
- SpillingOutputStream diverts post-mark bytes into a small in-memory
  buffer that overflows to an encrypted sibling temp file, so plaintext
  never reaches disk and small regions never touch disk at all.
- EncoderRollback flushes the encoder at each boundary and drives the
  spill; rollback flushes the tail into the spool and drops it, so the
  Kryo Output internals are untouched.
- WriteContext.beginRollbackScope() journals the inline WriteIdentities
  (current isolate + shared) and swaps in a BufferingProblemsListener
  that defers problems, replaying them on commit or returning them for
  inspection on rollback. circularReferences/PropertyTrace/reentrant
  self-heal on unwind.

The BufferingProblemsListener forwards onError immediately rather than
deferring it: onError carries control-flow meaning (the CC problems
listener re-throws IOException/ConfigurationCacheThrowable to abort
serialization), so deferring it would let a failing write appear to
succeed and surface only during replay.

Rollback is offered only for the plain KryoBackedEncoder path (Work),
not the sequential string-dedup encoder whose inline table is not yet
revertible. Nesting is rejected.

Covered by unit tests for WriteIdentities, SpillingOutputStream and
BufferingProblemsListener, plus a write/read round-trip test.

### DIFF
--- a/.teamcity/subprojects.json
+++ b/.teamcity/subprojects.json
@@ -709,7 +709,7 @@
   {
     "name": "graph-serialization",
     "path": "platforms/core-configuration/graph-serialization",
-    "unitTests": false,
+    "unitTests": true,
     "functionalTests": false,
     "crossVersionTests": false
   },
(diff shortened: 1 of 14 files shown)


mlopatkin:
@bot-gradle test PT

PR #38321, opened 2026-06-25
Consolidate all Best Practices to kebab-cased directory

tresat:
We had `/bestPractices` and `/best-practices`.


lkasso (approved):

hfhbd:
Is it also a best practice to have kebab-cased project names?

tresat:
> Is it also a best practice to have kebab-cased project names?

We don't have a stance on that, but it does seem fairly common.  This change is just internal docs cleanup.

PR #38327, opened 2026-06-26
Make native Swift incremental-compile tests Isolated Projects-compatible

blindpirate:
## What

Makes native Swift incremental-compile testing compatible with **Isolated Projects**. Two layers were involved:

### 1. Shared test fixture (`AbstractInstalledToolChainIntegrationSpec`)

The fixture injected an init script applying the toolchain plugin to every project via `allprojects { apply ... }`, which IP rejects from the root project:

```
* Where: Initialization script 'init.gradle' line: 3
Project ':' cannot access 'Project.apply' functionality on subprojects via 'allprojects'
```

This deterministically failed *every* multi-project native test run under `-Dorg.gradle.isolated-projects=true`, at init-script evaluation. Switched to the IP-compatible `gradle.lifecycle.beforeProject { }` hook (toolchain values are interpolated into the script text at generation time, so the action captures no external state).

### 2. `SwiftIncrementalCompileIntegrationTest`

Once the init-script violation was cleared, the multi-project test *changes to an unused dependency rebuilds everything* surfaced two further, test-local cross-project accesses:

- `setup()` configured tasks across projects: `allprojects { tasks.withType(SwiftCompile) { ... } }` → now configures only the current project (`tasks.withType(SwiftCompile).configureEach { }`).
- The test cross-applied the library plugin: `project(":unused") { apply plugin: 'swift-library' }` → now applied in the subproject's own `unused/build.gradle`.
- The `SWIFTC_4` test's init-script rewrite used the same `allprojects { apply }` pattern → updated to `gradle.lifecycle.beforeProject` for consistency.

## Verification

Reproduced the original CI error and confirmed each fix with faithful local repros (real multi-project Swift builds, swiftc, `-Dorg.gradle.isolated-projects=true`):

- **Init-script `allprojects { apply }`** → reproduced the exact IP error; gone after the fixture fix. ✅
- **Test-local violations** → CI run [#114408069](https://builds.gradle.org/build/114408069) on this branch (fixture fix only) confirmed the failure moved to `Project.tasks ... via allprojects` exactly as expected. ✅
- **After both fixes** → both branches of the test (`-PincludeDep` and removal) configure and compile under IP with **no configuration-cache problems** (cache stored and reused). ✅
- **Non-IP regression** → toolchain still applied and `compileDebugSwift` compiles normally. ✅

Originating failure: https://builds.gradle.org/build/114394092 — `SwiftIncrementalCompileIntegrationTest > changes to an unused dependency rebuilds everything` (swiftc 6.0.1), *Flaky Test Quarantine - IsolatedProjects*.

CI verification of the complete fix is running on branch `bz-swiftc`.

### DIFF
--- a/platforms/native/language-native/src/integTest/groovy/org/gradle/language/swift/SwiftIncrementalCompileIntegrationTest.groovy
+++ b/platforms/native/language-native/src/integTest/groovy/org/gradle/language/swift/SwiftIncrementalCompileIntegrationTest.groovy
@@ -33,10 +33,8 @@ class SwiftIncrementalCompileIntegrationTest extends AbstractInstalledToolChainI
     def setup() {
         // Useful for diagnosing swiftc incremental compile failures
         buildFile << """
-            allprojects {
-                tasks.withType(SwiftCompile) {
-                    compilerArgs.add('-driver-show-incremental')
-                }
+            tasks.withType(SwiftCompile).configureEach {
+                compilerArgs.add('-driver-show-incremental')
             }
         """
     }
@@ -211,10 +209,9 @@ class SwiftIncrementalCompileIntegrationTest extends AbstractInstalledToolChainI
         """
         buildFile << """
             apply plugin: 'swift-application'
-
-            project(":unused") {
-                apply plugin: 'swift-library'
-            }
+        """
+        file("unused/build.gradle") << """
+            apply plugin: 'swift-library'
         """
         file("unused/src/main/swift/Library.swift") << """
             public class Library {
@@ -289,10 +286,10 @@ class SwiftIncrementalCompileIntegrationTest extends AbstractInstalledToolChainI
         Assume.assumeNotNull(swiftc3, swiftc4)
 
         initScript.text = """
-            allprojects { p ->
-                apply plugin: ${swiftc3.pluginClass}
+            gradle.lifecycle.beforeProject { p ->
+                p.apply plugin: ${swiftc3.pluginClass}
 
-                toolChains {
+                p.toolChains {
                     ${swiftc3.buildScriptConfig}
                 }
             }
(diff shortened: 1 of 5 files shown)


platforms/native/platform-native/src/testFixtures/groovy/org/gradle/nativeplatform/fixtures/AbstractInstalledToolChainIntegrationSpec.groovy
```diff
@@ -39,13 +39,16 @@ abstract class AbstractInstalledToolChainIntegrationSpec extends AbstractIntegra
     AvailableToolChains.InstalledToolChain getToolchainUnderTest() { toolChain }
 
     def setup() {
+        // Use the Isolated Projects-compatible lifecycle hook instead of `allprojects { apply ... }`,
+        // which is rejected under Isolated Projects as a cross-project access from the root project.
+        // The toolchain values are interpolated into the script text here, so the action captures no external state.
```
alllex: 🤔  I doubt the value of these comments _in the code_. They might be useful in a commit message, and even then not very much.

Using `allprojects` or `.beforeProject` callbacks is only required here to reduce boilerplate of setting up the test. Therefore, I don't think we should explain why we use one boilerplate instead of the other. Similarly, we could have distributed these statements across individual build scripts.

platforms/native/language-native/src/integTest/groovy/org/gradle/language/swift/SwiftIncrementalCompileIntegrationTest.groovy
alllex, line 1: ❓ If we made some fixtures or setups more compatible, why weren't any of `@ToBeFixedForIsolatedProjects` annotations adjusted?

alllex: Could it be that these tests were not executed with regular auto-triggered CI and we should run something more extensive?

alllex:

blindpirate:

alllex (approved):

blindpirate:
@bot-gradle squash and merge

PR #38338, opened 2026-06-26
Adjust TaskInputs.files serialization in CC to handle absent better

### LINKED ISSUE #38330: Optional runtime file input from an absent provider fails fingerprinting with the configuration cache
### Current Behavior

When a task has a **runtime-defined optional file input** (`task.inputs.files(provider).optional()`) and the source `provider` is **absent** (has no value), the build **fails when the configuration cache is enabled**:

```
* What went wrong:
Execution failed for task ':myTask' (registered in build file 'build.gradle').
> Cannot query the value of this property because it has no value available.
```

Without the configuration cache, the same build **succeeds** — the absent provider is resolved leniently and the input is simply treated as empty.

(Related observation: a *required* runtime file input fed an absent provider also behaves differently between the two modes. Without CC it fails with the structured `value-not-set` validation problem — *"Property 'inputFiles' doesn't have a configured value … isn't marked as optional"*. With CC it instead fails with the same generic *"Cannot query the value of this property because it has no value available."* and emits no validation problem.)

### Expected Behavior

An optional file input whose source provider is absent should be ignored under the configuration cache exactly as it is without it, and the task should run successfully. Behavior should be consistent between CC and non-CC modes.

### Context (optional)

This surfaces in plugins that wire an optional input from another task/provider that may not produce a value (e.g. `inputs.files(other.flatMap { it.optionalOutput }).optional()`), so enabling the configuration cache breaks builds that work fine without it.

`settings.gradle`:
```groovy
rootProject.name = "absent-optional-input"
```

`build.gradle`:
```groovy
def absentSource = objects.fileProperty()  // no value -> absent

tasks.register("myTask") {
    inputs.files(absentSource).optional().withPropertyName("inputFiles")
    def out = layout.buildDirectory.file("out.txt")
    outputs.file(out)
    doLast {
        out.get().asFile.text = "done"
    }
}
```

```
gradle myTask --no-configuration-cache                       # SUCCEEDS
gradle myTask --configuration-cache    # FAILS: Cannot query the value of this property because it has no value available.
```

### Gradle version

9.7.0-milestone-1

### Build scan URL (optional)

_No response_

_No response_

mlopatkin:
This is all caused by a failure when an optional files() input has an
absent provider as its single source. Without CC, it is allowed, but the
new FileCollection wrapping broke this.

First problem is that CC didn't preserve file collection leniency, so
fingerprinting of the optional property was failing while resolving the
wrapping collection. This commit changes the codec to preserve the
strategy properly.

This uncovered the second issue. Property validation is not
exactly part of the fingerprinting and examines raw values rather than
inputs as a collection. It has special handling for absent providers.
In particular, an absent provider doesn't pass the check if the property
is not optional, but it works when it is wrapped in a file collection or
an Iterable. Thus, our unconditional wrapping tripped it over as it no
longer saw the provider. The fix is to exclude provider from wrapping,
as CC already knows how to handle it.

Confusingly, for users this means that `inputs.files(absent)` works,
while `inputs.files(absent, absent)` doesn't, but this is consistent
with vintage and out of scope for this PR.

Fixes #38330.

### DIFF
--- a/platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
+++ b/platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
@@ -28,7 +28,9 @@ import org.gradle.api.internal.provider.Providers
 import org.gradle.api.internal.tasks.TaskDestroyablesInternal
 import org.gradle.api.internal.tasks.TaskInputFilePropertyBuilderInternal
 import org.gradle.api.internal.tasks.TaskLocalStateInternal
+import org.gradle.api.provider.Provider
 import org.gradle.api.specs.Spec
+import org.gradle.api.tasks.TaskProvider
 import org.gradle.execution.plan.LocalTaskNode
 import org.gradle.execution.plan.TaskNodeFactory
 import org.gradle.internal.cc.base.serialize.IsolateOwners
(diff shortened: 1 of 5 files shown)


mlopatkin:
@bot-gradle test ACC

mlopatkin:
@bot-gradle test ACC

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -423,13 +425,18 @@ suspend fun WriteContext.writeRegisteredPropertiesOf(task: Task) {
  * mutable internals. This is semantically equivalent to what consumers of `TaskInputs` do at execution time
  * via `FileParameterUtils.resolveInputFileValue`.
  *
+ * One exception are [Provider] values. These are handled specially by the validation logic, so we have to preserve
+ * the shape. And yes, this means that `files(absentProvider)` fails and `files(listOf(absentProvider))` works.
```
ov7a, line 429: ❓ do we have an open issue for that?

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -445,13 +452,26 @@ fun WriteContext.adaptInputFileValueForSerialization(value: Any?, filePropertyTy
...
     }
     return isolate.owner.serviceOf<FileCollectionFactory>().resolvingLeniently(value)
 }
 
 
+/**
+ * Checks if the input value needs to be wrapped in a file collection to properly serialize.
+ *
+ * [FileCollection]s aren't wrapped as it is pointless.
+ * [Provider]s except [TaskProvider] aren't wrapped because input validation has special handling for them.
```
ov7a, line 466: :harold:

ov7a (approved):

PR #30028, opened 2024-07-26
Collect stacktraces for large number of IP violations

alllex:
Fixes https://github.com/gradle/gradle/issues/29441

by increasing the limit for the number of collected stacktraces to 5000 when Isolated Projects feature is enabled

PR #38344, opened 2026-06-26
Drop the 5k full-stack cap under Isolated Projects

alllex:
> **Spike / draft.** Implements the follow-up agreed in #32362 and the related Slack thread: now that the cheap bounded location capture has landed (#38127, #38251), drop the Isolated-Projects-specific 5000 full-stack-trace cap.

### Why
PR #30028 raised the full stack-trace capture cap from 50 to **5000** under Isolated Projects, as a corner-cutting way to give early adopters source locations on large builds. The number was arbitrary (loosely inspired by the CC 4096 problem-count cap).

Since then, #38127 and #38251 added a cheap **bounded** location capture: past the full-stack cap, a bounded `StackWalker` walk infers the source location (keeping the originating build logic down to the calling script, dropping the runtime frames) at negligible cost. That makes the 5000 full-stack cap redundant — it only forces expensive full captures where a bounded capture yields the same location.

### What changes
- Remove `ISOLATED_PROJECTS_MAX_STACKTRACE_COUNT = 5000` and the `isIsolatedProjects()` branch in `DefaultProblemDiagnosticsFactory`. IP now uses the standard **50** full-stack cap + the bounded location capture past it, like every other mode.
- The supplied-exception path (`forCurrentCaller(Supplier)`) past the full cap now also falls back to a location-only bounded capture, but **never** surfaces that bounded throwable as the problem's exception — a supplied throwable is preserved or dropped, never replaced. Factored out as `StackTraceCapturer.captureBoundedFallback()`.

**Behaviour under IP:**
- Problems 1–50: full trace (unchanged).
- Problems 51–2050: previously full trace (up to 5000), now bounded trace + location.
- Problems 2051+: no inferred location (same as non-IP); `--warning-mode=all`/`fail` still lifts the bounded budget so every problem gets a location.

No public API change. The within-cap behaviour is unchanged.

### Tests
- `IsolatedProjectsProblemReportingIntegrationTest` — rewritten for the bounded-past-50 contract (530 problems still all resolve to their line).
- `ProblemLocationPastCapIntegrationTest` — `@ToBeFixedForIsolatedProjects` removed; now exercises the bounded path under IP too.
- `DefaultProblemDiagnosticsFactoryTest` / `StackTraceCapturerTest` — cover the supplied-exception bounded fallback and budget exhaustion.

### Notes
- The CC problem-count caps (`maxCollectedProblems = 4096`, `configurationCacheMaxProblems = 512`) are a different axis (how many problems are reported, not per-problem trace cost) and are untouched.
- Out of scope (separate follow-ups per the thread): raising/removing the 2000 bounded budget; a knob to force full traces; the `largeAndroidBuild` IDE-sync regression.

Follow-up to #32362.

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsProblemReportingIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsProblemReportingIntegrationTest.groovy
@@ -22,7 +22,7 @@ import static org.gradle.integtests.fixtures.configurationcache.ConfigurationCac
 
 class IsolatedProjectsProblemReportingIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
 
-    def "stops reporting problems at certain limits collecting all stacktraces"() {
+    def "reports problems with locations past the stack-capture cap using bounded captures"() {
         settingsFile """
             include(":a")
         """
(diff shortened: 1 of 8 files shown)


alllex:
Superseeded by https://github.com/gradle/gradle/pull/39107

PR #38346, opened 2026-06-26
Fix issue with configureEach realizing pending elements

### LINKED ISSUE #36951: DomainObjectCollection realize lazily added elements too early
### Current Behavior

Any `DomainObjectCollection` realize lazily added elements (via `addAllLater`) as soon as `configureEach` is called (after at least one _eager_ element exist) thanks to the [call to `estimatedSize()`](https://github.com/gradle/gradle/blob/master/subprojects%2Fcore%2Fsrc%2Fmain%2Fjava%2Forg%2Fgradle%2Fapi%2Finternal%2FDefaultDomainObjectCollection.java#L205) hidden behind `hasNext()` of the _no flush iterator_.

### Expected Behavior

Lazily added elements should not be treated differently when mixed with eager elements.

### Context (optional)

There may be several path to the issue. For example, the size estimation on collection provider may be wrong (aka resolve element provider too early).

```
def c = objects.domainObjectSet(Object)
c.addAllLater(objects.listProperty(Object).value(provider { throw new RuntimeException() }))
c.configureEach { println('pass') }
c.add(new Object())
c.configureEach { println('boom') }
```

### Gradle version

any

### Build scan URL (optional)

_No response_

_No response_

ghale:
Fixes #36951

### Context

`DomainObjectCollection.configureEach` is [documented as lazy](https://docs.gradle.org/current/javadoc/org/gradle/api/DomainObjectCollection.html#configureEach(org.gradle.api.Action)) and registering a `configureEach` action should not realize elements that were added lazily via `addAllLater`. However, once a collection also contained at least one *eager* element, calling `configureEach` realized every pending `addAllLater` collection provider.

Root cause: DefaultDomainObjectCollection.configureEach iterates with the no-flush iterator (correctly yielding only realized elements), but presized its temporary copy list with estimatedSize(). estimatedSize() sums size() over all inserted elements, and size() on a pending collection provider resolves the provider value, realizing it. The asymmetry in the reproducer is that the presize is only reached once the no-flush iterator is non-empty, i.e. once at least one eager element exists.

Fix: the copy only ever holds the realized elements yielded by iteratorNoFlush(), so it is sized with a default-capacity ArrayList instead of estimatedSize(). This keeps configureEach lazy without changing the estimatedSize() contract or the eager all() / whenObjectAdded path (which iterates the flushing iterator and is legitimately eager).

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DefaultDomainObjectCollection.java
+++ b/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DefaultDomainObjectCollection.java
@@ -203,7 +203,11 @@ public class DefaultDomainObjectCollection<T> extends AbstractCollection<T> impl
         while (iterator.hasNext()) {
             // only create an intermediate collection if there's something to copy
             if (copied == null) {
-                copied = new ArrayList<>(estimatedSize());
+                // Don't size the copy with estimatedSize(): it realizes pending elements (size() on a
+                // collection provider resolves its value), which would make configureEach eager. This copy
+                // only ever holds the realized elements yielded by iteratorNoFlush(), so a default-capacity
+                // list is correct and keeps configureEach lazy, as documented.
+                copied = new ArrayList<>();
             }
             copied.add(iterator.next());
         }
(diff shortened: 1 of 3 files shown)


ghale:
@bot-gradle test this

big-guy (approved):

PR #38349, opened 2026-06-27
Surface clear cause for exec stream config error

### LINKED ISSUE #33858: Confusing failure when setting up standard streams for Providers.exec and javaexec
### Current Behavior

Build fails with a confusing message:
```

* Where:
Build file '/private/tmp/tests/1/build.gradle' line: 2

* What went wrong:
A problem occurred evaluating root project '1'.
> Could not create provider for value source ProcessOutputValueSource.
```

### Expected Behavior

The clarifying message `Standard streams cannot be configured for exec output provider` should be visible without specifying `--stacktrace`.

### Context (optional)

_No response_

```
providers.exec {
  standardInput = System.in
  commandLine("echo")
}
```

### Gradle version

n/a

### Build scan URL (optional)

_No response_

_No response_

mlopatkin:
Fixes #33858

### Context

When a build configures standard streams on a `providers.exec {}` / `providers.javaexec {}` spec, the build fails with a confusing message:

```
* What went wrong:
A problem occurred evaluating root project '...'.
> Could not create provider for value source ProcessOutputValueSource.
```

The actionable cause — `Standard streams cannot be configured for exec output provider` — was only visible with `--stacktrace`.

**Reproducer:**
```groovy
providers.exec {
    standardInput = System.in
    commandLine("echo")
}
```

### Root cause

`ProviderCompatibleBaseExecSpec` threw `UnsupportedOperationException` from its standard-stream getters/setters. That spec is configured inside the action run by `DefaultValueSourceProviderFactory.createProviderOf`, whose `catch (Exception e)` wraps any non-`GradleException` into a generic `GradleException("Could not create provider for value source …", e)`. The default console renderer shows the wrapper's message but collapses the nested `UnsupportedOperationException` cause, hiding the meaningful text.

### Change

Throw `InvalidUserCodeException` (a `GradleException` subtype) instead of `UnsupportedOperationException` from the six standard-stream getters/setters, keeping the message verbatim. Because `createProviderOf` rethrows `GradleException` causes **unwrapped** (`catch (GradleException e) { throw e; }`), the clarifying message now surfaces as the reported failure cause without `--stacktrace`. The exception type is also semantically correct: configuring standard streams here is invalid build-script usage.

No change to the generic `createProviderOf` wrapping path, so there is no blast radius for other value source types.

### Out of scope

Actually *supporting* input-stream / standard-stream configuration on `providers.exec`/`javaexec` (deferred until `ExecSpec` migrates to the Provider API, per the issue discussion). This PR only makes the existing error message visible.

### Tests

- New regression integration test `ProcessOutputProviderIntegrationTest."providers.exec reports a clear cause when configuring standard streams"` (annotated `@Issue`), asserting the clarifying cause is visible in the rendered failure (without `--stacktrace`). Red before the fix, green after, in both forking and configuration-cache variants.
- Updated the three `set*` forbidden-stream unit tests in `ProviderCompatibleBaseExecSpecTestBase` to assert the new exception type and exact message; added three new tests covering the previously-untested `get*` methods.

### DIFF
--- a/platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/ProcessOutputProviderIntegrationTest.groovy
+++ b/platforms/core-configuration/model-core/src/integTest/groovy/org/gradle/api/provider/ProcessOutputProviderIntegrationTest.groovy
@@ -20,6 +20,7 @@ import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.process.ShellScript
 import org.gradle.process.TestJavaMain
 import org.gradle.util.internal.TextUtil
+import spock.lang.Issue
 
 class ProcessOutputProviderIntegrationTest extends AbstractIntegrationSpec {
     def "providers.exec can be used during configuration time"() {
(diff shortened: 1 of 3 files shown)


mlopatkin:
@bot-gradle test this

PR #33585, opened 2025-05-22
Suppress and propagate deprecations in generated Kotlin DSL accessors

h0tk3y:
Fixes:
* https://github.com/gradle/gradle/issues/33386
* https://github.com/gradle/gradle/issues/33387

PR #38364, opened 2026-06-29
Fix margin marker leaking into generated Kotlin DSL ERROR-deprecation accessors

ljacomet:
Introduced in
* #33585


h0tk3y (approved):
LGTM, thanks for fixing this!

PR #38372, opened 2026-06-30
Label PRs where `bt-cortex` is a reviewer with `@cortex` label

mlopatkin:
This allows Project automation to automatically pick them up. This is slightly better than using `add-to-project` action to add directly, as no extra-privilege token is needed.

### DIFF
--- a/.github/workflows/label-cortex-review-prs.yml
+++ b/.github/workflows/label-cortex-review-prs.yml
@@ -1,0 +1,29 @@
+name: Label PRs reviewed by bt-cortex
+
+# The `@cortex`-labeled PRs are expected to be picked up by Project automation of the Cortex Board.
+
+on:
+  pull_request_target:
+    types: [review_requested, review_request_removed]
+
+permissions: {}
+
+jobs:
+  label:
+    if: github.event.requested_team.slug == 'bt-cortex'
+    permissions:
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add @cortex label
+        if: github.event.action == 'review_requested'
+        env:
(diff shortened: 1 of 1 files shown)


ov7a:
Ugh, Github.

The same query that works fine [in filters](https://github.com/gradle/gradle/pulls?q=is%3Aopen+is%3Apr+team-review-requested%3Agradle%2Fbt-cortex) does not work for workflows.

image

.github/workflows/label-cortex-review-prs.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add @cortex label
+        if: github.event.action == 'review_requested'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR: ${{ github.event.pull_request.html_url }}
+        run: gh pr edit "$PR" --add-label "@cortex"
+      - name: Remove @cortex label
+        if: github.event.action == 'review_request_removed'
```
ov7a, line 25: 🤔 I wonder if the workflows will always be able to react to the addition event properly and there's no subtle concurrency issue.

 But as long as it works, LGTM.

ov7a (approved):

.github/workflows/label-cortex-review-prs.yml
mlopatkin: > 🤔 I wonder if the workflows will always be able to react to the addition event properly and there's no subtle concurrency issue.
> 
> But as long as it works, LGTM.

I think add-remove-add are going to be rare in practice.

PR #38377, opened 2026-06-30
Avoid ProjectRegistry

jvandort:
ProjectRegistry is an old type. We remove all remaining usages of it in favor of the modern BuildProjectRegistry.

See https://github.com/gradle/gradle/issues/34469

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
```diff
@@ -72,7 +71,8 @@ class CrossProjectConfigurationReportingGradle(
         else -> gradle
     }
 
-    private val crossProjectModelAccess: CrossProjectModelAccess = delegate.serviceOf()
+    private val crossProjectModelAccess: CrossProjectModelAccess
+        get() = delegate.serviceOf()
```
reinsch82: ❓ How about this?

```suggestion
    private val crossProjectModelAccess: CrossProjectModelAccess by lazy { delegate.serviceOf() }
```

preserves the caching and adds laziness?

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultCrossProjectModelAccess.java
```diff
@@ -65,32 +66,40 @@ public ProjectInternal findProject(ProjectIdentity referrer, Path path) {
...
-            .collect(Collectors.toCollection(TreeSet::new));
+        ProjectState targetProjectState = projectRegistry.getProject(target.getProjectPath());
+        Set<ProjectInternal> allProjects = subprojectsFor(referrer, targetProjectState);
+        allProjects.add(accessFromState(referrer, targetProjectState));
+        return allProjects;
+    }
+
+    private Set<ProjectInternal> subprojectsFor(ProjectIdentity referrer, ProjectState targetProjectState) {
+        TreeSet<ProjectInternal> subprojects = new TreeSet<>();
+        for (ProjectState subproject : ProjectOrderingUtil.orderedSubprojectsOf(targetProjectState)) {
```
reinsch82: ❌ I think we could get and unordered subproject list, since `TreeSet` does it's own ordering.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt
```diff
@@ -140,25 +141,28 @@ fun log(message: String) {
 
 
 private
-fun Project.findProjectWithBuildFile(file: File) =
-    allprojects.find { it.buildFile == file }
+fun ProjectInternal.findProjectWithBuildFile(file: File) =
+    ProjectOrderingUtil.orderedAllProjectsOf(owner)
```
reinsch82, line 145: ❌ also looks like unused ordering. Does it help to have the projects ordered?

reinsch82 (changes requested):
some suggestions

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt
jvandort: In practice only a single file should match in this case, so order should not matter. 

Really though, it was sorted in the existing code. I would rather maintain behavior here. This code will be going away when we get rid of the non-ip safe IDE model builders so I'm not so concerned about this atm.

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultCrossProjectModelAccess.java
jvandort: Switched to `LinkedHashSet`

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/CrossProjectConfigurationReportingGradle.kt
jvandort: I don't think caching this or not is performance sensitive at all, though I've added lazy.

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

reinsch82 (approved):
🎉

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinDslScriptsModelBuilder.kt
```diff
@@ -166,10 +167,13 @@ fun Project.collectKotlinDslScripts(): List<File> = buildList {
     addAll(discoverInitScripts())
     addNotNull(discoverSettingScript())
 
-    allprojects.forEach { p ->
-        addNotNull(p.discoverBuildScript())
-        addAll(p.discoverPrecompiledScriptPluginScripts())
-    }
+    ProjectOrderingUtil.orderedAllProjectsOf((project as ProjectInternal).owner)
+        .asSequence()
+        .map { it.mutableModelEvenAfterFailure }
```
tresat, line 172: 💭 There's a lot of repetition of:

```
.orderedXProjectsOfY()
.asSequence()
.map { it.mutableModelEvenAfterFailure }
```

Perhaps it would be better to have something like:

`ProjectOrderingUtil.forEachMutableSub(project: ProjectInternal, actionOnSub: (Project) -> Unit)`

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectOrderingUtil.java
```diff
@@ -17,7 +17,53 @@
...
+
 public class ProjectOrderingUtil {
+
+    /**
+     * Returns a new collection, containing all direct and transitive child projects of
+     * the given project, in order.
+     * <p>
+     * If you are writing new code and want to use this, consider if you really need to.
+     * Most operations should be scoped to a single project. Why do you need all transitive
+     * subprojects at the same time?
```
tresat, line 32: 💭 Should these methods be marked deprecated to begin with then?  Even if they doesn't emit a warning?

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectOrderingUtil.java
```diff
@@ -17,7 +17,53 @@
...
+     */
+    public static List<ProjectState> orderedSubprojectsOf(ProjectState projectState) {
```
tresat, line 34: ❗ So these methods source their `Project` list from:

```
    /**
     * Immutable list of children, sorted by their identity.
     */
    List<ProjectIdentity> getChildren();
```

which is already sorted.  Then in:

```
    public Collection<ProjectState> getUnorderedChildProjects() {
        return CollectionUtils.transform(descriptor.getChildren(), this::getState);
    }
```

we are intentionally losing the sorting guarantee?  Why not return a `List` there, preserve the ordering in that method contract, and so remove the need to re-sort here?

subprojects/core/src/main/java/org/gradle/internal/service/scopes/BuildScopeServices.java
```diff
@@ -398,9 +403,9 @@ protected TaskStatistics createTaskStatistics() {
 
     @SuppressWarnings("deprecation")
     @UsedByScanPlugin("ImportJUnitXmlReports")
-    @Provides({ProjectRegistry.class, org.gradle.api.internal.project.DefaultProjectRegistry.class})
-    protected org.gradle.api.internal.project.DefaultProjectRegistry createProjectRegistry() {
-        return new org.gradle.api.internal.project.DefaultProjectRegistry();
+    @Provides({org.gradle.api.internal.project.ProjectRegistry.class, org.gradle.api.internal.project.DefaultProjectRegistry.class})
```
tresat, line 406: 💅 Use imports here

subprojects/core/src/main/java/org/gradle/execution/TaskNameResolver.java
```diff
@@ -36,11 +36,16 @@ public class TaskNameResolver {
...
+            return true;
+        }
+
+        for (ProjectState child : project.getUnorderedChildProjects()) {
+            if (tryFindUnqualifiedTaskCheaply(name, child)) {
                 return true;
             }
         }
+
         return false;
```
tresat: ❓ What is cheap about this?  Why is it unreliable?  The name "cheaply" or the javadoc doesn't answer this.

`findFirstTaskWithName`, `tryFindTaskInProjectAndSubs`?

subprojects/core/src/main/java/org/gradle/execution/TaskNameResolver.java
```diff
@@ -36,11 +36,16 @@ public class TaskNameResolver {
      */
     public boolean tryFindUnqualifiedTaskCheaply(String name, ProjectState project) {
         // don't evaluate children, see if we know it's without validating it
```
tresat:
```suggestion
        // don't evaluate children, see if we can locate the task without evaluating this project
```

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultCrossProjectModelAccess.java
```diff
@@ -65,32 +66,38 @@ public ProjectInternal findProject(ProjectIdentity referrer, Path path) {
...
         );
     }
 
     @Override
     public Set<? extends ProjectInternal> getSubprojects(ProjectIdentity referrer, ProjectIdentity target) {
-        return projectRegistry.getSubProjects(target.getProjectPath().asString()).stream()
-            .map(project -> LifecycleAwareProject.wrap(project, referrer, instantiator, gradleLifecycleActionExecutor))
-            .collect(Collectors.toCollection(TreeSet::new));
+        ProjectState targetProjectState = projectRegistry.getProject(target.getProjectPath());
+        Set<ProjectInternal> subprojects = new LinkedHashSet<>();
```
tresat, line 86: 💅 Why `LinkedHashSet` rather than `HashSet` if we're only returning a `Set` here?  No need to keep the ordering inside this method.

subprojects/core/src/test/groovy/org/gradle/api/internal/project/DefaultProjectRegistryTest.groovy
```diff
@@ -0,0 +1,57 @@
...
+    def "returns projects if present"() {
+        expect:
+        projectRegistry.getProject(":").is(rootMock.mutableModel)
+        projectRegistry.getProject(":child").is(childMock.mutableModel)
+        projectRegistry.getProject(":child:child").is(childChildMock.mutableModel)
+    }
+
+    def "returns null if no project"() {
+        expect:
+        projectRegistry.getProject(":foo") == null
```
tresat, line 45: Could be good to show some "close misses" here ("child" instead of ":child"), and also test `null` argument.

tresat (commented):
I think the sorting might be extraneous.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinDslScriptsModelBuilder.kt
jvandort: Generally we want internal code to operate on `ProjectState` objects rather than `Project` instances. If you want to operate on `Project` instances, you should be going through `allprojects`. 

Resilient sync is a special case. It needs these mutable models after failure, so can't go through allprojects even though it wants mutable project instances. 

In practice this code should not live long. I wanted to do the minimum to fix this. We have a separate implementation of these model builders for IP which only operate on a single project and we will be migrating to them soon and removing this old IP unsafe implementation.

subprojects/core/src/main/java/org/gradle/internal/service/scopes/BuildScopeServices.java
jvandort: `ProjectRegistry` is deprecated. the compiler will complain if we try to import it.

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultCrossProjectModelAccess.java
jvandort: The Project.allProjects and Project.subprojects, which this method backs, historically have ordering even though they return a `Set`. We could move where we enforce that ordering closer to the `Project` API, but that was not the intention of this PR. The old code did ordering here before, so we continue to that here.

subprojects/core/src/test/groovy/org/gradle/api/internal/project/DefaultProjectRegistryTest.groovy
jvandort: `DefaultProjectRegistry` is a shell of a service which only exists because a prior version of DV uses it. I've retained any relevant tests we had before, but I don't think its worth investing into testing this any further. In practice it is a light wrapper around `BuildProjectRegistry` which has its behavior verified in `DefaultProjectStateRegistryTest`

subprojects/core/src/main/java/org/gradle/execution/TaskNameResolver.java
jvandort: I imagine cheaply comes from that it only realizes as few projects as possible. Renamed. Renamed to `findFirstTaskWithName`

subprojects/core/src/main/java/org/gradle/execution/TaskNameResolver.java
jvandort: Didn't use this exact wording but reworded the comment.

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectOrderingUtil.java
jvandort: Prior to https://github.com/gradle/gradle/pull/35933, `getChildren` was not ordered and returned declaration order. That PR seems to have made `getUnorderedChildProjects` irrelevant. 

The difference between `ProjectState.getChildren()` which is ordered and `ProjectOrderingUtil.orderedSubprojectsOf(ProjectState)` is that `orderedSubprojectsOf` recursively visits the subproject tree. It has more than a single projects' children. 

Now it might be possible that if we just concat all projects ordered children as we visit them in one big list, we might end up with the desired ordering of `ProjectOrderingUtil::compare`. Though, I'd rather not rely on whatever ordering `ImmutableProjectDescriptor.getChildren()` has and ensure it here. 

The point is that `getUnorderedChildProjects` returns a collection, and while collections iterate in _some order_, consumers should not necessarily depend on that order.

I don't really think we should be relying on `ImmutableProjectDescriptor.getChildren` to return any particular order. If your use case (strangely) cares about the order of projects, you should ensure that yourself.

subprojects/core/src/main/java/org/gradle/api/internal/project/ProjectOrderingUtil.java
jvandort: At the moment there are some more legacy use cases for this code. I think if we designed things today we probably wouldn't need this method, and some rearchitecting could potentially get rid of existing usages of this method, I think deprecating without any planned removal would cause more noise than benefit.

PR #38385, opened 2026-07-01
Keep task-specific problems distinct across different tasks

### LINKED ISSUE #38147: Problems API may loose a reported similar problems for different tasks
### Current Behavior

When two tasks in parallel send a set of similar problems, the Gradle Problems API loses some of the events for these tasks - one task will show a subset of reported problems, and another one will show another non-intersecting subset.

For example:
```
> Task :mylibrary:compileDebugKotlin FAILED
e: [REDUNDANT_CLI_ARG] The argument '-Xcontext-parameters' is redundant for the current language version 2.4.
e: warnings found and -Werror specified

> Task :mylibrary:compileReleaseKotlin FAILED
e: [REDUNDANT_CLI_ARG] The argument '-Xcontext-parameters' is redundant for the current language version 2.4.
e: warnings found and -Werror specified
```
will end up in the HTML report as:
```
[REDUNDANT_CLI_ARG] The argument '-Xcontext-parameters' is redundant for the current language version 2.4.
Kotlin compiler error
        - Locations
- `:mylibrary:compileReleaseKotlin`
    - [error] Kotlin compiler error
    
warnings found and -Werror specified
Kotlin compiler error
        - Locations
- `:mylibrary:compileDebugKotlin`
```

### Expected Behavior

Both tasks show the reported set of problems, even if they are similar. As the problems are task-specific.

### Context (optional)

Originally reported here: [KT-85564](https://youtrack.jetbrains.com/issue/KT-85564) Problems API report: locations section for compiler errors/warnings doesn't contain all affected tasks

[gen-ide-check-min.zip](https://github.com/user-attachments/files/28743133/gen-ide-check-min.zip)

### Gradle version

9.5.1

### Build scan URL (optional)

_No response_

_No response_

donat:
Fixes #38147

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/kotlin-dsl-integ-tests/src/integTest/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPluginForOldestKotlinVersionTest.kt
+++ b/platforms/core-configuration/kotlin-dsl-integ-tests/src/integTest/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPluginForOldestKotlinVersionTest.kt
@@ -70,7 +70,12 @@ class KotlinDslPluginForOldestKotlinVersionTest : AbstractKotlinIntegrationTest(
 
         withDefaultSettings().appendText("""includeBuild("producer")""")
         withBuildScript("""plugins { id("some") }""")
-        executer.expectExternalDeprecatedMessage("    Language version $oldestKotlinLanguageVersion is deprecated and its support will be removed in a future version of Kotlin. Update the version to 2.2.")
+        // The Kotlin compiler reports this warning from both `compilePluginsBlocks` and `compileKotlin`. A problem
+        // carries the identity of the task that reported it, so the two are not deduplicated against each other.
+        val languageVersionDeprecation =
+            "    Language version $oldestKotlinLanguageVersion is deprecated and its support will be removed in a future version of Kotlin. Update the version to 2.2."
+        executer.expectExternalDeprecatedMessage(languageVersionDeprecation)
+        executer.expectExternalDeprecatedMessage(languageVersionDeprecation)
         build("help").apply {
             assertThat(output, containsString("some!"))
         }
--- a/platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblem.java
+++ b/platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblem.java
@@ -121,6 +121,7 @@ public class DefaultProblem implements Serializable, ProblemInternal {
             equal(contextualLabel, that.contextualLabel) &&
             equal(solutions, that.solutions) &&
             equal(originLocations, that.originLocations) &&
+            equal(contextualLocations, that.contextualLocations) &&
             equal(details, that.details) &&
             equal(exception, that.exception) &&
             equal(additionalData, that.additionalData);
@@ -128,7 +129,7 @@ public class DefaultProblem implements Serializable, ProblemInternal {
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(problemDefinition, contextualLabel, solutions, originLocations, details, exception, additionalData);
+        return Objects.hashCode(problemDefinition, contextualLabel, solutions, originLocations, contextualLocations, details, exception, additionalData);
     }
 
     @Override
--- a/platforms/ide/problems/src/main/java/org/gradle/problems/internal/services/DefaultProblemSummarizer.java
+++ b/platforms/ide/problems/src/main/java/org/gradle/problems/internal/services/DefaultProblemSummarizer.java
@@ -79,8 +79,9 @@ public class DefaultProblemSummarizer implements ProblemSummarizer {
 
     @Override
     public void emit(ProblemInternal problem, @Nullable OperationIdentifier id) {
+        // Attach the task location before the deduplication check so that identical problems reported by different tasks are not treated as duplicates.
+        problem = maybeAddTaskLocation(problem, id);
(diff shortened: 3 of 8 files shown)


donat:
@bot-gradle test this

donat:
@bot-gradle test this

reinsch82 (approved):

donat:
Turns out the fix needs more consideration.

reinsch82:
@bot-gradle test this

platforms/ide/problems/src/main/java/org/gradle/problems/internal/services/DefaultProblemSummarizer.java
```diff
@@ -79,8 +79,9 @@ public void report(File reportDir, ProblemConsumer validationFailures) {
 
     @Override
     public void emit(ProblemInternal problem, @Nullable OperationIdentifier id) {
```
jbartok, line 81: ❌ Task validation warnings are now reported TWICE for the SAME task:

 - **First report:** `DefaultNodeValidator.hasValidationProblems` calls `reporter.report(warnings)` while mutations are resolved. That happens before `DefaultTaskNodeExecutor` sets `ProblemTaskIdentityTracker`,
  and outside the execute-task operation, so `taskIdentityFor(id)` returns null and no task location is added.
  - **Second report:** `ValidateStep` reports the same problems again during task execution. This copy gets a `DefaultTaskLocation`.
  - **Why both get through:** `problem = maybeAddTaskLocation(problem, id)` now runs before `shouldEmit`, so the two copies hash differently.

Before the PR, the two copies were equal and the second one was dropped. After the PR they differ only by the task location, so both are emitted.

The test named "validation warnings are displayed once" was updated to expect an extra `receivedProblem(3)`. The new block doesn't assert locations, so it doesn't check which copy has
  the task location.

The resolve-mutations operation does know which task it's for: ResolveTaskMutationsDetails carries the TaskIdentity. A possible fix would be to attach the task location on that first report too. Both copies
  would then be equal again and deduplicate.

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblem.java
```diff
@@ -121,14 +121,15 @@ public boolean equals(Object o) {
             equal(contextualLabel, that.contextualLabel) &&
             equal(solutions, that.solutions) &&
             equal(originLocations, that.originLocations) &&
+            equal(contextualLocations, that.contextualLocations) &&
             equal(details, that.details) &&
             equal(exception, that.exception) &&
             equal(additionalData, that.additionalData);
     }
```
jbartok, line 128: 🤔 `equal(contextualLocations, that.contextualLocations)` and the matching `hashCode` cover every contextual location, not just the task location this PR is about.

  `DefaultProblemBuilder.addLocationsFromDiagnostics` adds `DefaultStackTraceLocation(fileLocation, diagnostics.getStack())` and `DefaultPluginIdLocation` to `contextLocations` in two cases:

  - problems reported from a build script with no explicit location
  - `LoggingDeprecatedFeatureHandler` deprecations that aren't `USER_CODE_DIRECT`

  For those problems, the stack trace is now part of the deduplication key:

  - The same problem reached through a different call stack or script line is now counted as distinct. Example: one indirect deprecation triggered from many project scripts. These used to collapse into a single
  entry.
  - Each copy counts toward the `SummarizerStrategy` threshold of 15 per `ProblemId` and adds an entry to `ProblemSummaryInfo.hashes`.
  - `shouldEmit` hashes the full `List<StackTraceElement>` on every report.

  The unit tests (the `DefaultProblemTest` "task location" row and `SummarizerStrategyTest`) only cover `DefaultTaskLocation`, so this wider change is untested. If only the task should matter, compare just the
  `TaskLocation` entries.

platforms/core-configuration/kotlin-dsl-integ-tests/src/integTest/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPluginForOldestKotlinVersionTest.kt
```diff
@@ -70,7 +70,12 @@ class KotlinDslPluginForOldestKotlinVersionTest : AbstractKotlinIntegrationTest(
 
         withDefaultSettings().appendText("""includeBuild("producer")""")
         withBuildScript("""plugins { id("some") }""")
-        executer.expectExternalDeprecatedMessage("    Language version $oldestKotlinLanguageVersion is deprecated and its support will be removed in a future version of Kotlin. Update the version to 2.2.")
+        // The Kotlin compiler reports this warning from both `compilePluginsBlocks` and `compileKotlin`. A problem
```
jbartok: 🤔 The PR says this change causes the second "Language version 2.0 is deprecated" message. Master already expects that message twice, without this fix:

  ```kotlin
  repeat(2) { executer.expectExternalDeprecatedMessage(...) }
  ```

  That line has the comment "emitted by both compilePluginsBlocks and compileKotlin" and was added in the Kotlin 2.4.20-Beta2 bump. This hunk conflicts with master, so recheck the expected count after rebasing
  rather than keeping the PR's version.

  Also, `ConsoleProblemEmitter` prints both copies with `ProblemWriter.simple()`, which leaves out the task path. Console users see two identical lines they can't tell apart.

jbartok (changes requested):

reinsch82:
@bot-gradle test this

PR #38388, opened 2026-07-01
Fix Kotlin object serialization

### LINKED ISSUE #22347: Support Sealed Objects in Configuration Cache Serializer
Given a custom kotlin Sealed class with some `object` implementations used in the Task Configuration state.

```kotlin
sealed class SealedObject {
  object A: SealedObject()
  object B: SealedObject()
}

abstract class FooTask : DefaultTask() {
  private val sealedObject: SealedObject = SealedObject.A

  @TaskAction
  fun action() {
        val a = when (sealedObject) {
            SealedObject.A -> "a"
            SealedObject.B -> "b"
            else -> "null"
        }
        println("Value of a=$a")
        println("Hashcode of enumVal: ${sealedObject.hashCode()}")
        println("HashCode of A: ${SealedObject.A.hashCode()}")
        println("HashCode of B: ${SealedObject.B.hashCode()}")
  }
}
```

### Expected Behavior
It is expected consistent task execution results with and without configuration cache.

### Current Behavior
When configuration cache is reused `FooTask.sealedObject` will have different hashCode than actual `SealedObject.A`
thus they will not be equal to each other. I.e. `FooTask.sealedObject === SealedObject.A` is FALSE.

### Context
It is well-known issue among Java serializers that create copies of object implementations of Kotlin's sealed classes.
In [Jackson](https://github.com/FasterXML/jackson-module-kotlin/issues/225)] for example.

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheSupportedKotlinTypesIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheSupportedKotlinTypesIntegrationTest.groovy
@@ -77,50 +77,128 @@ class ConfigurationCacheSupportedKotlinTypesIntegrationTest extends AbstractConf
         "org.jetbrains.kotlinx:kotlinx-datetime:0.4.0" | "kotlinx.datetime.LocalDateTime(2025, 1, 1, 1, 1, 1)" | "2025-01-01T01:01:01"
     }
 
-    def "preserves identity of Kotlin object singletons"() {
+    def "preserves the identity of a Kotlin object declared as #description across configuration cache reuse"() {
         buildKotlinFile """
-            sealed class SealedObject {
-                object A : SealedObject()
-                object B : SealedObject()
-            }
-
-            object TopLevelObject
+            $declaration
 
             abstract class FooTask : DefaultTask() {
                 @get:Internal
-                val sealedField: SealedObject = SealedObject.A
-                @get:Internal
-                val objectField: TopLevelObject = TopLevelObject
+                abstract val objectField: Property<Any>
 
                 @TaskAction
                 fun foo() {
-                    println("sealed identity = \${sealedField === SealedObject.A}")
-                    println("object identity = \${objectField === TopLevelObject}")
-                    val matched = when (sealedField) {
-                        SealedObject.A -> "A"
-                        SealedObject.B -> "B"
-                    }
-                    println("when matched = \$matched")
+                    println("identity = \${objectField.get() === $reference}")
+                    println("name = \${${reference}.NAME}")
                 }
             }
 
-            tasks.register("foo", FooTask::class)
+            tasks.register("foo", FooTask::class) {
+                objectField = $reference
+            }
         """
 
         when:
         configurationCacheRun "foo"
 
         then:
-        outputContains("sealed identity = true")
-        outputContains("object identity = true")
-        outputContains("when matched = A")
+        outputContains("identity = true")
+        outputContains("name = $name")
 
-        when: "the configuration cache entry is reused and the fields are deserialized"
+        when: "the configuration cache entry is reused and the field is deserialized"
         configurationCacheRun "foo"
 
         then:
-        outputContains("sealed identity = true")
-        outputContains("object identity = true")
-        outputContains("when matched = A")
+        outputContains("identity = true")
+        outputContains("name = $name")
+
+        where:
+        description                       | declaration                                                                                      | reference               | name
+        "a sealed object subtype"         | 'sealed class Sealed { object A : Sealed() { const val NAME = "sealed" }; object B : Sealed() }' | "Sealed.A"              | "sealed"
+        "a top-level object"              | 'object TopLevel { const val NAME = "top-level" }'                                               | "TopLevel"              | "top-level"
+        "an extension's companion object" | 'open class MyExtension { companion object { const val NAME = "extension" } }'                   | "MyExtension.Companion" | "extension"
+        "a named companion object"        | 'class Registry { companion object SomeName { const val NAME = "named" } }'                      | "Registry.SomeName"     | "named"
+        "a serializable object"           | 'object Marker : java.io.Serializable { const val NAME = "serializable" }'                       | "Marker"                | "serializable"
+    }
+
+    def "preserves the identity of a Kotlin object that is the task's own companion across configuration cache reuse"() {
+        buildKotlinFile """
+            abstract class FooTask : DefaultTask() {
+                companion object {
+                    const val NAME = "companion"
+                }
+
+                @get:Internal
+                abstract val objectField: Property<Any>
+
+                @TaskAction
+                fun foo() {
+                    println("identity = \${objectField.get() === FooTask.Companion}")
+                    println("name = \${FooTask.Companion.NAME}")
+                }
+            }
+
+            tasks.register("foo", FooTask::class) {
+                objectField = FooTask.Companion
+            }
+        """
+
+        when:
+        configurationCacheRun "foo"
+
+        then:
+        outputContains("identity = true")
+        outputContains("name = companion")
+
+        when: "the configuration cache entry is reused and the field is deserialized"
+        configurationCacheRun "foo"
+
+        then:
+        outputContains("identity = true")
+        outputContains("name = companion")
+    }
+
+    def "preserves the identity of a Kotlin object with custom Java serialization across configuration cache reuse"() {
+        buildKotlinFile """
+            object CustomSerializableObject : java.io.Serializable {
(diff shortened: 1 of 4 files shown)


ov7a:
@bot-gradle test this

platforms/core-configuration/stdlib-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/stdlib/KotlinObjectCodec.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    private
+    const val INSTANCE_FIELD_NAME = "INSTANCE"
+
+    override fun encodingForType(type: Class<*>): Encoding? =
+        KotlinObjectEncoding.takeIf { type.isKotlinObject() }
+
+    override suspend fun ReadContext.decode(): Any? {
+        val objectClass = readClass()
+        return objectClass.getDeclaredField(INSTANCE_FIELD_NAME).get(null)
```
ov7a: 🤔 Is it ok to bring kotlin-reflect as a runtime dependency and use kotlin helpers from it? https://github.com/gradle/gradle/compare/db88919c3aad206e653014b2494cd752626dcb48..b7e8bc59d8945a8b719e82f38d245685d0584837#diff-ced4b830833c5b6c731b69a15671f1f215f9cb3a49f0d788dd0a6cd5e0cc108dL32

(kotlin-reflect is already in daemon IIUC)

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheSupportedKotlinTypesIntegrationTest.groovy
```diff
@@ -76,4 +76,51 @@ class ConfigurationCacheSupportedKotlinTypesIntegrationTest extends AbstractConf
         dependency                                     | reference                                             | output
         "org.jetbrains.kotlinx:kotlinx-datetime:0.4.0" | "kotlinx.datetime.LocalDateTime(2025, 1, 1, 1, 1, 1)" | "2025-01-01T01:01:01"
     }
+
+    def "preserves identity of Kotlin object singletons"() {
```
tresat: Other test ideas:
- an object declared as a script-local variable remains identical to a task field
- an object declared as a companion to the task itself remains identical to the task field
- an object declared as a companion to an extension or plugin

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/serialize/ConfigurationCacheCodecs.kt
```diff
@@ -342,6 +343,7 @@ class DefaultConfigurationCacheCodecs(
         bind(ExternalizableCodec)
         bind(JavaObjectSerializationCodec(javaSerializationEncodingLookup))
         bind(ValueObjectCodec)
+        bind(KotlinObjectCodec)
```
tresat: ❗ What if the `object` is itself `Serializable`?  Wouldn't that `Codec` match first and result in losing the singleton guarantee?  I think this `Codec` needs to go first in the list.

platforms/core-configuration/stdlib-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/stdlib/KotlinObjectCodec.kt
```diff
@@ -0,0 +1,63 @@
...
+        return field.type == this &&
+            Modifier.isStatic(field.modifiers) &&
+            Modifier.isFinal(field.modifiers) &&
+            Modifier.isPublic(field.modifiers)
+    }
+}
+
+
+private
+object KotlinObjectEncoding : Encoding {
```
tresat: ❓ Can't `KotlinObjectCodec` implement `Encoding` as well?  Why keep this separate?

platforms/core-configuration/stdlib-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/stdlib/KotlinObjectCodec.kt
```diff
@@ -0,0 +1,63 @@
...
+import java.lang.reflect.Modifier
+
+
+/**
+ * Preserves the identity of Kotlin `object` singletons across the configuration cache.
+ */
+object KotlinObjectCodec : EncodingProducer, Decoding {
+
+    private
+    const val INSTANCE_FIELD_NAME = "INSTANCE"
```
tresat: ❗ Doesn't this return `false` for companion objects, since a companion instance lives in a `Companion` field on the enclosing class, not `INSTANCE`?

A test will prove it.

tresat (changes requested):

platforms/core-configuration/stdlib-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/stdlib/KotlinObjectCodec.kt
ov7a: It's an established pattern at this point:) See EnumEncoding, ProxyEncoding, OwnerServiceEncoding, etc.

IIUC, there's also a small benefit of returning a smaller isolated class

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheSupportedKotlinTypesIntegrationTest.groovy
ov7a: Great points! I've updated the PR

platforms/core-configuration/stdlib-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/stdlib/KotlinObjectCodec.kt
```diff
@@ -0,0 +1,72 @@
...
+                Modifier.isFinal(it.modifiers) &&
+                Modifier.isPublic(it.modifiers)
+        }
+}
+
+
+private
+object KotlinObjectEncoding : Encoding {
+    override suspend fun WriteContext.encode(value: Any) {
+        writeClass(value.javaClass)
```
abstratt: 💭 

Could be defensive and reject a value that is not the value of the `INSTANCE` static field.

platforms/core-configuration/stdlib-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/stdlib/KotlinObjectCodec.kt
```diff
@@ -0,0 +1,84 @@
...
+fun Class<*>.singletonInstance(): Any =
+    (singletonInstanceField() ?: error("Cannot find the singleton instance field of Kotlin object $name."))
+        .apply { isAccessible = true }
+        .get(null)
+
+
+private
+fun Class<*>.singletonInstanceField(): Field? {
+    if (!isAnnotationPresent(Metadata::class.java)) {
+        return null
```
abstratt, line 72:
```suggestion
        // not a Kotlin class
        return null
```

tresat (approved):

platforms/core-configuration/stdlib-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/stdlib/KotlinObjectCodec.kt
```diff
@@ -0,0 +1,85 @@
...
+ * Preserves the identity of Kotlin `object` singletons across the configuration cache.
+ */
+object KotlinObjectCodec : EncodingProducer, Decoding {
+
+    override fun encodingForType(type: Class<*>): Encoding? =
+        KotlinObjectEncoding.takeIf { type.singletonInstanceField() != null }
+
+    override suspend fun ReadContext.decode(): Any? {
+        val objectClass = readClass()
+        return objectClass.singletonInstance()
```
abstratt, line 40: 💭 

A singleton may have mutable state (even though that would be a bad idea - except for caching). IMO, it makes it worthwhile describing this change in behavior in the upgrade guide, and adding a note on that decision here.

abstratt (approved):
Looks good, just a couple of comments.

lkasso (approved):
Docs LGTM

Issue #35837, opened 2025-11-27
Deprecated 'api' and other accessors with "named argument notation" seem to miss "@Deprecated" annotations in bytecode


fedochet:
### Current Behavior

Recently, multiple methods for declaring dependencies with "named argument notation" were deprecated, see this [link](https://docs.gradle.org/current/userguide/best_practices_dependencies.html#single-gav-string).

The following method is also marked as deprecated:
```kt
@Deprecated("Use single-string notation instead")
fun DependencyHandler.`api`(
    group: String,
    name: String,
    version: String? = null,
    configuration: String? = null,
    classifier: String? = null,
    ext: String? = null,
    dependencyConfiguration: Action<ExternalModuleDependency>? = null
): ExternalModuleDependency = ...
```

However, if you look at the corresponding bytecode, it seems like the deprecation is actually missing there.

This affects tooling like IntelliJ IDEA, which is unable to highlight calls like `api(group = "org.mockito", name = "mockito-core", version = "1.9.5")` as deprecated.

### Expected Behavior

The bytecode should also contain all the information about the deprecation, since the tooling also relies on that information being present not only in the sources, but primarily in the bytecode.

### Context (optional)

The deprecations were introduced with [this commit](https://github.com/gradle/gradle/commit/8aa3362ab039942a34c4d53a72669cd676604b3d), and they seem to deprecate the “dependency map” notation across many methods.

I suspect that the reason for that is that the bytecode and the sources for the accessors are generated separately [here](https://github.com/gradle/gradle/blob/8aa3362ab039942a34c4d53a72669cd676604b3d/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorFragments.kt), and the `@Deprecated` annotation is omitted in the bytecode.

### Self-contained Reproducer Project

Attaching a simple reproducer project.

To reproduce, open it with a relatively fresh IntelliJ IDEA (2025.*), go to the `build.gradle.kts` in it, and observe that the deprecation is displayed for the `invoke` operator, but is not displayed for the explicit `api` call:

<img width="872" height="938" alt="Image" src="https://github.com/user-attachments/assets/b4796c5a-c31e-4ced-a06c-91432e5adb0d" />

[gradle-missing-deprecation-on-api-dependency.zip](https://github.com/user-attachments/files/23794915/gradle-missing-deprecation-on-api-dependency.zip)

### Gradle version

9.2.1

[...]

PR #38389, opened 2026-07-01
Deprecate KDSL multi-string dependency functions in bytecode

h0tk3y:
Add the missing deprecation in the Kotlin DSL accessors bytecode for multi-string dependency DSL functions, in addition to the deprecation in the generated sources (which were only effective in precompiled scripts but not in regular Kotlin DSL).

Fixes:
* https://github.com/gradle/gradle/issues/35837


hegyibalint (approved):
LGTM!

PR #38393, opened 2026-07-01
Reformat capabilities conflict error message for clarity

tresat:


platforms/documentation/docs/src/snippets/how-to/dependency-management/dependency-management/incompatible-variants/tests/runtimeClasspath.out
```diff
@@ -31,12 +31,12 @@ project ':producer'
    Failures:
       - Could not resolve project ':producer'.
           - Module 'org.gradle.demo:producer' has been rejected:
-               Cannot select module with conflict on capability 'org.gradle.demo:producer-db-support:1.0' also provided by ['project ':producer'' (postgresSupportRuntimeElements)]
+               Cannot select module because of conflict with project ':producer' (postgresSupportRuntimeElements). All provide capability 'org.gradle.demo:producer-db-support:1.0'.
```
lkasso: WDYT?
```suggestion
               Cannot select module because of conflict with project ':producer' (postgresSupportRuntimeElements). Both provide capability 'org.gradle.demo:producer-db-support:1.0'.
```

tresat: I made it "all" because there could actually be more than 2 conflicting capabilities.  But 2 is probably far more common, I'll special case it.

lkasso: It is much easier to understand what the issue is when it says "both".

lkasso (approved):

big-guy (approved):

PR #38394, opened 2026-07-01
Remove jetty dependency from test infrastructure

ov7a:
- **Removed:** all Jetty usage (28 `testLibs.jetty*`, `javax.servlet-api`, and Selenium (it was unused).

### Behavior-preservation notes (com.sun differs from Jetty)
- **Expect: 100-continue** — com.sun always auto-sends `100 Continue` before the handler with no way to reject; the two build-cache tests needing rejection/redirect use a small purpose-built raw-socket server.
- **SSL "read timeout"** — `sslPreHandler` now runs at request-handling time (like Jetty's `HttpConfiguration.Customizer`) rather than during the TLS handshake.
- **NTLM** — track authenticated *and* failed connections so a completed handshake isn't re-challenged (round-trip counts) and bad credentials don't loop.
- **Premature-end / dropped connections** — honor an explicit `Content-Length` and close the exchange mid-body to reproduce truncation; widened one build-cache error assertion to cover the full connection-drop message set (`Broken pipe`/`Connection reset`/…), which is nondeterministic at the TCP layer.
- **Directory listings** — `StaticFileHandler` serves Apache-style listings so perf tests' `latest.release` version enumeration keeps working.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ov7a:
@coderabbitai review

ov7a:
@bot-gradle test RfR

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpServer.groovy
blindpirate, line 1: toString() still reads server.started, but getServer() now returns com.sun.net.httpserver.HttpServer (no started property, and null before start()).

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HandlerChain.groovy
```diff
@@ -0,0 +1,42 @@
...
+    void addHandler(HttpResourceHandler handler) {
+        handlers.add(handler)
+    }
+
+    void clear() {
+        handlers.clear()
+    }
+
+    void handle(String target, HttpRequest request, HttpResponse response) throws IOException {
+        for (HttpResourceHandler handler : handlers) {
```
blindpirate, line 36: `handlers` is not thread-safe.

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpResponse.groovy
```diff
@@ -0,0 +1,165 @@
...
+package org.gradle.test.fixtures.server.http
+
+import com.sun.net.httpserver.HttpExchange
+
+import java.nio.charset.StandardCharsets
+import java.text.SimpleDateFormat
+
+class HttpResponse {
+    private final HttpExchange exchange
+    private final ByteArrayOutputStream buffer = new ByteArrayOutputStream()
```
blindpirate, line 30: The response body is fully buffered here, so the distribution integration tests would hold several hundreds of MBs byte array, while test JVM default is 512M.

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpServerDispatcher.groovy
```diff
@@ -0,0 +1,78 @@
...
+        try {
+            if (exchange instanceof HttpsExchange) {
+                sslPreHandler.handshakeStarted()
+            }
+            loggingHandler.log(request)
+            securityHandlerWrapper.handle(request.pathInfo, request, response)
+            if (!request.handled) {
+                customHandler.handle(request.pathInfo, request, response)
+            }
+        } catch (Throwable t) {
```
blindpirate, line 55: Shall we log the error here? It's only visible in the response body.

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpServerDispatcher.groovy
```diff
@@ -0,0 +1,78 @@
...
+            if (!response.isCommitted()) {
+                try {
+                    response.sendError(500, String.valueOf(t))
+                } catch (Exception ignore) {
```
blindpirate: Why ignored?

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpServerDispatcher.groovy
```diff
@@ -0,0 +1,78 @@
...
+            if (exchange instanceof HttpsExchange) {
+                sslPreHandler.handshakeStarted()
+            }
+            loggingHandler.log(request)
+            securityHandlerWrapper.handle(request.pathInfo, request, response)
+            if (!request.handled) {
+                customHandler.handle(request.pathInfo, request, response)
+            }
+        } catch (Throwable t) {
+            if (!response.isCommitted()) {
```
blindpirate, line 57: So, if the handler already starts to write response, then an error happens, it's permanently lost?

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpServerFixture.groovy
```diff
@@ -16,59 +16,61 @@
...
-    private ServerConnector connector
-    private ServerConnector sslConnector
+    private com.sun.net.httpserver.HttpServer httpServer
+    private HttpsServer httpsServer
+    private HttpServerDispatcher dispatcher
+    private LoggingHandler loggingHandler
+    private boolean running
     private final SslPreHandler sslPreHandler = new SslPreHandler()
     private final SecuredHandlerCollection securityHandlerWrapper = new SecuredHandlerCollection()
+    private final ExecutorService executor = Executors.newCachedThreadPool({ Runnable r ->
```
blindpirate: Stop this in `stop()`.

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpServer.groovy
ov7a: good catch, fixed!

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpResponse.groovy
```diff
@@ -0,0 +1,165 @@
...
+
+    HttpResponse(HttpExchange exchange) {
+        this.exchange = exchange
+    }
+
+    void setStatus(int status) {
+        this.status = status
+    }
+
+    void setStatus(int status, String reason) {
```
blindpirate: `reason` is lost.

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HandlerChain.groovy
ov7a: yes, but they shouldn't be, right? There's no mutithread access to them, IIUC

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/NtlmAuthenticator.groovy
```diff
@@ -17,113 +17,154 @@
...
+@CompileStatic
+class NtlmAuthenticator implements AuthScheme.Authenticator {
+    private static final int TYPE_1_NEGOTIATE = 1
+    private static final int TYPE_3_AUTHENTICATE = 3
+    private static final int SERVER_CHALLENGE_LENGTH = 8
+
+    private final TestUserRealm realm
+    private final Map<String, byte[]> challenges = new ConcurrentHashMap<>()
+    private final Set<String> authenticatedConnections = ConcurrentHashMap.newKeySet()
+    private final Set<String> failedConnections = ConcurrentHashMap.newKeySet()
```
blindpirate, line 37: authenticatedConnections and failedConnections should remove the key `remoteAddr + ":" + remotePort` when connection closed. Otherwise the OS recycles an ephemeral port (realistic in soak/stress runs), a brand-new connection is either treated as pre-authenticated (handshake silently skipped) or permanently 401'd via a stale failedConnections entry.

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/AuthScheme.groovy
```diff
@@ -46,114 +39,170 @@ enum AuthScheme {
...
-    private static class HttpHeaderAuthHandler extends AuthSchemeHandler {
-        @Override
-        protected String constraintName() {
-            return TestHttpHeaderAuthenticator.AUTH_SCHEME_NAME;
+        private boolean validate(String method, Map<String, String> params) {
+            String username = params.get("username")
+            if (username == null || username != realm.username) {
+                return false
+            }
+            String nonce = params.get("nonce")
```
blindpirate, line 173: This is not validated. The client could forge a nonce and md5.

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/AuthScheme.groovy
```diff
@@ -16,27 +16,20 @@
...
+import java.security.MessageDigest
+import java.security.SecureRandom
+import java.nio.charset.StandardCharsets
+
+@CompileStatic
 enum AuthScheme {
-    BASIC(new BasicAuthHandler()),
+    BASIC(new BasicAuthHandler(false)),
     DIGEST(new DigestAuthHandler()),
-    HIDE_UNAUTHORIZED(new HideUnauthorizedBasicAuthHandler()),
```
blindpirate, line 39: This is a behavior change. `HideUnauthorizedBasicAuthHandler` inherited `containsUnexpectedAuthentication()` from `AuthSchemeHandler`, which returns `false`, but the new behavior is returning `true`.

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpResponse.groovy
```diff
@@ -0,0 +1,165 @@
...
+    void addHeader(String name, String value) {
+        if (name.equalsIgnoreCase("Transfer-Encoding") || name.equalsIgnoreCase("Content-Length")) {
+            setHeader(name, value)
+            return
+        }
+        exchange.responseHeaders.add(name, value)
+    }
+
+    void setDateHeader(String name, long date) {
+        SimpleDateFormat format = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss 'GMT'", Locale.US)
```
blindpirate: Not thread-safe, use `DateTimeFormatter.RFC_1123_DATE_TIME` instead.

platforms/core-execution/build-cache-http/src/integTest/groovy/org/gradle/caching/http/internal/ExpectContinueHttpServer.groovy
```diff
@@ -0,0 +1,197 @@
...
+        StringBuilder response = new StringBuilder("HTTP/1.1 ").append(status).append(" ").append(reason).append("\r\n")
+        headers.each { name, value -> response.append(name).append(": ").append(value).append("\r\n") }
+        response.append("Content-Length: 0\r\n")
+        // Close after each response so redirect-following uses a fresh connection (no keep-alive bookkeeping).
+        response.append("Connection: close\r\n\r\n")
+        output.write(response.toString().getBytes(StandardCharsets.US_ASCII))
+        output.flush()
+    }
+
+    private static byte[] readBody(InputStream input, int length) {
```
blindpirate: `InputStream.readNBytes()`.

platforms/core-execution/build-cache-http/src/integTest/groovy/org/gradle/caching/http/internal/ExpectContinueHttpServer.groovy
```diff
@@ -0,0 +1,197 @@
...
+            if (count < 0) {
+                break
+            }
+            read += count
+        }
+        return body
+    }
+
+    /** Reads a single CRLF-terminated line, returning it without the trailing CRLF (null at end of stream). */
+    private static String readLine(InputStream input) {
```
blindpirate, line 158: `input` is raw socket, better read in buffer

testing/internal-performance-testing/src/main/groovy/org/gradle/performance/StaticFileHandler.groovy
```diff
@@ -0,0 +1,74 @@
...
+        exchange.responseHeaders.set("Content-Type", contentType)
+        exchange.sendResponseHeaders(200, head ? -1 : bytes.length)
+        if (!head) {
+            OutputStream out = exchange.responseBody
+            out.write(bytes)
+            out.close()
+        }
+    }
+
+    private static byte[] directoryListing(File dir, String requestPath) {
```
blindpirate: Use `HttpServer.sendDirectoryListing()` with correct escaping.

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpServerDispatcher.groovy
```diff
@@ -0,0 +1,78 @@
...
+            }
+            try {
+                response.commit(request.method)
+            } catch (Exception ignore) {
+            }
+            exchange.close()
+        }
+    }
+    
+    private static void drainRequestBody(HttpExchange exchange) {
```
blindpirate, line 73: Use Guava `ByteStreams.exhaust()` if it's already used on classpath.

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpServerDispatcher.groovy
ov7a: Exception is now logged and the state is cleared

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpResponse.groovy
ov7a: This is sorta expected because sun server always returns the standard reason. I removed the param to avoid confusion

ov7a:
@bot-gradle test RfN

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HandlerChain.groovy
blindpirate: >  yes, it's possible, and in the timeout tests it's even routine 
> - Mutation while running: HttpServer.resetExpectations() → reset() → SecuredHandlerCollection.reset() → handlers.clear() on the HandlerChain, and every server.expect(...)/allow(...) → collection.addHandler(...). All of these happen from the test thread with the server still running — that's the fixture's normal multi-build pattern (expect → build → resetExpectations() → expect → build again).
> - Iteration: HandlerChain.handle's for-each loop, executed on the http-fixture executor threads, one per in-flight exchange.

Previously Jetty uses `CopyOnWriteArrayList`, which could be a cheap solution.

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpServerFixture.groovy
```diff
@@ -117,104 +126,105 @@ trait HttpServerFixture {
...
-            shutdownConnector(connector)
-            connector = null
+        if (httpServer != null) {
+            httpServer.stop(0)
+            httpServer = null
         }
-
-        server?.stop()
+        if (executor != null) {
+            executor.shutdown()
```
blindpirate:
```suggestion
            executor.shutdownNow()
```

This also interrupts the threads.

blindpirate (approved):
LGTM with two tiny comments.

PR #38396, opened 2026-07-01
Sync should empty destination when source becomes empty

### LINKED ISSUE #37597: Sync task is skipped with NO-SOURCE if inputs become empty instead of deleting stale outputs

abstratt:
Sync should not be skipped if inputs are empty, as it should make the destination dir empty as well.  

This (when out of draft) shall supersede #37836, by @devareddy05 (it contains commits by @devareddy05 from that branch).

Issue: #37597

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -337,6 +337,14 @@ If your build requires shared, synchronized state, use <<build_services.adoc#bui
 
 For a complete list of types that are incompatible with the Configuration Cache, see the <<configuration_cache_requirements.adoc#config_cache:requirements:disallowed_types, documentation>>.
 
+[[sync_runs_when_source_is_empty]]
+==== `Sync` task is no longer skipped when its source files are empty
+
+Previously, a link:{javadocPath}/org/gradle/api/tasks/Sync.html[`Sync`] task with no source files was reported as `NO-SOURCE` and skipped, leaving stale files in its destination directory.
+The task now runs even when its sources are empty, ensuring stale outputs are deleted as `Sync` is documented to do.
+
+If you previously relied on the skip behavior, configure your task to short-circuit explicitly with `onlyIf { ... }`.
+
 [[v9_upgrade_to_kotlin_2_3_21]]
 ==== Upgrade to Kotlin 2.3.21
 
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -217,10 +217,13 @@ If you have custom code that branched on `getParameters() == null` to detect a p
 [[sync_runs_when_source_is_empty]]
 ==== `Sync` task is no longer skipped when its source files are empty
 
-Previously, a link:{javadocPath}/org/gradle/api/tasks/Sync.html[`Sync`] task with no source files was reported as `NO-SOURCE` and skipped, leaving stale files in its destination directory.
-The task now runs even when its sources are empty, ensuring stale outputs are deleted as documented.
+Previously, a link:{javadocPath}/org/gradle/api/tasks/Sync.html[`Sync`] task with no source files was reported as `NO-SOURCE` and skipped, leaving any previously synced files in its destination directory in place indefinitely.
 
-If you previously relied on the skip behavior, configure your task to short-circuit explicitly:
+`Sync` now always runs its copy action, even when its source is empty. Once a `Sync` task has successfully synced into a given destination directory at least once, a later run with an empty source deletes stale files from that destination — anything not covered by its `preserve()` filter — exactly as `Sync` is documented to do for any other source change.
+
+As a safety measure, this cleanup only happens for a destination the task has synced into before. The very first execution of a `Sync` task against a given destination directory is left untouched if its source is empty: nothing is deleted, and no assumption is made about pre-existing, unrelated content sitting at that destination.
+
+If you rely on the previous skip behavior for a `Sync` task that has already run against its destination, configure it to short-circuit explicitly:
 ====
 [.multi-language-sample]
 =====
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -77,10 +77,9 @@ If you relied on mutable state surviving a cache hit, move it into a task proper
 
 Previously, a link:{javadocPath}/org/gradle/api/tasks/Sync.html[`Sync`] task with no source files was reported as `NO-SOURCE` and skipped, leaving any previously synced files in its destination directory in place indefinitely.
 
-`Sync` now always runs its copy action, even when its source is empty, and behaves the same way whether or not its destination is owned by Gradle.
-Once a `Sync` task has successfully synced into a given destination directory at least once, a later run with an empty source deletes stale files from that destination — anything not matched by the task's `preserve { ... }` filter.
+`Sync` now always runs its copy action, even when its source is empty, deleting everything in its destination directory that is not matched by the task's `preserve { ... }` filter.
 
-If you rely on the previous skip behavior for a `Sync` task that has already run against a destination not owned by Gradle, configure it to short-circuit explicitly:
+If you rely on the previous skip behavior, configure the task to short-circuit explicitly:
 
 ====
 [.multi-language-sample]
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -341,9 +341,31 @@ For a complete list of types that are incompatible with the Configuration Cache,
 ==== `Sync` task is no longer skipped when its source files are empty
 
 Previously, a link:{javadocPath}/org/gradle/api/tasks/Sync.html[`Sync`] task with no source files was reported as `NO-SOURCE` and skipped, leaving stale files in its destination directory.
-The task now runs even when its sources are empty, ensuring stale outputs are deleted as `Sync` is documented to do.
+The task now runs even when its sources are empty, ensuring stale outputs are deleted as documented.
 
-If you previously relied on the skip behavior, configure your task to short-circuit explicitly with `onlyIf { ... }`.
+If you previously relied on the skip behavior, configure your task to short-circuit explicitly:
+====
+[.multi-language-sample]
+=====
+.build.gradle.kts
+[source,kotlin]
+----
+tasks.named<Sync>("mySync") {
+    onlyIf { !source.isEmpty }
+}
+----
+=====
+[.multi-language-sample]
+=====
+.build.gradle
+[source,groovy]
+----
+tasks.named('mySync') {
+    onlyIf { !source.empty }
+}
+----
+=====
+====
 
 [[v9_upgrade_to_kotlin_2_3_21]]
 ==== Upgrade to Kotlin 2.3.21
--- a/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
@@ -214,6 +214,36 @@ Null-checks against the result are no longer required.
 No source change is needed in typical builds.
 If you have custom code that branched on `getParameters() == null` to detect a parameterless action, replace it with `getParameters() instanceof WorkParameters.None` (or the equivalent `None` type).
 
+[[sync_runs_when_source_is_empty]]
+==== `Sync` task is no longer skipped when its source files are empty
+
+Previously, a link:{javadocPath}/org/gradle/api/tasks/Sync.html[`Sync`] task with no source files was reported as `NO-SOURCE` and skipped, leaving stale files in its destination directory.
+The task now runs even when its sources are empty, ensuring stale outputs are deleted as documented.
+
+If you previously relied on the skip behavior, configure your task to short-circuit explicitly:
+====
+[.multi-language-sample]
+=====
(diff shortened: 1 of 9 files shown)


abstratt:
@bot-gradle test RFN

abstratt:
@bot-gradle test RFN

lkasso (approved):
Docs LGTM.

subprojects/core/src/main/java/org/gradle/api/tasks/AbstractCopyTask.java
```diff
@@ -116,6 +116,13 @@ protected CopySpecInternal createRootSpec() {
 
     protected abstract CopyAction createCopyAction();
 
+    /**
+     * Whether this task should be skipped when its source files are empty.
+     */
+    /* package */ boolean shouldSkipWhenSourceIsEmpty() {
```
asodja, line 140: 💅 Why we need `/* package */`?

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
```diff
@@ -71,6 +80,61 @@ public abstract class Sync extends AbstractCopyTask {
...
+    @Override
+    boolean shouldSkipWhenSourceIsEmpty() {
+        return false;
+    }
+
+    @Override
+    protected void copy() {
+        File destinationDir = getDestinationDir();
+        if (destinationDir != null && !hasPreviousOutputFilesUnder(destinationDir)) {
+            // performance: check file system contents last
```
asodja: 💅 Do we need this comment?

```
// performance: check file system contents last
```

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
```diff
@@ -71,6 +80,61 @@ public abstract class Sync extends AbstractCopyTask {
...
+    }
+
+    /**
+     * Whether the source has no files and no directories. Unlike {@link org.gradle.api.file.FileCollection#isEmpty()},
+     * which counts regular files only, this also counts empty directories (needed to honor
+     * {@code ignoreEmptyDirectories(false)}).
+     */
+    private boolean isSourceEmpty() {
+        MutableBoolean found = new MutableBoolean();
+        getSource().getAsFileTree().visit(new EmptyFileVisitor() {
```
asodja: ❓ Can we simplify that to:
```
getSource().getAsFileTree().visit(fileDetails -> {
    found.set(true);
    fileDetails.stopVisiting();
});
```
?

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
```diff
@@ -71,6 +80,61 @@ public abstract class Sync extends AbstractCopyTask {
...
 
+    @Override
+    boolean shouldSkipWhenSourceIsEmpty() {
+        return false;
+    }
+
+    @Override
+    protected void copy() {
+        File destinationDir = getDestinationDir();
+        if (destinationDir != null && !hasPreviousOutputFilesUnder(destinationDir)) {
```
asodja: 💅 Can we shorten this a bit?

It could be just:
```
@Override
protected void copy() {
    File destinationDir = getDestinationDir();
    if (destinationDir != null && !hasPreviousOutputFilesUnder(destinationDir) && isSourceEmpty()) {
        // Empty source without history: skip to avoid wiping external files.
        // Untracked tasks have no history, so they always take this path when empty.
        setDidWork(false);
        return;
    }
    super.copy();
}
```

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -72,6 +72,40 @@ As a consequence, only the singleton's identity is restored, not any mutable sta
 Previously, each `object` was serialized as a regular bean and rebuilt as a _new_ instance on a cache hit, so `deserialized === TheObject` was `false`.
 If you relied on mutable state surviving a cache hit, move it into a task property or another bean type that is serialized by value.
 
+[[sync_runs_when_source_is_empty]]
+==== `Sync` task is no longer skipped when its source files are empty
+
+Previously, a link:{javadocPath}/org/gradle/api/tasks/Sync.html[`Sync`] task with no source files was reported as `NO-SOURCE` and skipped, leaving any previously synced files in its destination directory in place indefinitely.
```
asodja: ❓ Is this comment true only for dirs outside of `build/` https://github.com/gradle/gradle/issues/37597#issuecomment-4857944501 . So issue was not present for `build/`?

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
```diff
@@ -71,6 +80,61 @@ public abstract class Sync extends AbstractCopyTask {
...
+            @Override
+            public void visitDir(FileVisitDetails dirDetails) {
+                found.set(true);
+                dirDetails.stopVisiting();
+            }
+        });
+        return !found.get();
+    }
+
+    private boolean hasPreviousOutputFilesUnder(File destinationDir) {
```
asodja: ❓ Could there be any weird behavior if execution history is not present (e.g. Gradle version upgrade or user deletes .gradle)?

subprojects/core/src/main/java/org/gradle/api/tasks/AbstractCopyTask.java
abstratt: It is a convention to make it clear we really want the default visibility. We have other instances in the codebase, but don't follow it strictly. Some tools, like PMD,  have [even lint rules](https://pmd.github.io/pmd/pmd_rules_java_codestyle.html#commentdefaultaccessmodifier) for this:

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
abstratt: To avoid someone to merge the conditions into a single if, and potentially reordering the conditions, risking hitting a system call when one may not be necessary.

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
abstratt: Yes, we learned that afterwards. Do you think we should be explicit about that here?

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
```diff
@@ -71,6 +80,61 @@ public abstract class Sync extends AbstractCopyTask {
...
+                found.set(true);
+                fileDetails.stopVisiting();
+            }
+
+            @Override
+            public void visitDir(FileVisitDetails dirDetails) {
+                found.set(true);
+                dirDetails.stopVisiting();
+            }
+        });
```
abstratt:
```suggestion
        getSource().getAsFileTree().visit(fileDetails -> {
            found.set(true);
            fileDetails.stopVisiting();
        });
```

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
```diff
@@ -71,6 +80,61 @@ public abstract class Sync extends AbstractCopyTask {
...
+            // performance: check file system contents last
+            if (isSourceEmpty()) {
+                // Empty source and no record of a prior sync into this destination:
+                // possibly a misconfigured source pointing at an unrelated directory, so do nothing rather than
+                // wipe content this task never wrote. (Untracked tasks keep no history, so they always
+                // take this branch on an empty source.)
+                setDidWork(false);
+                return;
+            }
+        }
```
abstratt:
```suggestion
        if (destinationDir != null && !hasPreviousOutputFilesUnder(destinationDir) && isSourceEmpty()) {
            // Empty source without history: skip to avoid wiping external files.
            // Untracked tasks have no history, so they always take this path when empty.
            setDidWork(false);
            return;
        }
```

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
abstratt: This is not a regression. Without history, the old path did nothing either. The change only adds cleanup where history is present; where it isn't, behavior is byte-for-byte what it was.

It self-heals: The next run with a non-empty source syncs normally, deletes stale files (honoring preserve), and re-records history. From then on, empty-source runs clean up again. So the worst case is one "stale until the next real sync" window.

With no record of what it wrote, the task can't distinguish its own stale output from unrelated user content in the destination, so declining to delete is the safe branch - the same reasoning that makes untracked tasks always take this path.

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
abstratt: Done!

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
abstratt: Done, thanks!

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
abstratt: I made further changes to the upgrade guide to clarify that

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
asodja: To try to understand this a bit more, I see this JavaDoc on Sync:
```
 * This task is like the {@link Copy} task, except the destination directory will only contain the files
 * copied. All files that exist in the destination directory will be deleted before copying files, unless
 * a {@link #preserve(Action)} is specified.
```

So I guess old path was also doing different what JavaDoc says?

abstratt: Yes - that's the bug this PR fixes. `Sync` inherited `skipWhenEmpty(true)` from `AbstractCopyTask`, so an empty source meant `NO-SOURCE` and an untouched destination, contradicting that Javadoc every time (#37597).

Now the documented behavior is the rule. The only exception is when we can't prove the destination is ours:

- it is not build-owned,
- there are no recorded previous outputs under it, and
- the source is empty.

asodja: So from this Javadoc the idea is to always delete destination folder, and if user don't like that, they need to specify preserve filter.

What I want to understand specifically is, why we added this exception `The only exception is when we can't prove the destination is ours` in this PR?

Is it because of possible different behavior/backward incompatiblity or something else? It seems I am missing a bit of context from previous discussions

asodja (approved):
LGTM

abstratt:
@asodja Please take another look. The new behavior is now opt-in, via a new property in `Sync`

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
```diff
@@ -74,6 +75,38 @@ public abstract class Sync extends AbstractCopyTask {
...
+     *
+     * @return whether this task is skipped when its source is empty
+     * @since 9.8.0
+     */
+    @Incubating
+    @Input
+    public abstract Property<Boolean> getSkipWhenSourceIsEmpty();
+
+    @Override
+    boolean shouldSkipWhenSourceIsEmpty() {
```
asodja, line 111: 🤔 I wonder if we can avoid this method, by modifing inputs somehow. But I think that is not possible, right?

E.g. we would do in constructor of sync:
`getInput(<name>).skipWhenEmpty(getSkipWhenSourceIsEmpty())`

asodja (approved):
LGTM

subprojects/core/src/main/java/org/gradle/api/tasks/Sync.java
abstratt: I don't know of a way to look up registered inputs. The registration code at construction time needs to be able to determine whether the input may ever need to be *not* skipped, and that is different for `Copy` and other tasks and `Sync`.

asodja: I didn't find anything either. I think we should merge what we have.

I still think that this behavior is weird
> a destination inside the build directory is cleaned up, while one outside it keeps the files the source no longer contains

and maybe we should make `skipWhenSourceIsEmpty = false` default with Gradle 10

abstratt: It is a matter of safety. Gradle-owned locations have content that is derived from the source, so it is inherently disposable. That is not true for other locations. 

I am not sure we should ever change the default, not many people seem to complain about the current behavior.

PR #38401, opened 2026-07-02
Update Documentation: Document opportunistic Maven local artifact reuse

### LINKED ISSUE #38329: Artifacts are served from local Maven (`~/.m2`) during resolution from a remote repository, even when `mavenLocal()` is not declared
### Current Behavior

If a byte-identical copy of the artifact already exists in `~/.m2/repository`, Gradle uses
the local file and never downloads the artifact body from the remote repository, even
though `mavenLocal()` is not declared anywhere in the build.

Gradle still contacts the remote for metadata and checksums (a `HEAD` on the artifact and a
`GET` on its `.sha1`), learns the expected SHA-1, finds a file with that SHA-1 under
`~/.m2`, and serves the body from there. At `--info` it logs:

```
Found locally available resource with matching checksum: [http://.../widget-1.0.0.jar, <~/.m2>/.../widget-1.0.0.jar]
```
The only way I found to prevent it is to point `-Dmaven.repo.local` at an empty/nonexistent directory.

### Expected Behavior

When a build declares only a remote (HTTP) Maven repository and does not declare
`mavenLocal()`, resolving a dependency should download the artifact from the declared remote
repository. The local Maven repository (`~/.m2/repository`) should not be consulted as a
source of artifacts.

Is it intended that `~/.m2/repository` participates in resolution when `mavenLocal()` is not
declared? If so, is there a supported way to disable the Maven-local candidate finder for a
build (other than pointing `-Dmaven.repo.local` at an empty directory)?

### Context (optional)

This breaks any workflow that relies on observing or controlling what the declared remote
repository actually serves. For example, download proxies/ repository mirrors that capture artifacts as they are fetched. AndroidX's `development/importMaven` runs a local proxy in front of
  Google/Maven Central and writes the fetched bytes into a prebuilts mirror. When the
  requested artifact already exists in the developer's `~/.m2`, Gradle serves it locally, the
  proxy never sees the body, and nothing is written to the mirror. The import silently does
  nothing.

A self-contained, hermetic reproducer is attached (`gradle-m2-shortcircuit-repro.tgz`). It
uses a fake `~/.m2` via `-Dmaven.repo.local` and never touches the real one.

```bash
tar xzf gradle-m2-shortcircuit-repro.tgz && cd gradle-m2-repro
export JAVA_HOME=/path/to/jdk21
./repro.sh
```

What it does:

1. Publishes `com.example:widget:1.0.0` (plain POM + JAR, no Gradle Module Metadata) into a
   directory that is then served over HTTP. The only declared repository in the consumer
   build. `mavenLocal()` is not declared.
2. Copies the published artifact into a fake `~/.m2` (byte-identical → same SHA-1).
3. Resolves the dependency in the consumer build twice against a request-logging HTTP server,
   wiping `$GRADLE_USER_HOME/caches` before each run:
   - **Run A**: `maven.repo.local` = the populated fake `~/.m2`.
   - **Run B**: `maven.repo.local` = an empty directory (control).

The consumer build is essentially:

```kotlin
// consumer/build.gradle.kts
repositories {
    maven {
        url = uri("http://127.0.0.1:$repoPort/")
        isAllowInsecureProtocol = true
    }
}

val widget by configurations.creating
dependencies { widget("com.example:widget:1.0.0") }

tasks.register("resolveWidget") {
    val files = configurations["widget"]
    doLast { files.files.forEach { println(it) } }
}
```

Run with `-Dmaven.repo.local=<dir-containing-the-artifact>`.

### Observed output

```
HTTP requests the remote received when ~/.m2 WAS populated:
    GET  /
    HEAD /com/example/widget/1.0.0/widget-1.0.0.pom
    GET  /com/example/widget/1.0.0/widget-1.0.0.pom.sha1
    HEAD /com/example/widget/1.0.0/widget-1.0.0.jar
    GET  /com/example/widget/1.0.0/widget-1.0.0.jar.sha1
    --> GET of jar body (widget-1.0.0.jar): 0     # never downloaded from the remote

HTTP requests the remote received when ~/.m2 was EMPTY (control):
    GET  /
    GET  /com/example/widget/1.0.0/widget-1.0.0.pom
    GET  /com/example/widget/1.0.0/widget-1.0.0.jar
    --> GET of jar body (widget-1.0.0.jar): 1     # downloaded from the remote
```

## Root cause analysis

I traced this to the locally-available-resource lookup, which always includes
`~/.m2/repository` regardless of declared repositories:

* `LocallyAvailableResourceFinderFactory.java#create()` adds `~/.m2/repository` as a checksum-indexed candidate location 
*  `DefaultLocalMavenRepositoryLocator.java#getLocalMavenRepository()` resolves that location: `-Dmaven.repo.local` →  `~/.m2/settings.xml` `<localRepository>` → default `~/.m2/repository`.
* `DefaultCacheAwareExternalResourceAccessor.java#getResource()` fetches the remote checksum, calls
  `additionalCandidates.findByHashValue(sha1)`, and on a match copies the local file into the
  cache (`copyCandidateToCache`) and returns, skipping the body download. The
  `Found locally available resource with matching checksum` log line is emitted here.

This appears to be long-standing behavior, not a regression in 9.6.

[gradlem2shortcircuitrepro.tgz](https://github.com/user-attachments/files/29376656/gradlem2shortcircuitrepro.tgz)

### Gradle version

All

### Build scan URL (optional)

_No response_

_No response_

ljacomet:
Fixes #38329

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/reference/dependency-management/declaring-repositories/supported_repository_types.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/reference/dependency-management/declaring-repositories/supported_repository_types.adoc
@@ -68,7 +68,7 @@ include::sample[dir="snippets/reference/dependency-management/declaring-reposito
 include::sample[dir="snippets/reference/dependency-management/declaring-repositories/defineRepository/groovy",files="build.gradle[tags=maven-local]"]
 ====
 
-NOTE: Gradle manages its own <<dependency_caching.adoc#sec:dependency-cache,cache>> and doesn't need to declare the local Maven repository even if you resolve dependencies from a remote Maven repository.
+NOTE: Gradle manages its own <<dependency_caching.adoc#sec:dependency-cache,cache>> and doesn't need to declare the local Maven repository even if you resolve dependencies from a remote Maven repository. Even without `mavenLocal()` declared, Gradle may reuse already-downloaded artifacts from it as a cache — see <<dependency_caching.adoc#sec:cache-artifact-reuse,Artifact reuse>>.
 
 Gradle uses the same logic as Maven to identify the location of your local Maven cache.
 If a `settings.xml` file is defined in the user’s home directory (`~/.m2/settings.xml`), this location takes precedence over `__M2_HOME__/conf`


ghale (approved):
LGTM

Issue #36771, opened 2026-02-16
Make building commit distributions IP-compatible


alllex:
When running a performance test against a baseline that has IP enabled (like the current master or anything later), then building the distribution will fail with:
```
  Configuration Cache cannot be disabled when Isolated Projects is enabled.
```

That's because CC is currently being explicitly disabled (https://github.com/gradle/gradle/issues/36770). It will require some work to have it enabled. However, it's not enough.

The build logic involved in the BuildCommitDistribution is not currently IP-compatible, so it would have to be fixed.

---

For now, it's only a problem for local runs of performance tests, because they use the fork-point baseline, meaning that IP is already enabled there.

On CI, it will become a problem as soon as we update the baseline to a state, where IP is enabled in the gradle.properties

alllex:
As a first step, we should explicitly disable IP, when building distributions from a commit

PR #38402, opened 2026-07-02
Drop deprecated isolated-projects property names

### LINKED ISSUE #38400: Clean up remaining org.gradle.unsafe.isolated-projects property usages
### Summary

The Isolated Projects properties were renamed and the `org.gradle.unsafe.*` names are now deprecated aliases:

- `org.gradle.unsafe.isolated-projects` → `org.gradle.isolated-projects`
- `org.gradle.unsafe.isolated-projects.diagnostics` → `org.gradle.isolated-projects.diagnostics`
- `org.gradle.unsafe.isolated-projects.dangerously-ignore-problems` → `org.gradle.isolated-projects.dangerously-ignore-problems`

Several places in the repository still use the old `unsafe` names. This issue tracks cleaning them up to use the current names.

Spun off from https://github.com/gradle/gradle/pull/37759#discussion_r3511384341 ("Drop `unsafe`. There are other places like this").

### Candidates to update

- `.teamcity/src/main/kotlin/projects/CheckProject.kt`
- `.teamcity/src/main/kotlin/projects/GradleBuildToolRootProject.kt`
- `build-logic/performance-testing/src/main/kotlin/gradlebuild/performance/tasks/BuildCommitDistribution.kt`
- `gradle.properties` (repo root)
- `platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIntegrationTest.groovy`

### Needs per-file judgement

- `platforms/ide/tooling-api/src/crossVersionTest/.../r930|r940|r970/*.groovy` — cross-version tests may target older Gradle versions where only the `unsafe` name exists; verify before changing.
- `architecture/standards/0010-gradle-properties-naming.md` — uses the name as a naming-convention example; may be intentional.

### Must NOT change (they intentionally reference the deprecated name)

- `platforms/core-runtime/start-parameter/src/main/java/org/gradle/initialization/StartParameterBuildOptions.java` — defines `DEPRECATED_PROPERTY_NAME` for backwards-compatible handling.
- `platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc` — documents the deprecation/rename.
- `platforms/documentation/docs/src/docs/release/notes.md` — release notes for the change.

### Notes

- The two docs snippets flagged in the PR review (`isolatedProjects/lifecycleApplyFalse` and `isolatedProjects/lifecycleSettingsPlugin` `gradle.properties`) are being fixed directly in #37759 and are not part of this cleanup.
- `git grep "org.gradle.unsafe.isolated-projects"` is a good starting point (~15 files at time of filing).

reinsch82:
## Summary

Cleans up leftover usages of the deprecated `org.gradle.unsafe.isolated-projects` property name, using the current `org.gradle.isolated-projects` where it is safe to do so.

Addresses #38400. Spun off from the review discussion on #37759, and **targets that feature branch** (`reinhold/ip/36461-beforeproject-convention-plugin`) rather than `master`.

## Changed (outer build config, runs on the current Gradle)

- Root `gradle.properties` — also removes a deprecation warning we were emitting when building Gradle itself.
- `.teamcity` `GRADLE_OPTS` in `CheckProject.kt` and `GradleBuildToolRootProject.kt`.

## Deliberately left as `unsafe` (still required or intentional)

- **tooling-api cross-version tests** (`r930`/`r940`/`r970`) — run against older Gradle versions (`@TargetGradleVersion('>=9.3.0'/'>=9.7.0')`) that only understand the `unsafe` name.
- **`BuildCommitDistribution`** — builds/runs potentially old baseline commits for performance tests (older distributions don't recognise the new name); pre-existing `// TODO:isolated` (#36771).
- **`IsolatedProjectsIntegrationTest`** — these tests verify the deprecated names still work (incl. "can enable via deprecated property").
- **`StartParameterBuildOptions`** — defines the deprecated aliases.
- **Docs** — release notes, upgrade guide, and `isolated_projects.adoc` document the old names on purpose.

## Not touched (judgement call for reviewers)

- `architecture/standards/0010-gradle-properties-naming.md` uses the old name as a naming-convention example. Left as-is; happy to update if preferred.

## Test plan

- [ ] CI on this branch (the change is string-literal only; `.teamcity` config compile + Gradle build are covered by `sanityCheck`).

### DIFF
--- a/.teamcity/src/main/kotlin/projects/CheckProject.kt
+++ b/.teamcity/src/main/kotlin/projects/CheckProject.kt
@@ -27,7 +27,7 @@ class CheckProject(
 
         params {
             if (model.branch.isMaster || model.branch.isExperimental) {
-                param("env.GRADLE_OPTS", "-Dorg.gradle.unsafe.isolated-projects=%enableIsolatedProjects%")
+                param("env.GRADLE_OPTS", "-Dorg.gradle.isolated-projects=%enableIsolatedProjects%")
             }
             param("credentialsStorageType", "credentialsJSON")
             // Disallow Web UI changes to TeamCity settings
--- a/.teamcity/src/main/kotlin/projects/GradleBuildToolRootProject.kt
+++ b/.teamcity/src/main/kotlin/projects/GradleBuildToolRootProject.kt
@@ -32,6 +32,6 @@ class GradleBuildToolRootProject(
 
         params {
             param("enableIsolatedProjects", "false") // TODO: remove as soon as CI is compatible with enabling IP
-            param("env.GRADLE_OPTS", "-Dorg.gradle.unsafe.isolated-projects=false")
+            param("env.GRADLE_OPTS", "-Dorg.gradle.isolated-projects=false")
         }
     })
--- a/build-logic/performance-testing/src/main/kotlin/gradlebuild/performance/tasks/BuildCommitDistribution.kt
+++ b/build-logic/performance-testing/src/main/kotlin/gradlebuild/performance/tasks/BuildCommitDistribution.kt
@@ -195,7 +195,10 @@ abstract class BuildCommitDistribution @Inject internal constructor(
             "--no-daemon",
             if (gradleModeCompatibility == null) "--no-configuration-cache" else null,
             // TODO:isolated https://github.com/gradle/gradle/issues/36771
+            // Both names are needed: wrappers before Gradle 9.7 only understand the deprecated name,
+            // while 9.7+ wrappers prefer the current name when the checked-out commit enables IP with it.
             if (gradleModeCompatibility == "IP") null else "-Dorg.gradle.unsafe.isolated-projects=false",
+            if (gradleModeCompatibility == "IP") null else "-Dorg.gradle.isolated-projects=false",
             "--init-script",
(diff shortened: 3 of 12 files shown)


cobexer (approved):
Target branch?

cobexer (changes requested):
❌ Actually https://github.com/gradle/gradle/actions/runs/28586602674/job/84759674229?pr=38402#step:8:40 must be fixed first

Maybe there: https://github.com/actions/gradle-build-tools-actions/tree/main/dependency-submission but I don't know, I suppose @bigdaz can answer that?

cobexer:
https://github.com/gradle/gradle/blob/95803a7796b6815af62a2cf54367393162835351/build-logic/performance-testing/src/main/kotlin/gradlebuild/performance/tasks/BuildCommitDistribution.kt#L198 should be updated too, no version that does not understand the new name has the flag set?

Also, I wonder if that is all possible ways we could build that property name, did you do a thorough check?

And: could we update tests as well to pass the desired name based on the Gradle version? otherwise we'll have to do such a change again in the future while we could cleanup now.

cobexer (approved):

cobexer (approved):

PR #38378, opened 2026-06-30
Nag when rule source plugins are applied

big-guy:
Emit a deprecation warning when a rule-based / software model plugin is applied, pointing users to the Gradle 9 upgrade guide and noting these plugins will be removed in Gradle 10.

## Changes
- `RuleBasedPluginTarget.applyRules` now nags via `DeprecationLogger.deprecatePlugin(...)` with context "Rule-based/software model plugins are no longer supported."

Cherry-picked from `b872cf3aad5`.

PR #38404, opened 2026-07-02
Add reproducer for issue 36234

### LINKED ISSUE #36234: Importing a platform to testFixturesApi changes the results of dependency resolution when there is a fork in the tree
### Current Behavior

In a certain set of circumstances, when the java-test-fixtures plugin is used alongside the java-platform plugin, the dependency tree can be materially changed (i.e. dependencies are removed) by importing a platform to `testFixturesApi`

A minimal reproducer for this behaviour can be found in https://github.com/3ll3d00d/gradle-bom-fork-bug , see the readme for observed behaviour & review the dependencies in the project to see the precise circumstances in which the bug can surface or be avoided (specifically avoided when a fork in the dependency tree is not present and triggered when the fork is present and the platform is imported to testFixturesApi)

### Expected Behavior

The imported platform only specifies the version to use without disturbing the dependency tree itself

### Context (optional)

Use test-fixtures in conjunction with a platform in a complex dependency tree

https://github.com/3ll3d00d/gradle-bom-fork-bug

### Gradle version

9.2.1

### Build scan URL (optional)

https://scans.gradle.com/s/mxwsmbvry7wi4 - failure run (with platform import)
https://scans.gradle.com/s/bhmhb3uannk6i - successful run (without platform import)

_No response_

tresat:
The test exercises the scenario from the external
reproducer at https://github.com/3ll3d00d/gradle-bom-fork-bug — a fork in the dependency tree where the same module is reached via both a `testFixtures` capability path and a regular library path, combined with a java-platform imported to `testFixturesApi` on the consumer.

The test currently passes as both `apiElements` and
`testFixturesApiElements` variants of base are selected in the resolved `testCompileClasspath`.
The external reporter observed the defect against Gradle 9.2.1 - this test is committed as a regression guard.

Uses a publish-then-consume pattern: a producer sub-build applies real java-library / java-test-fixtures / java-platform / maven-publish plugins to publish faithful Gradle Module Metadata, then the consumer at the test root resolves against the resulting repository.

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/platforms/PlatformImportedToTestFixturesForkResolveIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/platforms/PlatformImportedToTestFixturesForkResolveIntegrationTest.groovy
@@ -1,0 +1,151 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.integtests.resolve.platforms
+
+import org.gradle.integtests.fixtures.AbstractIntegrationSpec
+import spock.lang.Issue
+
+/**
+ * Reproduces the defect from https://github.com/3ll3d00d/gradle-bom-fork-bug against the
+ * current Gradle build, using a "publish then consume" pattern: a producer sub-build applies
+ * the real {@code java-library}, {@code java-test-fixtures}, {@code java-platform}, and
+ * {@code maven-publish} plugins to publish faithful Gradle Module Metadata for five
+ * interdependent modules; a separate consumer build then resolves against that repository.
+ *
+ * Synthetic maven-fixture metadata (via {@code mavenRepo.module(...)}) does not reproduce the
+ * defect — the resolver appears to require the exact attribute set and variant shape that
+ * real Gradle publishing produces.
+ */
+@Issue("https://github.com/gradle/gradle/issues/36234")
+class PlatformImportedToTestFixturesForkResolveIntegrationTest extends AbstractIntegrationSpec {
+
+    def "importing a platform to testFixturesApi does not drop the testFixtures variant of a transitively required module when the dependency tree forks"() {
+        given:
+        // Producer sub-build — publishes the 5 modules with real Gradle plugins.
+        def producer = file("producer")
+        producer.file("settings.gradle") << """
+            rootProject.name = 'inner'
+            include('base', 'fork', 'middle', 'upper', 'bom')
+        """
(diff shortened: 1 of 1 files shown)


blindpirate:
@tresat This PR has been kicked out of the merge queue 3 times in the last 24h, and at least two of those are a **real, deterministic failure in this PR** — not flakiness. It has also blocked two unrelated PRs that happened to sit behind it in the queue.

### What fails

Isolated Projects bucket 8 (`Gradle_Master_Check_IsolatedProjects_21_bucket8`):

```
PlatformImportedToTestFixturesForkResolveIntegrationTest >
  importing a platform to testFixturesApi does not drop the testFixtures variant
  of a transitively required module when the dependency tree forks
```

```
* Where:
Build file '.../PlatformImp.Test/<dir>/producer/build.gradle' line: 3

* What went wrong:
A problem occurred evaluating root project 'inner'.
> Project ':' cannot access 'Project.group' functionality on subprojects
```

It failed **all 6 retry invocations** in every affected build, so it is reproducible, not a flake.

### Why

The producer fixture build script configures subprojects from the root:

```groovy
producer.file("build.gradle") << """
    subprojects {
        group = 'org.example'      // <- line 3, IP violation
        version = '1.0.0'
        apply plugin: 'maven-publish'
        afterEvaluate { ... }
    }
"""
```

`subprojects { ... }` is a cross-project configuration violation under Isolated Projects, and the IP integ-test bucket runs the executer with `-Dorg.gradle.isolated-projects=true` and `-Dorg.gradle.configuration-cache.max-problems=0`, so the producer build fails outright before the reproducer ever gets to the interesting part.

### Suggested fix

Either:

1. Move `group`, `version`, `apply plugin: 'maven-publish'` and the `publishing.repositories` block into each of the five subproject build scripts (they all already exist), dropping the `subprojects { }` / `afterEvaluate { }` block entirely — this keeps the reproducer valid and makes it IP-clean; or
2. Annotate the test with `@ToBeFixedForIsolatedProjects` if you'd rather not rework the fixture right now.

### Evidence

| Merge queue build | Queued PR | Result |
|---|---|---|
| [116397525](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_GitHubMergeQueueCheckPass/116397525) (08-18 20:08) | #38404 | IP bucket8 [116397433](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_IsolatedProjects_21_bucket8/116397433) — [scan](https://ge.gradle.org/s/kr67vqurwb2cg) |
| [116414265](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_GitHubMergeQueueCheckPass/116414265) (08-19 07:15) | #38404 | IP bucket8 [116414173](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_IsolatedProjects_21_bucket8/116414173) — [scan](https://ge.gradle.org/s/vi4g2tal26srg) |
| [116397858](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_GitHubMergeQueueCheckPass/116397858) (08-18 20:15) | #37924 | same test — collateral, this PR was ahead in the queue |
| [116411357](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_GitHubMergeQueueCheckPass/116411357) (08-19 05:49) | #38086 | same test — collateral |

(The third rejection, [116410561](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_GitHubMergeQueueCheckPass/116410561), was a set of native performance regressions and is *not* caused by this PR — that queue branch was built on base `2e9373c` = #38378, and every recorded occurrence of those scenarios so far sits on a branch containing #38378. I've flagged it separately.)

Since the head commit is unchanged, re-queueing as-is will fail the same way and keep taking other PRs down with it.

tresat:
@big-guy It looks like this needs another look before it can be merged.  It's not vital for 9.8, I'm delaying it to 9.9.

PR #38408, opened 2026-07-02
Deprecate starting processes at configuration time

### LINKED ISSUE #38399: Deprecate starting external processes at configuration time outside of ProviderFactory.{java}exec or ValueSource
Starting an external process at configuration time should emit a deprecation warning even when the configuration cache is **not** enabled.

With the configuration cache on, this is already reported as an error:

> Starting an external process '<command>' during configuration time is unsupported.

With the configuration cache off, the same code runs silently — the only trace is the lack of an optional end-of-build "promo" nudge suggesting the user try the configuration cache. There is no deprecation and nothing that pushes build authors to migrate.

As we move toward the configuration cache being on by default, behavior that is unsupported with CC on should be deprecated ahead of time, so that build and plugin authors get a clear, actionable warning during the grace period rather than discovering a hard failure only once CC becomes the default.

Given the impact and the timeline for Gradle 10, we should not aim to elevate this to an error there.

## Expected behavior

Calling these APIs at configuration time emits a deprecation warning when configuration cache is not enabled:
- Gradle's exec APIs: `ExecOperations.exec` / `ExecOperations.javaexec` and the underlying exec spec actions
- `java.lang.ProcessBuilder.start()` and `ProcessBuilder.startPipeline(...)`
- `java.lang.Runtime.exec(...)` (all overloads)
- Groovy process helpers: `String.execute(...)`, `String[].execute(...)`, `List.execute(...)` (`ProcessGroovyMethods`)

mlopatkin:
Starting an external process at configuration time is unsupported with the configuration cache and is already reported as an error when it is enabled. With the configuration cache off the process runs silently. As the configuration cache moves toward being enabled by default, emit a deprecation for this case so build and plugin authors get an actionable warning ahead of time. The behavior will become an error in Gradle 11.

The deprecation is emitted at the point where all external-process APIs converge, gated to the configuration-cache-off path so the existing error path is unchanged. It covers ExecOperations.exec/javaexec, ProcessBuilder.start/startPipeline, Runtime.exec, and the Groovy String/String[]/List.execute helpers. Processes started through ProviderFactory.exec/javaexec or a ValueSource remain supported and are unaffected.

Fixes #38399.

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
@@ -587,6 +587,10 @@ Because `obtain()` runs on every build, it is recommended to keep the implementa
 
 TIP: Plugin and build scripts should avoid running external processes at configuration time.
 
+Starting an external process at configuration time via `ExecOperations.exec` / `ExecOperations.javaexec`, `java.lang.ProcessBuilder`, `java.lang.Runtime.exec`, or the Groovy `String` / `String[]` / `List` `execute()` helpers is *deprecated* as of Gradle 9.7 and will become an error in Gradle 11.
+This deprecation only fires when the Configuration Cache is disabled; when it is enabled, the same behavior is already reported as a Configuration Cache problem.
+`providers.exec`, `providers.javaexec`, and `ValueSource` implementations are unaffected; those remain the supported way to run a process at configuration time.
+
 You should avoid using these APIs for running processes during configuration:
 
 - *Java/Kotlin*: `ProcessBuilder`, `Runtime.exec(...)`, etc...
--- a/platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
@@ -587,19 +587,17 @@ Because `obtain()` runs on every build, it is recommended to keep the implementa
 
 TIP: Plugin and build scripts should avoid running external processes at configuration time.
 
-Starting an external process at configuration time via `ExecOperations.exec` / `ExecOperations.javaexec`, `java.lang.ProcessBuilder`, `java.lang.Runtime.exec`, or the Groovy `String` / `String[]` / `List` `execute()` helpers is *deprecated* as of Gradle 9.7 and will become an error in Gradle 11.
-This deprecation only fires when the Configuration Cache is disabled; when it is enabled, the same behavior is already reported as a Configuration Cache problem.
-`providers.exec`, `providers.javaexec`, and `ValueSource` implementations are unaffected; those remain the supported way to run a process at configuration time.
-
-You should avoid using these APIs for running processes during configuration:
+You must not use these APIs directly for running processes during configuration:
 
 - *Java/Kotlin*: `ProcessBuilder`, `Runtime.exec(...)`, etc...
 - *Groovy*: `*.execute()`, etc...
 - *Gradle*: `ExecOperations.exec`, `ExecOperations.javaexec`, etc...
 
-The flexibility of these methods prevent Gradle from determining how the calls impact the build configuration, making it difficult to ensure that the Configuration Cache entry can be safely reused.
+The flexibility of these methods prevents Gradle from determining how the calls impact the build configuration, making it difficult to ensure that the Configuration Cache entry can be safely reused.
+Therefore, Gradle reports a Configuration Cache problem when it detects such an invocation.
+Even when the Configuration Cache is disabled, calling these methods at configuration time still emits a deprecation warning and will become an error in a future Gradle version.
 
-However, if running processes is required at configuration time, you can use the configuration-cache-compatible APIs detailed below.
+If running processes is required at configuration time, you can use the configuration-cache-compatible APIs detailed below.
 
 For simpler cases, when grabbing the output of the process is enough,
 link:{groovyDslPath}/org.gradle.api.provider.ProviderFactory.html#org.gradle.api.provider.ProviderFactory:exec(org.gradle.api.Action)[`providers.exec()`] and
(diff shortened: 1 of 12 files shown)


mlopatkin:
@bot-gradle test ACC

lkasso (approved):
Docs LGTM.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -557,6 +557,10 @@ Because `obtain()` runs on every build, it is recommended to keep the implementa
 
 TIP: Plugin and build scripts should avoid running external processes at configuration time.
 
+Starting an external process at configuration time via `ExecOperations.exec` / `ExecOperations.javaexec`, `java.lang.ProcessBuilder`, `java.lang.Runtime.exec`, or the Groovy `String` / `String[]` / `List` `execute()` helpers is *deprecated* as of Gradle 9.7 and will become an error in Gradle 11.
+This deprecation only fires when the Configuration Cache is disabled; when it is enabled, the same behavior is already reported as a Configuration Cache problem.
+`providers.exec`, `providers.javaexec`, and `ValueSource` implementations are unaffected; those remain the supported way to run a process at configuration time.
+
```
abstratt: ❓ 

One edge case seems to be *eager* evaluation of process-executing value sources. Those are not accepted by CC, but lead to deprecation when CC is off.

mlopatkin: > One edge case seems to be _eager_ evaluation of process-executing value sources. Those are not accepted by CC, but lead to deprecation when CC is off.

WDYM? Eager evaluation of process-running value sources (e.g. `process.exec {}.standardOutput.get()`) is allowed with CC, and, even with this PR, produces no deprecation in Vintage.

abstratt: Sorry, I meant that 

> Those ***are*** accepted by CC, but lead to deprecation when CC is off.

> produces no deprecation in Vintage.

But they do, that is what I meant.

According to our friend:

> With CC off, eagerly obtaining a ValueSource that starts a process via ProcessBuilder / Runtime.exec / Groovy execute() emits the deprecation — even though CC-on accepts the identical code, and even though ValueSource + ExecOperations stays silent in the same scenario.
> 
>  Root cause: those three APIs' exemption rides on input-tracking suppression during obtain(), which is driven by the CC fingerprint writer and therefore inert when CC is off. ExecOperations dodges it only because the value-source factory injects a no-listener variant. The requirements doc backs the intended contract explicitly (configuration_cache_requirements.adoc:497-499, 535, 541).

Test case:

```
package org.gradle.internal.cc.impl.inputs.process

import org.gradle.internal.cc.impl.AbstractConfigurationCacheIntegrationTest
import org.gradle.process.ShellScript
import spock.lang.Issue

@Issue("https://github.com/gradle/gradle/issues/38399")
class ValueSourceProcessAtConfigurationTimeIntegrationTest extends AbstractConfigurationCacheIntegrationTest {

    def "eagerly obtaining a ValueSource that starts a process with ProcessBuilder is not deprecated when the configuration cache is off"() {
        given:
        def script = ShellScript.builder().printText("Hello from script").writeTo(testDirectory, "script")
        def commandLine = ShellScript.cmdToVarargLiterals(script.commandLine)

        buildFile """
            import org.gradle.api.provider.ValueSource
            import org.gradle.api.provider.ValueSourceParameters

            abstract class RunProcess implements ValueSource<String, ValueSourceParameters.None> {
                @Override
                String obtain() {
                    String[] command = [$commandLine]
                    new ProcessBuilder(command).start().waitFor()
                    return "done"
                }
            }

            println("Result: " + providers.of(RunProcess) {}.get())

            tasks.register("noop") {}
        """

        expect:
        run(DISABLE_SYS_PROP, "noop")
    }
}

```

mlopatkin: Nice catch! We have tests that run ValueSource with Configuration Cache, and I got confused thinking that they working without deprecations mean something for this PR.

mlopatkin:
@bot-gradle test PRF, BD

platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_9.adoc
```diff
@@ -58,6 +58,34 @@ org.gradle.configuration-cache.unsafe.skip-task-logging-listeners-serialization=
...
+- `java.lang.ProcessBuilder.start()` and `ProcessBuilder.startPipeline(...)`.
+- `java.lang.Runtime.exec(...)` (all overloads).
+- The Groovy process helpers `String.execute(...)`, `String[].execute(...)`, and `List.execute(...)`.
+
+This does *not* affect uses of the link:{javadocPath}/org/gradle/api/provider/ProviderFactory.html#exec(org.gradle.api.Action)[`ProviderFactory.exec`] / link:{javadocPath}/org/gradle/api/provider/ProviderFactory.html#javaexec(org.gradle.api.Action)[`ProviderFactory.javaexec`] providers (`providers.exec` / `providers.javaexec`) or a <<configuration_cache_requirements.adoc#config_cache:requirements:external_processes, `ValueSource`>> that runs a process inside — those remain the supported way to run a process at configuration time.
+
+Starting an external process at configuration time in any of the deprecated ways is already an error when the Configuration Cache is enabled.
+To simplify preparation for the Configuration Cache being enabled by default in Gradle 10, and to open the way to it becoming the only mode of execution, Gradle now unconditionally deprecates behaviors that are unsupported with the Configuration Cache.
```
abstratt, line 115: 💭 

FYI: Our friend didn't like "unconditionally" when this is conditional to CC being disabled

I won't insist.

abstratt (approved):
LGTM, one minor nit

blindpirate:
@bot-gradle test qf

mlopatkin:
@bot-gradle test PRF without PTS

mlopatkin:
@bot-gradle test this

PR #38409, opened 2026-07-02
Deprecate invalid Attribute value types

### LINKED ISSUE #27816: Please document what the advantage is of having interfaces that implement `Named` for attributes
### Issue type

Missing information

### Problem description

What is the advantage of having interfaces that implement `Named` for attributes like `public interface Usage extends Named`?
I mean opposed to just having the attributes with type `String`.
In the module metadata they are just strings.
To set them you use the `String` constants in the interface.
You can also use any arbitrary other string.

So the only effect I can realize right now is, that it is more cumbersome to set the attributes by needing to use `objects.named` to create  the value instead of just setting a `String` as value.

### Context (optional)

_No response_

### Page with the problem

https://docs.gradle.org/current/userguide/variant_attributes.html#sec:declaring_attributes

### LINKED ISSUE #38242: Enforce that non basic dependency attributes implement Named
### Current Behavior

You can declare an attribute from an `Enum` that does not implement `Named`

### Expected Behavior

Declaring an attribute that does not implement `Named` or is a `String`, `Boolean` or `Integer` should give you a clear error immediately.

### Context (optional)

Discovered during the 9.6.0 RC phase that refactored code in making this an explicit error during resolution.

Related:
* #27816 - where we should really discourage using something else than primitive types.

N/A

### Gradle version

9.5.1 and before

### Build scan URL (optional)

_No response_

_No response_

tresat:
Plain `Enum` types, array types, and other types that are not `Boolean`, `String`, `Number` or `Named` are specifically **not** allowed as attribute values, even though they would work in some limited situations.  They can cause problems when published.

Any other types will now emit a deprecation warning.

Also adds tests to cover non-`Integer` numeric types, and properly implements serialization of them.  All allowed types are now tested.

This aligns the actual code with what is advertised as valid `Attribute` value types in the docs, and makes those docs exhaustively and precisely list all the allowed types.  

A custom deprecation warning is also added to cover older versions of KGP, which use raw `Enum` values (more recent KGP versions do not).

### DIFF
--- a/platforms/core-configuration/kotlin-dsl-integ-tests/src/integTest/kotlin/org/gradle/kotlin/dsl/plugins/KotlinPluginAppliedWithOlderGradleVersionsIntegrationTest.kt
+++ b/platforms/core-configuration/kotlin-dsl-integ-tests/src/integTest/kotlin/org/gradle/kotlin/dsl/plugins/KotlinPluginAppliedWithOlderGradleVersionsIntegrationTest.kt
@@ -162,20 +162,22 @@ class KotlinPluginAppliedWithOlderGradleVersionsIntegrationTest(
             .withStackTraceChecksDisabled() // The Kotlin compiler daemon intermittently crashes and logs a stack trace before falling back; that's not what we test here.
             .run()
 
-        // KGP 2.0.20 and 2.0.21 register an attribute whose type is a plain enum
-        // (KotlinNativeBundleArtifactFormat.KotlinNativeBundleArtifactsTypes). Gradle allows
-        // this via a targeted compatibility exception in Attribute.of and emits a deprecation
-        // warning. Verify the warning is present for those specific KGP versions and absent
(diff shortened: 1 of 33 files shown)


tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

lkasso (approved):
Docs LGTM. I added an entry in the upgrade guide.

tresat:
> Docs LGTM. I added an entry in the upgrade guide.

@lkasso - I may have force-pushed over your update without noticing, sorry.  Can you please push it again?

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

tresat:
@bot-gradle test RFN

tresat:
@bot-gradle test RFN

tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

big-guy (approved):

PR #38414, opened 2026-07-03
Upgradle 9.7.0-milestone-3

alllex:
### Context

Bump the Gradle wrapper distribution from `9.7.0-milestone-1` to `9.7.0-milestone-2` to dogfood the latest milestone.

Validated locally:
- `./gradlew help --dry-run` ✅
- `./gradlew sanityCheck` ✅

### Contributor Checklist

### DIFF
--- a/build-logic-commons/basics/build.gradle.kts
+++ b/build-logic-commons/basics/build.gradle.kts
@@ -33,12 +33,7 @@ dependencies {
 }
 
 configurations.testRuntimeClasspath {
-    // The kotlin-gradle-plugin jar bundles a partial, gradle-patched subset of the shaded
-    // `org.jetbrains.kotlin.com.intellij.*` classes. Its MockProject implements
-    // ComponentManagerEx, an interface that jar does NOT bundle, so when it shadows the complete
-    // classes from kotlin-compiler-embeddable, constructing KotlinCoreProjectEnvironment fails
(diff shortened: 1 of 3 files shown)


jbartok:
Root cause of the failures experienced on wrapper bump is a change in classpath ordering, which should be fixed separately. In the meantime this exclusion is good.

jbartok (approved):

alllex:
⚠️ **Blocked: the docs build fails on 9.7.0-M2.**

`:docs:userguideMultiPage`, `:docs:userguideSinglePageHtml`, and `:docs:samplesMultiPage` all fail with:

> The value for property 'workingDirectory' cannot be changed any further.

The Asciidoctor plugin runs out-of-process and copies its fork options onto the worker spec (`copyTo`), which re-sets `workingDirectory`. In 9.7.0-M2, `DefaultWorkerExecutor.processIsolation` now calls `disallowChanges()` on that property *before* the plugin action runs, so the copy throws — a regression introduced this milestone (works on M1). Details + minimal reproducer in #38455.

This upgrade can't land until #38455 is fixed (or we work around it in our docs build logic).

Example failing scan: https://ge.gradle.org/s/wtzbtyfqlk32w

ljacomet (approved):
Updated to milestone 3

PR #38419, opened 2026-07-03
Document applying third-party plugins from lifecycle callbacks

reinsch82:
## Summary

Documents applying **third-party (repository) plugins** from a `gradle.lifecycle.beforeProject`/`afterProject` callback under Isolated Projects, and migrating a third-party plugin away from an `allprojects {}` block.

Rather than adding a parallel section, this generalizes the existing "Applying convention plugins from an included build" section into **"Applying plugins from a lifecycle callback"**, since the limitation and fix are the same for included-build convention plugins and repository plugins.

## Changes (`isolated_projects.adoc`)

- Generalize the limitation: a lifecycle callback can only apply a plugin already on the project's classpath; it doesn't resolve plugins. This affects both included-build convention plugins **and** repository plugins.
- Fold the third-party case into the shared `apply false` workaround, adding the version-bearing snippet (`id("com.example.foo") version "1.2.3" apply false`) and the `pluginManagement { plugins { … } }` alternative.
- Add a "Migrating from `allprojects {}`" note: the declaration must move from the root build script into `settings.gradle(.kts)` so the settings-registered callback can see it.
- The `sec:lifecycle_callbacks_with_included_plugin_builds` anchor is preserved (still referenced by the diagnostic hint and an existing NOTE).

## Test plan

- [x] `./gradlew :docs:checkDeadInternalLinks` — passes (docs parse, cross-references resolve).
- Examples are illustrative `[source]` blocks (a real repository plugin can't be exercised offline in a docs sample); the included-build workaround still uses the existing executed samples.

### DIFF
--- a/platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
+++ b/platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
@@ -711,17 +711,18 @@ Applying a convention plugin from a build registered via `pluginManagement.inclu
 ====
 
 [[sec:lifecycle_callbacks_with_included_plugin_builds]]
-===== Applying convention plugins from an included build
+===== Applying plugins from a lifecycle callback
 
-Plugins from a build registered via `pluginManagement.includeBuild(...)` are resolved lazily so they are built only when explicitly requested in a `plugins {}` block.
-When Gradle sees `plugins { id("my.convention") }`, it builds `my.convention` from the included build and exports the resulting classpath into the surrounding script's classloader scope.
+A `gradle.lifecycle.beforeProject` (or `afterProject`) callback can only apply a plugin that is already on the project's plugin classpath.
+The callback merely registers an action; it does not resolve plugins itself.
+So `apply(plugin = "...")` from a callback fails with `Plugin with id '...' not found` for any plugin that is resolved lazily — that is, only built or downloaded when a `plugins {}` block requests it — and that has not already been resolved onto a classloader scope the project inherits.
+This affects two common cases:
 
-There is a limitation however:
+- a convention plugin from a build registered via `pluginManagement.includeBuild(...)`, and
+- a third-party plugin from a repository, such as the Gradle Plugin Portal or a Maven repository.
 
-- Using a `plugins {}` block in `settings.gradle(.kts)` or any project script *works*, because the block itself triggers the build of the plugin.
-- Using the plugin from inside a `gradle.lifecycle.beforeProject` (or `afterProject`) callback registered in the settings script *does not work*.
-The callback only registers an action; it doesn't request anything.
-By the time the callback runs at project configuration, nothing has asked Gradle to build the plugin, so `apply(plugin = "...")` fails with `Plugin with id '...' not found`.
+Using such a plugin from a `plugins {}` block in `settings.gradle(.kts)` or any project script *works*, because the block itself triggers resolution.
+Using it from a lifecycle callback *does not*, because nothing has triggered resolution by the time the callback runs at project configuration.
 
 This limitation does not apply to `buildSrc/`: it is built eagerly and its classpath is exported to every project, so a `gradle.lifecycle.beforeProject` callback can apply a `buildSrc` convention plugin directly, with none of the workarounds below.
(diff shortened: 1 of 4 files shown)


platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -753,17 +754,53 @@ Apply the settings convention plugin from `settings.gradle(.kts)`:
 include::sample[dir="snippets/isolatedProjects/lifecycleSettingsPlugin/kotlin",files="settings.gradle.kts[tags=settings-file]"]
 include::sample[dir="snippets/isolatedProjects/lifecycleSettingsPlugin/groovy",files="settings.gradle[tags=settings-file]"]
 
-*Alternative: pre-resolve via the settings `plugins {}` block*
+*Pre-resolve in the settings `plugins {}` block*
```
alllex: ❔ Btw, any reason why this one and sibling sections are not an h5 heading?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -711,17 +711,18 @@ Applying a convention plugin from a build registered via `pluginManagement.inclu
 ====
 
 [[sec:lifecycle_callbacks_with_included_plugin_builds]]
-===== Applying convention plugins from an included build
+===== Applying plugins from a lifecycle callback
```
alllex: 💅 We should probably rename the anchor to stay consistent. Given we are just incubating IP, I think it's okay to not care about potentially broken links just yet.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -753,17 +754,53 @@ Apply the settings convention plugin from `settings.gradle(.kts)`:
...
+[source,kotlin]
+.build.gradle.kts
+----
+plugins {
+    id("com.example.foo") version "1.2.3" apply false
+}
+
+allprojects {
+    apply(plugin = "com.example.foo")
+}
```
alllex, line 818: 💅 Given this is on the IP page, we should be explicit that this is not a compatible pattern. The usual reading flow is to look at snippets first, because they stand out, so that's were we should say that this is just showing an incompatible pattern. I'd add an inline comment before `allprojects`

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -753,17 +754,53 @@ Apply the settings convention plugin from `settings.gradle(.kts)`:
...
+.settings.gradle.kts
+----
+plugins {
+    id("com.example.foo") version "1.2.3" apply false
+}
+
+gradle.lifecycle.beforeProject {
+    apply(plugin = "com.example.foo")
+}
+----
```
alllex, line 783: ❓ Should we make this a tested snippet? We should be able to use a real plugin, I think. We do that in many places in the docs already.

I don't recall exactly, but we might have a way to make this a tested snippet without creating an entire sample build. Though, I might be confusing it with some other context.

alllex (approved):
LGTM with some suggestions

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
lkasso: I'm ok with this not being tested, but I did add a Groovy counterpart.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
lkasso: I updated this anchor and all references to it.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
lkasso: Updated.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
lkasso: Added a comment in the code snippet, hopefully that's good enough.

lkasso (approved):
Docs LGTM. Made a few improvements and addressed @alllex comments.

PR #38424, opened 2026-07-06
Declare input annotations for MavenPom

jvandort:
This allows us to use it as a real task input and avoid Transient.Var and Cached
like we were using before. Now, CC can see the providers that make up the maven
pom and if they have task dependencies, can properly serialize the providers
without eagerly realizing tasks.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/maven/src/main/java/org/gradle/api/publish/maven/MavenPomContributor.java
+++ b/platforms/software/maven/src/main/java/org/gradle/api/publish/maven/MavenPomContributor.java
@@ -19,6 +19,8 @@ package org.gradle.api.publish.maven;
 import org.gradle.api.provider.MapProperty;
 import org.gradle.api.provider.Property;
 import org.gradle.api.provider.SetProperty;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Optional;
 
 /**
  * A contributor of a Maven publication.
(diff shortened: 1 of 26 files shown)


jvandort:
@bot-gradle test this

lkasso (approved):
Docs LGTM

platforms/software/maven/src/main/java/org/gradle/api/publish/maven/tasks/GenerateMavenPom.java
```diff
@@ -26,33 +26,30 @@
...
-    private final Transient.Var<MavenPom> pom = varOf();
-    private final Cached<MavenPomFileGenerator.MavenPomSpec> mavenPomSpec = Cached.of(() ->
-        MavenPomFileGenerator.generateSpec((MavenPomInternal) getPom())
-    );
+    private MavenPom pom;
+
+    public GenerateMavenPom() {
+        this.doNotTrackStateIf(
+            "withXml actions cannot be snapshotted",
+            task -> !((MavenPomInternal) ((GenerateMavenPom) task).getPom()).getXmlAction().isEmpty()
```
ghale, line 52: 💅 I think we should have an integration test validating this to ensure this works as expected and (more importantly) to guard against regressions.  Getting a false positive up-to-date here on an actually changing pom would be ugly.

platforms/software/maven/src/main/java/org/gradle/api/publish/maven/MavenPomContributor.java
```diff
@@ -32,41 +34,56 @@ public interface MavenPomContributor {
...
      */
+    @Input
+    @Optional
     Property<String> getOrganizationUrl();
 
     /**
      * The roles of this contributor.
      */
+    @Input
+    @Optional
```
ghale: 💅 Since this is `SetProperty`, it always returns a value, right? (i.e. no value == empty Set) I don't think this `@Optional` is necessary here.

ghale (dismissed):
This looks good.  I think we should add the test coverage I suggested before merging, though.

platforms/software/maven/src/main/java/org/gradle/api/publish/maven/tasks/GenerateMavenPom.java
jvandort: Added a test

platforms/software/maven/src/main/java/org/gradle/api/publish/maven/MavenPomContributor.java
jvandort: Good catch

ghale (approved):

cobexer:
PTAL https://ge.gradle.org/s/f42mzdweyzgvc/console-log?page=19#L18991

```
* What went wrong:	
Execution failed for task ':maven:validatePlugins' (registered by plugin 'gradlebuild.task-properties-validation').	
 Plugin validation failed with 1 problem	
  Not cacheable without reason	
    Type 'org.gradle.api.publish.maven.tasks.GenerateMavenPom' must be annotated either with @CacheableTask or with @DisableCachingByDefault	
      The task author should make clear why a task is not cacheable	
      For more information, please refer to https://docs.gradle.org/9.8.0-gradleception-4d385ce11c51fd6d5c4074c8912e7618-19800202020202+0000/userguide/validation_problems.html#disable_caching_by_default in the Gradle documentation.	
      Possible solutions:	
        1. Add @DisableCachingByDefault(because = ...).	
        2. Add @CacheableTask.	
        3. Add @UntrackedTask(because = ...).
```

jvandort:
Strange how this error only shows up in gradleception

PR #37928, opened 2026-05-19
Make ResolutionResult API thread safe

jvandort:
When originally refactored, we expected this API not be accessed concurrently. Unfortunatley in some circumstances -- the github-dependency-graph-gradle-plugin -- reads to instances of a ResolvedComponentResult may execute in parallel, as the plugin uses the internal build operation API backed by DefaultBuildEventsListenerRegistry, which executes code in parallel without acquiring project locks.

We update the ResolutionResult API to be thread safe throughout to avoid data races when lazily calculating values.

Fixes https://github.com/gradle/gradle/issues/37927

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #38428, opened 2026-07-06
Fix deadlock when reading the resolution result concurrently

FinlayRJW:
### Context

Reading a `ResolutionResult` concurrently can deadlock the build. Making the API thread-safe in #37928 added `synchronized` to the lazily-computed caches, but the two read paths take the component and graph monitors in opposite orders: `getVariants()` locks the component then the graph, while `getDependents()` locks the graph then the component. Consumers that read the result from build-operation listeners run without the project lock and can race these two paths, deadlocking the build.

Each lazily-cached value is now computed outside the monitor via `Lazy.atomic()` and published once, so neither object holds its own lock while calling into the other. `DefaultResolvedComponentResult` and `ResolvedGraphResult` are made `final` so their `this`-capturing suppliers don't trip the `this-escape` lint.

### Contributor Checklist

- [ ] Provide integration tests (under \`<subproject>/src/integTest\`) to verify changes from a user perspective.
- [x] Provide unit tests (under \`<subproject>/src/test\`) to verify logic.
- [x] Ensure that tests pass sanity check: \`./gradlew sanityCheck\`.
- [x] Ensure that tests pass locally: \`./gradlew <changed-subproject>:quickTest\`.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/ivyservice/resolveengine/result/ResolutionResultConcurrencyTest.groovy
+++ b/platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/ivyservice/resolveengine/result/ResolutionResultConcurrencyTest.groovy
@@ -1,0 +1,222 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.api.internal.artifacts.ivyservice.resolveengine.result
+
+import org.gradle.api.artifacts.result.ComponentSelectionReason
+import org.gradle.api.internal.artifacts.DefaultImmutableModuleIdentifierFactory
+import org.gradle.api.internal.artifacts.DefaultModuleIdentifier
+import org.gradle.api.internal.artifacts.DefaultModuleVersionIdentifier
+import org.gradle.api.internal.artifacts.DependencyManagementTestUtil
+import org.gradle.api.internal.artifacts.capability.CapabilitySelectorSerializer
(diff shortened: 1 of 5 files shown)


platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/ivyservice/resolveengine/result/ResolutionResultConcurrencyTest.groovy
```diff
@@ -0,0 +1,91 @@
...
+package org.gradle.api.internal.artifacts.ivyservice.resolveengine.result
+
+import org.gradle.api.internal.artifacts.result.ResolvedGraphResult
+
+import java.lang.management.ManagementFactory
+import java.util.concurrent.CountDownLatch
+
+class ResolutionResultConcurrencyTest extends AbstractResolutionResultBuilderTest {
+
+    def "reading a component's variants and dependents concurrently does not deadlock"() {
```
FinlayRJW: On testing: I wanted something deterministic rather than a stress test, `ResolutionResultConcurrencyTest` forces the opposing lock order with latches and detects the cycle via `ThreadMXBean.findDeadlockedThreads()`, so it fails on the pre-fix code and passes with the fix (no timeouts). Not sure that's the most idiomatic approach here, happy to change it if there's a better pattern.

Also validated end-to-end on a real internal project that was hitting this: it deadlocks on `9.6.0` and completes with the fix.

ghale:

jvandort:
Thank you for reporting this issue and submitting a reproducer and proposed fix.

I've opened a separate [PR](https://github.com/gradle/gradle/pull/38446) with a revised set of changes for the following reasons:

1. Allocating `Lazy` wrappers inside `DefaultResolvedComponentResult` adds additional memory overhead. Recent refactorings to this code were specifically made to avoid additional memory overhead in this area. The separate PR avoids these additional allocations.
2. The Lazy.atomic around `computeResolvedEdgesByTarget` may potentially cause the relatively expensive reverse-dependents mapping to be calculated multiple times in concurrent scenarios. Keeping this under a lock avoids that.
3. Refactored tests avoids `StreamingResolutionResultBuilder` entirely, opting to directly construct a `GraphStructure` instance. 
4. We additionally resolve another similar deadlock with `getAllDependencies`. 

A separate PR avoids a verbose commit history which I like to avoid, given the changes diverged enough from your original PR. However, I have included you as a `Co-authored-by` entry in the commit there. 

PTAL at the linked PR. We will include this fix in the 9.7.0 release.

PR #38432, opened 2026-07-06
Don't use CC if it's relocated

### LINKED ISSUE #36392: Configuration cache entry does not get discarded when cache itself was moved
### Current Behavior

If I copy my project folder with existing configuration cache entries in `.gradle/configuration-cache`, the entries are reused in the new location. Oftentimes though, the entries contain absolute paths as part of the build configuration and the the build then points at files in the old location. So, for example, the UP-TO-DATE check will look at source files in the old location rather the new.

### Expected Behavior

A cache entry is discarded if the entry itself has a different absolute path than it had when it was created.

### Context (optional)

_No response_

In an empty folder `orig`:

```shell
# setup example project
gradle init --type java-library --use-defaults

# build example
gradle assemble

# copy example with caches
cd ..
cp -rf orig copy
cd copy

# modify a source file in copy
echo broken > lib/src/main/java/org/example/Library.java

# build example copy
gradle assemble

```

Despite the now broken sources you get:

```
BUILD SUCCESSFUL in 397ms
2 actionable tasks: 2 up-to-date
Configuration cache entry reused.
```

### Gradle version

9.3.0

### Build scan URL (optional)

_No response_

_No response_

ov7a:
fixes:
* https://github.com/gradle/gradle/issues/36392

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheDirIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheDirIntegrationTest.groovy
@@ -77,4 +77,71 @@ class ConfigurationCacheDirIntegrationTest extends AbstractConfigurationCacheInt
         then:
         configurationCache.assertStateLoaded()
     }
+
+    def "entry is not reused when the build directory is copied to a new location"() {
+        given:
+        def configurationCache = newConfigurationCacheFixture()
+        buildFile 'original/settings.gradle', """rootProject.name = 'relocated-build'"""
+        buildFile 'original/build.gradle', """task ok"""
+
+        when:
+        inDirectory 'original'
+        configurationCacheRun 'ok'
+
+        then:
+        configurationCache.assertStateStored()
+
+        when: 'running again in the original location'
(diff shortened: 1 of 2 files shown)


ov7a:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheDirIntegrationTest.groovy
```diff
@@ -77,4 +77,71 @@ class ConfigurationCacheDirIntegrationTest extends AbstractConfigurationCacheInt
         then:
         configurationCache.assertStateLoaded()
     }
+
+    def "entry is not reused when the build directory is copied to a new location"() {
+        given:
+        def configurationCache = newConfigurationCacheFixture()
+        buildFile'original/settings.gradle', """rootProject.name = 'relocated-build'"""
```
6hundreds: 💅 May we have a space before the first argument?

6hundreds (approved):
LGTM!

PR #38433, opened 2026-07-06
Deprecate Signature as PublishArtifact

jvandort:
Deprecates a number of signing APIs:

- All APIs on Signature that inherited from `PublishArtifact`
- APIs on `SignOperation` that return `Signature`
- Public constructors on `Signature`
- APIs on `SigningExtension` that deal with adding/removing artifacts to configurations
- Protected methods which should have been private

We could not come up with a reasonable use case for adding signatures as artifacts to a configuration. If this is desirable, it can be done manually with existing APIs to add files as artifacts to a consumable configuration.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

lkasso (approved):
Docs LGTM.

big-guy (approved):

PR #38445, opened 2026-07-08
Make GradleException implement ResolutionProvider

tresat:
This change reduces boilerplate by removing redundant `ResolutionProvider` implementations from numerous subclasses, allowing all `GradleException` instances and extensions to inherently carry resolution suggestions. It improves consistency and removes some code.  Some exception types which carried no extra data and we're tested for (which seemed to exist only to implement `ResolutionProvider`) are also deleted in favor of just using `GradleException`.


tresat:
@bot-gradle test this

tresat:
@bot-gradle test this

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/internal/exceptions/DefaultMultiCauseException.java
```diff
@@ -82,17 +91,6 @@ private ThreadLocal<Boolean> threadLocal() {
         return new HideStacktrace();
     }
 
-    @Override
-    public List<String> getResolutions() {
-        List<String> resolutions = new ArrayList<String>(causes.size()); // Typical case is 0 or 1 resolutions/cause
-        for (Throwable cause : causes) {
-            if (cause instanceof ResolutionProvider) {
-                resolutions.addAll(((ResolutionProvider) cause).getResolutions());
```
jbartok, line 90: This might be worth discussing. Until now you had a list of causes, now you have a flat list of resolutions, without a way of mapping resolution to problem it resolves. Feels wrong.

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/internal/exceptions/DefaultMultiCauseException.java
```diff
@@ -109,13 +107,30 @@ public List<? extends Throwable> getCauses() {
     public synchronized Throwable initCause(Throwable throwable) {
         causes.clear();
         causes.add(throwable);
+        addResolutionsFrom(throwable);
         return this;
     }
 
     public void initCauses(Iterable<? extends Throwable> causes) {
         this.causes.clear();
+        addCauses(causes);
```
jbartok, line 121: ❌ I think there should be a `clearResolutions()` here. I see it was removed in an intermediary commit, but that seems incorrect. Resolutions aren't lazily computed from causes any more, so not clearing their store will lead to duplication.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlReportGenerationException.java
```diff
@@ -18,26 +18,13 @@
...
  */
 @Contextual
-public class GenericHtmlReportGenerationException extends GradleException implements ResolutionProvider {
+public class GenericHtmlReportGenerationException extends GradleException {
     public GenericHtmlReportGenerationException(String message, Throwable cause) {
         super(message, cause);
     }
-
-    @Override
-    public List<String> getResolutions() {
```
jbartok, line 36: ❌ Just deleting this is a change in behavior; it should have been done as in `TestSuiteExecutionException`.

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/api/GradleException.java
```diff
@@ -16,21 +16,55 @@
...
+        super(message);
+        this.resolutions = new ArrayList<>(resolutions);
+    }
+
+    public GradleException(String message, @Nullable Throwable cause, List<String> resolutions) {
+        super(message, cause);
+        this.resolutions = new ArrayList<>(resolutions);
+    }
+
+    public final void addResolution(String resolution) {
```
jbartok, line 59: ❌ Public `addResolution()` and `clearResolution()` methods open the way for anybody to mutate/abuse the resolutions. Should be made at least package-private.

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/api/GradleException.java
```diff
@@ -16,21 +16,55 @@
...
+    }
+
+    public final void clearResolutions() {
+        resolutions.clear();
+    }
+
+    @NonNull
+    @Override
+    public final List<String> getResolutions() {
+        return Collections.unmodifiableList(resolutions);
```
jbartok: 🤔 Unmodifiable list, but the resolution objects themselves are still live and mutable. Not sure how to improve in an efficient way, though.

platforms/jvm/toolchains-jvm-shared/src/main/java/org/gradle/jvm/toolchain/internal/install/exceptions/ToolchainDownloadException.java
```diff
@@ -18,35 +18,23 @@
...
-
-import static java.util.Collections.emptyList;
 
 @Contextual
-public class ToolchainDownloadException extends GradleException implements ResolutionProvider {
-
-    private final List<String> resolutions;
+public class ToolchainDownloadException extends GradleException {
 
+    @SuppressWarnings("this-escape")
```
jbartok: `@SuppressWarnings("this-escape")` not needed here, only in constuctors where `addResolution(...)` is called in their body. Multiple instances, I'm only commenting on this one.

platforms/software/build-init/src/main/java/org/gradle/buildinit/plugins/internal/BuildInitException.java
```diff
@@ -16,30 +16,20 @@
...
 
     public BuildInitException(String message, Iterable<String> resolutions) {
-        super(message);
-        this.resolutions = ImmutableList.copyOf(resolutions);
-    }
-
-    @Override
-    public List<String> getResolutions() {
-        return resolutions;
+        super(message, Lists.newArrayList(resolutions));
```
jbartok: Redundant coppy, is copied again in `GradleException`.

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/internal/exceptions/DefaultMultiCauseException.java
jbartok: Please add test coverage for this.

platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/GenericHtmlReportGenerationException.java
jbartok: Please add test coverage for this.

jbartok (changes requested):

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/internal/exceptions/DefaultMultiCauseException.java
tresat: We used to have _the potential_ for mapping resolutions -> causes, but we never actually made use of it.  We always just flattened them into a single list.

I don't think there is a problem with this.  We only show a single list of resolutions at the end of a failed build.  So we'll need to flatten all the resolution data eventually regardless.  This PR just flattens the list at "add time" rather than waiting for "get time".

A multi-cause exception could _in theory_ create an abritrary "tree" of exceptions (where its own causes are also multi-cause exceptions) but _in practice_ seems to only ever have a single layer of causes - so I don't think this is really a big deal.

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/internal/exceptions/DefaultMultiCauseException.java
tresat: Will do.

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/api/GradleException.java
tresat: This was intentional.  I can imagine something farther down in the call stack catching a `GradleException` and adding an additional resolution to it - or, knowing better, completely replace any existing resolutions.

I don't see a reason to lock the resolutions collection.  Some of the (now deleted) `ResolutionProvider` implementations already provided for this with mutable resolutions lists.

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/api/GradleException.java
tresat: Discussed in thread above.

platforms/software/build-init/src/main/java/org/gradle/buildinit/plugins/internal/BuildInitException.java
tresat: This was actually needed to transform the `Iterable` to a `List` for the `GradleException` construction.  But, no reason `GradleException` can't just accept an `Iterable` directly.

platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/api/GradleException.java
abstratt: Can't we require folks to build a new `GradleException` in that case?

tresat: That's an option.  But I think it's more correct not to say a _new_ exceptional situation has been encountered.  We just know something else that might be helpful to add about the current failure - so we should append it to the existing resolutions already present.

big-guy:
Since Tom's working on something else right now, moving this to 9.9

PR #38446, opened 2026-07-08
Fix deadlock when reading the resolution result concurrently

jvandort:
DefaultResolvedComponentResult and ResolvedGraphResult took their monitors in opposite orders: computing a component's variants or dependencies held the component monitor while calling the graph's synchronized getVariant()/getComponent(), while computing incoming edges (getDependents()) held the graph monitor while calling back into every component's synchronized getAllDependencies(). Two threads reading a cold result concurrently — expected since build operation listeners read resolution results without holding the project lock — could deadlock the build.

Instead, compute the component-side values without holding any lock and publish them under a monitor that guards only the field access, with racing threads adopting the published value. This is sufficient because the graph monitor is now the only lock held during any computation, and nothing reachable from it acquires another lock: the component publish sections are leaf critical sections. It also keeps the whole-graph incoming-edge computation at-most-once (it still runs under the graph monitor) and preserves canonical DependencyResult identity across the getDependencies()/getDependents() views.

The new test pins the adversarial schedule by pausing the incoming- edge computation from inside its own GraphStructure reads, rather than by acquiring implementation locks, so it remains valid whatever locking mechanism the implementation uses. Both scenarios deadlock on the previous code and pass with this change.

Supersedes https://github.com/gradle/gradle/pull/38428

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/result/DefaultResolvedComponentResult.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/result/DefaultResolvedComponentResult.java
@@ -42,8 +42,8 @@ public class DefaultResolvedComponentResult implements ResolvedComponentResultIn
     private final IntList nodeIndices;
     private final ResolvedGraphResult graph;
 
-    private @Nullable ImmutableList<ResolvedVariantResult> variants;
-    private @Nullable ComponentDependencies dependencies;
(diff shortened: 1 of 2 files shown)


jvandort:
@bot-gradle test rfr

PR #38450, opened 2026-07-08
Fix TaskInputs.files absent provider validation

### LINKED ISSUE #38410: `TaskInputs.files()` silently accepts an absent `Provider` when it is wrapped in a collection, but fails when it is passed directly
### Current Behavior

Registering an absent (no-value) `Provider` as a required file input via `TaskInputs.files(...)` behaves differently depending on whether the provider is passed directly or wrapped in a collection:

- `inputs.files(absentProvider)` — **fails** validation:
  > Property 'p' doesn't have a configured value.
  > This property isn't marked as optional and no value has been configured.
- `inputs.files([absentProvider])` — the *same* absent provider wrapped in a `List` — **succeeds**: the provider is silently dropped and the input resolves to an empty file collection.

### Expected Behavior

An absent provider should be handled consistently regardless of whether it is passed directly or nested in a collection. Either both forms fail (a required file input has no value) or both are treated leniently. The two shapes should not diverge.

### Context (optional)

Task validation treats "raw" providers specially, but doesn't unwrap collections or FileCollections.

This is specific to `inputs.files(...)`. Verified sibling cases behave differently (all fail in both forms):

| Registration | direct absent provider | wrapped in a `List` |
| --- | --- | --- |
| `inputs.files` | fails — *value not set* | **succeeds** (silently empty) |
| `inputs.file` | fails — *value not set* | fails — *unsupported value: `ArrayList` cannot be converted to a file* |
| `inputs.dir` | fails — *value not set* | fails — *unsupported value: `ArrayList` cannot be converted to a directory* |
| `outputs.files` | fails — *value not set* | fails — *Cannot query the value … no value available* |

For contrast, `Project.files(...)` (i.e. a `ConfigurableFileCollection`, which resolves non-leniently) rejects an absent provider *uniformly* — `files(absent)`, `files([absent])`, and `files(absent, absent)` all fail on resolution with *"Cannot query the value of this property because it has no value available."* So the lenient, silently-empty behavior of `inputs.files([absent])` is inconsistent not only with `inputs.files(absent)` but also with the otherwise-equivalent `project.files(...)` collection.

`build.gradle`:
```groovy
def absent = objects.fileProperty()  // no value -> absent

// Passing the absent provider directly: FAILS validation ("value not set")
tasks.register("directFails") {
    inputs.files(absent).withPropertyName("p")
    doLast {}
}

// The same absent provider wrapped in a list: SUCCEEDS (silently treated as empty)
tasks.register("wrappedSucceeds") {
    inputs.files([absent]).withPropertyName("p")
    doLast {}
}
```

```
$ gradle directFails
...
A problem was found with the configuration of task ':directFails' (type 'DefaultTask').
Value not set
  Property 'p' doesn't have a configured value
    This property isn't marked as optional and no value has been configured
    For more information, please refer to https://docs.gradle.org/9.6.1/userguide/validation_problems.html#value_not_set in the Gradle documentation.
    Possible solutions:
      1. Assign a value to 'p'.
      2. Mark property 'p' as optional.
...

```

### Gradle version

Tested with 9.6, but likely it is there for a long time

### Build scan URL (optional)

_No response_

_No response_

DragonFSKY:
Fixes #38410.

Required `TaskInputs.files(...)` registrations now treat absent providers consistently when they are supplied directly, nested in ordinary collections, or supplied through varargs. These cases report the existing `value-not-set` validation problem instead of letting the provider be silently dropped by lenient file resolution.

Required `@InputFiles` properties returning ordinary collections or object arrays now reject nested absent providers as well, and the traversal rules are shared with the configuration-cache serialization so the two paths cannot diverge again.

Optional `inputs.files(...)` registrations keep the existing lenient behavior, and `inputs.file(...)` / `inputs.dir(...)` are unchanged. The configuration-cache task state path preserves provider-containing ordinary collection shapes so validation can still see nested providers after deserialization, while live Gradle collections and `TaskProvider` values continue to go through lenient `FileCollection` wrapping. Multi-valued output serialization keeps the previous wrapping behavior for non-map values and preserves mapped outputs as maps so their output identities survive restore.

Verification:
- `./gradlew :core-serialization-codecs:compileKotlin :configuration-cache:embeddedIntegTest --tests org.gradle.internal.cc.impl.ConfigurationCacheTaskSerializationIntegrationTest --no-daemon --stacktrace`
- `./gradlew :core:test --tests org.gradle.api.internal.tasks.DefaultFinalizingValidatingPropertyTest :core:embeddedIntegTest --tests org.gradle.api.file.TaskFilePropertiesIntegrationTest :core:checkstyleMain :core:codenarcIntegTest :configuration-cache:codenarcIntegTest --no-daemon --stacktrace`
- `git diff --check`

Final head (`c2d26b0a745`): `:core:test` 7 tests / 0 failures, `:core:embeddedIntegTest` 16 tests / 0 failures, `:configuration-cache:embeddedIntegTest` 26 tests / 0 failures — the configuration cache cases assert both the store and the reuse phase.

### AI assistance disclosure

I used Codex to help inspect the relevant Gradle code paths, draft the implementation, and run/interpret focused tests. I reviewed the resulting diff and validation output before submitting.


ljacomet:

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -449,22 +455,69 @@ fun WriteContext.adaptInputFileValueForSerialization(value: Any?, filePropertyTy
  * Only applied to the multi-valued [OutputFilePropertyType.FILES] and [OutputFilePropertyType.DIRECTORIES] —
  * [OutputFilePropertyType.FILE] and [OutputFilePropertyType.DIRECTORY] expect a single path-like value on read,
  * so wrapping in a [FileCollection] would break `outputs.file(...)` / `outputs.dir(...)`.
+ *
+ * [Map] values are kept as maps so mapped output property names survive configuration-cache restore.
  */
 private
 fun WriteContext.adaptOutputFileValueForSerialization(value: Any?, filePropertyType: OutputFilePropertyType): Any? {
-    if (value == null || filePropertyType == OutputFilePropertyType.FILE || filePropertyType == OutputFilePropertyType.DIRECTORY || !needsFileCollectionWrapper(value)) {
+    if (
```
asodja: ❌ Why is this change needed in this PR? This seems unrelated to the issue

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -436,10 +437,15 @@ suspend fun WriteContext.writeRegisteredPropertiesOf(task: Task) {
  */
 private
 fun WriteContext.adaptInputFileValueForSerialization(value: Any?, filePropertyType: InputFilePropertyType): Any? {
-    if (value == null || filePropertyType != InputFilePropertyType.FILES || !needsFileCollectionWrapper(value)) {
+    if (value == null || filePropertyType != InputFilePropertyType.FILES) {
         return value
     }
-    return isolate.owner.serviceOf<FileCollectionFactory>().resolvingLeniently(value)
+    val fileCollectionFactory = isolate.owner.serviceOf<FileCollectionFactory>()
+    return if (containsProviderRequiringValidation(value)) {
```
asodja: ❌ This traverses the value twice: once for containsProviderRequiringValidation and again in adaptProviderContainingInputFileValue(). Nested Callables may execute twice, and one-shot or mutating Iterables can produce different results between the two passes.

Could a single traversal directly return the adapted value? It should also handle Path before Iterable and avoid creating one resolving FileCollection for every provider-free leaf.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -449,22 +455,69 @@ fun WriteContext.adaptInputFileValueForSerialization(value: Any?, filePropertyTy
...
+        is Iterable<*> -> unpacked.map { adaptProviderContainingInputFileValue(it, fileCollectionFactory) }
+        else -> fileCollectionFactory.resolvingLeniently(unpacked)
+    }
+}
+
+
+private
+fun containsProviderRequiringValidation(value: Any?): Boolean {
+    val unpacked = DeferredUtil.unpackNestableDeferred(value)
+    return when (unpacked) {
```
asodja: ❌  This code and code in `adaptProviderContainingInputFileValue` looks suspicous.  Why `TaskNodeCodec` needs to know which type to validate and handle every type possible?

I think this needs a better solution

subprojects/core/src/integTest/groovy/org/gradle/api/file/TaskFilePropertiesIntegrationTest.groovy
```diff
@@ -215,14 +215,17 @@ class TaskFilePropertiesIntegrationTest extends AbstractIntegrationSpec {
         executedAndNotSkipped(":myTask")
 
         where:
-        description                                | declaration
-        "inputs.files(fileProperty)"               | 'inputs.files(objects.fileProperty()).optional().withPropertyName("inputProp")'
-        "inputs.files(fileProperty, fileProperty)" | 'inputs.files([objects.fileProperty()]).withPropertyName("inputProp")'
-        "inputs.file(fileProperty)"                | 'inputs.file(objects.fileProperty()).optional().withPropertyName("inputProp")'
-        "inputs.property(property)"                | 'inputs.property("inputProp", objects.property(String)).optional(true)'
+        description                                 | declaration
+        "inputs.files(fileProperty)"                | 'inputs.files(objects.fileProperty()).optional().withPropertyName("inputProp")'
```
asodja: ❓ Can we also test:
```
inputs.files(file("a.txt"), objects.fileProperty())
  .withPropertyName("inputProp")

inputs.files(Path.of("input.txt"))
.withPropertyName("inputProp")
```

subprojects/core/src/main/java/org/gradle/api/internal/tasks/properties/DefaultFinalizingValidatingProperty.java
```diff
@@ -16,15 +16,84 @@
...
 public class DefaultFinalizingValidatingProperty extends AbstractValidatingProperty {
     private final PropertyValue value;
+    private final boolean validateAbsentProviderElements;
     private LifecycleAwareValue lifecycleAware;
 
     public DefaultFinalizingValidatingProperty(String propertyName, PropertyValue value, boolean optional, ValidationAction validationAction) {
+        this(propertyName, value, optional, false, validationAction);
+    }
+
+    public DefaultFinalizingValidatingProperty(
```
asodja: ❌ I don't think we should modify `DefaultFinalizingValidatingProperty` to have special casing for files but I think this should be solved on a different level. 

For example, DefaultTaskProperties.ValidationVisitor could select a specialized action for required FILES properties:
```
ValidationAction action =
    filePropertyType == InputFilePropertyType.FILES && !optional
        ? ValidationActions.REQUIRED_INPUT_FILES
        : ValidationActions.inputValidationActionFor(filePropertyType);
```

Note that current solution also cause StackOverflow for `hasAbsentProvider()` with `Path.of("input.txt")`

asodja (changes requested):
Thanks for looking into this. 

This PR needs adjustments and we can't merge it as it is. 
I think we should not modify DefaultFinalizingValidatingProperty, but rather solve it on a level of `ValidationActions`. Also changes in TaskNodeCodec are suspicous and they don't look ok (TaskNodeCodec should not handle every type possible for file inputs).

I noticed some changes that don't seem related to the issue, are they needed here?

Additionally we should add some extra cases in tests.

See my comments below

DragonFSKY:
Thanks for the review, and sorry the initial approach was overcomplicated. I’ve reworked it as suggested, removed the unrelated changes, and added the requested tests.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/FileCollectionCodec.kt
```diff
@@ -184,7 +210,14 @@ class CollectingVisitor : AbstractVisitor() {
                 // being referenced from a different task.
                 val provider = fileCollection.provider
                 if (provider !is TaskProvider<*>) {
-                    elements.add(ProviderBackedFileCollectionSpec(fileCollection.resolver, fileCollection.providerResolutionStrategy, provider))
+                    elements.add(
+                        ProviderBackedFileCollectionSpec(
+                            fileCollection.resolver,
+                            fileCollection.providerResolutionStrategy,
+                            provider,
+                            retainInputFileValidation && fileCollection.providerResolutionStrategy == ProviderResolutionStrategy.ALLOW_ABSENT
```
asodja: ❌ It looks like this case is different with cc and without cc
```  
def sources = objects.domainObjectSet(Provider)
  sources.add(objects.fileProperty()) // absent

  tasks.register("myTask") {
      inputs.files(sources).withPropertyName("inputProp")
  }
```
  Observed behavior:

  - Without configuration cache: myTask succeeds.
  - With configuration cache: myTask fails with “Property 'inputProp' doesn't have a configured value.”

And supposedly because of this line. 

  DomainObjectCollection is skipped during normal validation, but this code loses that information and validates its providers after serialization. As a result, a task that succeeds without configuration cache can fail with value-not-set when configuration cache is enabled.

Is there a way to make behavior consistent?

asodja (changes requested):

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/FileCollectionCodec.kt
DragonFSKY: Thanks for catching this. I updated the serialization path to skip `DomainObjectCollection` when collecting nested validation providers, matching normal validation, and added a regression test covering no configuration cache, cache store, and cache reload. `ConfigurationCacheTaskSerializationIntegrationTest` passes.

subprojects/core/src/main/java/org/gradle/api/internal/tasks/properties/ValidationActions.java
```diff
@@ -44,6 +47,19 @@ public void doValidate(String propertyName, Object value, PropertyValidationCont
         public void validate(String propertyName, Supplier<Object> propertyValue, PropertyValidationContext context) {
         }
     },
+    REQUIRED_INPUT_FILES("file collection") {
+        @Override
+        public void doValidate(String propertyName, Object value, PropertyValidationContext context) {
+            if (hasNestedAbsentProvider(value)) {
+                AbstractValidatingProperty.reportValueNotSet(propertyName, context, true);
```
asodja, line 54: ❌ With `hasConfigurableValue` hardcoded to `true` here, the solution text differs between the direct and nested shapes for the same provider:

```groovy
inputs.files(providers.provider { null })    // "The value of 'inputProp' is calculated, make sure a valid value can be calculated"
inputs.files([providers.provider { null }])  // "Assign a value to 'inputProp'"
```

The first one is right — `providers.provider { }` isn't a configurable property. Could you derive the argument from the absent provider instead, the way `AbstractValidatingProperty.validate` does for the direct case? E.g. have the lookup return the absent provider rather than a boolean:

```java
Provider<?> absentProvider = findNestedAbsentProvider(value);
if (absentProvider != null) {
    AbstractValidatingProperty.reportValueNotSet(propertyName, context, AbstractValidatingProperty.hasConfigurableValue(absentProvider));
}
```

(`hasConfigurableValue` is private in `AbstractValidatingProperty`; package-private should be fine.) Existing cases keep the same message since `RegularFileProperty` is `HasConfigurableValue`; only non-configurable providers change.

It'd be good to add a `providers.provider { null }` row to the "required … fails" table that also asserts the solution text.

subprojects/core/src/main/java/org/gradle/api/internal/tasks/properties/DefaultTaskProperties.java
```diff
@@ -226,7 +226,11 @@ public void visitInputFileProperty(
             PropertyValue value,
             InputFilePropertyType filePropertyType
         ) {
-            taskPropertySpecs.add(new DefaultFinalizingValidatingProperty(propertyName, value, optional, ValidationActions.inputValidationActionFor(filePropertyType)));
+            ValidationAction validationAction =
+                filePropertyType == InputFilePropertyType.FILES && !optional
+                    ? ValidationActions.REQUIRED_INPUT_FILES
```
asodja, line 231: ❓ It looks like this also applies to annotated properties, since `AbstractInputFilePropertyAnnotationHandler` calls the same visitor: a required `@InputFiles` returning a `Collection`/array with an absent provider would now fail too. I think that's reasonable, but could you add an integration test for it (e.g. a task class with a required `@InputFiles List<Object>` getter returning `[absentProvider]`) and mention it in the PR description?

subprojects/core/src/integTest/groovy/org/gradle/api/file/TaskFilePropertiesIntegrationTest.groovy
```diff
@@ -249,30 +250,13 @@ class TaskFilePropertiesIntegrationTest extends AbstractIntegrationSpec {
...
-        "inputs.files(fileProperty)" | 'inputs.files(objects.fileProperty()).withPropertyName("inputProp")'
-        "inputs.file(fileProperty)"  | 'inputs.file(objects.fileProperty()).withPropertyName("inputProp")'
-        "inputs.property(property)"  | 'inputs.property("inputProp", objects.property(String))'
+        description                                | declaration
+        "inputs.files(fileProperty)"               | 'inputs.files(objects.fileProperty()).withPropertyName("inputProp")'
+        "inputs.files([fileProperty])"             | 'inputs.files([objects.fileProperty()]).withPropertyName("inputProp")'
+        "inputs.files(fileProperty, fileProperty)" | 'inputs.files(objects.fileProperty(), objects.fileProperty()).withPropertyName("inputProp")'
+        "inputs.files(file, fileProperty)"         | 'inputs.files(file("input.txt"), objects.fileProperty()).withPropertyName("inputProp")'
+        "inputs.file(fileProperty)"                | 'inputs.file(objects.fileProperty()).withPropertyName("inputProp")'
+        "inputs.property(property)"                | 'inputs.property("inputProp", objects.property(String))'
```
asodja: ❓ Could you add a test for a required `@InputFiles List<Object>` getter returning `[absentProvider]` on a task class? Also, there seems to be a stray blank line before the closing brace.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -595,6 +616,33 @@ suspend fun ReadContext.readInputPropertiesOf(task: Task) =
     }
 
 
+private
+fun collectNestedInputFileProviders(value: Any?): List<Provider<*>> {
+    val providers = mutableListOf<Provider<*>>()
+
+    fun collect(value: Any?, nested: Boolean) {
```
asodja, line 623: ❌ This duplicates the shape rule in `ValidationActions.hasNestedAbsentProvider` (recurse into `Collection`/arrays, skip `DomainObjectCollection`, treat nested `Provider`s specially). The two copies already went out of sync once during this PR — that was the `DomainObjectCollection` inconsistency from the last review. 

Could you extract one shared helper and use it from both places, so the CC and non-CC paths can't diverge again? `FileParameterUtils` looks like a good home, e.g. `List<Provider<?>> findNestedProviders(Object value)` — then validation can check presence on the result and this codec can serialize it.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -595,6 +616,33 @@ suspend fun ReadContext.readInputPropertiesOf(task: Task) =
...
+    }
+
+    collect(value, false)
+    return providers
+}
+
+
+private
+fun requiredInputFilesValidationProvider(providers: List<Provider<*>>): Provider<*> {
+    val merged = MergeProvider<Any>(providers.uncheckedCast<List<Provider<Any>>>())
```
asodja: ❓ If I read it right, `MergeProvider.calculatePresence` computes the full values once all items are present, so after a CC reload validation would resolve every nested provider, while without CC only `isPresent()` is called (as the new unit test asserts). Would a presence-only provider work here to keep both paths equivalent?

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/TaskNodeCodec.kt
```diff
@@ -371,8 +373,11 @@ suspend fun WriteContext.writeRegisteredPropertiesOf(task: Task) {
         writePropertyValue(kind, propertyName, propertyValue)
     }
 
-    suspend fun writeInputProperty(propertyName: String, propertyValue: Any?) =
-        writeProperty(propertyName, propertyValue, PropertyKind.InputProperty)
+    suspend fun writeInputProperty(propertyName: String, propertyValue: Any?, validationProviders: List<Provider<*>>) {
```
asodja: ❓ This writes an (empty) list for every input property. Could we only write it for `FILES && !optional`?

subprojects/core/src/main/java/org/gradle/api/internal/tasks/properties/AbstractValidatingProperty.java
```diff
@@ -69,8 +67,7 @@ public void validate(PropertyValidationContext context) {
         Object unnested = DeferredUtil.unpackNestableDeferred(value.call());
         if (isPresent(unnested)) {
             // only resolve deferred values if actually required by some action
-            Supplier<Object> valueSupplier = Suppliers.memoize(() -> DeferredUtil.unpack(unnested));
-            validationAction.validate(propertyName, valueSupplier, context);
+            validationAction.validate(propertyName, () -> unnested, context);
```
asodja: ❓ Since the supplier no longer yields an unpacked value, could you update the `ValidationAction.validate` javadoc accordingly?

subprojects/core/src/main/java/org/gradle/api/internal/tasks/properties/ValidationActions.java
```diff
@@ -273,10 +318,11 @@ private static void validateNotInReservedFileSystemLocation(String propertyName,
 
     @Override
     public void validate(String propertyName, Supplier<Object> value, PropertyValidationContext context) {
+        Object resolvedValue = DeferredUtil.unpack(value.get());
```
asodja: ❓ Nit: `unpack` used to run inside the `try`, so an `UnsupportedNotationException` from resolving became the friendly problem. Could you move it back inside?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheTaskSerializationIntegrationTest.groovy
```diff
@@ -539,6 +540,70 @@ class ConfigurationCacheTaskSerializationIntegrationTest extends AbstractConfigu
         outputContains("inputs = [task1.txt, task2.txt]")
     }
 
+    @Issue("https://github.com/gradle/gradle/issues/38410")
+    def "required TaskInputs.files with an absent provider in a list fails after configuration cache reload"() {
```
asodja: ❓ Could you also cover the varargs (`inputs.files(file, absent)`) and nested-list (`[[absent]]`) shapes here? They seem to work, but only the non-CC tests exercise them.

subprojects/core/src/main/java/org/gradle/api/internal/tasks/properties/ValidationActions.java
asodja, line 321: 💅 Nit: `unpack` used to run inside the `try`, so an `UnsupportedNotationException` from resolving became the friendly problem. Could you move it back inside?

asodja (changes requested):

asodja:
@bot-gradle test this

asodja:
Thanks, I think this is close to the finish line: the `ValidationActions` approach and the CC parity look good to me now. 

Could you take a look at a few remaining things? Mainly the error message for calculated providers, a test for annotated `@InputFiles` (this change seems to affect them too), and sharing the traversal logic between validation and the CC codec. Details inline.

Since this changes behavior for existing builds it will need an upgrade-guide note. If you want to contribute that too, please open a separate PR, otherwise I will add it later.

DragonFSKY:
Thanks — all four items are addressed in the last two commits:

- the nested provider traversal is shared between validation and the configuration cache codec now
- the `value-not-set` solution comes from the absent provider itself, and the cache keeps that distinction per provider across store and reuse
- `DeferredUtil.unpack` is back inside the `try`
- added the annotated `@InputFiles` test, and the runtime and cache tables now assert the solution text

Ran `:core:test` (9), `:core:embeddedIntegTest` (18) and `:configuration-cache:embeddedIntegTest` (30) locally, all green — the cache cases check both the store and the reuse phase.

The upgrade guide note is in #39209.

asodja:
@bot-gradle test this

asodja (approved):
LGTM

Thanks for the contribution! We have some CI issues that we are resolving, so I may need a day or two to get it merged

PR #38456, opened 2026-07-08
Skip SIGINT taskbar test where the signal is undeliverable

reinsch82:
### Context

`TaskbarProgressResetFunctionalTest > sends OSC 9;4;0 reset sequence when build receives SIGINT` fails on CI (nightly, macOS agents): `kill -SIGINT` has no effect, so the build never cancels.

Cause (confirmed by a CI probe): on background/service processes like CI agents, SIGINT is inherited as `SIG_IGN` — a disposition that survives `exec` and that HotSpot leaves in place — so the forked client ignores it. Resending cannot help; it is not a product bug (an interactive Ctrl+C is deliverable).

Fix: `Assume`-skip the test when SIGINT is undeliverable in the current process tree (detected by forking a child that signals itself). It still runs for real wherever SIGINT works.

Issue: https://github.com/gradle/gradle-private/issues/5153

### Contributor Checklist
- [x] [Review Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md)
- [x] Commits are [signed off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff)
- [x] Contributed code is distributable under Apache License 2.0
- [x] Integration tests provided (test-only change)

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/logging/src/integTest/groovy/org/gradle/internal/logging/console/TaskbarProgressResetFunctionalTest.groovy
+++ b/platforms/core-runtime/logging/src/integTest/groovy/org/gradle/internal/logging/console/TaskbarProgressResetFunctionalTest.groovy
@@ -22,6 +22,7 @@ import org.gradle.test.fixtures.ConcurrentTestUtil
 import org.gradle.test.preconditions.TestExecutionPreconditions
 import org.gradle.test.preconditions.OsTestPreconditions
 import org.gradle.test.precondition.Requires
+import org.junit.Assume
 import spock.lang.Issue
 
 /**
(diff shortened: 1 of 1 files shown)


platforms/core-runtime/logging/src/integTest/groovy/org/gradle/internal/logging/console/TaskbarProgressResetFunctionalTest.groovy
```diff
@@ -90,6 +101,34 @@ class TaskbarProgressResetFunctionalTest extends AbstractIntegrationSpec {
...
+     * On background/service processes such as CI build agents, SIGINT is inherited as
+     * SIG_IGN — a disposition that survives exec and that HotSpot leaves in place — so
+     * the forked client ignores it and cannot be cancelled by {@code kill -SIGINT},
+     * regardless of how often the signal is resent. Detect this by forking a child the
+     * same way the executer forks the client and having it signal itself with SIGINT:
+     * if it survives (prints the marker), SIGINT is inherited-ignored; if it is killed,
+     * SIGINT is deliverable. Works on both Linux and macOS.
+     * See https://github.com/gradle/gradle-private/issues/5153.
+     */
+    private static boolean sigintDeliverableToForkedChild() {
```
jbartok, line 109: 🤔 I wouldn't make this a method, but something even more static. Should be done only once per JVM instance, right? Yes, now it's called in a single test, but that might change in the future.

platforms/core-runtime/logging/src/integTest/groovy/org/gradle/internal/logging/console/TaskbarProgressResetFunctionalTest.groovy
```diff
@@ -90,6 +101,34 @@ class TaskbarProgressResetFunctionalTest extends AbstractIntegrationSpec {
         gradle.standardOutput.contains(OSC_RESET)
     }
 
+    /**
+     * Whether SIGINT can actually be delivered to a child process forked by this test
```
jbartok: 🤔 Duplicates the comment from the assume above to some extent. Too verbose, please at least remove the duplication.

jbartok (approved):
Have some requests, but they can be in a follow-up PR, if this one is urgent.

PR #38458, opened 2026-07-08
Proceed with build and discarding the cache when CC cache entry is corrupted

### LINKED ISSUE #26663: Auto-clear the config cache when it is corrupted.
### Expected Behavior

If the config cache is corrupted and fails with a `ClassNotFoundException` or other exception that indicates a corrupted config cache, then just clear the corrupted cache automatically. 

See https://github.com/gradle/gradle/issues/25864

It just fails repeatedly until you manually do `rm -rf .gradle/configuration-cache` 

### Context

Of course, it is best to fix the root cause of the corruption issues.

But rather than just fail repeatedly and leaving to the developer to figure out that `rm -rf .gradle/configuration-cache` will get them unstuck, Gradle could just detect the corruption and clear the corrupted cache.

ov7a:
Fixes
* https://github.com/gradle/gradle/issues/26663

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ov7a:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheAwareBuildTreeWorkController.kt
```diff
@@ -87,15 +93,41 @@ class ConfigurationCacheAwareBuildTreeWorkController(
...
+                }
+            } ?: run {
+                maybeDumpHeap("cc-miss-store")
+                storeAndReload(scheduleTaskSelectorPostProcessing)
+            }
+        } catch (recoveryFailure: Throwable) {
+            logger.info("Recomputing the configuration cache entry after a failed load did not succeed", recoveryFailure)
+            return TaskRunResult.ofScheduleFailure(originalFailure)
         }
+        logger.warn("The configuration cache entry could not be loaded and has been recomputed.", originalFailure)
```
ov7a: 🤔 I'm not sure if it's the best way to expose the original failure.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultConfigurationCache.kt
```diff
@@ -295,6 +295,24 @@ class DefaultConfigurationCache internal constructor(
...
+    override fun recoverFromFailedLoad(failure: Throwable): Boolean {
+        if (!startParameter.isRecoverFromCacheCorruption) {
+            return false
+        }
+        resetState()
+        return true
+    }
+
+    private fun resetState() {
+        scopeRegistryListener.reattach()
```
ov7a: 👇 Please take extra care about this part, I'm not really sure about it.

lkasso (approved):
Docs LGTM

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/fixtures/AbstractConfigurationCacheOptInFeatureIntegrationTest.groovy
```diff
@@ -37,4 +40,30 @@ abstract class AbstractConfigurationCacheOptInFeatureIntegrationTest extends Abs
...
+    protected static void corrupt(TestFile file) {
+        assert file.exists()
+        file.bytes = "corrupt".bytes
+    }
+
+    /**
+     * Corrupts the stored work graph while keeping {@code entry.bin} and the fingerprints valid, so the entry is still
+     * selected and loaded (exercising the load-time recovery path rather than the fingerprint check).
+     */
+    protected static void corruptWorkState(TestFile cacheDir) {
```
alllex: 💅 The arg is always `.gradle/configuration-cache` and the method always assumes the layout of the provided dir. We can make this method non-static and resolve the cache dir in a single place

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheIntegrationTest.groovy
```diff
@@ -218,6 +218,183 @@ class ConfigurationCacheIntegrationTest extends AbstractConfigurationCacheIntegr
...
+        configurationCache.assertStateStored()
+        outputContains("Hello")
+
+        when: "the stored state is corrupted while the fingerprint stays valid, so the entry is selected for loading"
+        corruptWorkState(file(".gradle/configuration-cache"))
+        executer.withStackTraceChecksDisabled()
+        configurationCacheRun "hello"
+
+        then: "the failure is reported with its cause, the entry is recomputed in place, and the build runs to completion"
+        outputContains("The configuration cache entry could not be loaded and has been recomputed.")
```
alllex: 🤔 I don't think the `recomputed` terminology is used anywhere. We should align with what we already have.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheIntegrationTest.groovy
```diff
@@ -218,6 +218,183 @@ class ConfigurationCacheIntegrationTest extends AbstractConfigurationCacheIntegr
...
+        failureCauseContains("BOOM from readResolve")
+
+        when: "the unreadable entry is selected on the next run, so recovery is attempted and fails"
+        executer.withStackTraceChecksDisabled()
+        configurationCacheFails "broken"
+
+        then: "the original failure is reported, not the secondary error from the failed recompute"
+        failureDescriptionContains("Could not load the value of field `prop`")
+        failureCauseContains("BOOM from readResolve")
+        outputDoesNotContain("Cannot create a new id")
```
alllex: 💭 This is an assertion that can easily go stale silently

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultConfigurationCache.kt
```diff
@@ -295,6 +295,24 @@ class DefaultConfigurationCache internal constructor(
         return loadWorkGraph(graph, graphBuilder, true)
     }
 
+    override fun recoverFromFailedLoad(failure: Throwable): Boolean {
```
alllex: 💅 The failure parameter is never used

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheAwareBuildTreeWorkController.kt
```diff
@@ -87,15 +93,41 @@ class ConfigurationCacheAwareBuildTreeWorkController(
...
+    }
 
-        // Store and reload the graph for the execution.
-        cache.finalizeCacheEntry()
-        buildRegistry.visitBuilds { build ->
-            build.beforeModelReset().rethrow()
-        }
-        buildRegistry.visitBuilds { build ->
-            build.resetModel()
+    private fun recomputeAfterFailedLoad(
```
alllex: ❌ The code of this method significantly duplicates the logic of the caller `scheduleAndRunRequestedTasks` method, while no longer carrying the same explanations of conditions and copying complex conditionals.

This will likely be better off being a new method altogether that has a parameter allowing it to care for failed-load recovery. It will dictate, for instance, what to do in case the `loadOrScheduleRequestedTasks` is unsuccessful. In the initial case, it will call itself without recovery. In the recorver path it will immediately rethrow the failure, but will do so explicitly.

Another thing that seems like a smell is a nested call to `workGraph.withNewWorkGraph`, while keeping the outer one alive. I don't fully grasp the consequences of this. I'd much prefer if on load failure we'd back out of the first `workGraph.withNewWorkGraph` call and then attempted the second one.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheAwareBuildTreeWorkController.kt
alllex: Can we report it as a Problems API warning instead?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheAwareBuildTreeWorkController.kt
```diff
@@ -87,15 +93,41 @@ class ConfigurationCacheAwareBuildTreeWorkController(
...
+                    null
+                } else {
+                    TaskRunResult.ofExecutionResult(workExecutor.execute(finalizedGraph.graph))
+                }
+            } ?: run {
+                maybeDumpHeap("cc-miss-store")
+                storeAndReload(scheduleTaskSelectorPostProcessing)
+            }
+        } catch (recoveryFailure: Throwable) {
+            logger.info("Recomputing the configuration cache entry after a failed load did not succeed", recoveryFailure)
```
alllex: 🤔 At this point, we don't know whether the failure was caused by the fact of recomputation, or if it is a regular build failure. I am suspicious of catching an arbitrary throwable here and reporting it.

Also, I think this outer try-catch might not be necessary, if the logic of this method is unified with `scheduleAndRunRequestedTasks` as per the comment above.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheAwareBuildTreeWorkController.kt
```diff
@@ -104,6 +136,15 @@ class ConfigurationCacheAwareBuildTreeWorkController(
         }
     }
 
+    private fun resetBuildModels() {
```
alllex: 💭 It's the only place where we reset the models. It feels like the abstraction is leaking somewhat, and this concern should be on the side of the `buildRegistry` and not this build-tree work controller.

I understand this PR just moved this code so far, but I think it will have better time living even as a default method on the `BuildStateRegistry`, which is about resetting the model of the entire build tree.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheIntegrationTest.groovy
```diff
@@ -218,6 +218,183 @@ class ConfigurationCacheIntegrationTest extends AbstractConfigurationCacheIntegr
         outputContains("Recreating configuration cache")
     }
 
+    @Issue("https://github.com/gradle/gradle/issues/26663")
```
alllex: ❓ How does the DV plugin handle potentially observing two CC Load build operations?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/BuildTreeConfigurationCache.kt
```diff
@@ -46,6 +46,16 @@ interface BuildTreeConfigurationCache {
...
     fun loadRequestedTasks(graph: BuildTreeWorkGraph, graphBuilder: BuildTreeWorkGraphBuilder?): BuildTreeWorkGraph.FinalizedGraph
 
+    /**
+     * Invoked when [loadOrScheduleRequestedTasks] fails while reusing a cache entry, e.g. because it is corrupted.
+     *
+     * When the failure is recoverable and recovery is enabled, reports it, discards the broken entry so it is never
+     * reused, and reconfigures this cache so the rest of this invocation recomputes the requested work without reusing
+     * or storing a cache entry (as if the cache had missed), and returns `true`. Otherwise returns `false` and the
+     * failure should be reported as a build failure.
+     */
```
alllex: 🤔 This javadoc doesn't fully make sense to me. Maybe some logic has moved during refactors?

In terms of API design, it feels like asking about the recover and doing a reset of the cache's inner structures are two significantly different operations. I'd split the second one out.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultConfigurationCache.kt
```diff
@@ -295,6 +295,24 @@ class DefaultConfigurationCache internal constructor(
...
+        return true
+    }
+
+    private fun resetState() {
+        scopeRegistryListener.reattach()
+        loadedSideEffects.clear()
+        cacheEntryRequiresCommit = false
+        entryDiscardRequested = false
+        gradlePropertiesController.unloadAll()
+        cacheAction = Store
```
alllex: 🤔 This feels suspicious, because determining an action is not as trivial. We are basically dealing with a situation of we thought we are loading, but that failed. So we need to re-determine the action as if the entry wasn't on disk.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultConfigurationCache.kt
```diff
@@ -295,6 +295,24 @@ class DefaultConfigurationCache internal constructor(
...
+        modelSideEffectExecutor.sideEffectStore = buildTreeModelSideEffects
```
alllex: 🤔 This feels like a no-op

alllex (changes requested):
Requires a number of changes, and I also have a number of questions.

I think this change also pushes further some of the leaky abstractions and couplings that already existed, especially with the reset/reattach logic of the build-tree CC cache itself. I definitely don't like how this is looking architecturally, but it might be too big of an ask to try and tackle in this PR. The question for me is whether we need a preparation PR that makes the contracts and state transitions clear(er) before we ship the current change.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIntegrationTest.groovy
```diff
@@ -19,9 +19,43 @@ package org.gradle.internal.cc.impl.isolated
...
+        when: "the stored work graph is corrupted"
+        corruptWorkState(file(".gradle/configuration-cache"))
+        executer.withStackTraceChecksDisabled()
+        isolatedProjectsRun("thing")
+
+        then: "the entry is recomputed and the build completes"
+        outputContains("The configuration cache entry could not be loaded and has been recomputed.")
+        outputContains("thing ran")
+
+        when: "the build runs again"
```
6hundreds: 💅 🦾 I find such comments useless. This comments must grab reader's attention when something exceptional happens, not to duplicate what the code says

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/fixtures/AbstractConfigurationCacheOptInFeatureIntegrationTest.groovy
```diff
@@ -19,10 +19,13 @@ package org.gradle.internal.cc.impl.fixtures
 import org.gradle.initialization.StartParameterBuildOptions
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
 import org.gradle.integtests.fixtures.configurationcache.ConfigurationCacheProblemsExecutionResultFixture
+import org.gradle.test.fixtures.file.TestFile
 
 abstract class AbstractConfigurationCacheOptInFeatureIntegrationTest extends AbstractIntegrationSpec {
```
6hundreds, line 24: ❓ Not sure this abstract class is a good place for putting entry corruption logic, as I don't think dozens of its implementations must be aware of it. May be we should consider keep it as a `trait`?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIntegrationTest.groovy
```diff
@@ -19,9 +19,43 @@ package org.gradle.internal.cc.impl.isolated
...
 import org.gradle.test.fixtures.file.TestFile
 import org.gradle.util.internal.ToBeImplemented
+import spock.lang.Issue
 
 class IsolatedProjectsIntegrationTest extends AbstractIsolatedProjectsIntegrationTest implements TasksWithInputsAndOutputs {
 
+    @Issue("https://github.com/gradle/gradle/issues/26663")
+    def "recovers from a corrupted entry with isolated projects enabled"() {
+        given:
+        settingsFile << ""
```
6hundreds: 💅 This can be omitted

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsIntegrationTest.groovy
```diff
@@ -19,9 +19,43 @@ package org.gradle.internal.cc.impl.isolated
 import org.gradle.api.tasks.TasksWithInputsAndOutputs
 import org.gradle.test.fixtures.file.TestFile
 import org.gradle.util.internal.ToBeImplemented
+import spock.lang.Issue
 
 class IsolatedProjectsIntegrationTest extends AbstractIsolatedProjectsIntegrationTest implements TasksWithInputsAndOutputs {
 
+    @Issue("https://github.com/gradle/gradle/issues/26663")
+    def "recovers from a corrupted entry with isolated projects enabled"() {
```
6hundreds: 💭  This test is almost a verbatim copy of test with CC Encryption, and I believe it should be other plain CC test. Maybe it's better to structure it in a way introducing a dedicated test that parametrized by a particular mode?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingModelsWithDependencyResolutionIntegrationTest.groovy
```diff
@@ -750,4 +753,117 @@ class IsolatedProjectsToolingModelsWithDependencyResolutionIntegrationTest exten
             modelsReused(":", ":b", ":c", ":d", ":buildSrc")
         }
     }
+
+    @Issue("https://github.com/gradle/gradle/issues/26663")
+    def "discards a corrupted reused project state entry so the next build recovers"() {
+        given:
+        withSomeToolingModelBuilderPluginThatPerformsDependencyResolutionInBuildSrc()
+        settingsFile << """
```
6hundreds: 💅 Use `settingsFile(@GroovySettingsScriptLanguage String append)` for new tests

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsToolingModelsWithDependencyResolutionIntegrationTest.groovy
```diff
@@ -750,4 +753,117 @@ class IsolatedProjectsToolingModelsWithDependencyResolutionIntegrationTest exten
...
+            include("a")
+            include("b")
+        """
+        file("a/build.gradle") << """
```
6hundreds: 💅 Use `buildFile(@GroovySettingsScriptLanguage String append)` for new tests

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheIntegrationTest.groovy
```diff
@@ -218,6 +218,183 @@ class ConfigurationCacheIntegrationTest extends AbstractConfigurationCacheIntegr
...
+        corruptEntry(file(".gradle/configuration-cache"))
+        configurationCacheFails "hello", DISABLE_CC_RECOVERY
+
+        then: "the build fails instead of recovering"
+        outputDoesNotContain("has been recomputed")
+        outputDoesNotContain("will be discarded")
+        outputDoesNotContain("Hello")
+
+        where:
+        corruptedPart | corruptEntry
```
6hundreds: ❓ Can fingerprint be corrupted?

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultConfigurationCache.kt
6hundreds: 💅 I'd prefer the name to be prefixed by `is`, otherwise it looks like a command that returns it result(was is successfull or not)

6hundreds: UPD. Sorry, IT'S a command indeed

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheIntegrationTest.groovy
ov7a: Note: with the current master, a build with corrupted CC does not produce a Build Scan.

[Sample DV scan](https://ge.gradle.org/s/hhtysug34hte4) for cache corruption recovery.

I don't see any notable issues with the scan, nor any highlights that CC was loaded twice.

ov7a:
I extracted out preparation refactoring into 
* https://github.com/gradle/gradle/pull/38811

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheIntegrationTest.groovy
ov7a: https://github.com/gradle/gradle/issues/38814

ov7a:
@bot-gradle test this

ov7a:
@bot-gradle test this

ov7a:
@bot-gradle test this

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheValueSourceIntegrationTest.groovy
```diff
@@ -67,6 +68,44 @@ class ConfigurationCacheValueSourceIntegrationTest extends AbstractConfiguration
...
+
+        then:
+        configurationCache.assertStateStored()
+
+        when:
+        executer.withStackTraceChecksDisabled()
+        configurationCacheFails(integrityCheck, "ok")
+
+        then:
+        failureCauseContains("cannot deserialize Poison")
```
alllex: 💅 Let's assert the error as it is presented to the user. In this case, it's this:
```
* What went wrong:
Could not load the value of field `value` of `org.gradle.internal.Try$Success` bean found in field `value` of `org.gradle.api.internal.provider.DefaultValueSourceProviderFactory$DefaultObtainedValue` bean found in class `PoisonValueSource`.
> cannot deserialize Poison 
```

The first part should be asserted as well, since it is a horrible show of internals

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_enabling.adoc
```diff
@@ -281,6 +281,17 @@ To temporarily revert to the previous behavior, set the Gradle property:
...
 
+3. Starting with Gradle 9.9.0, when a Configuration Cache entry cannot be loaded (for example because it is corrupted), Gradle discards the broken entry and recomputes the requested work as a cache miss, instead of failing the build.
++
+Prior to Gradle 9.9.0, such an entry caused the build to fail repeatedly until the cache was deleted manually (for example with `rm -rf .gradle/configuration-cache`).
++
+To restore the previous behavior and surface corruption as a build failure, set the Gradle property:
++
+[source,properties]
+----
+org.gradle.internal.configuration-cache.recover-from-corruption=false
```
alllex: ❌ Internal properties must not be documented in the public docs

https://github.com/gradle/gradle/blob/master/architecture/standards/0010-gradle-properties-naming.md?plain=1#L42

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheEntryReadException.kt
```diff
@@ -0,0 +1,41 @@
...
+ * (such as a value source throwing or an integrity check detecting a problem).
+ */
+internal class ConfigurationCacheEntryReadException(cause: Throwable) : RuntimeException(cause)
+
+
+/**
+ * Reads a part of a stored cache entry. An unclassified failure becomes a [ConfigurationCacheEntryReadException] unless
+ * integrity checking is enabled.
+ */
+internal inline fun <T> readStoredState(isIntegrityCheckEnabled: Boolean, read: () -> T): T =
```
alllex: 💭 It's a bit off for this function(ality) to live in the file named after an exception.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheCorruptionRecoveryIntegrationTest.groovy
```diff
@@ -0,0 +1,378 @@
...
+
+        when:
+        configurationCacheRun("hello")
+
+        then:
+        configurationCache.assertStateStored()
+        outputContains("Hello")
+
+        when:
+        "$corruptState"()
```
alllex: 💅 I saw that in the original PR and didn't like that this makes the helper functions (at the bottom of the class) unused, and also ties the function name to the test name. You also can't navigate via cmd+click to such a function. I addressed this (in the test file that gets removed by this PR) by splitting the two.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheEntrySelector.kt
```diff
@@ -50,8 +51,17 @@ internal class ConfigurationCacheEntrySelector(
     private val virtualFileSystem: BuildLifecycleAwareVirtualFileSystem,
     private val buildOperationRunner: BuildOperationRunner,
     private val gradlePropertiesController: GradlePropertiesController,
-    private val isolateOwner: IsolateOwner
+    private val isolateOwner: IsolateOwner,
+    private val problems: ConfigurationCacheProblems
 ) {
+    private
+    val isRecoveryEnabled: Boolean
+        get() = startParameter.isRecoverFromCacheCorruption && !startParameter.isIntegrityCheckEnabled
```
alllex: 🤔 This is duplicated logic, which should be consolidated in `ConfigurationCacheStartParameter`

PR #38462, opened 2026-07-08
ProjectInternal no longer implements DomainObjectContext

jvandort:
DomainObjectContext intends to model some abstract mutable 'context' under which 'domain objects' are created. It includes accessors to determine the context's identity and type (are you a script, project, etc).

In practice, only DependencyManagement uses this to figure out what sort of 'environment' it is running under, as behavior might change if DM is running in a settings context vs project.

We should expand this to better model contexts other than Project, but this PR specifically intends to decouple the domain object context for a Project from that Project itself. Currently, ProjectInternal _is_ a DomainObjectContext, which means users of this interface may unknowingly be carrying around a
 Project instance without knowing.

This is the first step to cleaning up this pattern, decoupling the DOC from the Project itself.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ProblemReportingCrossProjectModelAccess.kt
@@ -70,7 +70,6 @@ import org.gradle.internal.logging.StandardOutputCapture
 import org.gradle.internal.metaobject.DynamicInvokeResult
 import org.gradle.internal.metaobject.DynamicObject
 import org.gradle.internal.metaobject.HierarchicalDynamicObject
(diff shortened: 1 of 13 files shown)


jvandort:
@bot-gradle test this

bamboo (approved):
Good move.

One comment, in general, the rules for writing good commit message subject lines and bodies also apply to PR titles and summaries:
- [Use the imperative mood in the subject line](https://cbea.ms/git-commit/#imperative)

jvandort:
Yeah I was unsure on how to better phrase this PR title. 

`Remove implementation of DomainObjectContext from ProjectInternal`
`Un-implement DomainObjectContext on ProjectInternal`
`Stop implementing DomainObjectContext on ProjectInternal`.

I suppose any could work.

In any case, thanks for the review

Issue #35996, opened 2025-12-10
Require explicit user opt-in into Parallel Tooling Model Building


alllex:
### Context

Parallel Tooling Model Building or "Parallel Model Fetching" (PMF) is a feature that allows Tooling API clients executing work more efficiently on multi-core machines. The most common application of this is the IDE sync scenarios, where a user build information can be imported faster due to the parallelism, unlocking many IDE features developers are used to.

Historically, Gradle users didn't have direct control over PMF. Instead, it was enabled implicitly when `org.gradle.parallel` aka "Parallel Execution" (PE) was enabled in the gradle properties.

Gradle 9.4.0 [introduces](https://github.com/gradle/gradle/pull/35948) a new property `org.gradle.tooling.parallel`, which directly controls PMF and takes precendence over `org.gradle.parallel`. However, to preserve "performance backward-compatibility", the old behavior is still honored, meaning that Parallel Execution implies PMF.

To allow additional flexibility for IDE integrations, another option was introduced in the same release -- `org.gradle.tooling.parallel.ignore-legacy-default`. When set to `true` it disables the legacy default behavior, removing the connection between PE and PMF options, making them independent.

### Problem

While the introduction of new properties is a required measure for improving the PMF story in IDEA, it leads to added complexity on both sides: for Gradle as well as for IDEs. The root of that complexity is the legacy default behavior, where PE implies PMF.

Removing the legacy default behavior in a minor Gradle version is not appropriate, because many Android Studio users rely on `org.gradle.parallel` today to enable PMF and deliver their current IDE Sync performance. Disabling PMF can lead to significant performance regressions, especially for larger builds.

### Future considerations

In the future, both `org.gradle.parallel` and `org.gradle.parallel.tooling` can become absolute with the stabilization of Isolated Projects (IP). Due to additional constraints in place, IP can allow at least the same level of parallelism, while keeping compatible builds safe.

However, it will take time for IP to become stable, and even more time until the ecosystem and individual users can adopt it. So we should probably find a way to improve the situation until then.

### Proposal

We can make `org.gradle.parallel` and `org.gradle.parallel.tooling` independent options in a major Gradle release.

However, before we do that, we should warn users about the change.

One strategy we can apply is to **deprecate not being explicit about the PMF option if PE option is enabled.**

[...]

PR #38470, opened 2026-07-09
Deprecate implicit opt-in into parallel model building

### LINKED ISSUE #36001: Deprecate implicit opt-in into Parallel Tooling Model Building
Context:
- https://github.com/gradle/gradle/issues/35996

---

We can make `org.gradle.parallel` and `org.gradle.parallel.tooling` independent options in a major Gradle release.

However, before we do that, we should warn users about the change, and do it much in advance, given the breadth of the adoption for the `org.gradle.parallel` property.

One strategy we can apply is to deprecate not being explicit about the PMF option if PE option is enabled.

In other words, starting from some Gradle version, we issue a deprecation, when we detect `org.gradle.parallel=true` in the build, while no explicit value for `org.gradle.parallel.tooling` is provided. To avoid the deprecation, all the users have to do is either enable or disable the PMF property explicitly.

reinsch82:
Fixes #36001

### Context

Parallel model building (used by the Tooling API, most notably during IDE sync) was historically enabled implicitly whenever parallel execution (`org.gradle.parallel` / `--parallel`) was enabled. Gradle 9.4.0 (#35948) introduced the dedicated `org.gradle.tooling.parallel` property, which defaults to the value of `org.gradle.parallel` when unset to preserve backward compatibility.

Per #35996, we intend to make the two properties independent in Gradle 10, and to make it an error to enable `org.gradle.parallel` without an explicit `org.gradle.tooling.parallel` during model building (so users don't silently lose IDE-sync performance on upgrade). This PR is the deprecation that must precede that change.

**What it does:** when a build action builds models in Vintage mode with `org.gradle.parallel` enabled and `org.gradle.tooling.parallel` not set explicitly, Gradle emits a deprecation warning advising the user to set `org.gradle.tooling.parallel` explicitly. The remedy (setting the property) both silences the warning and pre-empts the Gradle 10 error.

**Trigger scope:** only model-building (Tooling API) invocations are affected — plain `--parallel` CLI builds never nag. The nag fires regardless of the transitional `org.gradle.tooling.parallel.ignore-legacy-default` system property, because that property will no longer be honored in Gradle 10 (it becomes redundant once the properties are independent). This is documented in the upgrade guide.

**Implementation note:** the deprecation is emitted from `RootBuildLifecycleBuildActionExecutor` (after the `DeprecationLogger` is initialized) rather than during build-model-parameter computation, where the logger is not yet available.

Relevant issues:
- #36001 (this deprecation)
- #35996 (the eventual Gradle 10 behavior change)
- #35948 (introduced `org.gradle.tooling.parallel`)

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### Decisions made (please confirm)

- **The nag fires regardless of `org.gradle.tooling.parallel.ignore-legacy-default`.** This is intentionally broader than the "narrowest case" sketched in #36001 (which would exempt the flag and thereby spare IntelliJ 2026.1.1+ users). Rationale: it mirrors the intended Gradle 10 error, so no one passes clean in 9.x and then breaks on upgrade. This assumes the Gradle 10 error fires even when the flag is set — please confirm.
- **`org.gradle.tooling.parallel.ignore-legacy-default` will no longer be honored in Gradle 10** (documented in the upgrade guide). It is not separately nagged: it is IDE-passed and undocumented, so a runtime nag would spam IDE syncs for something users did not set.
- **End state = require explicitness → error in Gradle 10** (alllex's option 2 from #36001), which is what `willBecomeAnErrorInGradle10()` encodes.

### Open questions

- **Exact Gradle 10 error semantics**, especially the interaction with `ignore-legacy-default` — this is the keystone for the first decision above.
- **Does a standard deprecation warning actually surface to Android Studio sync users?** They are the primary audience, and IDE sync UIs often swallow end-of-build deprecation warnings. If not, a Problems-API report or a louder channel may be needed (alllex's "louder / in more contexts" point).
- **IDE coordination:** how/when IntelliJ stops passing `ignore-legacy-default`, and whether Android Studio should start writing an explicit `org.gradle.tooling.parallel` rather than the nag landing on end users.

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
@@ -70,6 +70,9 @@ object BuildModelParametersProvider {
      *
      * It exists as a transitionary measure to allow IDEs to effectively require a separate user opt-in
      * into parallel model building via the explicit `org.gradle.tooling.parallel` property.
+     *
+     * This is transitional: once `org.gradle.parallel` and `org.gradle.tooling.parallel` become
+     * independent in Gradle 10, this property is redundant and will no longer be read (it becomes a no-op).
      */
     @JvmStatic
     val parallelModelBuildingIgnoreLegacyDefault = "org.gradle.tooling.parallel.ignore-legacy-default"
--- a/platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/RootBuildLifecycleBuildActionExecutor.java
+++ b/platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/RootBuildLifecycleBuildActionExecutor.java
@@ -120,6 +120,10 @@ public class RootBuildLifecycleBuildActionExecutor {
      * This applies only to Vintage model building, which is the only mode where the implicit
      * {@code org.gradle.parallel} -> parallel model building link exists.
      * <p>
+     * "Vintage" refers to the resolved {@link BuildModelParameters} mode, not the user-facing opt-in:
+     * model building with Configuration Cache enabled currently falls back to Vintage (models cannot be
+     * cached yet), so such builds are in scope of this deprecation. Isolated Projects model building is not.
+     * <p>
      * The nag is intentionally emitted regardless of the {@code org.gradle.tooling.parallel.ignore-legacy-default}
      * system property: in a future major, having {@code org.gradle.parallel} enabled without an explicit value
      * for {@code org.gradle.tooling.parallel} during model building will be an error, so users relying on the
--- a/platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/RootBuildLifecycleBuildActionExecutor.java
+++ b/platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/RootBuildLifecycleBuildActionExecutor.java
@@ -21,6 +21,7 @@ import org.gradle.api.internal.BuildDefinition;
 import org.gradle.api.internal.StartParameterInternal;
 import org.gradle.api.logging.configuration.ShowStacktrace;
 import org.gradle.api.problems.internal.ProblemsInternal;
+import org.gradle.initialization.StartParameterBuildOptions;
 import org.gradle.internal.build.BuildStateRegistry;
 import org.gradle.internal.build.RootBuildState;
 import org.gradle.internal.buildtree.BuildActionRunner;
@@ -93,6 +94,7 @@ public class RootBuildLifecycleBuildActionExecutor {
             try {
                 initDeprecationLogging(startParameter);
                 maybeNagOnDeprecatedJavaRuntimeVersion();
+                maybeNagOnImplicitParallelModelBuildingOptIn(startParameter);
                 RootBuildState rootBuild = buildStateRegistry.createRootBuild(BuildDefinition.fromStartParameter(startParameter, null));
                 return rootBuild.run(buildController -> buildActionRunner.run(action, buildController));
             } finally {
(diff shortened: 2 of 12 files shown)


chrispaulwu:
你好，邮件已收到，我将在空余时间尽快给你回复，谢谢！

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
```diff
@@ -70,6 +70,9 @@ object BuildModelParametersProvider {
      *
      * It exists as a transitionary measure to allow IDEs to effectively require a separate user opt-in
      * into parallel model building via the explicit `org.gradle.tooling.parallel` property.
+     *
+     * This is transitional: once `org.gradle.parallel` and `org.gradle.tooling.parallel` become
+     * independent in Gradle 10, this property is redundant and will no longer be read (it becomes a no-op).
```
jbartok, line 75: ❓ Just curious: why no-op and not error (if set explicitly)?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r980/ParallelModelBuildingDeprecationCrossVersionSpec.groovy
```diff
@@ -0,0 +1,69 @@
...
+        succeeds { connection ->
+            connection.model(GradleProject).withArguments(["--parallel"] + args).get()
+        }
+
+        where:
+        // The deprecation must fire regardless of ignore-legacy-default, because in Gradle 10 having
+        // 'org.gradle.parallel' enabled without an explicit 'org.gradle.tooling.parallel' will be an error.
+        args << [
+            [],
+            ["-Dorg.gradle.tooling.parallel.ignore-legacy-default=true"],
```
jbartok, line 46: ❓ Just for completeness, I would throw in the `false` case too, because, why not?

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r980/ParallelModelBuildingDeprecationCrossVersionSpec.groovy
```diff
@@ -0,0 +1,69 @@
...
+package org.gradle.integtests.tooling.r980
+
+import org.gradle.integtests.tooling.fixture.TargetGradleVersion
+import org.gradle.integtests.tooling.fixture.ToolingApiSpecification
+import org.gradle.tooling.model.GradleProject
+import spock.lang.Issue
+
+@Issue("https://github.com/gradle/gradle/issues/36001")
+@TargetGradleVersion("current")
+class ParallelModelBuildingDeprecationCrossVersionSpec extends ToolingApiSpecification {
```
jbartok, line 26: 🤔 Would be good to also have a test asserting that a non-model build with `--parallel` (a plain CLI build, or a TAPI task run) does not nag.

platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/exec/RootBuildLifecycleBuildActionExecutor.java
```diff
@@ -111,6 +113,34 @@ private void initDeprecationLogging(StartParameterInternal startParameter) {
         DeprecationLogger.init(startParameter.getWarningMode(), eventEmitter, problemsService, problemsStream);
     }
 
+    /**
+     * Deprecates relying on the legacy default where parallel model building (used by the Tooling API,
+     * e.g. during IDE sync) is derived from the {@code org.gradle.parallel} property.
+     * <p>
+     * This applies only to Vintage model building, which is the only mode where the implicit
```
jbartok, line 120: 🤔 Worth clarifying the scope of "Vintage" here, since it's the resolved `BuildModelParameters` mode rather than "the user didn't enable CC/IP." A model build with `org.gradle.configuration-cache=true` still falls back to vintageMode (CC can't cache models yet), so `isVintage()` returns `true` and this nag fires even for CC-enabled projects — which makes the practical reach "any org.gradle.parallel=true project on IDE sync," CC or not. 

Consider adding a sentence noting that CC-with-models degrades to vintage and is therefore in scope, while Isolated Projects is not.

jbartok (approved):

lkasso (approved):
Added a few more refs in the docs. LGTM now.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/buildtree/control/BuildModelParametersProvider.kt
reinsch82: it is a decision we can take. one could pose the question: why is a property that is not used anymore different from one that never existed in the first place. We'd have code checking for not used properties lingering around forever?

platforms/documentation/docs/src/docs/userguide/optimizing-builds/isolated_projects.adoc
```diff
@@ -201,8 +201,8 @@ However, both Gradle and the IDE need to be configured to allow it.
 
 On the Gradle side, parallel model building must be allowed in the build definition.
 It is *automatically allowed with Isolated Projects*.
-It can also be <<performance.adoc#sec:configure_tooling_api_actions_parallelism,allowed without Isolated Projects>>
-via `org.gradle.tooling.parallel` or `org.gradle.parallel` Gradle properties.
+It can also be <<performance.adoc#sec:configure_tooling_api_actions_parallelism,allowed without Isolated Projects>> by setting the `org.gradle.tooling.parallel` Gradle property to `true`.
+Historically it was also enabled implicitly by `org.gradle.parallel`, but relying on that default is <<upgrading_version_9.adoc#deprecate_implicit_parallel_model_building, deprecated>>.
```
eskatos: 💅  I would not state the past in the reference manual. This content will rot over time.

platforms/documentation/docs/src/docs/userguide/optimizing-builds/performance.adoc
```diff
@@ -194,6 +194,10 @@ At the same time, it is rare for builds to have problems with tasks _and_ Toolin
 Since Gradle 9.4.0, there is an additional option `org.gradle.tooling.parallel`
 that allows controlling parallelism of Tooling API actions independently of task execution parallelism.
 
+NOTE: Relying on `org.gradle.parallel` alone to enable parallel model building is deprecated and will become an error in Gradle 10.
```
eskatos: 💅 Same comment. The reference user manual should explain how to set things now, not explain the history of things.

platforms/documentation/docs/src/docs/userguide/reference/runtime-configuration/build_environment.adoc
```diff
@@ -481,6 +481,8 @@ Allows running <<service_injection.adoc#toolingmodelbuilderregistry,Tooling Mode
 An explicit value will override `org.gradle.parallel` in this context.
 +
 _Default is the value of `org.gradle.parallel`._
+Relying on that default during model building is deprecated and will become an error in Gradle 10.
```
eskatos: 💅 Same comment again.

eskatos (approved):
LGTM code wise

I think the documentation changes are not ideal. The reference user manual should explain what flags are available and what they do. Not the history of them. This might require some rephrasing instead of just adding the deprecation in several pages. This should be addressed before merging.

The deprecation is surfaced to IDEs in the exact same way as for any deprecation of API/features/flags that may be used only from Tooling Builders. This is enough for this change, I would not treat this deprecation as special in this regard.

Issue #22879, opened 2022-11-28
Lift some execution time restrictions imposed by the configuration cache on accessing methods and variables defined in build script


mlopatkin:
Consider the following task, defined in `build.gradle`:
```gradle
def dir = file('data')

def listFiles(File dir) {
    dir.listFiles({ file -> file.isFile() } as FileFilter).name.sort()
}

tasks.register('listFiles') {
    doLast {
        println listFiles(dir)
    }
}
```
### Expected Behavior
Calling a method defined in the build script should be allowed at execution time if the method doesn't call `Project` methods. Accessing variables defined in the build script should also be allowed if these values are serializable by the configuration cache.

### Current Behavior
In case of the Groovy DSL, the configuration cache is safely saved, but the next run fails to run the serialized task, often with the "method not found" error. This behavior comes from #20126. Accessing script-defined variables works fine.

In case of the Kotlin DSL, the configuration cache entry cannot be serialized, because the captured script reference cannot be serialized. An instance of this is filed under #16080.

### Context
Extracting methods is a useful feature to break up/reuse pieces of build script logic, recommended in the [Gradle user guide](https://docs.gradle.org/nightly/userguide/tutorial_using_tasks.html#sec:using_methods).

The workaround is to use utility classes/objects as containers for these methods. This approach is, however, problematic for the variables in the Kotlin DSL, as these variables cannot be initialized with project-derived values at configuration time.

mlopatkin:
Some of the approaches we're considering:
* Substitute a "broken" object for script reference that allows to call methods defined in the script itself, but not the method that delegate to e.g. `Project` or `Settings`. We do something similar for Groovy, but the "broken" object is too broken to e.g. call methods on it.
* Use a compiler plugin to "staticalize" methods and variables that do not depend on `Project`.
* Hack Kotlin script compilation to consider the script body a method rather than a special thing.

TheGoesen:
I would argue that another way to look at it would be to say that currently local variables are very often actually not local variables and even the user guide explicitly lies about this:
https://docs.gradle.org/current/userguide/writing_build_scripts.html#sec:local_variables

```
val dest = "dest" //this is actually not a local variable dest. this creates a property dest for the script object-> it cannot be used in a doLast block

tasks.register<Copy>("copy") {
    from("source")
    into(dest)
}
```

```
if (true){
     val dest = "dest" //now this is actually a local variable dest which can be used in a doLast block/execution time

    tasks.register<Copy>("copy") {
[...]

Issue #16080, opened 2021-02-09
Task.onlyIf is hardly compatible with Configuration Cache, due to kotlin lambda compilation


stephanenicolas:
Using Gradle 6.8+, consider a plugin:

```
 fun apply(project: Project) {
   val b: Boolean = project.hasProperty("foo")
    project.tasks.register("task", MyTask::class.java) {
       it.onlyIf { b }
    }
 }
```
because of the way kotlin compiles lambdas, it will create a lambda that contains a reference to the outer scope of the `onlyIf` lambda, which will include the plugin instance. This will most often break as plugins are allowed to use non serializable types such as projects, tasks, etc. 
See : https://stackoverflow.com/q/44538627/693752

as a workaround, one can create a subclass of `Spec` that accepts a boolean as a constructor parameter and pass an instance of this class to `onlyIf`, so that there is no leak of the plugin instance.

To solve this, it would be nice to be able to provide an overload of onlyIf that would accept a provider of boolean.

stephanenicolas:
Actually, another workaround is to do 
```
project.tasks.register("task", MyTask::class.java) {
       val b: Boolean = it.project.hasProperty("foo")
       it.onlyIf { b }
    }
```
So that the lambda is more self contained. 

But yes, definitely `onlyIf` could accept a provider.

eskatos:
Note that this workaround above will miss a build logic input. It should be:

```kotlin
project.tasks.register<MyTask>("task") {
    val foo = project.providers.gradleProperty("foo")
    onlyIf { foo.isPresent }
}
```

:+1: on having `onlyIf(Provider<Boolean>)`.

tasomaniac:
Hi folks, I hit this problem too and would appreciate any feedback if my workaround is ok. The last workaround is better but only applicable to `property` usage.

In my case, I was accessing a boolean property from my extension. I resolved it by assigning the extension to a "redundant" local variable. https://github.com/GradleUp/auto-manifest/pull/14

h0tk3y:
To be more precise, the scenario that this issue affects is when the lambda that is SAM-converted to a `Spec` (or another interface) and stored in the configuration cache captures the properties of the enclosing class that reference non-serializable model objects.
 
It does not reproduce under the following conditions:
* If the lambda does not reference the properties of any class – e.g. if it only references local variables of the method where it is declared, which is one of the workarounds for the issue;
* If the enclosing class does not attempt to store the references to non-serializable model objects, for example:
[...]

PR #38485, opened 2026-07-10
Scrub Kotlin & Groovy scripts for the configuration cache

h0tk3y:
Task actions and other lambdas in a Gradle script capture the
whole compiled script instance. The configuration cache rejected it as
a Gradle script object or fails at runtime, so any task action that touched a script val or fun failed to store (#22879, #16080).

Serialize the script instead of rejecting it. Keep its
Project-independent state - user val/var/by lazy and the base-class
service delegates - and drop the bridge to the build model. Build-model
references (Project/Settings/Gradle) and the script host are replaced on
load with broken proxies that report a clear problem when touched at
execution, mirroring the Groovy BrokenScript. Services (logger, file
operations, ...) keep working - they are re-resolved from the isolate
owner.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/core-configuration/core-serialization-codecs/docs/script-scrubbing-spec.md
```diff
@@ -0,0 +1,190 @@
...
+for something that reads fine and has no inherent problems.
+
+## Functional design
+
+Both codecs make the same moves:
+
+1. Keep the compiled script instance rather than rejecting it.
+1. Sever the build model — replace known references to from the script object it with a "broken object" stand-in, similar to the "broken script" object.
+   These broken objects are proxies that report a clear CC problem and throw on any method invocation
+1. Keep the script's own state and ambient services working (`val`/`var`, `file()`, `logger`). The services cannot be accessed from the broken objects,
```
mlopatkin, line 55: ❓ What if the script has a `val` with something that holds an unsupported type inside? E.g.
```kotlin
class Foo(val project: Project)

val foo = Foo(project) // Bean codec will fail
val groupProvider = provider { project.group }  // legit
tasks.register("bar") {
  doLast { println("group = ${groupProvider.get()}") }
}
```

There `doLast` captures the script instance, and we'll try to serialize `foo`. Eventually, we'll fail.

h0tk3y: Hmm... I don't see why anything should change in the example you provided. I'd say it's different from the script reference captured by the lambdas. In your example, the `Project` reference is explicitly captured in the object stored in the task, which is similar to explicitly putting it into a field of the task. 

So my answer is that we should still fail to serialize it, because it's by no means an implicit usage of the script object, it's an explicitly stored reference to the project. 

Am I missing something?

mlopatkin: > So my answer is that we should still fail to serialize it, because it's by no means an implicit usage of the script object, it's an explicitly stored reference to the project.

The task here only needs a benign provider reference, but due to the way capturing works it captures the whole script instance. But the mere presence of `val foo`, even if it is unused at execution time at all, makes the script instance non-serializable, so the user would be forced to go back to workarounds. We cannot scrub it off, because we only do that for fields we own, not user-defined ones. At least that's how I understand the current approach.

h0tk3y: UPD: I guess your question is about the field capturing the project reference being an unused one. Yes, my design does not cover picking only the fields used by the task state. If the script object is captured, all its field will get serialized.

I see how it might be undesirable, if you have multiple vals at the top level and you change one lambda to start using the script object, which will now lead to all the fields getting into the serialized state (or failing).

It would be good to improve the granularity, also taking into account that different tasks might capture different sets of the script's fields. However, doing it seems to need one of:
* statically analyzing the links between classes to find out which fields are used by the task state or anything that it incokes; or
* scoping the serialization of the subgraph reachable from the script object's fields in a way that it stores "broken objects" in place of Project (and other model) references reachable from the script object's fields.
* bytecode manipulation on the compiled script's classes, probably separating the fields into several nested classes and making the lambdas only capture the classes having the fields they are using?

Do you have any thoughts on the feasibility of this? Or other approaches?

PR #38494, opened 2026-07-11
Update Documentation: clarify project property resolution

### LINKED ISSUE #35979: Document exact resolution order of properties and extra properties
### Context

There are many ways to obtain the values of "properties" in Gradle, and these properties can come from many different sources. What makes the situation more complicated is that each way of accessing may use a different look-up chain to determine the value of the property.

This has created a lot of confusion and might have been considered bugs at the beginning, but they have now become legacy.

See:
- https://github.com/gradle/gradle/issues/23572
- https://github.com/gradle/gradle/issues/29600

### Problem

At the same time, the public documentation for these is either missing the explanation of the look-up chains or that explanation does not exactly match the implementation, causing confusion when users need to understand the subtleties of the contract.

We should document in detail at least these in the javadocs / user guide:

- `Project.findProperty()` and friends
- `ExtensionContainer.getExtraProperties()`
- `Project.extra` (Kotlin DSL extension)
- `ProviderFactory.gradleProperty()`

### Pointers

The resolution of "Gradle properties" comprises a big chunk of the look-up chain in all cases. It already has detailed internal documentation, which should be made public:
https://github.com/gradle/gradle/blob/50bfd96817f175abee4527ab8142004d70cece95/subprojects/core/src/main/java/org/gradle/api/internal/properties/GradlePropertiesController.java#L66-L75

Note, there are subtleties when it comes to resolving project's "Gradle properties" as opposed to build's "Gradle properties".

This [part-discovery spec on Dynamic Groovy resolution](https://docs.google.com/document/d/1ID_daltidUDkGcUw0-vThu8btRZ2IWC2E2TcOV_QfZU) explains the dynamic look-up chain that is also applicable to the `Project.findProperty()` lookups.

qiu-tiandev:
Fixes #35979

## What changed

Clarifies the documented resolution order for project properties, Gradle properties, and extra properties.

## Why

The previous documentation mixed build-level Gradle-property precedence with the dynamic project-property lookup APIs and described an outdated convention-object order.
It also used an invalid `gradle.properties` example for a system property.

## Impact

Build authors can now choose the appropriate API and property source with a clear, accurate reference.
Kotlin DSL extra-property access is distinguished from the broader project-property lookup.

## Validation

- `git diff --check`
- `:docs:tasks`

### DIFF
--- a/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/ExtraPropertiesExtensions.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/ExtraPropertiesExtensions.kt
@@ -31,6 +31,11 @@ import kotlin.reflect.KProperty
 /**
  * The extra properties extension in this object's extension container.
  *
+ * For a [org.gradle.api.Project], this accesses only that project's extra properties. Values set explicitly
+ * on the extension take precedence over Gradle properties with the same name when those properties are not mapped
+ * directly to a `Project` property. This does not perform the broader lookup used by
+ * [org.gradle.api.Project.findProperty].
+ *
  * @see [ExtensionContainer.getExtraProperties]
  */
 val ExtensionAware.extra: ExtraPropertiesExtension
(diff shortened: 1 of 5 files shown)


lkasso (approved):
LGTM. I applied some minor cleanup and will now merge.

PR #38502, opened 2026-07-13
Add @since check for non-public classes

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic/binary-compatibility/src/main/groovy/gradlebuild.binary-compatibility.gradle
+++ b/build-logic/binary-compatibility/src/main/groovy/gradlebuild.binary-compatibility.gradle
@@ -16,6 +16,7 @@
 
 
 import gradlebuild.EnrichedReportRenderer
+import me.champeau.gradle.japicmp.report.GroovyReportRenderer
 import gradlebuild.basics.GradleModuleApiAttribute
 import gradlebuild.basics.PublicApi
 import gradlebuild.basics.PublicKotlinDslApi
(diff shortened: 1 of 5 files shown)


ov7a:
There might be a caveat when a non-public api with `@since` becomes public. IIUC, the method will be checked by the public api checker, and the since annotation will be forced to be updated

blindpirate (commented):
Moving internal packages around is something we do pretty often — precisely because they're internal, so there's no compatibility pressure holding them in place. To japicmp a move is remove + NEW, so the moved class gets re-checked, its preserved `@since` no longer matches the current version, and the build breaks. The only fix is to rewrite the tag to the version of the move, which makes it less accurate than it was.

What makes this worse is that there's no opt-out. For public API changes we have `accepted-public-api-changes.json`; here there's nothing — `reject()` bypasses `acceptOrReject()` when sinceRequired is false, and no `AcceptedRegressionsRule` is registered.

How do you plan to handle this?

ov7a:
🤔 Good point. 

The idea was that if there's a `@since` annotation, then it should be correct. And if it's incorrect, it's easy to fix directly in the code (or just remove it). Going via accepted api changes would be a hassle. However, I missed the package move case. 

Technically, the annotation should be updated (because it's under a new package and the api is different). At this point, the question arises, what are the scenarios for recording `since`? If it's just informational (this method EXISTS since x.x.), it's one thing, but if it's API-like (this method CAN BE USED as-is since x.y. by internal plugins/DV), it should be a bit stricter. Maybe the way out is multiple `@since`?

ov7a:
After discussing this internally, we've decided that capturing the "EXISTS since" part is not worth it. Also, the `@since` annotation on internal methods may hint that the method is used by internal plugins/DV.

The ideal situation is to have a public API isolated from the internals, and a clear protocol for DV. Plugins should rely on the public API. However, it's a much bigger goal.

Given the discussion proceeding with the current approach.

Issue #37409, opened 2026-04-02
Investigate making StartParameter read-only


mlopatkin:
Mutability of `StartParameter` type is problematic, especially considering that it is available through `Gradle.getStartParameter()` AND is read by internals to drive behavior.

Mutating start parameter is a semi-supported pattern, but today there are no guardrails, so applied configuration can be silently ignored, when it e.g. comes too late.

Known uses include:
* adding tasks to [always run them](https://github.com/JetBrains/intellij-community/blob/8e9b60096552c4eb0eae3ac9cf880299d8a839e1/plugins/kotlin/gradle/gradle-tooling/impl/src/org/jetbrains/kotlin/idea/gradleTooling/KotlinDslScriptTaskModelBuilder.kt#L27-L31)
* enabling build cache from a plugin.

`GradleBuild` task also uses a mutable StartParameter as its input, so there is a requirement to provide some way to build arbitrary StartParameter instances.

Moving towards readonly will likely involve deprecation and removal of setters and replacing them with a builder. We may also need replacements to serve existing use cases.

asodja:
FYI: IntelliJ is setting some tasks to run on their side when there are building models, see:
https://github.com/JetBrains/intellij-community/blob/5a5b8b369f0790ee0bcfa37eaab464cd9907d6f3/plugins/kotlin/gradle/gradle-tooling/impl/src/org/jetbrains/kotlin/idea/gradleTooling/KotlinDslScriptTaskModelBuilder.kt#L29-L31

lukebemish:
> adding tasks to [always run them](https://github.com/JetBrains/intellij-community/blob/8e9b60096552c4eb0eae3ac9cf880299d8a839e1/plugins/kotlin/gradle/gradle-tooling/impl/src/org/jetbrains/kotlin/idea/gradleTooling/KotlinDslScriptTaskModelBuilder.kt#L27-L31)

My greatest worry with making `StartParameter` read-only is making sure there's an alternative for this use case; I've seen a few plugins use this for IDE integration reasons -- it's the simplest, most-reliable, and most-compatible way that I and others have found to ensure that a generated dependency is correctly created at IDE sync, so that it's actually usable from within the IDE.

[...]

Issue #37408, opened 2026-04-02
Investigate deprecation of GradleBuild task


mlopatkin:
### Context

The `GradleBuild` task type allows users to invoke a Gradle build from within another Gradle build. While this was a convenient mechanism, it has proven to be problematic in practice.

This issue tracks the **investigation** into deprecating `GradleBuild` and eventually removing it.

### Problems with `GradleBuild`

**Poor isolation from the outer build**
`GradleBuild` runs the nested build inside the same JVM and Gradle daemon as the outer build. This means the two builds share classloaders, logging infrastructure, and other global state, leading to subtle and hard-to-diagnose failures.

**`StartParameter` API clash**
`GradleBuild` reuses the `StartParameter` of the outer build as a starting point and mutates it for the nested invocation. This creates conflicts between "invocation" parameters that belong to the parent build (e.g. parallel execution settings, max workers, dependency verification flags) and those intended for the nested build. Distinguishing between the two is fragile and error-prone.

**Maintenance burden**
Supporting an in-process nested build adds significant complexity to the Gradle codebase. Every change to build execution, caching, or daemon lifecycle must account for the `GradleBuild` code path, which is rarely tested in comparison to the normal execution path.

### Proposed direction

Drive users towards `Exec`-based invocations of the Gradle wrapper (`./gradlew`). This gives full process-level isolation and avoids all of the issues above at the cost of spawning a new daemon (which is the expected and well-tested execution model).

### Scope of this investigation

- [ ] Catalogue known use cases for `GradleBuild` in the ecosystem (community plugins, build scans, user reports).
- [ ] Identify edge cases where `Exec` is **not** a sufficient replacement and determine whether those gaps can be closed.
- [ ] Assess the impact on Gradle's own integration tests and internal usage.
- [ ] Propose a deprecation timeline and migration guidance (to be tracked in a follow-up issue).

### Known edge cases where `Exec` falls short

**Environment setup convenience** — `GradleBuild` automatically inherits the Gradle environment of the outer build (wrapper location, Gradle version, JVM args, etc.). With `Exec`, users must manually locate the wrapper script, forward relevant system properties, and ensure the nested invocation uses the correct Gradle distribution. This is boilerplate-heavy and error-prone, especially across platforms.

[...]

PR #38512, opened 2026-07-14
Store build cache flag in CC when it's enabled in settings

### LINKED ISSUE #37088: Changing the startParameter programatically in settings is not CC-compatible
### Current Behavior

Given 
```kotlin
gradle.startParameter.buildCacheEnabled = true
```
in settings, the buildCache is not enabled when CC is restored

### Expected Behavior

Either this method of enabling Build Cache (and other startParameter flags) should be supported with CC or it should be deprecated

### Context (optional)

_No response_

See this test
https://github.com/gradle/gradle/blob/e1d9eb448927badae1d8fec6115f3ffe1e9d6984/subprojects/core/src/integTest/groovy/org/gradle/caching/configuration/internal/BuildCacheCompositeConfigurationIntegrationTest.groovy#L42-L50

### Gradle version

latest

### Build scan URL (optional)

_No response_

_No response_

ov7a:
But also, depreciate setting buildCache via startParameter in settings

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
+++ b/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
@@ -820,11 +820,13 @@ public class StartParameter implements LoggingConfiguration, ParallelismConfigur
 
     /**
      * Enables/disables the build cache.
+     * <p>
+     * Calling this on the start parameter of a running build is deprecated: use the {@code org.gradle.caching}
+     * Gradle property instead. It is not annotated as deprecated because it remains the only way to configure the
+     * build a {@code GradleBuild} task runs.
      *
      * @since 3.5
-     * @deprecated Use {@code org.gradle.caching} Gradle property instead.
      */
-    @Deprecated
     public void setBuildCacheEnabled(boolean buildCacheEnabled) {
         onMutableCall("setBuildCacheEnabled(boolean)");
         this.buildCacheEnabled = buildCacheEnabled;
--- a/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
+++ b/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
@@ -112,7 +112,7 @@ public class StartParameter implements LoggingConfiguration, ParallelismConfigur
     private boolean offline;
     private File projectCacheDir;
     private boolean refreshDependencies;
-    protected boolean buildCacheEnabled;
+    private boolean buildCacheEnabled;
     private boolean buildCacheDebugLogging;
     private boolean configureOnDemand;
     private boolean continuous;
--- a/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
+++ b/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
@@ -112,7 +112,7 @@ public class StartParameter implements LoggingConfiguration, ParallelismConfigur
     private boolean offline;
     private File projectCacheDir;
     private boolean refreshDependencies;
-    private boolean buildCacheEnabled;
+    protected boolean buildCacheEnabled;
     private boolean buildCacheDebugLogging;
     private boolean configureOnDemand;
     private boolean continuous;
--- a/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
+++ b/platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
@@ -822,7 +822,9 @@ public class StartParameter implements LoggingConfiguration, ParallelismConfigur
      * Enables/disables the build cache.
      *
      * @since 3.5
+     * @deprecated Use {@code org.gradle.caching} Gradle property instead.
      */
+    @Deprecated
     public void setBuildCacheEnabled(boolean buildCacheEnabled) {
         onMutableCall("setBuildCacheEnabled(boolean)");
         this.buildCacheEnabled = buildCacheEnabled;
--- a/platforms/core-runtime/start-parameter/src/main/java/org/gradle/api/internal/StartParameterInternal.java
+++ b/platforms/core-runtime/start-parameter/src/main/java/org/gradle/api/internal/StartParameterInternal.java
@@ -108,11 +108,25 @@ public class StartParameterInternal extends StartParameter {
     }
 
     /**
+     * Runs the action with the mutation listener detached, so internal machinery can go through the public
+     * setters without its own mutations being reported as violations.
+     */
+    private void withoutMutationListener(Runnable action) {
+        Consumer<String> previous = this.mutationListener;
+        this.mutationListener = null;
+        try {
+            action.run();
+        } finally {
+            this.mutationListener = previous;
+        }
+    }
+
+    /**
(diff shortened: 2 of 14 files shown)


ov7a:
@bot-gradle test this

ov7a:
@bot-gradle test this

lkasso (approved):
Docs LGTM.

platforms/core-runtime/start-parameter/src/main/java/org/gradle/api/internal/StartParameterInternal.java
```diff
@@ -106,6 +107,28 @@ protected void onMutableCall(String methodSignature) {
         }
     }
 
+    /**
+     * Sets build cache enablement from a source other than user build logic
+     */
+    public void setBuildCacheEnabledInternal(boolean buildCacheEnabled, boolean configuredByBuildLogic) {
+        this.buildCacheEnabled = buildCacheEnabled;
```
abstratt: ❓ 

Instead of making this field protected (and exposing it to users/javadoc) just so you could avoid the `onMutableCall` check, did you consider adding a way to temporally disable the check (like a `withoutMutationListener(Runnable)` method) and then be able to call the regular setter?

That is possibly going to be useful for future needs to override the base setters.

platforms/core-runtime/start-parameter/src/main/java/org/gradle/api/internal/StartParameterInternal.java
```diff
@@ -106,6 +107,28 @@ protected void onMutableCall(String methodSignature) {
...
+        this.buildCacheEnabledConfiguredByBuildLogic = configuredByBuildLogic;
+    }
+
+    @Override
```
abstratt, line 134: ❓ 

Shouldn't `StartParameter.setBuildCacheEnabled()` be marked as `@Deprecated`?

abstratt (dismissed):
I would rather we didn't add a protected field to public API class.

platforms/core-runtime/start-parameter/src/main/java/org/gradle/api/internal/StartParameterInternal.java
ov7a: Good point, thanks!

platforms/core-runtime/start-parameter/src/main/java/org/gradle/internal/deprecation/StartParameterDeprecations.java
```diff
@@ -36,6 +36,14 @@ public static void nagOnStartParameterCopy(String methodWithParams) {
             .nagUser();
     }
 
+    public static void nagOnSetBuildCacheEnabled() {
+        DeprecationLogger.deprecateMethod(StartParameter.class, "setBuildCacheEnabled(boolean)")
+            .withAdvice("Use the 'org.gradle.caching' Gradle property to enable the build cache instead.")
```
abstratt:
```suggestion
            .withAdvice("Use the 'org.gradle.caching' Gradle property to enable or disable the build cache instead.")
```

platforms/core-runtime/start-parameter/src/main/java/org/gradle/api/internal/StartParameterInternal.java
```diff
@@ -106,6 +107,45 @@ protected void onMutableCall(String methodSignature) {
         }
     }
 
+    /**
+     * Runs the action with the mutation listener detached, so internal machinery can go through the public
+     * setters without its own mutations being reported as violations.
+     */
+    private void withoutMutationListener(Runnable action) {
+        Consumer<String> previous = this.mutationListener;
+        this.mutationListener = null;
```
abstratt: 💭 I think listener disablement should be done via a static ThreadLocal<Boolean> or else multiple threads touching this object may affect each other.

abstratt:
❓ 

@ov7a  I was thinking of `GradleBuild`'s `startParameter`. Do we want to issue deprecation for that as well? 
(we do now)

ov7a:
@abstratt I think it's out of scope for this PR. 

This problem is tracked in
* https://github.com/gradle/gradle/issues/37409

abstratt:
@ov7a Thanks for the pointer. However, considering the changes we are making here, we are going to be issuing deprecations for `StartParameter` being used as a configuration knob in `GradleBuild`. It seems to me that in the case of `GradleBuild`, mutating the build configuration before it is started is benign. Or are we mandating folks rely exclusively on `gradle.properties` for that use case as well? (which would seem excessive)

```
    def "mutating the start parameter of a GradleBuild task should not issue a deprecation"() {
        buildFile << """
            tasks.register('nested', GradleBuild) {
                dir = file('other')
                tasks = ['customTask']
                startParameter.buildCacheEnabled = true
            }
        """
        file("other/settings.gradle") << """
            buildCache {
                local {
                    directory = '$cacheDir'
                }
            }
        """
        file("other/build.gradle") << customTaskCode()

        expect:
        // TODO there should be no deprecation in this case
        executer.expectDocumentedDeprecationWarning(SET_BUILD_CACHE_ENABLED_DEPRECATION)
        succeeds("nested")
        !localBuildCache.empty
    }

```

It seems we should somehow exempt setting `GradleBuild.startParameter.buildCacheEnabled` from deprecations.

ov7a:
@abstratt I didn't think about that, thanks for pointing this out and for providing a test.

I pushed a limited fix to address that. There's a caveat that if a nested build mutates startParameter.buildCache in its settings script, then the deprecation is not triggered. I think this is acceptable given that we are considering the deprecation of the GradleBuild task:
* https://github.com/gradle/gradle/issues/37408

PTAL

abstratt:
Thanks, Vlad! One last thing - with the `GradleBuild` scenario in mind, I think it was bad advice to ask you to add a `@Deprecated` annotation, since the deprecation can only be established at runtime.

abstratt (approved):
LGTM, but I'd undo the bit around the `@Deprecated` annotation

ov7a:
FTR, with `@Deprecation`, the setting script is marked correctly, but the problem manifests in the configuration of `GradleBuild` task because it reuses the `startParameter` class verbatim instead of using a custom configuration model. 

This is a tradeoff between two options: either be helpful for most users, but misleading for `GradleBuild` task users, or be silent about deprecation on a source level, but be correct for `GradleBuild` scenarios. Despite a presumably small user base of the `GradleBuild` task, we decided to be conservative. Thus, source-level annotation was removed.

platforms/core-runtime/start-parameter/src/main/java/org/gradle/StartParameter.java
```diff
@@ -820,6 +820,10 @@ public boolean isBuildCacheEnabled() {
 
     /**
      * Enables/disables the build cache.
+     * <p>
+     * Calling this on the start parameter of a running build is deprecated: use the {@code org.gradle.caching}
+     * Gradle property instead. It is not annotated as deprecated because it remains the only way to configure the
```
alllex, line 825: 💅  In the `It is not annotated as ...` the `It` probably refers to the current method. But English-wise it reads like `It` refers to the property from the previous sentence.

PR #38515, opened 2026-07-14
Improve ServeDocs port conflict handling

tresat:
Previously, the `ServeDocs` task would report success even if its requested port was already in use, leading to a non-functional server and a confusing user experience.

This change introduces a synchronous pre-flight check for port availability. If the port is occupied, the task now fails proactively with a `GradleException`, instead of reporting a successful build.

The error message provides helpful suggestions for alternative free ports, streamlining problem resolution. It also exposes the port via a `--port` option.

```

* What went wrong:
Execution failed for task ':docs:serveDocs' (registered by plugin 'gradlebuild.documentation').
> Port 8000 is already in use.
  Try running with a different port, e.g.:
    ./gradlew :docs:serveDocs --port 8100
```

### DIFF
--- a/build-logic/documentation/src/main/groovy/gradlebuild/docs/ServeDocs.java
+++ b/build-logic/documentation/src/main/groovy/gradlebuild/docs/ServeDocs.java
@@ -17,6 +17,7 @@
 package gradlebuild.docs;
 
 import org.gradle.api.DefaultTask;
+import org.gradle.api.GradleException;
 import org.gradle.api.file.DirectoryProperty;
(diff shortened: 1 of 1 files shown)


tresat:
I kept forgetting other things were running on 8000, and it was annoying to remember to find them and shut them down first.  This makes it easy to use an alternate port.

lkasso:
So much better than always trying to run on 8000! Thanks @tresat!

lkasso:
I tested this, and it worked well for me.

lkasso (approved):

PR #38519, opened 2026-07-14
Stop publishing SHA-256/SHA-512 checksums for Maven and Ivy

big-guy:
## What this does

Maven and Ivy publications now publish only the checksums that repositories universally expect:

- **Maven**: `.sha1` + `.md5`
- **Ivy**: `.sha1`

The `.sha256` and `.sha512` checksum files are no longer published for any artifact (including `.asc` signatures and `maven-metadata.xml`).

### Why

These extra checksums were rarely consumed and were frequently rejected by repositories that do not recognize them, adding noise without a corresponding benefit. Consuming builds are unaffected — Gradle has never required SHA-256 or SHA-512 checksums to be present in order to resolve or verify artifacts.

### Notes

- The internal, undocumented `org.gradle.internal.publish.checksums.insecure` system property (which previously suppressed these extra checksums) is removed, as it no longer has any effect.
- Only two code paths generated these files: `AbstractMavenPublisher` (Maven) and `ExternalResourceResolver` (Ivy). Both are updated.

## Changes

- Production: `AbstractMavenPublisher`, `ExternalResourceResolver`
- Test fixtures: `extraChecksums`/`withExtraChecksums` now default to `false`; the machinery is retained so cross-version tests can still model older Gradle distributions that do publish the extra checksums.
- Cross-version `r47` test: expects extra checksums only for target versions in `[6.0, 9.7.0)`.
- Removed hardcoded SHA-256/SHA-512 expectations across HTTP/HTTPS/S3/GCS/SFTP/Console/Issues/Retries/ConfigurationCache publish tests.
- Docs: upgrade-guide and release-notes entries.

## Testing

Ran with `embeddedIntegTest` + test distribution, all green:

- Maven + Ivy HTTP publish tests
- Basic / Java / Console / Issues publish tests
- Signing publications, ConfigurationCache, Retries, HTTPS, SFTP publish tests
- Cross-version tooling-api test source compiles

S3/GCS integ tests were not run locally (require emulators); those edits are mechanical expectation removals over the same transport-agnostic code path exercised by the HTTP tests.


platforms/software/maven/src/integTest/groovy/org/gradle/api/publish/maven/MavenPublishHttpIntegTest.groovy
```diff
@@ -463,39 +422,31 @@ class MavenPublishHttpIntegTest extends AbstractMavenPublishIntegTest {
     private static void expectPublishModuleWithCredentials(MavenHttpModule module, HttpServer.PasswordCredentials credentials) {
         module.artifact.expectPut(credentials)
         module.artifact.sha1.expectPut(credentials)
-        module.artifact.sha256.expectPut(credentials)
```
jvandort, line 466: We removed many/most callers of these `sha256/512` methods. Is anyone else calling them now or can we delete the getters?
https://github.com/gradle/gradle/blob/1439b8dbe998010e410b4926acbb9bfcc9e6a4fd/testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/resource/RemoteArtifact.java#L26-L28

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpArtifact.groovy
```diff
@@ -78,7 +78,7 @@ abstract class HttpArtifact extends HttpResource implements RemoteArtifact {
         }
     }
 
-    void expectPublish(boolean extraChecksums = true, HttpServer.PasswordCredentials credentials = null) {
+    void expectPublish(boolean extraChecksums = false, HttpServer.PasswordCredentials credentials = null) {
```
jvandort: Is anyone calling this with true now? Or can we remove this parameter and the if statement below?

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/maven/AbstractMavenModule.groovy
```diff
@@ -56,7 +56,7 @@ abstract class AbstractMavenModule extends AbstractModule implements MavenModule
     private final List dependencies = []
     private Map<String, ?> mainArtifact = [:]
     private final List artifacts = []
-    private boolean extraChecksums = true
+    private boolean extraChecksums = false
```
jvandort: I am not sure anyone calls `withExtraChecksums()` on the ivy or maven modules. I think we can strip this all out since I think this is always false now

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/server/http/HttpArtifact.groovy
jvandort: It looks like our cross version tests do. Though, I would rather inline this logic there if this is the only place that verifies sha256/512 publishing -- or just don't verify that these files get published at all in the CV tests

jvandort (approved):

testing/internal-distribution-testing/src/main/groovy/org/gradle/test/fixtures/maven/AbstractMavenModule.groovy
big-guy: Yep, I was stripping these out too.

platforms/software/maven/src/integTest/groovy/org/gradle/api/publish/maven/MavenPublishHttpIntegTest.groovy
big-guy: I'd like to get rid of these too because there are other related methods that support these, but the cross-version test uses it more directly now.

big-guy:
@bot-gradle test this

Issue #38361, opened 2026-06-29
Dependency resolution failure - "Problems loading the resolution results"


ulrikls:
### Current Behavior

After upgrading to Gradle 9.6.1, IntelliJ cannot import our Gradle project:

```

CONFIGURE SUCCESSFUL in 235ms
> Task :prepareKotlinBuildScriptModel UP-TO-DATE
Internal error in Develocity Gradle plugin: finished notification
com.gradle.develocity.DevelocityException: Internal error in Develocity Gradle plugin: finished notification
	at com.gradle.develocity.DevelocityException.a(SourceFile:25)
	at com.gradle.scan.plugin.internal.operations.d.a(SourceFile:138)
	at com.gradle.scan.plugin.internal.operations.h.a(SourceFile:51)
	at com.gradle.scan.plugin.internal.s.a$a.a(SourceFile:31)
	at com.gradle.scan.plugin.internal.s.a$a.a(SourceFile:20)
	at com.gradle.scan.plugin.internal.s.a.c(SourceFile:67)
	at java.base@25.0.2/java.lang.Thread.run(Thread.java:1474)
Caused by: java.lang.RuntimeException: Problems reading data from Binary store in D:\.gradle\.tmp\gradle7461146297053433701.bin offset 1806560 exists? true
	at org.gradle.api.internal.artifacts.ivyservice.resolveengine.store.DefaultBinaryStore$SimpleBinaryData.read(DefaultBinaryStore.java:137)
	at org.gradle.api.internal.artifacts.ivyservice.resolveengine.result.StreamingResolutionResultBuilder$GraphFactory.lambda$create$1(StreamingResolutionResultBuilder.java:386)
	at org.gradle.api.internal.artifacts.ivyservice.resolveengine.store.CachedStoreFactory$SimpleStore.load(CachedStoreFactory.java:101)
	at org.gradle.api.internal.artifacts.ivyservice.resolveengine.result.StreamingResolutionResultBuilder$GraphFactory.create(StreamingResolutionResultBuilder.java:384)
	at org.gradle.api.internal.artifacts.ivyservice.resolveengine.graph.results.DefaultVisitedGraphResults.lambda$new$0(DefaultVisitedGraphResults.java:70)
	at org.gradle.api.internal.artifacts.configurations.ResolveConfigurationResolutionBuildOperationResult.getRootComponent(ResolveConfigurationResolutionBuildOperationResult.java:60)
	at java.base/java.util.Optional.map(Optional.java:260)
[...]

PR #38522, opened 2026-07-15
Fix stale capability conflict replacements after component eviction

jvandort:
When a node loses a capability conflict, it records a pointer to the winning node, and its incoming edges are retargeted to the winner. A later version or module conflict may then evict the winner's component, selecting a different component of the same module in its place. Nothing updates the loser's pointer when this happens, so edges recomputed after the eviction would follow the stale pointer and re-attach to a component no longer in the graph. This corrupted the resolution graph, surfacing as "Cannot find node index for target node" during result assembly, or as a graph-shape assertion failure when assertions are enabled.

The replacement pointer is now validated each time it is followed: a replacement whose component is no longer its module's selected component is stale, and is cleared so it cannot reactivate later. The affected edge then re-resolves to its original target, which re-enters capability conflict detection against the newly selected component and is re-resolved by the existing conflict resolution machinery.

Also:
- Rename NodeState.replacement to capabilityReplacement, to disambiguate node-level capability replacement from module replacement (replacedBy rules).
- Strengthen the post-retarget assertion in restartIncomingEdges to state the invariant this bug violated: an edge may only remain attached to a node whose component is its module's current selection.

Fixes https://github.com/gradle/gradle/issues/38361

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesConflictResolutionIssuesIntegrationTest.groovy
+++ b/platforms/software/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/capabilities/CapabilitiesConflictResolutionIssuesIntegrationTest.groovy
@@ -1261,6 +1261,253 @@ class CapabilitiesConflictResolutionIssuesIntegrationTest extends AbstractIntegr
         output.contains("project ':reproducer' (c)")
     }
 
+    @Issue("https://github.com/gradle/gradle/issues/38361")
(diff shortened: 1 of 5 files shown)


jvandort:
@bot-gradle test rfr

ljacomet:
@bot-gradle test this

ljacomet (approved):

PR #38525, opened 2026-07-15
Use model classes and kotlinx.serialization to emit CC/problems report

mlopatkin:
This commit replaces handwritten serialization code living far from the consumer with report-defined classes that are read and written through the same serialization pipeline.

The model remains mostly the same, except that to keep streaming possible, we emit diagnostics and total counts separately and rely on JS code to reassemble them into the object the report can understand. This is intended to be a temporary measure.

This is a foundation for simpler report evolution in the future.

Report size of the changes:
https://github.com/gradle/configuration-cache-report/pull/62

Design and stage overview: https://github.com/gradle/gradle/issues/39055#issuecomment-5546080451


platforms/ide/problems/src/main/kotlin/org/gradle/problems/internal/impl/DefaultProblemsReportCreator.kt
```diff
@@ -67,29 +68,23 @@ class DefaultProblemsReportCreator(
         distinctReports = false
     )
     private val taskNames = startParameter.taskNames
-    private val problemCount = AtomicInteger(0)
```
mlopatkin, line 70: ⚠️  Unlike CC report, the problems report doesn't use `totalProblemCount`, so it is removed.

reinsch82 (approved):

reinsch82 (approved):

mlopatkin:
Will land in https://github.com/gradle/gradle/pull/39144

PR #38527, opened 2026-07-15
Fix Worker API metaspace leak by allowing cross-session cache Class keys to be GC'ed

### LINKED ISSUE #18313: 'java.lang.OutOfMemoryError: Metaspace' when using the Worker API with classloader isolation
### Expected Behavior
When using the Worker API with classloader isolation, classes should either be cached/shared between worker actions, or alternatively be cleaned at the end of an action.

### Current Behavior
When using the Worker API with classloader isolation, classes that are loaded in worker actions are not cached nor cleaned, therefore causing `java.lang.OutOfMemoryError: Metaspace` exception rather quickly and also making the entire build stuck.

### Context
https://github.com/gradle/gradle/issues/15200#issuecomment-917546781

> I encountered this problem while trying to make the [scoverage (scala coverage tool) plugin](https://github.com/scoverage/gradle-scoverage) run its tasks in class-loader isolation mode, instead of adding the necessary classpath to the main/default gradle class loader.
> 
> My attempt can be seen [in this PR](https://github.com/scoverage/gradle-scoverage/pull/170), and the failed tests (with OOM Metaspace) can be seen in [this Github action](https://github.com/scoverage/gradle-scoverage/runs/3575971231).
> 
> The failed test is [a functional test](https://github.com/scoverage/gradle-scoverage/blob/master/src/functionalTest/java/org/scoverage/ScalaMultiModuleTest.java) which applies the plugin to a (relatively small) [multi-module project with only 3 modules](https://github.com/scoverage/gradle-scoverage/tree/master/src/functionalTest/resources/projects/scala-multi-module). The plugin creates multiple tasks for the project, and the classpath that is required for running these tasks consists of the [scalac-plugin](https://github.com/scoverage/scalac-scoverage-plugin) plus the entire scala library.
> 
> Unfortunately, this behavior makes the worker API unappealing for the purpose of class-loader isolation. May I add that parallelism is perhaps of lesser importance in this scenario; meaning, if there were a sequential API, which would perhaps make it easier to implement a cache, then that solution would be sufficient for our use case.

### Steps to Reproduce 
1. Create a new gradle project with the files below (and nothing else).

This project creates a `foo` task per module, which then uses the Worker API with classloader isolation to loads all the classes in the `foo` configuration (containing two arbitrary packages).

`build.gradle`
```gradle
import java.util.jar.JarFile

allprojects {
    repositories {
        mavenCentral()
    }
}

abstract class FooTask extends DefaultTask {
    private final WorkerExecutor workerExecutor

    @InputFiles
    abstract ConfigurableFileCollection getConfig()

    @Inject
    FooTask(WorkerExecutor workerExecutor) {
        this.workerExecutor = workerExecutor
    }

    @TaskAction
    void foo() {
        WorkQueue workQueue = workerExecutor.classLoaderIsolation() { spec ->
            spec.getClasspath().from(this.getConfig())
        }
        workQueue.submit(FooAction.class) {params ->
            params.config.from(this.getConfig())
        }
        workQueue.await()
    }

    static interface FooParameters extends WorkParameters {
        ConfigurableFileCollection getConfig()
    }

    static abstract class FooAction implements WorkAction<FooParameters> {
        @Override
        void execute() {
            println('hello')
            loadAllClasses()
            println('world')
        }

        void loadAllClasses() {
            ClassLoader cl = getClass().getClassLoader()
            getParameters().getConfig().getFiles().forEach { file ->
                JarFile jar = new JarFile(file)
                jar.entries().asIterator().forEachRemaining { entry ->
                    String entryName = entry.getName()
                    if (!entry.isDirectory() && entryName.endsWith('.class')) {
                        String className = entryName.substring(0, entryName.length()-6).replace('/','.')
                        cl.loadClass(className)
                    }
                }
            }
        }
    }
}


subprojects {
    configurations {
        foo
    }

    dependencies {
        foo group: 'org.scala-lang', name: 'scala-library', version: '2.13.6'
        foo group: 'com.google.guava', name: 'guava', version: '30.1.1-jre'
    }

    it.tasks.register('foo', FooTask.class) {
        config.from(configurations.foo)
    }
}
```

`settings.gradle`
```gradle
include 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k'
```

2. Execute `./gradlew foo`, after which the build will fail and also get stuck, requiring to force-kill the daemon and manually clearing cache locks:
```
$ gradle --stop
$ rm -rf .gradle
$ find ~/.gradle -type f -name "*.lock" -delete
```

### Your Environment
```
$ ./gradlew --version

Gradle 7.2

Build time:   2021-08-17 09:59:03 UTC
Revision:     a773786b58bb28710e3dc96c4d1a7063628952ad

Kotlin:       1.5.21
Groovy:       3.0.8
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          11.0.11 (AdoptOpenJDK 11.0.11+9)
OS:           Mac OS X 11.3 x86_64
```

Build scan URL: https://scans.gradle.com/s/arufy4mdd3pbu
(this one was only with three modules so it succeeded)

bamboo:
Fixes #18313

Before this change, `newClassCache()` and `newClassMap()` stored values in the
session-lived, strongly-referenced `valuesForThisSession` map inherited from
`AbstractCrossBuildInMemoryCache`. For a class cache, the key is a `Class`, so
every `classLoaderIsolation` worker's `Class` (and hence its `ClassLoader` and
all classes it defined) stayed pinned for the whole build session. A single
build that spawns many worker classloaders therefore exhausts Metaspace
(#18313).

This PR backs both factories with a new `ClassValueCache` that stores each value
on its key `Class` via `java.lang.ClassValue`. The value now lives exactly as
long as its class, imposing no external retention, so the class and its
ClassLoader become collectible as soon as they are otherwise unused -- including
within a single build session. This also subsumes the previous soft/weak
cross-session tiers for class caches and the `DefaultClassMap`
ClassLoader-user-data scheme, which is removed.

### DIFF
--- a/subprojects/core/src/main/java/org/gradle/cache/internal/DefaultCrossBuildInMemoryCacheFactory.java
+++ b/subprojects/core/src/main/java/org/gradle/cache/internal/DefaultCrossBuildInMemoryCacheFactory.java
@@ -84,7 +84,6 @@ public class DefaultCrossBuildInMemoryCacheFactory implements CrossBuildInMemory
         return cache;
     }
 
-
     @Override
     public <K, V> CrossBuildInMemoryCache<K, V> newCacheRetainingDataFromPreviousBuild(Predicate<V> retentionFilter) {
         CrossBuildCacheRetainingDataFromPreviousBuild<K, V> cache = new CrossBuildCacheRetainingDataFromPreviousBuild<>(retentionFilter);
@@ -253,7 +252,7 @@ public class DefaultCrossBuildInMemoryCacheFactory implements CrossBuildInMemory
          * re-introduce the pin this cache exists to avoid.
          */
         private static class Slot<V> {
-            volatile V value;
+            volatile @Nullable V value;
             volatile int epoch;
         }
 
@@ -276,15 +275,17 @@ public class DefaultCrossBuildInMemoryCacheFactory implements CrossBuildInMemory
             }
             synchronized (slot) {
                 epoch = currentEpoch.get();
-                if (slot.value == null || slot.epoch != epoch) {
+                V cached = slot.value;
+                if (cached == null || slot.epoch != epoch) {
                     V computed = factory.apply(key);
                     if (computed == null) {
                         throw new IllegalStateException("Factory '" + factory + "' failed to produce a value for key '" + key + "'!");
                     }
                     slot.value = computed;
                     slot.epoch = epoch;
+                    return computed;
                 }
-                return slot.value;
+                return cached;
             }
         }
 
--- a/subprojects/core/src/main/java/org/gradle/cache/internal/DefaultCrossBuildInMemoryCacheFactory.java
+++ b/subprojects/core/src/main/java/org/gradle/cache/internal/DefaultCrossBuildInMemoryCacheFactory.java
@@ -28,7 +28,6 @@ import java.lang.ref.SoftReference;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
-import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -37,7 +36,6 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 
-import static java.util.Collections.synchronizedMap;
 
 /**
  * A factory for {@link CrossBuildInMemoryCache} instances.
@@ -63,14 +61,14 @@ public class DefaultCrossBuildInMemoryCacheFactory implements CrossBuildInMemory
 
     @Override
     public <K, V> CrossBuildInMemoryCache<K, V> newCache() {
-        DefaultCrossBuildInMemoryCache<K, V> cache = new DefaultCrossBuildInMemoryCache<>(KeyRetentionPolicy.STRONG);
+        DefaultCrossBuildInMemoryCache<K, V> cache = new DefaultCrossBuildInMemoryCache<>();
         listenerManager.addListener(cache);
         return cache;
     }
 
     @Override
     public <K, V> CrossBuildInMemoryCache<K, V> newCache(Consumer<V> onReuse) {
-        DefaultCrossBuildInMemoryCache<K, V> cache = new DefaultCrossBuildInMemoryCache<K, V>(KeyRetentionPolicy.STRONG) {
+        DefaultCrossBuildInMemoryCache<K, V> cache = new DefaultCrossBuildInMemoryCache<K, V>() {
             @Nullable
             @Override
             protected V maybeGetRetainedValue(K key) {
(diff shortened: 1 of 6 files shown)


bamboo:
@bot-gradle test APT, AST

bamboo:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/cache/internal/DefaultCrossBuildInMemoryCacheFactory.java
```diff
@@ -251,40 +230,85 @@ protected V maybeGetRetainedValue(K key) {
...
-        protected void retainValuesFromCurrentSession(Stream<V> values) {
-            // Ignore
+        /**
+         * Holds the cached value on the key {@code Class}. The {@code epoch} lets {@link #clear()} invalidate every
+         * entry in O(1) without keeping a registry of keys -- such a registry would reference the keys and so
+         * re-introduce the pin this cache exists to avoid.
+         */
+        private static class Slot<V> {
+            @Nullable
+            volatile V value;
```
vlsi:
```suggestion
            volatile @Nullable V value;
```

Nit: `@Nullable` is a type annotation (it's target is `TYPE_USE` not `FIELD`), so it should belong to type rather than a field.

For instance, you would write `volatile List<@Nullable String> listOfNullableStrings` (you can't move the annotation to a top-level for such use case), so when the list itself is nullable, you would write `volatile @Nullable List<String> nullableListOfStrings`.

bamboo:
@bot-gradle test PRF

bamboo:
@bot-gradle test RFN

subprojects/core/src/test/groovy/org/gradle/cache/internal/DefaultCrossBuildInMemoryCacheFactoryTest.groovy
```diff
@@ -96,4 +103,68 @@ class DefaultCrossBuildInMemoryCacheFactoryTest extends AbstractCrossBuildInMemo
...
+
+    @TempDir
+    File tempDir
+
+    // Compiles a trivial pure-Java class and loads it in a fresh, isolated URLClassLoader (a pure-Java class avoids
+    // Groovy's global ClassInfo/globalClassSet retention), then caches a value against it. Stays dynamic Groovy --
+    // the JavaCompiler.run varargs call does not resolve under @CompileStatic -- but never dispatches a method ON
+    // the throwaway Class; that reflection is isolated in cacheConstructor() below. Returns ONLY a weak probe on
+    // the loader, so no strong local escapes to the caller.
+    private WeakReference<ClassLoader> populate(CrossBuildInMemoryCache<Class<?>, Object> cache) {
```
mlopatkin: :nail_care: I'd rather use a prebuilt java class next to the test and a custom classloader like
```groovy
new URLClassLoader([Throwaway.protectionDomain.codeSource.location] as URL[], null as ClassLoader)
```
rather than invoking a compiler. Then we probably can have the whole method as `@CompileStatic`.

mlopatkin (approved):
Looks good, one minor suggestion.

subprojects/core/src/test/groovy/org/gradle/cache/internal/DefaultCrossBuildInMemoryCacheFactoryTest.groovy
bamboo: Thanks, @mlopatkin!

bamboo:
@bot-gradle test PRF, APT

PR #38530, opened 2026-07-15
Do not publish checksums for signature files

big-guy:
This is Part 2 of reducing the number of files Gradle publishes. Part 1 (#38519) stopped publishing SHA-256/SHA-512 checksums for Maven and Ivy publications. This change stops publishing checksum files (`.sha1`, `.md5`) for signature files (`.asc`, `.sig`).

Checksums of signature files provide no value: a signature already protects the integrity of the artifact it signs, and consumers verify the signature directly rather than its checksum. Publishing them just adds noise to the repository.

## Changes

- **maven-publish** (`AbstractMavenPublisher`): skip checksum generation when the published resource is a signature file.
- **ivy-publish** (`ExternalResourceResolver`): same guard for the Ivy publishing path.
- Both use an `isSignatureFile()` helper matching `.asc`/`.sig` suffixes.
- Added assertions to `SigningPublicationsIntegrationSpec` verifying that signature files have no checksums while artifacts still do.
- Documented in the 9.x upgrade guide.

## Reviewing

Take each commit in isolation.

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/resolver/ExternalResourceResolver.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/resolver/ExternalResourceResolver.java
@@ -321,7 +321,18 @@ public abstract class ExternalResourceResolver implements ConfiguredModuleCompon
 
     private void put(File src, ExternalResourceName destination) {
         repository.withProgressLogging().resource(destination).put(new FileReadableContent(src));
-        publishChecksums(destination, src);
+        if (!isSignatureFile(destination)) {
+            publishChecksums(destination, src);
+        }
+    }
+
+    /**
+     * Signature files (e.g. {@code .asc}, {@code .sig}) are themselves integrity artifacts, so we do not
+     * publish checksums for them.
+     */
+    private static boolean isSignatureFile(ExternalResourceName destination) {
+        String path = destination.getPath();
+        return path.endsWith(".asc") || path.endsWith(".sig");
     }
 
     private void publishChecksums(ExternalResourceName destination, File content) {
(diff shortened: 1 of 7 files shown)


big-guy:
@bot-gradle test this

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/resolver/ExternalResourceResolver.java
```diff
@@ -321,7 +321,18 @@ public void publish(ModuleComponentArtifactMetadata artifact, File src) {
...
+        }
+    }
+
+    /**
+     * Signature files (e.g. {@code .asc}, {@code .sig}) are themselves integrity artifacts, so we do not
+     * publish checksums for them.
+     */
+    private static boolean isSignatureFile(ExternalResourceName destination) {
+        String path = destination.getPath();
+        return path.endsWith(".asc") || path.endsWith(".sig");
```
jvandort, line 335: I just developed two new file formats:
- Amazing Super Compression (.asc)
- Scientific Information Graph (.sig)

But when I try to publish these to a repository they get rejected since they don't have checksums.

It's strange. I was also considering Super Amazing Compression (.sac) and that works without issue.

jvandort (commented):
I think the "does this artifact need checksums" should be a property of an artifact rather than inferred from the extension

lkasso (approved):
Docs LGTM.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/resolver/ExternalResourceResolver.java
ljacomet: I think we can do a quick fix now and a proper one in 10.

For me we should fully leverage artifacts and derived artifacts:
* Checksum only regular artifacts
* Sign only regular artifacts

Everything added to derived artifacts must be complete - that is if your derived artifact needs a signature or a checksum, handle that.

ljacomet (approved):

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/resolver/ExternalResourceResolver.java
aSemy: I have created a PR that fixes this in #39246. It replaces the file extension check with a 'DerivedArtifact' check. Only regular artifacts get signatures, derived artifacts do not.

PR #38536, opened 2026-07-15
DomainObjectContext updates

jvandort:
Updates to attempt to make this type more generic to better model things like build scoped state. Still more work to do. See individual commits for specific details:

- Move CalculatedModelValue out of ModelContainer. Was only used in Configuration, no need for this to be tied to DomainObjectContext specifically.
- Remove identityPath(String) and projectPath(String) in favor of exposing the raw identity path and letting consumers create paths as necessary
- Remove DomainObjectContext.getProject(). Migrate users to not require any project at all or otherwise move to getProjectState().getMutableModel()

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

PR #25619, opened 2023-07-05
Apply coarse locking on all projects to solve deprecation warning

hegyibalint:
* Fixes #21492

PR #38537, opened 2026-07-15
Address unsafe project lock avoidance mechanisms

jvandort:
There are two mechanism that permit unsafe access to projects without acquiring their locks:
- ProjectLeaseRegistry.allowUncontrolledAccessToAnyProject
- ModelContainer.forceAccessToMutableState

allowUncontrolledAccessToAnyProject was used in two places.
1. ToolingModelBuilderRegistry.getBuilder
- Git history states this was added to restore historical unsafe behavior that the android plugin relied on.
- A search of AGP source code shows this method is no longer referenced anywhere in AGP
2. AbstractKotlinDslScriptsModelBuilder.buildAll
- Git history shows this was added to avoid a depecation during model bulding in 8.x. Added in https://github.com/gradle/gradle/pull/25619
- The accompanying test that was introduced as part of this PR does not fail when the workaround is removed.
- No other explanation in git/PR history was included as to _why_ this solution was chosen.

forceAccessToMutableState is only used in DefaultTransform to workaround a case where reaching into another project and resolving a configuration from that project that triggers artifact transforms would cause transform parameter isolation to occur without proper lock acquired. When transforms are exected on the task graph, lock are properly acquired. This can only occur when undeclared transform execution is performed. Undeclared transform execution was deprecated in https://github.com/gradle/gradle/pull/38173 This method can be removed in Gradle 10 when undeclared transform execution becomes a failure.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/moduleconverter/dependencies/DefaultLocalVariantGraphResolveStateBuilder.java
```diff
@@ -206,19 +206,15 @@ private static CalculatedValue<ImmutableList<LocalComponentArtifactMetadata>> ge
         ModelContainer<?> model,
         CalculatedValueContainerFactory calculatedValueContainerFactory
     ) {
-        return calculatedValueContainerFactory.create(Describables.of(displayName, "artifacts"), context -> {
-            if (sourceArtifacts.isEmpty()) {
-                return ImmutableList.of();
```
jvandort, line 211: Calling `sourceArtifacts.isEmpty` can execute code in the target project. It is unsafe to make this check without a lock on that project. We instead always acquire the lock on the target project. 

This was discovered while investigating the usage of `forceAccessToMutableState` in `DefaultTransform`.

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/AbstractResourceLockRegistry.java
```diff
@@ -119,12 +108,11 @@ public void lockReleased(ResourceLock resourceLock) {
     }
 
     public boolean mayAttemptToChangeLocks() {
-        ThreadLockDetails<T> details = detailsForCurrentThread();
-        return details.mayChange && !details.canAccessAnything;
+        return detailsForCurrentThread().mayChange;
     }
 
     public boolean isAllowedUncontrolledAccessToAnyResource() {
```
octylFractal: ❓ Why not clean this method up?

jvandort: An oversight. I should address this

jvandort: Removed

jvandort:
@bot-gradle test this

octylFractal (approved):

jvandort:
@bot-gradle test this

PR #38538, opened 2026-07-16
Deprecate legacy lock file format

jvandort:
Prior to Gradle 6.4, each configuration wrote their own lock state to their own file. Starting in 6.4, we began writing a unified lockfile for each project, with all lock state for all configurations in that project. However, we continued reading these legacy lock files. Starting in Gradle 10, we will no longer maintain the machinery to read lock files in this legacy format.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

lkasso (approved):
Docs LGTM.

PR #38545, opened 2026-07-16
Make `@since` check more strict and backfill the annotation for existing APIs

### LINKED ISSUE #27113: Public API should always annotate each method with `@since` annotation
The current practice is to only include it for methods, when the class already exist. When an entirely new class/interface is added, then the since-tag only appears in the class-javadoc.

This creates friction in multiple ways:
- When working on Gradle itself and arriving at a method it’s useful to quickly understand when it was added. Having to scroll to the class-javadoc is unnecessary work
- For the same reason, the since-tag is not shown for methods in Quick Doc view in the IDE
  - This is also relevant for build engineers working on a Gradle plugin, for instance
- In the userguide, we often link to specific methods in the DSL (Groovy historically, Kotlin more recently), and when users arrive at that page view, it’s not clear since when the method is available.

The team agreed that this should be enforced at merge time, for instance, with an architecture test.

ov7a:
Fixes:
* https://github.com/gradle/gradle/issues/27113

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ov7a:
@coderabbitai review

ov7a:
Mismatches report for cases where git tag & the check & existing annotation disagree 
[since-mismatch-report.txt](https://github.com/user-attachments/files/30086343/since-mismatch-report.txt)

PR #38547, opened 2026-07-16
Refactoring: identify .gradle usages as project cache dir

ov7a:
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/idea/IdeaPlugin.java
+++ b/platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/idea/IdeaPlugin.java
@@ -326,6 +326,9 @@ public abstract class IdeaPlugin extends IdePlugin {
         conventionMapping.map("excludeDirs", new Callable<Set<File>>() {
             @Override
             public Set<File> call() {
+                // ".gradle" is the default project cache dir name (see BuildScopeCacheDir). Hardcoding it here is a
+                // historical accident: it should honor the user-configurable --project-cache-dir instead.
+                // We deliberately leave it as-is, as this IDE model generation is scheduled for removal in Gradle 10.
                 excludeDirs.add(project.file(".gradle"));
                 excludeDirs.add(project.getLayout().getBuildDirectory().getAsFile().get());
                 return excludeDirs;
--- a/platforms/software/build-init/src/main/java/org/gradle/buildinit/plugins/internal/GitIgnoreGenerator.java
+++ b/platforms/software/build-init/src/main/java/org/gradle/buildinit/plugins/internal/GitIgnoreGenerator.java
@@ -61,6 +61,9 @@ public class GitIgnoreGenerator implements BuildContentGenerator {
 
     @SuppressWarnings("DefaultCharset") //TODO: evaluate errorprone suppression (https://github.com/gradle/gradle/issues/35864)
     private static Set<String> getGitignoresToAppend(File gitignoreFile) {
+        // .gradle - project cache directory, see https://docs.gradle.org/current/userguide/directory_layout.html#dir:project_root
+        //  build  - build output directory
+        // .kotlin - Kotlin Gradle Plugin caches/metadata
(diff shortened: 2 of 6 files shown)


platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/idea/IdeaPlugin.java
```diff
@@ -326,6 +326,7 @@ public Set<File> call() throws Exception {
         conventionMapping.map("excludeDirs", new Callable<Set<File>>() {
             @Override
             public Set<File> call() {
+                // Literal ".gradle" is the default project cache dir name
```
alllex: 🤔  I don't think it's immediately clear what this means without context. I'd link to the `BuidlScopeCacheDir`, if only by mentioning it. Though, I think even more valuable would be to say that this directory is hardcoded here, and this is likely a historical accident, because it should have been referencing user-configurable project-cache-dir (link to userguide?). However, we choose not to fix it, because the owner implementation is scheduled for removal in Gradle 10.

platforms/software/build-init/src/main/java/org/gradle/buildinit/plugins/internal/GitIgnoreGenerator.java
```diff
@@ -61,6 +61,7 @@ public void generate(InitSettings settings, BuildContentGenerationContext buildC
 
     @SuppressWarnings("DefaultCharset") //TODO: evaluate errorprone suppression (https://github.com/gradle/gradle/issues/35864)
     private static Set<String> getGitignoresToAppend(File gitignoreFile) {
+        // Literal ".gradle" is the default project cache dir name
```
alllex: 🤔  This is a strange one. It suffers from the same problem as the other location. Though, in this case it's probably better to link to the user guide, because generating projects is somewhat an "outside" activity, rather than an internal implementation. Ideally, we provide a reason for each directory that lives here, however obvious. Though, I am not 100% sure why `.kotlin` is here. Case in point, I guess.

subprojects/core/src/main/java/org/gradle/cache/internal/BuildScopeCacheDir.java
```diff
@@ -28,6 +28,7 @@
 
 public class BuildScopeCacheDir {
     public static final String UNDEFINED_BUILD = "undefined-build/";
+    private static final String DEFAULT_PROJECT_CACHE_DIR = ".gradle";
```
alllex: 💭  The name of this field is directly misaligned with the name of the class

I still maintain that the directory far outgrew the "project cache dir" name, to the point where the user-facing property is a misleading misnomer. We don't have to do anything about it now, necessarily. But I'd rather we have field names that represent reality, while we can't change the user-facing properties that easily.

alllex (approved):
LGTM with some suggestions

platforms/software/build-init/src/main/java/org/gradle/buildinit/plugins/internal/GitIgnoreGenerator.java
ov7a: .kotlin is for kotlin daemon

subprojects/core/src/main/java/org/gradle/cache/internal/BuildScopeCacheDir.java
ov7a: 🤔 I changed it to refer to the class. 

I think for the sake of ubiquitous language, we should stick to "project cache dir". I think the reference in the code should be aligned with user-facing concepts (cli arg/docs) and the surrounding code (e.g. BuildScopeCacheDir).

However, I'm fine with having it refer to the BuildScopeCacheDir class for now.

ov7a:
Despite ongoing discussion, I'm merging this: even if the PR isn't ideal, it improves things. Further polishing may happen in follow-up PRs.

Issue #34483, opened 2025-07-30
It's too verbose to `ExecOperations.exec` in `doFirst`/`doLast`


dfabulich:
### Expected Behavior

In Gradle 8, I can do this:

```kotlin
tasks.register("myTask") {
  doLast {
    exec { commandLine("sh", "-c", "myscript.sh") }
    exec { commandLine("sh", "-c", "myscript2.sh") }
  }
}
```

Today, I get a deprecation warning when I do that.

> The Project.exec(Action) method has been deprecated. This is scheduled to be removed in Gradle 9.0. Use ExecOperations.exec(Action) or ProviderFactory.exec(Action) instead. Consult the upgrading guide for further information: https://docs.gradle.org/8.11.1/userguide/upgrading_version_8.html#deprecated_project_exec

The recommended replacement is to use an injected `ExecOperation`. https://docs.gradle.org/8.11.1/userguide/service_injection.html#execoperations

The documentation gives this example:

> With some ceremony, it is possible to use `ExecOperations` in an ad-hoc task defined in a build script:
>
> ```kotlin
> interface InjectedExecOps {
>     @get:Inject val execOps: ExecOperations
> }
> 
> tasks.register("myAdHocExecOperationsTask") {
>     val injected = project.objects.newInstance<InjectedExecOps>()
> 
>     doLast {
>         injected.execOps.exec {
>             commandLine("ls", "-la")
>         }
>     }
> }
> ```

That is a _lot_ of ceremony, and it requires knowing how Gradle injected services work in order to understand it. I wish there were briefer and simpler way to write this.

I'd love to at least be able to write this, without defining my own `InjectedExecOps` interface:

```
val execOps = ExecOperations.create()
tasks.register("myTask") {
  doLast {
    execOps.exec {
      commandLine("myscript.sh")
    }
  }
}
```

Even better if a `execOps` instance were just one of Gradle's many automatic imports. The documentation claims that `ExecOperations.exec` is a "drop-in replacement" for `Project.exec`, but a true drop-in replacement would look like this:

```
tasks.register("myTask") {
  doLast {
    execOps.exec {
      commandLine("myscript.sh")
    }
  }
}
```

### Context

[...]

Issue #22536, opened 2022-10-27
Simplify the use of injectable services in ad-hoc tasks


mlopatkin:
The snippet mentioned at the ["Working with files" chapter](https://docs.gradle.org/7.5/userguide/working_with_files.html#sec:project_copy_method) is using `Project.copy` at the execution time, this is not supported with the configuration cache. The currently supported alternative is to use the `@inject`-able `FileSystemOperations`, but the code for this is cumbersome: a build author has to write the task as a class or declare an interface with `@Inject` getter and use `ObjectFactory.newInstance` to obtain the instance of the interface. Both approaches require much more Gradle knowledge than the user starting the journey has at this point.

We're reconsidering if the injection should be the only possible approach to obtain `FileSystemOperations`. The chapter should be revisited after the decision is made.

mlopatkin:
As we work on this, we should also rewrite Groovy/Kotlin API usages in the doc snippets, like
https://github.com/gradle/gradle/pull/23033#pullrequestreview-1209999360

mlopatkin:
Actually, we have an issue for this already: #13121.

Issue #21517, opened 2022-08-12
Init scripts need access to `ProviderFactory` to allow for configuration cache- safe logic branches


mcumings:
With the configuration cache enabled, configuration cache hits result in a lack of init script execution. As a result, any branching logic contained within the init script will fail to be re-evaluated unless:
- The logic happens to be based upon an input that would be detected as changed via other means (e.g., command line changes)
-  The init script file content itself changes

I don't see any way to obtain a `ProviderFactory` instance from within an init script.  As a result. init scripts are unable to leverage the `ValueSource` API and therefore have no means to communicate changes in conditions that might result in a need to be reevaluated.

A simple example of this might be the test for a file's presence at a particular path.

### Expected Behavior

Init scripts have access to `ProviderFactory` and can register custom implementations.

### Current Behavior

Init scripts don't appear to be able to access the `ProviderFactory` and therefore cannot take advantage of the existing providers (e.g., `environmentVariable(...)`) nor can it register custom value sources.

### Context

Our build utilizes a sentinel file's presence as an indicator that a particular build feature should be enabled.  When running with configuration cache enabled, the presence of this file is only tested once and the result is stored in the configuration cache.  If the sentinel file is removed and the build re-run then Gradle does not detect the necessary change in behavior since the init script is not reinvoked.

mcumings:
One clarification: `gradle.rootProject { providers... }` can be used but in my understanding this is a deferred invocation versus direct/inline.

mlopatkin:
This is going to be solved as part of the more general #13121.

PR #38552, opened 2026-07-16
Add public service lookup to Project, Settings, Gradle and Task

### LINKED ISSUE #13121: Reduce ceremony to use common public services in Gradle scripts
Using common public Gradle service from scripts requires unnecessary ceremony in some cases. This issue is about removing the ceremony thus making it easier to write Gradle scripts, including binary plugins authored as precompiled scripts.

### Expected Behavior

Injected services [listed in the userguide](https://docs.gradle.org/current/userguide/custom_gradle_types.html#service_injection) can all be used from all Gradle scripts without ceremony.

Let's say for example that in a project script I want to add a `doLast` action to an existing task that does some file system operations.

```kotlin
// The existing task
tasks.register("some")

// Adding my doLast task action
tasks.named("some") {
    doLast {
        fs.delete { delete("thing") }
    }
}
```

The above is with an hypothetical access to the `FileSystemOperations` service as `DefaultTask.fs`. This is one potential option but not an implementation goal. Deciding how the services can be accessed depending on the use cases should be considered as part of this issue.

### Current Behavior

In project scripts one can already directly use `ObjectFactory`, `ProjectLayout` and `ProviderFactory`. None of the public service can be used directly in scripts settings or init scripts. Note that `ProjectLayout` does make sense when targeting a `Project` only.

With no possible direct usage one need to write a full blown plugin class in order to get the service injected:

```kotlin
// The existing task
tasks.register("some")

// Adding my doLast task action
abstract class MyPlugin : Plugin<Project> {

    @get:javax.inject.Inject
    abstract val fs: FileSystemOperations

    override fun apply(project: Project): Unit = project.run {
        tasks.named("some") {
            doLast {
                fs.delete { delete("thing") }
            }
        }
    }
}
apply<MyPlugin>()
```

This is a lot of ceremony.

### Context

Faced while moving away from `project.delete()` in favor of the `FileSystemOperations` service because the former isn't allowed when the configuration cache is enabled.

Also faced while writing settings scripts that access system properties, wanting to use `ProviderFactory.systemProperty(..)`.

reinsch82:
### Context

Part of #13121 ("Reduce ceremony to use common public services in Gradle scripts") and its related issues (#14171, #21517, #22536, #34483): since Gradle 9 removed `Project.exec` (and `project.delete`, while still present, cannot be called at execution time with the configuration cache), even a simple `doLast { exec }` requires writing a plugin class or the `ObjectFactory.newInstance` injection ceremony, which is incomprehensible to newcomers.

This PR introduces the service-lookup design the team leaned towards in [the issue discussion](https://github.com/gradle/gradle/issues/13121#issuecomment-4542596503) (option 3): a `service(Class)` method on `Project`, `Settings`, `Gradle` and `Task`, plus reified `service<T>()` extensions in the Kotlin DSL.

```groovy
tasks.register("cleanThing") {
    doLast {
        service(FileSystemOperations).delete {
            delete("thing")
        }
    }
}
```

Availability per scope is expressed with **per-scope marker interfaces** — `ProjectService`, `TaskService`, `SettingsService`, `GradleService` — used as upper bounds on the corresponding `service(...)` method. As a result, looking up a service that is not available in the current scope, or a type that is not a service at all, is a **compile error in the Kotlin and Java DSLs**. An internal runtime allowlist backs the same restriction for the Groovy DSL (where the generic bound is not enforced) and for reflective `Class` lookups, failing with an `InvalidUserDataException` whose message enumerates the services available in the current scope; `BuildService` subtypes are redirected to the shared-build-service APIs.

It works in build, settings and init scripts (Groovy and Kotlin DSL) and inside task actions, and is configuration-cache safe: the `Task` entry point resolves against the owning project's registry, which is live at execution time after CC loading, and captured service instances round-trip via the existing `ServicesCodec`.

**Exposed services (v1):**

- All scopes: `ObjectFactory`, `ProviderFactory`, `FileSystemOperations`, `ArchiveOperations`
- Projects and tasks: `ProjectLayout`
- Tasks only: `ExecOperations` — resolving it outside a task action (i.e. at configuration time) is incompatible with the configuration cache, so it is restricted to the task scope
- Settings only: `BuildLayout`

The list is deliberately conservative — build-scope services still under IP-safety review (#38028) are deferred.

**Isolated Projects:** cross-project lookups (`project(":other").service(...)`) are reported as cross-project access, consistent with `getObjects()`/`getProviders()`.

### Work plan (phased commits on this PR)

- [x] Phase 1 — `service(Class)` API + allowlist validation + Groovy DSL coverage (unit + integration tests incl. configuration cache)
- [x] Phase 2 — Kotlin DSL reified `service<T>()` extensions
- [x] Phase 3 — Isolated Projects + Groovy dispatch edge-case tests
- [x] Phase 4 — User guide (service_injection.adoc) + release notes
- [x] Review feedback — per-scope marker interfaces + generic bounds for compile-time validation; `ExecOperations` restricted to tasks

### Contributor Checklist
- [x] Ensure that tests pass locally: `:core:test`, `:core:forkingIntegTest`, `:core:configCacheIntegTest`, and `:kotlin-dsl:forkingIntegTest` for the new test classes.

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupConfigurationCacheIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupConfigurationCacheIntegrationTest.groovy
@@ -20,44 +20,14 @@ import org.gradle.integtests.fixtures.configurationcache.ConfigurationCacheFixtu
 import spock.lang.Issue
 
 /**
- * Pins the Groovy closure dispatch and configuration-cache behavior of the public {@code service(Class)}
- * lookup in the corner cases where the receiver resolution or the captured instance interacts with the
- * configuration cache. Happy-path and error-message coverage lives in the core
- * {@code PublicServiceLookupIntegrationTest}.
+ * Pins how a public {@code service(Class)} instance captured in a settings script survives the
+ * configuration cache and remains usable from a task action. Happy-path and error-message coverage
+ * lives in the core {@code PublicServiceLookupIntegrationTest}.
  */
 @Issue("https://github.com/gradle/gradle/issues/13121")
-class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCacheIntegrationTest {
+class ServiceLookupConfigurationCacheIntegrationTest extends AbstractConfigurationCacheIntegrationTest {
 
     def configurationCache = new ConfigurationCacheFixture(this)
-
-    def "service lookup in an owner-first task closure fails as an unsupported script reference at execution time"() {
-        given:
-        buildFile """
-            tasks.register("some") {
-                onlyIf {
-                    service(ProviderFactory)
-                    throw new IllegalStateException("UNREACHABLE")
-                }
-                doFirst {
-                }
-            }
-        """
-
-        when:
-        configurationCacheFails ":some"
-
-        then:
-        failure.assertHasFailure("Invocation of 'service' references a Gradle script object from a Groovy closure at execution time, which is unsupported with the configuration cache.") {
-            // The cause is not reported
-        }
-        outputDoesNotContain("UNREACHABLE")
-
-        configurationCache.assertStateStoredAndDiscarded {
-            hasStoreFailure = false
-            reportedOutsideBuildFailure = true
-            problem "Task `:some` of type `org.gradle.api.DefaultTask`: invocation of 'service' references a Gradle script object from a Groovy closure at execution time, which is unsupported with the configuration cache."
-        }
-    }
 
     def "a settings-scoped service captured in a settings script survives the configuration cache and is usable in a task action"() {
         given:
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
@@ -96,6 +96,8 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
             gradle.rootProject {
                 tasks.register("useLayout") {
                     doLast {
+                        // Tripwire on the first line: if the action is ever entered, this prints.
+                        println("REACHED ACTION")
                         println("settings dir: " + captured.settingsDirectory.asFile.name)
                     }
                 }
@@ -106,12 +108,16 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
         run ":useLayout"
 
         then:
+        outputContains("REACHED ACTION")
         outputContains("settings dir: " + testDirectory.name)
 
         when: "run with the configuration cache, the captured service is re-resolved from the task's own project registry, which cannot see the settings scope"
         configurationCacheFails ":useLayout"
 
         then:
+        // The failure happens while loading the cache entry, before the task action runs...
         failure.assertHasCause("No service of type BuildLayout available in project services.")
+        // ...so the action is never entered on the reused build.
+        outputDoesNotContain("REACHED ACTION")
     }
 }
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
@@ -82,7 +82,7 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
         then:
         // The rejection happens while the task is being created, not while it runs...
         failure.assertHasCause("Could not create task ':useLayout'.")
-        failure.assertHasCause("org.gradle.api.file.BuildLayout is not available in tasks. It is available in settings scripts and settings plugins.")
+        failure.assertHasCause("org.gradle.api.file.BuildLayout is not available in tasks.\nIt is available in settings scripts and settings plugins.")
         // ...so the task action is never entered.
         outputDoesNotContain("REACHED ACTION")
     }
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
@@ -82,7 +82,7 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
         then:
         // The rejection happens while the task is being created, not while it runs...
         failure.assertHasCause("Could not create task ':useLayout'.")
-        failure.assertHasCause("org.gradle.api.file.BuildLayout is not available in tasks. It is available in settings scripts and plugins.")
+        failure.assertHasCause("org.gradle.api.file.BuildLayout is not available in tasks. It is available in settings scripts and settings plugins.")
         // ...so the task action is never entered.
         outputDoesNotContain("REACHED ACTION")
     }
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl
+
+import org.gradle.integtests.fixtures.configurationcache.ConfigurationCacheFixture
+import spock.lang.Issue
+
+/**
+ * Pins the Groovy closure dispatch and configuration-cache behavior of the public {@code service(Class)}
+ * lookup in the corner cases where the receiver resolution or the captured instance interacts with the
+ * configuration cache. Happy-path and error-message coverage lives in the core
+ * {@code PublicServiceLookupIntegrationTest}.
+ */
+@Issue("https://github.com/gradle/gradle/issues/13121")
+class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCacheIntegrationTest {
+
+    def configurationCache = new ConfigurationCacheFixture(this)
+
+    def "service lookup in an owner-first task closure fails as an unsupported script reference at execution time"() {
+        given:
+        buildFile << """
+            tasks.register("some") {
+                onlyIf {
+                    service(ProviderFactory)
+                    throw new IllegalStateException("UNREACHABLE")
+                }
+                doFirst {
+                }
+            }
+        """
+
+        when:
+        configurationCacheFails ":some"
+
+        then:
+        failure.assertHasFailure("Invocation of 'service' references a Gradle script object from a Groovy closure at execution time, which is unsupported with the configuration cache.") {
+            // The cause is not reported
+        }
+        outputDoesNotContain("UNREACHABLE")
+
+        configurationCache.assertStateStoredAndDiscarded {
+            hasStoreFailure = false
+            reportedOutsideBuildFailure = true
+            problem "Task `:some` of type `org.gradle.api.DefaultTask`: invocation of 'service' references a Gradle script object from a Groovy closure at execution time, which is unsupported with the configuration cache."
+        }
+    }
+
+    def "looking up a settings-only service through a task closure resolves to the task and is rejected at configuration time"() {
+        given:
+        settingsFile << """
+            gradle.rootProject {
+                tasks.register("useLayout") {
+                    // `service` here resolves to the Task receiver, which does not expose the
+                    // settings-only BuildLayout, so this fails while the task is being configured.
+                    def captured = service(BuildLayout)
+                    doLast {
+                        println("settings dir: " + captured.settingsDirectory.asFile.name)
+                    }
+                }
+            }
+        """
+
+        when:
+        fails ":useLayout"
+
+        then:
+        failure.assertHasCause("org.gradle.api.file.BuildLayout is not available in tasks. It is available in settings scripts and plugins.")
+    }
+
+    def "a settings-scoped service captured in a settings script and used in a task action cannot be re-resolved under the configuration cache"() {
+        given:
+        settingsFile << """
+            // Captured at settings-script scope, where `service` resolves to the Settings receiver
+            // and BuildLayout is available. The instance is then used from a task action.
+            def captured = service(BuildLayout)
+            gradle.rootProject {
+                tasks.register("useLayout") {
+                    doLast {
+                        println("settings dir: " + captured.settingsDirectory.asFile.name)
+                    }
+                }
+            }
+        """
+
+        when: "run without the configuration cache, the captured settings-scoped service is used directly"
+        run ":useLayout"
+
+        then:
+        outputContains("settings dir: " + testDirectory.name)
+
+        when: "run with the configuration cache, the captured service is re-resolved from the task's own project registry, which cannot see the settings scope"
+        configurationCacheFails ":useLayout"
+
+        then:
+        failure.assertHasCause("No service of type BuildLayout available in project services.")
+    }
+}
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
@@ -87,7 +87,7 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
         outputDoesNotContain("REACHED ACTION")
     }
 
-    def "a settings-scoped service captured in a settings script and used in a task action cannot be re-resolved under the configuration cache"() {
+    def "a settings-scoped service captured in a settings script survives the configuration cache and is usable in a task action"() {
         given:
         settingsFile """
             // Captured at settings-script scope, where `service` resolves to the Settings receiver
@@ -96,7 +96,7 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
             gradle.rootProject {
                 tasks.register("useLayout") {
                     doLast {
-                        // Tripwire on the first line: if the action is ever entered, this prints.
+                        // Tripwire on the first line: prints only if the action is actually entered.
                         println("REACHED ACTION")
                         println("settings dir: " + captured.settingsDirectory.asFile.name)
                     }
@@ -111,13 +111,20 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
         outputContains("REACHED ACTION")
         outputContains("settings dir: " + testDirectory.name)
 
-        when: "run with the configuration cache, the captured service is re-resolved from the task's own project registry, which cannot see the settings scope"
-        configurationCacheFails ":useLayout"
+        when: "the configuration cache entry is stored, BuildLayout is captured by value"
+        configurationCacheRun ":useLayout"
 
         then:
-        // The failure happens while loading the cache entry, before the task action runs...
-        failure.assertHasCause("No service of type BuildLayout available in project services.")
-        // ...so the action is never entered on the reused build.
-        outputDoesNotContain("REACHED ACTION")
+        configurationCache.assertStateStored()
+        outputContains("REACHED ACTION")
+        outputContains("settings dir: " + testDirectory.name)
+
+        when: "the configuration cache entry is reused, the captured BuildLayout is restored by value"
+        configurationCacheRun ":useLayout"
+
+        then:
+        configurationCache.assertStateLoaded()
+        outputContains("REACHED ACTION")
+        outputContains("settings dir: " + testDirectory.name)
     }
 }
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
@@ -68,6 +68,8 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
                     // settings-only BuildLayout, so this fails while the task is being configured.
                     def captured = service(BuildLayout)
                     doLast {
+                        // Tripwire on the first line: if the action is ever entered, this prints.
+                        println("REACHED ACTION")
                         println("settings dir: " + captured.settingsDirectory.asFile.name)
                     }
                 }
@@ -78,7 +80,11 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
         fails ":useLayout"
 
         then:
+        // The rejection happens while the task is being created, not while it runs...
+        failure.assertHasCause("Could not create task ':useLayout'.")
         failure.assertHasCause("org.gradle.api.file.BuildLayout is not available in tasks. It is available in settings scripts and plugins.")
+        // ...so the task action is never entered.
+        outputDoesNotContain("REACHED ACTION")
     }
 
     def "a settings-scoped service captured in a settings script and used in a task action cannot be re-resolved under the configuration cache"() {
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
@@ -32,7 +32,7 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
 
     def "service lookup in an owner-first task closure fails as an unsupported script reference at execution time"() {
         given:
-        buildFile << """
+        buildFile """
             tasks.register("some") {
                 onlyIf {
                     service(ProviderFactory)
@@ -61,7 +61,7 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
 
     def "looking up a settings-only service through a task closure resolves to the task and is rejected at configuration time"() {
         given:
-        settingsFile << """
+        settingsFile """
             gradle.rootProject {
                 tasks.register("useLayout") {
                     // `service` here resolves to the Task receiver, which does not expose the
@@ -89,7 +89,7 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
 
     def "a settings-scoped service captured in a settings script and used in a task action cannot be re-resolved under the configuration cache"() {
         given:
-        settingsFile << """
+        settingsFile """
             // Captured at settings-script scope, where `service` resolves to the Settings receiver
             // and BuildLayout is available. The instance is then used from a task action.
             def captured = service(BuildLayout)
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
@@ -59,34 +59,6 @@ class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCa
         }
     }
 
-    def "looking up a settings-only service through a task closure resolves to the task and is rejected at configuration time"() {
-        given:
-        settingsFile """
-            gradle.rootProject {
-                tasks.register("useLayout") {
-                    // `service` here resolves to the Task receiver, which does not expose the
-                    // settings-only BuildLayout, so this fails while the task is being configured.
-                    def captured = service(BuildLayout)
-                    doLast {
-                        // Tripwire on the first line: if the action is ever entered, this prints.
-                        println("REACHED ACTION")
-                        println("settings dir: " + captured.settingsDirectory.asFile.name)
-                    }
-                }
-            }
-        """
-
-        when:
-        fails ":useLayout"
-
-        then:
-        // The rejection happens while the task is being created, not while it runs...
-        failure.assertHasCause("Could not create task ':useLayout'.")
-        failure.assertHasCause("org.gradle.api.file.BuildLayout is not available in tasks.\nIt is available in settings scripts and settings plugins.")
-        // ...so the task action is never entered.
-        outputDoesNotContain("REACHED ACTION")
-    }
-
     def "a settings-scoped service captured in a settings script survives the configuration cache and is usable in a task action"() {
         given:
         settingsFile """
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
@@ -109,4 +109,26 @@ class IsolatedProjectsServiceLookupIntegrationTest extends AbstractIsolatedProje
         }
         outputContains("settings dir name: " + testDirectory.name)
     }
+
+    def "can capture a settings-scoped service and use it from an isolated project action"() {
+        createDirs("a")
+        settingsFile """
+            include("a")
+            // Captured at settings scope and used from an isolated 'beforeProject' action,
+            // which is serialized per project. The captured BuildLayout must survive that.
+            def captured = service(BuildLayout)
+            gradle.lifecycle.beforeProject { project ->
+                println(project.path + " settings dir: " + captured.settingsDirectory.asFile.name)
+            }
+        """
+
+        when:
+        isolatedProjectsRun(":a:help")
+
+        then:
+        fixture.assertStateStored {
+            projectsConfigured(":", ":a")
+        }
+        outputContains(":a settings dir: " + testDirectory.name)
+    }
 }
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.internal.cc.impl.isolated
+
+import spock.lang.Issue
+
+@Issue("https://github.com/gradle/gradle/issues/13121")
+class IsolatedProjectsServiceLookupIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
+
+    def "can look up a service in a task action of the owning project"() {
+        createDirs("a")
+        settingsFile """
+            include("a")
+        """
+        file("a/thing.txt").text = "content"
+        file("a/build.gradle") << """
+            tasks.register("cleanThing") {
+                doLast {
+                    service(FileSystemOperations).delete {
+                        delete("thing.txt")
+                    }
+                }
+            }
+        """
+
+        when:
+        isolatedProjectsRun(":a:cleanThing")
+
+        then:
+        fixture.assertStateStored {
+            projectsConfigured(":", ":a")
+        }
+        !file("a/thing.txt").exists()
+    }
+
+    def "can look up a service at configuration time of the owning project"() {
+        createDirs("a")
+        settingsFile """
+            include("a")
+        """
+        file("a/build.gradle") << """
+            def dirName = service(ProjectLayout).projectDirectory.asFile.name
+            tasks.register("show") {
+                doLast {
+                    println("project dir name: " + dirName)
+                }
+            }
+        """
+
+        when:
+        isolatedProjectsRun(":a:show")
+
+        then:
+        fixture.assertStateStored {
+            projectsConfigured(":", ":a")
+        }
+        outputContains("project dir name: a")
+    }
+
+    def "reports a problem when a build script looks up a service on another project"() {
+        createDirs("a", "b")
+        settingsFile """
+            include("a")
+            include("b")
+        """
+        buildFile """
+            project(':a').service(ObjectFactory)
+        """
+
+        when:
+        isolatedProjectsFailsUsing(mode, "help")
+
+        then:
+        fixture.assertIsolatedProjectsProblems(mode) {
+            projectsConfigured(":", ":a", ":b")
+            problem("Build file 'build.gradle': line 2: Project ':' cannot access 'Project.service' functionality on another project ':a'")
+        }
+
+        where:
+        mode << ALL_MODES
+    }
+
+    def "can look up a service in a settings script"() {
+        settingsFile """
+            def layout = service(BuildLayout)
+            println("settings dir name: " + layout.settingsDirectory.asFile.name)
+        """
+
+        when:
+        isolatedProjectsRun("help")
+
+        then:
+        fixture.assertStateStored {
+            projectsConfigured(":")
+        }
+        outputContains("settings dir name: " + testDirectory.name)
+    }
+}
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
@@ -26,7 +26,7 @@ class IsolatedProjectsServiceLookupIntegrationTest extends AbstractIsolatedProje
             include("a")
         """
         file("a/thing.txt").text = "content"
-        file("a/build.gradle") << """
+        buildFile("a/build.gradle", """
             tasks.register("cleanThing") {
                 doLast {
                     service(FileSystemOperations).delete {
@@ -34,7 +34,7 @@ class IsolatedProjectsServiceLookupIntegrationTest extends AbstractIsolatedProje
                     }
                 }
             }
-        """
+        """)
 
         when:
         isolatedProjectsRun(":a:cleanThing")
@@ -43,22 +43,22 @@ class IsolatedProjectsServiceLookupIntegrationTest extends AbstractIsolatedProje
         fixture.assertStateStored {
             projectsConfigured(":", ":a")
         }
+        and:
         !file("a/thing.txt").exists()
     }
 
     def "can look up a service at configuration time of the owning project"() {
-        createDirs("a")
         settingsFile """
             include("a")
         """
-        file("a/build.gradle") << """
+        buildFile("a/build.gradle", """
             def dirName = service(ProjectLayout).projectDirectory.asFile.name
             tasks.register("show") {
                 doLast {
                     println("project dir name: " + dirName)
                 }
             }
-        """
+        """)
 
         when:
         isolatedProjectsRun(":a:show")
@@ -67,6 +67,7 @@ class IsolatedProjectsServiceLookupIntegrationTest extends AbstractIsolatedProje
         fixture.assertStateStored {
             projectsConfigured(":", ":a")
         }
+        and:
         outputContains("project dir name: a")
     }
 
@@ -106,6 +107,7 @@ class IsolatedProjectsServiceLookupIntegrationTest extends AbstractIsolatedProje
         fixture.assertStateStored {
             projectsConfigured(":")
         }
+        and:
         outputContains("settings dir name: " + testDirectory.name)
     }
 
@@ -128,6 +130,7 @@ class IsolatedProjectsServiceLookupIntegrationTest extends AbstractIsolatedProje
         fixture.assertStateStored {
             projectsConfigured(":", ":a")
         }
+        and:
         outputContains(":a settings dir: " + testDirectory.name)
     }
 }
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
@@ -22,7 +22,6 @@ import spock.lang.Issue
 class IsolatedProjectsServiceLookupIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
 
     def "can look up a service in a task action of the owning project"() {
-        createDirs("a")
         settingsFile """
             include("a")
         """
--- a/platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/BuildLayoutCodec.kt
+++ b/platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/BuildLayoutCodec.kt
@@ -18,12 +18,9 @@ package org.gradle.internal.serialize.codecs.core
 
 import org.gradle.api.file.BuildLayout
 import org.gradle.api.file.Directory
-import org.gradle.api.internal.GeneratedSubclasses
+import org.gradle.internal.serialize.graph.Codec
 import org.gradle.internal.serialize.graph.ReadContext
 import org.gradle.internal.serialize.graph.WriteContext
-import org.gradle.internal.serialize.graph.codecs.Decoding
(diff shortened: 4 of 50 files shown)


reinsch82:
@coderabbitai review

subprojects/core-api/src/main/java/org/gradle/api/Task.java
```diff
@@ -811,4 +811,42 @@ Task doLast(@DelegatesTo(Task.class)
...
+     * use {@link org.gradle.api.services.ServiceReference} to access those.</p>
+     *
+     * @param serviceType the type of the service to look up
+     * @param <T> the service type
+     * @return the service instance. Never returns null.
+     * @throws InvalidUserDataException when the given type is not one of the services available in this scope
+     * @since 9.8.0
+     */
+    @Incubating
+    <T> T service(Class<T> serviceType);
```
mlopatkin: :x: Same comment about the upper bound.

subprojects/core-api/src/main/java/org/gradle/api/Project.java
```diff
@@ -1112,6 +1112,35 @@ Project project(String path,
...
+     *
+     * @param serviceType the type of the service to look up
+     * @param <T> the service type
+     * @return the service instance. Never returns null.
+     * @throws InvalidUserDataException when the given type is not one of the services available in this scope
+     * @since 9.8.0
+     */
+    @Incubating
+    @HiddenInDefinition
+    <T> T service(Class<T> serviceType);
```
mlopatkin: :x: Same comment about the upper bound.

subprojects/core-api/src/main/java/org/gradle/api/initialization/Settings.java
```diff
@@ -306,6 +306,32 @@ default void includeFlat(String... projectNames) {
...
+     *
+     * @param serviceType the type of the service to look up
+     * @param <T> the service type
+     * @return the service instance. Never returns null.
+     * @throws org.gradle.api.InvalidUserDataException when the given type is not one of the services available in this scope
+     * @since 9.8.0
+     */
+    @Incubating
+    @HiddenInDefinition
+    <T> T service(Class<T> serviceType);
```
mlopatkin: :x: I don't think we should even start with unrestricted service type. It makes the API much harder to navigate. A marker interface like `SettingsInjectable` and an upper bound `<T extends SettingsInjectable>` would improve usability and discoverability a lot.

We already suffer with `Object` accepting methods and overly permissive bounds, let's not add another one. It is much easier to relax it in the future than to tighten.

subprojects/core-api/src/main/java/org/gradle/api/invocation/Gradle.java
```diff
@@ -417,4 +417,28 @@ public interface Gradle extends PluginAware, ExtensionAware {
...
+     * use {@link #getSharedServices()} to access those.</p>
+     *
+     * @param serviceType the type of the service to look up
+     * @param <T> the service type
+     * @return the service instance. Never returns null.
+     * @throws org.gradle.api.InvalidUserDataException when the given type is not one of the services available in this scope
+     * @since 9.8.0
+     */
+    @Incubating
+    <T> T service(Class<T> serviceType);
```
mlopatkin: ❌ Same comment about the upper bound.

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/service_injection.adoc
```diff
@@ -141,6 +141,59 @@ The following services are available for injection in settings plugins, extensio
...
+
+Relative paths passed to a looked-up `FileSystemOperations` or `ArchiveOperations` are resolved the same way as for the enclosing scope: against the project directory for a project script or task action, and against the build's root directory for a settings or init script.
+
+=== Cross-project access and Isolated Projects
+
+`service(...)` resolves against the scope in which it is called.
+Calling it on another project — for example `project(":other").service(ObjectFactory)` — reaches across the project boundary and is reported as a problem under <<isolated_projects.adoc#isolated_projects,Isolated Projects>>, exactly like `project(":other").getObjects()`.
+Look services up from the owning project instead.
+
+NOTE: A service that is only available in the settings scope (such as `BuildLayout`) can be captured in a settings script and referenced from a task action, but it cannot be restored when that task runs from the Configuration Cache, because it is re-resolved against the task's own project scope. Look up project-scoped services (or `ProjectLayout`) from the task instead.
```
mlopatkin: ❓ Can we make it work instead? FWIW, the data exposed in `BuildLayout` is immutable anyway and there is no risk of confusion.

Mixing different kind of path-aware services can be a problem, though. I'd say we should make it work properly too (preserving the original scope).

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/service_injection.adoc
```diff
@@ -141,6 +141,59 @@ The following services are available for injection in settings plugins, extensio
...
+
+[%header,cols="2,1,1,1"]
+|===
+| Service | Project scripts, plugins & task actions | Settings scripts & plugins | Init scripts & `Gradle` plugins
+
+| <<objectfactory,`ObjectFactory`>>               | ✓ | ✓ | ✓
+| <<providerfactory,`ProviderFactory`>>           | ✓ | ✓ | ✓
+| <<filesystemoperations,`FileSystemOperations`>> | ✓ | ✓ | ✓
+| <<archiveoperations,`ArchiveOperations`>>       | ✓ | ✓ | ✓
+| <<execoperations,`ExecOperations`>>             | ✓ | ✓ | ✓
```
mlopatkin: 🤔  There is little-to-no point in resolving `ExecOperations` from scopes other than `Task` if we take CC compatibility into account. Using it at configuration time is going to be deprecated. `buildFinished` and other deprecated callbacks may still use it in Vintage, but they're also on the way out. Let's start restrictive and consistent with our long-term vision rather than try to patch the new thing afterwards.

mlopatkin (commented):
Before I dive into reviewing the implementation, I have some conceptual concerns about the shape of the API.

subprojects/core-api/src/main/java/org/gradle/api/initialization/Settings.java
reinsch82: I get the point, I'm not sure how much more discoverable it is though.

A user would probably got to the definition of the method and either see `SettingsInjectable`, click on that, try to find derived classes.
Or he would got to the definition and see the list of injectable classes in the java doc.

It would be runtime checked in this case.

lkasso (approved):
Docs LGTM. I added an upgrade guide note, cleaned up the new docs and release notes, and added some cross-references. @reinsch82, please double-check my changes.

subprojects/core-api/src/main/java/org/gradle/api/initialization/Settings.java
eskatos: This would give the IDE enough information to filter a completion list of types, also compilation-time checks would catch wrong services much earlier. +1 to marker interfaces

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/service_injection.adoc
```diff
@@ -141,6 +141,64 @@ The following services are available for injection in settings plugins, extensio
 | Provides access to important locations for a Gradle build, such as the root directory and settings file. Use it from a settings plugin to resolve paths relative to the root settings directory.
 |===
 
+[[looking_up_services]]
+== Looking up services in scripts incubating-label:[]
+
+Instead of declaring an `@Inject` point, build, settings, and init scripts, as well as task actions, can look up a curated set of Gradle services directly with `service(Class)` in the Groovy DSL or `service<Type>()` in the Kotlin DSL.
+This avoids the ad-hoc `objects.newInstance(...)` ceremony shown in the per-service sections below, while remaining compatible with the <<configuration_cache.adoc#config_cache,Configuration Cache>> and <<isolated_projects.adoc#isolated_projects,Isolated Projects>>.
```
eskatos, line 148: ❓  "in the per-service sections below"

I think these sections should be updated to use the new service lookup instead of the problematic ceremony. Maybe we keep a dedicated section to demonstrate the alternative with more ceremony.

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/service_injection.adoc
```diff
@@ -141,6 +141,64 @@ The following services are available for injection in settings plugins, extensio
...
+| <<objectfactory,`ObjectFactory`>>               | ✓ | ✓ | ✓
+| <<providerfactory,`ProviderFactory`>>           | ✓ | ✓ | ✓
+| <<filesystemoperations,`FileSystemOperations`>> | ✓ | ✓ | ✓
+| <<archiveoperations,`ArchiveOperations`>>       | ✓ | ✓ | ✓
+| <<execoperations,`ExecOperations`>>             | ✓ | ✓ | ✓
+| <<sec:projectlayout,`ProjectLayout`>>           | ✓ |   |
+| <<buildlayout,`BuildLayout`>>                   |   | ✓ |
+|===
+
+Looking up a type that is not available in the current scope, an internal type, or a <<build_services.adoc#build_services,shared build service>> fails with an `InvalidUserDataException`.
```
eskatos: ❌ This is not valid anymore with the introduction of marker interfaces

subprojects/core-api/src/main/java/org/gradle/api/Project.java
```diff
@@ -1112,6 +1113,35 @@ Project project(String path,
...
+    /**
+     * Looks up a service provided by Gradle for use in this project.
+     *
+     * <p>The following services are available in project scripts and plugins:</p>
+     * <ul>
+     * <li>{@link org.gradle.api.model.ObjectFactory}</li>
+     * <li>{@link org.gradle.api.provider.ProviderFactory}</li>
+     * <li>{@link org.gradle.api.file.FileSystemOperations}</li>
+     * <li>{@link org.gradle.api.file.ArchiveOperations}</li>
+     * <li>{@link org.gradle.process.ExecOperations}</li>
```
eskatos: ❓  It seems to me that the list in this comment is already wrong with the last edits. This is also true on `Settings` and `Gradle`.

subprojects/core-api/src/main/java/org/gradle/api/services/TaskInjectable.java
```diff
@@ -0,0 +1,32 @@
...
+ * Marks a Gradle service that can be looked up from a task via
+ * {@link org.gradle.api.Task#service(Class)}, including from task actions at execution time.
+ *
+ * <p>This is a marker interface with no members. It exists so the service lookup can reject
+ * unavailable types at compile time; it is not meant to be implemented by user code.</p>
+ *
+ * @since 9.8.0
+ */
+@Incubating
+public interface TaskInjectable {
```
eskatos: ❓ The name of the marker interfaces is kind of confusing. The name uses "Injectable" but this is about lookup. A candidate that comes to mind: simply `TaskService`.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
```diff
@@ -0,0 +1,123 @@
...
+class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCacheIntegrationTest {
+
+    def configurationCache = new ConfigurationCacheFixture(this)
+
+    def "service lookup in an owner-first task closure fails as an unsupported script reference at execution time"() {
+        given:
+        buildFile << """
+            tasks.register("some") {
+                onlyIf {
+                    service(ProviderFactory)
```
mlopatkin: ❓ Why does it delegate to owner rather than the delegate `Task`? Is it consistent with `doFirst`/`doLast`?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupConfigurationCacheIntegrationTest.groovy
```diff
@@ -0,0 +1,123 @@
...
+                }
+            }
+        """
+
+        when: "run without the configuration cache, the captured settings-scoped service is used directly"
+        run ":useLayout"
+
+        then:
+        outputContains("REACHED ACTION")
+        outputContains("settings dir: " + testDirectory.name)
```
mlopatkin, line 54: ❓ What's the point of this "no-configuration-cache works" test?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
```diff
@@ -0,0 +1,123 @@
...
+                        // Tripwire on the first line: if the action is ever entered, this prints.
+                        println("REACHED ACTION")
+                        println("settings dir: " + captured.settingsDirectory.asFile.name)
+                    }
+                }
+            }
+        """
+
+        when:
+        fails ":useLayout"
```
mlopatkin: :x: This doesn't use configuration cache at all, why is the test here?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
```diff
@@ -0,0 +1,123 @@
...
+import spock.lang.Issue
+
+/**
+ * Pins the Groovy closure dispatch and configuration-cache behavior of the public {@code service(Class)}
+ * lookup in the corner cases where the receiver resolution or the captured instance interacts with the
+ * configuration cache. Happy-path and error-message coverage lives in the core
+ * {@code PublicServiceLookupIntegrationTest}.
+ */
+@Issue("https://github.com/gradle/gradle/issues/13121")
+class ServiceLookupGroovyDispatchIntegrationTest extends AbstractConfigurationCacheIntegrationTest {
```
mlopatkin: 💅 Please use `buildFile(String)` and `settingsFile(String)` helpers rather than `<<`.

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/service_injection.adoc
```diff
@@ -141,6 +141,66 @@ The following services are available for injection in settings plugins, extensio
...
+
+=== Cross-project access and Isolated Projects
+
+`service(...)` resolves against the scope in which it is called.
+Calling it on another project — for example `project(":other").service(ObjectFactory)` — reaches across the project boundary and is reported as a problem under <<isolated_projects.adoc#isolated_projects,Isolated Projects>>, exactly like `project(":other").getObjects()`.
+Look services up from the owning project instead.
+
+NOTE: A service that is only available in the settings scope (such as `BuildLayout`) can be captured in a settings script and referenced from a task action, but the captured reference cannot be restored when that task runs from the Configuration Cache.
+The reference is re-resolved against the task's own project scope, which does not expose settings-only services.
+Look up project-scoped services (or `ProjectLayout`) from the task instead.
```
mlopatkin: 🤔 IMO, this should be fixed rather than documented. Doesn't have to happen in this same PR. 

And this is not something new, if you get the reference through old-style injection trick, you'll get the same result, right? Maybe worth opening an issue and linking to it from "Not Yet Implemented" section of the CC documentation.

subprojects/core-api/src/main/java/org/gradle/api/services/GradleService.java
```diff
@@ -0,0 +1,32 @@
...
+package org.gradle.api.services;
+
+import org.gradle.api.Incubating;
+
+/**
+ * Marks a Gradle service that can be looked up from the build (Gradle) scope via
+ * {@link org.gradle.api.invocation.Gradle#service(Class)}, including from init scripts and Gradle plugins.
+ *
+ * <p>This is a marker interface with no members. It exists so the service lookup can reject
+ * unavailable types at compile time; it is not meant to be implemented by user code.</p>
```
mlopatkin: 🤔 I wouldn't put that much detail in the Javadoc. `it is not meant to be implemented by user code` is the only relevant part, IMO.

The same applies to the rest of the markers.

subprojects/core-api/src/main/java/org/gradle/api/Project.java
```diff
@@ -1112,6 +1113,34 @@ Project project(String path,
...
+     * <p>The following services are available in project scripts and plugins:</p>
+     * <ul>
+     * <li>{@link org.gradle.api.model.ObjectFactory}</li>
+     * <li>{@link org.gradle.api.provider.ProviderFactory}</li>
+     * <li>{@link org.gradle.api.file.FileSystemOperations}</li>
+     * <li>{@link org.gradle.api.file.ArchiveOperations}</li>
+     * <li>{@link org.gradle.api.file.ProjectLayout}</li>
+     * </ul>
+     *
+     * <p>The lookup can be used both at configuration time and at execution time. The returned instance
```
mlopatkin: :x: Maybe we shouldn't document that the lookup on `Project` can be used at execution time. This is not going to be CC-compatible.

subprojects/core-api/src/main/java/org/gradle/api/Project.java
```diff
@@ -1112,6 +1113,34 @@ Project project(String path,
...
+     * may be captured in a task action and is safe to use with the configuration cache.</p>
+     *
+     * <p>This method does not provide access to {@link org.gradle.api.services.BuildService shared build services};
+     * use {@link org.gradle.api.services.ServiceReference} to access those.</p>
```
mlopatkin: 💅  We can also link `Gradle#getSharedServices` as it is available in the Project scope. But not for the `Task.service`.

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,275 @@
...
+            ]
+            println("resolved services: " + services.count { it != null })
+        """
+
+        expect:
+        succeeds("help")
+        outputContains("resolved services: 5")
+    }
+
+    def "can delete files with FileSystemOperations looked up inside a task action"() {
```
mlopatkin: 💅 Why is `delete` special? I suppose it is just an arbitrary service to demonstrate lookup? IMO, the test name should reflect that then.

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,275 @@
...
+import org.gradle.integtests.fixtures.AbstractIntegrationSpec
+import org.gradle.internal.jvm.Jvm
+import org.gradle.util.internal.TextUtil
+import spock.lang.Issue
+
+@Issue("https://github.com/gradle/gradle/issues/13121")
+class PublicServiceLookupIntegrationTest extends AbstractIntegrationSpec {
+
+    def "all documented services can be looked up in a build script"() {
+        buildFile << """
```
mlopatkin: 💅 `buildFile()`.
```suggestion
        buildFile """
```

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,275 @@
...
+        "ArchiveOperations"    | 'println("out: " + captured.zipTree("stuff.zip").files*.name.sort())'                        | "out: [entry.txt]"
+        "ExecOperations"       | 'def r = captured.exec { commandLine("JAVA_EXE", "-version") }; println("out: " + r.exitValue)' | "out: 0"
+        "ProjectLayout"        | 'println("out: " + captured.projectDirectory.asFile.name)'                                   | "out: @PROJECT_DIR_NAME@"
+    }
+
+    private static String getJvmPath() {
+        return TextUtil.escapeString(Jvm.current().javaExecutable.absolutePath)
+    }
+
+    def "FileSystemOperations from a project script resolves paths relative to the project directory"() {
```
mlopatkin, line 155: 🤔 It would be great to have three scopes - `Gradle`, `Project`, `Task` and verify that each resolves properly.

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,275 @@
...
+    def "looking up a project-only service from a settings script fails with a helpful message"() {
+        settingsFile << """
+            service(ProjectLayout)
+        """
+
+        when:
+        fails("help")
+
+        then:
+        failure.assertHasCause("org.gradle.api.file.ProjectLayout is not available in settings scripts and plugins. It is available in project scripts and plugins and tasks.")
```
mlopatkin: 🎉 Nice

💅 To me `It is available in X and Y and Z` feels a bit verbose, I'd rather stick with just `in X, Y, Z`.

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,275 @@
...
+            }
+
+            service(CounterService)
+        """
+
+        when:
+        fails("help")
+
+        then:
+        failure.assertHasCause("CounterService is a shared build service, which cannot be obtained with service().")
```
mlopatkin: 🤔 The test name promised a reference to build service api, but there is none.

subprojects/core/src/main/java/org/gradle/api/internal/services/PublicServiceLookups.java
```diff
@@ -0,0 +1,134 @@
...
+        .build();
+
+    private PublicServiceLookups() {
+    }
+
+    /**
+     * Visible for testing: the authoritative service→scopes allowlist, so the marker/bound consistency
+     * test can assert the compile-time markers and method bounds agree with it.
+     */
+    static ImmutableMap<Class<?>, ImmutableSet<EntryPoint>> availableServices() {
```
mlopatkin, line 96: 💅 `@VisibleForTesting` should be available here. We already use it in places.

subprojects/core/src/main/java/org/gradle/api/internal/services/PublicServiceLookups.java
```diff
@@ -0,0 +1,134 @@
...
+        return AVAILABLE_SERVICES;
+    }
+
+    public static <T> T lookup(@Nullable Class<T> serviceType, EntryPoint entryPoint, ServiceRegistry services) {
```
mlopatkin, line 118: ❓ A `@Nullable` that throws is a bit questionable, as this isn't exactly a public API boundary. OTOH, it centralizes the argument handling. Let's at least have a javadoc here.

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,275 @@
...
+        outputContains(expectedOutput.replace("@PROJECT_DIR_NAME@", testDirectory.name))
+
+        where:
+        serviceType            | action                                                                                        | expectedOutput
+        "ObjectFactory"        | 'def p = captured.property(String); p.set("captured-value"); println("out: " + p.get())'     | "out: captured-value"
+        "ProviderFactory"      | 'println("out: " + captured.provider { "provided-value" }.get())'                            | "out: provided-value"
+        "FileSystemOperations" | 'captured.delete { delete(targetFile) }; println("out: exists=" + targetFile.exists())'      | "out: exists=false"
+        "ArchiveOperations"    | 'println("out: " + captured.zipTree("stuff.zip").files*.name.sort())'                        | "out: [entry.txt]"
+        "ExecOperations"       | 'def r = captured.exec { commandLine("JAVA_EXE", "-version") }; println("out: " + r.exitValue)' | "out: 0"
+        "ProjectLayout"        | 'println("out: " + captured.projectDirectory.asFile.name)'                                   | "out: @PROJECT_DIR_NAME@"
```
mlopatkin: 💅 column alignment is a bit off

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,275 @@
...
+        file("zip-src/entry.txt").text = "entry"
+        file("zip-src").zipTo(file("stuff.zip"))
+        buildFile << """
+            tasks.register("use") {
+                def captured = service($serviceType)
+                def targetFile = file("doomed.txt")
+                doLast {
+                    ${action.replace("JAVA_EXE", jvmPath)}
+                }
+            }
```
mlopatkin: 💅 Perhaps writing all the tasks in full and parameterizing by the name of the task to invoke would be more readable.

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,275 @@
...
+
+        when:
+        succeeds(":sub:cleanLocal")
+
+        then:
+        !file("sub/local.txt").exists()
+        file("local.txt").exists()
+    }
+
+    def "services can be looked up in a settings script"() {
```
mlopatkin: 🤔 I'd expect the test to verify every available service, like the Project-scoped test above does.

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,275 @@
...
+            println("settings property: " + property.get())
+        """
+
+        expect:
+        succeeds("help")
+        outputContains("settings dir name: " + testDirectory.name)
+        outputContains("settings property: from-settings")
+    }
+
+    def "services can be looked up in an init script"() {
```
mlopatkin: 🤔 I'd expect the test to verify every available service, like the Project-scoped test above does.

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,275 @@
...
+        buildFile << """
+            service(org.gradle.api.internal.project.ProjectInternal)
+        """
+
+        when:
+        fails("help")
+
+        then:
+        failure.assertHasCause("org.gradle.api.internal.project.ProjectInternal is not a service that is available for lookup with service(). " +
+            "The following services are available in project scripts and plugins: " +
```
mlopatkin: 💅 Given how many are there, maybe use EOL instead of a `,` for the available service enumeration?

subprojects/core/src/test/groovy/org/gradle/api/internal/services/PublicServiceLookupMarkerConsistencyTest.groovy
```diff
@@ -0,0 +1,65 @@
...
+import org.gradle.api.services.SettingsService
+import org.gradle.api.services.TaskService
+import spock.lang.Specification
+
+/**
+ * Pins the two representations of "which service is available in which scope" together: the runtime
+ * allowlist in {@link PublicServiceLookups} (authoritative), the per-scope compile-time marker interfaces,
+ * and the generic bound on each {@code service(Class)} member. If any drifts from the others this fails.
+ */
+class PublicServiceLookupMarkerConsistencyTest extends Specification {
```
mlopatkin, line 34: 🎉 Good tests here

mlopatkin (changes requested):
Looks good overall! I spotted few quirks in the tests and scaffolding, but the implementation is solid.
 
A common theme:
- 💅 It might be worth it to cross-link each service's Javadoc to the `#service` methods.

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
reinsch82: you mean like

```
failure.assertHasCause("org.gradle.api.file.ProjectLayout is not available in settings scripts and settings plugins." +
"\nIt is available in project scripts, project plugins, and tasks.")
```
?

mlopatkin: > you mean like
> 
> ```
> failure.assertHasCause("org.gradle.api.file.ProjectLayout is not available in settings scripts and settings plugins." +
> "\nIt is available in project scripts, project plugins, and tasks.")
> ```
> 
> ?

Not that one - it has only a short list, but the "The following services are available in project scripts and plugins:" specifically. I thing something like
```
"org.gradle.api.internal.project.ProjectInternal is not a service that is available for lookup with service(). The following services are available in project scripts and plugins:\n" +
" - org.gradle.api.model.ObjectFactory\n" +
" - org.gradle.api.provider.ProviderFactory\n" +
" - org.gradle.api.file.FileSystemOperations\n" +
" - org.gradle.api.file.ArchiveOperations\n" +
" - org.gradle.api.file.ProjectLayout"
```

subprojects/core/src/main/java/org/gradle/api/internal/services/PublicServiceLookups.java
```diff
@@ -0,0 +1,147 @@
...
+
+    private static String wrongScopeMessage(Class<?> serviceType, EntryPoint entryPoint, Set<EntryPoint> availableIn) {
+        return String.format(
+            "%s is not available in %s. It is available in %s.",
+            serviceType.getName(), entryPoint.getDisplayName(),
+            availableIn.stream().map(EntryPoint::getDisplayName).collect(Collectors.joining(" and ")));
+    }
+
+    private static String servicesAvailableIn(EntryPoint entryPoint) {
+        return AVAILABLE_SERVICES.entrySet().stream()
```
mlopatkin, line 156: 💅  Maybe also sort the services by name?

subprojects/core/src/main/java/org/gradle/api/internal/services/PublicServiceLookups.java
reinsch82: Good point — it's not a normal public-API boundary, but for this parameter it effectively is one. serviceType is the only argument that carries user input: it's forwarded straight from the service(x) call, whereas entryPoint and services are always supplied by our own call sites (EntryPoint.PROJECT etc. + getServices()), so those stay non-null by construction.

The reason `null` can reach serviceType specifically is Groovy: service(null) in a build script is dynamically dispatched, so none of the compile-time protection applies. Kotlin rejects it (the reified service<T>() extension takes no argument, and the Java overload's parameter is non-null under @NullMarked), but Groovy has no such gate — so lookup is the single place that can catch it and turn it into a helpful InvalidUserDataException. There's a test for exactly this ("looking up a null service type fails with a helpful message").

So the @Nullable-that-throws is deliberate: it documents that null is an expected user input here, not an internal precondition. Agreed it deserves to be spelled out though — I've added a @param explaining the above.

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/service_injection.adoc
reinsch82: made it work.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
reinsch82: to my understanding this is a long standing inconsistency.

jvandort:
So I understand why we'd put this on `Project`, `Gradle`, and `Settings`, they are targets of build scripts, when writing build logic in a script there is no "object" to instantiate, and therefore no place to inject. This applies to both precompiled script plugins and non-precompiled regular scripts. I think this justifies putting this method on `Project`, `Gradle`, and `Settings`. 

But for task, I do not completely follow the justification. For a task you are either extending `DefaultTask`, in which case you can inject whatever you want, or you are doing the `doLast` thing. If this is _only_ to support `doLast` in tasks, I would first want to verify that we as a team think building tasks "adhoc" like this using `doLast` is _is good practice at all_. 

It is my understanding that we generally promote extending DefaultTask rather than adhoc task definitions like this. I would prefer we don't add things that make it easier to do things we don't want to do. If I am misunderstanding and the sort of "adhoc doLast task definition" isn't quite as "taboo" as I thought, then I suppose I am not as concerned about adding this to `Task` as well.

mlopatkin:
> It is my understanding that we generally promote extending DefaultTask rather than adhoc task definitions like this. 

`doLast` is not only for purely "ad hoc" tasks. Quite often this is the only (or the most straightforward) way to customize outputs of the task you as a build author don't own. A few examples:
- Customizing generated plugin descriptors in plugin-compatibility-plugin: https://github.com/gradle/gradle-plugin-compatibility-plugin/blob/95d55dd925ea89a8ac790162ac141ccefff5e49c/src/main/java/org/gradle/plugin/compatibility/internal/CompatibilityPlugin.java#L48-L61
- Customizing Windows start scripts: https://blog.mrhaki.com/2015/04/gradle-goodness-alter-start-scripts.html (dated, but probably still the only option)

"Hack everything in the build script" can quickly become a mess, I agree, and the bigger the project goes, the stricter it should be with its build logic. However, for quick starts this approach is still appealing. The overhead of introducing a first proper plugin is still there.

And, IMO, even in environments where `@Inject` is available, the new locator API can be a more straightforward choice and might improve discoverability of the services.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
mlopatkin: > to my understanding this is a long standing inconsistency.

🤔 I looked a bit more into that, and you're right. I suggest to refactor the test. Rather than capturing the "broken" behavior, we can have a non-CC-specific test in `PublicServiceLookupIntegTest` marked with `@ToBeFixedForConfigurationCache`.

subprojects/core/src/main/java/org/gradle/api/internal/services/PublicServiceLookups.java
mlopatkin: > 💅 Maybe also sort the services by name?

This still holds

subprojects/core/src/integTest/groovy/org/gradle/api/PublicServiceLookupIntegrationTest.groovy
mlopatkin: > > you mean like
> > ```
> > failure.assertHasCause("org.gradle.api.file.ProjectLayout is not available in settings scripts and settings plugins." +
> > "\nIt is available in project scripts, project plugins, and tasks.")
> > ```
> ```
> "org.gradle.api.internal.project.ProjectInternal is not a service that is available for lookup with service(). The following services are available in project scripts and plugins:\n" +
> " - org.gradle.api.model.ObjectFactory\n" +
> " - org.gradle.api.provider.ProviderFactory\n" +
> " - org.gradle.api.file.FileSystemOperations\n" +
> " - org.gradle.api.file.ArchiveOperations\n" +
> " - org.gradle.api.file.ProjectLayout"
> ```

❓ Not sure if that got a resolution.

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/working_with_files.adoc
```diff
@@ -641,6 +641,7 @@ Keep an eye on the number of arguments to distinguish between them.
 
 WARNING: Using the `Project.copy` method at execution time, as described here, is not compatible with the <<configuration_cache_requirements.adoc#config_cache:requirements:use_project_during_execution,configuration cache>>.
 A possible solution is to implement the task as a proper class and use link:{javadocPath}/org/gradle/api/file/FileSystemOperations.html#copy-org.gradle.api.Action-[FileSystemOperations.copy] method instead, as described in the <<configuration_cache_debugging.adoc#config_cache:troubleshooting,configuration cache chapter>>.
+Scripts and ad-hoc tasks can also obtain `FileSystemOperations` (or `ArchiveOperations`) directly with the <<service_injection.adoc#looking_up_services,`service(...)` lookup>>.
```
mlopatkin, line 644: 🤔  That example should be rewritten after the new API stabilizes. I think we usually don't hesitate to use Incubating APIs in the documentation, but I might be wrong. If so, a follow-up issue like "Promote service lookup API to stable" with the list of the docs to update can be helpful.

However, given that we're updating "service injection" doc page, we can rewrite this snippet too.

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/service_injection.adoc
```diff
@@ -141,6 +141,64 @@ The following services are available for injection in settings plugins, extensio
...
+Look services up from the owning project instead.
+
+NOTE: A service that is only available in the settings scope (such as `BuildLayout`) can be captured in a settings script and referenced from a task action, and the captured instance is preserved when that task runs from the Configuration Cache.
+
+=== When to use a lookup
+
+Prefer a looked-up `FileSystemOperations`, `ArchiveOperations`, or `ExecOperations` over hand-rolled `java.io`/`java.nio` code or `Runtime.exec(...)`.
+The Gradle services give you path resolution, copy and sync specs, and robust deletion, and they keep your code compatible with the Configuration Cache.
+
+`ExecOperations` can be looked up only from a task action, reflecting that running external processes at configuration time is incompatible with the Configuration Cache.
```
mlopatkin: 🤔  This isn't actually enforced, right?

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/BuildLayoutCodec.kt
```diff
@@ -0,0 +1,72 @@
...
+ * task action, so it must survive the configuration cache. The general [org.gradle.internal.serialize.graph.codecs.ServicesCodec]
+ * cannot help here: it re-resolves a service from the isolate owner's registry on load, and a task's
+ * registry is Project-scoped, a sibling of the Settings scope, so the settings-scoped [BuildLayout]
+ * is unreachable (and the Settings registry no longer exists at execution time). Instead, this codec
+ * writes the two directories and reconstructs an immutable value on load.
+ *
+ * Must be bound ahead of [org.gradle.internal.serialize.graph.codecs.ServicesCodec] so that it wins
+ * for this type (first matching binding is used).
+ */
+object BuildLayoutCodec : EncodingProducer, Decoding {
```
mlopatkin: 🤔 A bit kludgy, but okay for now, until we add a proper scoping support there.

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/BuildLayoutCodec.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    override fun encodingForType(type: Class<*>): Encoding? =
```
mlopatkin: ❓ Do we actually need the `encodingForType` dance? It looks like just extending `Codec<BuildLayout>` would be sufficient.

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -79,6 +79,25 @@ Gradle provides an intuitive [command-line interface](userguide/command_line_int
 ### Build authoring improvements
 Gradle provides [rich APIs](userguide/getting_started_dev.html) for build engineers and plugin authors, enabling the creation of custom, reusable build logic and better maintainability.
 
+#### Look up Gradle services from scripts and task actions
```
mlopatkin, line 82: ❓ We usually don't do upgrade guide entries for new functions, even though they may cause name clashes. Do you think there is an increased risk this time?

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/service_injection.adoc
```diff
@@ -321,17 +382,8 @@ include::sample[dir="snippets/reference/gradle-types/services/groovy",files="bui
 
 The `ArchiveOperations` service is injected into the `MyArchiveOperationsTask` task's constructor using the `@Inject` annotation.
 
-With some ceremony, it is possible to use `ArchiveOperations` in an ad-hoc task defined in a build script:
-
-====
-include::sample[dir="snippets/reference/gradle-types/services/kotlin",files="build.gradle.kts[tags=archive-op-adhoc]"]
```
mlopatkin, line 327: :x: If we're removing the use of the snippet, let's remove the snippet itself. And maybe we can add a new snippet that uses `lookup()` here.

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/service_injection.adoc
```diff
@@ -321,17 +382,8 @@ include::sample[dir="snippets/reference/gradle-types/services/groovy",files="bui
...
-include::sample[dir="snippets/reference/gradle-types/services/groovy",files="build.gradle[tags=archive-op-adhoc]"]
-====
-
-First, you need to declare an interface with a property of type `ArchiveOperations`, here named `InjectedArcOps`, to serve as an injection point.
-Then call the method link:{javadocPath}/org/gradle/api/model/ObjectFactory.html#newInstance(java.lang.Class,java.lang.Object%2E%2E%2E)[`ObjectFactory.newInstance`] to generate an implementation of the interface that holds an injected service.
-
-TIP: This is a good time to consider extracting the ad-hoc task into a proper class.
+To use `ArchiveOperations` in an ad-hoc task without writing a dedicated class, look it up with `service()` — see <<looking_up_services,Looking up services in scripts>>.
+For a service that is not available through the lookup, use the `ObjectFactory.newInstance` injection point shown for <<filesystemoperations,`FileSystemOperations`>>.
```
mlopatkin: ❓ This section is about `ArchiveOperations`, why should we talk about other services?

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/service_injection.adoc
```diff
@@ -348,17 +400,9 @@ include::sample[dir="snippets/reference/gradle-types/services/groovy",files="bui
...
-include::sample[dir="snippets/reference/gradle-types/services/groovy",files="build.gradle[tags=exec-op-adhoc]"]
-====
-
-First, you need to declare an interface with a property of type `ExecOperations`, here named `InjectedExecOps`, to serve as an injection point.
-Then call the method link:{javadocPath}/org/gradle/api/model/ObjectFactory.html#newInstance(java.lang.Class,java.lang.Object%2E%2E%2E)[`ObjectFactory.newInstance`] to generate an implementation of the interface that holds an injected service.
-
-TIP: This is a good time to consider extracting the ad-hoc task into a proper class.
+To run an external process from an ad-hoc task, look `ExecOperations` up with `service()` from the task action — see <<looking_up_services,Looking up services in scripts>>.
+It can be looked up only from a task action, not at configuration time.
+For a service that is not available through the lookup, use the `ObjectFactory.newInstance` injection point shown for <<filesystemoperations,`FileSystemOperations`>>.
```
mlopatkin: ❓ Same as above, this is about `ExecOperations`.

mlopatkin (changes requested):
Just a couple more comments.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
reinsch82: but can we actually fix it?

it would break existing code probably and deprecation warnings are also tricky, we'd need to introduce an alternative API for that right?

platforms/documentation/docs/src/docs/release/notes.md
reinsch82: there is precedent for this kind of entry
https://github.com/gradle/gradle/blob/bf157a90413033e232bc5a1a04ecb15a2c78787c/platforms/documentation/docs/src/docs/userguide/releases/upgrading/upgrading_version_8.adoc#L3108
cc @lkasso

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
mlopatkin: > but can we actually fix it?
> 
> it would break existing code probably and deprecation warnings are also tricky, we'd need to introduce an alternative API for that right?

This can be tricky, but I think we won't actually change the observed lookup order in `onlyIf`.

reinsch82: Digging in, the only way service() hits this CC failure is onlyIf's owner-first closure resolving to the script — design-aligned usage (config block, doFirst/doLast) always resolves to the task and is CC-safe. So the failure is really the general "no script reference at execution" rule, not something specific to service(), and @ToBeFixedForConfigurationCache would enshrine the accidental owner-first resolution as intended. I'll remove the test rather than pin a behavior that's at odds with the design; the design-aligned cases stay covered by the capture and by-value tests.

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/service_injection.adoc
reinsch82: reworded

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
mlopatkin: >Digging in, the only way service() hits this CC failure is onlyIf's owner-first closure resolving to the script — design-aligned usage (config block, doFirst/doLast) always resolves to the task and is CC-safe. 

I'd argue that `onlyIf` is also intended usage for `service` (and should resolve to a task), so it is worth to cover it by a test too.

>@ToBeFixedForConfigurationCache would enshrine the accidental owner-first resolution as intended

I think AI got this completely backwards.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,134 @@
...
+
+package org.gradle.internal.cc.impl.isolated
+
+import spock.lang.Issue
+
+@Issue("https://github.com/gradle/gradle/issues/13121")
+class IsolatedProjectsServiceLookupIntegrationTest extends AbstractIsolatedProjectsIntegrationTest {
+
+    def "can look up a service in a task action of the owning project"() {
+        createDirs("a")
```
6hundreds: 💅 Do you need to create it explicitly tho? I'm not sure

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,134 @@
...
+        settingsFile """
+            include("a")
+        """
+        file("a/thing.txt").text = "content"
+        file("a/build.gradle") << """
```
6hundreds: 💅 Use `buildFile` API instead

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,134 @@
...
+        """
+
+        when:
+        isolatedProjectsRun(":a:cleanThing")
+
+        then:
+        fixture.assertStateStored {
+            projectsConfigured(":", ":a")
+        }
+        !file("a/thing.txt").exists()
```
6hundreds, line 47: 💅 I'd say
```
and:
!file("a/thing.txt").exists()
```

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
6hundreds: The tests below are also may be fixed for consistency

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
```diff
@@ -0,0 +1,134 @@
...
+        outputContains("settings dir name: " + testDirectory.name)
+    }
+
+    def "can capture a settings-scoped service and use it from an isolated project action"() {
+        createDirs("a")
+        settingsFile """
+            include("a")
+            // Captured at settings scope and used from an isolated 'beforeProject' action,
+            // which is serialized per project. The captured BuildLayout must survive that.
+            def captured = service(BuildLayout)
```
6hundreds, line 120: ❓ Do we have a test for services that not survive CC boundary?

reinsch82: The only service that didn't survive was the settings-scoped BuildLayout; that's the case this test now covers (it used to assert the failure). Every other lookupable service re-resolves from the task's registry, so there's no remaining negative case.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
reinsch82: I read this as with the annotation we would declare this as intended behavior, while I don't think it is. It's an accident in API history. an unfixable for that matter.

we could only fix `onlyIf` if we made it `DELEGATE_FIRST`, which is a behavioral change. we'd need to introduce a new API for that purpose and deprecating the usage of `onlyIf` and thereby allowing the users to address the deprecation as soon as it shows up.
Therefore `onlyIf` will not be fixed, so declaring it `ToBeFixed` is a wrong notion to apply here.

What am I getting wrong?

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/IsolatedProjectsServiceLookupIntegrationTest.groovy
reinsch82: there are none now, that don't survive the boundary.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ServiceLookupGroovyDispatchIntegrationTest.groovy
mlopatkin: > I read this as with the annotation we would declare this as intended behavior, while I don't think it is. It's an accident in API history. an unfixable for that matter.

The `@ToBeFixedForCC` annotation means "this test fails with CC today, but we believe its a bug and it must work". It ensures that the test fails, indeed, but that only to remind to remove the annotation when the underlying issue is fixed - not to encode "this feature breaks CC and must stay that way". 
If we simply don't test - there is:
- no test for vintage
- no track that there is a corner case we wanted to address
- no safety net to indicate the change in the behavior

> we could only fix `onlyIf` if we made it `DELEGATE_FIRST`, which is a behavioral change.

FWIW, for `register { onlyIf { ... }}` there will be no behavior change, as the owner is the enclosing Closure that has the Task as its delegate and `DELEGATE_FIRST` strategy. With CC, we can be clever and restore the lookup change somehow for such Closures only. Other Closures, that look up symbols in the script first can stay broken.

reinsch82:
@bot-gradle test rfn

reinsch82:
@bot-gradle test this

mlopatkin (approved):

PR #38555, opened 2026-07-16
Include component metadata rules in ResolutionParameters

jvandort:
Component metadata rules and the variant derivation strategy are now captured as an immutable snapshot and passed through ResolutionParameters, rather than being pulled out of a service as hidden inputs to resolution.

- Settings-scoped rules are now passed as immutable data. Previously DependencyResolutionManagementInternal.applyRules mutated a project-scoped handler passed into a settings-scoped service, rather than that service exposing its own immutable settings-scoped rules.
- Remove the ComponentMetadataHandlerInternal.createFactory indirection through delegate handlers. The logic for choosing between a project's and the settings' rules is now stated plainly in one place.
- Wire component metadata rules and the variant derivation strategy explicitly through ResolutionParameters instead of embedding them in a service where they acted as hidden inputs to resolution.
- Split DefaultComponentMetadataHandler into a rule DSL builder and a separate processor factory. The DSL element can now be created in the settings scope without dragging in services it doesn't need.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test apt

PR #38556, opened 2026-07-17
Remove core dependency from workers

jvandort:
Workers no longer depend on core -- at least at compile time. 
This does not necessarily mean workers don't load core classes at runtime. We will find out soon enough though. 

To do thi:
- Move WorkerShared*ScopedServices to new project: `worker-shared`, containing classes that the worker shares with the daemon
- Move `MixInLegacyTypesClassLoader` to `worker-shared`, as currently workers continue to support loading legacy types compiled against groovy APIs
- Split `ant-impl` into `isolated-ant-builder`, moving DefulatIsolatedAntBuilder, which is used in workers, away from classes like `AntBuilder` and `AntTask` which only run in the daemon
- Move `RelativePathUtil`, `VerificationException`, and `MarkedVerificationException` out of core, so that the compiler workers and code quality workers no longer depend on core

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

blindpirate:
it failed with a known flakiness.

jvandort:
@bot-gradle test ast

jvandort:
@bot-gradle test ast

PR #7931, opened 2018-12-05
Remove build-receipt.properties from Wrapper jar

marcphilipp:
Prior to this commit, the `gradle-wrapper.jar` was created when
executing the `wrapper` task by copying the `gradle-wrapper.jar`
classpath resource and adding a Wrapper-specific
`build-receipt.properties` to it. The properties file only contained a
`versionNumber` property that was solely used to build the User-Agent
for downloading the distribution.

Now, the User-Agent no longer includes the version number. Thus, the
Wrapper-specific `build-receipt.properties` file is no longer needed.

This change will make it easier to publish the Wrapper jar or, as a
first step, its checksum alongside the distributions.

Resolves #6632.

Issue #7930, opened 2018-12-05
Provide better support for difference in case sensitivity of the file system


lacasseio:
### Expected Behavior

On case-sensitive system, when a file is not found but a different cased file exists, we could suggest using that file instead.

On case-insensitive system, we could warn about different cased file open.

All this is to help notify the user that his build system contains potential cross-platform issues.

### Current Behavior

Gradle delegate all file operation to the operating system. When developing cross-platform builds, a different case can result in explicit or, worse, silent failures. This cause builds to be invalid or lengthy debug time.

### Context

See https://discuss.gradle.org/t/case-sensitivity-in-configurablefilecollection/29506

### Your Environment

 * Build scan URL:

lacasseio:
@wolfs @lptr This should not be close. It's a real issue with caching and up-to-date checks.

PR #38563, opened 2026-07-17
Deprecate restricting KotlinDslScriptsModel to explicit scripts

6hundreds:
Part of #38820

## Context

The original plan of this PR — switching all builds to the `IsolatedProjectsSafeKotlinDslScriptsModelBuilder` and removing the `org.gradle.tooling.model.kotlin.dsl.scripts` property outright — introduced a behavioral breaking change without a deprecation cycle (see review discussion below). This PR is now narrowed to **step 1 of the agreed path**: cleanly deprecate the property and its behavior, changing nothing else.

The builder switch and the removal of the property are deferred to Gradle 10:
1. ~Explicitly specify affected version ranges of IDEs~ (done, see Due diligence below)
2. **This PR**: cleanly deprecate the flag and its usage regardless of IP — delivered in **Gradle 9.9**
3. (optionally in 9.x) Switch model builders not affected by the flag to the IP-safe implementation
4. (in 10) Remove the flag and switch the remaining model builders

## Changes

- `KotlinDslScriptsModel.SCRIPTS_GRADLE_PROPERTY_NAME` is now `@Deprecated`, scheduled for removal in Gradle 10. Builder selection is untouched: the legacy builder still serves non-IP builds and still honors the property.
- Setting the property emits a deprecation warning (`DeprecationLogger`, upgrade guide section `kotlin_dsl_scripts_model_explicit_scripts`) and then behaves exactly as before.
- Under Isolated Projects, setting the property fails the model request as it always has — now with `InvalidUserDataException` instead of `IllegalArgumentException`, and an IP-scoped message.
- Upgrade guide entry added under a new `Upgrading from 9.8.0 and earlier` section (this is the first 9.9-targeted entry).
- Cross-version coverage: the existing `r60`/`r68` explicit-scripts specs keep running against all Gradle versions with their full assertions; a shared `maybeExpectExplicitScriptsDeprecation()` helper (same pattern as `maybeExpectAccessorsDeprecation`) expects the warning when the target version is 9.9+.

## Due diligence: no IDE ever set the property

Checked every known open-source Gradle IDE integration for both the property string and the constant name, at release-tag granularity:

| Repo | Coverage | Result |
|---|---|---|
| JetBrains/kotlin (hosted the Kotlin IDE plugin 2019–2022, the property's entire early life) | Exhaustive: full master history + all 284 `v*` release tags (4.4M git objects) | Only hit: vendored copy of Gradle's own `KotlinDslScriptsModel.java` (the constant declaration) |
| JetBrains/intellij-community | All 29 release-branch heads (`idea/173` → `idea/2026.2.1-rc`) + recent master (1.5M objects) | Zero hits |
| eclipse/buildship | Exhaustive: all tags + all objects | Zero |
| apache/netbeans | Exhaustive: all tags + all objects | Only hit: generated `.sig` API snapshot transcribing Gradle's constant declaration |
| microsoft/vscode-gradle | Exhaustive: all tags + all objects | Zero |
| microsoft/build-server-for-gradle | Exhaustive: all tags + all objects | Zero |
| GitHub-wide code search | Both strings | Only vendored Gradle distributions and gradle/gradle forks |

Only IntelliJ's Kotlin plugin fetches `KotlinDslScriptsModel` at all. Its resolver — from v1.3.70 (the first version to fetch the model, contemporaneous with Gradle 6.0 introducing the property) through current master — passes only the correlation id and classpath mode, never the scripts property. The affected IDE version range is the empty set; the deprecation window covers the residual unknown of closed-source TAPI clients.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


6hundreds:
@bot-gradle test this

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/model/kotlin/dsl/KotlinDslScriptsModel.java
```diff
@@ -49,11 +46,6 @@
  */
 public interface KotlinDslScriptsModel {
 
-    /**
-     * Gradle property name for the set of scripts to be queried for.
-     */
-    String SCRIPTS_GRADLE_PROPERTY_NAME = "org.gradle.tooling.model.kotlin.dsl.scripts";
```
reinsch82, line 56: ❌ I'd say we deprecate it and remove it with 10.
there are a few less well known users of the TAPI. It doesn't really hurt that much to deprecate it.

6hundreds: @reinsch82 To be clear, now the PR just removes the flag from public API. It's even simpler than go through deprecation cycle I believe. Do you think we should got via proper deprecation cycle?

6hundreds:
@bot-gradle test this

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinDslScriptsModelBuilder.kt
```diff
@@ -32,6 +32,15 @@ import org.gradle.tooling.provider.model.ToolingModelBuilder
 import java.io.File
 
 
+/**
+ * Gradle property that used to restrict the set of scripts the model is built for.
+ * No longer part of the public [KotlinDslScriptsModel] API: it is only honored by the legacy builder
+ * and dies with it.
+ */
+internal
+const val SCRIPTS_GRADLE_PROPERTY_NAME = "org.gradle.tooling.model.kotlin.dsl.scripts"
```
jbartok: There should be an entry in the upgrade guide describing this information (and others that are currently only found in KDoc).

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/isolated/AbstractIsolatedProjectsToolingApiIntegrationTest.groovy
```diff
@@ -25,6 +25,15 @@ class AbstractIsolatedProjectsToolingApiIntegrationTest extends AbstractIsolated
 
     static final String CONFIGURE_ON_DEMAND_FOR_TOOLING = "-Dorg.gradle.internal.isolated-projects.configure-on-demand=tooling"
     static final String CACHING_FOR_TOOLING = "-Dorg.gradle.internal.isolated-projects.caching=tooling"
+    static final String LEGACY_KOTLIN_DSL_SCRIPTS_MODEL_BUILDER = "-Dorg.gradle.internal.legacy-kotlin-dsl-scripts-model=true"
+
+    /**
+     * The IP-safe KotlinDslScriptsModel builder is the default even without Isolated Projects, so parity
+     * tests must force the legacy builder when fetching the model they compare the IP-safe one against.
```
jbartok: Also add to upgrade guide.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/crossVersionTest/groovy/org/gradle/kotlin/dsl/tooling/builders/r980/KotlinDslGivenScriptsModelCrossVersionSpec.groovy
```diff
@@ -0,0 +1,102 @@
...
+
+import static org.gradle.kotlin.dsl.tooling.fixtures.KotlinScriptModelParameters.SCRIPTS_GRADLE_PROPERTY_NAME
+import static org.gradle.kotlin.dsl.tooling.fixtures.KotlinScriptModelParameters.setModelParameters
+
+/**
+ * From Gradle 9.8 onwards, the Isolated Projects safe builder is the default for {@link KotlinDslScriptsModel}.
+ * It always builds the model for all the Kotlin DSL scripts of the build and rejects the
+ * {@code org.gradle.tooling.model.kotlin.dsl.scripts} property (removed from the public API) that used to
+ * restrict the requested scripts.
+ * The legacy builder can be temporarily restored with the {@code org.gradle.internal.legacy-kotlin-dsl-scripts-model}
```
jbartok: Also add to upgrade guide.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/model/kotlin/dsl/KotlinDslScriptsModel.java
jbartok: Definitely, I can't come up with any justification or precedent for "just removing something from the public API" without deprecating it first.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
```diff
@@ -69,8 +70,11 @@ class IsolatedProjectsSafeKotlinDslScriptsModelBuilder(
 ) : AbstractKotlinDslScriptsModelBuilder() {
 
     override fun prepareParameter(rootProject: Project): KotlinDslScriptsParameter {
-        require(rootProject.findProperty(KotlinDslScriptsModel.SCRIPTS_GRADLE_PROPERTY_NAME) == null) {
-            "Property ${KotlinDslScriptsModel.SCRIPTS_GRADLE_PROPERTY_NAME} is not supported with Isolated Projects"
+        require(rootProject.findProperty(SCRIPTS_GRADLE_PROPERTY_NAME) == null) {
+            "Property $SCRIPTS_GRADLE_PROPERTY_NAME is no longer supported: " +
+                "the model is always built for all the Kotlin DSL scripts of the build. " +
+                "Set the '${KotlinScriptingModelBuildersRegistrant.LEGACY_SCRIPTS_MODEL_BUILDER.propertyName}' system property to 'true' " +
```
jbartok: Tells the user to set `org.gradle.internal.legacy-kotlin-dsl-scripts-model=true`, but `KotlinScriptingModelBuildersRegistrant` computes `useLegacyBuilder = !isIsolatedProjects && flag`, so with IP enabled the advice does nothing and the user gets the same error again. Either condition the hint on `!isIsolatedProjects`, or word it "…(without Isolated Projects)". Optionally warn when the flag is set alongside IP.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/internal/IsolatedProjectsSafeKotlinDslScriptsModelBuilder.kt
```diff
@@ -69,8 +70,11 @@ class IsolatedProjectsSafeKotlinDslScriptsModelBuilder(
 ) : AbstractKotlinDslScriptsModelBuilder() {
 
     override fun prepareParameter(rootProject: Project): KotlinDslScriptsParameter {
-        require(rootProject.findProperty(KotlinDslScriptsModel.SCRIPTS_GRADLE_PROPERTY_NAME) == null) {
-            "Property ${KotlinDslScriptsModel.SCRIPTS_GRADLE_PROPERTY_NAME} is not supported with Isolated Projects"
+        require(rootProject.findProperty(SCRIPTS_GRADLE_PROPERTY_NAME) == null) {
```
jbartok: `require` throws an `IllegalArgumentException`; for such a user-facing misconfiguration an `InvalidUserDataException` of `GradleException` would read as a build problem rather than a programming error and would make more sense.

jbartok (dismissed):

alllex (changes requested):
After taking a fresh look at this, I don't think we can proceed with the proposed plan, because this PR introduces a behavioral breaking change, which needs a deprecation. 

The fact that we provide an _internal_ flag as a replacement is a mediation for someone who would have already been hit by that change. The replacement flag is a valid strategy for a change in a major, where we do change the behavior (remove support for the original flag).

We must treat the public API/behavior strictly. The fact that we did the research and didn't find any major usages of the flag gives us confidence that its deprecation and the consequent removal wouldn't create a negative impact. The only thing this allows us is to have a very short deprecation window. Fortunately, Gradle 10 is right around the corner, so the path of deprecating this in a late 9.x and removing it in 10 seems feasible.

The PR description says that we verified against "sources of" the popular IDEs. This flag/behavior is cross-version, so changing it can have impact on older versions of the IDEs. Therefore, being explicit about the version ranges of the affected IDEs is mandatory for due diligence.

It frustrates me that we can't switch to the IP builders more easily. But the complete alternative to implementing the support for this flag would be a wasted effort. Having said that, we can look into switching all model builders not affected by this flag, which might already be useful.

The path forward, as I see it:

0. Explicitly specify affected version ranges of IDEs
1. Cleanly deprecate this flag and its usage regardless of IP
2. (optionally in 9.x) Switch all model builders that don't depend on this flag to the IP
3. (in 10) Remove the flag and switch the remaining model builders

6hundreds:
@alllex @jbartok If we find an evidence that the flag was never used, or was used only by really ancient version of IDEA, will it untie our hands?
I have no history proofs yet, only gut feel that the flag doesn't deserve deprecation cycle.

alllex:
@6hundreds, imo we should follow the public API changes contract and remove the flag support only in a major.

https://docs.gradle.org/current/userguide/feature_lifecycle.html#sec:public:~:text=A%20public%20feature%20will%20never%20be%20removed%20or%20intentionally%20changed%20without%20undergoing%20deprecation.

As I said previously, in case no recent versions of IDEA/Studio are affected, this makes us confident that a short deprecation cycle is enough. But there still has to be a deprecation cycle.

6hundreds:
@bot-gradle test this

6hundreds:
@alllex @jbartok PTAL at the deprecation-only version of the PR

lkasso (approved):

alllex:
@6hundreds, can you make sure the PR title and description accurately reflect the change?

6hundreds:
@alllex done!

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/model/kotlin/dsl/KotlinDslScriptsModel.java
```diff
@@ -28,10 +28,9 @@
...
  * See {@link KotlinDslModelsParameters#PREPARATION_TASK_NAME}
  * <p>
- * The set of scripts can be provided as a Gradle property named <code>org.gradle.kotlin.dsl.provider.scripts</code>,
- * as a list of absolute paths separated by <code>|</code>.
- * If none are provided, then the model is built for all the Kotlin DSL scripts known to belong to this build.
- * See {@link KotlinDslScriptsModel#SCRIPTS_GRADLE_PROPERTY_NAME}.
+ * By default, the model is built for all the Kotlin DSL scripts known to belong to this build.
+ * Restricting the model to an explicit set of scripts via {@link #SCRIPTS_GRADLE_PROPERTY_NAME} is deprecated
+ * and will be removed in Gradle 10.
  * <p>
```
alllex, line 34: 💅 The deprecation note in the middle of the javadoc is hard to find and is probably not that useful. I'd leave the javadoc as is.

platforms/core-configuration/kotlin-dsl-tooling-builders/src/crossVersionTest/groovy/org/gradle/kotlin/dsl/tooling/builders/r980/KotlinDslGivenScriptsModelCrossVersionSpec.groovy
alllex, line 1: ❌  We are losing coverage for some of the functionality by not duplicating all tests here, after skipping them in the older test classes. Overall, we expect the behavior to stay functionally the same until we actually remove the property. So there is no reason to even try to duplicate coverage in the first place.

In case of deprecations, we normally add a conditional verification of that. E.g. see `org.gradle.kotlin.dsl.tooling.builders.r60.KotlinDslScriptsModelCrossVersionSpec#maybeExpectAccessorsDeprecation`.

platforms/ide/tooling-api/src/main/java/org/gradle/tooling/model/kotlin/dsl/KotlinDslScriptsModel.java
```diff
@@ -51,8 +50,12 @@ public interface KotlinDslScriptsModel {
 
     /**
      * Gradle property name for the set of scripts to be queried for.
+     *
```
alllex, line 53: 💅  since we are reformatting the file, can we reformat the whole file? E.g. the javadoc of the other method

alllex (changes requested):

alllex:
@bot-gradle test ACT

alllex (approved):

6hundreds:
@bot-gradle test act

blindpirate:
@6hundreds This PR is broken against current `master` and will be ejected from the merge queue again if it is re-queued as-is. It has already been ejected twice, and the second time it took #38925 down with it as the stacked entry below it.

### Evidence

Two independent merge-queue runs, both containing this PR's merge commit, failed on the same 5 tests in `QuickFeedbackCrossVersion_5_bucket7`:

| Merge-queue run | Queue branch | Revision | Failing bucket |
|---|---|---|---|
| [#7930](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_GitHubMergeQueueCheckPass/116606983) | `pr-38563-c5853843184` | `8bd4a189` (this PR's merge commit) | [116606973](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_QuickFeedbackCrossVersion_5_bucket7/116606973) |
| [#7931](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_GitHubMergeQueueCheckPass/116607563) | `pr-38925-8bd4a189d14` (stacked on top of this PR) | `b19921d5` | [116607553](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_QuickFeedbackCrossVersion_5_bucket7/116607553) |

Both runs were cancelled by `bot-gradle` with `This build won't succeed because .../Gradle_Master_Check_QuickFeedbackCrossVersion_5_Trigger/… fails`, and both list the identical 5 failures:

```
r60.KotlinDslScriptsModelCrossVersionSpec.multi-scripts model is dehydrated over the wire
r60.KotlinDslGivenScriptsModelCrossVersionSpec.can fetch model for a given set of scripts
r60.KotlinDslGivenScriptsModelCrossVersionSpec.can fetch model for a given set of scripts of a build in lenient mode
r68.KotlinDslGivenScriptsModelCrossVersionSpec.can fetch model for a given set of init scripts
r68.KotlinDslGivenScriptsModelCrossVersionSpec.can fetch model for a given set of init scripts of a build in lenient mode
```

all in the `TAPI current -> Gradle 9.8.0-20260824131806+0000` variant, all with:

```
java.lang.AssertionError: Expected the following deprecation warnings:
 - Requesting the KotlinDslScriptsModel for an explicit set of scripts using the
   'org.gradle.tooling.model.kotlin.dsl.scripts' Gradle property. This behavior has been
   deprecated. This is scheduled to be removed in Gradle 10. ...
```

i.e. the test declared the deprecation as expected, but the Gradle under test never emitted it. Two runs, two different base commits, same five tests, same assertion — this is deterministic, not flaky.

### Root cause

`master` was bumped to `9.9.0` at 10:34 UTC on 2026-08-24 by #38940, and #38941 moved `released-versions.json` `latestReleaseSnapshot` to `9.8.0-20260824131806+0000`. The cross-version matrix therefore now runs `TAPI current -> Gradle 9.8.0-…`, and that distribution predates your deprecation.

The guard added in this PR does not exclude it — `AbstractKotlinScriptModelCrossVersionTest.groovy:244`:

```groovy
if (targetVersion >= GradleVersion.version("9.8")) {
```

`DefaultGradleVersion.compareTo` compares the dot-separated version part first and returns `1` as soon as one side has *more* segments:

```java
if (majorVersionParts.length > otherMajorVersionParts.length) {
    return 1;
}
```

`"9.8.0-20260824131806+0000"` has three segments (`9.8.0`) against `"9.8"`'s two, so the comparison short-circuits to "greater" before the snapshot suffix is ever considered. The guard fires for every 9.8.x nightly, none of which carry the deprecation.

This is also why the full cross-version run you triggered with `@bot-gradle test act` [passed at 10:27 UTC](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_Stage_ReadyforRelease_CrossVersionTests_Trigger/116593860): it started at 09:59, ~35 minutes *before* the 9.9.0 bump landed, so the target was still a 9.7.x snapshot and the guard correctly evaluated to `false`. The PR has not changed since (`cd416cf`) — the ground moved under it.

### Fix

```groovy
if (targetVersion.baseVersion >= GradleVersion.version("9.9")) {
```

The deprecation first ships in 9.9.0, so the bound has to be `9.9`. `.baseVersion` is the idiom used throughout `crossVersionTest` (e.g. `ToolingApiEclipseModelClasspathContainerCrossVersionSpec:193`, `AbstractProgressCrossVersionSpec:41`) and keeps the comparison correct for `9.9-rc-N`-shaped targets too.

---

<sub>For the record, the *first* ejection at 13:02 UTC was unrelated to this PR: `:architecture-test:checkBinaryCompatibility` was red across all of `master` in the window between #38940 (which bumped the version and cleaned `accepted-public-api-changes.json`) and #38941 (which advanced the japicmp baseline). Only the second ejection is on you.</sub>

PR #38572, opened 2026-07-19
Keep copy button visible while scrolling long stack traces

### LINKED ISSUE #38531: Preserve "copy"-button in top right corner while scrolling down long stacktraces in HTML test reports
### Expected Behavior

My proposal is to keep the copy button visible on the top right corner of the stack trace area while scrolling long stack traces

Currently the copy-button has fixed location and goes out of the visible are of the screen while scrolling down long stack traces.

### Context

This feature request is related to **the** copy button introduced in #30449.

The problem is that in the JVM world (especially for enterprise JVM apps) stack traces are pretty long, and you usually scroll to the bottom of them to see the root `Caused by` first. After that, I quite often want to copy the stack trace to analyze it in IntelliJ, but the copy button was left at the top and moved out of the visible area of the screen, so I need to scroll back up ⬆️.

My proposal is to keep the copy button visible on the top right corner of the stack trace area while scrolling long stack traces

jkrannich:
Fixes #38531 

### Context

In the HTML test report, the "Copy" button was always statically positioned in the top right. Due to JVM stack traces being very long in the case of a failure occuring, users often had to scroll way further down. This caused the button to go out of view since it was in a fixed position, so if a user wanted to copy the stack trace they would then have to scroll all the way back up.

My approach to fix this was to make the button sticky so it will always stay within the user's viewport even when scrolling all the way down.

### Changes

First approach was to just change the styling of the button to `position: sticky`, however since it was wrapped in a span with `position: relative` that wasn't possible. Therefore I wrapped the button in a new div with zero-height and `position: sticky` and the button keeps the original layout. I also had to move it so it would render before the `<pre>` because otherwise the button would sit below the entire stack trace.

Files I changed:

- `HtmlWriterTools` - wraps the button in `div.clipboard-copy-btn-container`
- `CodePanelRenderer` / `PerRootTabRenderer` - emit the button before the `<pre>`
- `generic/style.css` - adds the `.clipboard-copy-btn-container` rule

### Testing

I did not add any unit/integration tests since this is a pure layout/css change. However I did test it by first trying it out via DevTools in my Chrome Browser. 

I then implemented the code changes and built it locally via `./gradlew install` and ran tests in my test repository with the generated binary. Below you can find a short screen recording of the result:

https://github.com/user-attachments/assets/31f65c6e-bcae-42f9-bbf6-376065ce13bc

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/report-rendering/src/main/java/org/gradle/reporting/CodePanelRenderer.java
+++ b/platforms/core-runtime/report-rendering/src/main/java/org/gradle/reporting/CodePanelRenderer.java
@@ -37,9 +37,9 @@ public class CodePanelRenderer extends ReportRenderer<CodePanelRenderer.Data, Si
     @Override
     public void render(Data data, SimpleHtmlWriter htmlWriter) throws IOException {
         // Wrap in a <span>, to work around CSS problem in IE
-        htmlWriter.startElement("span").attribute("class", "code")
-            .startElement("pre").attribute("id", data.codePanelId).characters(data.text).endElement();
+        htmlWriter.startElement("span").attribute("class", "code");
         addClipboardCopyButton(htmlWriter, data.codePanelId);
+        htmlWriter.startElement("pre").attribute("id", data.codePanelId).characters(data.text).endElement();
         htmlWriter.endElement();
     }
 }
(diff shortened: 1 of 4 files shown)


ghale:

big-guy:
Thanks, I've queued this for merge into 9.8

PR #38574, opened 2026-07-20
Match corner radius of media to codeblocks

alllex:
Before this change, they had no radius, with default rectangular corners.

Now both images and gifs have the same 4px corner radius as codeblocks:

Image:
image

Gif:
image


alllex:
@bot-gradle test bd

lkasso (approved):

PR #38579, opened 2026-07-20
Speed up package cycle check from ~6min to ~2s by replacing per-project PackageCycleTest with a single whole-codebase check

blindpirate:
### Reason for the change

The package cycle check ran once in every JVM library project's `archTest` task (~240 invocations per build), each paying the full ArchUnit class import cost. This accumulated ~6 minutes of test time per build and contributed to slow SanityCheck runs.

Fixes https://github.com/gradle/gradle-private/issues/4682
Relates to https://github.com/gradle/gradle-private/issues/4333

### Change description

The check now runs as a single test in `:architecture-test`, where the ArchUnit class import is already shared with the other architecture tests. On top of that shared import, the check itself takes about a second.

- Added `PackageCycleTest` to `testing/architecture-test`. It evaluates the cycle rule **once per module**, assigning slices only to that module's own classes (identified by class file location). This preserves the exact semantics of the previous per-project setup: only dependencies between classes of the same module can form a reported cycle, and exclude patterns only apply to the module that declared them. Package cycles spanning multiple modules are not reported, matching the previous behavior.
- The per-project `packageCycles { excludePatterns }` declarations moved into the test as a per-module map, faithfully preserving each module's original patterns.
- Removed the per-project machinery: the `gradlebuild.arch-test` plugin, `PackageCyclesExtension`, the `testing/internal-architecture-testing` project, and all 58 `packageCycles {}` blocks.
- Raised `:architecture-test:test` heap from 1g to 2g: the per-module slice evaluations need headroom on top of the imported class graph (1g OOMed intermittently).

Note: merging all classes into one graph would in principle allow detecting cross-module package cycles too, but there is a substantial pre-existing backlog (ArchUnit caps reporting at 100 cycles, dominated by hubs like `org.gradle.util.internal`, `org.gradle.internal.exceptions` and `org.gradle.api`). That is left as a possible follow-up; this PR is a performance-only change.

### Verification

- SanityCheck passed on this branch (initial version: https://builds.gradle.org/build/115354552).
- Verified failure detection by planting an intentional cycle between two packages in `:files`: the test fails and reports the violating module and the exact cycle with the members causing each edge.
- Full `:architecture-test:test` suite ran three consecutive times without failures; `:architecture-test:codeQuality` and `checkSubprojectsInfo` pass locally.


blindpirate:
@bot-gradle test this

tresat (approved):

PR #38584, opened 2026-07-21
Add lazy evaluation overloads for `onlyIf` task skip reasons

### LINKED ISSUE #38488: [Performance optimization] allow onlyIf reasons to be lambdas
### Expected Behavior

allow the reason String in an `onlyIf` call to be a lambda such as `Supplier<String>` or `Provider<String>` in case we want to do something expensive to generate the error message. Then, only call this reason if the condition fails

if the String is constructed using an expensive ValueSource, this will happen during configuration phase when the task is instantiated, and it will _always_ run

### Context

Our specific example is an onlyIf that checks if you are on the default branch, and we want the message to display what we detect your current branch to be. that logic uses env vars and git operations. ENV are preferred as they are cheaper, but they are only reliably set in CI, so git operations are needed on local runs

tresat:
Introduces new `onlyIf` and `setOnlyIf` overloads on the `Task` interface that accept a `Provider` for the skip reason.

This defers the computation of the reason until the task is actually skipped, avoiding unnecessary work during configuration. It particularly benefits expensive reason calculations and improves configuration cache behavior when using `ValueSource`-backed providers.

Fixes #38488

### DIFF
--- a/platforms/documentation/docs/src/docs/release/notes.md
+++ b/platforms/documentation/docs/src/docs/release/notes.md
@@ -82,10 +82,12 @@ Gradle provides [rich APIs](userguide/getting_started_dev.html) for build engine
 #### Lazy `onlyIf` reasons
 [`Task.onlyIf`](dsl/org.gradle.api.Task.html#onlyIf) and [`Task.setOnlyIf`](dsl/org.gradle.api.Task.html#setOnlyIf) now accept a `Provider<String>` that reports the reason for the task being skipped.
 
-This provider is only computed when the task is actually skipped.  
-This will avoid expensive I/O such as running an external process or reading environment state, unless the reason is needed.
+This provider is only computed when the task is actually skipped.
+This avoids expensive I/O such as running an external process or reading environment state, unless the reason is needed.
 
-Using a [`ValueSource`](userguide/configuration_cache_requirements.html#config_cache:requirements:value_source) provider will keep the computation lazy even when the task is serialized in the [Configuration Cache](userguide/configuration_cache.html).
+Using a [`ValueSource`](userguide/configuration_cache_requirements.html#config_cache:requirements:value_source) provider keeps the computation lazy even when the task is serialized in the [Configuration Cache](userguide/configuration_cache.html).
+
+See the [Using a predicate](userguide/controlling_task_execution.html#sec:using_a_predicate) section in the Gradle User Manual for more details.
 
 ### Platform and toolchain management
 Gradle provides comprehensive support for [Native development](userguide/building_cpp_projects.html) and [JVM languages](userguide/building_java_projects.html), featuring automated [Toolchains](userguide/toolchains.html) for seamless JDK management.
--- a/platforms/documentation/docs/src/docs/release/notes.md
+++ b/platforms/documentation/docs/src/docs/release/notes.md
@@ -79,6 +79,14 @@ Gradle provides an intuitive [command-line interface](userguide/command_line_int
 ### Build authoring improvements
 Gradle provides [rich APIs](userguide/getting_started_dev.html) for build engineers and plugin authors, enabling the creation of custom, reusable build logic and better maintainability.
 
+#### Lazy `onlyIf` reasons
+[`Task.onlyIf`](dsl/org.gradle.api.Task.html#onlyIf) and [`Task.setOnlyIf`](dsl/org.gradle.api.Task.html#setOnlyIf) now accept a `Provider<String>` that reports the reason for the task being skipped.
+
+This provider is only computed when the task is actually skipped.  
+This will avoid expensive I/O such as running an external process or reading environment state, unless the reason is needed.
+
+Using a [`ValueSource`](userguide/configuration_cache_requirements.html#config_cache:requirements:value_source) provider will keep the computation lazy even when the task is serialized in the [Configuration Cache](userguide/configuration_cache.html).
+
 ### Platform and toolchain management
 Gradle provides comprehensive support for [Native development](userguide/building_cpp_projects.html) and [JVM languages](userguide/building_java_projects.html), featuring automated [Toolchains](userguide/toolchains.html) for seamless JDK management.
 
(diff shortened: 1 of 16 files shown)


tresat:
@wakingrufus - How does this look?

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -79,6 +79,14 @@ Gradle provides an intuitive [command-line interface](userguide/command_line_int
 ### Build authoring improvements
 Gradle provides [rich APIs](userguide/getting_started_dev.html) for build engineers and plugin authors, enabling the creation of custom, reusable build logic and better maintainability.
 
+#### Lazy `onlyIf` reasons
+[`Task.onlyIf`](dsl/org.gradle.api.Task.html#onlyIf) and [`Task.setOnlyIf`](dsl/org.gradle.api.Task.html#setOnlyIf) now accept a `Provider<String>` that reports the reason for the task being skipped.
+
+This provider is only computed when the task is actually skipped.  
+This will avoid expensive I/O such as running an external process or reading environment state, unless the reason is needed.
+
+Using a [`ValueSource`](userguide/configuration_cache_requirements.html#config_cache:requirements:value_source) provider will keep the computation lazy even when the task is serialized in the [Configuration Cache](userguide/configuration_cache.html).
```
hfhbd: What about the case the ValueSource is mapped, like `myValueSource.map { }`?

It is unrelated to this PR, but the CC user guide does not mention it (ValueSources at all).

tresat: This is also handled lazily.  I added a test to charactorize the behavior [here](https://github.com/gradle/gradle/pull/38584/changes/f4320d280d64522ff3088b86e8640a8669b71da7).

tresat: @gradle/bt-cortex, @lkasso - Perhaps something missing from the docs here.

wakingrufus:
> @wakingrufus - How does this look?

This looks great! Thanks!

bamboo:
I wonder why @gradle/bt-cortex was not automatically added as reviewer...

platforms/documentation/docs/src/docs/release/notes.md
lkasso: Added to the docs in this PR.

lkasso (approved):
Docs LGTM.

mlopatkin:
> I wonder why @gradle/bt-cortex was not automatically added as reviewer...

Nobody owns `core-api` or `core`

big-guy (changes requested):
The use case given in the issue is:
> Our specific example is an onlyIf that checks if you are on the default branch, and we want the message to display what we detect your current branch to be. that logic uses env vars and git operations. ENV are preferred as they are cheaper, but they are only reliably set in CI, so git operations are needed on local runs

That means they have something like this:
```
currentBranch=expensiveCall()
myTask.onlyIf("$currentBranch == default branch") {
    currentBranch == "main"
}
```
The way we use the reason today would result in this kind of message in DV:
> 'not-main == default branch' not satisfied

The crappy alternative is:
```
myTask.onlyIf("on default branch") {
    currentBranch=expensiveCall()
    currentBranch == "main"
}
```
This hides the result of `expensiveCall()`, but it looks like this in DV:
> 'on default branch' not satisified

I'm not sure about this for a few reasons:

1. This is yet another `onlyIf(...)` API with two unusual parameters (Provider + Spec).
2. The Spec and the Provider really belong together as you wouldn't want inputs to the Provider to be different from the Spec or vice versa. You also don't want to duplicate the expensive calculation between them. We're compensating for this in the PR by caching the result for the self describing spec, but we still have to calculate it all again for the isSatisfiedBy call. 
3. `ValueSource` feels like overkill for just a `Provider<String>`.
4. Maybe this is something we'd want generically across all kinds of Specs. 

A few ideas:
1. Maybe cramming diagnostics into `Provider<String>` is the wrong approach and instead `onlyIf` reason should be written in a declarative way and diagnostics should be captured elsewhere. 
2. We should be able to go from a `Provider` to a `Spec`. This doesn't address the reason directly, but it lets users create a ValueSource -> Provider -> Spec chain. We could capture diagnostics from the ValueSource. 
3. Maybe `Spec` is wrong and we should have a `Spec.isSatisfiedBy(ReasonConsumer<T>)` that lets you report a reason to the consumer. I'm not sure if there's a way to do this that would be compatible without a new `onlyIf` method.  Maybe we could have a `onlyIf(BiSpec<T, Consumer<String>>)` that still returns boolean, but when there are two parameters to the closure, we select a different method. 


(2) looks something like:
```
myTask.onlyIf("on default branch", providers.environmentVariable("GIT_BRANCH").asSpec {
    it == "main"
})
```
Since `environmentVariable` is a ValueSource Gradle creates, we could introspect this deeper and record its value.

Maybe a user defined type would look something like:
```
branchInfo = providers.of(BranchInfo) {
    gitBranch = providers.environmentVariable("GIT_BRANCH")
    // other things
}
myTask.onlyIf("on default branch", branchInfo.asSpec {
    gitBranch.get() == "main"
})
```

`Provider.map` is close to this, except you get a `Provider<Boolean>`.  We could do something like:
```
myTask.onlyIf("on default branch", providers.environmentVariable("GIT_BRANCH").map {
    it == "main"
}.asSpec())
```
...to avoid introducing a `asSpec(Transformer)` API.

But none of this is quite right because `onlyIf` needs a `Spec<Task>`. 

(3) looks something like:
```
interface ReasonConsumer<T> {
    void report(String reason)
    T getElement()
}
```
and you'd have to write things like:
```
gitBranch = providers.environmentVariable("GIT_BRANCH")
myTask.onlyIf("on default branch", { ReasonConsumer consumer ->
    String actualBranch = gitBranch.get()
    consumer.report("branch is " + actualBranch)
    actualBranch == "main"
})
```

This isn't great because `gitBranch` is implicitly captured.

PR #35335, opened 2025-10-16
Fix FlakyTestQuarantine

blindpirate:
This PR resolves two issues:

1. FlakyTestQuarantine for smoke tests [fails with error](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_SmokeTest_GradleBuildSmokeTests_FlakyTestQuarantine/104308179?hideTestsFromDependencies=true): `No tests found for given includes`.

2. Remove FlakyTestQuarantine for AllCrossVersionTests because of [the timeout](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_FlakyQuarantine_Check_AllVersionsCrossVersion_11/104308161?buildTab=overview&hideTestsFromDependencies=true) and ["too many artifacts" error](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_FlakyQuarantine_Check_AllVersionsCrossVersion_10/104308173?buildTab=overview&hideTestsFromDependencies=true). We need a better design (probably bucket splits) for AllCrossVersionTest_FlakyTestQuarantine.

PR #38592, opened 2026-07-22
Re-enable cross-version flaky quarantine and un-flag stable kotlin-dsl cross-version tests

blindpirate:
## Why

gradle/gradle#35335 removed **all** cross-version tests from `FlakyTestQuarantine` with the over-broad filter `!it.testType.crossVersionTests`. That is `true` for QUICK, PLATFORM, and QuickFeedbackCrossVersion coverage as well, so those were dropped from quarantine as collateral damage — not just the `AllVersionsCrossVersion` coverage that motivated the removal.

Re-enabling the quarantine for `AllVersionsCrossVersion` surfaced two further findings:
- One "flaky" failure was actually **deterministic** — `KotlinBuildScriptModelCrossVersionSpec."sourcePath includes Gradle sources"` fails in the "old TAPI → current Gradle" direction because the snapshot build under development has no published `gradle-<version>-src.zip` to resolve.
- The `kotlin-dsl-tooling-builders` cross-version specs are otherwise **stable** — across full quarantine runs the only real failure was the one above; the rest passed against every version.

## What

1. **Restore the flaky quarantine for all cross-version coverage** — relax the filter back to `it.os == os`, so QUICK/PLATFORM/QuickFeedbackCrossVersion/AllVersionsCrossVersion are quarantined again. The quarantine mechanism stays enabled for every cross-version test type.

2. **Fix the deterministic sources assertion** — guard `"sourcePath includes Gradle sources"` with `assumeFalse(targetDist.version.snapshot)`, so it is skipped only when the model provider is an unreleased snapshot (whose sources can't be resolved) and still runs/asserts for every released provider.

3. **Un-flag the stable `kotlin-dsl-tooling-builders` cross-version specs** — remove `@Flaky` from the 11 classes so they run in the normal cross-version builds again. They haven't run for a long time so now we can check if they're still flaky - if they're, we can do further actions - they're anyway in ReadyForRelease stage and won't blocker user feedback.

## Context

Addresses gradle-private#4895.

### DIFF
--- a/.teamcity/src/main/kotlin/configurations/FlakyTestQuarantine.kt
+++ b/.teamcity/src/main/kotlin/configurations/FlakyTestQuarantine.kt
@@ -53,7 +53,7 @@ class FlakyTestQuarantineProject(
         model.stages
             .filter { it.stageName <= StageName.READY_FOR_RELEASE }
             .flatMap { it.functionalTests }
-            .filter { it.os == os && !it.testType.crossVersionTests }
+            .filter { it.os == os }
             .forEach {
                 buildType(FlakyTestQuarantine(model, stage, it))
             }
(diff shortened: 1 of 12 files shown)


blindpirate:
@bot-gradle squash

PR #38569, opened 2026-07-18
Consolidate daemon bootstrap communication

jvandort:
Move all logic for communicating between client and daemon over stdin/out into `DaemonStartupCommunication` in `:daemon-messaging`. 

Goal is to allow this logic to be easily reused so that embedded daemons can be migrated to communicate the same way process isolated daemons are.

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #38594, opened 2026-07-22
Execute embedded daemon in separate classloader

jvandort:
Goal: Decouple the client from compiling against the daemon.

There are three major technical places we do this, all related to the embedded daemon
* CLI --no-daemon when the target JVM args matches the current process's JVM args, we start the daemon in the client process
* CLI --foreground, we start a long-running daemon in the client process.
* TAPI embedded mode, driven by TestKit `withDebug`, we start the daemon in the client.

In all three cases, the current implementation directly compiles against `BuildProcessState` or `DaemonProcessState`, which are the top-level entry points implementing what we know as the "daemon". This implementation of an embedded daemon requires the client to compile against the entire daemon classpath.

This PR implements the first step of migrating away from this pattern by reimplementing the CLI --no-daemon path to instead load the daemon classpath in a separate classloader. The TAPI and --foreground paths will be migrated in future changes. 

Instead of directly instantiating a new `BuildProcessState`, we instead start up "real" daemon in a separate classloader, communicating with in initially over a "phony" stdin/stdout, like we do a forked daemon. The embedded daemon starts a real server, opening a real socket, and we communicate with the classloader-isolated daemon like we do a forked daemon.

The main classes of interest are `EmbeddedDaemonStarter`, which parallels the existing `DefaultDaemonStarter`, and the `EmbeddedDaemonMain`, which parallels the existing `DaemonMain`.

This work is required to fully decouple the client from the daemon, and after completion will allow us to compile many more daemon classes to Java 17. 

Prerequisites:
* https://github.com/gradle/gradle/pull/38569
* https://github.com/gradle/gradle/pull/38610
* https://github.com/gradle/gradle/pull/38608

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test rfr

jvandort:
@bot-gradle help

jvandort:
@bot-gradle test rfr

jvandort:
@bot-gradle test rfr

jvandort:
@bot-gradle test rfr

jvandort:
@bot-gradle test rfr

platforms/core-runtime/daemon-server/src/main/java/org/gradle/launcher/daemon/bootstrap/EmbeddedDaemonMain.java
```diff
@@ -0,0 +1,103 @@
...
+            // Native services are not initialized here. They are a per-process singleton that the
+            // hosting client has already initialized, and which the daemon sees through the shared
+            // parent classloader.
+
+            ServiceRegistry loggingRegistry = LoggingServiceRegistry.newEmbeddableLogging();
+
+            // There is not necessarily a process exit to rely on for cleanup after a forced stop.
+            // Embedded TAPI clients may execute multiple embedded daemons sequentially in the same
+            // process.
+            boolean closeServicesOnForcedStop = true;
```
asodja, line 63: ❓ 🤖  A forced stop means the build did not respond to cancellation and may still be running. Closing the daemon services here then waits for that same build worker to finish, with an effectively unbounded timeout.

For the CLI this becomes a one-minute delay before `System.exit` terminates the process. 

For an embedded Tooling API client, it could hang indefinitely. Could we keep the existing skip-on-forced-stop behavior, or use bounded cleanup that does not wait forever for the abandoned build?

platforms/core-runtime/daemon-server/src/main/java/org/gradle/launcher/daemon/bootstrap/EmbeddedDaemonMain.java
```diff
@@ -0,0 +1,103 @@
...
+     * no log file, since all of its output leaves the daemon over its connections. It captures
+     * the process' system sources while it runs, restoring them when the daemon stops.
+     */
+    private static class EmbeddedDaemonServerEnvironment implements DaemonServerEnvironment {
+
+        @Override
+        public void beforeStart(LoggingManagerInternal loggingManager, File daemonLog, ServiceRegistry daemonServices) {
+            // Capture System.out and System.err so they are forwarded to the client as logging events.
+            // The previous streams are restored when the logging manager stops.
+            loggingManager.captureSystemSources();
```
asodja, line 91: ❓ 🤖 
The embedded daemon temporarily takes ownership of the JVM's shared logging state, but its logging manager has no console or file attached. Build output is visible while `LogToClient` is active, but client
  messages emitted before or after that relay window are dropped.

  For example, the timeout warning in `SingleUseDaemonClient` is logged while the embedded daemon still owns the shared logger, so the user cannot see it. `captureSystemSources()` also redirects the client's
  `System.out` and `System.err` for the daemon's lifetime.

  Could we preserve the client's logging destination or otherwise provide an output sink outside the `LogToClient` window?

platforms/core-runtime/client-services/src/main/java/org/gradle/launcher/daemon/client/SingleUseDaemonClient.java
```diff
@@ -46,20 +47,31 @@ public SingleUseDaemonClient(
...
+            // connection must be closed first, so the daemon is not kept alive by this client.
+            daemonConnection.stop();
+            awaitDaemonTermination(daemonConnection);
+        }
+    }
 
-        return executeBuild(build, daemonConnection, buildRequestContext.getCancellationToken(), buildRequestContext.getEventConsumer());
+    private static void awaitDaemonTermination(DaemonClientConnection daemonConnection) {
+        DaemonHandle daemonHandle = daemonConnection.getDaemonHandle();
+        if (daemonHandle != null && !daemonHandle.awaitTermination(1, TimeUnit.MINUTES)) {
```
asodja, line 72: ❓ 🤖   The client assumes every started single-use daemon has reached `RequestStopIfSingleUsedDaemon` and will therefore stop itself. However, the build command can be rejected or fail before reaching that action. Closing the connection does not stop the daemon, so this wait then lasts the full minute.

Additionally: The embedded branch also skips the two-minute safety cap, leaving such an idle daemon with the configured idle timeout - three hours by default. 

Could we explicitly stop the handle when the build was not accepted, rather than waiting for a stop that was never requested?

platforms/core-runtime/daemon-server/src/main/java/org/gradle/launcher/daemon/bootstrap/DaemonServerBootstrap.java
```diff
@@ -0,0 +1,121 @@
...
+            daemon = daemonServices.get(Daemon.class);
+            daemon.start();
+
+            Long pid = daemonServices.get(DaemonContext.class).getPid();
+            DaemonStartupCommunication.writeDaemonStartupInfo(
+                startupOut,
+                new DaemonStartupInfo(
+                    daemon.getUid(),
+                    daemon.getAddress(),
+                    new DaemonDiagnostics(daemonLog, pid)
```
asodja, line 96: ❓ 🤖  This sends the client a `DaemonDiagnostics` log path even though the embedded environment deliberately never creates that file. The path is used both when `doStartDaemon()` times out connecting and when the daemon disappears during a build.

  If the file is absent, `DaemonDiagnostics` formats `NoSuchFileException.getCause()`, which is normally null, so the user gets `because of: null`. Because the filename is derived from the shared client PID, PID reuse can also cause an unrelated stale daemon log to be tailed.

Could embedded daemons explicitly report that no daemon log exists, rather than advertising this synthetic path?

platforms/core-runtime/client-services/src/main/java/org/gradle/launcher/daemon/client/DefaultDaemonConnector.java
```diff
@@ -216,12 +219,13 @@ public DaemonClientConnection startDaemon(ExplainingSpec<DaemonContext> constrai
     private DaemonClientConnection doStartDaemon(ExplainingSpec<DaemonContext> constraint, boolean singleRun) {
         ProgressLogger progressLogger = progressLoggerFactory.newOperation(DefaultDaemonConnector.class)
             .start("Starting Gradle Daemon", "Starting Daemon");
-        final DaemonStartupInfo startupInfo = daemonStarter.startDaemon(singleRun);
+        final DaemonHandle daemonHandle = daemonStarter.startDaemon(singleRun);
```
asodja, line 222: ❓ 🤖  Once `startDaemon()` returns, this method has no cleanup path if registry discovery or connection fails. For an embedded daemon, the handle is lost while its non-daemon thread continues until idle expiration.

The CLI currently masks this by calling `System.exit`, but a reusable in-process client would leak the daemon and its resources.

Could `DaemonHandle` expose a stop operation and be stopped/awaited from this failure path?

asodja (changes requested):
Added few comments mostly from things 🤖 found. So please double check

Also the new embedded starter does not appear to have direct coverage for its important failure paths:

  - connector failure after startup;
  - command failure before `RequestStopIfSingleUsedDaemon`;
  - forced cancellation with an unresponsive worker;
  - restoration of client logging after shutdown or timeout.

The compatible cases in `SingleUseDaemonIntegrationTest` exercise the successful embedded path, but not these cleanup and diagnostic paths. Could we add/do we need to add targeted coverage for them?

PR #38608, opened 2026-07-23
Warn instead of failing when the daemon cannot mutate its environment

jvandort:
A daemon running without the JPMS --add-opens flags, such as a foreground daemon started by the launch scripts, fails the build with 'Unable to get mutable environment variable map.' when applying the client's environment variables, instead of printing the intended warning. Wrap the native calls in try/catch and warn on failure.

Also re-enable three foreground daemon tests that have been disabled since 2018 because of this. They additionally needed the foreground daemon and the client to request the same daemon JVM args, since the test executer's default JVM args differ between the two, preventing the client from reusing the foreground daemon.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

PR #38610, opened 2026-07-23
Restore logger source state when closing a nested logging manager

jvandort:
When a nested logging manager was closed, the SLF4J and java.util.logging sources did not restore the state installed by the outer logging manager. Log events kept being forwarded to the nested manager's closed destination instead of the top-level owning logging manager. Track that state in the source's snapshot and reinstate it on restore.

For SLF4J, fold Slf4jLoggingConfigurer and LoggingSystemAdapter into a single Slf4jLoggingSystem whose snapshot captures the binding's current listener and level. For java.util.logging, snapshot and reinstate the root logger's handlers instead of resetting the LogManager, closing only the handlers installed during the capturing scope.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

PR #38612, opened 2026-07-23
Introduce `plugin-development-ecosystem` and `xdclGradlePlugin`

h0tk3y:


build-logic/packaging/src/main/kotlin/gradlebuild.distributions.gradle.kts
```diff
@@ -203,7 +219,13 @@ dependencies {
 }
 val gradleApiKotlinExtensions = tasks.register<GenerateKotlinExtensionsForGradleApi>("gradleApiKotlinExtensions") {
     sharedRuntimeClasspath.from(kotlinDslSharedRuntimeClasspath)
-    classpath.from(runtimeClasspath)
+    classpath.from(
+        runtimeClasspath.filter { jar ->
```
jvandort, line 223: This looks like it performs dependency resolution at configuration time

jvandort (commented):
Skimmed a bit

build-logic/packaging/src/main/kotlin/gradlebuild.distributions.gradle.kts
h0tk3y: Hmm... I don't think this resolves dependencies at configuration time. `filter` resolves to the member `FileCollection.filter(Spec<File>)` returning a lazy live view.
`classpath.from(...)` doesn't query the collection either until after configuration phase.

If the member-vs-extension ambiguity is the concern, I'm happy to pin the overload explicitly — `runtimeClasspath.filter(Spec { jar -> ... })` — so the lazy semantics are visible at the call site.

bamboo:
Superseded by https://github.com/gradle/gradle/pull/38847

PR #38615, opened 2026-07-24
Reattach non-abstract property owners after configuration cache restore

### LINKED ISSUE #37421: Private non-abstract properties lose their owner after Configuration Cache roundtrip
Private non-abstract Property with a non-final getter doesn't have its owner restored after being deserialized by Configuration Cache, failing this test:
https://github.com/gradle/gradle/blob/5c4dca965b9792403dbd0ecdfec732374e1dcd39/subprojects/core/src/integTest/groovy/org/gradle/api/tasks/TaskPropertiesIntegrationTest.groovy#L216-L251

Expected failure message: `Cannot query the value of task ':thing' property 'count' because it has no value available.`
Actual failure message: `Cannot query the value of this property because it has no value available.`

The root cause seems to be the attachment mechanism that is injected into the getter, so internal access (which goes through the field) does not re-attach the property to its owner after restoring from CC.

See #33215, #24614.

Develop-KIM:
Fixes #37421

### Context

A non-abstract `Property` (or `ConfigurableFileCollection`) getter that returns a backing field has its owner attached **lazily**, by the getter override that the class generator injects. That override only runs when the getter is actually invoked during configuration.

When the owning object is restored from the configuration cache, configuration is skipped, so the getter is never called and the owner is never attached. Internal access through the backing field then reports a generic error instead of naming the owning task and property:

| Execution | Error message |
| --- | --- |
| No configuration cache | `Cannot query the value of task ':thing' property 'count' because it has no value available.` |
| With configuration cache | `Cannot query the value of this property because it has no value available.` |

This is the divergence tracked by the `@ToBeImplemented` test in `TaskPropertiesIntegrationTest`.

### Changes

Abstract managed properties are already re-attached on cache read through the generated `attachModelProperties()` method — `ManagedPropertiesHandler` registers them for on-demand reattachment (extended to `ConfigurableFileCollection` in bab3822). Non-abstract *convention* properties whose owner is attached lazily through an overriding getter were never registered for this reattachment, so they lost their owner after a cached run.

`ExtensibleTypePropertyHandler` now registers such convention properties (a reattachable type with an attachable overriding getter) for on-demand reattachment, so `attachModelProperties()` restores their owner after deserialization — mirroring the existing behavior for abstract managed properties. This only affects the configuration-cache restore path; construction behavior is unchanged.

### Tests

- Updated the `@ToBeImplemented` integration test in `TaskPropertiesIntegrationTest` to assert the now-consistent message under both executers.
- Added a unit test in `AsmBackedClassGeneratorDecoratedTest` that reads the property back through the field (as happens on cache restore) and verifies `attachModelProperties()` reattaches its owner.

<sub>Prepared with AI assistance.</sub>

### Contributor Checklist
- [x] Ensure that tests pass locally: `./gradlew model-core:test --tests "*AsmBackedClassGeneratorDecoratedTest"`.

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
@@ -426,6 +426,15 @@ abstract class AbstractClassGenerator implements ClassGenerator {
         return isAttachableType(metadata.getReturnType(), metadata.method::isAnnotationPresent);
     }
 
+    private static boolean hasAttachableOverridableGetter(PropertyMetadata property) {
+        for (MethodMetadata getter : property.getOverridableGetters()) {
+            if (isAttachableMethod(getter)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
     private static boolean isAttachableType(Class<?> type, Predicate<Class<? extends Annotation>> propertyHasAnnotation) {
         // This should apply to all 'managed' types however only the ConfigurableFileCollection and Provider types and @Nested value current implement OwnerAware
         return Provider.class.isAssignableFrom(type) || isConfigurableFileCollectionType(type) || hasNestedAnnotation(propertyHasAnnotation);
@@ -1039,9 +1048,16 @@ abstract class AbstractClassGenerator implements ClassGenerator {
                 visitor.mixInConventionAware();
             }
             for (PropertyMetadata property : conventionProperties) {
-                boolean applyRole = isAttachProperty(property) && isLazyAttachPropertyIfNeeded(property) && isRoleType(property);
+                boolean attachProperty = isAttachProperty(property) && isLazyAttachPropertyIfNeeded(property);
+                boolean applyRole = attachProperty && isRoleType(property);
                 if (applyRole) {
                     visitor.instantiatesNestedObjects();
+                }
+                if (attachProperty && isReattachProperty(property) && hasAttachableOverridableGetter(property)) {
+                    // The owner is attached lazily by the overriding getter, which is not invoked when the
+                    // object is restored from the configuration cache. Reattach it on demand so the property
+                    // keeps its display name after a cached run. See https://github.com/gradle/gradle/issues/37421.
+                    visitor.attachOnDemand(property, applyRole);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


abstratt:
@Develop-KIM I had very similar changes sitting locally, the only differences between your changes and mine were:
- my changes had some additional test coverage
- your changes have a stricter guard for eligible properties (I will post a question about that in a minute)

Not sure I will merge your PR or go ahead with a new one, but either way, you will be credited.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
```diff
@@ -426,6 +426,15 @@ private static boolean isAttachableMethod(MethodMetadata metadata) {
         return isAttachableType(metadata.getReturnType(), metadata.method::isAnnotationPresent);
     }
 
+    private static boolean hasAttachableOverridableGetter(PropertyMetadata property) {
```
abstratt, line 429: ❓ 

@Develop-KIM What concrete scenario is this meant to protect against?

Develop-KIM: It keeps the on-demand reattach in lockstep with how the owner gets attached in a normal (non-CC) run.

On this decorated/convention path the owner is only ever attached through a generated overriding getter, and the getter loop in the `ClassGenerationVisitor` overload only attaches it when `attachProperty && isAttachableMethod(getter)`. So a property actually gets an owner-attaching getter iff `attachProperty && hasAttachableOverridableGetter(property)`. `isReattachProperty` is exactly the moment that getter is bypassed — on cache restore the value comes back through the field, so I want to reattach only for properties that (a) would have had their owner attached by a generated getter and (b) lose it because the field is re-read.

`isReattachProperty` on its own isn't enough because it's a pure type check (`Property`/`MapProperty`/`ConfigurableFileCollection`). The case it lets through: a read-only `ConfigurableFileCollection` (or `Property`) with a `final`/non-overridable getter. `isLazyAttachPropertyIfNeeded` still returns true for it (it falls through to `!hasPropertyType`) and `isAttachProperty` is true, so `attachProperty && isReattachProperty` both hold — but with no overridable getter a normal run never attaches an owner to it. Without `hasAttachableOverridableGetter`, the restore path would hand it an owner it never has in a non-cached run, i.e. a CC-vs-non-CC difference in the other direction.

The managed/abstract path lower down doesn't need the extra check because every property there is one the generator fully owns, so that mismatch can't arise; this path sees the user's concrete class, so it can.

And no worries either way on whose branch lands — if you'd rather go with yours for the extra test coverage, that's totally fine by me.

abstratt:
Superceded by #38649

PR #38564, opened 2026-07-17
Add TestKit API for verifying that configuration cache is a hit/miss/discarded

jbartok:
Fixes
* https://github.com/gradle/gradle/issues/38223

### Contributor Checklist
- [ ] [Review Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md).
- [ ] Make sure that all commits are [signed off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff) to indicate that you agree to the terms of [Developer Certificate of Origin](https://developercertificate.org/).
- [ ] Make sure all contributed code can be distributed under the terms of the [Apache License 2.0](https://github.com/gradle/gradle/blob/master/LICENSE), e.g. the code was written by yourself or the original code is licensed under [a license compatible to Apache License 2.0](https://apache.org/legal/resolved.html).
- [ ] Check ["Allow edit from maintainers" option](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) in pull request so that additional changes can be pushed by Gradle team.
- [ ] Provide integration tests (under `<subproject>/src/integTest`) to verify changes from a user perspective.
- [ ] Provide unit tests (under `<subproject>/src/test`) to verify logic.
- [ ] Update User Guide, DSL Reference, and Javadoc for public-facing changes.
- [ ] Ensure that tests pass sanity check: `./gradlew sanityCheck`.
- [ ] Ensure that tests pass locally: `./gradlew <changed-subproject>:quickTest`.

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

Issue #35595, opened 2025-11-07
TestKit should provide API for verifying configuration cache compatibility and inputs used


joshfriend:
### Expected Behavior

As a plugin author, I should be able to write tests using TestKit that verify that configuration cache is stored/reused and verify the exact inputs tracked.

### Current Behavior (optional)

We have to resort to parsing the report manually. I have some code for that [in Spotlight](https://github.com/joshfriend/spotlight/blob/99d1b03be708134e0da9c011dcfede899d96115f/spotlight-gradle-plugin/src/functionalTest/kotlin/com/fueledbycaffeine/spotlight/functionaltest/fixtures/BuildResult.kt#L25-L28), and [androidx has a similar parsing tool](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/integration-test/framework/src/main/java/com/android/build/gradle/integration/common/fixture/ConfigurationCacheReportParser.kt;l=236-248?q=begin-report-data).

Since this is digging deep into internal stuff that Gradle isn't supporting officially or intending to be used this way, it can break at any time.

### Context

With stable configuration cache and upcoming isolated projects features, it is important that plugin authors have a robust test suite that verifies compatibility with these features. This has been a pain point for us for a long time as we write tests when [a gradle version update breaks this parsing](https://github.com/joshfriend/spotlight/pull/58#discussion_r2503851450) we do by changing the logged message or [formatting in the report](https://github.com/gradle/gradle/pull/34562).

adam-enko:
Parseable CC data would also be useful for Dokka. The integration tests parse the CC report JSON.
https://github.com/Kotlin/dokka/blob/v2.1.0/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/gradleCCReportData.kt

ljacomet:
This feature request is in the backlog of the relevant team and is prioritized by them.

ov7a:
Related:
* https://github.com/gradle/gradle/issues/27964

eskatos:
For verifying hit/miss and the reason we already have build operations. We could relay them via the Tooling API so TestKit can consume these and offer an API to make assertions.

[...]

Issue #26019, opened 2023-08-03
Allow tracking configuration cache inputs


mathjeff:
### Expected Behavior

We think it would be nice if Gradle could help us detect unexpected inputs to the configuration cache by offering an api (whether text in a file, or an object in a callback to a BuildListener, or something like that) that enumerates the inputs that were read at configuration time.

### Current Behavior (optional)

Gradle doesn't currently seem to provide a conveniently machine-readable list of configuration cache inputs.

### Context

In AndroidX we'd like to avoid inadvertently invalidating the configuration cache by reading an unexpected input.

We think having a machine-readable list of configuration cache inputs would help us to detect unexpected inputs in the future (we could fail the build if unexpected inputs are detected), and potentially avoid them or report issues.

We found that the Android Gradle Plugin uses something like this: https://cs.android.com/android-studio/platform/tools/base/+/762ea0ce6e05742fb83a87cb701c80f7ad411265:build-system/integration-test/framework/src/main/java/com/android/build/gradle/integration/common/fixture/ConfigurationCacheReportChecker.kt

but it operates by directly parsing the configuration cache html report and we're not sure whether this format will be consistent over time.

Thanks!

ov7a:
Thank you for your interest in Gradle!

This issue needs a decision from the team responsible for that area. They have been informed. Response time may vary.

***

Your use case is reasonable. Whether this would be implemented or how is on the responsible team.

mlopatkin:
Thank you for your interest in Gradle!

This feature request is in the backlog of the relevant team.

mlopatkin:
Another use of these can be to expose the list of inputs to an IDE so it knows when to re-sync.

PR #38618, opened 2026-07-24
Add build logic API for verifying that configuration cache is a hit/miss/discarded

### LINKED ISSUE #27964: Make it easy to confirm configuration cache was reused
### Expected Behavior

It would be nice if there were an easy way for a build to programmatically confirm that the configuration cache had been reused, and if the cache couldn't be reused, to see a brief reason why.

At the moment it seems that it is possible to search the output of Gradle for the message `Reusing configuration cache.` although hard to be sure that that message will be stable over time (and it also requires some effort to programmatically access the build logs).

### Context

In AndroidX we run our CI Gradle builds twice in a row to check for out-of-date tasks. When we updated to Gradle 8.1, we [noticed](https://issuetracker.google.com/278767328) our configuration cache entries weren't being used in the second build anymore. It was sort of a coincidence that we noticed this, because the text of the message was noticed as suspicious by another system.

It could be nice if we could validate that the configuration cache entries get used in our up-to-date builds, to make sure that we can detect configuration cache invalidations even if the text of the message isn't one we've seen before.

This issue is similar to https://github.com/gradle/gradle/issues/26019 since they both relate to the idea of making it easy to verify that the configuration cache wasn't unexpectedly invalidated - implementing one might make the other one less important.

Thanks!

reinsch82:
Fixes
* https://github.com/gradle/gradle/issues/27964

> [!NOTE]
> Was stacked on #38564 (now merged); rebased onto `master` and retargeted. The branch contains only this PR's commits.

### Context

Gives build logic a stable way to observe the configuration cache outcome of the current build invocation ([product decision](https://github.com/gradle/bt-product-decisions/issues/29)):

- New public enum `org.gradle.api.configuration.ConfigurationCacheOutcome` — value-identical to TestKit's `org.gradle.testkit.runner.ConfigurationCacheOutcome` from #38564 (all seven values, including `NOT_ENABLED`). Only the Tooling API *event* layer has no `NOT_ENABLED`: the outcome operation is emitted solely when the configuration cache is on, so a CC-off build sends no event, and TestKit and this provider both totalize that silence into `NOT_ENABLED`.
- New `FlowProviders.getConfigurationCacheOutcome()`, modeled on `getBuildWorkResult()`: value available once scheduled work has completed, consumable only from dataflow actions (outside the cached task graph). `NOT_ENABLED` when CC is off. One outcome per invocation, observed by all builds of a composite.
- The outcome is memoized in `ConfigurationCacheProblems`, so console message, entry-outcome build operation and this provider cannot disagree. Serialized by reference like `BuildWorkResultProvider`; rejected as task input and in the fingerprint context.

Primary use case (AndroidX): fail the second of two identical CI runs when the entry was not reused — see the init-script integration test and the new user-guide section, which shows how to read the CI toggle safely (at flow-action execution time).

### Design notes for reviewers

- Querying the outcome freezes the fate and its problem count; `beforeComplete` reuses them. Memoization is lazy — builds that never query keep exact #38564 behavior.
- With the parallel configuration cache, a partial reuse reports `UPDATED` — strict reuse gates should accept only `REUSED` (tested and documented).
- Invalidation *reasons* are out of scope (#35595 / #26019). TestKit's enum stays separate (`TaskOutcome` precedent, cross-version `UNDETERMINED` semantics).

### Verification

11 new integration tests covering every outcome and the misuse guards; existing flow-scope and outcome-op suites green; `sanityCheck` passes (`@Incubating` + `@since 9.8.0`, no accepted-api-changes entries).

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


lkasso (approved):
Docs LGTM.

eskatos:
@bot-gradle test this

mlopatkin:
Please don't proceed without a review from bt-cortex.

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -91,6 +91,19 @@ The outcome is also available to any Tooling API client through a new [`CONFIGUR
 
 See the [Testing with the Configuration Cache](userguide/test_kit.html#sub:test-kit-configuration-cache) section in the Gradle User Manual for more details.
 
+#### Build logic can verify the Configuration Cache outcome
+
+Build logic can now observe the same outcome from within the build, through a new [`FlowProviders.getConfigurationCacheOutcome()`](javadoc/org/gradle/api/flow/FlowProviders.html#getConfigurationCacheOutcome--) lifecycle event provider that can be used as an input to [dataflow actions](userguide/dataflow_actions.html#verifying_configuration_cache_reuse):
```
eskatos: 💅 
```suggestion
Build logic can now observe the same outcome from within the build, through a new [`FlowProviders.getConfigurationCacheOutcome()`](javadoc/org/gradle/api/flow/FlowProviders.html#getConfigurationCacheOutcome--) lifecycle event provider that can be used as an input to [dataflow actions](userguide/dataflow_actions.html#dataflow_action):
```

platforms/documentation/docs/src/docs/release/notes.md
```diff
@@ -91,6 +91,19 @@ The outcome is also available to any Tooling API client through a new [`CONFIGUR
...
+
+```groovy
+flowScope.always(VerifyConfigurationCacheReuse) {
+    parameters.outcome = flowProviders.configurationCacheOutcome
+}
+```
+
+This makes it possible, for example, to fail a CI build when an expected Configuration Cache hit did not happen, without parsing console output.
+When the Configuration Cache is not enabled, the provider reports [`NOT_ENABLED`](javadoc/org/gradle/api/configuration/ConfigurationCacheOutcome.html#NOT_ENABLED).
+
```
eskatos, line 124: 💅 

```suggestion

For more information, see the [dedicated section](userguide/dataflow_actions.html#verifying_configuration_cache_reuse) in the user manual.

```

subprojects/core-api/src/main/java/org/gradle/api/configuration/ConfigurationCacheOutcome.java
```diff
@@ -0,0 +1,90 @@
...
+ * <p>
+ * TestKit exposes the same information to plugin tests via
+ * {@code org.gradle.testkit.runner.ConfigurationCacheOutcome}, which additionally reports
+ * outcomes across Gradle versions.
+ *
+ * @see org.gradle.api.flow.FlowProviders#getConfigurationCacheOutcome()
+ * @since 9.8.0
+ */
+@Incubating
+public enum ConfigurationCacheOutcome {
```
mlopatkin: 🤔 Having this as an enum is going to limit our ability to evolve the result. E.g. there is no way to expose the invalidation message through it if the users ask us to.

We cannot use sealed classes here yet, though they'll be a good fit. We can emulate them with nested child classes though:

```java
public abstract class ConfigurationCacheOutcome {
    private ConfigurationCacheOutcome() {}
    
    public static final class Stored extends ConfigurationCacheOutcome {
        private Stored() {}
    }
    
    public static Stored stored() { return new Stored(); }
    
    public static final class NotEnabled extends ConfigurationCacheOutcome {
        private NotEnabled() {}
    }
    
    public static NotEnabled notEnabled() { return new NotEnabled(); }
    
    // Basic equivalence checks assuming no subclasses carry data.
    public final boolean equals(@Nullable Object rhs) {
        return (rhs != null && getClass().equals(rhs.getClass()));
    }
    
    public final int hashCode() { return getClass().hashCode(); }
}
```

Of course, the recommended way of checking then becomes
```kotlin
require(result is ConfigurationCacheOutcome.Stored)
```

It is less prominent for TestKit use case, because there we can just add more stuff into the BuildResult, though we may decide to align that API too.

subprojects/core-api/src/main/java/org/gradle/api/configuration/ConfigurationCacheOutcome.java
```diff
@@ -0,0 +1,90 @@
...
+    DISCARDED,
+
+    /**
+     * No configuration cache entry was stored on purpose: no reusable entry was found while the cache
+     * is in read-only mode, or configuration caching was degraded gracefully because incompatible
+     * tasks were scheduled.
+     *
+     * @since 9.8.0
+     */
+    NOT_STORED,
```
mlopatkin: :question: I'm a bit concerned about the differences between DISCARDED and NOT_STORED, and whether we're exposing too much internal details today.

Neither maps cleanly onto user-facing concepts. For example, if CC entry is deliberately not saved, you can see both DISCARDED (for an incompatible task) and NOT_STORED (graceful degradation, read-only). OTOH, DISCARDED can also be caused by a failure to save the entry (typically accompanied by a build failure).

WDYT if we change it to `STORE_FAILED` (hard failure to save) and `STORE_SKIPPED` (deliberately skipped because of an incompatible task/feature/read-only-mode-miss)?

subprojects/core-api/src/main/java/org/gradle/api/configuration/ConfigurationCacheOutcome.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * The configuration cache was enabled, but the build finished before the outcome was
+     * determined, e.g. because it failed early.
+     *
+     * @since 9.8.0
+     */
+    UNDETERMINED
```
mlopatkin: :question: I don't think we run user code that can observe the state before the cache action is determined. So maybe we can skip exposing it in the public API.

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/dataflow_actions.adoc
```diff
@@ -94,6 +94,40 @@ flowProviders.buildWorkResult.map {
...
+            def outcome = parameters.outcome.get()
+            if (outcome != ConfigurationCacheOutcome.REUSED) {
+                throw new GradleException("Expected the configuration cache entry to be reused but the outcome was: $outcome")
+            }
+        }
+    }
+}
+----
+
+NOTE: A partially reused entry — possible for example with the parallel configuration cache or under <<isolated_projects.adoc#isolated_projects,isolated projects>> — is reported as `UPDATED`, not `REUSED`. A strict reuse check like the one above should only accept `REUSED`.
```
mlopatkin: :x: "parallel configuration cache" has nothing to do with partially reused entries. This is IP-exclusive.

platforms/core-configuration/core-flow-services-api/src/main/java/org/gradle/api/flow/FlowProviders.java
```diff
@@ -79,4 +80,57 @@ public interface FlowProviders {
...
+     * <p>
+     * The returned {@link Provider#get() provider's value} becomes available after the scheduled work
+     * has completed, so it can only be used as an input to {@link FlowAction dataflow actions}.
+     * When the configuration cache is not enabled, the value is
+     * {@link ConfigurationCacheOutcome#NOT_ENABLED}.
+     * </p>
+     * <p>
+     * There is a single outcome per build invocation: in a composite build, dataflow actions of all
+     * builds observe the outcome of the invocation as a whole. A nested build started by a
+     * {@code GradleBuild} task is a separate invocation without configuration caching, so its dataflow
```
mlopatkin: :nail_care: I would avoid bringing `GradleBuild` into the conversation here. Sure, it cannot start a build with CC enabled, but this is accidental and not related to the contract of the flow provider - the provider just faithfully represent the (sad) reality.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/FlowProvidersConfigurationCacheOutcomeIntegrationTest.groovy
```diff
@@ -0,0 +1,329 @@
...
+        settingsFile ''
+        buildFile """
+            tasks.register('nested', GradleBuild) {
+                dir = 'nested'
+                tasks = ['help']
+            }
+        """
+
+        when:
+        configurationCacheRunLenient 'nested'
```
mlopatkin: :nail_care: AFAICT, no need to `RunLenient` here, this is CC-compatible.

platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/FlowProvidersConfigurationCacheOutcomeIntegrationTest.groovy
```diff
@@ -0,0 +1,329 @@
...
+        then: 'serializing the provider as a task input is reported as a problem'
+        failure.assertHasFailures(2)
+        failureDescriptionContains 'can only be used as input to flow actions'
+
+        and: 'reading the provider at execution time fails because the value is not yet available'
+        failureDescriptionContains "Execution failed for task ':printOutcome'"
+        failureCauseContains "Cannot access the value of 'ConfigurationCacheOutcome' before it becomes available!"
+    }
+
+    def "flow action observes UPDATED when the entry is partially reused"() {
```
mlopatkin: :x: This test relies on a bug in CC. `UPDATED` should be restricted to Isolated Projects and only to model building. In this test, we emit Update message by mistake - not saying it should be fixed in this PR. 

However, I suggest we replace UPDATED with STORE/STORE_FAILED for the first release, until we get more IP coverage. But in any case, some IP-specific test coverage is necessary.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
mlopatkin: 🤔 It looks like we may also discard the entry if the problems are not ignored, but we don't yet check for it here. We would need to tweak this to distinguish STORED vs. STORE_FAILED then.

mlopatkin (changes requested):
I did a bit more thorough review of the outcomes, both for the actual set and the shape of the API. It might be beneficial to tweak the TAPI-exposed model later too.

subprojects/core-api/src/main/java/org/gradle/api/configuration/ConfigurationCacheOutcome.java
eskatos: I would be in favor of aligning the TestKit API

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

reinsch82:
@bot-gradle test this

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/dataflow_actions.adoc
```diff
@@ -94,6 +94,40 @@ flowProviders.buildWorkResult.map {
...
+    interface Parameters extends FlowParameters {
+        @Input Property<ConfigurationCacheOutcome> getOutcome()
+    }
+
+    void execute(Parameters parameters) {
+        // Read the toggle inside the action, at execution time: reading it at configuration
+        // time would make it a configuration input and invalidate the very entry to verify.
+        if (System.getenv('VERIFY_CC_REUSE') != null) {
+            def outcome = parameters.outcome.get()
+            if (outcome != ConfigurationCacheOutcome.reused()) {
```
mlopatkin: :nail_care: It is Groovy, so it is probably okay (it falls back to equals), but I'm afraid Java folks can blindly copy-paste and get it "working" until we actually start producing multiple instances.

I'd change the check to

```groovy
if (outcome !instanceof ConfigurationCacheOutcome.Reused) {
```

platforms/core-configuration/flow-services/src/main/kotlin/org/gradle/internal/flow/services/ConfigurationCacheOutcomeSource.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import org.gradle.api.configuration.ConfigurationCacheOutcome
+import org.gradle.internal.service.scopes.Scope
+import org.gradle.internal.service.scopes.ServiceScope
+
+/**
+ * Provides the outcome of configuration caching for the current build invocation.
+ *
+ * Registered by the configuration cache infrastructure only when the configuration cache is
+ * enabled; when absent, the outcome is [ConfigurationCacheOutcome.NOT_ENABLED].
```
mlopatkin: :x: `ConfigurationCacheOutcome.NotEnabled`

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
```diff
@@ -497,13 +517,69 @@ class ConfigurationCacheProblems(
...
+        !::cacheAction.isInitialized -> null
+        cacheAction is Load -> EntryOutcomeKind.REUSED
+        cacheAction == SkipStore -> EntryOutcomeKind.STORE_SKIPPED
+        // Store or Update from here on
+        entryFinalizeResult == EntryFinalizeResult.COMMITTED -> EntryOutcomeKind.STORED
+        seenSerializationErrorOnStore -> EntryOutcomeKind.STORE_FAILED
+        shouldDegradeGracefully() -> EntryOutcomeKind.STORE_SKIPPED
+        incompatibleTasks.isNotEmpty() && summary.consoleProblemCount == 0 -> EntryOutcomeKind.STORE_SKIPPED
+        // Discarded due to problems, too many problems, or never stored because the build failed early
+        else -> EntryOutcomeKind.STORE_FAILED
```
mlopatkin, line 604: :x: there we have `incompatibleTasks.isEmpty() || summary.consoleProblemCount > 0`. `summary.consoleProblemCount > 0` doesn't necessary mean that the store failed. With `--configuration-cache-problems=warn` and an incompatible task present, we'll not fail the build and skip store deliberately too.

This should be handled properly, otherwise we're breaking the API contract.

FWIW, the console message is a bit confusing ("Entry discarded with N problems" without any mention of an incompatible task), but it is factually correct; users may conclude that the problems to discard the entry, but the message doesn't say so. I'm not saying we should change the console message.

platforms/core-configuration/core-flow-services-api/src/main/java/org/gradle/api/flow/FlowProviders.java
```diff
@@ -79,4 +80,55 @@ public interface FlowProviders {
...
+     *
+     *     {@literal @}Inject
+     *     VerifyConfigurationCacheReusePlugin(FlowScope flowScope, FlowProviders flowProviders) {
+     *         this.flowScope = flowScope;
+     *         this.flowProviders = flowProviders;
+     *     }
+     *
+     *     {@literal @}Override
+     *     public void apply(Settings target) {
+     *         flowScope.always(VerifyReuse.class, spec -&gt;
```
mlopatkin, line 141: :nail_care: It would be good to add `VerifyReuse` and its parameters interface here too.

mlopatkin (changes requested):
The current shape looks good. There are some corner cases and rough edges I'd like to see addressed before we can merge this, but no more dramatic changes.

I'd also appreciate an explicit review for TAPI changes from @gradle/bt-tide, to make sure we're on the same page with regard to how the APIs look now.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/problems/ConfigurationCacheProblems.kt
jbartok: I've attempted a fix, please check.

jbartok: The test failures on the PR are related to this change. @mlopatkin is the change wrong or the tests?

mlopatkin: > The test failures on the PR are related to this change. [@mlopatkin](https://github.com/mlopatkin) is the change wrong or the tests?

Both, actually. [One](https://ge.gradle.org/s/bgfseuirrslou/tests/task/:configuration-cache:embeddedIntegTest/details/org.gradle.internal.cc.impl.ConfigurationCacheEntryOutcomeBuildOperationIntegrationTest/emits%20STORE_FAILED%20when%20there%20are%20too%20many%20problems?top-execution=1) is elevating the warnings to failure as there are too many of those, but we report STORE_SKIPPED instead of FAILED due to the new condition. 

[Others](https://ge.gradle.org/scans/tests?search.names=gitCommitId%2CPartOf&search.tags=CI&search.timeZoneId=Europe%2FBerlin&search.values=6b6b2844cfea9b8f57b333237f8ef0ffa3a463c0%2CQuickFeedbackLinuxOnly&tests.container=org.gradle.internal.cc.impl.ConfigurationCacheIncompatibleTasksIntegrationTest) are passing builds, so it would be incorrect to expect store to fail per the guidelines. The detected problem is ignored by the warning mode and the incompatible task present causes the entry to be discarded.

[IP test failures](https://ge.gradle.org/scans/tests?search.names=gitCommitId%2CPartOf&search.tags=CI&search.timeZoneId=Europe%2FBerlin&search.values=6b6b2844cfea9b8f57b333237f8ef0ffa3a463c0%2CQuickFeedbackLinuxOnly&tests.container=org.gradle.internal.cc.impl.isolated.IsolatedProjectsIgnoreProblemsIntegrationTest) also seem to have their expectations backwards.

platforms/documentation/docs/src/docs/userguide/reference/gradle-types/dataflow_actions.adoc
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2024 Gradle, Inc.
+    // Copyright (C) 2024 Gradle, Inc.
```
mlopatkin, line 1: 💅  Accidental whitespace

mlopatkin (commented):
We'll have to fix the tests before merging; likely fixing the "too many warnings lead to failure"  should be enough for that.

PR #38622, opened 2026-07-24
Honor explicit opt-out for Isolated Projects options

### LINKED ISSUE #38598: 9.7 RC: org.gradle.unsafe.isolated-projects=false does not disable IP when opt-in via org.gradle.isolated-projects=true
### Current Behavior

Project Isolation is active, breaking builds when the project is in fact not compatible

### Expected Behavior

No Project isolation, the unsafe option is a StartParameter alias

### Context (optional)

I use the new property name `org.gradle.isolated-projects=true` for my JVM project https://github.com/hfhbd/adventOfCode. I also use the GitHub dependency submission plugin that is not compatible with Project Isolation yet:
https://github.com/gradle/github-dependency-graph-gradle-plugin/issues/109, so it uses the (old) property: `org.gradle.unsafe.isolated-projects=false` causing a failing build when the main project uses the new property name as mentioned in the Gradle user guide.

For my use-case, the dependency submission plugin could just pass the old and the new property name as a workaround, but Gradle StartParameters should support the alias.

Use a project that is incompatible with Project Isolation

```properties
# gradle properties
org.gradle.isolated-projects=true
```

```shell
./gradlew build org.gradle.unsafe.isolated-projects=false
```

### Gradle version

9.7.0-rc-1

9.7.0-milestone-2

### Build scan URL (optional)

https://gradle.com/s/fk4qhwh3c6hc2

_No response_

alllex:
Fixes #38598

### Summary

Isolated Projects can now always be disabled through its deprecated `org.gradle.unsafe.isolated-projects=false` property, even when the build enables the feature under the stable name. A stop-gap scoped to the three Isolated Projects options; Configuration Cache alias behavior is unchanged.

### Why

Build options resolve against a single map into which all property sources (`gradle.properties`, command-line `-D`, …) have already been merged, and an alias pair resolves by name: the stable name, when present in any source, shadows the deprecated one. Tooling that still opts out under the deprecated name — the `dependency-submission` GitHub Action unconditionally passes `-Dorg.gradle.unsafe.isolated-projects=false` ([main.ts#L49](https://github.com/gradle/actions/blob/6550634d3eb14a20275549de1588a83267023d42/sources/src/actions/dependency-submission/main.ts#L49)) — therefore cannot disable the feature once a build adopts `org.gradle.isolated-projects=true` as the 9.7 docs recommend, and fails on IP problems despite an explicit opt-out. This is not an RC1 code regression: resolution has worked this way since the alias pair was introduced; 9.7 promoting the new name is what exposed it.

### What

- New `DeprecatedAliasDisableWinsBooleanBuildOption`: when both names of an alias pair are set, the effective value is their conjunction — an explicit `false` from either name wins
- `IsolatedProjectsOption`, `IsolatedProjectsDiagnosticsOption`, `IsolatedProjectsDangerouslyIgnoreProblemsOption` migrate to it; no other option touched
- Characterization test pins the unchanged Configuration Cache alias behavior
- The rule is documented in `isolated_projects.adoc` and the 9.7 upgrading guide

### Details

Consumer impact — exactly one resolution outcome changes (values parse per `BooleanOptionUtil.isTrue`, so any non-`"true"` value counts as an explicit false):

| stable name | deprecated name | before | after |
| --- | --- | --- | --- |
| `true` | `false` | IP enabled | **IP disabled** |
| any other combination | | unchanged | unchanged |

`--isolated-projects` / `--no-isolated-projects` continue to trump all properties.

Deliberate choices:

- **Conjunction, not source precedence.** The proper solution is to make the deprecated name an actual alias, resolved against the main option within each property source rather than after all sources are merged. That touches shared launcher/daemon property-merging machinery — too invasive for an RC fix. The accepted downside of the conjunction: it ignores source precedence, so a disable from a lower-precedence source beats an enable from a higher-precedence one when the two use different names. It fails safe — the rule can only ever turn the feature off.
- **Guarded override rather than "deprecated false always wins".** The override delegates to the base class except in the single case where base resolution gives the wrong value. This keeps origin attribution precise (the deprecated name is only credited when it changed the outcome) and leaves the base class's deprecated-name fallback path intact for the planned deprecation nag, which records usage there.
- **Configuration Cache options not migrated.** Their stable-name-wins behavior has stood since Gradle 8.1; changing it is out of scope for an RC fix. The new characterization test turns any accidental leakage red.
- **No mixed-usage warning in this change.** Property resolution runs in the launcher JVM and is not once-per-build; a reliable warning needs lifecycle-point detection and is deferred to a follow-up.
- **Explicit 16-row truth table over a `combined:` cross-product.** The table (name placements file/cli × values) is the specification; a what-if run against future per-source semantics showed the acceptance delta is exactly 4 cells, so migrating to the proper solution later means flipping those cells and nothing else.

Evidence: 16-iteration integration truth table plus unit truth table including origin attribution (`:build-option`, 106 tests); full `IsolatedProjectsIntegrationTest` green (35 tests); CC pin test green; `sanityCheck` green.

Review: scrutinize `DeprecatedAliasDisableWinsBooleanBuildOption.getFromProperties` (the entire behavior change, ~8 lines) and the `expected` column of the truth-table test; the three base-class swaps and doc sentences are skimmable.

### Contributor Checklist

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheEnablementIntegrationTest.groovy
+++ b/platforms/core-configuration/configuration-cache/src/integTest/groovy/org/gradle/internal/cc/impl/ConfigurationCacheEnablementIntegrationTest.groovy
@@ -18,6 +18,7 @@ package org.gradle.internal.cc.impl
 
 import org.gradle.initialization.StartParameterBuildOptions.ConfigurationCacheOption
 import org.gradle.integtests.fixtures.configurationcache.ConfigurationCacheFixture
+import spock.lang.Issue
 
 import static org.gradle.initialization.StartParameterBuildOptions.ConfigurationCacheProblemsOption.Value.WARN
 
@@ -99,6 +100,24 @@ class ConfigurationCacheEnablementIntegrationTest extends AbstractConfigurationC
         cliOrigin         | cliArgument
         "long option"     | DISABLE_CLI_OPT
         "system property" | DISABLE_SYS_PROP
+    }
+
+    @Issue("https://github.com/gradle/gradle/issues/38598")
+    def "deprecated property set to false does not disable configuration cache enabled via gradle.properties"() {
+        // Characterization test: for Configuration Cache the stable property name deliberately wins over
+        // the deprecated spelling regardless of value, unchanged since Gradle 8.1. The Isolated Projects
+        // options resolve the same situation as a conjunction (explicit disable wins) since issue 38598;
+        // flipping this test is a conscious decision to migrate CC to those semantics.
+        given:
+        file('gradle.properties') << """
+            $ENABLE_GRADLE_PROP
+        """
+
+        when:
+        run 'help', "-D${ConfigurationCacheOption.DEPRECATED_PROPERTY_NAME}=false"
+
+        then:
+        fixture.assertStateStored()
     }
 
     def "can enable configuration cache using incubating and final property variants"() {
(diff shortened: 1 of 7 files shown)


6hundreds (commented):
LGTM!

cobexer (approved):

PR #38628, opened 2026-07-25
Add @HasInternalProtocol to FileCollectionDependency and guard internal cast

### LINKED ISSUE #37544: `FileCollectionDependency` is not `@HasInternalProtocol`, but hard casts to internal type.
### Current Behavior

`FileCollectionDependency`s are hard cast to `SelfResolvingDependencyInternal` when resolved, leading to ClassCastEceptions like the following:
```
class io.github.cputnama11y.deptransformation.wiring.CapabilityInfoCarrierDependency_Decorated cannot be cast to class org.gradle.api.internal.artifacts.dependencies.SelfResolvingDependencyInternal (io.github.cputnama11y.deptransformation.wiring.CapabilityInfoCarrierDependency_Decorated is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader$InstrumentingVisitableURLClassLoader @634d6ef3; org.gradle.api.internal.artifacts.dependencies.SelfResolvingDependencyInternal is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader @5010be6)
```

### Expected Behavior

`FileCollectionDependency` should be annotated with `@HasInternalProtcol`, a sensible default should be picked if the `FileCollectionDependency` doesn't implement `SelfResolvingDependencyInternal` or, alternatively, `SelfResolvingDependencyInternal` could be exposed as api. Regardless of the chosen solution, there should not be an unexplained `ClassCastException` when resolving the dependency. Other options probably include things I haven't thought about.

### Context (optional)

`FileCollectionDependency` is useful for adding non-published dependency info to a configuration to allow resolution of custom dependencies later down the line, as the kotlin gradle plugin does.

```kotlin
plugins {
    java
}

abstract class CustomFileCollectionDependency : FileCollectionDependency {
    @get:Inject
    abstract val objects: ObjectFactory

    override fun getFiles(): FileCollection = objects.fileCollection()

    override fun getGroup(): String? = null

    override fun getName(): String = "CustomFileCollectionDependency"

    override fun getVersion(): String? = null

    override fun copy(): Dependency = objects.newInstance<CustomFileCollectionDependency>()
    override fun getReason(): String? = null

    override fun because(reason: String?) = Unit
}

dependencies {
    implementation(objects.newInstance<CustomFileCollectionDependency>())
}
```

### Gradle version

9.4.1

### Build scan URL (optional)

_No response_

_No response_

ghost:
### Context

When a custom implementation of FileCollectionDependency is added to a configuration and resolved, Gradle performs an unchecked cast to SelfResolvingDependencyInternal in DefaultLocalVariantGraphResolveStateBuilder, resulting in a ClassCastException. This happens because the public FileCollectionDependency API interface lacks the @HasInternalProtocol annotation that would signal to plugin authors that Gradle's internal machinery expects a specific implementation type, and the getComponentId method in DefaultLocalFileDependencyMetadata unconditionally casts to SelfResolvingDependencyInternal without verifying the runtime type. The fix addresses both aspects by annotating the FileCollectionDependency interface with @HasInternalProtocol to document the internal protocol expectation, and by guarding the cast with an instanceof check so that custom FileCollectionDependency implementations that do not extend SelfResolvingDependencyInternal are handled gracefully by returning null for the component identifier rather than throwing a ClassCastException.

Closes #37544

### Contributor Checklist

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/moduleconverter/dependencies/DefaultLocalVariantGraphResolveStateBuilder.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/moduleconverter/dependencies/DefaultLocalVariantGraphResolveStateBuilder.java
@@ -380,7 +380,10 @@ public class DefaultLocalVariantGraphResolveStateBuilder implements LocalVariant
 
         @Override @Nullable
         public ComponentIdentifier getComponentId() {
-            return ((SelfResolvingDependencyInternal) fileDependency).getTargetComponentId();
+            if (fileDependency instanceof SelfResolvingDependencyInternal) {
+                return ((SelfResolvingDependencyInternal) fileDependency).getTargetComponentId();
+            }
+            return null;
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


ghale:

PR #38633, opened 2026-07-26
Avoid ignoring source and documentation directories named build

### LINKED ISSUE #18737: Default gitignore generated by `init` ignores all directories called `build`
### Expected Behavior

After running `gradle init`, gradle-generated files should be ignored by git, but source code or resources in a directory called `build` should not be ignored.

This could be as simple as ignoring `/build`, at first. It could be more complicated, e.g. [gradle's own .gitignore](https://github.com/gradle/gradle/blob/b430a9a6e44b1b41e42539a73668bbd6fea60c05/.gitignore#L4-L8).

I recognise this is only an initial `.gitignore`, so maybe `/build`, which would ignore the artifacts of a flat project only, is appropriate.

### Current Behavior
`./src/main/java/com/company/build/MyGradlePlugin.java` is ignored by `git add`.

### Context
My build plugin source code was ignored by git until I updated `.gitignore`. It wasn't immediately obvious why.

### Steps to Reproduce 
```sh
$ gradle init --type=basic --dsl=kotlin --project-name=git-test
$ git init
$ mkdir -p build module/build src/main/java/com/company/build
$ touch build/a module/b module/build/c src/main/java/com/company/build/d
$ git add .
$ git status
```
You would expect `a`, `b` and `d` to be added, but `d` is not. It's up for debate if `c` should be included or not, with a basic gitignore.

### Your Environment
```------------------------------------------------------------
Gradle 7.2

Build time:   2021-08-17 09:59:03 UTC
Revision:     a773786b58bb28710e3dc96c4d1a7063628952ad

Kotlin:       1.5.21
Groovy:       3.0.8
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          16.0.2 (Homebrew 16.0.2+0)
OS:           Mac OS X 11.5 x86_64
```

DragonFSKY:
Fixes #18737

### Context

`gradle init` currently generates a bare `build` pattern. Git treats it as matching any file or directory named `build`, so source packages and resources such as `src/main/java/com/example/build/` are ignored as well.

This change follows the root `.gitignore` approach discussed in the issue. Newly generated files use:

```gitignore
build/
!**/docs/**/build/
!**/src/**/build/
```

This keeps root and subproject build output ignored while allowing directories named `build` under source and documentation paths. The parent-directory limitation of Git ignore rules still applies, so a path below an already ignored outer `build/` remains ignored.

Merging into an existing `.gitignore` is deliberately conservative. If the file already contains any rule with a literal `build` path component, including unignore rules or case variants, the generator leaves all build-related rules untouched. This avoids changing the ordering semantics of user-owned rules. As a tradeoff, existing projects with a build rule are not automatically migrated when `init` is rerun.

### Testing

Passed locally:

- `./gradlew :build-init:test --tests org.gradle.buildinit.plugins.internal.GitIgnoreGeneratorTest`
- `./gradlew :build-init:embeddedIntegTest --tests org.gradle.buildinit.plugins.BasicTypeInitIntegrationTest`
- `./gradlew :build-init:checkstyleMain :build-init:codenarcTest :build-init:codenarcIntegTest`

The integration test initializes a real Git repository and uses `git check-ignore --no-index` to verify root and nested build output, source and documentation paths, ignored parent-directory behavior, and non-matching names such as `mybuild`. It also isolates the test from user Git templates and global excludes.

I also ran `./gradlew :build-init:quickTest`. It completed 357 tests with one unrelated failure in `BuildInitSpecsIntegrationTest > gives decent error message when triggered with unknown init-type after loading project specs`: `JavaToolchainFixture.groovy:76` throws an NPE while reading toolchain installations. The focused unit and integration tests above pass.

### AI assistance disclosure

I used Codex extensively to help analyze the issue, implement and review the change, and design the tests. I reviewed the complete diff, validated the Git behavior, and ran the checks listed above.

### Contributor Checklist

- [x] Reviewed the contribution guidelines.
- [x] All commits are signed off under the Developer Certificate of Origin.
- [x] The contributed code can be distributed under the Apache License 2.0.
- [x] Allow edits from maintainers.
- [x] Added integration tests to verify the user-visible behavior.
- [x] Added unit tests for generation, merging, and idempotency.
- [x] No user guide, DSL reference, or Javadoc update is needed because this changes generated ignore defaults without changing public API.
- [ ] `./gradlew sanityCheck`
- [ ] `./gradlew :build-init:quickTest` (one unrelated local failure described above)

### Reviewing cheatsheet

- ❌ ❓ **must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/BasicTypeInitIntegrationTest.groovy
+++ b/platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/BasicTypeInitIntegrationTest.groovy
@@ -101,7 +101,9 @@ build/
     def "generated .gitignore preserves source and documentation directories named build"() {
         given:
         run('init', '--project-name', 'someApp', '--overwrite')
-        assertGitSucceeds('init')
+        def emptyGitTemplate = targetDir.file('empty-git-template').createDir()
+        assertGitSucceeds('-c', "init.templateDir=${emptyGitTemplate.absolutePath}", 'init')
+        assertGitSucceeds('config', 'core.excludesFile', targetDir.file('empty-global-excludes').createFile().absolutePath)
 
         def ignoredPaths = [
             'build/output.txt',
--- a/platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/BasicTypeInitIntegrationTest.groovy
+++ b/platforms/software/build-init/src/integTest/groovy/org/gradle/buildinit/plugins/BasicTypeInitIntegrationTest.groovy
@@ -97,4 +97,41 @@ build/
         where:
         scriptDsl << ScriptDslFixture.SCRIPT_DSLS
     }
+
+    def "generated .gitignore preserves source and documentation directories named build"() {
+        given:
+        run('init', '--project-name', 'someApp', '--overwrite')
+        assertGitSucceeds('init')
+
+        def ignoredPaths = [
+            'build/output.txt',
+            'module/build/output.txt',
+            'nested/module/build/output.txt',
+            'build/src/build/output.txt'
+        ]
+        def visiblePaths = [
+            'src/build/source.txt',
+            'src/main/resources/build/fixture.txt',
+            'module/src/test/resources/build/fixture.txt',
+            'docs/build/page.md',
+            'docs/guide/assets/build/page.md',
+            'module/docs/reference/build/page.md',
+            'mybuild/output.txt'
+        ]
+        (ignoredPaths + visiblePaths).each { targetDir.file(it).createFile() }
+
+        expect:
+        ignoredPaths.every { gitExitValue('check-ignore', '--no-index', it) == 0 }
+        visiblePaths.every { gitExitValue('check-ignore', '--no-index', it) == 1 }
+    }
+
+    private void assertGitSucceeds(String... arguments) {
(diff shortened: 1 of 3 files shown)


ghale:

reinsch82:
Thanks a lot for this contribution, @DragonFSKY — the work here is careful and well-tested, and the deliberately conservative merging into existing `.gitignore` files was a thoughtful touch.

After investigating, we've concluded the pattern-based approach unfortunately can't be made correct: the `!**/docs/**/build/` and `!**/src/**/build/` negations re-include real build output in common layouts (a `docs` subproject's `docs/build/`, and `src/<project>/build/` in monorepos that keep projects under `src/`), which regresses into silently committing build artifacts. Since that's inherent to what ignore patterns can express — not to your implementation — we don't want to send you through further iterations on this route.

The direction we favor is the per-directory self-ignoring `.gitignore` you mentioned being open to, but that is a Gradle-core runtime change that needs a design spec and internal alignment before any implementation, and it isn't currently prioritized. I've written up the full reasoning in https://github.com/gradle/gradle/issues/18737#issuecomment-5115026224.

I'm closing this PR accordingly. That's no reflection on the quality of the work — thank you for digging into a genuinely tricky problem, and for the clear write-up and tests that made this easy to evaluate.

PR #38641, opened 2026-07-27
Remove stale Test Distribution diagnostic properties

cobexer:
Removes two stale Test Distribution diagnostic properties from `gradle.properties`:

- `develocity.internal.testdistribution.logWorkerReadyMessage`
- `develocity.internal.testacceleration.enableCustomValues`

Both were added in a7e7e71 (~18 months ago) to diagnose remote test executors being enqueued in the wrong state. The investigation is complete, and the WorkerReadyMessage logging is now just noise in every build.


cobexer:

PR #38642, opened 2026-07-27
Report problems from threads without a current build operation

### LINKED ISSUE #35885: Problems that are reported from threads without active build opertation are ignored
### Current Behavior

If there is no `OperationIdentifier`, the following code simply discards the problem:

https://github.com/gradle/gradle/blob/1468e41e448ec0c31f6f5b6999825da00e22c931/platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemReporter.java#L128-L131

### Expected Behavior

All problems reported via `problems.reporter.report()` should be added to the problems report.

### Context (optional)

_No response_

**settings.gradle.kts**

```kotlin
import org.gradle.kotlin.dsl.support.serviceOf
import org.gradle.tooling.events.FinishEvent
import org.gradle.tooling.events.OperationCompletionListener

val observer = gradle.sharedServices.registerIfAbsent("buildObserver", BuildObserver::class)

serviceOf<BuildEventsListenerRegistry>().onTaskCompletion(observer)

abstract class BuildObserver : BuildService<BuildServiceParameters.None>, OperationCompletionListener {
    @get:Inject abstract val problems: Problems

    override fun onFinish(event: FinishEvent) {
        val problemGroup = ProblemGroup.create("issues", "issues")
        val id = ProblemId.create("finished",event.displayName, problemGroup)
        problems.reporter.report(id) {}
    }
}
```

### Gradle version

9.3.0 RC1

### Build scan URL (optional)

_No response_

_No response_

reinsch82:
Fixes #35885

### Context

Problems reported from a thread without a current build operation — e.g. an `OperationCompletionListener` build service, dispatched on an executor thread that is never enrolled in the thread-local `CurrentBuildOperationRef` — were silently discarded by `DefaultProblemReporter`. The null-id guard existed to avoid the `IllegalArgumentException` of the progress-event emitter, trading a crash for silent data loss.

### How it works

- **`BuildOperationIdRef`** (new): provides the id of a build operation to attribute events to. Implemented by `CurrentBuildOperationRef`, the holders below, and a `FallbackBuildOperationIdRef` composite (preferred ref, then fallback).
- **`RootBuildOperationRef`** (new, CrossBuildSession scope): captures the root "Run build" operation id from `DefaultBuildTreeActionExecutor` — same pattern `LoggingBuildOperationProgressBroadcaster` uses for logging from unenrolled threads. Overwritten per root-operation start, never cleared, discarded with the per-invocation `CrossBuildSessionState`.
- **`DefaultProblemReporter`** depends on a single composed ref: current operation, falling back to the root operation — which is still running whenever such problems arrive, so all sinks observe them (build operation stream, HTML report, console, TAPI/Develocity).
- **Worker processes** have no root operation; `WorkRequestBuildOperationRef` (new, immutable, created per work request on the request thread) provides the work request's operation as the fallback, so problems from user-spawned threads in process-isolated workers are forwarded to the daemon like any other worker problem. In-process workers are covered by the daemon-side fallback.
- **Discard logging**: if no operation is available at all (near-impossible now), the problem is rendered via `ProblemWriter.simple()` in a warning instead of being lost silently. To reuse the renderer without a dependency cycle, `DefaultProblems`/`DefaultProblemReporter` moved out of `:problems-api` into the new **`:problems-reporting`** module (shared by daemon and worker processes) — a first step toward a thinner, interface-only `:problems-api`, which also loses its slf4j dependency.
- **`ProblemBodyWriter`** now renders the file location behind `StackTraceLocation`s (as the HTML report and TAPI conversion already did), so problems reported without an explicit location show their origin. This adds a `Location:` line to console problem output and build-failure problem blocks for stack-located problems — an intended behavior change.

### Testing

- Reproducer from the issue (`ProblemsServiceWithoutBuildOperationIntegrationTest`): red before, green after, in forking + configuration-cache variants; also asserts the HTML report content.
- `WorkerExecutorProblemsApiIntegrationTest`: problem from a user-spawned thread in all three isolation modes, asserting attribution (work request op for process isolation, root op otherwise).
- Unit tests for the fallback composite, the root holder, the reporter (incl. the rendered discard message and its origin line), and the stack-trace-location rendering; TAPI cross-version test (`r980`) proving an IDE-style client receives the event.
- Render-surface suites green after the `ProblemBodyWriter` change: problems console rendering, plugin validation (incl. failure rendering), java compilation problems, work-validation warnings.
- `sanityCheck` passes; internal-only change.


platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemReporter.java
```diff
@@ -132,16 +136,22 @@ public ProblemInternal internalCreate(Action<? super ProblemSpecInternal> action
...
      */
     @Override
     public void report(Problem problem) {
-        OperationIdentifier id = currentBuildOperationRef.getId();
+        OperationIdentifier id = operationIdRef.getId();
         if (id != null) {
             report(problem, id);
+        } else {
+            ProblemId problemId = ((ProblemInternal) problem).getDefinition().getId();
+            LOGGER.info("Discarding problem '{}:{}': no build operation is available to attribute it to on this thread", problemId.getGroup().getName(), problemId.getName());
```
eskatos: ❓ The problem ID might not be enough for locating its origin making this log statement useless. Is there a way to render more information about the problem? Can we reuse some console rendering infrastructure easily here?

eskatos: ❓ Why `INFO`? I would have expected at least `WARNING`

platforms/core-execution/workers/src/integTest/groovy/org/gradle/workers/internal/WorkerExecutorProblemsApiIntegrationTest.groovy
```diff
@@ -225,6 +226,102 @@ class WorkerExecutorProblemsApiIntegrationTest extends AbstractIntegrationSpec {
...
+
+        when:
+        run("reportProblem")
+
+        then:
+        verifyAll(receivedProblem) {
+            definition.id.fqid == 'generic:spawned'
+            definition.id.displayName == 'problem from spawned thread'
+            if (isolationMode == "'${WorkerExecutorFixture.IsolationMode.PROCESS_ISOLATION.method}'") {
+                // In a worker process the problem is attributed to the work request's build operation
```
eskatos, line 317: 💅 It would be good to assert on the build operation attribution in other isolation modes too

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemReporter.java
reinsch82: Did reuse some, to reuse more we need a larger refactoring of the problems projects.

platforms/ide/problems-api/src/main/java/org/gradle/api/problems/internal/DefaultProblemReporter.java
```diff
@@ -132,17 +139,48 @@ public ProblemInternal internalCreate(Action<? super ProblemSpecInternal> action
...
+            LOGGER.warn(
+                "Discarding problem '{}' ({}){}: no build operation is available to attribute it to on this thread",
+                problemId.getDisplayName(),
+                problemId,
+                describeOrigin(problemInternal)
+            );
+        }
+    }
+
+    private static String describeOrigin(ProblemInternal problem) {
```
eskatos: ❌ This is duplicating a lot of logic for displaying a problem on the console while still not doing it properly. I'd rather we reuse `ProblemWriter.simple()`.

I understand that there is a project dependency complication because `DefaultProblemReporter` lives in `:problems-api` and `ProblemWriter` lives in `:problems-rendering`. But, this is an opportunity to reduce the amount of internals that are currently present in `:problems-api`.

eskatos (changes requested):

reinsch82:
@bot-gradle test this

eskatos (approved):

PR #38643, opened 2026-07-27
Respect third-party agents on CC hit

### LINKED ISSUE #38607: ClassLoader mismatch for BuildService references with Isolated Projects when debugging Gradle daemon
### Current Behavior

When the Gradle daemon runs with a JVMTI/Java agent attached — which any debug session does via `-Dorg.gradle.debug=true` (jdwp `-agentlib:`) — a build that uses **Isolated Projects** and a shared `BuildService` referenced from serialized state fails to load the configuration cache entry.

The same class is loaded twice by two different instances of the *same* classloader scope (`root-project[:](export)`), so setting the service provider onto the consumer's property fails:

```
* What went wrong:
Could not load the value of field `__buildFusServiceProperty__` of `org.jetbrains.kotlin.gradle.plugin.BuildFinishAndConfigurationTimeMetricsFlowAction$Parameters_Decorated` bean found in class `org.jetbrains.kotlin.gradle.plugin.BuildFinishAndConfigurationTimeMetricsFlowAction`.
> Cannot set the value of a property of type org.jetbrains.kotlin.gradle.plugin.statistics.FlowActionBuildFusService loaded with InstrumentingVisitableURLClassLoader(ClassLoaderScopeIdentifier.Id{coreAndPlugins:settings[:]:buildSrc[:]:root-project[:](export)}) using a provider of type org.jetbrains.kotlin.gradle.plugin.statistics.FlowActionBuildFusService loaded with InstrumentingVisitableURLClassLoader(ClassLoaderScopeIdentifier.Id{coreAndPlugins:settings[:]:buildSrc[:]:root-project[:](export)}).
  This can be caused by a plugin being applied to two sibling projects and then using a shared build service. To fix this, use `@ServiceReference` or add the problematic plugin with `apply false` to the root build script.
```

Note both the property type and the provider type report the **identical** classloader scope id — the suggested fixes in the message do not apply.

<details>
<summary>Full stacktrace</summary>

```
org.gradle.api.GradleException: Could not load the value of field `__buildFusServiceProperty__` of `org.jetbrains.kotlin.gradle.plugin.BuildFinishAndConfigurationTimeMetricsFlowAction$Parameters_Decorated` bean found in class `org.jetbrains.kotlin.gradle.plugin.BuildFinishAndConfigurationTimeMetricsFlowAction`.
        at org.gradle.internal.serialize.graph.BeanPropertyExtensionsKt.readPropertyValue(BeanPropertyExtensions.kt:70)
        at org.gradle.internal.serialize.beans.services.BeanPropertyReader.readFieldOf(BeanPropertyReader.kt:95)
        at org.gradle.internal.serialize.beans.services.BeanPropertyReader.readFieldsOf(BeanPropertyReader.kt:75)
        at org.gradle.internal.serialize.codecs.core.RegisteredFlowActionCodec.decode(ProviderCodecs.kt:193)
        at org.gradle.internal.cc.impl.ConfigurationCacheState.readFlowScopeOf(ConfigurationCacheState.kt:608)
        at org.gradle.internal.cc.impl.ConfigurationCacheState.readBuildContent$org_gradle_configuration_cache(ConfigurationCacheState.kt:554)
        at org.gradle.internal.cc.impl.ConfigurationCacheState.readBuildState(ConfigurationCacheState.kt:388)
        at org.gradle.internal.cc.impl.ConfigurationCacheState.readRootBuild(ConfigurationCacheState.kt:296)
        at org.gradle.internal.cc.impl.ConfigurationCacheState.readRootBuildState(ConfigurationCacheState.kt:197)
        at org.gradle.internal.cc.impl.DefaultConfigurationCacheIO.readRootBuildStateFrom(DefaultConfigurationCacheIO.kt:228)
        at org.gradle.internal.cc.impl.DefaultConfigurationCache.loadWorkGraph(DefaultConfigurationCache.kt:771)
        at org.gradle.internal.cc.impl.DefaultConfigurationCache.loadRequestedTasks(DefaultConfigurationCache.kt:295)
        at org.gradle.internal.cc.impl.ConfigurationCacheAwareBuildTreeWorkController.scheduleAndRunRequestedTasks(ConfigurationCacheAwareBuildTreeWorkController.kt:100)
        ...
Caused by: java.lang.IllegalArgumentException: Cannot set the value of a property of type org.jetbrains.kotlin.gradle.plugin.statistics.FlowActionBuildFusService loaded with InstrumentingVisitableURLClassLoader(ClassLoaderScopeIdentifier.Id{coreAndPlugins:settings[:]:buildSrc[:]:root-project[:](export)}) using a provider of type org.jetbrains.kotlin.gradle.plugin.statistics.FlowActionBuildFusService loaded with InstrumentingVisitableURLClassLoader(ClassLoaderScopeIdentifier.Id{coreAndPlugins:settings[:]:buildSrc[:]:root-project[:](export)}).
This can be caused by a plugin being applied to two sibling projects and then using a shared build service. To fix this, use `@ServiceReference` or add the problematic plugin with `apply false` to the root build script.
        at org.gradle.api.services.internal.BuildServiceProvider.asSupplier(BuildServiceProvider.java:141)
        at org.gradle.api.internal.provider.DefaultProperty.set(DefaultProperty.java:118)
        at org.gradle.api.internal.provider.DefaultProperty.provider(DefaultProperty.java:110)
        at org.gradle.internal.serialize.codecs.core.PropertyCodec.decodeThis(ProviderCodecs.kt:420)
        at org.gradle.internal.serialize.codecs.core.AbstractPropertyCodec.decode(ProviderCodecs.kt)
        at org.gradle.internal.serialize.graph.BeanPropertyExtensionsKt.readPropertyValue(BeanPropertyExtensions.kt:61)
        ... 174 more
```

</details>

### Expected Behavior

The build succeeds regardless of whether an agent is attached to the daemon. The shared `BuildService` class should resolve to a single classloader, so its provider and consumers remain type-compatible.

### Context (optional)

This blocks debugging a Gradle daemon (`-Dorg.gradle.debug=true`) for any build that uses Isolated Projects together with a build service registered as a build-event listener — for example the Kotlin Gradle Plugin's FUS service. It reproduces only under Isolated Projects; plain configuration cache is unaffected.

I will add a test. It reproduces with Gradle's own build

### Gradle version

9.7.0-rc-1

### Build scan URL (optional)

N/A

_No response_

### LINKED ISSUE #38619: Transformation applied to build logic classes by a third-party Java agent is lost on configuration cache hit
### Current Behavior

On Configuration Cache hit, the transformation applied by a third-party agent is lost. Only Gradle's own transformation is applied.

### Expected Behavior

All transformations should be applied both on CC hit and CC miss.

### Context (optional)

This can be problematic for plugin authors testing Configuration Cache compatibility with Jacoco enabled.

TBD (test)

### Gradle version

9.7.0-rc-1

### Build scan URL (optional)

_No response_

_No response_

mlopatkin:
When a third-party agent is attached to the build JVM (e.g. while debugging the daemon), we don't use the pre-transformed JARs for classpath, but instead replay the transformation at runtime, during classloading; this way we can cover whatever tranformations the other agents have applied, and they see the original bytecode.

However, we weren't rebuilding the runtime transformation pipeline upon configuration cache hits, falling back to the pre-transformed JARs instead. This caused two issues:
- Third-party agent modifications were lost when the Gradle's agent replaced them with the pre-transformed bytecode.
- ClassLoader reuse was broken, resulting in BuildService references not being restorable properly.

This commit adds the necessary plumbing for CC hit. We store information needed to set up the instrumentation in the classloader spec, and use it to rebuild the runtime transformation path upon recreating the classloader from the CC entry.

Fixes #38619, #38607


mlopatkin:
@bot-gradle test this without PTS, PT

mlopatkin:
@coderabbitai review

mlopatkin:
@bot-gradle test this

eskatos (approved):

PR #38646, opened 2026-07-28
Model dependency management instance behavior with explicit parameters

jvandort:
Previously, a DomainObjectContext acted as a parameters object for each dependency management instance, which derived its behavior from details of the context -- whether it was a script, a root script, or a plugin context. Instead, we now model those behaviors explicitly: each instance is created from a DependencyManagementParameters carrying its display name, dependency locking prefix, whether it runs in a JVM environment, and whether it ignores repositories and component metadata rules from dependency resolution management.

DomainObjectContext is repurposed to model a set of mutable objects sharing a domain, such as a Gradle build or a Gradle project. It no longer exposes per-instance details. Build scope now registers a BuildDomainObjectContext, which non-project dependency management instances inherit, and StandaloneDomainObjectContext is reduced to its anonymous model container.

Since all non-project instances within a build now share the build's context, the context can no longer identify a single instance or describe its owner. Instead, each instance registers its own DependencyManagementInstanceIdentity, created from the parameters' display name. This identity backs the same-context check for extendsFrom and the owner descriptions used in resolution failure messages, configuration container display names, and dependency locking output, preserving the previous per-instance display names.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

PR #38649, opened 2026-07-28
Re-attach owners for un-managed properties on CC restore

### LINKED ISSUE #37421: Private non-abstract properties lose their owner after Configuration Cache roundtrip

abstratt:
Non-abstract Property and ConfigurableFileCollection getters (a.k.a. unmanaged properties) used to only attach their owner lazily in the generated getter override, which never ran when the owner was restored from the configuration cache. This PR ensures owner attachment happens for properties whose getters are never invoked (they are accessed  via direct access to the underlying property).  

Co-authored-by: Develop-KIM <kimdonghwan913@gmail.com> 
Co-authored-by: rchaves@gradle.com

Fixes:  #37421
Supersedes: #38615

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AsmBackedClassGenerator.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AsmBackedClassGenerator.java
@@ -512,6 +512,7 @@ public class AsmBackedClassGenerator extends AbstractClassGenerator {
         private static final String RETURN_VOID_FROM_OBJECT_MODEL_OBJECT = getMethodDescriptor(VOID_TYPE, OBJECT_TYPE, MODEL_OBJECT_TYPE);
         private static final String RETURN_VOID_FROM_DEFAULT_PROPERTY_SERVICE_LOOKUP_STRING = getMethodDescriptor(VOID_TYPE, DEFAULT_PROPERTY_TYPE, SERVICE_LOOKUP_TYPE, STRING_TYPE);
         private static final String RETURN_VOID_FROM_MODEL_OBJECT_DISPLAY_NAME = getMethodDescriptor(VOID_TYPE, MODEL_OBJECT_TYPE, DISPLAY_NAME_TYPE);
+        private static final String RETURN_VOID_FROM_EXCEPTION_MODEL_OBJECT_STRING = getMethodDescriptor(VOID_TYPE, EXCEPTION_TYPE, MODEL_OBJECT_TYPE, STRING_TYPE);
         private static final String RETURN_OBJECT_FROM_TYPE = getMethodDescriptor(OBJECT_TYPE, JAVA_LANG_REFLECT_TYPE);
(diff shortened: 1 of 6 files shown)


abstratt:
Supersedes #38615

abstratt:
@Develop-KIM This is the PR we will try to merge. If you have any feedback about the bits not in your original PR, please do chime in.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AsmBackedClassGenerator.java
```diff
@@ -1240,18 +1241,60 @@ public LocalMethodVisitorScope(MethodVisitor methodVisitor) {
...
+             * Same as {@link #attachProperty(AttachedProperty)}, but tolerates a getter that fails.
+             *
+             * <p>The getter is invoked only to obtain the property so its owner can be re-attached, so a failure
+             * here is not reported: the property is simply left without an owner. A later direct call to the
+             * getter still fails as usual.</p>
+             */
+            protected void attachPropertyIfPossible(AttachedProperty attached) {
+                // GENERATE
+                //     <type> value;
+                //     try { value = get<prop>(); } catch (Exception ignored) { <skip this property> }
```
asodja: ❓ What is consequence of try/catch here? What does that mean in practice?

What is the risk of this (e.g. some side effects)?

abstratt: It means that getters that may not be invocable at some point in time (like properties not meant to be used after serialization, like [this](https://github.com/gradle/gradle/blob/2066d26f38b892c20dff48f52f35bc7cc5b6da78/platforms/software/publish/src/main/java/org/gradle/api/publish/tasks/GenerateModuleMetadata.java#L89)) are tolerated (the result is a missing owner). 

I changed the code to log information from the exception to the debug output.

asodja: Ok, let's hope users don't put some funny code in there :)

asodja (approved):
LGTM

PR #38656, opened 2026-07-29
Store cross-build task references as node ids in the Configuration Cache

jvandort:
Groundwork for eventually replacing the per-build work-graph coordination
(`TaskInAnotherBuild`, `ExportedTaskNode`) with work graphs that span the build tree. 

Node IDs are now assigned from a single ID space spanning all builds in the tree, and a
`TaskInAnotherBuild` reference is stored as the node id of its target instead of
`(taskPath, buildIdentifier)`. These node IDs are assigned up-front and written to the CC at store time. During load, we bind tasks to restored `TaskInAnotherBuild` directly using the global node ID rather than looking up the tasks in the target build's work graph.

It seems a stored `TaskInAnotherBuild` always references a real `TaskNode`. The `TaskInAnotherBuild` is excluded from the execution plan if the target node in the other build has already executed. This means we can always be sure the target task is present in the restored CC work graph when binding a task to a restored `TaskInAnotherBuild`. This allowed us to remove the previous path-based task lookup entirely.

This change makes it so CC no longer cares which build a task is from. This opens the door to the scheduler creating unified work graphs for a single build tree rather than segregating what would be a single work graph into several build-scoped work graphs.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
+++ b/platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
@@ -73,7 +73,9 @@ import org.gradle.internal.file.FileSystemDefaultExcludesProvider
 import org.gradle.internal.flow.services.BuildFlowScope
 import org.gradle.internal.operations.BuildOperationProgressEventEmitter
 import org.gradle.internal.scopeids.id.BuildInvocationScopeId
+import org.gradle.internal.serialize.codecs.core.IdForNode
 import org.gradle.internal.serialize.codecs.core.IsolateContextSource
+import org.gradle.internal.serialize.codecs.core.assignNodeIds
 import org.gradle.internal.serialize.graph.MutableReadContext
 import org.gradle.internal.serialize.graph.ReadContext
 import org.gradle.internal.serialize.graph.WriteContext
@@ -312,7 +314,8 @@ class ConfigurationCacheState(
             }
         }
         val transformedProjectsPerBuild = collectTransformedProjectsPerBuild(scheduledWorkPerBuild)
-        val buildTreeState = StoredBuildTreeState(requiredBuildServicesPerBuild, scheduledWorkPerBuild, transformedProjectsPerBuild)
+        val idForNode = assignNodeIds(scheduledWorkPerBuild.values)
+        val buildTreeState = StoredBuildTreeState(requiredBuildServicesPerBuild, scheduledWorkPerBuild, idForNode, transformedProjectsPerBuild)
         writeCollection(builds.values) { build ->
             writeBuildState(build, buildTreeState)
         }
@@ -524,7 +527,7 @@ class ConfigurationCacheState(
             if (scheduledWork.scheduledNodes.isNotEmpty()) {
                 writeBoolean(true)
                 withDebugFrame({ "Work Graph" }) {
-                    writeWorkGraphOf(gradle, scheduledWork)
+                    writeWorkGraphOf(gradle, scheduledWork, buildTreeState.idForNode)
                 }
             } else {
                 writeBoolean(false)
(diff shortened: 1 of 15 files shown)


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

bamboo:
Have you benchmarked the effect of these changes on `gradle/gradle` cache hits and cache misses?

jvandort:
I have not. Is there some benchmark suite other than our standard perf tests?

jvandort:
@bot-gradle test apt

jvandort:
I can try running gradle profiler locally with `sanityCheck --dry-run` vs master

bamboo:
Our performance tests are only imperfect gates. Whenever we touch highly optimized/benchmark areas like this one we should take on ourselves to do some benchmarking/profiling. I suggest making sure we don't regress on `codeQuality --dry-run` with cache hits first, cache misses second.

jvandort:
Initial performance testing for CC hits and misses on `sanityCheck --dryRun` showed a ~1-2% regression depending on the scenario. I've pushed up a commit which removes the regression and brings this PR back to performance parity with master. PTAL

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/ConfigurationCacheState.kt
```diff
@@ -255,6 +260,31 @@ class ConfigurationCacheState(
...
+    /**
+     * Binds each restored reference to a task in another build to its restored target node.
+     * This can only be done once all builds in the tree have been loaded, as a reference may
+     * point into the work graph of a build that is loaded after the referencing build.
+     */
+    private
+    fun bindRestoredTaskReferences(builds: List<CachedBuildState>) {
+        val buildsWithWork = builds.filterIsInstance<BuildWithWork>()
+        // Node ids form a single dense, contiguous id space spanning the whole build tree, so each build's
+        // nodes can be scattered into one flat array indexed by global node id, giving O(1) target lookups.
```
bamboo: 🤔 `O(1)` lookup cost is nice but at the cost of building this large array when, in fact, every workGraph already carries an id dense array. Why not simply lookup as needed which should be `O(len(buildsWithWork))` right? My hypothesis is that restored task references should be rare enough that there's no point building the big array.

bamboo (changes requested):

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/WorkNodeCodec.kt
```diff
@@ -215,14 +272,20 @@ class WorkNodeCodec(
...
+    fun ReadContext.readNodes(baseNodeId: Int, buildNodeCount: Int): Array<Node> {
+        // Node ids of a work graph form a dense, contiguous range, so build into a temporary null-filled
+        // buffer (batches arrive out of order) and then validate every slot was filled exactly once.
+        val buildNodesById = arrayOfNulls<Node>(buildNodeCount)
+        readNodeBatchesInParallel().forEach { batch ->
+            batch.forEach { (node, globalNodeId) ->
+                buildNodesById[globalNodeId - baseNodeId] = node
+            }
+        }
+        return Array(buildNodeCount) { buildNodeId ->
```
bamboo: ❓ Could this be a `for-loop` validation plus an unsafe cast instead of a copying operation?

bamboo (changes requested):

platforms/core-configuration/core-serialization-codecs/src/main/kotlin/org/gradle/internal/serialize/codecs/core/WorkNodeCodec.kt
```diff
@@ -272,20 +273,17 @@ class WorkNodeCodec(
...
-            }
-        }
-        return Array(buildNodeCount) { buildNodeId ->
-            requireNotNull(buildNodesById[buildNodeId]) {
-                "No node with id ${baseNodeId + buildNodeId} was restored"
+                nullableNodesById[globalNodeId - baseNodeId] = node
             }
         }
+        val buildNodesById: Array<Node> = nullableNodesById.requireNoNulls()
+        return { globalNodeId -> buildNodesById[globalNodeId - baseNodeId] }
```
bamboo, line 286: 💅 I think it would be ok to expose the array directly to the consumer and avoid the extra indirection, up to you.

bamboo (approved):

PR #38658, opened 2026-07-29
Update cross-build task references to operate on task nodes

jvandort:
https://github.com/gradle/gradle/pull/38656 guarantees all cross-build task graph references hold on to a reference to the target `TaskNode`. Now `TaskInAnotherBuild` can operate directly on this `TaskNode` instead of the `IncludedBuildTaskResource` facade.

* Deleted `ExportedTaskNode`. This acted as a cross-build reference to some `TaskNode` and was used as a common reference to some other task node in some other build. Now that other builds can directly reference the original `TaskNode`, this is no longer necessary.
* Deleted `IncludedBuildTaskResource`. Acted as a proxy/facade to the actual `TaskNode`. No longer necessary now that we have an actual reference to the node.
* Deleted `TaskIdentifier`. This facilitated discovering `ExportedTaskNode`s, but is no longer necessary now that we already reference the underlying `TaskNode` we wanted to lookup.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

bamboo:
It would be better to have at least someone from @gradle/bt-cortex approving the PR before the merge...

jvandort:
Agreed. I have a number of changes pending after this one, so we should discuss a procedural change to avoid PRs sitting in review for many weeks.

bamboo:
> I have a number of changes pending after this one

What is this work connected to by the way? What's the big picture?

subprojects/composite-builds/src/main/java/org/gradle/composite/internal/DefaultBuildController.java
```diff
@@ -167,14 +166,15 @@ private static void checkForCyclesFor(TaskInternal task, Set<TaskInternal> visit
...
-    private static void visitDependenciesOf(TaskInternal task, Consumer<TaskInternal> consumer) {
-        TaskNodeFactory taskNodeFactory = ((GradleInternal) task.getProject().getGradle()).getServices().get(TaskNodeFactory.class);
-        TaskNode node = taskNodeFactory.getOrCreateNode(task);
-        for (Node dependency : node.getAllSuccessors()) {
-            if (dependency instanceof TaskNode) {
-                consumer.accept(((TaskNode) dependency).getTask());
+    private static void visitDependenciesOf(Node node, Consumer<? super Node> visitor) {
+        node.getAllSuccessors().forEach(dep -> {
+            if (dep instanceof TaskInAnotherBuild) {
+                // Hop over the included build reference so we can detect cycles across build work graphs.
```
bamboo, line 172: Is this new behavior? Do we have enough test coverage for cross-build cycles?

subprojects/composite-builds/src/main/java/org/gradle/composite/internal/DefaultBuildControllers.java
```diff
@@ -37,7 +36,7 @@
 
 class DefaultBuildControllers implements BuildControllers {
     // Always iterate over the controllers in a fixed order
-    private final Map<BuildIdentifier, BuildController> controllers = new TreeMap<>(idComparator());
```
bamboo, line 40: ❓ Does it still need to be `TreeMap`?

jvandort: The cross-build scheduling fixed-point (`populateWorkGraphs`) iterates this map in order. Changing the ordering of this map would change the order work is discovered and scheduled across builds.

The commit that added `Always iterate over the controllers in a fixed order` and the `TreeMap` does not mention this change in the commit message, so I don't really know the _reason_ we have historically always iterated over the controllers in a fixed order.

I can only imagine changing this to an unordered map will make scheduling work across builds non-deterministic. 

Are you proposing a different ordered map other than `TreeMap`? Or are you proposing that this map may not need to be ordered at all?

subprojects/composite-builds/src/main/java/org/gradle/composite/internal/DefaultBuildController.java
jvandort: No, this is existing behavior. The prior code, for some input `task` attempts to find its dependencies. This method is part of a cycle detection check for a graph of `Task` objects:
```
        TaskNodeFactory taskNodeFactory = ((GradleInternal) task.getProject().getGradle()).getServices().get(TaskNodeFactory.class);
        TaskNode node = taskNodeFactory.getOrCreateNode(task);
        for (Node dependency : node.getAllSuccessors()) {
            if (dependency instanceof TaskNode) {
                consumer.accept(((TaskNode) dependency).getTask());
            }
        }
```

We used the _target_ tasks's `TaskNodeFactory` to get `node`. Since the task and the `TaskNodeFactory` are from the same build, we always get a `LocalTaskNode` back. Then, we get all dependencies, and find all `TaskNode`s (both `LocalTaskNode` and `TaskInAnotherBuild`). 

So, the `dependency instanceof TaskNode` + `getTask` effectively does the same "hop" as my current code makes explicit (nodes in this graph we're walking are `Tasks`, and `getTask` returns tasks from both the same and other builds since `dependency` is either a local or other build task node).

In terms of tests, we have `CompositeBuildDependencyCycleIntegrationTest`, plus one test in `CompositeBuildPluginDevelopmentIntegrationTest`.

jvandort: The whole purpose of this cycle detection algorithm are to detect cycles of work _between_ execution plans in the same build tree work graph.

The "normal" cycle detection algorithm only works between nodes in a single plan: https://github.com/gradle/gradle/blob/0f12f8106bb62cf1b77f081a22c531d3d2563008/subprojects/core/src/main/java/org/gradle/execution/plan/DetermineExecutionPlanAction.java#L295

This code we're talking about is here precisely since we've divided what is a single work graph into a number of execution plans, where the "edges" between these plans of a single work graph are `TaskInAnotherBuild`. We need a separate cycle detection algorithm that treats these cross-plan edges are real edges. Ideally, this code we're commenting on shouldn't exist at all. We should have one plan per build tree work graph.

jvandort:
@bamboo The main motivation behind this series of changes is to simplify the way Gradle builds a work graph when there are multiple builds involved. If some project depends on work defined in a separate build, the `ExecutionPlan` of that project's node does not have an edge to the node it depends on. There are two execution plans, two "build work graphs", all owned by a "build tree work graph" and the edge between the two build work graphs are represented by a `TaskInAnotherBuild`.

This PR stream aims to address the existing structure by consolidating multiple build work graphs into a single build tree work graph without the complexity overhead of juggling and stitching multiple plans together.

jvandort:
@bot-gradle test this

PR #38664, opened 2026-07-29
Add deprecation nag to Configuration.setVisible()

### LINKED ISSUE #34290: Add deprecation nag to Configuration.setVisible()
### Issue type

Deprecation

### Problem description

Add a deprecation nag for `Configuration.setVisisble()` once https://youtrack.jetbrains.com/issue/KT-78754 is resolved.

### Context (optional)

The Kotlin plugin currently calls `Configuration.setVisible()` in a number of places.  We then use the kotlin plugin in various places such as the `kotlin-dsl` and `kotlin-embeddable` plugins.  So, when we add a deprecation nag, we get hundreds of failing tests.  We would end up adding expected deprecations in about a hundred places only to then removed them once https://youtrack.jetbrains.com/issue/KT-78754 was resolved.  Similarly, users would get this deprecation with no meaningful way to resolve it until that issue was resolved.  

Because of this, we've opted not to add this deprecation nag for now, and only add it once a version of Kotlin is available that does not use the deprecated API.

ghale:
The Kotlin plugin previously called Configuration.setVisible() in a number of places (KT-78754), which meant adding a runtime nag would have produced failing tests across kotlin-dsl, kotlin-embeddable and user builds with no meaningful way to resolve them. Now that the embedded Kotlin no longer uses the deprecated API, add the nag, mirroring the existing isVisible() deprecation.  We also expect the deprecation in various places where we test 3rd party plugins that also use this method.

Fixes #34290

### Related Followup PRs

* https://github.com/spotbugs/spotbugs-gradle-plugin/issues/1773
* https://github.com/detekt/detekt/issues/9558
* https://github.com/google/protobuf-gradle-plugin/issues/815
* https://issuetracker.google.com/issues/540783786
* https://github.com/gradle/playframework/issues/218
* https://github.com/tbroyer/gradle-errorprone-plugin/issues/188
* https://github.com/palantir/gradle-consistent-versions/issues/1700
* https://github.com/jenkinsci/gradle-jpi-plugin/issues/415

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/DefaultConfiguration.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/DefaultConfiguration.java
@@ -361,7 +361,7 @@ public abstract class DefaultConfiguration extends AbstractFileCollection implem
     @Deprecated
     public boolean isVisible() {
         DeprecationLogger.deprecateMethod(Configuration.class, "isVisible")
-            .willBeRemovedInGradle10()
+            .willBeRemovedInGradle11()
             .withUpgradeGuideSection(9, "deprecate-visible-property")
             .nagUser();
         return visible;
@@ -371,7 +371,7 @@ public abstract class DefaultConfiguration extends AbstractFileCollection implem
     @Deprecated
     public Configuration setVisible(boolean visible) {
         DeprecationLogger.deprecateMethod(Configuration.class, "setVisible(boolean)")
-            .willBeRemovedInGradle10()
+            .willBeRemovedInGradle11()
             .withUpgradeGuideSection(9, "deprecate-visible-property")
             .nagUser();
         validateMutation(MutationType.BASIC_STATE);
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/DefaultConfiguration.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/DefaultConfiguration.java
@@ -370,8 +370,11 @@ public abstract class DefaultConfiguration extends AbstractFileCollection implem
     @Override
     @Deprecated
     public Configuration setVisible(boolean visible) {
+        DeprecationLogger.deprecateMethod(Configuration.class, "setVisible(boolean)")
+            .willBeRemovedInGradle10()
+            .withUpgradeGuideSection(9, "deprecate-visible-property")
+            .nagUser();
         validateMutation(MutationType.BASIC_STATE);
-        // TODO: Create a deprecation warning once https://youtrack.jetbrains.com/issue/KT-78754 is resolved
         this.visible = visible;
         return this;
     }
--- a/platforms/software/platform-base/src/integTest/groovy/org/gradle/api/plugins/BasePluginIntegrationTest.groovy
+++ b/platforms/software/platform-base/src/integTest/groovy/org/gradle/api/plugins/BasePluginIntegrationTest.groovy
@@ -16,6 +16,7 @@
 package org.gradle.api.plugins
 
 import org.gradle.integtests.fixtures.AbstractIntegrationSpec
+import org.gradle.integtests.fixtures.configuration.ConfigurationAPIDeprecations
 import org.gradle.test.precondition.Requires
 import org.gradle.test.preconditions.FileSystemTestPreconditions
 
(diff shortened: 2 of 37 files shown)


ghale:
@bot-gradle test this

ghale:
@bot-gradle test rFR

platforms/software/platform-base/src/integTest/groovy/org/gradle/api/plugins/BasePluginIntegrationTest.groovy
```diff
@@ -189,6 +190,7 @@ class BasePluginIntegrationTest extends AbstractIntegrationSpec {
         """
 
         expect:
+        ConfigurationAPIDeprecations.expectSetVisibleMethodDeprecation(executer)
```
jvandort, line 193: How come the base plugin is emitting a deprecation?

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/DefaultConfiguration.java
```diff
@@ -370,8 +370,11 @@ public boolean isVisible() {
     @Override
     @Deprecated
     public Configuration setVisible(boolean visible) {
+        DeprecationLogger.deprecateMethod(Configuration.class, "setVisible(boolean)")
+            .willBeRemovedInGradle10()
+            .withUpgradeGuideSection(9, "deprecate-visible-property")
```
jvandort, line 375: Looks like most existing upgrade guide sections use underscores instead of dashes

jvandort: Also looks like we're missing the upgrade guide section in the adoc

jvandort (commented):
Looks like a lot of plugins are affected by this deprecation. 

Would it make sense to push the actual removal to 11 given the blast radius and timeframe>?

platforms/software/platform-base/src/integTest/groovy/org/gradle/api/plugins/BasePluginIntegrationTest.groovy
ghale: The test is "artifacts on legacy configurations are not built by default even if visible is set".  It was introduced when we neutered this method and was intended as a check to guard against regression before we removed the property altogether (i.e. it just confirms that setting visible has no effect).  Now that the method nags, we need an `expectDeprecation()` call.

jvandort: Aw crap, should have read the actual test

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/DefaultConfiguration.java
ghale: > Also looks like we're missing the upgrade guide section in the adoc

No, it's actually there (pre-existing from when we deprecated `getVisible()`).  It was written as a generic "the 'visible' property has been deprecated" section rather than individual methods.  The `setVisible()` method has been deprecated since then, there just wasn't a nag yet.  I double-checked it to be sure and the link it produces looks good.

> Looks like most existing upgrade guide sections use underscores instead of dashes

Of the 9.1 "deprecation" sections in the upgrade guide, only one of them uses underscores, the other six use dashes.  I suspect it's much more inconsistent throughout the rest of the doc, though.  I think what happens is someone adds the first section and then everyone else follows whatever pattern they use for subsequent sections.  I suspect that's what happened here, too.  I guess the question is - does it matter?  I don't mind changing it if it makes a difference, but if not, it doesn't seem worth going back and changing.

ghale:
> Would it make sense to push the actual removal to 11 given the blast radius and timeframe>?

Yeah, I was wondering the same thing and was planning on raising the question during our next meeting.

platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/DefaultConfiguration.java
jvandort: Yeah doesn't matter, we can leave as is, especially if there are plenty of other sections with that format

lkasso (approved):
Docs LGTM. Thanks for updating the BPs.

ghale:
@bot-gradle test this

jvandort (approved):

cobexer:
@bot-gradle test APT

PR #38674, opened 2026-07-30
Relocate test-retry plugin into gradle/gradle

tresat:
Imports the [Gradle test-retry plugin](https://github.com/gradle/test-retry-gradle-plugin) into gradle/gradle as a bundled plugin. The plugin lives at `platforms/software/test-retry/`, ships in `distributions-jvm`, and is applied via the new plugin id **`org.gradle.test-retry-bundled`**. The id is deliberately distinct from the portal-published `org.gradle.test-retry` so builds and tests using the standalone plugin are not silently shadowed by the bundled version.

## What's included

**Prod sources.** All 28 Java classes and the `testRetry.kt` Kotlin DSL extension copied from `gradle/test-retry-gradle-plugin@main:plugin/src/main/`, then modified to fit gradle/gradle conventions: 
- nullability annotations migrated from JSR-305 / JetBrains to jspecify
- `@NullMarked` `package-info.java` files added for six packages
-  four errorprone findings surfaced during migration (missing `@Override`, ambiguous `||`/`&&` precedence, mixed-mutability return, reflection-helper generics) all fixed
- ASM is used unshaded from `libs.asm` because the bundled plugin runs inside Gradle's own classloader — the standalone plugin's shadowJar relocation of `org.objectweb.asm.*` is unnecessary here

**Unit tests.** 

Four unit tests copied from upstream (`TestNamesTest`, `GlobPatternTest`, `RetryFilterTest`, `AnnotationInspectorImplTest`) run under `:test-retry:test`. `AnnotationInspectorImplTest` was rewritten to compile Java via `javax.tools.JavaCompiler` instead of TestKit's `GradleRunner` — this eliminates the need for `gradleTestKit()` on the test classpath.

**Integration tests.** 

All 27 `*FuncTest.groovy` classes from upstream ported to gradle/gradle's `AbstractIntegrationSpec` fixture stack. The upstream files were committed verbatim first as a git-history checkpoint (`fbb4c7fc848`) before rewriting. The migration collapsed the standalone plugin's per-Gradle-version parameterization (`GRADLE_VERSIONS_UNDER_TEST`, `gradleRunner(gradleVersion)`) — a bundled plugin only runs against one Gradle version by definition — and modernized to Spock 2.4-groovy-4.0 with Groovy 4 idioms throughout.

**Sample.** 
- Ported into gradle/gradle's snippets infrastructure at `platforms/documentation/docs/src/snippets/testing/testRetry/` with `common/`, `kotlin/`, and `groovy/` directories per convention. 
- Runs automatically as part of `:docs:docsTest` via the exemplar / `SamplesTestEngine` framework. The upstream `sample-tests/` project was not ported — `:docs:docsTest` uses the same `org.gradle.exemplar:samples-check` library under the hood and completely subsumes it.

## Architecture and cross-cutting checks

The plugin's classes are structurally JVM-testing-specific (they use `Test`, `JvmTestExecutionSpec`, and the `JUnit*/TestNG*` test framework types from `platforms/jvm`) but live in `platforms/software` alongside `testing-base`. This required 25 entries in `platform-coupling.txt`, plus the standard entries for the new subproject: `PackageCycleTest` exemption, `ApiNullabilityTest` freeze entries, two `public-api-methods-return-allowed-types.txt` entries for the Kotlin DSL accessors, four binary-compatibility baseline accepts, and a regenerated `.teamcity/subprojects.json`. `sanityCheck` passes with errorprone fully enabled.

## Known follow-ups

Two upstream plugin bugs surfaced by the migration are captured as `@PendingFeature` rather than papered over:

- **`TestsReader.classFileName` NPE** — TestNG static-initializer failures produce a `null` `className` that the plugin's bytecode reader doesn't guard against. Tracked in `TestNGPlainFuncTest.does not handle flaky static initializers`.
- **`TestExecuter` interception gap** — modern Gradle throws `TestSuiteExecutionException` at a level above the plugin's retry hook when JUnit Platform discovery fails with a `NoClassDefFoundError`. The plugin never gets to retry. Tracked in `JUnit5FuncTest.handles setup failure caused by errors in discovery`.

Both are latent bugs that also affect the portal-published plugin against modern Gradle; the migration surfaced them rather than introducing them.

The upstream `testGradle<N>Releases` cross-version matrix tasks were intentionally dropped — the bundled plugin only exists in Gradle 9.8+, so testing against older Gradle versions is not applicable. Deprecation of the standalone `org.gradle.test-retry` on the Plugin Portal is a separate product/comms decision, out of scope here.


jvandort:
Do we plan to keep bundled in the plugin name forever?

Can we add some kind of alias so if you don't have the external test retry plugin on the classpath you can apply using the non bundled name?

jvandort:
I also wonder if we can maintain git history when merging all this in. We should be able to add the test retry repo as a remote to your local Gradle repo, merge that remote's master branch into Gradle master, have another commit to move the files into the correct location, then rebase all your existing changes on top of those moved files 

This sort of pattern is pretty well documented https://gfscott.com/blog/merge-git-repos-and-keep-commit-history/ and I'm sure Claude can take care of it pretty easily

tresat:
> Do we plan to keep bundled in the plugin name forever?
> 
> Can we add some kind of alias so if you don't have the external test retry plugin on the classpath you can apply using the non bundled name?

This is not this work's final form, and is not meant to be merged.  It is just something helpful to consider when the DV + GBT team plans this plugin's future. 

> I also wonder if we can maintain git history when merging all this in. We should be able to add the test retry repo as a remote to your local Gradle repo, merge that remote's master branch into Gradle master, have another commit to move the files into the correct location, then rebase all your existing changes on top of those moved files

I guess we could, but I don't see much value in that.  I'd expect this entire new plugin subproject to vanish, as functionality is migrated piece-by-piece into the base `Test` class.

jvandort:
Git history is very valuable, otherwise there wouldn't be as much reason to track it ourselves for gradle/gradle

I often find myself looking at git history to determine _why_ code is the way it is, and even for AI based workflows that's translated as an extra data source for agents to do the same. The commit messages and "flow" of code over time can help determine if certain code is the way it is by design or if its current state is incidental rather than intentional. 

If we were to "lump sum" merge all this in at once I would definitely vote to maintain history.

But if as you say we ultimately plan to merge this in piece by piece, more "clean state yet influenced by the original plugin" my opinion would likely differ

PR #38675, opened 2026-07-30
Exempt well-known debug/profile agents from third-party agent handling

mlopatkin:
Attaching a debugger passes -agentlib:jdwp (or -agentpath:.../libjdwp) to the daemon JVM. Detection flagged this conservatively as a third-party agent, which switched buildscript classpath instrumentation onto the load-time "compose" path. As a result, simply attaching a debugger made Gradle developers debug a different code path than the one that runs in production, where no third-party agent is present.

Unlike an arbitrary JVMTI agent, JDWP never rewrites bytecode during normal class loading; it only redefines classes on an explicit debugger command (Hot Code Replace). So it does not need the compose path. Exempt it from detection so a plain debugging session exercises the same substitution-based instrumentation as a regular build.

The trade-off is that the substitution path serves bytecode captured during the artifact transform and ignores the buffer the JVM supplies, so a debugger-initiated redefinition of an instrumented buildscript or plugin class is silently discarded. Gradle cannot veto the debugger's redefinition, so instead of misleading the developer, warn when a hot-swap of an instrumented class has no effect. The classloader now reports whether it honors redefinition (only the compose path does), and the agent's transformer logs a warning otherwise.
### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/instrumentation-agent-services/src/main/java/org/gradle/internal/instrumentation/agent/AgentUtils.java
+++ b/platforms/core-runtime/instrumentation-agent-services/src/main/java/org/gradle/internal/instrumentation/agent/AgentUtils.java
@@ -16,6 +16,8 @@
 
 package org.gradle.internal.instrumentation.agent;
 
+import java.util.function.Predicate;
+
 /**
  * Common utilities used for Gradle's own Java agents.
  */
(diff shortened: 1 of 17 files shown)


mlopatkin:
@coderabbitai review

mlopatkin:
@bot-gradle test this

mlopatkin:
@bot-gradle test PRF, APT

mlopatkin:
@coderabbitai review

platforms/documentation/docs/src/docs/userguide/optimizing-builds/configuration-cache/configuration_cache_requirements.adoc
```diff
@@ -722,6 +722,10 @@ When process isolation is used, the worker action is executed in a separate work
 
 In simple cases, when the libraries also provide command-line entry points (`public static void main()` method), you can also use the link:{javadocPath}/org/gradle/api/tasks/JavaExec.html[`JavaExec`] task to isolate the library.
 
+The agent also affects hot-swapping build logic classes in a running build.
+It prevents Hot Code Replace of such classes: Gradle logs a warning and ignores the recompiled definition.
+You should not rely on Hot Code Replace when debugging build logic; restart the build to pick up such a change.
```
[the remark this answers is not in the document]
mlopatkin: >The statement is too broad. Hot Code Replace has no effect only when Gradle uses the substitution path. A classloader on the runtime-transformation path can reinstrument the supplied redefinition bytecode.

Disabling the substitution path is an internal debugging feature, not a supported mode of operation.

mlopatkin:
PRF failures seem to be unrelated to the PR

lkasso (approved):
Docs LGTM.

platforms/core-runtime/instrumentation-agent-services/src/main/java/org/gradle/internal/instrumentation/agent/AgentUtils.java
```diff
@@ -33,29 +35,101 @@ private AgentUtils() {}
...
-            return !jvmArg.contains(AGENT_MODULE_NAME);
+        // Java Agents usually transform bytecode.
+        if (isJavaAgentSwitchMatching(jvmArg, agentJar -> !isGradleInstrumentationAgent(agentJar))) {
+            // This is `-javaagent:` and not our own agent.
+            return true;
+        }
+
+        // Native JVMTI agents attached via `-agentlib:` or `-agentpath:`
+        // can transform bytecode at class load or redefine classes, the same way a Java agent can.
+        // They run after Gradle's own instrumentation so their transformations are safe (though we don't detect inputs added by them).
```
blindpirate, line 58: I think the ordering here is inverted.

`DefaultDaemonStarter` appends `daemonOpts` — which carry `org.gradle.jvmargs`, so any user `-agentlib:`/`-agentpath:` — at [line 143](https://github.com/gradle/gradle/blob/master/platforms/core-runtime/client-services/src/main/java/org/gradle/launcher/daemon/client/DefaultDaemonStarter.java#L143), and Gradle's `-javaagent:gradle-instrumentation-agent…jar` last, at line 165. HotSpot creates JVMTI environments in command-line order and chains `ClassFileLoadHook` in that same order, so a native agent supplied via `org.gradle.jvmargs` runs **before** Gradle's transformer, not after. That's precisely why the compose path exists.

The exemption may still be defensible for JDWP on other grounds — redefinition is the real risk there, and the new warning in `DefaultClassFileTransformer` covers it — but this sentence shouldn't be the justification for it.

platforms/core-runtime/instrumentation-agent-services/src/main/java/org/gradle/internal/instrumentation/agent/AgentUtils.java
```diff
@@ -33,29 +35,101 @@ private AgentUtils() {}
...
+        // Without runtime transformation capability, Gradle's agent will revert the changes back to the pre-instrumented original versions.
+        // We cannot figure out if the native agent redefines classes, so we err on the side of caution.
+
+        // That said, some JVMTI agents are used for debugging and profiling work, and we know they only redefine classes in some uncommon workflows.
+        // Treating them as third-party agents would force buildscript instrumentation onto runtime transformation capability,
+        // so a debugging or profiling session would exercise a different code path than a regular build does in production.
+        // Exempt the ones we recognize so those sessions run the same instrumentation as production.
+        // So far, we exclude:
+        //  - JDWP (java debugger): it may redefine classes when attempting Hot Code Replace, but we emit a warning there, and it is not used often for Gradle work.
+        //  - AsyncProfiler: it can only transform classes, but never redefines them (as of 4.5).
```
blindpirate, line 69: "it can only transform classes, but never redefines them" — transformation is the capability that makes this exemption unsafe, not the one that makes it safe.

Concrete case:

```
org.gradle.jvmargs=-agentpath:/opt/async-profiler/lib/libasyncProfiler.so=start,event=com.example.MyTask.run
```

async-profiler's `instrument` engine enables `ClassFileLoadHook` and rewrites the target method's class. Before this PR that was detected as third-party, so the classpath went on the compose path and Gradle re-instrumented on top of the profiler's output. With the exemption, `InstrumentingVisitableURLClassLoader.instrumentClass` takes the substitution branch and returns the artifact-transform bytes wholesale, **discarding** the profiler's rewrite of buildscript and plugin classes. Method-level profiling of build logic then silently yields no samples — and the new "had no effect" warning doesn't fire, since it only triggers on `redefineClasses`.

Since the PR title and description only claim JDWP, the simplest fix is to drop the async-profiler entries here and in `isExemptAgentShortName` / `isExemptAgentLibraryFileName`, and ship the JDWP exemption on its own.

mlopatkin: >so the classpath went on the compose path and Gradle re-instrumented on top of the profiler's output.

The [spec](https://docs.oracle.com/en/java/javase/17/docs/api/java.instrument/java/lang/instrument/ClassFileTransformer.html) states:
>Transformations are applied in the following order:
>1. Retransformation incapable transformers  _(ML: this is where Gradle transformer runs)_
>2. Retransformation incapable native transformers 
>3. Retransformation capable transformers 
>4. Retransformation capable native transformers _(ML: this is where async profiler runs)_
>
>For retransformations, the retransformation incapable transformers are not called, instead the result of the previous transformation is reused. In all other cases, this method is called. **Within each of these groupings, transformers are called in the order registered**. Native transformers are provided by the ClassFileLoadHook event in the Java Virtual Machine Tool Interface).



So, according to the spec, async profiler always runs *after* Gradle's agent and the latter cannot override its processing, regardless of path chosen.

platforms/core-runtime/instrumentation-agent-services/src/main/java/org/gradle/internal/instrumentation/agent/AgentUtils.java
mlopatkin: >a native agent supplied via org.gradle.jvmargs runs before Gradle's transformer

As I stated in my other comment, the native agents run after Gradle's transformer per the [spec](https://docs.oracle.com/en/java/javase/17/docs/api/java.instrument/java/lang/instrument/ClassFileTransformer.html#:~:text=Transformations%20are%20applied%20in%20the%20following%20order).

Moreover, Gradle only registers its transform during daemon process startup, unlike other agents that do that it in their `premain` (where command-line order matters).

asodja (approved):
LGTM

blindpirate:
It seems like a real performance regression with this PR...

mlopatkin:
@bot-gradle test APT

mlopatkin:
> It seems like a real performance regression with this PR...

It isn't this PR. Most likely, the regression came with #38664:
* [There is no difference when comparing this to its merge-base](https://builds.gradle.org/buildConfiguration/Gradle_Master_Util_Performance_AdHocPerformanceScenarioLinuxAmd64/116181489?buildTab=tests&expandedTest=build%3A%28id%3A116181489%29%2Cid%3A2000000003&showLog=116181489_4688_99.4659.4676.4683&logFilter=debug&logView=flowAware#%2Fresults%2Fperformance%2Fbuild%2Ftest-results-nowInAndroidBuildPerformanceAdHocTest.zip;%2Fresults%2Fperformance%2Fbuild%2Ftest-results-nowInAndroidBuildPerformanceAdHocTest.zip!%2Fperformance-tests%2Freport;%2Fresults%2Fperformance%2Fbuild%2Ftest-results-nowInAndroidBuildPerformanceAdHocTest.zip)
* [Master regresses slightly compared to the current baseline](https://builds.gradle.org/buildConfiguration/Gradle_Master_Util_Performance_AdHocPerformanceScenarioLinuxAmd64/116181344?buildTab=tests&expandedTest=build%3A%28id%3A116181344%29%2Cid%3A2000000003&showLog=116181344_2753_136.2721.2739.2747&logFilter=debug&logView=flowAware)
* Nagging is visible in the flamegraph:
image

PR #29686, opened 2024-06-25
Use aws-sdk-java-v2.

pkania:
Fixes #27774

### Context

This improves security [CVE-2024-21634](https://nvd.nist.gov/vuln/detail/CVE-2024-21634) and exposes more S3 functionality such as S3 Object Lambda.

### Contributor Checklist
- [✅  ] [Review Contribution Guidelines](https://github.com/gradle/gradle/blob/master/CONTRIBUTING.md).
- [ ✅ ] Make sure that all commits are [signed off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff) to indicate that you agree to the terms of [Developer Certificate of Origin](https://developercertificate.org/).
- [✅  ] Make sure all contributed code can be distributed under the terms of the [Apache License 2.0](https://github.com/gradle/gradle/blob/master/LICENSE), e.g. the code was written by yourself or the original code is licensed under [a license compatible to Apache License 2.0](https://apache.org/legal/resolved.html).
- [ ] Check ["Allow edit from maintainers" option](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) in pull request so that additional changes can be pushed by Gradle team.
- [ ] Provide integration tests (under `<subproject>/src/integTest`) to verify changes from a user perspective.
- [ ] Provide unit tests (under `<subproject>/src/test`) to verify logic.
- [ ] Update User Guide, DSL Reference, and Javadoc for public-facing changes.
- [✅  ] Ensure that tests pass sanity check: `./gradlew sanityCheck`.
- [ ✅ ] Ensure that tests pass locally: `./gradlew <changed-subproject>:quickTest`.

### Reviewing cheatsheet

Before merging the PR, comments starting with 
- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

PR #38687, opened 2026-07-30
Use aws-sdk-java-v2 latest

tresat:
Continues https://github.com/gradle/gradle/pull/29686, updating to the latest AWS version and, fixing the build, adding tests and adding numerous other small fixes and improvements.


platforms/software/resources-s3/src/main/java/org/gradle/internal/resource/transport/aws/s3/S3Client.java
```diff
@@ -16,106 +16,186 @@
...
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 
-@SuppressWarnings("deprecation")
 public class S3Client {
     private static final Logger LOGGER = LoggerFactory.getLogger(S3Client.class);
+    private static final Map<String, Region> KNOWN_BUCKET_REGIONS = new HashMap<>();
```
blindpirate, line 68: Why static? Also, there's potential concurrency issues?

platforms/software/resources-s3/src/main/java/org/gradle/internal/resource/transport/aws/s3/S3Client.java
```diff
@@ -127,132 +207,238 @@ public void put(InputStream inputStream, Long contentLength, URI destination) {
...
+        return listBuilder.build();
+    }
+
     private void configureClient(S3RegionalResource s3RegionalResource) {
-        Optional<URI> endpoint = s3ConnectionProperties.getEndpoint();
-        if (endpoint.isPresent()) {
-            amazonS3Client.setEndpoint(endpoint.get().toString());
+        // Reset before recomputing so a stale region or endpoint from a previous call
+        // doesn't leak into requests for a different bucket (e.g. cross-region uploads).
+        this.endpoint = null;
```
blindpirate, line 381: The mutable instance states don't look good. I'm not sure about concurrency access, but better to make this function return a value immutable object and feed it to `build(config)` method.

platforms/software/resources-s3/src/main/java/org/gradle/internal/resource/transport/aws/s3/S3Client.java
```diff
@@ -127,132 +207,238 @@ public void put(InputStream inputStream, Long contentLength, URI destination) {
...
+        }
+        return location.orElse(originalUri);
+    }
+
+    // A HEAD 404 has no response body, so SDK v2 leaves errorCode null; a GET 404 with an
+    // XML body populates it (typically "NoSuchKey"). Only fall back to a bare-404 check when
+    // the errorCode is absent, so unrelated 404s (NoSuchBucket, endpoint proxies, etc.) still
+    // bubble up as errors instead of being silently swallowed as "resource missing".
+    private static boolean isNoSuchKey(S3Exception e) {
+        String errorCode = e.awsErrorDetails().errorCode();
```
blindpirate, line 420: > awsErrorDetails() can be null — isNoSuchKey, locationFromRedirectException, regionFromRedirectException

> All three dereference it unguarded. SDK v2 leaves it null for exceptions built without service detail (connection-level failures surfacing as S3Exception, some proxy responses). Your own test at S3ClientTest builds S3Exception.builder().statusCode(404).awsErrorDetails(...) — always populated, so the tests can't catch this. An NPE here converts a clean "resource not found" into a build crash.

platforms/software/resources-s3/src/main/java/org/gradle/internal/resource/transport/aws/s3/S3ResourceConnector.java
```diff
@@ -53,52 +50,26 @@ public List<String> list(ExternalResourceName parent) {
...
-            }
-        } catch (IOException e) {
-            // Don't complain loudly to the user about the error there because we were discarding the response anyway.
-            LOGGER.debug("Exception while consuming empty object content from metadata request", e);
-        } finally {
-            // This also closes objectContent, no need to close it explicitly.
-            IoActions.closeQuietly(s3Object);
-        }
+        return new DefaultExternalResourceMetaData(location.getUri(),
+            headObjectResponse.lastModified().toEpochMilli(),
```
blindpirate, line 68: Check `headObjectResponse.lastModified()` null here

platforms/software/resources-s3/src/main/java/org/gradle/internal/resource/transport/aws/s3/S3Client.java
```diff
@@ -127,132 +207,238 @@ public void put(InputStream inputStream, Long contentLength, URI destination) {
...
-        ImmutableList.Builder<String> builder = ImmutableList.builder();
-        builder.addAll(resourceResolver.resolveResourceNames(objectListing));
-
-        while (objectListing.isTruncated()) {
-            objectListing = amazonS3Client.listNextBatchOfObjects(objectListing);
-            builder.addAll(resourceResolver.resolveResourceNames(objectListing));
+        try (software.amazon.awssdk.services.s3.S3Client amazonS3Client = build()) {
+            return amazonS3Client.headObject(objectRequest -> objectRequest.bucket(bucketName).key(s3BucketKey));
+        } catch (S3Exception e) {
+            if (exceptionIsRedirect(e)) {
```
blindpirate, line 319: The endpoint might be any S3-compatible server, so avoid infinite redirection exhaust our stack.

blindpirate (commented):
Some discoveries by Claude which look reasonable:

1. A new HTTP client and connection pool per operation. Every getMetaData, getResource, listDirectChildren, and put calls build(), which constructs a fresh ApacheHttpClient — new pool, new threads, new TLS handshakes — then closes it. v1 held one client for the connector's lifetime. Resolving a module with several hundred artifacts now stands up and tears down several hundred pools, each paying full TLS setup with no connection reuse.
 
2. Full-body buffering on upload. putSingleObject does readExactly(inputStream, Math.toIntExact(contentLength)). The single-object path is taken when contentLength < getMultipartThreshold(), and that threshold is partSize * 2 = 100 MiB by default (S3ConnectionProperties:114, DEFAULT_PART_SIZE = 50 * 1024 * 1024). So publishing a 99 MiB artifact now allocates a 99 MiB byte[]. Multipart buffers partSize = 50 MiB per part. v1 streamed both.
The comment explains why buffering was needed (v2 re-reads for signing/retries with non-markable streams) and that reasoning is sound — but RequestBody.fromContentProvider with a re-openable provider, or spooling to a temp file, gets the same re-readability without the heap. Worth fixing before this ships to users publishing large artifacts.

3. s3control is a test dependency for no reason. build.gradle.kts adds testImplementation(libs.awsS3Control) solely so S3ClientTest can write:

assert putObjectRequest.acl == S3CannedAccessControlList.BUCKET_OWNER_FULL_CONTROL.toString()

That compares an ObjectCannedACL (from s3) against a string constant borrowed from a different service module. Use ObjectCannedACL.BUCKET_OWNER_FULL_CONTROL directly and drop the dependency — it's a whole extra AWS artifact pulled in to spell one enum.

tresat:
@blindpirate - Thanks for the review, I think most of those issues are very relevant.

This PR is definitely still incomplete and needs to be Team Triaged for priority again.  The most important thing here was to restart the work on the linked PR, which was very out of date.  It's a least working with the latest AWS version now.

PR #38717, opened 2026-08-03
Confine Kotlin script compiler stdout redirect to its own thread

### LINKED ISSUE #38702: Build-script println() is silently discarded during Isolated Projects parallel project configuration
### Current Behavior

With Isolated Projects enabled, a plain `println(...)` (i.e. anything written to `System.out`/`System.err`) from a **subproject's** build script at configuration time is **nondeterministically discarded** — roughly 50% of the time in my measurements.

The output is not merely misrouted: it appears neither on the client console nor in the daemon log. It is silently dropped.

Output written through the logging API (`logger.quiet(...)`, `logger.warn(...)`) from the exact same place is always delivered, and root/settings-script `println` (which runs on the main thread) always survives.

### Expected Behavior

`println` from a subproject's configuration should be delivered to the build's console output, as it is without Isolated Projects.

### Context / impact

This is how the CI flakiness of `ProjectSchemaAccessorsIntegrationTest.can access configurations registered by declared plugins via jit accessor` manifests — that test prints a resolved classpath at configuration time in a subproject and asserts on the build output. Failure rates on `gradle/gradle` CI:

| Build type | FAIL | PASS |
|---|---|---|
| `Check_IsolatedProjects_*_bucket8` (master) | 152 | 45 |
| `Check_IsolatedProjects_*_bucket8` (release) | 28 | 11 |
| `Check_ConfigCache_*_bucket9` | 1 | 201 |
| `Check_Quick_1_bucket6` | 2 | 302 |
| `Check_Parallel_*` / `Check_NoDaemon_*` | 0 | 39 / 39 |

~76% under Isolated Projects vs ~0.5% elsewhere, going back to at least 2026-06-18. But the user-facing impact is broader than the test: any configuration-time `println` can vanish under Isolated Projects.

### Steps to Reproduce

Three subprojects, `println` in a subproject build script, run `help` with Isolated Projects and a fresh Gradle user home each attempt:

`settings.gradle.kts`
```kotlin
rootProject.name = "minrepro"
include("a", "b", "c")
```

`a/build.gradle.kts`
```kotlin
plugins { `java-library` }
println("MARKER-A-STDOUT")        // nondeterministically LOST
logger.quiet("MARKER-A-LOGGER")   // always delivered
```

`b/build.gradle.kts`, `c/build.gradle.kts` — same, with their own markers.

```bash
for i in 1 2 3 4 5 6; do
  rm -rf .gradle
  gradle help --console=plain -g /tmp/gh$i \
    -Dorg.gradle.isolated-projects=true \
    -Dorg.gradle.configuration-cache.max-problems=0 \
  | grep -c MARKER-A-STDOUT
done
```

Observed (master `9.8.0-*`, also reproduces on `9.7.0-rc-1`):

```
attempt 1: [MARKER-A-LOGGER-QUIET MARKER-A-LOGGER-WARN MARKER-A-STDOUT]
attempt 2: [MARKER-A-LOGGER-QUIET MARKER-A-LOGGER-WARN]              <- stdout lost
attempt 3: [MARKER-A-LOGGER-QUIET MARKER-A-LOGGER-WARN MARKER-A-STDOUT]
attempt 4: [MARKER-A-LOGGER-QUIET MARKER-A-LOGGER-WARN]              <- stdout lost
attempt 5: [MARKER-A-LOGGER-QUIET MARKER-A-LOGGER-WARN MARKER-A-STDOUT]
attempt 6: [MARKER-A-LOGGER-QUIET MARKER-A-LOGGER-WARN]              <- stdout lost
```

Adding `--max-workers=1` (which forces sequential project configuration) makes it pass every time.

I confirmed the subproject's script really does execute and resolve correctly in the failing runs by having it also append to a file — the file gets all the expected content while stdout shows nothing. So only the output is lost.

### Analysis

Under Isolated Projects, projects are configured in parallel: `DefaultProjectsPreparer.configureHierarchyInParallel` → `TaskPathProjectEvaluator.scheduleProjectsJustInTime` / `scheduleProjectsAheadOfTime`, each project's configuration running inside a `RunnableBuildOperation` on a "Build operations Thread N" worker.

`DefaultScriptRunnerFactory` wraps each script run in `getStandardOutputCapture().start()` / `.stop()` (`subprojects/core/src/main/java/org/gradle/groovy/scripts/internal/DefaultScriptRunnerFactory.java:89,95`). For a project script that resolves to the **per-project** `DefaultLoggingManager` (`ProjectScopeServices.java:174-177` creates one per project). All of those per-project managers mutate the **same process-global** `PrintStreamLoggingSystem` singletons (`LoggingServiceRegistry.java:153-166`).

`PrintStreamLoggingSystem` (`platforms/core-runtime/logging/src/main/java/org/gradle/internal/logging/source/PrintStreamLoggingSystem.java`) has no synchronization at all: `original`, `enabled` and `logLevel` are plain non-volatile fields, and `snapshot()`/`restore()`/`startCapture()`/`install()`/`uninstall()` are unguarded. `DefaultLoggingManager.start()`/`stop()` are likewise unsynchronized, and each `StartableLoggingSystem` keeps its **own** `originalState` snapshot of that shared state (`DefaultLoggingManager.java:337-345, 380-390`) — last-writer-wins, with no nesting count.

Two concrete defects follow:

1. **`install()` can snapshot the capture stream as its own "original"** (`PrintStreamLoggingSystem.java:121-133`):
   ```java
   if (original == null) {
       PrintStream originalStream = get();      // no guard against == outstr
       original = new PrintStreamDestination(originalStream);
   }
   ```
   If a concurrent `uninstall()` has already set `original = null` while `System.out` is still `outstr`, then `get()` returns `outstr` and `original` becomes `PrintStreamDestination(outstr)`. A later `uninstall()` then does `System.setOut(outstr)` with `destination = PrintStreamDestination(outstr)` — a self-referential sink, so everything written to `System.out` loops back into the same buffering stream and is discarded. That matches the observation that the lost output is in neither the console nor the daemon log. Note line 130 already guards `if (get() != outstr)` for the stream itself; lines 122-125 have no equivalent guard.

2. **Sibling scopes tear down each other's capture.** Project A snapshots (`enabled == false`), installs capture; project B (parallel) snapshots (`enabled == true`), so its `startCapture()` is a no-op; project A finishes and `restore()`s its `enabled == false` snapshot → `uninstall()` while B is still configuring and printing. There is no reference counting, so "last restore wins" disables capture process-wide.

The same per-scope `start()`/`stop()` pattern is used for **parallel task execution** (`TaskExecution.java:224,235` with a per-task logging manager from `AbstractTask.java:593-601`), so the hazard is not confined to project configuration.

There is no existing concurrency coverage: `DefaultLoggingManagerTest.groovy` and `PrintStreamLoggingSystemTest.groovy` are single-threaded, and build-script stdout capture is only integ-tested sequentially.

### Suggested fix

In `PrintStreamLoggingSystem`: guard `snapshot`/`restore`/`setLevel`/`startCapture`/`install`/`uninstall` with a monitor, never snapshot `outstr` as `original`, and make capture **reference-counted** so one scope's `stop()` cannot uninstall while sibling scopes still have capture active.

### Your Environment

Reproduced on Gradle `9.8.0-20260730160000+0000` (master) and `9.7.0-rc-1`, macOS (darwin 25.5.0), JDK 17.

blindpirate:
Fixes #38702.

## Problem

`redirecting()` in `KotlinCompiler.kt` swaps the **process-global** `System.out`/`System.err` for a per-compilation `PrintStream` while it compiles a Kotlin script, and `ignoringOutputOf` points that stream at `NullOutputStream.INSTANCE`.

Under Isolated Projects, projects are configured — and their scripts compiled — in parallel. While thread T1 holds the redirect, thread T2 running another project's build script or a task action writes to `System.out` and lands in T1's null sink. The output is **destroyed, not misrouted**: it reaches neither the client console nor the daemon log, while the build still reports `BUILD SUCCESSFUL`.

## Fix

Route the redirect through a thread-confined `OutputStream`: writes from the thread that installed the redirect go to the redirect target, writes from any other thread go to the stream that was current before. One call-site change plus a small private class.

## Evidence

Diagnosis (recorded on #38702):

- Probing `System.out` identity at print time, 16 observations, perfect correlation: `System.out == LinePerThreadBufferingOutputStream` → delivered; `System.out ==` a plain `PrintStream` → lost.
- Kotlin DSL lost output in 6/8 runs vs Groovy DSL 0/8, back to back on the same reproducer — the redirect only exists on the Kotlin script compilation path.
- `logger.quiet` is always delivered (the logging API never touches `System.out`), and `--max-workers=1` always delivers.

Validation of this change:

- Interleaved A/B of two distributions differing only in this commit: **baseline 9/14 rounds lost stdout, patched 0/14**.
- Kotlin compile-error reporting still routes through the collector (checked separately).

CI corroboration from Develocity — `WorkingWithFilesIntegrationTest` over 30 days, 2685 runs, 274 flaky container executions (~10%). Flakiness lands **only** on iterations that compile a Kotlin script not already in the script cache:

| Iteration (per scenario, in `where:` order) | flaky |
|---|---|
| `(set defaults: false) (set values: true)` | 89–100 |
| `(set defaults: true) (set values: false)` | 103–123 |
| `(set defaults: true) (set values: true)` | **0** |

The third row's settings file is identical to row 2's and its project file identical to row 1's, so both scripts are already compiled by the time it runs — no compilation, no redirect window, no lost output. Every flaky iteration is a Kotlin-DSL one; no Declarative-DSL iteration has ever flaked. Other victims in the same `Check_IsolatedProjects_*` buckets: `ProjectSchemaAccessorsIntegrationTest` (~76% in bucket8), `PrecompiledScriptPluginAccessorsTest`, `NestedInputKotlinImplementationTrackingIntegrationTest`.

## Notes

- This is the **root-cause** fix. #38712 (`Make process-global stdout/stderr capture thread safe`) hardens `PrintStreamLoggingSystem`, which is independently correct but is *not* what loses the output — `LineBufferingOutputStream.flush()` drains its buffer before invoking the handler, so the arrangement suspected there would recurse into `StackOverflowError` rather than discard silently. The two changes are complementary and independent.
- `GeneratePrecompiledScriptPluginAccessors.kt` (~lines 684, 689) performs the same global `System.setOut` swap. Same hazard, separate site, not addressed here.
- No test-side change is needed for the affected integration tests — they were correct; the product was dropping their output.

## Still to do before review

- [ ] Add regression coverage asserting a build-script `println` survives parallel configuration under Isolated Projects.
- [ ] Full green CI run.

### DIFF
--- a/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
+++ b/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
@@ -275,11 +275,42 @@ inline fun <T> redirecting(
     to: OutputStream,
     action: () -> T
 ): T = try {
-    set(PrintStream(to, true))
+    set(PrintStream(ThreadConfinedOutputStream(Thread.currentThread(), to, stream), true))
     action()
 } finally {
     set(stream)
     to.flush()
+}
+
+
+/**
+ * Routes writes made by [owner] to [redirected], and writes made by any other thread to [passthrough].
+ *
+ * [redirecting] mutates the process-global `System.out`/`System.err`. Under Isolated Projects, projects are
+ * configured — and their scripts compiled — concurrently, so another thread may be running a build script that
+ * writes to `System.out` while the redirect is installed. Redirecting unconditionally destroys that output:
+ * [ignoringOutputOf] points the target at a [NullOutputStream], so an unrelated `println` disappears without
+ * reaching either the console or the daemon log. Confining the redirect to the installing thread keeps the
+ * compiler's own output captured while leaving every other thread writing to the stream that was current before.
+ */
+private
+class ThreadConfinedOutputStream(
+    private val owner: Thread,
+    private val redirected: OutputStream,
+    private val passthrough: OutputStream
+) : OutputStream() {
+
+    private
+    fun target(): OutputStream =
+        if (Thread.currentThread() === owner) redirected else passthrough
+
+    override fun write(b: Int) = target().write(b)
+
+    override fun write(b: ByteArray) = target().write(b)
+
+    override fun write(b: ByteArray, off: Int, len: Int) = target().write(b, off, len)
+
+    override fun flush() = target().flush()
 }
 
 


blindpirate:
Independent verification of this change, arrived at from the `ProjectSchemaAccessorsIntegrationTest` side. Same conclusion, different evidence — plus one hypothesis ruled out and a measurement caveat.

## `ProjectSchemaAccessorsIntegrationTest` is fixed by this change alone

Tree = `master` + this PR's diff and nothing else. `:kotlin-dsl-integ-tests:isolatedProjectsIntegTest --tests "…ProjectSchemaAccessorsIntegrationTest" --rerun-tasks`, three consecutive runs:

```
run 1  BUILD SUCCESSFUL   27 tests, 0 failed
run 2  BUILD SUCCESSFUL   27 tests, 0 failed
run 3  BUILD SUCCESSFUL   27 tests, 0 failed
```

Baseline on the same machine reproduced the failure first: `can access configurations registered by declared plugins via jit accessor`, `AssertionError at ProjectSchemaAccessorsIntegrationTest.kt:698`, with the build reporting success and only the `help` task text in `result.output`.

## Corroborating evidence for the diagnosis

Two observations that independently point at the Kotlin compiler redirect rather than the logging system:

- **The capture system is never torn down during a losing run.** Instrumenting `PrintStreamLoggingSystem` to dump every state transition shows `count=1 installed=true level=QUIET` for the whole build, and — decisively — **no `text()` call for the lost marker at all**. The output never reaches the capture stream, so nothing downstream of it can be responsible.
- **`System.out` identity at print time correlates perfectly with delivery.** When a subproject's `println` lands, `System.out` is either `LinePerThreadBufferingOutputStream` (delivered) or a bare `java.io.PrintStream` with a fresh identity each run (lost). The bare stream is the `PrintStream(NullOutputStream.INSTANCE)` installed by `redirecting()`.

The Groovy-vs-Kotlin control reproduces here too: converting the same three-subproject reproducer to the Groovy DSL removes the loss entirely.

## Ruled out: missing auto-flush in `LinePerThreadBufferingOutputStream`

Worth recording so nobody spends time on it. `LinePerThreadBufferingOutputStream.getStream()` builds its per-thread stream as `new PrintStream(new LineBufferingOutputStream(...))` with **no auto-flush**, and `flush()` only ever flushes the *calling* thread's stream — which looks like it should strand worker-thread output. It does not.

Interleaved A/B, two distributions differing only by adding `, true` to that constructor, alternating rounds against fresh Gradle user homes:

| Arm | Complete | Lost |
|---|---|---|
| this PR only | 10/14 | 4/14 |
| this PR + auto-flush | 10/14 | 4/14 |

No effect. (An earlier non-interleaved comparison of mine appeared to show a large effect; it was confounded by load and by a stale distribution directory, see below.)

## Caveat on measuring the residual

Both arms above lost output in 4/14 rounds, but that is an artifact of the harness rather than a gap in this fix: the interleaved runs alternate two distributions, so two daemons compete. Running the same arm on its own immediately afterwards gave **8/8 clean**. Loss rate is very sensitive to machine load, so a `0/N` result is worth annotating with how loaded the box was — otherwise it is hard to compare across sessions.

Two methodology notes that cost me real time here, in case they help with the remaining checklist items:

- `:distributions-full:binDistributionZip` does **not** refresh `packaging/distributions-full/build/bin distribution/`, which is the directory the integration tests execute. `:distributions-full:binInstallation` does. Measuring against the stale directory silently reports pre-change behaviour.
- Build the A/B arms by copying the whole installed distribution aside per arm, rather than rebuilding between rounds.

## On #38712

I opened #38712 earlier in this investigation under the reference-counting theory. That theory is refuted by the trace above — capture is never torn down — so #38712 does not fix anything here. It also carries a regression risk this PR does not: its `install()` gains an `if (installed) return;` early exit, which removes the old code's incidental self-healing `if (get() != outstr) set(outstr)` on every install. I'd suggest closing it rather than treating it as complementary; the synchronization argument can be made separately on its own merits if wanted.

jbartok (approved):

cobexer (approved):

jbartok:
@bot-gradle test this

Issue #2754, opened 2017-08-18
:(


rabulhusn:
### Your Environment

 * Build scan URL:

PR #38718, opened 2026-08-03
Handle plain Enum values in DesugaringAttributeContainerSerializer

blindpirate:
### Context

`Gradle_Master_Check_ForceRealizeDependencyManagement_15_bucket3` has been red since build #2754 (2026-07-30) on:

```
EnumAttributeIntegrationTest > component metadata rule adds a plain Enum attribute to a resolved module
  ClassCastException: class MyEnum cannot be cast to class org.gradle.api.Named
```

`Attribute.of` only *supports* `String`, `Boolean`, a `Number` subtype or a `Named` subtype. Since #38409 an unsupported value type is **deprecated rather than rejected**, so unsupported values now survive to places that previously could not see them. `DesugaringAttributeContainerSerializer.write` still assumed its fallback branch was reachable only by `Named` values and cast blindly — the cast itself predates #38409, where it was an `assert` plus a cast.

This is not only a test-harness artifact. A plain enum reaches module metadata via a component metadata rule, and the same serializer backs the `@CacheableRule` result cache, so a rule that sets a plain-enum attribute fails for real users once that cache is written.

The bisect points at f167756dea1 (#38409), which is the commit that *adds* the failing test. It is only red under `-Dorg.gradle.integtest.force.realize.metadata=true` (`DefaultComponentMetadataProcessor.maybeForceRealization` → `forceSerialization`), which is why PR CI and the normal `Check` pipeline stayed green.

Review of the first fix surfaced a second, independent round-trip defect, fixed here as well: an enum that *also* implements `Named` could never be read back. The write path desugars any `Named` value with `getName()`, but `CoercingStringValueSnapshot.coerce` checks `type.isEnum()` **before** its `Named` branch, so read resolved the value with `Enum.valueOf` and threw `No enum constant …` whenever `getName()` differs from `name()`. This predates #38409 — the `Named` write branch is original — and is what a full round-trip test exposes.

### Changes

- `DesugaringAttributeContainerSerializer` writes a plain (non-`Named`) `Enum` as its `name()`, mirroring what `CoercingStringValueSnapshot.coerce` accepts on the way back in, so the value round-trips. This is the same representation the publishing path already uses for a plain enum (`ModuleMetadataSpecBuilder.attributeValueFor`).
- Writing such a value also **nags**, `willBecomeAnErrorInGradle10()`, against the same `unsupported_attribute_value_type` upgrade-guide section as `Attribute.of`. `Attribute.of` only reaches whoever declares the type; a consumer can read a plain-enum value back out of a cache an earlier build wrote without ever declaring it, and this covers that path.
- `CoercingStringValueSnapshot` resolves an enum type that implements `Named` by matching constants on `getName()` first, falling back to `Enum.valueOf`. `getName()` therefore stays the single serialized form for every `Named` value — module metadata cache and publishing alike — so a cache hit and a cache miss cannot desugar the same value differently.
- Value types that genuinely cannot be serialized now throw an `IllegalArgumentException` naming the attribute and the accepted types, instead of a bare `ClassCastException`.
- `desugar` is renamed to `toSerializedString`, documenting that it captures only the identity `read` needs to recover a value, not the value itself: state carried on an enum constant is not preserved.
- Every site carrying partial raw-`Enum` support is marked `// TODO: Remove support for raw Enums in Gradle 10.0.0` — `Attribute.validateSupportedType`, `AttributeTypeValidator.isKGPSpecialCase`, `DefaultImmutableAttributesEntry.desugar`, `ModuleMetadataSpecBuilder.attributeValueFor`, `DesugaringAttributeContainerSerializer.toSerializedString`, `CoercingStringValueSnapshot.coerce` — with the marker on `Attribute` naming the downstream sites, so the 10.0 scrub has one entry point.

Plain enums stay deprecated — this only makes the deprecation period behave as documented, rather than failing with an internal cast error before the deprecation can take effect in Gradle 10.

### Verification

Locally, with the fix applied:

- `./gradlew :core:integForceRealizeTest --tests "*EnumAttributeIntegrationTest"` — 45 tests, 0 failures, 0 skipped (was 1 failure, the one above); still green with the new serializer deprecation in place.
- `./gradlew :dependency-management:test --tests "*DesugaringAttributeContainerSerializerTest"` — 32 tests, 0 failures, including the round-trip specs and the new deprecation spec. The `Named`-enum spec fails without the `CoercingStringValueSnapshot` fix.
- `./gradlew :model-core:test --tests "*Snapshot*" --tests "*Isolat*"` and `--tests "*Attribute*"` in `:core` and `:dependency-management` — green.

### Contributor Checklist

- [x] Provide integration tests (under `<subproject>/src/integTest`) to verify changes from a user perspective — covered by the existing `EnumAttributeIntegrationTest`, which this PR turns green under force-realize.
- [ ] Update User Guide, DSL Reference, and Javadoc for public-facing changes — not applicable, internal classes; Javadoc on the changed methods updated.
- [ ] Ensure that tests pass sanity check: `./gradlew sanityCheck` — left to CI.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/snapshot/impl/CoercingStringValueSnapshot.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/snapshot/impl/CoercingStringValueSnapshot.java
@@ -36,7 +36,10 @@ public class CoercingStringValueSnapshot extends StringValueSnapshot {
             return type.cast(this);
         }
         if (type.isEnum()) {
-            return type.cast(Enum.valueOf(Cast.uncheckedNonnullCast(type.asSubclass(Enum.class)), getValue()));
+            // TODO: Remove support for raw Enums in Gradle 10.0.0
+            // Once every attribute value implements Named, this branch and findEnumConstant below
+            // both go away: an enum type implementing Named is served by the Named branch.
+            return type.cast(findEnumConstant(Cast.uncheckedNonnullCast(type.asSubclass(Enum.class)), getValue()));
         }
         if (Named.class.isAssignableFrom(type)) {
             return type.cast(instantiator.named(type.asSubclass(Named.class), getValue()));
@@ -45,5 +48,26 @@ public class CoercingStringValueSnapshot extends StringValueSnapshot {
             return type.cast(Integer.parseInt(getValue()));
         }
         return null;
+    }
+
+    /**
+     * Resolves an enum constant from the String form a value of that enum type is written as.
+     * <p>
+     * A value is written as {@link Named#getName()} when its type implements {@link Named} and as
+     * {@link Enum#name()} otherwise. An enum type can be both, so a match on {@code getName()} is
+     * attempted first: {@link Enum#valueOf} alone cannot read back an enum whose {@code getName()}
+     * differs from its {@code name()}.
+     */
+    private static <S extends Enum<S>> S findEnumConstant(Class<S> enumType, String value) {
+        if (Named.class.isAssignableFrom(enumType)) {
+            for (S constant : enumType.getEnumConstants()) {
+                if (((Named) constant).getName().equals(value)) {
+                    return constant;
+                }
+            }
+        }
+        // Also covers a Named enum whose getName() does return name(), and reports the unknown
+        // value the same way as before for anything that matches no constant at all.
+        return Enum.valueOf(enumType, value);
     }
 }
--- a/platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
@@ -37,8 +37,10 @@ import java.util.Optional;
 /**
  * A thread-safe and reusable attribute container serializer that will desugar typed attributes.
  *
- * Attributes that are of types different than {@code String} or {@code boolean} will be desugared
- * before serialization. The process requires the attribute type to implement {@link Named}.
+ * Attributes that are of types different than {@code String}, {@code boolean} or {@link Number} will
+ * be desugared to their name before serialization. The process requires the attribute value to
+ * implement {@link Named}, or — for the deprecated types {@code Attribute.of} still accepts — to be
+ * an {@link Enum}.
  */
 public class DesugaringAttributeContainerSerializer implements AttributeContainerSerializer {
     private final AttributesFactory attributesFactory;
@@ -107,12 +109,28 @@ public class DesugaringAttributeContainerSerializer implements AttributeContaine
                 encoder.writeString(attribute.getType().getName());
                 encoder.writeString(attributeValue.toString());
             } else {
-                // Attribute.of only accepts String, Boolean, a Number subtype, or a Named subtype;
-                // the first three are handled above, so this branch is reached only via Named subtypes.
-                Named attributeValue = (Named) container.getAttribute(attribute);
                 encoder.writeByte(DESUGARED_ATTRIBUTE);
-                encoder.writeString(attributeValue.getName());
+                encoder.writeString(desugar(attribute, container.getAttribute(attribute)));
             }
+        }
+    }
+
+    /**
+     * Desugars an attribute value to the string form {@link #read} coerces back, mirroring
+     * {@link CoercingStringValueSnapshot#coerce(Class)}.
+     * <p>
+     * {@code Attribute.of} only <em>supports</em> {@code String}, {@code Boolean}, a {@link Number}
+     * subtype or a {@link Named} subtype — the first three are handled by the caller. Any other type
+     * is deprecated rather than rejected until Gradle 10, so a plain (non-{@link Named}) {@link Enum}
+     * can still reach this point, most commonly by way of a component metadata rule.
+     */
+    private static String desugar(Attribute<?> attribute, Object value) {
+        if (value instanceof Named) {
+            return ((Named) value).getName();
+        } else if (value instanceof Enum) {
+            return ((Enum<?>) value).name();
+        } else {
+            throw new IllegalArgumentException("Cannot serialize attribute '" + attribute.getName() + "': values of type '" + attribute.getType().getName() + "' must be of type String, Boolean, a subtype of Number, or implement " + Named.class.getName() + ".");
         }
     }
 
--- a/platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
@@ -22,6 +22,7 @@ import org.gradle.api.internal.artifacts.ivyservice.resolveengine.result.Attribu
 import org.gradle.api.internal.attributes.AttributesFactory;
 import org.gradle.api.internal.attributes.ImmutableAttributes;
 import org.gradle.api.internal.model.NamedObjectInstantiator;
+import org.gradle.internal.deprecation.DeprecationLogger;
 import org.gradle.internal.serialize.Decoder;
 import org.gradle.internal.serialize.Encoder;
 import org.gradle.internal.snapshot.impl.CoercingStringValueSnapshot;
@@ -37,8 +38,10 @@ import java.util.Optional;
 /**
  * A thread-safe and reusable attribute container serializer that will desugar typed attributes.
  *
- * Attributes that are of types different than {@code String} or {@code boolean} will be desugared
- * before serialization. The process requires the attribute type to implement {@link Named}.
+ * Attributes that are of types different than {@code String}, {@code boolean} or {@link Number} will
+ * be desugared to their name before serialization. The process requires the attribute value to
+ * implement {@link Named}, or — for the deprecated types {@code Attribute.of} still accepts — to be
+ * an {@link Enum}.
  */
 public class DesugaringAttributeContainerSerializer implements AttributeContainerSerializer {
     private final AttributesFactory attributesFactory;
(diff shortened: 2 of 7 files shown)


platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
```diff
@@ -107,15 +109,31 @@ public void write(Encoder encoder, AttributeContainer container) throws IOExcept
...
+     * <p>
+     * {@code Attribute.of} only <em>supports</em> {@code String}, {@code Boolean}, a {@link Number}
+     * subtype or a {@link Named} subtype — the first three are handled by the caller. Any other type
+     * is deprecated rather than rejected until Gradle 10, so a plain (non-{@link Named}) {@link Enum}
+     * can still reach this point, most commonly by way of a component metadata rule.
+     */
+    private static String desugar(Attribute<?> attribute, Object value) {
+        if (value instanceof Named) {
+            return ((Named) value).getName();
+        } else if (value instanceof Enum) {
```
tresat, line 139: ❗ We can't generally assume the `name()` of a "raw" enum is safe to be serialized, though.  Different versions of the enum across the producer + consumer builds could change the name - this is one of the reasons why raw enums aren't safe.  I think we should either fail here, or at least emit _another_ deprecation warning.

platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
```diff
@@ -107,15 +109,31 @@ public void write(Encoder encoder, AttributeContainer container) throws IOExcept
...
+    /**
+     * Desugars an attribute value to the string form {@link #read} coerces back, mirroring
+     * {@link CoercingStringValueSnapshot#coerce(Class)}.
+     * <p>
+     * {@code Attribute.of} only <em>supports</em> {@code String}, {@code Boolean}, a {@link Number}
+     * subtype or a {@link Named} subtype — the first three are handled by the caller. Any other type
+     * is deprecated rather than rejected until Gradle 10, so a plain (non-{@link Named}) {@link Enum}
+     * can still reach this point, most commonly by way of a component metadata rule.
+     */
+    private static String desugar(Attribute<?> attribute, Object value) {
```
tresat: ❗ We shouldn't name this "desugar" - it's not necessarily completely representing the attribute value as a `String`.  Consider an `Enum` that does _not_ implement `Named`, but _does_ contain mutable state.  That state will be reset to the default by this.

platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
```diff
@@ -107,15 +109,31 @@ public void write(Encoder encoder, AttributeContainer container) throws IOExcept
                 encoder.writeString(attribute.getType().getName());
                 encoder.writeString(attributeValue.toString());
             } else {
-                // Attribute.of only accepts String, Boolean, a Number subtype, or a Named subtype;
-                // the first three are handled above, so this branch is reached only via Named subtypes.
-                Named attributeValue = (Named) container.getAttribute(attribute);
                 encoder.writeByte(DESUGARED_ATTRIBUTE);
-                encoder.writeString(attributeValue.getName());
+                encoder.writeString(desugar(attribute, container.getAttribute(attribute)));
```
tresat: ❗ This is not quite symmetrical with the `read` path for `DESUGARED_ATTRIBUTE`.  I think deserialization will fail.  We should add a full round-trip to the test.

tresat (changes requested):
I completely agree with the analysis about what's happening here, but I'm not fully convinced we make this change vs. just patching/ignoring the test until Gradle 10.0.0.

Raw enums wouldn't work pre-9.7.  Post 9.7, they work, but emit a deprecation.  Is it really a big deal if they continue to fail (after now causing a deprecation warning) until 10.0, considering this?

platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
blindpirate: You were right, and it turned out to be a real defect rather than a test gap — though not in the plain-enum path. Plain enums do round-trip: `coerce` resolves an enum-typed attribute with `Enum.valueOf`, which mirrors the `name()` written here.

The break is an enum that *also* implements `Named`. Write desugars it via `getName()` (the `Named` branch wins), but `CoercingStringValueSnapshot.coerce` checks `type.isEnum()` **before** its `Named` branch, so read does `Enum.valueOf("vanilla")`:

```
IllegalArgumentException: No enum constant ...NamedFlavor.vanilla
  at CoercingStringValueSnapshot.coerce(CoercingStringValueSnapshot.java:39)
  at DefaultImmutableAttributesContainer.getAttribute(DefaultImmutableAttributesContainer.java:147)
```

This predates #38409 — the `Named` write branch is original — and it needs no deprecated type to reproduce.

Fixed in `CoercingStringValueSnapshot`: for an enum type implementing `Named`, match constants on `getName()` first, falling back to `Enum.valueOf`. I fixed it on the read side rather than writing `name()` here on purpose — `getName()` then stays the single serialized form for every `Named` value, matching `ModuleMetadataSpecBuilder.attributeValueFor`, so a cache hit and a cache miss can't desugar the same value differently.

Added the full round-trip specs: the `Named`-enum one (fails without the `coerce` fix), and the mixed-container spec now carries a `Named` and an enum value through `DESUGARED_ATTRIBUTE` alongside the other types.

platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
blindpirate: Agreed, renamed to `toSerializedString`, and the Javadoc now says so explicitly:

> This is not a lossless representation of the value, only of the identity `read` needs to recover it: for a `Named` value the name, and for a plain `Enum` the name of the constant. Any other state a value carries — fields on an enum constant, for instance — is not preserved, and a value read back is the constant itself rather than the instance written.

platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
blindpirate: I'd argue this specific serializer is not where that risk lives, so I've kept `name()` — happy to add the warning if you still want it after this.

Two things pointed me here:

- `ModuleMetadataSpecBuilder.attributeValueFor` already writes `((Enum<?>) value).name()` for a plain enum, and that's the **publishing** path — the actual producer→consumer boundary where enum-name skew between builds can bite. So `name()` is already the established representation for these values; this change doesn't introduce a new convention, it makes an internal serializer agree with the one we ship.
- Every consumer of `DesugaringAttributeContainerSerializer` is a local cache — the module metadata cache, the value-snapshotter registry, resolution-result serialization. Those are written and read within one Gradle home, so there's no cross-build producer/consumer to disagree about the enum's constants.

On a second deprecation specifically: it would fire only when a cache write happens to occur, so the same build would warn or not depending on cache state (and only under force-realize in tests). #38409 already nags deterministically at `Attribute.of` and in the rule chains, which is where a user can actually act on it. Failing here instead would leave `EnumAttributeIntegrationTest` red under force-realize and turn the deprecation into an error early for anyone whose rule sets a plain-enum attribute.

Which brings me to your top-level point about whether to fix this at all: reviewing this did surface a genuine round-trip bug that has nothing to do with deprecated types — a `Named` enum whose `getName()` differs from `name()` could never be deserialized (details in the thread on line 113). That one I think is worth fixing regardless of 10.0.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/snapshot/impl/CoercingStringValueSnapshot.java
```diff
@@ -46,4 +46,25 @@ public <S> S coerce(Class<S> type) {
...
+     * <p>
+     * A value is written as {@link Named#getName()} when its type implements {@link Named} and as
+     * {@link Enum#name()} otherwise. An enum type can be both, so a match on {@code getName()} is
+     * attempted first: {@link Enum#valueOf} alone cannot read back an enum whose {@code getName()}
+     * differs from its {@code name()}.
+     */
+    private static <S extends Enum<S>> S findEnumConstant(Class<S> enumType, String value) {
+        if (Named.class.isAssignableFrom(enumType)) {
+            for (S constant : enumType.getEnumConstants()) {
+                if (((Named) constant).getName().equals(value)) {
```
tresat, line 64: 💅 Probably doesn't matter much, but `((Named) constant).getName()` can be lifted out of the loop.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/snapshot/impl/CoercingStringValueSnapshot.java
```diff
@@ -36,7 +36,7 @@ public <S> S coerce(Class<S> type) {
             return type.cast(this);
         }
         if (type.isEnum()) {
-            return type.cast(Enum.valueOf(Cast.uncheckedNonnullCast(type.asSubclass(Enum.class)), getValue()));
+            return type.cast(findEnumConstant(Cast.uncheckedNonnullCast(type.asSubclass(Enum.class)), getValue()));
```
tresat, line 42: 🤔 Both the new and the old code seem like they would allow incorrectly coercing between different types of `Enum`s.

```
public enum TrafficLightColors {
  RED,
  GREEN;
}

public enum AntTypes {
  BLACK,
  RED;
}

// Given: this = AntTypes.RED
TrafficLightColors color = coerce(TrafficLightColors.class); // works, but shouldn't work
```

Which is another motivation for dropping raw `Enum` support and just using the `Named` interface.

platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
tresat: > ModuleMetadataSpecBuilder.attributeValueFor already writes ((Enum<?>) value).name() for a plain enum, and that's the publishing path — the actual producer→consumer boundary where enum-name skew between builds can bite.

Yes, and we consider publishing of raw `Enum`s broken due to this, as `public enum MyValue implements Named` will fail if `name()` and `getName()` return different values.  If a raw enum later acquires `Named` in a later version of a consumer, then there are other potential problems.

> Every consumer of DesugaringAttributeContainerSerializer is a local cache

This is true now, but I don't believe it was always true in older versions of Gradle.

`Enum` attributes are just full of footguns, and the best solution is banning raw `Enum` values in favor of just using `Named` and not making any assumptions about whether it's a class or enum implementing it.  

I'm less concerned about what happens with them prior to 10.0, since they were broken before 9.7 and are now deprecated.  For 10.0, we can thoroughly scrub this code to remove all traces of raw enums as `Attribute` values.

I don't think it would hurt to throw another deprecation here - just in case anyone is merely deserializing these instead of creating new `Attribute`s.  It might make the removal of raw `Enum`s in Gradle 10.0 less surprising.

tresat (changes requested):
Can you add some markers throughout this: `// TODO: Remove support for raw Enums in Gradle 10.0.0` first, before we call this done.  Let's make sure we don't forget to scrub the partial `Enum` support then.

blindpirate:
@bot-gradle squash

platforms/software/dependency-management/src/main/java/org/gradle/internal/resolve/caching/DesugaringAttributeContainerSerializer.java
blindpirate: Fair points — I'd been reasoning from what's true today, and "this was true in older Gradle" plus "someone may only ever be deserializing" both cut the other way.

Added the nag in `toSerializedString`, `willBecomeAnErrorInGradle10()`, against the same `unsupported_attribute_value_type` upgrade-guide section as `Attribute.of`, with a spec asserting it fires. `EnumAttributeIntegrationTest` under force-realize is still green with it in place (45 tests, 0 failures), so the extra nag doesn't disturb the test this PR was fixing.

And agreed on the endgame: the 10.0 scrub is the real fix, this only keeps the deprecation window behaving as documented in the meantime.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/snapshot/impl/CoercingStringValueSnapshot.java
blindpirate: `getName()` is called on `constant`, which changes each iteration, so it can't be lifted — but the per-iteration `(Named)` cast can go, since the array's runtime component type already implements `Named`:

```java
Named[] constants = (Named[]) enumType.getEnumConstants();
for (Named constant : constants) {
    if (constant.getName().equals(value)) {
        return enumType.cast(constant);
    }
}
```

Happy to take that if it's what you meant, though the current form reads more plainly to me and this only runs on a cache miss for a `Named` enum. Let me know which you'd prefer.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/snapshot/impl/CoercingStringValueSnapshot.java
blindpirate: Confirmed, and it's not new or enum-specific — it's what desugaring *is*. Once a value has been through `DESUGARED_ATTRIBUTE` it's a bare `String`; the type is gone, and `DefaultImmutableAttributesEntry.coerce` matches by attribute **name** and then coerces to whatever type *the other side* declares.

The `Named` branch two lines down is exactly as permissive: `instantiator.named(TrafficLightColors.class, "RED")` will manufacture a value from any string a producer happened to write, with no check that the producer's type had such a constant at all. So switching to `Named`-only narrows nothing here.

That looseness is deliberate for desugaring — the consumer by construction doesn't have the producer's class — and the real guard is that two builds agreeing on an attribute *name* are expected to agree on its meaning. So I'd leave it as is in this PR; tightening it means changing the desugared representation itself to carry the declared type, which is well outside this fix. Marked with a TODO either way.

blindpirate:
> Can you add some markers throughout this: `// TODO: Remove support for raw Enums in Gradle 10.0.0` first, before we call this done.

Done. Markers are on every site that carries partial raw-`Enum` support:

- `Attribute.validateSupportedType`
- `AttributeTypeValidator.isKGPSpecialCase`
- `DefaultImmutableAttributesEntry.desugar`
- `ModuleMetadataSpecBuilder.attributeValueFor`
- `DesugaringAttributeContainerSerializer.toSerializedString`
- `CoercingStringValueSnapshot.coerce`

The one on `Attribute` spells out that the two deprecation branches become hard failures and names the downstream sites, so the scrub has a single entry point rather than six independent discoveries.

Also in this push, per the thread on the serializer: writing a plain `Enum` now nags on its own, so a consumer that only ever deserializes such a value out of an older build's cache is covered too.

Verified locally:

- `:core:integForceRealizeTest --tests "*EnumAttributeIntegrationTest"` — 45 tests, 0 failures, 0 skipped (was 1 failure). Still green with the new nag.
- `:dependency-management:test --tests "*DesugaringAttributeContainerSerializerTest"` — 32 tests, 0 failures, including the new deprecation spec.

Two threads I've replied to rather than changed code on — the 💅 hoist (can't be lifted as written; offered the variant I think you meant) and the 🤔 cross-enum coercion (pre-existing and equally true of the `Named` branch, so `Named`-only doesn't narrow it). Happy to act on either if you disagree.

platforms/core-configuration/model-core/src/main/java/org/gradle/internal/snapshot/impl/CoercingStringValueSnapshot.java
tresat: My mistake - this is fine.

tresat (approved):
Thanks for all the followup, LGTM now.

PR #38722, opened 2026-08-03
Check that the IDEA project SDK supports the project language level

cobexer:
`.idea/misc.xml` commits both the project language level and the project SDK name, and they have to agree. IntelliJ maintains the language level from the Gradle model, but the SDK name is a reference into the developer's own SDK table, so it can silently fall behind — which is how the repo ended up reporting "Module JDK is misconfigured" on import until #38721 fixed it by hand.

- Add a `checkIdeaJdkConfiguration` task that fails when the committed project SDK cannot support the committed language level, and wire it into `sanityCheck`.
- The SDK name is a free-form label, so the task infers a version by convention and skips the check when it cannot.
- Extract `createSecureDocumentBuilder()` into `gradlebuild:basics` so the new task reuses it rather than duplicating the parser hardening. No behaviour change.

The check compares the two values in `misc.xml` against each other rather than against `gradle/gradle-daemon-jvm.properties`, because Gradleception rewrites that file before running `sanityCheck`. The trade-off is that it won't catch both values drifting downwards together.

### DIFF
--- a/build-logic-commons/basics/src/main/kotlin/gradlebuild/basics/XmlFactories.kt
+++ b/build-logic-commons/basics/src/main/kotlin/gradlebuild/basics/XmlFactories.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


ov7a (approved):

asodja (approved):
LGTM

PR #38737, opened 2026-08-04
Delete DependencyState

jvandort:
This class was mostly leftover from a prior graph engine architecture where it was in charge of managing dependency substitutions for an edge and as the basis for a SelectorState. Most of this logic has been extracted out, so we can inline the class now

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

PR #38738, opened 2026-08-04
Track user code application timings in UserCodeApplicationContext

jvandort:
Previously, per-plugin configuration timings for TAPI PROJECT_CONFIGURATION events were reconstructed from the build operation tree by two trackers in tooling-api-builders, which attributed each ApplyPlugin/listener/callback operation's wall time to the ConfigureProject operation enclosing it.

UserCodeApplicationContext now records this data directly. Each application is registered against the target it is applied to (project or other), and the time spent executing each application is accumulated via a thread-local timer, broken down by code type (general, collection callback, listener). A recording is started around the build tree action and stopped when it completes; ProjectBuilder starts a recording that is never stopped, since it has no lifecycle. Applying user code outside a recording is an error.

This removes the need to track build operations for TAPI events and lays the groundwork for reporting per-application timings to Develocity via the Run Build build operation (see TODO in DefaultBuildTreeActionExecutor). Develocity currently derives these timings from the build operations emitted by CollectionCallbackActionDecorator and
ListenerBuildOperationDecorator; once it consumes the recorded timings instead, those per-callback build operations can be removed entirely, which initial testing shows can improve configuration time by ~10%.

This changes the semantics of ProjectConfigurationOperationResult:

- Plugin applications are attributed to the project the plugin is applied to, rather than the project being configured when the application happened to execute. A plugin applied to subproject :b from the root's allprojects block is now reported under :b, not the root.
- Durations are exclusive: nested plugin applications are billed to themselves rather than being included in the applying plugin's wall time, so durations no longer double count and can be meaningfully summed.
- Applications that happen outside project configuration (e.g. buildSrc's auto-applied plugins at projectsLoaded) are now visible. As before, plugins applied by class rather than by id report a null plugin id.
- Listener/callback time that accrues after the owning project's configuration has completed (e.g. a root plugin's subprojects afterEvaluate block) is no longer included in any per-project event. It remains part of the application's total recorded time.

Results are still merged per plugin identifier with summed durations, as required by the ProjectConfigurationOperationResult contract.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle help

jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test apt

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/DefaultUserCodeApplicationContext.java
```diff
@@ -16,67 +16,148 @@
...
+    public DefaultUserCodeApplicationContext(NanoTimeProvider timeProvider) {
+        this.timeProvider = timeProvider;
+    }
 
     @Override
-    @Nullable
-    public Application current() {
-        return this.currentApplication.get();
+    public Recording startRecording() {
+        if (this.recording != null) {
```
alllex, line 67: 💭 Starting/stopping recordings isn't expected to happen concurrently, but it might still be clearer to synchronize those operations on the Context instance. The current implementation of start is racy in theory, if two threads call it.

Also, see my other comments on the suggested update to the shape of this API

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/DefaultUserCodeApplicationContext.java
```diff
@@ -92,41 +173,201 @@ public UserCodeSource getSource() {
...
+
+        /**
+         * Return all applications applied to the given target.
+         * <p>
+         * This method is thread-safe, but not atomic. This method may be called concurrently while
+         * user code applications are being registered or executed. However, applications registered
+         * while this method is executing may or may not be included in the returned list.
+         */
+        public ImmutableList<Application> getApplicationsFor(Target target) {
+            return ImmutableList.copyOf(applications.getOrDefault(target, Collections.emptyList()));
```
alllex, line 353: 💭 Here and below, we are doing a `copyOf` a synchronized list. The behavior contract of the latter requires to manually synchronize on the instance of the list in case of iterator operations. In the current version of Guava, `copyOf` calls `.toArray()` on the collection, which makes the operation safe, however, this is rather opaque. It's possible in the future to refactor this innocent-looking code into something with a potential race condition. 

At the expense of being verbose, we could have a holder class that makes the synchronization explicit. The only operation we are calling on the list is `add()` and then copying, so it shouldn't be too bad.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeApplicationContext.java
```diff
@@ -65,18 +95,124 @@ interface Application {
...
+            }
+
+        }
+
+        /**
+         * A target modeling some non-project gradle domain.
+         */
+        class Other implements Target {
+
+            public static final Other INSTANCE = new Other();
```
alllex, line 209: ❓ If instances of Target ever get serialized, e.g. via CC-serialization, which can handle beans, this will result in more than one "singleton" instance. Do we need to take care of this now?

In other place, we've been using enums with a single value for similar purposes.

subprojects/core/src/main/java/org/gradle/configuration/BuildOperationScriptPlugin.java
```diff
@@ -62,7 +64,8 @@ public void apply(final Object target) {
         } else {
             URI uri = resource.getFile() != null ? resource.getFile().toURI() : resource.getLocation().getURI();
             UserCodeSource source = new UserCodeSource.Script(getSource().getShortDisplayName(), uri);
-            userCodeApplicationContext.apply(source, userCodeApplicationId -> buildOperationRunner.run(new RunnableBuildOperation() {
+            UserCodeApplicationContext.Target projectIdentityPath = getApplicationTargetFor(target);
```
alllex, line 67: 💅 The variable name is off: we don't know if it's a project even

platforms/core-runtime/build-operations/src/main/java/org/gradle/internal/operations/BuildOperationRunner.java
```diff
@@ -18,12 +18,14 @@
...
 import org.gradle.internal.service.scopes.Scope;
 import org.gradle.internal.service.scopes.ServiceScope;
+import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
 /**
  * Runs build operations: the pieces of work that make up a build.
  * Build operations can be nested inside other build operations.
  */
+@NullMarked
```
alllex, line 28: ❔ Do we need this? The entire package seems to be null-marked?

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/DefaultUserCodeApplicationContext.java
```diff
@@ -16,67 +16,148 @@
...
+    /**
+     * Tracks the current application in progress for the current thread. Null when the current
+     * thread has never executed user code for this context.
+     * <p>
+     * This is deliberately not static. Each context tracks its own applications independently,
+     * so that user code running in one context is transparent to the application timings of
+     * another. In production, this scenario is not expected to occur, but this is likely more
+     * correct in testing scenarios involving multiple {@code ProjectBuilder} instances.
+     * <p>
+     * This intentioanlly does not use {@link ThreadLocal#withInitial}. Threads that only ever
```
alllex, line 45: 💅 `intentioanlly` → `intentionally`

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/ToolingApiBuildEventListenerFactory.java
```diff
@@ -67,18 +79,16 @@ public Iterable<Object> createListeners(BuildEventSubscriptions subscriptions, B
             listeners.add(new BuildPhaseOperationListener(progressEventConsumer, idFactory));
         }
 
-        listeners.add(createClientBuildEventGenerator(subscriptions, consumer, progressEventConsumer));
+        listeners.add(createClientBuildEventGenerator(subscriptions, consumer, progressEventConsumer, userCodeApplicationContext));
         return listeners.build();
     }
 
-    private ClientBuildEventGenerator createClientBuildEventGenerator(BuildEventSubscriptions subscriptions, BuildEventConsumer consumer, ProgressEventConsumer progressEventConsumer) {
+    private ClientBuildEventGenerator createClientBuildEventGenerator(BuildEventSubscriptions subscriptions, BuildEventConsumer consumer, ProgressEventConsumer progressEventConsumer, UserCodeApplicationContext userCodeApplicationContext) {
```
alllex, line 86: 💅 Put params on lines for readability

subprojects/core/src/main/java/org/gradle/internal/build/event/BuildEventListenerRegistryInternal.java
```diff
@@ -52,6 +52,9 @@ public Provider<?> getListener() {
 
     /**
      * Subscribes the given listener to build operation completion events. Note that no start events are forwarded to the listener.
+     *
```
alllex, line 55: 💅 `<p>`

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r51/ProjectConfigurationProgressEventCrossVersionSpec.groovy
```diff
@@ -136,12 +136,25 @@ class ProjectConfigurationProgressEventCrossVersionSpec extends ToolingApiSpecif
         runBuild("tasks")
 
         then:
-        containsPluginApplicationResultsForJavaPlugin(":buildSrc")
-        doesNotContainPluginApplicationResultsForJavaPlugin(":buildSrc:a")
-        containsPluginApplicationResultsForJavaPlugin(":")
-        doesNotContainPluginApplicationResultsForJavaPlugin(":b")
-        containsPluginApplicationResultsForJavaPlugin(":included")
-        doesNotContainPluginApplicationResultsForJavaPlugin(":included:c")
+        if (targetVersion >= GradleVersion.version("9.8")) {
```
alllex, line 139: ❓This and similar places have to change to `9.9.0`, right?

platforms/extensibility/unit-test-fixtures/src/main/java/org/gradle/testfixtures/internal/ProjectBuilderImpl.java
```diff
@@ -164,6 +165,11 @@ public ProjectInternal createProject(String name, @Nullable File inputProjectDir
         BuildRequestMetaData buildRequestMetaData = new DefaultBuildRequestMetaData(Time.currentTimeMillis());
         File userActionRootDir = globalServices.get(BuildLayoutFactory.class).getLayoutFor(startParameter.toBuildLayoutConfiguration()).getRootDirectory();
         CrossBuildSessionState crossBuildSessionState = new CrossBuildSessionState(globalServices, startParameter, userActionRootDir);
+
+        // User code applications require a recording to be in progress. ProjectBuilder has no
+        // lifecycle, so the recording is started here and never stopped.
```
alllex, line 170: 💭 There is a tear down procedure below that is used by tests (`org.gradle.testfixtures.internal.ProjectBuilderImpl#stop`), so we can consider stopping the recording there. I also wouldn't be surprised if plugins out there use the same method for faster tests.

 I know we should have abandoned ProjectBuilderImpl a long time ago, but I keep encountering plugins actively using it, so while the maintenance is not expensive, we should not make it potentially worse.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeApplicationContext.java
```diff
@@ -65,18 +95,124 @@ interface Application {
...
+
+        /**
+         * Get a snapshot of the total time spent executing this application, in nanoseconds.
+         */
+        long getTotalDurationNs();
+
+        /**
+         * Get a snapshot of the time spent executing code of the given {@link CodeType}, in nanoseconds.
+         */
+        long getDurationNsForType(CodeType codeType);
```
alllex, line 130: 💅 If this is for testing purposes only, I'd annotate it as such

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeApplicationContext.java
```diff
@@ -16,47 +16,77 @@
...
 @ServiceScope(Scope.CrossBuildSession.class)
 public interface UserCodeApplicationContext {
+
+    /**
+     * Begins recording user code applications. All {@link #apply} actions executed on all
+     * threads are recorded until the returned recording is stopped.
+     *
+     * @throws IllegalStateException If a recording is already in progress.
+     */
+    Recording startRecording();
```
alllex, line 44: 💭 I am not sure this API surface needs the `Recording` exposed.

The Context instance cannot have more than one recording at a time, and stopping of the recording changes the state of the Context instance, which seems off. On top of that, the actual reporting of the applications goes through the Context instance anyway. That also includes the intermediate reporting of the applications, before all of the recording is stopped.

Given this mismatch, I suggest to lift the lifecycle methods onto the context itself. Then make it explicit in the doc strings of other methods their relation with the lifecycle.

I don't feel strongly about the naming, but I have a weak preference for something like `start/stopTrackingApplications`. The stop method can still return the aggregate to avoid keeping that state in the Context, after the tracking has been stopped.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeApplicationContext.java
```diff
@@ -16,47 +16,77 @@
...
-    @Nullable
-    Application current();
+    ImmutableList<Application> getApplicationsFor(Target target);
 
     /**
-     * Immutable representation of the application of some user code.
+     * Representation of the application of some user code. Tracks the amount of time spent
+     * executing the application.
      */
     interface Application {
```
alllex, line 85: 🤔 What feels off about this type is that it tries to be both the mutable tracker of time and also the result that we return from `getApplicationsFor` and from the `stop()`.

On one side, whoever obtains an instance of this via `current()` should not be calling methods to get current durations. On the other side, whoever gets the results via `getApplicationsFor()` or via `stop()` should have no affordance to call `reapply*`.

Another related concern is about potentially missing applications that were reported after `getApplicationsFor()` was called. Today, the API states that the contract is lenient, but I am not sure there is a reason for us not to make it strict.

It seems that each application also has a lifecycle and that after it has finished, we expect no new re-applications and can safely collect the durations.

I propose to split this interface in two, introducing something like `ApplicationSummary`, which only exposes the identity and the results. Similar to how the higher-level recording works, the `Application` itself can have a `stop()` method, which will return the summary, but also, importantly, enforce the invariant that stopped applications can no longer collect data.

With this setup, the `current()` continues to return `Application` and it's clear that its callers shouldn't necessarily stop the application. Although, maybe they actually can in some cases?

Then the `getApplicationsFor` and the full recording result return summaries instead of live `Applications`. And, importantly, obtaining those summaries makes it so all corresponding applications have been stopped. 

Also, correct me if I'm wrong, but the number of `Applications` isn't expected to be very high, because it is roughly N-projects * M-plugins. Then there is no need to be very strict about reusing object instances for different purposes to potentially avoid memory pressure.

alllex (changes requested):
The implementation seems solid overall, I mainly have questions about the shape of the internal API and some more minor things.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r51/ProjectConfigurationProgressEventCrossVersionSpec.groovy
alllex, line 1: ❓ I'd also add to this coverage tests that apply plugins to projects from settings via callbacks like `gradle.rootProject {}`, `gradle.allprojects {}` and `gradle.lifecycle.beforeProject {}`

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/DefaultUserCodeApplicationContext.java
jvandort: Added a `recordingLock` which must be held when mutating the `recording`. `recording` is still volatile, so reading its value does not require acquiring the lock.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/DefaultUserCodeApplicationContext.java
jvandort: Replaced `Collections.synchronizedList` with a small custom `CopyOnReadArrayList` which wraps an `ArrayList`, makes synchronization explicit, and performs the list copy while synchronized.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeApplicationContext.java
jvandort: I don't foresee this ever getting serialized. This is only used as a key in the recording's application map. But, it was easy enough to make an enum so I did.

subprojects/core/src/main/java/org/gradle/configuration/BuildOperationScriptPlugin.java
jvandort: Good catch. Renamed to `appTarget` like the other call site.

platforms/core-runtime/build-operations/src/main/java/org/gradle/internal/operations/BuildOperationRunner.java
jvandort: No, removed. Something about IJ's handling of generics and nullability is broken and this was me attempting to fix it. But, given the package annotation this did nothing

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/DefaultUserCodeApplicationContext.java
jvandort: Fixed

platforms/ide/tooling-api-builders/src/main/java/org/gradle/tooling/internal/provider/runner/ToolingApiBuildEventListenerFactory.java
jvandort: Done

subprojects/core/src/main/java/org/gradle/internal/build/event/BuildEventListenerRegistryInternal.java
jvandort: Added

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r51/ProjectConfigurationProgressEventCrossVersionSpec.groovy
jvandort: Yes, assuming we get this into 9.9.

Updated to 9.9

platforms/extensibility/unit-test-fixtures/src/main/java/org/gradle/testfixtures/internal/ProjectBuilderImpl.java
jvandort: Stopped the recording as part of project builder tear down.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeApplicationContext.java
jvandort: This method will be used in the follow-up PR when we actually integrate with DV by emitting a build operation. DV wants per-type timings, and this is the intended method to expose them.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeApplicationContext.java
jvandort: Removed Recording as an API. Moved the stop method to `UserCodeApplicationContext`. Renamed to `start/stopTrackingApplications`.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/code/UserCodeApplicationContext.java
jvandort: Agreed, a consumer of an application should either be mutating timings or reading timings, never both. Split `Application` into a separate `ApplicationSnapshot`. I don't foresee an issue with creating an additional object instance per Application (for its snapshot), as the number of applications are indeed relatively low.

However, your proposal to finalize an application after it has been snapshotted will not work in the current model. `getApplicationsFor` is called in production by  `ProjectConfigurationOperationMapper` in response to a `ConfigureProjectBuildOperationType` build operation being emitted. That means we snapshot a project's timings when that project has finished evaluating (see `LifecycleProjectEvaluator`). Just because a project has finished evaluating does not mean code within it is finished executing. 

Consider a build like:
```
// build.gradle
plugins {
    id("org.foo.a")
}

// plugin org.foo.a
tasks.register("foo") {
    // This callback will be recorded as COLLECTION_CALLBACK timing for plugin org.foo.a
}

// child/build.gradle
project(":").tasks.all { } // Executes code in root after root is done evaluating
```

Note, you don't need to use IP unsafe things, to do this. Simply running `./gradlew :foo` will realize `foo` when building the work graph, a step which is performed after project evaluation. Or, you can use resolve a Configuration in one project to trigger code in another project.

Now, we _could_ change when we emit the TAPI `InternalProjectConfigurationResult` to sometime after we configure all projects. However, I would not want to do this here. It is an existing issue that the TAPI progress events miss timings.

alllex: I see, the timing of the operations is unfortunate. Maybe we'll be able to finalize snapshots once we migrate to the aggregated reporting. I agree it doesn't make sense to go against the current now.

platforms/ide/tooling-api/src/crossVersionTest/groovy/org/gradle/integtests/tooling/r51/ProjectConfigurationProgressEventCrossVersionSpec.groovy
jvandort: Added tests

jvandort:
@bot-gradle help

jvandort:
@bot-gradle test ACT

alllex (approved):
LGTM!

jvandort:
@bot-gradle test apt

Issue #13835, opened 2020-07-17
Explain memory leak-like slowdown when FSW is enabled on Gradle build


lptr:
@lptr commented on [Wed Jun 03 2020](https://github.com/gradle/gradle-private/issues/3111)

This is from the Gradle build using 6.5-rc-1:

![image](https://user-images.githubusercontent.com/495366/83620105-26507400-a58d-11ea-858e-e9944f2df4f7.png)

Raw results:
[profile-out.zip](https://github.com/gradle/gradle-private/files/4722793/profile-out.zip)

---
cc: @gradle/build-cache

---

@lptr commented on [Wed Jun 03 2020](https://github.com/gradle/gradle-private/issues/3111#issuecomment-638122687)

I'm running some tests with build scans to see what's happening.

---

@wolfs commented on [Tue Jun 09 2020](https://github.com/gradle/gradle-private/issues/3111#issuecomment-641384959)

The problem is very strange. The heap memory seems to be fine, this is from 5 builds which took about a minute each (instead of 7s):
![image](https://user-images.githubusercontent.com/423186/84165833-6d9aa100-aa74-11ea-8013-a8f0c4a1c1d2.png)

This looks pretty normal. GC activity is never more than 10%.

![image](https://user-images.githubusercontent.com/423186/84166236-ef8aca00-aa74-11ea-94d6-5ef509b0575f.png)

Looking at the hotspots, a very long time is spent on `CurrentBuildOperationRef.get` (146s). I have no explanation for this.

The whole slowdown is most apparent when capturing build operations from Gradle profiler. Then there are builds which take minutes. When not capturing build operations, the builds also seem to become slower. After 200 iterations there are also builds which take up to 40 seconds.

---

@wolfs commented on [Wed Jun 10 2020](https://github.com/gradle/gradle-private/issues/3111#issuecomment-641997596)

I captured a build scan when this problem happened: https://e.grdev.net/s/r2iejo5xak2ic/performance/configuration

It seems like configuration is taking insanely long, especially applying the `gradle-distribution` plugin:
![image](https://user-images.githubusercontent.com/423186/84271623-a007d500-ab2c-11ea-9389-8abe71c73482.png)

Here is an even more extreme case: https://e.grdev.net/s/of77cjtgku4lw/performance/configuration

---

@wolfs commented on [Wed Jun 10 2020](https://github.com/gradle/gradle-private/issues/3111#issuecomment-642002510)

[...]

PR #38745, opened 2026-08-05
Avoid ThreadLocal initialization on read-only paths

jvandort:
EvaluationContext and DefaultMutationGuard used ThreadLocal.withInitial, which materializes and stores a value on every thread that merely queries the state, and routes initialization through a supplier call site shared by all such thread locals in the JVM, which the JIT cannot inline or devirtualize. DefaultMutationGuard additionally removed the entry again whenever it held the default value, paying the full miss-initialize-remove cycle on every read in the default state.

Represent the default state as an absent entry instead: reads never write to the thread local map, and an entry exists only while a non-default state is active on the thread. For DefaultMutationGuard, this preserves the gradle/gradle#13835 guarantee that unused entries do not accumulate on long-lived daemon threads, without the read-side removal. EvaluationContext's thread local also becomes static, since the class is a singleton.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DefaultMutationGuard.java
+++ b/platforms/core-configuration/domain-object-collections/src/main/java/org/gradle/api/internal/DefaultMutationGuard.java
@@ -17,13 +17,25 @@
 package org.gradle.api.internal;
 
 import org.gradle.api.Action;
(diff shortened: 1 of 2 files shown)


jvandort:
@bot-gradle test apt

PR #38747, opened 2026-08-05
Improve performance of nanoTime on Windows

jvandort:
Our Windows CI machines report the following System.nanoTime stats:

```
os: Windows 10 10.0 (amd64), cores: 10
jvm: OpenJDK 64-Bit Server VM 17.0.20+8
nanoTime cost, sample 1: 12553 ns/call
nanoTime cost, sample 2: 18071 ns/call
nanoTime cost, sample 3: 19765 ns/call
currentTimeMillis cost: 276 ns/call (sink 1)
parkNanos(1ms) achieved: 15602 us average
Thread.sleep(1) achieved: 1734 us average
Time.clock().getCurrentTime() cost: 41118 ns/call over 5000 calls
```

nanoTime takes on the order of tens of microseconds, and calls to Clock.getCurrentTime as a result
take the same order of magnitude time. Our Clock is used for for build operations timestamps, tapi
progress events, and logging, all of which happen on the fast path. This slow timer causes builds
on our CI machines to take _seconds_ of extra time just due to the slow timer. This is likely
happening in real world builds on Windows machines with similarly configured timers.

To resolve this, we detect slow timers on process startup, by probing the platform timer in a separate
thread. If the system timer is determined to be slow, we use that thread to periodically, every 1ms,
update a volatile field with the current time. We then install a TimeSource backed by this volatile field
instead of the existing System.nanoTime backed timer

Windows perf test results: 
image

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-execution/worker-shared/src/main/java/org/gradle/internal/service/scopes/WorkerSharedGlobalScopeServices.java
+++ b/platforms/core-execution/worker-shared/src/main/java/org/gradle/internal/service/scopes/WorkerSharedGlobalScopeServices.java
@@ -69,6 +69,7 @@ import org.gradle.internal.state.DefaultManagedFactoryRegistry;
 import org.gradle.internal.state.ManagedFactoryRegistry;
 import org.gradle.internal.time.Clock;
 import org.gradle.internal.time.Time;
+import org.gradle.internal.time.TimeSourceManager;
 
 import java.util.Collections;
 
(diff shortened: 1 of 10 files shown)


jvandort:
@bot-gradle help

jvandort:
@bot-gradle test QuickFeedback on windows

jvandort:
@bot-gradle test apt

jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

big-guy (approved):

PR #36037, opened 2025-12-15
Run buildFinished action even if tasks fail

asodja:
This PR changes how build actions are executed if tasks are present.

Now if a task fails we still continue and run a build action. 

If task and build action fails we then report both errors to console. TAPI client gets only build action error, but they can listen to task failures with:
```
ProjectConnection.addProgressListener(new ProgressListener(), OperationType.TASK)
```
We can improve that in the future if needed, but reporting mutliple exceptions to TAPI client seems untrivial.

This PR also fixes resilient queries with "default target" (e.g. using `fetch(Model)` instead of `fetch(projectTarget, Model)`.

PR #38761, opened 2026-08-06
Simplify BuildTreeWorkController to return a plain ExecutionResult

asodja:
TaskRunResult is not needed anymore since we unified resilient and non-resilient logic in previous versions. So we can just return ExecutionResult.


asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

6hundreds (approved):
Wonderful! What're the previous changes allow us to do this?

asodja:
@6hundreds In #36037 commit 08d03af57b8 we simplified the logic around task configuration and execution failures. Before we swallowed the configuration failure and report only the task execution failure. Now the configuration and execution failure is treated in the same way

alllex (approved):

PR #38766, opened 2026-08-06
make DistributionIntegrationSpec more permissive for releases

cobexer:


ov7a (approved):
Note: this is a temporary solution

PR #38774, opened 2026-08-07
Update Tooling API docs: reading intermediate models with a phased build action

asodja:
### Context

Follow-up to the resilient sync documentation in the Tooling API user guide:

- Adds a full example of a two-phased build action (`ProjectConnection.action()` with `projectsLoaded` and `buildFinished`) showing how models are read through `IntermediateResultHandler`s. The example lives in the existing `fetchModel` snippet, so it is compiled by the existing sample test.
- Documents that starting with the Gradle 9.7 daemon, a resilient sync still fails as a whole when any part of it fails: `run()` throws after the intermediate handlers have received their models, so clients must read models via intermediate handlers of a phased build action.
- Moves "Running tasks together with a build action" below the new section, so phased actions and `IntermediateResultHandler` are introduced before they are referenced.

Preview locally with `./gradlew serveDocs -PquickDocs` (http://localhost:8000/userguide/tooling_api.html); snippet test: `./gradlew :docs:docsTest --tests '*integration-fetchModel*'`.


asodja:
@bot-gradle test BD

lkasso (approved):

PR #38777, opened 2026-08-07
Decouple DefaultTaskExecutionGraph from plan execution

jvandort:
DefaultTaskExecutionGraph implements TaskExecutionGraph, a public API for users to query the state of the graph prior to execution. The class also was in the 'main path' of actual execution, which is undesirable.

This commit separates the class's concerns so that it only acts as a public API implementation. All logic related to actually executing a graph has been moved out of the class into SelectedTaskExecutionAction. DefaultTaskExecutionGraph no longer has any ownership claim to the FinalizedExecutionPlan it wraps. Configuration Cache still gets its FinalizedExectionPlan throgh TaskExecutionGraphInternal. This should be relatively easy to change with some extra wiring later on. We should then deprecate querying any methods on TaskExecutionGraph outside of the whenReady callback, allowing DefaultTaskExecutionGraph to remove any final long-lived reference to FinalizedExecutionPlan.

DefaultTaskExecutionGraphSpec was more of an integration test than anything, combining DefaultPlanExecutor and DefaultExecutionPlan. Most of its existing tests were duplicated by tests in DefaultExecutionPlanTest, and in practice those duplicate tests did not belong in that file, as they actually verified behavior of DefaultExecutionPlan. Duplicate and misplaced tests have been deleted and moved to their appropriate files. SelectedTaskExecutionActionTest has been created to test execution logic that was moved out of DefaultTaskExecutionGraph.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

PR #38781, opened 2026-08-07
Consolidate BuildWorkExecutor implementations

jvandort:
There were a number of implementations of BuildWorkExecutor. In practice they all delegated to one another. We merge all logic into a single Default implementation, making the logic easier to follow.

We additionally inline logic used in the 'real' execution path, as it was only used in a single place. We were able to delete a number of interfaces, including NodeExecutor and SelfExecutingNode. Now, all nodes are 'self executing'

Finally, we move the responsibility of carrying the current build operation between threads to DefaultPlanExecutor, as it is the one actually creating threads. Previously, the work executor was responsible for this even though it didn't necessarily switch threads. Tests were added in the appropriate test files to verify this moved behavior.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

PR #38786, opened 2026-08-09
Remove all projects lock

jvandort:
Exclusive access to all projects of a build was granted by AllProjectsLock, an
additional lock in a build that every project lock had to consult before it could
be acquired. Coordinating the two made things more complicated than necessary.
ProjectState.runWithModelLock required additional bookkeeping rather than using 
standard locking primitives. This made it difficult to pass around just a `ResourceLock`
to model a Project's state lock, instead necessitating using a `ModelContainer<?>` which
provided methods that also checked if the all projects lock was acquired already.

Take the locks of all the projects instead. The coordination service already
acquires a set of locks atomically and rolls back on failure, so this waits on
exactly the condition AllProjectsLock.canAcquire() tested, and a thread still
cannot accumulate project locks one at a time and deadlock. That makes both the
second lock and the one-project-lock-per-thread rule unnecessary.

runWithModelLock and hasMutableState now just ask whether this project's lock is
held. holdsProjectLock answers that without copying the thread's locks,
which matters now that a thread can hold one per project.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


jvandort:
@bot-gradle test apt

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/AbstractResourceLockRegistry.java
```diff
@@ -64,60 +46,28 @@ public boolean hasOpenLocks() {
...
-        lockDetails.locks.remove(resourceLock);
-    }
+    /**
+     * Return true if the current thread is allowed to acquire or release locks of this registry.
+     */
+    public abstract boolean mayAttemptToChangeLocks();
 
-    public boolean mayAttemptToChangeLocks() {
-        return detailsForCurrentThread().mayChange;
+    @SuppressWarnings("deprecation") // Thread.getId() is deprecated since JDK 19, but the replacement Thread.threadId() does not exist on JDK 17 (production target).
```
octylFractal: 💭 Might be worth mentioning here that `getId` is not deprecated for removal, so this is probably OK for the foreseeable future.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLeaseRegistry.java
```diff
@@ -108,4 +107,5 @@ public interface ProjectLeaseRegistry extends BlockingNotifier {
      * For now, this is an opt-in behaviour.</p>
      */
     <T> T whileDisallowingProjectLockChanges(Factory<T> action);
+
```
octylFractal, line 110: 💅 Remove extra whitespace change.

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
```diff
@@ -266,23 +267,36 @@ public Set<? extends ProjectState> getAllProjects() {
 
         @Override
         public <T extends @Nullable Object> T fromMutableStateOfAllProjects(Function<AllProjectsAccess, T> factory) {
-            ResourceLock allProjectsLock = workerLeaseService.getAllProjectsLock(owner.getIdentityPath());
-            Collection<? extends ResourceLock> locks = workerLeaseService.getCurrentProjectLocks();
-            return workerLeaseService.withReplacedLocks(locks, allProjectsLock, () ->
-                factory.apply(new AllProjectsAccessImpl(owner, workerLeaseService, allProjectsLock))
+            Collection<? extends ResourceLock> currentLocks = workerLeaseService.getCurrentProjectLocks();
+            return workerLeaseService.withReplacedLocks(currentLocks, accessLocksOfAllProjects(), () ->
+                factory.apply(new AllProjectsAccessImpl(owner, workerLeaseService))
```
octylFractal, line 272: We need to reword `AllProjectsAccess` javadoc now that there is no all projects lock.

platforms/core-runtime/base-services/src/main/java/org/gradle/internal/resources/ProjectLeaseRegistry.java
jvandort: This was intentional. I generally like to put a blank line after the first line of a class to separate the fields from the class definition line. For symmetry, I also add a line at the end of a class.

octylFractal (approved):

subprojects/core/src/main/java/org/gradle/api/internal/project/DefaultProjectStateRegistry.java
octylFractal: FWIW it still mentions "all projects lock" but I guess that might be OK as this is a bit of an internal detail.

jvandort: Updated to "locks of all projects"

PR #2778, opened 2017-08-23
Add hint about --stacktrace to deprecation warnings

jjohannes:
See: https://github.com/gradle/gradle/issues/899

Issue #2774, opened 2017-08-23
Build zips/jars in parallel


melix:
In order to reduce build times, it is possible to build jars and zips in parallel. A [spike](https://github.com/gradle/gradle/tree/cc/build-cache/parallel-jar) has been created for the build cache team, and shows significant improvements (up to 40% faster) when there are large archives. The spike makes use of the Apache Commons Compress library, and integrates with our worker pool.

There are several issues with the spike that need to be addressed before making it into `master`:

- archives do not have reproducible order, so parallel zip/jar creation is disabled when reproducible order is requested
- jars are treated exactly like zips. But there are slight differences (manifests must come first, magic number). Determine if it is an issue or not, and if yes, fix it.

See gradle/build-cache#765

melix:
This is still relevant

Kukis13:
For what is worth, I developed the plugin that tackles the issue - In some cases it speeds up the `zip` task 9x times. https://github.com/Kukis13/parallel-zip-gradle-plugin

PR #38794, opened 2026-08-11
Fix Groovy 4.0.x max target Java version in test coverage fixture

blindpirate:
Fixes the `AllVersionsIntegMultiVersion` failure in
[Gradle_Master_Check_AllVersionsIntegMultiVersion_33_bucket4 #2778](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_AllVersionsIntegMultiVersion_33_bucket4/116076846):

```
GroovyCompileToolchainIntegrationTest.can compile source and run tests using Java 26 for Groovy  [4.0.29]

Condition not satisfied:

JavaVersion.forClass(groovyClassFile("GroovyBar.class").bytes) == groovyTarget
|           |                                                     |
|           1.8                                                   26
|                                                                 false
```

## Root cause

`GroovyCoverage.maxTargetJavaVersion` claims Groovy 4.x releases before 4.0.33 can target Java 26. They cannot — Groovy only added Java 26/27 to `CompilerConfiguration`'s JDK-to-bytecode map in **4.0.33**. For 4.0.29 a requested target of 26 is silently discarded, and because 26 is also absent from the map, `defaultTargetBytecode()` falls back to `1.8` too — hence major version 52 in the class file.

The bound was correct (`VERSION_25`) until 0ee6eca13cd0 ("Support Java 27"), which bumped it to `VERSION_26` at the same time as introducing the 4.0.33 boundary. The 4.0.33 boundary itself is right; only the pre-4.0.33 value was wrong.

This is why the failure looks flaky: the test is skipped unless the agent actually has a Java 26 JDK installed, so it only fails on the subset of agents that do.

## Verification

Ran `CompilerConfiguration.setTargetBytecode` against the real jars:

| Groovy | request 25 | request 26 | request 27 |
|---|---|---|---|
| 4.0.29 | 25 | **25** | **25** |
| 4.0.32 | 25 | **25** | **25** |
| 4.0.33 | 25 | 26 | 27 |

And confirmed the JDK-to-bytecode map contents directly (`javap` on `CompilerConfiguration`):

- 3.0.25 → max 17 (matches `case 3`)
- 4.0.29 / 4.0.32 → max 25 (**was 26**, now 25)
- 4.0.33 → max 27 (matches)
- 5.0.8 → max 27 (matches `case 5`)
- 6.0.0-alpha-2 → max 27 (matches `case 6`)

So the other branches of the switch are all correct; this changes only the one wrong value.

### DIFF
--- a/platforms/jvm/language-groovy/src/testFixtures/groovy/org/gradle/testing/fixture/GroovyCoverage.groovy
+++ b/platforms/jvm/language-groovy/src/testFixtures/groovy/org/gradle/testing/fixture/GroovyCoverage.groovy
@@ -90,7 +90,8 @@ class GroovyCoverage {
         return switch (groovyVersion.major) {
             case 6 -> JavaVersion.VERSION_27
             case 5 -> JavaVersion.VERSION_27
-            case 4 -> groovyVersion >= VersionNumber.parse('4.0.33') ? JavaVersion.VERSION_27 : JavaVersion.VERSION_26
+            // Groovy 4.0.33 added target support for Java 26 and 27; earlier 4.0.x releases top out at Java 25
+            case 4 -> (groovyVersion >= VersionNumber.parse('4.0.33') ? JavaVersion.VERSION_27 : JavaVersion.VERSION_25)
             case 3 -> JavaVersion.VERSION_17
             default -> throw new IllegalArgumentException("Computing effective target for Groovy version $groovyVersion is not supported")
         }


blindpirate:
Verification branch build: https://builds.gradle.org/build/116115387 (`Gradle_Master_Check_AllVersionsIntegMultiVersion_33_bucket4`, the exact config that has been red since #2774).

blindpirate:
Branch build **green**: https://builds.gradle.org/build/116115387 — `Tests passed: 1284, ignored: 202`, zero failures ([scan](https://ge.gradle.org/s/pf3cmuxluhkys)).

The previously failing case now passes rather than being skipped:

```
✓ GroovyCompileToolchainIntegrationTest.can compile source and run tests using Java 26 for Groovy  4.0.29
✓ ... 4.0.33   ✓ ... 5.0.8   ✓ ... 6.0.0-alpha-2   ○ ... 3.0.25 (skipped, as expected)
```

For reference, master's #2778 on the same config was `Tests failed: 1, passed: 1283, ignored: 202`.

Note: the branch was since force-pushed (`c50a8de` → `1999d88`) to add parentheses around the ternary — no behavioural change from the verified commit.

Issue #38787, opened 2026-08-09
BaseExecSpec streams seem to no longer conform with API on Gradle 9.7.0


chadlwilson:
### Current Behavior

it used to be possible to redirect `standardError` to `standardOutput` in execs, but no longer works:

```groovy
interface Injected {
  @Inject ExecOperations getExecOps()
}

tasks.register('some task') {
  def injected = project.objects.newInstance(Injected)
  doFirst {
    injected.execOps.exec {
      commandLine = ['ls', 'non-existent']
      errorOutput = standardOutput
    }
  }
}
```

Now you get the below:

```
Caused by: java.lang.IllegalArgumentException: outputStream == null!
        at org.gradle.process.internal.ProcessStreamsSpec.setErrorOutput(ProcessStreamsSpec.java:55)
        at org.gradle.process.internal.DefaultExecSpec.setErrorOutput(DefaultExecSpec.java:167)
        at org.gradle.process.internal.DefaultExecSpec_Decorated.setErrorOutput(Unknown Source)
        at org.gradle.process.internal.DefaultExecAction.setErrorOutput(DefaultExecAction.java:240)
        at org.gradle.internal.metaobject.BeanDynamicObject$MetaClassAdapter.setProperty(BeanDynamicObject.java:459)
        at org.gradle.internal.metaobject.BeanDynamicObject$MetaClassAdapter.setProperty(BeanDynamicObject.java:417)
        at org.gradle.internal.metaobject.BeanDynamicObject.trySetProperty(BeanDynamicObject.java:208)
        at org.gradle.internal.metaobject.ConfigureDelegate.setProperty(ConfigureDelegate.java:92)
        at org.gradle.internal.classpath.declarations.GroovyDynamicDispatchInterceptors.callInstrumentedSetProperty(GroovyDynamicDispatchInterceptors.java:101)
        at org.gradle.internal.classpath.declarations.GroovyDynamicDispatchInterceptors.intercept_setProperty(GroovyDynamicDispatchInterceptors.java:88)
```

### Expected Behavior

The documentation says the default is `System.out` and there's no hint that it's lazy, so it seems it should be populated and work as before rather than `null` (or documentation corrected)?

https://docs.gradle.org/9.7.0/javadoc/org/gradle/process/BaseExecSpec.html#getStandardOutput()

> Returns the output stream to consume standard output from the process executing the command. Defaults to System.out.

### Self-contained Reproducer Project

[...]

PR #38799, opened 2026-08-11
Fix BaseExecSpec regression with stdin/stout/sterr redirects

asodja:
Fixes https://github.com/gradle/gradle/issues/38787

https://github.com/gradle/gradle/pull/38799/changes/35cd904794c41315f72c9ddb27f247696bc3026f fixes also a possible race condition where we put the same stream object to different threads and whichever reaches EOF first closes it while the other may still be writing: Replicated with 9.6.0 distribution where 1 in 8 runs logged IOException: flush() failed: stream is closed.

### DIFF
--- a/platforms/core-runtime/process-services-base/src/main/java/org/gradle/process/internal/DefaultClientExecHandleBuilder.java
+++ b/platforms/core-runtime/process-services-base/src/main/java/org/gradle/process/internal/DefaultClientExecHandleBuilder.java
@@ -278,11 +278,19 @@ public class DefaultClientExecHandleBuilder implements ClientExecHandleBuilder, 
         return buildWithEffectiveArguments(argumentsSpec.getAllArguments());
     }
 
+    // One stream for both outputs means the caller wants them merged, so let the process merge them.
+    // Reading them separately would give two threads the same stream, and the one that finishes
+    // first closes it while the other is still writing.
+    private boolean isRedirectingErrorStream() {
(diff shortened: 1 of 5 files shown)


asodja:
@bot-gradle test this

asodja:
@bot-gradle test this

platforms/core-runtime/process-services-base/src/main/java/org/gradle/process/internal/DefaultClientExecHandleBuilder.java
```diff
@@ -278,11 +278,19 @@ public ExecHandle build() {
         return buildWithEffectiveArguments(argumentsSpec.getAllArguments());
     }
 
+    // One stream for both outputs means the caller wants them merged, so let the process merge them.
+    // Reading them separately would give two threads the same stream, and the one that finishes
+    // first closes it while the other is still writing.
+    private boolean isRedirectingErrorStream() {
+        return redirectErrorStream || streamsSpec.getErrorOutput() == streamsSpec.getStandardOutput();
```
abstratt, line 286: 💭 

this won't detect the case where one stream is a wrapper to the other - maybe record it as a known limitation

abstratt (approved):
LGTM, one nit

platforms/core-runtime/process-services-base/src/main/java/org/gradle/process/internal/DefaultClientExecHandleBuilder.java
asodja: Makes sense, added a comment

PR #38810, opened 2026-08-12
Keep type-use annotations in extracted ABI classes

### LINKED ISSUE #38792: Gradle 9.7.0 breaks a working Transformer implementation
### Current Behavior

I have this task working fine in Gradle 9.6.1:

```
private class LibRenameTransformer(
    private val foo: Boolean,
) : Transformer<String?, String>, Serializable {
    override fun transform(name: String): String? = when(foo) {
        true -> name.replace("foo", "bar")
        else -> name
    }
}
```

Usage is here in a copy task:

```
val copyStuff = tasks.register<Copy>("copyStuff") {
    group = "build"
    from(layout.projectDirectory.dir("lib")) { include("*.jar", "*.so", "*.dll", "*.dylib") }
    into(layout.buildDirectory.dir("new-lib"))
    rename(LibRenameTransformer(foo = true))
}
```

When updating Gradle to 9.7.0 I get a strange error on the `LibRenameTransformer` declaration:
`Type argument is not within its bounds: type parameter 'OUT (of interface Transformer<OUT : Any, IN : Any>' must be subtype of 'Any', but actual 'String?'`

Fixing this problem by making it `<String, String>` does not help, because the `rename` call does not take this anymore.

### Expected Behavior

Should not be a problem at all!

### Context (optional)

_No response_

Just copy the transformer class as defined into an empty Gradle project. Thanks!

### Gradle version

9.7.0

### Build scan URL (optional)

_No response_

_No response_

eskatos:
* Follows up
    * #35658
    * #33846
* Fixes https://github.com/gradle/gradle/issues/38792

### Why

Gradle 9.7.0 compiles Kotlin DSL build scripts against a new ABI stub jar (`gradle-public-api-legacy`). That stub jar lost the type-use annotations of every class and every field. Kotlin therefore read the bound of `Transformer<OUT extends @Nullable Object, IN>` as non-null. A build script that declares `Transformer<String?, String>` failed with this error:

```
Type argument is not within its bounds: type parameter 'OUT (of interface Transformer<OUT : Any, IN : Any>' must be subtype of 'Any', but actual 'String?'
```

The real class keeps the annotation. Only the stub jar lost it, and the Kotlin compiler is correct about the stub it reads.

### What changes

- The API extractor keeps the type-use annotations of classes and fields. Before, it kept only the annotations of methods.
- The API extractor keeps two type-use annotations that have the same name on one member. Before, the second annotation replaced the first. For this reason the 9.7.0 stub of `DependencyFactory.create` kept one `@Nullable` mark of four.
- The build extracts the stubs with the API extractor of this source tree. Before, it used the extractor of the Gradle distribution that runs the build.

### Implementation

- `ApiMemberSelector` implements `visitTypeAnnotation` for the class visitor and for the field visitor. It implemented the method visitor only.
- `AnnotatableMember` holds the type-use annotations, so classes, fields and methods share one collection.
- `TypeAnnotationMember` compares `typeRef` and `typePath`. The name and the visibility gave an annotation its identity before, and the sorted set kept one annotation per name.
- `IsolatedApiClassExtractor` loads the extractor from `extractorClasspath` under the platform class loader. The package `org.gradle.internal.tools.api` resolves parent-first, and the class loader reads the distribution first. Only JDK types cross this boundary, therefore the calls are reflective.
- A check in the constructor stops the build when the extractor comes from another class loader. A broken isolation writes wrong output and reports no error, which is why the original `FIXME` survived a release.
- The worker process stays. It keeps extraction failures and heap out of the build process.

### How to review

1. Read `IsolatedApiClassExtractor` first. Without it, the two extractor fixes do not reach the stub jar.
2. To see the fault, get `lib/api/gradle-public-api-legacy-9.7.0.jar` from a 9.7.0 distribution. Run `javap -v -p org/gradle/api/Transformer.class` on it. The attribute `RuntimeVisibleTypeAnnotations` is absent. The same class in `gradle-base-services-9.7.0.jar` has it.
3. To prove that the isolation is necessary, replace `ClassLoader.getPlatformClassLoader()` with `IsolatedApiClassExtractor::class.java.classLoader`. Then run the integration test. It fails with the error of the issue.

Type-use annotations in the stub jar of the core distribution, for the 11628 classes that the 9.7.0 jar and this jar share:

| Target | 9.7.0 | This PR |
|---|---|---|
| `CLASS_TYPE_PARAMETER_BOUND` | 0 | 18 |
| `FIELD` | 0 | 27 |
| `METHOD_FORMAL_PARAMETER` | 1481 | 1863 |
| `METHOD_RETURN` | 1872 | 1872 |
| `METHOD_TYPE_PARAMETER_BOUND` | 49 | 50 |

No count decreases, so the stubs lose nothing. `METHOD_RETURN` does not change, so no public method gets a nullable return type. Only parameters and bounds accept more.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ContentFilter.kt
+++ b/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ContentFilter.kt
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package gradlebuild.packaging.tasks
+
+import java.io.File
+
+
+private
+val KOTLIN_MODULE_PATH = Regex("META-INF/.*\\.kotlin_module")
+
+
+internal
+enum class ContentFilter {
+    VERBATIM,
+    API_ONLY,
+    SKIP
+}
+
+
+/**
+ * What to do with [file] of a classes directory [classDir].
+ */
+internal
+fun contentFilterFor(classDir: File, file: File): ContentFilter =
+    contentFilterFor(file.relativeTo(classDir).path)
+
+
+/**
+ * What to do with the entry at [relativePath] of a classes directory.
+ *
+ * The path takes the separator of either operating system. The rules below need `/`, and
+ * a relative path on Windows comes with `\`.
+ */
+internal
+fun contentFilterFor(relativePath: String): ContentFilter {
+    val path = relativePath.replace('\\', '/')
+    // Extraction keeps the module and the package annotations, so module-info and
+    // package-info need no copy of their own
+    if (path.endsWith(".class")) {
+        return ContentFilter.API_ONLY
+    }
+    if (path == "META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule") {
+        return ContentFilter.VERBATIM
+    }
+    if (path == "META-INF/services/org.codehaus.groovy.transform.ASTTransformation") {
+        return ContentFilter.VERBATIM
+    }
+    if (path.matches(KOTLIN_MODULE_PATH)) {
+        return ContentFilter.VERBATIM
+    }
+    return ContentFilter.SKIP
+}
--- a/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ContentFilter.kt
+++ b/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ContentFilter.kt
@@ -36,8 +36,8 @@ enum class ContentFilter {
  */
 internal
 fun contentFilterFor(relativePath: String): ContentFilter {
-    // Extraction keeps the module and the package annotations, so module-info and
-    // package-info need no copy of their own
+    // The extractor includes the package-private members, so it keeps module-info and
+    // package-info with the module directives and the package annotations of those files
     if (relativePath.endsWith(".class")) {
         return ContentFilter.API_ONLY
     }
--- a/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ContentFilter.kt
+++ b/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ContentFilter.kt
@@ -16,8 +16,6 @@
 
 package gradlebuild.packaging.tasks
 
-import java.io.File
-
 
 private
 val KOTLIN_MODULE_PATH = Regex("META-INF/.*\\.kotlin_module")
@@ -32,34 +30,24 @@ enum class ContentFilter {
 
 
 /**
- * What to do with [file] of a classes directory [classDir].
- */
-internal
-fun contentFilterFor(classDir: File, file: File): ContentFilter =
-    contentFilterFor(file.relativeTo(classDir).path)
-
-
-/**
  * What to do with the entry at [relativePath] of a classes directory.
  *
- * The path takes the separator of either operating system. The rules below need `/`, and
- * a relative path on Windows comes with `\`.
+ * The caller gives a path with `/` as separator on every operating system.
  */
 internal
 fun contentFilterFor(relativePath: String): ContentFilter {
-    val path = relativePath.replace('\\', '/')
     // Extraction keeps the module and the package annotations, so module-info and
     // package-info need no copy of their own
-    if (path.endsWith(".class")) {
+    if (relativePath.endsWith(".class")) {
         return ContentFilter.API_ONLY
     }
-    if (path == "META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule") {
+    if (relativePath == "META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule") {
         return ContentFilter.VERBATIM
     }
-    if (path == "META-INF/services/org.codehaus.groovy.transform.ASTTransformation") {
+    if (relativePath == "META-INF/services/org.codehaus.groovy.transform.ASTTransformation") {
         return ContentFilter.VERBATIM
     }
-    if (path.matches(KOTLIN_MODULE_PATH)) {
+    if (relativePath.matches(KOTLIN_MODULE_PATH)) {
         return ContentFilter.VERBATIM
     }
     return ContentFilter.SKIP
--- a/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
+++ b/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
@@ -19,10 +19,8 @@ package gradlebuild.packaging.tasks
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.DirectoryProperty
-import org.gradle.api.provider.SetProperty
 import org.gradle.api.tasks.CacheableTask
 import org.gradle.api.tasks.Classpath
-import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.OutputDirectory
 import org.gradle.api.tasks.TaskAction
 import org.gradle.workers.WorkAction
@@ -36,16 +34,13 @@ import javax.inject.Inject
  *
  * Keeps only the following:
  *
- * -  API stubs of public classes in the specified packages
+ * - API stubs of the classes, including the package-private ones
  * - `META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule`
  * - `META-INF/services/org.codehaus.groovy.transform.ASTTransformation`
  * - `META-INF/\*.kotlin_module`
  */
 @CacheableTask
 abstract class ExtractJavaAbi : DefaultTask() {
-
-    @get:Input
-    abstract val packages: SetProperty<String>
 
     // Do not change this to CompileClasspath. It hashes the input with the ABI extractor of the
     // distribution that runs this build, and that extractor drops the details this task keeps.
@@ -69,7 +64,6 @@ abstract class ExtractJavaAbi : DefaultTask() {
         // The worker process keeps extraction failures and heap out of the build process,
         // IsolatedApiClassExtractor keeps the extractor away from the distribution running this build
         workerExecutor.processIsolation().submit(ExtractJavaAbiAction::class.java) {
-            packages.set(task.packages)
             classesDirectories.setFrom(task.classesDirectories)
             outputDirectory.set(task.outputDirectory)
             extractorClasspath.setFrom(task.extractorClasspath)
@@ -79,7 +73,6 @@ abstract class ExtractJavaAbi : DefaultTask() {
     abstract class ExtractJavaAbiAction @Inject constructor() : WorkAction<ExtractJavaAbiAction.Params> {
 
         interface Params : WorkParameters {
-            val packages: SetProperty<String>
             val classesDirectories: ConfigurableFileCollection
             val outputDirectory: DirectoryProperty
             val extractorClasspath: ConfigurableFileCollection
@@ -87,7 +80,7 @@ abstract class ExtractJavaAbi : DefaultTask() {
 
         override fun execute() {
             val outputDirectory = parameters.outputDirectory.get().asFile
-            IsolatedApiClassExtractor.runUsing(parameters.extractorClasspath, parameters.packages.get()) { extractor ->
+            IsolatedApiClassExtractor.runUsing(parameters.extractorClasspath) { extractor ->
                 parameters.classesDirectories.forEach { classDir ->
                     classDir.walk().forEach { inputFile ->
                         val relativePath = inputFile.relativeTo(classDir).invariantSeparatorsPath
--- a/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
+++ b/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
@@ -85,7 +85,7 @@ abstract class ExtractJavaAbi : DefaultTask() {
 
         override fun execute() {
             val outputDirectory = parameters.outputDirectory.get().asFile
-            IsolatedApiClassExtractor(parameters.extractorClasspath, parameters.packages.get()).use { extractor ->
+            IsolatedApiClassExtractor.runUsing(parameters.extractorClasspath, parameters.packages.get()) { extractor ->
                 parameters.classesDirectories.forEach { classDir ->
                     classDir.walk().forEach { inputFile ->
                         val outputFile = outputDirectory.resolve(inputFile.relativeTo(classDir).path)
--- a/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
+++ b/build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
@@ -92,10 +92,10 @@ abstract class ExtractJavaAbi : DefaultTask() {
                 parameters.classesDirectories.forEach { classDir ->
                     classDir.walk().forEach { inputFile ->
                         val outputFile = outputDirectory.resolve(inputFile.relativeTo(classDir).path)
-                        when (inputFile.filtering()) {
+                        when (contentFilterFor(classDir, inputFile)) {
                             ContentFilter.VERBATIM -> {
                                 outputFile.parentFile.mkdirs()
-                                inputFile.copyTo(outputFile)
+                                inputFile.copyTo(outputFile, overwrite = true)
                             }
 
                             ContentFilter.API_ONLY -> {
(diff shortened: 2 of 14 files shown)


eskatos:
@bot-gradle test this

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
jbartok: There is a bootstrapping problem here: @CompileClasspath tells Gradle to fingerprint the task's input using ABI hashing, which is done by the old, buggy extractor. That normalization code runs inside the Gradle distribution executing the build, i.e. the wrapper version gradle/gradle builds itself with, not the code in the PR. And the wrapper's copy of the extractor still has the bug: it drops type-use annotations when computing the ABI. 

Maybe use `@get:Classpath` instead and then put back `@get:CompileClasspath` when the fix is in the wrapper?

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
```diff
@@ -85,52 +80,38 @@ abstract class ExtractJavaAbi : DefaultTask() {
...
             }
         }
 
-        private
-        fun ApiClassExtractor.extractApiClassFrom(input: InputStream) =
-            extractApiClassFrom(input.readAllBytes())
-
         private
         fun File.filtering(): ContentFilter {
             if (name.endsWith(".class")) {
```
jbartok: This is incorrect. "name" is used (which is just the last path segment), but it's being worked with as if it would be the full path.

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
```diff
@@ -85,52 +80,38 @@ abstract class ExtractJavaAbi : DefaultTask() {
...
-                    when (inputClassFile.filtering()) {
-                        ContentFilter.VERBATIM -> {
-                            outputClassFile.parentFile.mkdirs()
-                            inputClassFile.copyTo(outputClassFile)
-                        }
+            val outputDirectory = parameters.outputDirectory.get().asFile
+            IsolatedApiClassExtractor(parameters.extractorClasspath, parameters.packages.get()).use { extractor ->
+                parameters.classesDirectories.forEach { classDir ->
+                    classDir.walk().forEach { inputFile ->
+                        val outputFile = outputDirectory.resolve(inputFile.relativeTo(classDir).path)
```
jbartok: The filtering() extension function is wrong, see my comment on it. 
Pass this relative path to it, make it work on that.

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
jbartok: Consider using normalized file separators, so it works on Windows too.

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
```diff
@@ -85,52 +80,38 @@ abstract class ExtractJavaAbi : DefaultTask() {
...
+                        when (inputFile.filtering()) {
+                            ContentFilter.VERBATIM -> {
+                                outputFile.parentFile.mkdirs()
+                                inputFile.copyTo(outputFile)
```
jbartok: This defaults to overwrite=false, is that ok?

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
```diff
@@ -85,52 +80,38 @@ abstract class ExtractJavaAbi : DefaultTask() {
...
-                classDir.walk().forEach { inputClassFile ->
-                    val relativePath = inputClassFile.relativeTo(classDir).path
-                    val outputClassFile = parameters.outputDirectory.get().asFile.resolve(relativePath)
-                    when (inputClassFile.filtering()) {
-                        ContentFilter.VERBATIM -> {
-                            outputClassFile.parentFile.mkdirs()
-                            inputClassFile.copyTo(outputClassFile)
-                        }
+            val outputDirectory = parameters.outputDirectory.get().asFile
+            IsolatedApiClassExtractor(parameters.extractorClasspath, parameters.packages.get()).use { extractor ->
```
jbartok: Efficiency: this could be reused while the daemon is alive.

jbartok: Possible resource leak: `use` will close it only after it has been constructed. If the construction of `IsolatedApiClassExtractor` fails, it will never be closed.

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
jbartok: Also: due to how `AbiExtractingClasspathResourceHasher.appendConfigurationToHasher` works, cached ABI resource hashes computed before this fix will get reused.

platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/ApiMemberSelector.java
```diff
@@ -113,7 +120,7 @@ public AnnotationVisitor visitAnnotation(String desc, boolean visible) {
                 }
```
jbartok, line 121: visitRecordComponent omitted intentionally, right?

jbartok (changes requested):

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
eskatos: It could. I tried but measures said it is not worth the complexity of caching something that can hold file descriptors, memory etc... and the invalidation is tricky. Also, the overhead is minimal.

platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/ApiMemberSelector.java
eskatos: Yes, this is another issue, see

* https://github.com/gradle/gradle/issues/38823

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
eskatos: Good catch, fixed in af049bb4efb733f0c64a03fefa157b9088bdaef6

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
eskatos: Good catch, and it reproduces. On this branch, with `@CompileClasspath`, I added a
second bound annotation to `Transformer`:

```java
// before
public interface Transformer<OUT extends @Nullable Object, IN>
// after
public interface Transformer<OUT extends @Nullable Object, IN extends @Nullable Object>
```

```
> Task :base-services:compileJava        (re-ran)
> Task :base-services:extractJavaAbi UP-TO-DATE
real class: 2 CLASS_TYPE_PARAMETER_BOUND      generated ABI: 1
```

The task keeps a stale stub. With `@Classpath` the same change re-runs the task and the
stub gets both annotations. Switched to `@Classpath`.

One refinement on the plan. I suggest we keep `@Classpath` also after the fix reaches
the wrapper, rather than a temporary switch, for two reasons.

The gap is wider than type-use annotations. ABI normalization hides every detail that
the extractor of the running distribution drops. My audit of the extractor found four
more such details, all reported:

* record components
   #38823
* method parameter names
   #38827
* the `value` member of an annotation with an array member
   https://github.com/gradle/gradle/issues/38828
* `THROWS` type annotations
   #38825

Each fix re-opens the same hole for the next release. A task that writes ABI details
must not take its own input through an older copy of itself.

The cost is small. Measurements on this branch:

* Fingerprinting reads the same bytes either way. ABI normalization reads the full
  class file and then runs the extractor over it, so `@Classpath` does less work, not
  more. Both are cached, so an unchanged file is not read again.
* Re-execution of the biggest module, `dependency-management` with 1873 classes and
  11.4 MB, takes 300 ms to 400 ms. I measured 1954 ms and 2038 ms against an
  up-to-date baseline of 1644 ms for the same command.
* All 147 modules together extract in 8 s to 9 s cold.
* `@Classpath` hashes content, so a recompilation with equal bytes changes nothing.
* No compilation depends on the stubs. `publicApiAbiStubs` in `PublicApiJars.kt` is the
  only consumer, for the ABI jar and its publication, so a re-run reaches the jar tasks
  and stops there with equal content.

On your second point about `AbiExtractingClasspathResourceHasher.appendConfigurationToHasher`:
the resource hash caches are version scoped. `DefaultCacheScopeMapping.getCacheDir` uses
`version.getVersion()`, and this build tree holds `.gradle/9.7.0/fileHashes`. A different
Gradle version therefore starts with an empty cache. Two distributions with one version
string and different extractors still share entries, which a local build gives you, so
the point stands there. With `@Classpath` this task no longer uses ABI normalization for
its input, so the concern no longer applies to it.

One more note for the release. The 472 type annotations that this PR adds to the stub jar
do arrive in 9.7.1 even with the old annotation, because `extractorClasspath` is
`@Classpath` and the extractor jar changes here. The stale path only affects annotations
that somebody adds after the release.

eskatos: Fixed in 6831fa1cf50bce5cdaa4597da9846abae5565254

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
eskatos: Correct, and it never worked.

The rules come from CopyPublicApiClassesTransform, which classified ZipEntry objects. The name of a zip entry is the full entry path. Commit 8ad1bbe9526 moved the same expressions onto File.name, which holds only the last segment, so every rule with a path stopped to match. The task copied none of the META-INF entries that its own KDoc promises.

The rules now take the relative path of the entry.

I removed the two rules for module-info.class and package-info.class instead of a repair. Extraction already handles both files. The extracted module-info keeps its Module attribute with requires and exports, and package-info keeps its annotations. A verbatim copy puts SourceFile back into 397 package-info entries of a jar that holds normalized signatures, which is a step backwards.

The classification moved to ContentFilter.kt with unit tests, because the fault was the input of the rules and not the rules. A test of the rules alone stays green while the task is broken.

The stub jar of the distribution comes out byte identical after the change, so the behaviour of the release stays the same.

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
eskatos: Done, the rules now get file.relativeTo(classDir).path.

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
eskatos: Done. contentFilterFor replaces \ with / before it applies the rules, so the result is the same on every operating system.

A unit test passes a path with backslashes, so the Windows behaviour of the rules is under test on Linux and macOS too. Thanks for catching it.

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
eskatos: No it is not. Fixed

eskatos:
@bot-gradle test this

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ContentFilter.kt
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * What to do with the entry at [relativePath] of a classes directory.
+ *
+ * The path takes the separator of either operating system. The rules below need `/`, and
+ * a relative path on Windows comes with `\`.
+ */
+internal
+fun contentFilterFor(relativePath: String): ContentFilter {
+    val path = relativePath.replace('\\', '/')
+    // Extraction keeps the module and the package annotations, so module-info and
```
jbartok: The claim "extraction keeps them" depends entirely on which of the extractor's two modes is active, and that's controlled by the task's `packages` input property (ExtractJavaAbi builds the extractor as either `includePackagePrivateMembers()` when packages is empty, or `includePackagesMatching(...)` when it's not). 

This works now only because the `packages` property is not being set, would break if it would be set.

To fix it `module-info.class` (has no package) and `package-info.class` (isn't public) should be special cased to VERBATIM copy.

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
```diff
@@ -85,84 +83,36 @@ abstract class ExtractJavaAbi : DefaultTask() {
...
-                    when (inputClassFile.filtering()) {
-                        ContentFilter.VERBATIM -> {
-                            outputClassFile.parentFile.mkdirs()
-                            inputClassFile.copyTo(outputClassFile)
-                        }
+            val outputDirectory = parameters.outputDirectory.get().asFile
+            IsolatedApiClassExtractor.runUsing(parameters.extractorClasspath, parameters.packages.get()) { extractor ->
+                parameters.classesDirectories.forEach { classDir ->
+                    classDir.walk().forEach { inputFile ->
+                        val outputFile = outputDirectory.resolve(inputFile.relativeTo(classDir).path)
```
jbartok: `inputFile.relativeTo` computed twice (other one is inside `contentFilterFor`)

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ContentFilter.kt
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * What to do with the entry at [relativePath] of a classes directory.
+ *
+ * The path takes the separator of either operating system. The rules below need `/`, and
+ * a relative path on Windows comes with `\`.
+ */
+internal
+fun contentFilterFor(relativePath: String): ContentFilter {
+    val path = relativePath.replace('\\', '/')
```
jbartok: Use `File.invariantSeparatorsPath` instead?

platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/TypeAnnotationMember.java
```diff
@@ -32,7 +34,26 @@ public int getTypeRef() {
...
+     * The path to the annotated type, or {@code null} when the annotated type is the whole type.
+     */
+    @Nullable
     public TypePath getTypePath() {
         return typePath;
     }
+
+    @Override
+    public int compareTo(AnnotationMember o) {
+        TypeAnnotationMember other = (TypeAnnotationMember) o;
```
jbartok, line 56: This cast is meh, makes comparison assymetric:
* plainAnnotation.compareTo(typeAnnotation) dispatches to the base implementation, which happily compares descriptor and visibility. It succeeds — and returns 0 if the descriptors match.
* typeAnnotation.compareTo(plainAnnotation) hits the cast and throws ClassCastException.

Maybe this would be better, although not perfect:

```
if (!(o instanceof TypeAnnotationMember)) {
    int result = super.compare(o).result();
    return result != 0 ? result : 1;  // type annotations sort after plain ones
}
```

jbartok (changes requested):

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ContentFilter.kt
eskatos: Done. Better than my replace, and it belongs at the boundary rather than inside the rules.

The rules now document that they take /. I dropped the test with backslashes, because that conversion is a no-op outside Windows and the test checked my own normalization, which is gone.

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ExtractJavaAbi.kt
eskatos: Done, one relativeTo now feeds both the output path and the rules.

This also makes the original bug harder to write again. Both uses read the same value, so a name instead of a path lands the output in the wrong place, which is loud, rather than only misclassifying an entry, which is silent.

platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/TypeAnnotationMember.java
eskatos: Your suggestion removes the exception without fixing the asymmetry violating the `Comparable` contract.
This issue predates this PR and is present in other related types.

A full fix requires the parent class to collaborate. It's quite cheap so I did it in this PR.

build-logic/packaging/src/main/kotlin/gradlebuild/packaging/tasks/ContentFilter.kt
eskatos: I took another fix. Nothing sets packages. The registration in gradlebuild.distribution-module sets the classes directories, the output directory and the extractor classpath, and that is the only one. So the property and the second mode of the extractor are gone, and the extractor always includes the package-private members. The comment now gives that reason instead of a claim for one mode.

jbartok (approved):
Looks good!

PR #38811, opened 2026-08-12
Refactor DefaultConfigurationCache

ov7a:
1. Extract ConfigurationCacheCandidateEntries and ConfigurationCacheEntrySelector
2. simplify ConfigurationCacheClassLoaderScopeRegistryListener
3. split loading and scheduling tasks
4. simplify fingerprint handling

Extracted out of
* #38458 

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ov7a:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/internal/build/BuildStateRegistry.java
```diff
@@ -133,4 +133,9 @@ default void onRootBuildInclude(RootBuildState rootBuild, BuildState referrer, b
      * Visits all registered builds, ordered by {@link BuildState#getIdentityPath()}
      */
     void visitBuilds(Consumer<? super BuildState> visitor);
+
+    /**
+     * Discards the model of every registered build, so that the build tree can be configured again.
+     */
+    void resetModels();
```
bamboo, line 140: 👍

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultConfigurationCache.kt
```diff
@@ -195,29 +203,36 @@ class DefaultConfigurationCache internal constructor(
...
+    fun beginEntry(action: ConfigurationCacheAction, entryId: String, actionDescription: StructuredMessage) {
+        this.cacheAction = action
+        this.entryId = entryId
+        initializeCacheEntrySideEffects(action)
+        problems.action(action, actionDescription)
     }
 
     private
     fun initializeCacheEntrySideEffects(cacheAction: ConfigurationCacheAction) {
+        if (cacheAction.isReadOnly) {
```
bamboo, line 231: 🤔 Is this behavior new?

bamboo (commented):
Overall the direction looks good. I'll need to give it another pass to understand what happened to the subCodecs transitions.

platforms/core-configuration/configuration-cache/src/main/kotlin/org/gradle/internal/cc/impl/DefaultConfigurationCache.kt
ov7a: If you mean changes around `scopeRegistryListener`, then no, it's my attempt to reduce the `scopeRegistryListener.dispose()` calls to address this todo:
```
            // TODO:configuration-cache find a way to make `dispose` unnecessary;
            //  maybe by extracting an `ConfigurationCacheBuildDefinition` service
            //  from DefaultConfigurationCacheHost so a decision based on the configured
            //  configuration cache strategy (none, store or load) can be taken early on.
            //  The listener only needs to be attached in the `store` state.
```

alllex (changes requested):
Thanks for pulling this out of #38458 — the changes are useful and we want them in. But as one commit it's hard to review: four independent, non-trivial changes, three touching semantics. No behaviour change can be reverted without the mechanical work.

I propose closing this PR after splitting it into multiple, based on the concerns and risks.

### Some concerns

- ❓ **The `Try.ofFailable` boundary moves both ways.** It gains `finalizeCacheEntry`/`resetModels`/`loadRequestedTasks` (good — a throwable escaping there skips `reportProblems(...)`), but `loadAndRun` falls outside, so a corrupt entry propagates instead of becoming `ExecutionResult.failed`. In #38458 that hoist is safe because `LoadOutcome` carries the failure and the controller recovers. Keep the widening, move `loadAndRun` back in.
- ❓ **`start`/`stopRecording()` narrows the recording window** — `afterStart()` → the `DefaultRootBuildState` constructor, clearing state first. The gap isn't empty: instrumenting showed the buildSrc scope chain landing in it. I couldn't make a build fail, but the invariant isn't asserted and `assertNotDisposed` went in the same commit.
- 🤔 `runCatching` catches `Throwable`, so rollback now runs on `Error`/cancellation too. Intended?
- 💭 The entries-per-key LRU still has no test, and is now harder to reach.

`readFingerprintFrom` is conceptual, not practical: it puts `ConfigurationCacheFingerprintController.Host` on a general serialization interface. The isolate and host belong on the fingerprint side, as an extension.

`subCodecs` is half-done: there's no sub-codecs concept — it's the destructured second element, the same instance the field holds, so a no-op that reads like a semantic change. And `ChildContextSource`'s read/write sides mirror each other; only read changed. If the `Pair` goes, both should. We should simplify this eventually, but do it more thoroughly.

### What we'll do

In order to even be able to review this, I had to split it multiple commits. This eventually helped me to split it into multiple PRs.

- #39017 — the mostly mechanical parts. Merge first.
- #39027 — the load/schedule split, so the `Try` boundary gets settled with tests.
- #39025 — `start`/`stopRecording()`, pending an answer on the window.

The latter two need tests first; #38458's `ConfigurationCacheCorruptionRecoveryIntegrationTest` covers that load path and may be worth pulling forward.

ov7a:
Thanks, @alllex! I like you changes and I think you should proceed with merging them. 
The subcodes part can be dropped for now, indeed.

I'll close this PR and will rebase https://github.com/gradle/gradle/pull/38458 after your PRs chain is merged.

ov7a:
Superceeded by the refactor chain above

PR #38813, opened 2026-08-12
Fix the Kotlin version warning of embedded-kotlin

### LINKED ISSUE #38809: `embedded-kotlin` warns on the Kotlin Gradle Plugin version, not on the Kotlin compiler version
### Current Behavior

The `embedded-kotlin` plugin compares the version of the applied Kotlin Gradle Plugin with the Kotlin version embedded in Gradle. When the two versions differ, the plugin logs `WARNING: Unsupported Kotlin plugin version.`. The `kotlin-dsl` plugin applies `embedded-kotlin`, so build logic projects also get this warning.

The comparison reads `getKotlinPluginVersion()` in [`EmbeddedKotlinPlugin.kt:45`](https://github.com/gradle/gradle/blob/b700dade3bfd82368004732e0690d98c077324bc/platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt#L45). This function returns the version of the Kotlin Gradle Plugin. It does not return the version of the Kotlin compiler that compiles the code. A build selects that compiler with the [`kotlinExtension.compilerVersion`](https://kotlinlang.org/api/kotlin-gradle-plugin/kotlin-gradle-plugin-api/org.jetbrains.kotlin.gradle.dsl/-kotlin-top-level-extension/compiler-version.html) property.

As a result, the warning is wrong in both directions. The attached reproducer shows both, with Gradle 9.7.0 and its embedded Kotlin 2.4.0.

**Case A: no warning, and the build fails.** The applied Kotlin Gradle Plugin is 2.4.0, and the compiler is 2.3.21:

```
--- case-a ---
Kotlin embedded in Gradle : 2.4.0
Applied Kotlin plugin     : 2.4.0
Kotlin compiler           : 2.3.21

e: java.lang.NoSuchFieldError: Class org.jetbrains.kotlin.config.CompilerConfigurationKey does not have member field 'org.jetbrains.kotlin.config.CompilerConfigurationKey$Companion Companion'

* What went wrong:
Execution failed for task ':compilePluginsBlocks' (registered by plugin class 'org.gradle.kotlin.dsl.plugins.precompiled.PrecompiledScriptPlugins').
> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.btapi.BuildToolsApiCompilationWork
   > Compilation error. See log for more details
```

Gradle logs no warning. The compiler plugins of the `kotlin-dsl` plugin are version 2.4.0, and the Kotlin 2.3.21 compiler cannot load them. This is the exact risk that the warning must report. The same build without a `compilerVersion` value compiles without an error.

**Case B: a warning, and the build works.** The applied Kotlin Gradle Plugin is 2.4.10, and the compiler is 2.4.0:

```
WARNING: Unsupported Kotlin plugin version.
--- case-b ---
Kotlin embedded in Gradle : 2.4.0
Applied Kotlin plugin     : 2.4.10
Kotlin compiler           : 2.4.0

BUILD SUCCESSFUL
```

Gradle logs the warning, but the compiler is the one that Gradle expects. The warning is noise.

### Expected Behavior

The warning informs the user about one risk: the code compiles with a Kotlin compiler that differs from the compiler that Gradle expects. `compilerVersion` is the source of truth for that compiler version. Gradle compares `kotlinExtension.compilerVersion` with `embeddedKotlinVersion`, and logs the warning only when these two values differ.

Points to decide:

- `compilerVersion` is experimental in Kotlin Gradle Plugin 2.4.x. It needs two opt-in annotations: `@ExperimentalKotlinGradlePluginApi` and `@ExperimentalBuildToolsApi`. Gradle must accept this dependency on an experimental API.
- `compilerVersion` is a `Property<String>`. An eager read finalizes the value and stops later configuration of the property. The plugin does this comparison at apply time, so the new comparison must read the value lazily.
- The Kotlin documentation says that a `compilerVersion` value only has an effect when the Gradle property `kotlin.compiler.runViaBuildToolsApi` is `true`. The default value of the property is the version of the Kotlin Gradle Plugin, so builds without this property keep the current behavior.
- The property exists since Kotlin Gradle Plugin 2.1.20. A build can apply an older version, where this property is absent. Then the plugin must use `getKotlinPluginVersion()` as a default value, or Gradle must raise the minimum supported version.

### Context (optional)

This comes from a direct report by JetBrains on 2026-08-10. Users see the warning in projects that combine build logic with a recent version of the Kotlin Gradle Plugin. The warning is noise when the compiler version matches the Kotlin version embedded in Gradle.

The attached archive `kotlin-dsl-compiler-version-warning.zip` contains a Gradle wrapper for 9.7.0, a `README.md`, and two independent builds.

[kotlin-dsl-compiler-version-warning.zip](https://github.com/user-attachments/files/30975210/kotlin-dsl-compiler-version-warning.zip)

Run the two cases:

```
./gradlew -p case-a build
./gradlew -p case-b build
```

Case A fails and logs no warning. Case B succeeds and logs the warning.

Both builds set `kotlin.compiler.runViaBuildToolsApi=true` in `gradle.properties`, because `compilerVersion` only has an effect with this Gradle property. The build script of case A is:

```kotlin
import org.jetbrains.kotlin.buildtools.api.ExperimentalBuildToolsApi
import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi

plugins {
    `kotlin-dsl`
}

repositories {
    mavenCentral()
}

// Compile with a Kotlin compiler that differs from the Kotlin version embedded in Gradle
@OptIn(ExperimentalKotlinGradlePluginApi::class, ExperimentalBuildToolsApi::class)
fun useOldCompiler() = kotlin {
    compilerVersion.set("2.3.21")
}
useOldCompiler()
```

Case B is the same, with `kotlin("jvm") version "2.4.10"` added to the `plugins` block, and with `compilerVersion` set to `embeddedKotlinVersion`.

### Gradle version

9.7.0. The check is also unchanged on `master` (9.8.0, embedded Kotlin 2.4.10), and it exists in earlier releases too.

### Build scan URL (optional)

(none)

_No response_

eskatos:
* Fixes #38809

### Why

The `embedded-kotlin` plugin warns when the Kotlin version of a build differs from the Kotlin version
embedded in Gradle. The check reads the version of the applied Kotlin Gradle Plugin. That version is
not the version of the compiler that compiles the code, because a build selects the compiler with
`kotlinExtension.compilerVersion`.

The warning is wrong in both directions:

- Gradle logs no warning when the compiler version differs. The build then fails with a
  `NoSuchFieldError` from the compiler, and the user gets no explanation.
- Gradle logs the warning when the compiler version is the embedded one. The message is noise. This
  is the case that JetBrains reported.

### What changes

- The check compares `kotlinExtension.compilerVersion` with `embeddedKotlinVersion`.
- The first line of the message reads `WARNING: Unsupported Kotlin compiler version.`, and the text
  names the compiler version.
- The warning comes from the Kotlin compile tasks. A build that configures no Kotlin compile task
  gets no warning.

### Implementation

- A read of `compilerVersion` finalizes the value. A read at plugin apply time breaks every build
  that sets the property, so the check runs at the configuration of the compile tasks.
- A project can have several Kotlin compile tasks. A flag keeps the warning to one message.
- The code has no version check for old plugins. Gradle supports Kotlin Gradle Plugin 2.0.0 and
  newer, and every version in that range has the property.
- `compilerVersion` needs two opt-in annotations, so `kotlin-build-tools-api` gets a `compileOnly`
  declaration.

### How to review

1. Read `EmbeddedKotlinPlugin.kt`. The whole behavior change is there.
2. Read the two tests in `EmbeddedKotlinPluginIntegTest`. The first test covers the missed warning.
   The second test covers the false warning.
3. The second test forces an older Kotlin Gradle Plugin version with a `resolutionStrategy` rule. A
   `plugins` block cannot lower that version, because the distribution puts the embedded version on
   the same classpath and conflict resolution selects the higher one. The rule must cover every
   module with the `kotlin-gradle-plugin` prefix, or the plugin apply fails.
4. Both tests run `classes` on a project without Kotlin sources. The compile task is then
   `NO-SOURCE`, so the build needs no compiler artifact.

To see the change in a real build, publish the plugins and use the attached reproducer of the issue:

    ./gradlew :kotlin-dsl-plugins:publishToMavenLocal
    ./gradlew :distributions-full:install -Pgradle_installPath=/tmp/gradle-9.8-fix

The `kotlin-dsl` plugins are not part of a distribution, so the reproducer needs `mavenLocal()` in
its `pluginManagement` repositories and a rule that selects the version of the local build.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/CheckKotlinCompilerVersion.kt
+++ b/platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/CheckKotlinCompilerVersion.kt
@@ -75,6 +75,9 @@ internal abstract class CheckKotlinCompilerVersion : DefaultTask() {
     /**
      * The Kotlin Gradle Plugin runs the compilation through the Build Tools API since version 2.3.20.
      * Older versions need the Gradle property, and every version accepts it.
+     *
+     * This reads the Gradle properties only. The Kotlin Gradle Plugin also accepts the property from
+     * the extra properties of the project, and from the `local.properties` file of the root project.
      */
     private fun Project.runsCompilerViaBuildToolsApi(kotlinPluginVersion: String): Provider<Boolean> =
         providers.gradleProperty(RUN_COMPILER_VIA_BUILD_TOOLS_API_PROPERTY)
--- a/platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/CheckKotlinCompilerVersion.kt
+++ b/platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/CheckKotlinCompilerVersion.kt
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.gradle.kotlin.dsl.plugins.embedded
+
+import org.gradle.api.DefaultTask
+import org.gradle.api.Project
+import org.gradle.api.provider.Property
+import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.Console
+import org.gradle.api.tasks.TaskAction
+import org.gradle.api.tasks.UntrackedTask
+import org.gradle.util.internal.VersionNumber
+import org.jetbrains.kotlin.buildtools.api.ExperimentalBuildToolsApi
+import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
+import org.jetbrains.kotlin.gradle.dsl.kotlinExtension
+import org.jetbrains.kotlin.gradle.plugin.getKotlinPluginVersion
+
+
+/**
+ * Compares the Kotlin compiler version of a project with the Kotlin version embedded in Gradle.
+ *
+ * The Kotlin compile tasks depend on this task, and the comparison then happens after the
+ * configuration of the project. A read of the compiler version finalizes that value, and an earlier
+ * read breaks every build that configures the value.
+ */
+@UntrackedTask(because = "Produces only console output")
+internal abstract class CheckKotlinCompilerVersion : DefaultTask() {
+
+    companion object {
+        private const val RUN_COMPILER_VIA_BUILD_TOOLS_API_PROPERTY = "kotlin.compiler.runViaBuildToolsApi"
+        private val FIRST_VERSION_WITH_BUILD_TOOLS_API_BY_DEFAULT = VersionNumber.parse("2.3.20")
+    }
+
+    @get:Console
+    abstract val kotlinCompilerVersion: Property<String>
+
+    init {
+        kotlinCompilerVersion.set(project.effectiveKotlinCompilerVersion())
+        kotlinCompilerVersion.finalizeValueOnRead()
+    }
+
+    @TaskAction
+    fun checkKotlinCompilerVersion() {
+        logger.warnOnDifferentKotlinVersion(kotlinCompilerVersion.get())
+    }
+
+    /**
+     * The `compilerVersion` property selects the compiler only when the compilation runs through the
+     * Build Tools API. The legacy path always compiles with the compiler of the plugin version.
+     *
+     * The result stays a provider, because a read of `compilerVersion` finalizes that value.
+     */
+    @OptIn(ExperimentalKotlinGradlePluginApi::class, ExperimentalBuildToolsApi::class)
+    private fun Project.effectiveKotlinCompilerVersion(): Provider<String> =
+        getKotlinPluginVersion().let { kotlinPluginVersion ->
+            runsCompilerViaBuildToolsApi(kotlinPluginVersion)
+                .zip(kotlinExtension.compilerVersion.orElse(kotlinPluginVersion)) { viaBuildToolsApi, compilerVersion ->
+                    if (viaBuildToolsApi) compilerVersion else kotlinPluginVersion
+                }
+        }
+
+    /**
+     * The Kotlin Gradle Plugin runs the compilation through the Build Tools API since version 2.3.20.
+     * Older versions need the Gradle property, and every version accepts it.
+     */
+    private fun Project.runsCompilerViaBuildToolsApi(kotlinPluginVersion: String): Provider<Boolean> =
+        providers.gradleProperty(RUN_COMPILER_VIA_BUILD_TOOLS_API_PROPERTY)
+            .map { it.toBoolean() }
+            .orElse(VersionNumber.parse(kotlinPluginVersion).baseVersion >= FIRST_VERSION_WITH_BUILD_TOOLS_API_BY_DEFAULT)
+}
--- a/platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt
+++ b/platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt
@@ -95,11 +95,11 @@ abstract class EmbeddedKotlinPlugin @Inject internal constructor(
 }
 
 
-fun Logger.warnOnDifferentKotlinVersion(kotlinVersion: String?) {
-    if (kotlinVersion != embeddedKotlinVersion) {
+fun Logger.warnOnDifferentKotlinVersion(kotlinCompilerVersion: String?) {
+    if (kotlinCompilerVersion != embeddedKotlinVersion) {
         val warning =
             """|WARNING: Unsupported Kotlin compiler version.
-               |The `embedded-kotlin` and `kotlin-dsl` plugins rely on features of Kotlin `$embeddedKotlinVersion` that might work differently than in the compiler version `$kotlinVersion`.
+               |The `embedded-kotlin` and `kotlin-dsl` plugins rely on features of Kotlin `$embeddedKotlinVersion` that might work differently than in the compiler version `$kotlinCompilerVersion`.
                |Using the `kotlin-dsl` plugin together with a different Kotlin version (for example, by using the Kotlin Gradle plugin (`kotlin(jvm)`)) in the same project is not recommended.
                |
                |See https://docs.gradle.org/${GradleVersion.current().version}/userguide/kotlin_dsl.html#sec:kotlin-dsl_plugin for more details on how the `kotlin-dsl` plugin works (same applies to `embedded-kotlin`).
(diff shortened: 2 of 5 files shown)


eskatos:
@bot-gradle test this

platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt
```diff
@@ -59,6 +63,24 @@ abstract class EmbeddedKotlinPlugin @Inject internal constructor(
         }
     }
 
+    /**
+     * A read of `compilerVersion` finalizes the value, so the read waits for the task configuration.
+     * A project can have several Kotlin compile tasks, and the warning appears one time.
+     */
+    private fun Project.warnOnDifferentKotlinCompilerVersion() {
+        var warned = false
+        tasks.withType<KotlinCompile>().configureEach {
```
jbartok: This might not be late enough (considering that `kotlinExtension.compilerVersion` finalizes the value), if, for example other plugins eagerly realize compile tasks during configuration. We might need to use a provider and defer the warning until execution time.

platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt
```diff
@@ -59,6 +63,24 @@ abstract class EmbeddedKotlinPlugin @Inject internal constructor(
...
+            if (!warned) {
+                warned = true
+                logger.warnOnDifferentKotlinVersion(kotlinCompilerVersion())
+            }
+        }
+    }
+
+    @OptIn(ExperimentalKotlinGradlePluginApi::class, ExperimentalBuildToolsApi::class)
+    private fun Project.kotlinCompilerVersion(): String =
+        kotlinExtension.compilerVersion.orNull ?: getKotlinPluginVersion()
```
jbartok: `compilerVersion` contains a meaningful value only when compilation runs through the Build Tools API (see [here](https://github.com/JetBrains/kotlin/blob/v2.4.20-RC/libraries/tools/kotlin-gradle-plugin-api/src/common/kotlin/org/jetbrains/kotlin/gradle/dsl/KotlinTopLevelExtension.kt#L62-L73) and [here](https://github.com/JetBrains/kotlin/blob/2b6fec71675fc8b671376a92be7fb9d0e6da24b2/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPluginWrapper.kt#L123-L151)). 

BTA became the default only in KGP `2.3.20`, and users can still opt out via `kotlin.compiler.runViaBuildToolsApi=false` in `gradle.properties` (deprecated but supported through `2.4.x`). When BTA is off, KGP ignores `compilerVersion` entirely and compiles with `kotlin-compiler-embeddable:${kotlinPluginVersion}`,so the compiler version is the plugin version, whatever `compilerVersion` says.

platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt
```diff
@@ -76,8 +98,8 @@ abstract class EmbeddedKotlinPlugin @Inject internal constructor(
 fun Logger.warnOnDifferentKotlinVersion(kotlinVersion: String?) {
```
jbartok: Maybe the parameter should be renamed to be more specific (compiler or plugin version?).

jbartok (changes requested):

platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt
eskatos: Good idea, done.

platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt
eskatos: Good catch, fixed

platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt
eskatos: Right! I moved the check to a task to avoid ordering problems and have the warning only once per project.

eskatos:
@bot-gradle test this

platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt
jbartok: I wonder, could we do something to make the warning pop up during IDEA sync too?

platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/CheckKotlinCompilerVersion.kt
```diff
@@ -0,0 +1,83 @@
...
+                .zip(kotlinExtension.compilerVersion.orElse(kotlinPluginVersion)) { viaBuildToolsApi, compilerVersion ->
+                    if (viaBuildToolsApi) compilerVersion else kotlinPluginVersion
+                }
+        }
+
+    /**
+     * The Kotlin Gradle Plugin runs the compilation through the Build Tools API since version 2.3.20.
+     * Older versions need the Gradle property, and every version accepts it.
+     */
+    private fun Project.runsCompilerViaBuildToolsApi(kotlinPluginVersion: String): Provider<Boolean> =
```
jbartok, line 82: KGP resolves it in a more convoluted way (through PropertiesBuildService with precedence extras -> Gradle properties -> root local.properties), see [here](https://github.com/JetBrains/kotlin/blob/v2.4.20-RC/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/internal/properties/PropertiesBuildService.kt#L67-L71).

That being said, I'm not sure we also need to jump through all those hoops...

jbartok (approved):

platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt
eskatos: It mostly does already is the output.
Eventually we could use the Problems API for it so it is surfaced in the IDE.
Out of scope of this PR though.

platforms/core-configuration/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/CheckKotlinCompilerVersion.kt
eskatos: Good catch on the precedence. I kept providers.gradleProperty and documented what it misses: the flag set through project extra properties or through the root local.properties.

The reasoning: when we miss it we report the wrong version inside a warning, we never break a build. Matching the full precedence means reading local.properties through a ValueSource to stay configuration cache correct, plus an eager read of the extras, which duplicates a KGP internal for a property that is deprecated and goes away in Kotlin 2.5.0.

PR #38817, opened 2026-08-12
Intern repository content filtering rules

### LINKED ISSUE #38816: Repository content filtering allocates one `ContentSpec` per rule per repository, with no sharing
### Current Behavior

`DefaultRepositoryContentDescriptor.addInclude`/`addExclude` allocate a fresh `ContentSpec` for
every rule on every descriptor, and each descriptor retains its own private `HashSet` of them for
the life of the build. Nothing is shared between descriptors, so the retained cost is
`projects × repositories × rules`.

On a build with 500 projects, 4 repositories each, and a 697-rule shared repository policy (the
shape of a centrally managed repository setup), `ContentSpec` becomes **the single largest class in
the heap**:

```
$ jcmd <daemon> GC.class_histogram | grep -E 'ContentSpec|^Total'
   1:       1396788       67045824  ...DefaultRepositoryContentDescriptor$ContentSpec
Total       5173650      215559312
```

That is 1,396,788 objects representing **697 distinct rules**, a duplication factor of 2,004×, which
is exactly the repository count. Declaring the same repositories without the rules gives a 90.3 MiB
live set, so content filtering accounts for **115.3 MiB of a 205.6 MiB configuration-time heap
(56%)**.

This was found while profiling a real 519-project build, where the same arithmetic produced
1,473,703 `ContentSpec` objects

### Expected Behavior

Declaring the same content filtering rules on many repositories should cost approximately the size
of the rule set, not the size of the rule set multiplied by the number of repositories.
`ContentSpec` is immutable, caches its `hashCode`, and defines `equals`, so two identical rules on
two repositories are indistinguishable and one instance could serve both.

### Suggested fix

Intern the specs. All `ContentSpec` fields are `private final`, so this is a change to allocation
only:

```java
 class DefaultRepositoryContentDescriptor implements RepositoryContentDescriptorInternal {
+    private static final Interner<ContentSpec> SPEC_INTERNER = Interners.newWeakInterner();

     private void addExclude(String group, String module, String version, MatcherKind matcherKind) {
         assertMutable();
         if (excludeSpecs == null) {
             excludeSpecs = new HashSet<>();
         }
-        excludeSpecs.add(new ContentSpec(matcherKind, group, module, version,
-                versionSelectorScheme, versionSelectors, false));
+        excludeSpecs.add(SPEC_INTERNER.intern(new ContentSpec(matcherKind, group, module, version,
+                versionSelectorScheme, versionSelectors, false)));
     }
```

and the same in `addInclude`. A *weak* interner means entries are collected once the last descriptor
referencing them is gone, so nothing is retained across builds in the daemon.

This is a technique Gradle already applies elsewhere, so it needs no new vocabulary or dependency:
`org.gradle.api.internal.cache.StringInterner` implements `com.google.common.collect.Interner<String>`
over a Guava interner, and `org.gradle.internal.model.InMemoryInterner<T>` is Gradle's own
`intern(T)`/`invalidate()` abstraction for the session-scoped case. Routing this through
`InMemoryInterner` instead of a static field, threaded in via
`AbstractArtifactRepository.createRepositoryDescriptor`, would be tidier and would scope the pool to
the build rather than the daemon, at the cost of some plumbing. The static weak interner needs none.

**Sharing `ContentSpec` between descriptors is already part of the design.** `asMutableCopy()` does:

```java
copy.includeSpecs = includeSpecs == null ? null : Sets.newHashSet(includeSpecs);
copy.excludeSpecs = excludeSpecs == null ? null : Sets.newHashSet(excludeSpecs);
```

It copies the `Set` and shares the elements by reference, so two descriptors already hold the same
`ContentSpec` instances today. Interning extends that from copies to independently-declared
descriptors.

**Sharing `ContentSpec` between descriptors is already part of the design.** `asMutableCopy()` does:

```java
copy.includeSpecs = includeSpecs == null ? null : Sets.newHashSet(includeSpecs);
copy.excludeSpecs = excludeSpecs == null ? null : Sets.newHashSet(excludeSpecs);
```

It copies the `Set` and shares the elements by reference, so two descriptors already hold the same
`ContentSpec` instances today. Interning extends that from copies to independently-declared
descriptors.

### One behavioural consequence worth stating explicitly

`ContentSpec.equals` deliberately excludes the `versionSelectorScheme` and `versionSelectors`
fields. Interning therefore means the surviving instance carries whichever descriptor created it
first, so its `versionSelectors` cache becomes shared across repositories. That looks safe:
`versionSelectors` is a pure `ConcurrentHashMap` memo keyed by the version selector string, and
`versionSelectorScheme` is constructed identically per descriptor as
`new DefaultVersionSelectorScheme(new DefaultVersionComparator(), versionParser)`, where
`DefaultVersionComparator` is stateless. But it is a real change in what is shared, and maintainers
should confirm it rather than take my word for it.

The reproducer checks it empirically. With `-Pintern=true` it interns every spec in the build, then
compares a descriptor whose specs are now shared with 2,003 others against a control descriptor
built independently through the public API only, probing each rule's exact coordinate, a near-miss
module, and a subgroup:

```
=== interning =====================================================================
  spec slots across all descriptors         1,396,788
  distinct specs retained                         697
  duplicate references replaced             1,396,091
  filtering vs independent descriptor    identical on 2091 probes
===================================================================================

$ jcmd <daemon> GC.class_histogram | grep -E 'ContentSpec|^Total'
 206:           697          33456  ...DefaultRepositoryContentDescriptor$ContentSpec
Total       3839145      159476952
```

`ContentSpec` goes from rank 1 to rank 206, 63.9 MiB down to 0.03 MiB, and the live set drops from
205.6 MiB to 152.1 MiB. Interning after the fact cannot avoid the *allocation*, since the objects
already exist by then, so a fix inside `addInclude`/`addExclude` should do at least this well and
also cut the garbage.

### Context (optional)

Large monorepos typically centralize repository declarations and content filtering rules in a
convention plugin, so *every* project gets *every* rule. The rule set grows with the number of
modules being routed (Gradle Module Metadata migrations, mirrors, internal-vs-public splits), and
the project count grows independently, so this scales as a product of two things that both only go
up. It is pure configuration-time floor: paid before any task runs, and paid again per build in the
daemon.

### Why centralizing the repositories is not a substitute

The obvious response is "declare the repositories once in `settings.gradle`'s
`dependencyResolutionManagement` and you will have 4 descriptors instead of 2,076." That is good
advice and worth doing independently, and it would remove most of the cost in the build measured
above. But it narrows the multiplier rather than removing it, and it is not something a build author
can fully rely on:

- **Plugins add repositories to projects, and the build author does not own that code.** Any plugin
  that calls `project.repositories { … }` in its `apply`, whether to reach a private toolchain,
  resolve a detached configuration, or add `google()`, reintroduces a project-level
  `RepositoryHandler`. Moving those to settings means changing every such plugin, including
  third-party ones.
- **Forcing it converts the problem into a build failure.** `FAIL_ON_PROJECT_REPOS` is the only way
  to guarantee centralization, and it makes any plugin that adds a project repository fail the build
  outright. So a large build that wants this is blocked on every plugin in its graph, which is an
  ecosystem coordination problem rather than a local decision.

None of this argues against centralizing. The point is that descriptor count is only partly under
the build author's control, whereas interning inside `addInclude`/`addExclude` makes the cost
proportional to the rule set no matter where or by whom the repositories are declared, and it still
helps the fully-centralized case at a smaller multiplier.

### Two related things, if they are of interest

Both are separate from the above and I am happy to file them separately.

1. **Each descriptor still keeps its own `HashSet`.** After interning, 1,396,788 `HashMap$Node`
   entries remain. That is 42.6 MiB in the reproducer, and most of the gap between the 152.1 MiB
   interned figure and the 90.3 MiB no-rules baseline. It is also what `asMutableCopy`'s
   `Sets.newHashSet` is protecting against, so a copy-on-write set would be the shape of a fix
   rather than a shared immutable one.

2. **`content(Action)` is eager while `toContentFilter()` is lazy.** The `Action` runs immediately in
   `AbstractArtifactRepository.content`, materialising every spec for every repository, but the
   `SpecMatcher` list is only built on first `getContentFilter()` and cached in `cachedAction`. In a
   heap dump of the real build there were 28 `RepositoryFilterAction` instances against 2,076
   repositories, so the overwhelming majority of descriptors were populated and never consulted.
   Storing the `Action` and running it on first `toContentFilter()` would skip the allocation
   entirely for those. One caveat: that measurement is from `help`, which resolves almost nothing,
   so 28 is a floor rather than a typical figure.

[`gradle-contentspec-repro`](https://github.com/rpalcolea/gradle-contentspec-repro). Pure OSS Gradle, no plugins, and fully offline:
`help` resolves nothing and the repository URLs are `https://example.invalid/`, so they can never be
contacted.

```
./measure.sh
```

That runs `./gradlew help` three times, each on a fresh daemon, and cross-checks its own reflective
count against `jcmd GC.class_histogram`:

| run | `ContentSpec` | live set |
|---|---|---|
| `./gradlew help` | 1,396,788 (63.9 MiB), **rank 1** | 205.6 MiB |
| `./gradlew help -Pcontent=false` | 0 | 90.3 MiB |
| `./gradlew help -Pintern=true` | 697 (0.03 MiB), rank 206 | 152.1 MiB |

Full captured output is in `RESULTS.md`. Knobs: `-PprojectCount` (500), `-PrepoCount` (4),
`-PruleCount` (697), `-PgroupCount` (227).

Two details that make the measurement trustworthy without requiring a reviewer to run `jcmd`:

- The reflective count and the class histogram agree to the object, 1,396,788 both ways.
- Group and module names are `String.intern()`ed when the rules are built, so every repository
  shares them. Duplicate `String`s are therefore excluded as a confounder and the measurement is
  unambiguously about the `ContentSpec` objects.

### Gradle version

9.6.1

### Build scan URL (optional)

_No response_

- macOS 15 (Darwin 24.6.0), aarch64
- Reproducer daemon: `-Xmx2g`, compressed oops on, so 48 bytes per `ContentSpec`
- The real build: 519 projects, `-Xmx32g` daemon. Note that `-Xmx32g` is exactly at the
  `UseCompressedOops` threshold, so that daemon runs with compressed oops **off** and every
  reference field costs 8 bytes instead of 4. Worth knowing when reading heap numbers from large
  builds.

rpalcolea:
addInclude() and addExclude() allocated a fresh ContentSpec per rule per descriptor and retained it in that descriptor's own set, so the retained cost of content filtering was projects x repositories x rules with nothing shared between descriptors. Large multi-project builds usually declare repositories and their filtering rules from a convention plugin, so every project gets every rule.

On a 500-project build with 4 filtered repositories each and 697 rules, ContentSpec was the largest class in the heap: 1,396,788 instances holding 63.9 MiB of a 205.6 MiB configuration-time live set, all representing those 697 distinct rules. Interning them takes the class to 697 instances and the live set to 152.1 MiB.
Fixes #38816

### Context

`DefaultRepositoryContentDescriptor.addInclude`/`addExclude` allocate a fresh `ContentSpec` for every
rule and retain it in that descriptor's own `HashSet`. Nothing is shared between descriptors, so the
retained cost of repository content filtering is `projects x repositories x rules`.

That product grows fast in the setup this pattern is designed for. Content filtering is recommended
precisely so that builds with several repositories stop hitting all of them for every module, and the
natural way to apply it is a convention plugin that declares the same rule set on every repository of
every project. Each project then pays for its own private copy of a rule set that is identical
everywhere.

Measured on the reproducer below (500 projects, 4 repositories each, 697 rules, `-Xmx2g`, compressed
oops on), from `jcmd GC.class_histogram` after configuration:

| | objects | retained | live set |
|---|---|---|---|
| rules declared | 1,396,788 `ContentSpec` | 63.9 MiB | 205.6 MiB |
| same repositories, no rules | 0 | 0 | 90.3 MiB |
| every spec canonicalized after the fact | 697 `ContentSpec` | 0.03 MiB | 152.1 MiB |

`ContentSpec` is the single largest class in the heap, and those 1,396,788 objects represent 697
distinct rules, a duplication factor of 2,004. Content filtering accounts for 115.3 MiB of the
205.6 MiB live set, and roughly half of that is pure duplication. A real build with 519 projects shows
the same shape at 1,473,703 instances and 101.2 MiB, higher per instance because a heap above the
compressed oops threshold widens every reference.

A spec is an immutable value, so one instance can serve every descriptor that declares that rule.
This PR canonicalizes them through a weak interner, which makes the retained cost proportional to the
number of distinct rules instead of to rules times repositories.

Centralizing repositories in `dependencyResolutionManagement` is not a substitute here. The default
`RepositoriesMode.PREFER_PROJECT` leaves project repositories in charge wherever a project declares
any, plugins routinely add repositories to individual projects for their own resolution needs, and
`buildscript` repositories are outside `dependencyResolutionManagement` entirely. A build cannot
generally guarantee that no project or plugin declares a repository, so the per-project cost stays.

### Implementation

Two things happen in one commit, because the first is what makes the second safe.

**`ContentSpec` becomes a pure value.** It held `versionSelectorScheme` and a `versionSelectors` memo
belonging to the descriptor that created it, only so `toMatcher()` could pass them to
`SimpleSpecMatcher`'s constructor. `equals` already ignored both. Since `createSpecMatchers` is static
and is called from the instance method `toContentFilter()`, the owning descriptor can pass both down
as parameters instead. After that the spec carries no descriptor-scoped state at all, so equal specs
are genuinely interchangeable and each descriptor keeps using its own scheme and its own memo even
when the spec objects are shared. This is worth having on its own: it drops the object from six
reference fields to four, 48 bytes to 40 with compressed oops and 72 to 56 without.

**The specs are interned.** A `static final Interner<ContentSpec>` from `Interners.newWeakInterner()`,
wrapped around the two `new ContentSpec(...)` call sites. Weak means an entry goes away once the last
descriptor referencing it does, so nothing accumulates across builds in the daemon, and Guava's weak
interner is concurrent, so it is safe under parallel project configuration. This mirrors
`IncludeWithSimpleExpression` in `language-native`, which interns the same way for the same reason.

Sharing `ContentSpec` across descriptors is already how this code works:
`DefaultMavenRepositoryContentDescriptor.asMutableCopy()` does `Sets.newHashSet(getIncludeSpecs())`,
copying the set and sharing the elements by reference. No public API changes, no change to
`equals`/`hashCode`, no build file changes, and one production file touched.

An alternative worth mentioning: `InMemoryInterner` threaded through
`AbstractArtifactRepository.createRepositoryDescriptor` would scope the pool to the build rather than
the daemon, which is arguably tidier, at the cost of plumbing through a constructor that already
carries a `this-escape` suppression. Happy to switch to that if you prefer it.

### Verification

Four unit tests in `DefaultRepositoryContentDescriptorTest`:

- the same rule declared on two independently constructed descriptors yields one instance, and a
  different rule on the same descriptor does not collide with it
- an include and an exclude of the same coordinates stay distinct, which is the assertion that
  covers the `inclusive` flag in `equals`
- two descriptors sharing a spec still filter independently and correctly, including for a version
  range, built with separate `VersionParser` instances. This is the regression guard for the field
  removal: without it, one descriptor's scheme would leak into the other
- `asMutableCopy()` on a descriptor holding shared specs can still take further rules without
  disturbing the original

Checked that these tests are not vacuous: with the interner removed and everything else unchanged,
exactly the two interning-dependent tests fail.

Existing coverage is unchanged, including `RepositoryContentFilteringIntegrationTest`,
`BuildscriptRepositoryContentFilteringIntegrationTest`, and
`ExclusiveRepositoryContentFilteringIntegrationTest`. No new integration test: the change is invisible
from the build script level by design.

### AI involvement

I used Claude Code while working on this. It helped build the reproducer, run and cross-check the heap
measurements, and draft the tests and this description. The design, in particular removing the two
descriptor-scoped fields so that interning is behavior-neutral by construction rather than by
argument, and every claim above, I reviewed and verified myself, and I can explain and defend the
change in review.

### Reproducer

https://github.com/rpalcolea/gradle-contentspec-repro

Fully offline: `help` resolves nothing and every repository URL points at `https://example.invalid/`.
`./measure.sh` runs three configurations on fresh daemons and prints the table above. Project count,
repository count, rule count, and group count are all `-P` knobs.

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/DefaultRepositoryContentDescriptor.java
+++ b/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/DefaultRepositoryContentDescriptor.java
@@ -18,6 +18,8 @@ package org.gradle.api.internal.artifacts.repositories;
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Interner;
+import com.google.common.collect.Interners;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import org.gradle.api.Action;
@@ -43,6 +45,15 @@ import java.util.function.Supplier;
 import java.util.regex.Pattern;
 
 class DefaultRepositoryContentDescriptor implements RepositoryContentDescriptorInternal {
+    /**
+     * Rules are commonly declared identically on every repository of every project, which without
+     * sharing costs one {@code ContentSpec} per rule per repository. Since a spec is an immutable
+     * value, one instance can serve every descriptor declaring that rule. The interner is weak, so
+     * an entry is collected once the last descriptor referencing it is gone and nothing accumulates
+     * across builds in the daemon.
+     */
+    private static final Interner<ContentSpec> SPEC_INTERNER = Interners.newWeakInterner();
+
     private enum MatcherKind {
         SIMPLE,
         REGEX,
@@ -91,7 +102,10 @@ class DefaultRepositoryContentDescriptor implements RepositoryContentDescriptorI
             // no filtering in place
             return Actions.doNothing();
         }
-        cachedAction = new RepositoryFilterAction(createSpecMatchers(includeSpecs), createSpecMatchers(excludeSpecs));
+        cachedAction = new RepositoryFilterAction(
+            createSpecMatchers(includeSpecs, versionSelectorScheme, versionSelectors),
+            createSpecMatchers(excludeSpecs, versionSelectorScheme, versionSelectors)
+        );
         return cachedAction;
     }
 
@@ -117,12 +131,16 @@ class DefaultRepositoryContentDescriptor implements RepositoryContentDescriptorI
     }
 
     @Nullable
-    private static ImmutableList<SpecMatcher> createSpecMatchers(@Nullable Set<ContentSpec> specs) {
+    private static ImmutableList<SpecMatcher> createSpecMatchers(
+        @Nullable Set<ContentSpec> specs,
+        VersionSelectorScheme versionSelectorScheme,
+        ConcurrentHashMap<String, VersionSelector> versionSelectors
+    ) {
         ImmutableList<SpecMatcher> matchers = null;
         if (specs != null) {
             ImmutableList.Builder<SpecMatcher> builder = ImmutableList.builderWithExpectedSize(specs.size());
             for (ContentSpec spec : specs) {
-                builder.add(spec.toMatcher());
+                builder.add(spec.toMatcher(versionSelectorScheme, versionSelectors));
             }
             matchers = builder.build();
         }
@@ -188,7 +206,7 @@ class DefaultRepositoryContentDescriptor implements RepositoryContentDescriptorI
         if (includeSpecs == null) {
             includeSpecs = new HashSet<>();
         }
-        includeSpecs.add(new ContentSpec(matcherKind, group, moduleName, version, versionSelectorScheme, versionSelectors, true));
+        includeSpecs.add(SPEC_INTERNER.intern(new ContentSpec(matcherKind, group, moduleName, version, true)));
     }
 
     @Override
@@ -244,7 +262,7 @@ class DefaultRepositoryContentDescriptor implements RepositoryContentDescriptorI
         if (excludeSpecs == null) {
             excludeSpecs = new HashSet<>();
         }
-        excludeSpecs.add(new ContentSpec(matcherKind, group, moduleName, version, versionSelectorScheme, versionSelectors, false));
+        excludeSpecs.add(SPEC_INTERNER.intern(new ContentSpec(matcherKind, group, moduleName, version, false)));
     }
 
     @Override
@@ -328,23 +346,25 @@ class DefaultRepositoryContentDescriptor implements RepositoryContentDescriptorI
         this.requiredAttributes = requiredAttributes;
     }
 
+    /**
+     * A rule declared on a descriptor. This is a pure value: it carries no state belonging to the
+     * descriptor that created it, so equal specs are interchangeable and a single instance can be
+     * shared by every descriptor declaring that rule. The version selector scheme and the selector
+     * cache are supplied by the owning descriptor when the matcher is created.
+     */
     private static class ContentSpec {
         private final MatcherKind matcherKind;
         private final String group;
         private final String module;
         private final String version;
-        private final VersionSelectorScheme versionSelectorScheme;
-        private final ConcurrentHashMap<String, VersionSelector> versionSelectors;
         private final boolean inclusive;
         private final int hashCode;
 
-        private ContentSpec(MatcherKind matcherKind, String group, @Nullable String module, @Nullable String version, VersionSelectorScheme versionSelectorScheme, ConcurrentHashMap<String, VersionSelector> versionSelectors, boolean inclusive) {
+        private ContentSpec(MatcherKind matcherKind, String group, @Nullable String module, @Nullable String version, boolean inclusive) {
             this.matcherKind = matcherKind;
             this.group = group;
             this.module = module;
             this.version = version;
-            this.versionSelectorScheme = versionSelectorScheme;
-            this.versionSelectors = versionSelectors;
             this.inclusive = inclusive;
             this.hashCode = Objects.hashCode(matcherKind, group, module, version, inclusive);
         }
@@ -371,7 +391,7 @@ class DefaultRepositoryContentDescriptor implements RepositoryContentDescriptorI
             return hashCode;
         }
 
-        SpecMatcher toMatcher() {
+        SpecMatcher toMatcher(VersionSelectorScheme versionSelectorScheme, ConcurrentHashMap<String, VersionSelector> versionSelectors) {
             switch (matcherKind) {
                 case SIMPLE:
                 case SUB_GROUP:
--- a/platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/repositories/DefaultRepositoryContentDescriptorTest.groovy
+++ b/platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/artifacts/repositories/DefaultRepositoryContentDescriptorTest.groovy
@@ -486,4 +486,112 @@ class DefaultRepositoryContentDescriptorTest extends Specification {
         0 * details.notFound()
     }
 
+    def "the same rule declared on independent descriptors is a single spec instance"() {
+        given:
+        def first = new DefaultRepositoryContentDescriptor({ "first" }, new VersionParser())
+        def second = new DefaultRepositoryContentDescriptor({ "second" }, new VersionParser())
+
+        when:
+        first.includeGroup("org")
+        second.includeGroup("org")
+        second.includeGroup("other")
+
+        then:
+        first.includeSpecs.size() == 1
+        second.includeSpecs.size() == 2
+        second.includeSpecs.any { it.is(first.includeSpecs.first()) }
+        !second.includeSpecs.every { it.is(first.includeSpecs.first()) }
+    }
+
+    def "an include and an exclude of the same coordinates are not the same spec"() {
+        given:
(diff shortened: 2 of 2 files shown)


ov7a:

big-guy:
Thanks Roberto, we'll take a look at this

PR #16619, opened 2021-03-23
Keep a project scope input node access hierarchy

wolfs:
to improve performance. This seems to mitigate a big part of #16581.

Keeping the input access hierarchy project scoped may mean that we may miss some missing dependencies of tasks writing to the input location of tasks in another project.

PR #38818, opened 2026-08-13
Detect overlapping outputs between tasks in different builds

jvandort:
Previously, we did not detect overlapping outputs between tasks in different builds. By moving ExecutionNodeAccessHierarchies to the build tree scope, we can now detect these scenarios where we previously did not.
This means
- Tasks between builds with overlapping outputs will no longer run concurrently.
- Tasks that consume outputs from a task in another build will emit a deprecation warning if there is no declared dependency between those tasks.

Additionally, we also move NodeValidator to the build tree scope, as it only depends on ProblemsInternal, a build tree scoped service.

These two changes open the door to TaskNodeFactory becoming a build tree scoped service in an upcoming change set.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildOverlappingOutputsIntegrationTest.groovy
+++ b/subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildOverlappingOutputsIntegrationTest.groovy
@@ -41,7 +41,13 @@ class CompositeBuildOverlappingOutputsIntegrationTest extends AbstractCompositeB
 
         def buildB = singleProjectBuild("buildB") {
             buildFile << """
+                tasks.register("includedGate") {
+                    doLast {
+                        ${server.callFromBuild("includedGate")}
+                    }
+                }
                 tasks.register("includedPing") {
+                    dependsOn("includedGate")
                     outputs.dir(file("$sharedDir"))
                     doLast {
                         ${server.callFromBuild("includedPing")}
@@ -52,7 +58,13 @@ class CompositeBuildOverlappingOutputsIntegrationTest extends AbstractCompositeB
         includeBuild(buildB)
 
         buildA.buildFile << """
+            tasks.register("rootGate") {
+                doLast {
+                    ${server.callFromBuild("rootGate")}
+                }
+            }
             tasks.register("rootPing") {
+                dependsOn("rootGate")
                 outputs.file(file("$sharedDir/file.txt"))
                 doLast {
                     ${server.callFromBuild("rootPing")}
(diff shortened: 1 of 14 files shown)


jvandort:
@bot-gradle test this

jvandort:
@bot-gradle test this

lkasso (approved):
Docs LGTM.

subprojects/core/src/main/java/org/gradle/execution/plan/MissingTaskDependencyDetector.java
```diff
@@ -48,6 +54,7 @@
  * consuming a parent directory of the produced output.
  * </p>
  */
+@ServiceScope(Scope.BuildTree.class)
 public class MissingTaskDependencyDetector {
     private final ExecutionNodeAccessHierarchy outputHierarchy;
     private final ExecutionNodeAccessHierarchies.InputNodeAccessHierarchy inputHierarchy;
```
ghale, line 56: ❓ Prior to this change, `MissingTaskDependencyDetector` was a project scoped service, which meant inputHierarchy was discarded with every build.  Now it's a build-tree scoped service.  `ExecutionNodeAccessHierarchies` implements `HoldsProjectState` and its `discardAll()` clears out `outputHierarchy` and `destroyableHierarchy`, but not inputHierarchy, which was created on a per-project basis.

Now, we have moved both `MissingTaskDependencyDetector` and the creation of `inputHierarchy` to build-tree scope, which raises two questions:
1) should we now create and clear out `inputHierarchy` in a similar way (i.e. as a single instance, tracked by `ExecutionNodeAccessHierarchies` and cleared out in `discardAll()`?  (I think yes)
2) The comment on `createInputHierarchy()` suggests that this was project-scoped "for performance reasons".  Do the same performance concerns still exist?  And if so, are we creating a problem by making this build-tree scoped? (I don't know)

subprojects/core/src/main/java/org/gradle/execution/plan/MissingTaskDependencyDetector.java
```diff
@@ -176,4 +208,30 @@ private static void collectValidationProblem(Node producer, Node consumer, TypeV
...
         );
     }
+
+    private static boolean mayBeDifferentBuilds(Node producer, Node consumer) {
+        Path producerBuild = buildPathOf(producer);
+        if (producerBuild == null) {
+            return true;
+        }
+        Path consumerBuild = buildPathOf(consumer);
+        if (consumerBuild == null) {
```
ghale: ❓ Doesn't the `isEnabled` filter preclude this?  i.e. it is only enabled if it is an instance of a `LocalTaskNode` and so `collectValidationProblem()` only gets called if the node is a LocalTaskNode.  Similarly, the `LocalTaskNode` argument for producer in `collectValidationProblemsForProducer` and `collectValidationProblemsForConsumer` guarantees that the other value is also `LocalTaskNode`.

In other words, couldn't we make `collectValidationProblem()` and therefore `maybeDifferentBuilds()` take `LocalTaskNode` as arguments?  Then buildPathOf can always follow the `TaskNode` path and doesn't have to return null and we don't need this handling?  Or am I missing some detail that still requires this?

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildOverlappingOutputsIntegrationTest.groovy
```diff
@@ -0,0 +1,189 @@
...
+class CompositeBuildOverlappingOutputsIntegrationTest extends AbstractCompositeBuildIntegrationTest {
+
+    @Rule
+    BlockingHttpServer server = new BlockingHttpServer()
+
+    def setup() {
+        server.start()
+    }
+
+    def "tasks with overlapping outputs in different builds do not run in parallel"() {
```
ghale, line 38: 💅  We should add a `--max-workers` argument to this test.  We do this in `CompositeBuildParallelIntegrationTest`.  I don't think we're guaranteed multiple workers without it (although it is likely that we will get multiple workers).

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildOverlappingOutputsIntegrationTest.groovy
```diff
@@ -0,0 +1,189 @@
...
+            .withTasks(":buildB:includedPing", ":rootPing")
+            .start()
+
+        then:
+        // One of the two tasks is now frozen mid-action, holding its claim on the shared output location
+        handle.waitForAllPendingCalls()
+        // This is the detection window. If the mutual exclusion of overlapping outputs across builds
+        // were broken, the other task would start while the first is frozen, its call would arrive at
+        // the server, and the test would fail immediately with an unexpected request error.
+        Thread.sleep(500)
```
ghale: 🤔 I wonder if there is a better way to ensure this.  A sleep only makes it _more_ likely that two tasks that might run in parallel would in the case that this is broken, but it doesn't guarantee that they will.  On the other hand, it introduces an arbitrary 500ms sleep.  Another way to make it more likely without needing a sleep is to block on two dependencies that _can_ run concurrently.  When you release them, the other tasks should be more likely to run in parallel if they were allowed to.  There's still a race there, but it's smaller and its _maybe_ better than an arbitrary sleep.  Or maybe you have another clever idea?

This shouldn't block the PR, but its something we should at least think about.

ghale (changes requested):
I think this looks good.  I left a few questions where I think we might have some rough edges.  Nothing that seems like a major problem, but I think we should at least answer those questions before merging.

jvandort:
@bot-gradle test this

subprojects/core/src/main/java/org/gradle/execution/plan/MissingTaskDependencyDetector.java
jvandort: Good catch, I didn't realize `MissingTaskDependencyDetector` held a per-project instance of input hierarchies. Note, #16619, which added the per-project input hierarchies mentions `we may miss some missing dependencies of tasks writing to the input location of tasks in another project.`, which is not ideal. 

However, the goal of this PR was to raise ExecutionNodeAccessHierarchies to the build tree scope, so I reverted lifting of `MissingTaskDependencyDetector` in this PR. We now still detect some cross-build outputs between tasks, but with the same corner cases for same-build detections we have today. 

This should resolve any concerns about performance, though we should look into this in the future to try to close some of these corner cases.

subprojects/core/src/main/java/org/gradle/execution/plan/MissingTaskDependencyDetector.java
jvandort: Yeah I missed that this only executed against local task nodes. I updated the logic to cast early to `LocalTaskNode` when we check it so we can remove this dead code.

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildOverlappingOutputsIntegrationTest.groovy
jvandort: Added explicit `--max-workers`, though in practice I think we likely always get parallelism in the real world.

subprojects/composite-builds/src/integTest/groovy/org/gradle/integtests/composite/CompositeBuildOverlappingOutputsIntegrationTest.groovy
jvandort: The sleep definitely was not ideal. I applied your suggestion which seems to work just as well without the sleep.

ghale (approved):

jvandort:
@bot-gradle test apt

PR #38829, opened 2026-08-13
Allow adding setters for upgraded lazy properties

### LINKED ISSUE #38038: Remove redundant `Property<URI>` special case from class generators
### Issue type

Refactoring

### Problem description

`AbstractClassGenerator#claimPropertyImplementation` has a special-case guard: `!isLazyUriProperty(setter, property)` in [`AbstractClassGenerator.java`](https://github.com/gradle/gradle/blob/master/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java#L1118).

This guard short-circuits the disqualification step, so a property with an `abstract Property<URI> getX()` and a hand-written `void setX(String)` still gets claimed (i.e. the generator provides the managed `Property<URI>` backing).

After the setter reintroduction in #37783, the same case is covered by the general `!allAbstract` clause and can be folded back into the regular path instead of remaining an exception.

---

Also: the stale comment pointing at `AsmBackedClassGenerator#isLazyUriPropertyWithSetObject` (the method doesn't exist anymore) is worth deleting as part of the same cleanup — maybe a one-liner in the scope.

### Context (optional)

Follow-up from review of #37783, [see the thread](https://github.com/gradle/gradle/pull/37783/files#r3302735750). The URI special case predates the generalized "concrete setters on all-abstract-getter properties are
  forwarders" logic and is now subsumed by it.

asodja:
This continues work of https://github.com/gradle/gradle/pull/38196.

The main change is that we allow setters only for `@ReplacesEagerProperty` getters while for the rest we have the old behavior.

### What is handled, and how

**The annotation may sit anywhere on the getter in the hierarchy.** A method annotation is not
inherited by an override, and `ClassDetails` keeps only the most concrete declaration per name and
return type — so an annotated supertype getter disappears as soon as a subtype re-declares it. Both
shapes below are common in the migration and both are handled, by looking the annotation up over every
getter declaration rather than the one that survived:

```java
// annotation on the super-interface, re-declared with the same signature
public interface HasProp {
    @ReplacesEagerProperty
    Property<String> getProp();
}

public abstract class Bean implements HasProp {
    @Override
    public abstract Property<String> getProp();          // annotation not inherited

    public void setProp(String value) {
        getProp().set(value);
    }
}
```

```java
// covariant override: javac emits a concrete bridge getter, which is not user-managed state
public interface HasProp {
    @ReplacesEagerProperty
    Provider<String> getProp();
}

public abstract class Bean implements HasProp {
    @Override
    public abstract Property<String> getProp();

    public void setProp(String value) {
        getProp().set(value);
    }
}
```

**The setter may be declared below the getter.**

```java
public abstract class Base {
    @ReplacesEagerProperty
    public abstract Property<String> getProp();
}

public abstract class Bean extends Base {
    public void setProp(String value) {
        getProp().set(value);
    }
}
```

**The setter may take the property type rather than the value type.** This used to make the property
look like a plain mutable one: a setter was generated over the hand-written body and the property was
never created, so the getter returned `null`.

```java
@ReplacesEagerProperty
public abstract ConfigurableFileCollection getFiles();

public void setFiles(ConfigurableFileCollection files) {
    getFiles().setFrom(files);       // runs; not replaced by generated state
}
```

**A type that already declares `set<Name>(Object)`** no longer also gets the generated
`setFromAnyValue`-backed overload, which would be a duplicate method. This replaces the old
`Property<URI>` special case.

### What is deliberately not handled

**No annotation, no relaxation.** A concrete setter still means "hands off", so this is rejected
exactly as on `master`:

```java
public abstract class Bean {
    public abstract Property<String> getProp();          // no @ReplacesEagerProperty

    public void setProp(String value) {
        getProp().set(value);
    }
}
```

**A property type that cannot be created.** `Provider` is not a `@ManagedType`, so there is no lazy
property to forward into. Rather than generate over the setter, the property stays unclaimed and you
get the `Cannot have abstract method` error at generation time:

```java
@ReplacesEagerProperty
public abstract Provider<String> getProp();

public void setProp(String value) { ... }
```

**A concrete getter**, annotated or not. A getter with a body reads state the author manages, so the
property is not ours to implement.

Fixes https://github.com/gradle/gradle-private/issues/5282
Also fixes https://github.com/gradle/gradle/issues/38038

### DIFF
--- a/platforms/core-configuration/model-core/build.gradle.kts
+++ b/platforms/core-configuration/model-core/build.gradle.kts
@@ -31,6 +31,7 @@ dependencies {
     implementation(projects.baseServicesGroovy)
     implementation(projects.baseAsm)
     implementation(projects.classloaders)
+    implementation(projects.internalInstrumentationApi)
     implementation(projects.logging)
     implementation(projects.serviceRegistryBuilder)
 
--- a/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
+++ b/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java
@@ -357,9 +357,26 @@ abstract class AbstractClassGenerator implements ClassGenerator {
                 propertyMetaData.field(property.getBackingField());
             }
         }
-        // ClassDetails keeps a single declaration per getter signature, preferring the most concrete one, so a
-        // supertype getter that a subtype re-declares (directly, or via the bridge of a covariant override) is
-        // dropped. Record every declaration separately, so that annotations on the dropped ones remain visible.
+        recordGetterDeclarations(classDetails, classMetaData);
+        for (Method method : classDetails.getInstanceMethods()) {
+            Class<?>[] parameterTypes = method.getParameterTypes();
+            if (parameterTypes.length == 1) {
+                PropertyMetadata propertyMetaData = classMetaData.getProperty(method.getName());
+                if (propertyMetaData != null) {
+                    propertyMetaData.addSetMethod(method);
+                }
+            }
+        }
+    }
+
+    /**
+     * {@link ClassDetails} keeps a single declaration per getter signature, preferring the most concrete one, so a
+     * supertype getter that a subtype re-declares — directly, or via the bridge of a covariant override — is dropped.
+     * Record every declaration too, so that annotations on the dropped ones remain visible.
+     *
+     * <p>Mirrors how {@link ClassInspector} classifies methods into properties.
+     */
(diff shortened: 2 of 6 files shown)


asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this

asodja:
@bot-gradle test this APT

PR #38840, opened 2026-08-14
Fix ABI extraction defects that miss recompiles

### LINKED ISSUE #38823: Compile avoidance misses a change of record component order
### Current Behavior

Gradle uses the extracted ABI of each compile classpath entry for compile avoidance.
The extractor drops every record component, so a change of the component order does
not change the hash. Gradle then keeps the consumer up to date.

A record pattern binds each variable to a component position. The consumer keeps the
old positions after the change, and it gives a wrong result.

The reproducer has a library with `record Point(int x, int y)` and a consumer with a
record pattern. A swap of the two components gives this output:

```
> Task :lib:compileJava
> Task :app:compileJava UP-TO-DATE
f(new Point(1, 2)) = 21
```

A clean build of the same sources prints 12.

The cause is in `platforms/core-configuration/java-api-extractor`. `ApiMemberSelector`
does not implement `visitRecordComponent`, so the extracted class holds an empty
`Record` attribute. Both component orders keep the same accessors, the same canonical
constructor and the same private fields, therefore the extracted bytes are equal.

### Expected Behavior

The extracted ABI holds the record components. A change of the component order changes
the hash, and Gradle recompiles the consumer. An incremental build gives the same
result as a clean build.

### Context (optional)

Stumbled upon this while looking at
* https://github.com/gradle/gradle/pull/38810

Related issues
* https://github.com/gradle/gradle/issues/38825
* https://github.com/gradle/gradle/issues/38827
* https://github.com/gradle/gradle/issues/38828

The attached `records-abi-repro.zip` holds a two-module build. The wrapper uses Gradle
9.7.0. Java 21 or later is necessary for the record pattern.

[records-abi-repro.zip](https://github.com/user-attachments/files/31025625/records-abi-repro.zip)

Run `./reproduce.sh`, or follow these steps:

1. Run `./gradlew :app:run`. The output is 12.
2. Run `cp lib/variants/Point-swapped.java lib/src/main/java/Point.java`.
3. Run `./gradlew :app:run`. Gradle reports `:app:compileJava UP-TO-DATE`, and the
   output is 21.
4. Run `./gradlew :app:clean :app:run`. The output is 12 again.

`new Point(1, 2)` always puts 1 in the first component and 2 in the second component.
The pattern always reads the first component and then the second component. The
correct output is therefore 12 for both component orders.

### Gradle version

9.7.0. Gradle 9.6.1 gives the same result.

### Build scan URL (optional)

_No response_

_No response_

### LINKED ISSUE #38825: Compile avoidance misses a move of a type-use annotation in a throws clause
### Current Behavior

Gradle uses the extracted ABI of each compile classpath entry for compile avoidance.
The extractor writes the thrown exceptions of a method in sorted order. The
`type_index` of a `THROWS` type annotation points into the `Exceptions` attribute in
declaration order.

The extractor keeps the original index and writes a sorted list. The annotation
therefore moves to another exception when the sort changes the order. Two methods with
different annotated exceptions extract to the same bytes:

| source | extracted |
| --- | --- |
| `void m() throws @X ArithmeticException, IOException` | `throws java.io.IOException, java.lang.ArithmeticException` with `THROWS, type_index=0` |
| `void m() throws @X IOException, ArithmeticException` | `throws java.io.IOException, java.lang.ArithmeticException` with `THROWS, type_index=0` |

An annotation processor reads such an annotation with
`ExecutableElement.getThrownTypes()` and `TypeMirror.getAnnotationMirrors()`. The
annotation is therefore a real input of the compilation of the consumer. Gradle keeps
the consumer up to date, and the processor keeps its old output:

```
> Task :lib:compileJava
> Task :app:compileJava UP-TO-DATE
annotated thrown type = java.lang.@X ArithmeticException
```

A clean build of the same sources prints `java.io.@X IOException`.

The wrong index is also a defect on its own, without the hash. The extracted class
attaches the annotation to an exception that the source does not annotate.

The cause is in `platforms/core-configuration/java-api-extractor`. `MethodMember` holds
the exceptions in a `TreeSet`, and `JavaApiMemberWriter.writeMethod` writes
`method.getExceptions().toArray(new String[0])`.

### Expected Behavior

The extractor writes the `type_index` of a `THROWS` type annotation for the order of
the exceptions that it writes. The annotation stays on the exception of the source.
Two methods with different annotated exceptions extract to different bytes, and Gradle
recompiles the consumer.

### Context (optional)

Stumbled upon this while working on
* https://github.com/gradle/gradle/pull/38810

Related issues
* https://github.com/gradle/gradle/issues/38823
* https://github.com/gradle/gradle/issues/38827
* https://github.com/gradle/gradle/issues/38828

The attached `throws-abi-repro.zip` holds a three-module build with a small annotation
processor. The wrapper uses Gradle 9.7.0. Java 17 or later is necessary.

[throws-abi-repro.zip](https://github.com/user-attachments/files/31025728/throws-abi-repro.zip)

Run `./reproduce.sh`, or follow these steps:

1. Run `./gradlew :app:run`. The output names `ArithmeticException`.
2. Run `cp lib/variants/T-annotation-on-ioexception.java lib/src/main/java/T.java`.
3. Run `./gradlew :app:run`. Gradle reports `:app:compileJava UP-TO-DATE`, and the
   output still names `ArithmeticException`.
4. Run `./gradlew :app:clean :app:run`. The output names `IOException`.

### Gradle version

9.7.0. Gradle 9.6.1 gives the same result.

### Build scan URL (optional)

_No response_

_No response_

### LINKED ISSUE #38827: Compile avoidance misses a rename of a method parameter
### Current Behavior

Gradle uses the extracted ABI of each compile classpath entry for compile avoidance.
The extractor drops the `MethodParameters` attribute, so a rename of a method parameter
does not change the hash. Gradle then keeps the consumer up to date.

An annotation processor reads parameter names with `ExecutableElement.getParameters()`
and `VariableElement.getSimpleName()`. The names are therefore a real input of the
compilation of the consumer. The processor keeps its old output after the rename:

```
> Task :lib:compileJava
> Task :app:compileJava UP-TO-DATE
parameter names = firstName, repeatCount
```

A clean build of the same sources prints `parameter names = givenName, times`.

The cause is in `platforms/core-configuration/java-api-extractor`. `ApiMemberSelector`
does not implement `visitParameter`, therefore the extracted class holds no parameter
names. A method with the parameters `(String firstName, int repeatCount)` and the same
method with `(String givenName, int times)` extract to the same bytes.

### Expected Behavior

The extracted ABI holds the parameter names of a method. A rename changes the hash, and
Gradle recompiles the consumer. An incremental build gives the same result as a clean
build.

### Context (optional)

Many annotation processors read parameter names. Mapping libraries, dependency
injection libraries and binding libraries all use them to generate code.

`javac` does not need the parameter names of a dependency for its own code generation.
The names are still an input of the compile task, because the task also runs the
annotation processors. Gradle keeps the type annotations of a method for the same
reason since Gradle 9.3.0, which came with #35658.

The build gives stale generated code without an error and without a warning. The result
of an incremental build differs from the result of a clean build. The stale state also
survives a change back to the original names, because that change does not modify the
hash either.

The library needs the `-parameters` compiler option for the names to reach the class
file. Several build setups add this option by default.

Stumbled upon this while working on
* https://github.com/gradle/gradle/pull/38810

Related issues
* https://github.com/gradle/gradle/issues/38825
* https://github.com/gradle/gradle/issues/38823
* https://github.com/gradle/gradle/issues/38828

The attached `paramnames-abi-repro.zip` holds a three-module build with a small
annotation processor. The wrapper uses Gradle 9.7.0. Java 17 or later is necessary.

[paramnames-abi-repro.zip](https://github.com/user-attachments/files/31028661/paramnames-abi-repro.zip)

Run `./reproduce.sh`, or follow these steps:

1. Run `./gradlew :app:clean :app:run`. The output is `firstName, repeatCount`.
2. Run `cp lib/variants/T-renamed.java lib/src/main/java/T.java`.
3. Run `./gradlew :app:run`. Gradle reports `:app:compileJava UP-TO-DATE`, and the
   output still shows `firstName, repeatCount`.
4. Run `./gradlew :app:clean :app:run`. The output is `givenName, times`.

Step 1 starts from a clean state on purpose, because a stale state of `app` survives an
ordinary incremental build.

### Gradle version

9.7.0. Gradle 9.6.1 gives the same result.

### Build scan URL (optional)

_No response_

_No response_

### LINKED ISSUE #38828: Extracted ABI replaces an annotation value member with an array element
### Current Behavior

The API extractor writes a wrong annotation into the extracted class. The annotation
loses the content of its `value` member, and the extractor puts the first element of an
array member there:

```
real class:      Bar(value="v", names=["x","y"])
extracted class: Bar(names=["x","y"], value="x")
```

This has two effects.

**Wrong bytes in the ABI jar of the distribution.** The `lib/api` directory of the
distribution holds the extracted classes of the Gradle API. The Kotlin DSL compiles
build scripts against that jar. In Gradle 9.7.0 the class
`org.gradle.api.artifacts.transform.InputArtifact` carries this annotation:

```
real:      InjectionPointQualifier(supportedProviderTypes=[FileSystemLocation])
extracted: InjectionPointQualifier(supportedProviderTypes=[FileSystemLocation],
                                   value=FileSystemLocation)
```

`InjectionPointQualifier` declares `supportedTypes()` and `supportedProviderTypes()`. It
declares no `value()` member.

**A missed recompilation.** Gradle hashes the extracted class for compile avoidance. Two
classes with a different `value` member extract to the same bytes, so the hash is equal.
An annotation processor reads annotations with `Element.getAnnotationMirrors()`, so the
`value` member is a real input of the compilation of the consumer. Gradle keeps the
consumer up to date, and the processor generates output from the old value:

```
> Task :lib:compileJava
> Task :app:compileJava UP-TO-DATE
Bar value = v
```

A clean build of the same sources prints `Bar value = w`.

Three shapes of annotation give three results:

| annotation | result |
| --- | --- |
| a `value` member and an array member | the content of `value` is lost |
| no `value` member, one array member | an extra member with the name `value` appears |
| the `value` member is the array | correct content |

The `@Target` annotation and the `@SuppressWarnings` annotation belong to the third
shape, which is one reason why the problem stayed hidden.

### Expected Behavior

The extracted annotation holds the members of the real annotation, and it holds no other
member. Two classes with a different `value` member extract to different bytes, and
Gradle recompiles the consumer.

### Context (optional)

The cause is in `SortingAnnotationVisitor.visitEnd` in
`platforms/core-configuration/java-api-extractor`. The array branch wraps the collected
elements in an `ArrayAnnotationValue`. The method then continues to
`annotation.addValues(annotationValues)`, which adds each element to the annotation as a
member. Each element carries the name `value`, because an array element has no name of
its own.

The members live in a `TreeSet`, and `AnnotationValue.compareTo` compares only the name.
The visitor of the array ends before the visitor of the annotation. The wrong `value`
therefore enters the set first, and the set rejects the real `value`.

A stop after the array branch corrects all three results at once.

`AnnotationValue.compareTo` also deserves a look. A comparison on the name alone drops
data without a message whenever two members share a name.

The build gives stale generated code without an error and without a warning. The result
of an incremental build differs from the result of a clean build.

Stumbled upon this while working on:
* https://github.com/gradle/gradle/pull/38810

Related issues
* https://github.com/gradle/gradle/issues/38825
* https://github.com/gradle/gradle/issues/38823
* https://github.com/gradle/gradle/issues/38827

The attached `annvalue-abi-repro.zip` holds a three-module build with a small annotation
processor. The wrapper uses Gradle 9.7.0. Java 17 or later is necessary.

[annvalue-abi-repro.zip](https://github.com/user-attachments/files/31029208/annvalue-abi-repro.zip)

Run `./reproduce.sh`, or follow these steps:

1. Run `./gradlew :app:clean :app:run`. The output is `Bar value = v`.
2. Run `cp lib/variants/T-value-w.java lib/src/main/java/T.java`.
3. Run `./gradlew :app:run`. Gradle reports `:app:compileJava UP-TO-DATE`, and the output
   is still `Bar value = v`.
4. Run `./gradlew :app:clean :app:run`. The output is `Bar value = w`.

For the wrong bytes alone, unzip `lib/api/gradle-public-api-legacy-9.7.0.jar` from a
Gradle 9.7.0 distribution. Then run:

    javap -v -p org/gradle/api/artifacts/transform/InputArtifact.class

### Gradle version

9.7.0. Gradle 9.6.1 gives the same result.

### Build scan URL (optional)

_No response_

_No response_

big-guy:
Fixes #38823
Fixes #38825
Fixes #38827
Fixes #38828

### Context

Gradle compares the extracted ABI of each compile classpath entry to decide whether a consumer has to be recompiled. Four kinds of API change left the extracted bytes unchanged, so Gradle kept the consumer up to date and an incremental build produced a different result than a clean build — with no error and no warning. All four were reported by the same contributor while working on #38810, on top of which this builds.

**Annotation members displaced by array elements (#38828).** The array branch of `SortingAnnotationVisitor.visitEnd` wrapped the collected elements in an `ArrayAnnotationValue` and then fell through to `annotation.addValues(annotationValues)`, adding every element to the annotation as a member. An array element carries no name of its own, so each one arrived as `value` and displaced the real `value` member in the `TreeSet`. This also writes wrong bytes into the ABI jar of the distribution: in 9.7.0, `InputArtifact` carries an `InjectionPointQualifier(value=FileSystemLocation)` that its source never declares.

**Parameter names dropped (#38827).** `ApiMemberSelector` never implemented `visitParameter`, so `(String firstName, int repeatCount)` and `(String givenName, int times)` extracted to the same bytes. Two things were needed: collecting the parameters, and reading the class with `SKIP_CODE` instead of `SKIP_DEBUG` — ASM gates the `MethodParameters` attribute behind the latter. `SKIP_CODE` still discards method bodies and all the debug information inside them, and skips more work than before.

**Record components dropped (#38823).** `ApiMemberSelector` never implemented `visitRecordComponent`. Swapping the components of `record Point(int x, int y)` keeps every accessor, the canonical constructor and every private field, so both orders extracted to the same bytes — while a record pattern in the consumer binds by position and silently produced a wrong result.

**`THROWS` type annotation on the wrong exception (#38825).** The `type_index` of a `THROWS` type annotation points into the `Exceptions` attribute in declaration order, but the extractor writes the exceptions sorted and kept the original index. The annotation therefore moved to another exception, which is a defect on its own beyond the hash. `MethodMember` now maps the index onto the written order.

Parameter names, record components and type annotations in a `throws` clause are all real inputs of the consumer's compilation, since the compile task also runs annotation processors, which read them via `VariableElement.getSimpleName()`, `RecordComponentElement` and `ExecutableElement.getThrownTypes()`. This is the same reasoning that kept type annotations in #35658 and #38810.

Note that the extracted ABI of every classpath entry changes, so the first build after this change recompiles everything.

### Contributor Checklist
- [x] Ensure that tests pass locally.

### Testing

8 unit tests in `ApiClassExtractorTest` / `ApiClassExtractorAnnotationsTest`, and 4 integration tests in `AbstractJavaCompileAvoidanceIntegrationSpec`, which run across all three subclasses (incremental/non-incremental × jar/classes-dir).

Every test was verified against a stashed fix: all 8 unit tests and all 8 jar-variant integration tests fail without it.

Green locally: `sanityCheck`, `:java-api-extractor:test`, `:normalization-java:test`, `:kotlin-dsl:test`, and all compile-avoidance integration tests in `plugins-java`, `plugins-groovy`, `language-java` and `kotlin-dsl`.

### DIFF
--- a/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/ApiClassExtractor.java
+++ b/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/ApiClassExtractor.java
@@ -98,7 +98,9 @@ public class ApiClassExtractor {
         ApiMemberSelector visitor;
         try {
             visitor = new ApiMemberSelector(originalClassReader.getClassName(), apiMemberWriterFactory.makeApiMemberWriter(apiClassWriter), apiIncludesPackagePrivateMembers);
-            originalClassReader.accept(visitor, ClassReader.SKIP_DEBUG | ClassReader.SKIP_FRAMES);
+            // SKIP_CODE discards method bodies and all the debug information they contain, but
+            // unlike SKIP_DEBUG it keeps the MethodParameters attribute, which is part of the API.
+            originalClassReader.accept(visitor, ClassReader.SKIP_CODE | ClassReader.SKIP_FRAMES);
         } catch (ApiClassExtractionException e) {
             throw e.withClass(originalClassReader.getClassName());
         }
--- a/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/ApiMemberWriter.java
+++ b/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/ApiMemberWriter.java
@@ -22,6 +22,7 @@ import org.gradle.internal.tools.api.impl.ClassMember;
 import org.gradle.internal.tools.api.impl.FieldMember;
 import org.gradle.internal.tools.api.impl.InnerClassMember;
 import org.gradle.internal.tools.api.impl.MethodMember;
+import org.gradle.internal.tools.api.impl.RecordComponentMember;
 import org.jspecify.annotations.Nullable;
 import org.objectweb.asm.AnnotationVisitor;
 import org.objectweb.asm.FieldVisitor;
@@ -40,6 +41,8 @@ public interface ApiMemberWriter {
     ModuleVisitor writeModule(String name, int access, @Nullable String version);
 
     void writeClass(ClassMember classMember, Set<MethodMember> methods, Set<FieldMember> fields, Set<InnerClassMember> innerClasses);
+
+    void writeRecordComponent(RecordComponentMember recordComponent);
 
     void writeMethod(ClassMember classMember, @Nullable InnerClassMember declaringInnerClass, MethodMember method);
 
--- a/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/ApiMemberSelector.java
+++ b/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/ApiMemberSelector.java
@@ -24,6 +24,7 @@ import org.objectweb.asm.FieldVisitor;
 import org.objectweb.asm.MethodVisitor;
 import org.objectweb.asm.ModuleVisitor;
 import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.RecordComponentVisitor;
 import org.objectweb.asm.TypePath;
 
 import java.util.LinkedHashSet;
@@ -123,6 +124,12 @@ public class ApiMemberSelector extends ClassVisitor {
             methods.add(methodMember);
             return new MethodVisitor(Opcodes.ASM9) {
                 @Override
+                public void visitParameter(@Nullable String name, int access) {
+                    methodMember.addParameter(new ParameterMember(name, access));
+                    super.visitParameter(name, access);
+                }
+
+                @Override
                 public AnnotationVisitor visitAnnotation(String desc, boolean visible) {
                     AnnotationMember ann = new AnnotationMember(desc, visible);
                     methodMember.addAnnotation(ann);
@@ -131,7 +138,10 @@ public class ApiMemberSelector extends ClassVisitor {
 
                 @Override
                 public AnnotationVisitor visitTypeAnnotation(int typeRef, @Nullable TypePath typePath, String desc, boolean visible) {
-                    TypeAnnotationMember ann = new TypeAnnotationMember(desc, visible, typeRef, typePath);
+                    // The extracted class writes the thrown exceptions sorted, so a THROWS type
+                    // reference of the original class points at the wrong exception.
+                    int writtenTypeRef = methodMember.mapTypeReferenceToWrittenExceptionOrder(typeRef);
+                    TypeAnnotationMember ann = new TypeAnnotationMember(desc, visible, writtenTypeRef, typePath);
                     methodMember.addTypeAnnotation(ann);
                     return new SortingAnnotationVisitor(ann, super.visitTypeAnnotation(typeRef, typePath, desc, visible));
                 }
@@ -176,6 +186,29 @@ public class ApiMemberSelector extends ClassVisitor {
             };
         }
         return null;
+    }
+
+    @Override
+    public RecordComponentVisitor visitRecordComponent(String name, String descriptor, @Nullable String signature) {
+        // Record components are always part of the API: their order determines the parameter order
+        // of the canonical constructor and the binding order of a record pattern.
+        RecordComponentMember recordComponent = new RecordComponentMember(name, descriptor, signature);
+        classMember.getRecordComponents().add(recordComponent);
+        return new RecordComponentVisitor(Opcodes.ASM9) {
+            @Override
+            public AnnotationVisitor visitAnnotation(String desc, boolean visible) {
+                AnnotationMember ann = new AnnotationMember(desc, visible);
+                recordComponent.addAnnotation(ann);
+                return new SortingAnnotationVisitor(ann, super.visitAnnotation(desc, visible));
+            }
+
+            @Override
+            public AnnotationVisitor visitTypeAnnotation(int typeRef, @Nullable TypePath typePath, String desc, boolean visible) {
+                TypeAnnotationMember ann = new TypeAnnotationMember(desc, visible, typeRef, typePath);
+                recordComponent.addTypeAnnotation(ann);
+                return new SortingAnnotationVisitor(ann, super.visitTypeAnnotation(typeRef, typePath, desc, visible));
+            }
+        };
     }
 
     @Override
--- a/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/ClassMember.java
+++ b/platforms/core-configuration/java-api-extractor/src/main/java/org/gradle/internal/tools/api/impl/ClassMember.java
@@ -30,6 +30,7 @@ public class ClassMember extends AnnotatableMember {
     private final String[] interfaces;
 
     private final List<String> permittedSubclasses = new ArrayList<>();
+    private final List<RecordComponentMember> recordComponents = new ArrayList<>();
 
     public ClassMember(int version, int access, String name, @Nullable String signature, @Nullable String superName, String @Nullable [] interfaces) {
         super(access, name, signature);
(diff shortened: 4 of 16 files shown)


big-guy:
@bot-gradle test this

PR #38847, opened 2026-08-15
WIP: XDCL (Gradle 9.7.1)

bamboo:
Requires a sibling ../xdcl checkout.

# PR — Gradle builds written in XDCL, with ecosystems shipped in the box

**Branch:** `bamboo/xdcl-scripting-language/9.7.1` → release (9.7.1)
**Title:** `feat: XDCL as a Gradle scripting language, with built-in ecosystem plugins`

**What you get.** A Gradle build written entirely in XDCL — no Groovy, no Kotlin script,
no build-logic — against schemas that ship *inside the distribution* and activate only
when the build opts in. The settings own everything the build gets things from (plugins,
toolchains, repositories); project files describe only the component:

```
// settings.gradle.xdcl                   // build.gradle.xdcl
settings {                                javaLibrary {
  plugins [                                 javaVersion 21
    { id "java-ecosystem" }                 dependencies {
  ]                                           implementation ["com.acme:lib:1.0"]
  dependencyResolutionManagement {          }
    repositories [:mavenCentral]          }
  }
}
```

`settings.gradle.xdcl` / `build.gradle.xdcl` route to the XDCL provider (`.gradle.xdcl`
wins file-precedence over `.gradle`/`.kts`); `java-ecosystem` is a distribution plugin
applied by bare id — no version, no repository. Authoring a *plugin* is declarative too:
`plugin-development-ecosystem` gives `xdclGradlePlugin { }` and `xdclGradlePluginKotlin { }`
project types, so plugin projects (standalone or included build-logic) carry no
`build.gradle[.kts]` either.

**Shipped ≠ active — the completion-scoping property.** Every built-in ecosystem rides in
the image, but a build only sees the schemas of the ecosystems it applied: name a template
from an unapplied one and evaluation fails, so completion never fills with types the build
can't use. The machinery is deliberately ordinary dependency resolution end to end: each
carrier's plugin descriptor declares its schema library as a `distribution-companion-module`
(a new, generic descriptor property core plugin resolution honors), an embedded Maven
repository at `<gradleHome>/repo` serves the published ecosystem libraries offline at the
distribution's exact version, and an `eachDependency` pin makes the running distribution
win over any published copy a third-party plugin compiled against — explained by
`dependencyInsight`, not silently dropped. Build evaluation and IDE sync consume one frozen
universe composition, so the LSP sees exactly what the build evaluated.

**What ships as ecosystems is deliberately minimal:** the `common` schema fragment
(dependencies + the `HasDependencies` trait), a minimal `java-ecosystem`, and
`plugin-development-ecosystem`. Three further POC pairs (checkstyle, instrumentation,
Groovy) proved schema composition and fragmentation across module boundaries and were then
removed; their binding surfaces were trimmed back out of the java ecosystem, and the
demo-era `org.gradle.demos.*` packages are renamed to `org.gradle.xdcl.ecosystem.*` (still
outside the public-API spec, with generated Kotlin DSL accessors opted out by marker).
Repositories are deliberately **not** part of the ecosystem surface: they are declared
once per build in the settings (`dependencyResolutionManagement`, bound by the provider in
Gradle's default `PREFER_PROJECT` mode), so the project types carry only the dependency
surface — no `HasRepositories`, no per-project `repositories`.

**How it behaves / not.**

- Project scripts contribute no plugins and no repositories — both live in settings only;
  a project-local plugins block is a located evaluation error.
- The Kotlin plugin-development toolchain is not bundled: settings declare
  `{ id "embedded-kotlin", apply false }` (an XDCL front-end alias, version filled in by
  the provider) or `org.jetbrains.kotlin.jvm` at an explicit version, plus the
  settings-level `repositories [:mavenCentral]` (neither toolchain configures
  repositories).
- Deferred, documented in the spec: dependency-verification exemption for the embedded
  repository, publishing reaction code for reuse, empty runtime variants of the published
  schema libraries, a per-project repository override (additive if ever needed).

**For reviewers.** 62 commits, 72 files, +4047/−127 on the 9.7.1 release base. The touch
on existing Gradle code is small and generic — `.gradle.xdcl` registration in
`ScriptingLanguages`, the `distribution-companion-modules` descriptor property and its
`CorePluginResolver` handling, the `distributionRepositoryOnly` packaging bucket — with
everything else in new `xdcl-*` modules and build-logic packaging conventions
(carrier manifests and the embedded repo are convention-derived, no hardcoded id lists).
Proven by 6 forking integ-test classes / 31 tests against the real distribution (language
smoke incl. defaults and included-build plugins, ecosystem activation, the embedded
repository and version pin, sync hand-off, Java and Kotlin plugin authoring — with
repositories declared at settings level throughout), plus the distribution content specs
with jar counts spelled out and size ceilings re-measured. Pairs with the sibling
org.xdcl checkout (`includeBuild("../xdcl")`, overridable via `xdcl-checkout.txt`); the
`embedded-kotlin` alias, the `dependencyResolutionManagement` settings surface, and the
reaction-dispatch scope-lock fix live on its `bamboo/declarative-plugin-skill` branch.


bamboo:
Superseded by #38870

PR #38849, opened 2026-08-15
Asodja/provider api perf

asodja:
## Reduce the cost of reading a Property

  Six optimisations to the read path of `Property`, `ListProperty`, `SetProperty` and `MapProperty`. No public API or behaviour change.

  ### What changed

  1. Cycle detection ran on every read. `AbstractProperty` opened an `EvaluationContext` scope for every `get()`, `isPresent()`, `calculateExecutionTimeValue()`, `getProducer()`
  and `visitContentProducerTasks()` — 3.4ns of a 5.1ns scalar read. A supplier that already holds its value evaluates nothing and runs no user code, so it cannot re-enter the property
  and the scope can never fire.

  2. `map {}` rebuilt the producer graph on every read. `TransformBackedProvider.beforeRead` called `getProducer()`, which for a collection property allocated a `ValueProducer`
  per collector plus a stream pipeline, only to find there was no producer task. Suppliers now walk their inputs directly.

  3. Map values were materialised twice. `MapProperty` built a `LinkedHashMap` and copied it into an `ImmutableMap`. It now builds straight into an `ImmutableMap.Builder` and
  finishes with `buildKeepingLast()`, which has the same last-wins semantics.

  4. Collection builders grew from their default capacity, now presized from a cheap `Collector.sizeHint()`.

  ### Which reads get faster

  The scope is skipped when the property's supplier already holds its value: an explicit or convention **constant**, **any finalized property** (`withFinalValue()` collapses almost
  every provider to a fixed value, so finalized task inputs qualify), and **a property wired to another such property** (`b.set(a)`).

  It is not skipped for a non-finalized provider-backed property, a collection property that still has collectors, or a build service provider — `RegisteredBuildServiceProvider`
  survives finalization and instantiates the service on read, so "finalized" alone is not sufficient grounds to skip cycle detection.

  Changes 2–4 are not limited to constants and speed up mapped and collection reads generally.

  ### Measurements

  JMH, JDK 25, 5 forks × (5 warmup + 5 measurement) × 1s, `-prof gc`. Percentages rounded down.

  ```
  Operation                 Property value                  before             after             time  alloc
  ────────────────────────  ──────────────────────────────  ─────────────────  ────────────────  ────  ─────
  property.isPresent()      set("value")                    4.835 ns / 0 B     0.978 ns / 0 B    -79%      —
  property.get()            set("value")                    5.083 ns / 0 B     1.212 ns / 0 B    -76%      —
  property.get()            convention("value")             5.294 ns / 0 B     1.199 ns / 0 B    -77%      —
  property.get()            set("value") + finalizeValue()  4.96 ns / 0 B      1.104 ns / 0 B    -77%      —
  property.get()            set(other), other.set("value")  19.86 ns / 24 B    6.119 ns / 24 B   -69%     0%
  property.get()            chain of 3 properties           24.873 ns / 24 B   12.194 ns / 24 B  -50%     0%
  property.get()            chain of 9 properties           78.669 ns / 24 B   64.763 ns / 24 B  -17%     0%
  property.map{}.get()      set("value")                    45.776 ns / 48 B   5.84 ns / ~0 B    -87%  -100%
  listProperty.get()        add("e") x10                    155 ns / 296 B     77.42 ns / 115 B  -50%   -61%
  listProperty.get()        add(provider) x10               204 ns / 512 B     111.9 ns / 336 B  -45%   -34%
  listProperty.map{}.get()  add("e") x10                    238.1 ns / 627 B   117.4 ns / 125 B  -50%   -80%
  listProperty.isPresent()  add("e") x10                    11.047 ns / 0 B    10.021 ns / 0 B    -9%      —
  mapProperty.get()         put("k","v") x10                309 ns / 1070 B    276.2 ns / 490 B  -10%   -54%
  mapProperty.get()         put("k", provider) x10          377.7 ns / 1288 B  193 ns / 704 B    -48%   -45%
  ```

  Notes:

  - Property-to-property wiring keeps its 24 B: the `Present` produced by the inner property escapes into the outer one's return value, so escape analysis cannot remove it. Only the
  scope cost goes away.
  - `chain of 9 properties` runs past the lookahead bound, so its outer links walk four suppliers, answer false, and open a scope anyway. It is still faster, because the inner links
  skip theirs.
  - `mapProperty.get()` with constant entries is the noisiest row (±37 ns); its allocation drop is the solid result. Isolating the map build step: 144.7 → 71.0 ns, 2× faster, 55% less
  allocation.
  - `—` in the `alloc` column means the operation allocated nothing before or after.

  Benchmarks included: `ProviderHotPathBenchmark`, `ScalarPropertyGetBenchmark`. Retained heap per property is unchanged.

  ### Notes for reviewers

  - `ValueSupplier.isSelfContained(int)` is a correctness surface: it must stay `false` for anything that can reach user code. The `int` is a lookahead bound, and any implementation
  that delegates must spend a link — a property can be reachable from its own supplier (`p.set(p)`, or `a.set(b)` with `b.set(a)`), so unbounded delegation would not terminate.
  Answering `false` is always safe.
  - `calculateValueFrom` and `calculateOwnExecutionTimeValue` lost their `EvaluationScopeContext` parameter — a token asserting a scope was open, which the fast path deliberately does
  not do. No implementation used it.
  - `MapCollector` / `MapEntryCollector` take an `ImmutableMap.Builder` instead of a `Map`. Internal, no usages outside the package.
  - `sizeHint()` deliberately does not reuse `Collector.size()`, which for a provider-backed collector realizes the whole upstream value. It is floored at the builder's default
  capacity, since presizing to an unknown contributor's hint of zero is worse than not presizing at all.

### DIFF
--- a/platforms/core-configuration/model-core/src/jmh/java/org/gradle/api/internal/provider/ProviderHotPathBenchmark.java
+++ b/platforms/core-configuration/model-core/src/jmh/java/org/gradle/api/internal/provider/ProviderHotPathBenchmark.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.gradle.api.internal.provider;
+
+import org.gradle.api.provider.ListProperty;
+import org.gradle.api.provider.MapProperty;
+import org.gradle.api.provider.Property;
+import org.gradle.api.provider.Provider;
+import org.gradle.internal.evaluation.EvaluationContext;
+import org.gradle.internal.evaluation.EvaluationOwner;
+import org.gradle.internal.evaluation.EvaluationScopeContext;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Benchmarks the query hot paths of properties whose values are constants.
+ *
+ * <p>These are the shapes that dominate a real build: a property assigned a constant in a build
+ * script, and the same property read through a user {@code map {}} chain. Run with {@code -prof gc}
+ * to see the per-read allocation, which is what most of these cases are really measuring.
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Mode.AverageTime)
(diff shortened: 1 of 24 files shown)


asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test this APT

asodja:
@bot-gradle test APT

asodja:
@bot-gradle test this

PR #38880, opened 2026-08-18
Add ADR capturing all public API requirements

ghale:
This ADR captures all of the existing public API requirements spread across other ADRs and architecture tests into a coherent list that can be referenced by authors and agents.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/architecture/standards/0013-requirements-for-public-api-classes.md
+++ b/architecture/standards/0013-requirements-for-public-api-classes.md
@@ -14,7 +14,7 @@ There is no single document a contributor or reviewer can consult to answer "wha
 This ADR consolidates those requirements into one checklist.
 It does not propose new rules; it collects the existing ones and records the expectations that all public API classes should meet.
 
-Throughout this ADR, "public API" means a type that resides in one of the packages designated as public API (see `PublicAPI.kt`) together with its `public` and `protected` members.
+Throughout this ADR, "public API" means a type that resides in one of the packages designated as public API — see `PublicApi.kt` and `PublicKotlinDslApi.kt` in `build-logic-commons/basics/src/main/kotlin/gradlebuild/basics/` — together with its `public` and `protected` members.
 
 The keywords MUST, MUST NOT, SHOULD, SHOULD NOT and MAY are used as defined in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
 
--- a/architecture/standards/0013-requirements-for-public-api-classes.md
+++ b/architecture/standards/0013-requirements-for-public-api-classes.md
@@ -27,7 +27,8 @@ They fall into two groups: requirements enforced automatically by architecture t
 
 1. **Signatures reference only approved types.**
    Public API methods MUST NOT reference Gradle internal types as parameter or return types.
-   Permitted types are other Gradle public API types, primitives, a curated allow-list of JDK types (e.g. `java.lang`, `java.util`, `java.io`, `java.nio.file`, `java.time`, `URI`, `URL`, `Duration`, `BigDecimal`), and a small set of Kotlin and slf4j types (e.g. `KClass`, `KProperty`, `Pair`, `Unit` and `Marker`).
+   The permitted types are other Gradle public API types, primitives, and a curated allow-list of JDK, Kotlin, and slf4j types.
+   The authoritative allow-list is defined in `testing/architecture-test/src/test/java/org/gradle/architecture/test/PermittedPublicApiTypes.java`.
    Notably, `java.util.function` types are intentionally excluded — Gradle exposes its own functional types such as `org.gradle.api.Action`, `org.gradle.api.specs.Spec`, and `org.gradle.api.Transformer` instead.
 
 2. **No Groovy types; Closure-taking methods have a Gradle-type equivalent.**
--- a/architecture/standards/0013-requirements-for-public-api-classes.md
+++ b/architecture/standards/0013-requirements-for-public-api-classes.md
@@ -25,8 +25,6 @@ They fall into two groups: requirements enforced automatically by architecture t
 
 ### Structural requirements
 
-These are the requirements that *must* be met for public API classes:
-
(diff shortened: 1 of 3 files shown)


architecture/standards/0013-requirements-for-public-api-classes.md
```diff
@@ -0,0 +1,110 @@
...
+    Every public API type and member MUST carry an explicit `@since` tag recording the Gradle version in which it was introduced, per [ADR-0012](0012-since-on-public-api.md).
+
+13. **American English.**
+    All documentation and identifiers MUST use American English spelling, per [ADR-0009](0009-use-american-english.md).
+
+### Release notes (enforced during API review)
+
+14. **Reflect public API changes in the release notes.**
+    Any change to the public API — a new type or member, a deprecation, or a behavioral change — SHOULD be reflected in the release notes for the corresponding release.
+    New capabilities and deprecations in particular MUST be called out, since these notes are the primary way users learn what changed in the API they depend on.
```
mlopatkin: ❓ This looks like a new rule, though, especially deprecations, which go to the upgrade guide instead.

architecture/standards/0013-requirements-for-public-api-classes.md
```diff
@@ -0,0 +1,110 @@
...
+
+Gradle's public API is a contract.
+Build authors and plugin authors — writing in Java, Kotlin, Groovy, and other JVM languages — depend on it being stable, consistent, usable, and well documented.
+Once a type or member is published as public API, it is expensive to change, so these classes should be held to a high standard.
+
+There is no single document a contributor or reviewer can consult to answer "what does a class have to do to be part of the public API?".
+This ADR consolidates those requirements into one checklist.
+It does not propose new rules; it collects the existing ones and records the expectations that all public API classes should meet.
+
+Throughout this ADR, "public API" means a type that resides in one of the packages designated as public API (see `PublicAPI.kt`) together with its `public` and `protected` members.
```
alllex: This misses `PublicKotlinDslApi.kt`

architecture/standards/0013-requirements-for-public-api-classes.md
```diff
@@ -0,0 +1,110 @@
...
+The keywords MUST, MUST NOT, SHOULD, SHOULD NOT and MAY are used as defined in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Decision
+
+A public API class MUST satisfy all of the following requirements.
+They fall into two groups: requirements enforced automatically by architecture tests (a violation fails the build), and requirements enforced during API review (a violation blocks the PR).
+
+### Structural requirements
+
+These are the requirements that *must* be met for public API classes:
```
alllex: 💅  Seems like a duplicate of the first line of the Decision, and also uses `*must*` instead of `MUST` spelling. Seems like it's okay to simply drop it.

architecture/standards/0013-requirements-for-public-api-classes.md
```diff
@@ -0,0 +1,110 @@
...
+
+1. **Signatures reference only approved types.**
+   Public API methods MUST NOT reference Gradle internal types as parameter or return types.
+   Permitted types are other Gradle public API types, primitives, a curated allow-list of JDK types (e.g. `java.lang`, `java.util`, `java.io`, `java.nio.file`, `java.time`, `URI`, `URL`, `Duration`, `BigDecimal`), and a small set of Kotlin and slf4j types (e.g. `KClass`, `KProperty`, `Pair`, `Unit` and `Marker`).
```
alllex: ❓ What would be the plan for when this list evolves? Should it be updated inline here?

The wording here is not prescriptive and uses `e.g.`, hinting at the allowed incompleteness. If such a list exists and it is a hard requirement, we should link to the location of that list. If we don't have it, maybe we should introduce it as a standalone file.

architecture/standards/0013-requirements-for-public-api-classes.md
ghale: Added.

architecture/standards/0013-requirements-for-public-api-classes.md
ghale: Removed.

architecture/standards/0013-requirements-for-public-api-classes.md
ghale: Yeah, I used "e.g." simply because it was not the source of truth and the items might change, but you're right, we should be less passive here.  

I think it makes sense to pull this out into its own file.  It's not ideal that this information only resides in a test.  I've added `PermittedPublicApiTypes` which captures the permitted types and acts as the source of truth for this information.  Currently, it resides in `architecture-test` but we can pull this up into a more common place if we ever need to reference it elsewhere.  The ADR now references this file and does not enumerate any of the values.

testing/architecture-test/src/test/java/org/gradle/architecture/test/PermittedPublicApiTypes.java
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * JDK packages whose types are all permitted in public API method signatures.
+     *
+     * <p>Note: {@code java.util.function} is intentionally excluded. Gradle exposes its own
+     * functional types such as {@link org.gradle.api.Action}, {@link org.gradle.api.specs.Spec},
+     * and {@link org.gradle.api.Transformer} instead. Mixing these custom types with
+     * {@code java.util.function} types would make the public API harder to use, especially for
+     * plugin authors.
+     */
+    public static final List<String> PERMITTED_JDK_PACKAGES = Arrays.asList(
```
mlopatkin: 💅 `Arrays.asList` is mutable; `List.of` should be available here, let's use it instead (here and in the other field declarations in the file).

PR #38883, opened 2026-08-18
filter out variants lacking a requested attribute when disambiguating

wakingrufus:
### Context
This prevents a counter-intuitive ambiguity error when 2 variants have extra unrequested attributes, but 1 has more matched attributes. this is only an issue when the variant with more matched attributes has a "peer" variant which contains the same attributes, except some are only _compatible_, not exact matches (eg java-api and java-runtime) as this case causes the variant we want to not be considered a superset, and therefore the superset rule does not kick in and we drop down to disambiguation

Since this new filtering happens as part of the disambiguating process, this should not result in a failure to find a variant, as these are processed in precedence order, and will stop once there is only 1 candidate left

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/attributes/matching/DefaultAttributeMatcherTest.groovy
```diff
@@ -249,7 +249,7 @@ class DefaultAttributeMatcherTest extends Specification {
         expect:
         matcher.matchMultipleCandidates([candidate1, candidate2, candidate3, candidate4, candidate5, candidate6], requested) == [candidate5]
         matcher.matchMultipleCandidates([candidate1, candidate2, candidate3, candidate4, candidate6], requested) == [candidate1, candidate3, candidate4, candidate6]
-        matcher.matchMultipleCandidates([candidate1, candidate2, candidate4, candidate6], requested) == [candidate1, candidate4, candidate6]
+        matcher.matchMultipleCandidates([candidate1, candidate2, candidate4, candidate6], requested) == [candidate1, candidate6]
```
wakingrufus, line 252: this behavior changed. I'll admit I don't understand why we would have ever wanted candidate4 in this case, so maybe this is OK?

ljacomet:

big-guy:
We need to take a look at this deeper.

PR #38885, opened 2026-08-18
Honor Maven settings.xml mirrors 

big-guy:
### Context

The whole feature is behind an off-by-default Gradle property, `org.gradle.mirror.maven.settings`, implemented as a build option (`StartParameterBuildOptions.SharedMavenMirrorSettingsOption`). With it unset, `mirrorFor()` returns immediately: no settings are read and no configuration cache input is registered. Being a build option, it is set via `gradle.properties` or `-D`, not `-P`.

The user-facing story is documented in `centralizing_repositories.adoc`, under "Reusing mirror settings from Maven", with a runnable sample. Enabling the feature logs `Reusing Maven mirror settings is an incubating feature.` once per build.

### What it does

- **Parses the settings in one place.** `MavenSettingsProvider` is the only thing that reads `settings.xml`. It owns the mirror model (`MirroredRepository`, `MirrorCredentials`, `MirrorHttpHeader`), resolves and decrypts credentials, and caches the result. `MavenMirrorResolver` only matches repositories against what the provider produced.
- **Applies Maven's real `mirrorOf` semantics** — `*`, exact ids, `!id` negation, comma lists, `external:*`, `external:http:*` — ported from `org.eclipse.aether.util.repository.DefaultMirrorSelector`. A repository's "effective id" is `central` when its URL is Maven Central's, otherwise its Gradle repository name.
- **Rewrites the repository URL** in `DefaultMavenArtifactRepository.validateUrl()`, and fails the build for a mirror marked `<blocked>true</blocked>`.
- **Never sends the original repository's credentials to the mirror.** Mirror credentials come from the `<server>` entry matching the mirror id, including Maven-encrypted passwords decrypted with the master password from `settings-security.xml` — following a `<relocation>` to another file, and declaring every file in that chain as a build input. `httpHeaders` on the server entry are supported (first header only).
- **`mavenLocal()` is never mirrored** — by rule rather than by wiring. The resolver is injected like any other Maven repository and declines, because `mirrorFor` only rewrites `http`/`https` URLs and checks the scheme before reading any settings.
- **Everything is build-tree scoped** (`MavenFileLocations`, `MavenSettingsProvider`, `MavenMirrorResolver`), so a composite parses `settings.xml` once for the whole tree.
- **Configuration cache**: the settings files are declared as build inputs via `FileResourceListener.fileObserved`, the same mechanism dependency-management already uses for `verification-metadata.xml` and dependency lock files. `settings-security.xml` is declared only on the decryption path, since the settings builder never reads it. The feature flag is part of the cache *key* rather than a fingerprint input, because with the feature off there is no input that turning it on could invalidate.

### Interactions with Gradle's own behaviour

Three cases where a mirror could quietly override something Gradle already guarantees. Each has an integration test that was written failing first.

- **Publishing is unaffected.** Mirrors apply to resolution only, as in Maven. `getTransport()` keeps declared-URL semantics for publishing and resolution uses a separate path. Before this, publishing to a repository matched by `mirrorOf=*` sent the *mirror's* credentials to the deployment host and 401'd.
- **`allowInsecureProtocol` means what it says.** The insecure-protocol check now runs against the URL actually contacted, so an `http` mirror of an `https` repository fails without an opt-in. Previously the verifier was built from the declared URL, so the mirror bypassed the check entirely.
- **`maven-default-http-blocker` is ignored.** Maven ships it in its own `conf/settings.xml`; it matches on the declared repository URL, which Gradle already refuses unless the build opts in. Honouring it added no protection while overriding that opt-in, and only on machines with `M2_HOME` set. Mirrors *you* declare as blocked are still honoured.

### What it deliberately does not do

Summarised in the user guide. The main ones: only the first `<httpHeaders>` entry is applied, Maven's `settings.security` system property is not honoured, `<mirrorOfLayouts>` is ignored, the Maven 4 encryption format is unsupported, and the deprecated `artifactUrls` hosts are not mirrored.

Two reporting details worth a reviewer's attention. `getUrl()` deliberately keeps returning the *declared* URL, so reading back what you configured never lies — but `createDescriptor()` goes through the same `validateUrl()` as resolution, so build scans and error messages show the *mirror's* URL against your repository's name. And that descriptor still carries `setAuthenticated(usesCredentials())`, which reports the original repository's auth state, so a repository with no credentials of its own mirrored to an authenticated mirror reports `authenticated: false`.

`buildSettings()` also passes `System.getProperties()` to Maven, which interpolates system properties into the settings; that is not tracked as an input.

### Gaps before this could be a real feature

1. **Auth**: mostly closed — leak prevention, `<server>` credentials with decryption including relocated master passwords, first-header `<httpHeaders>` mapping. Remaining: multiple headers per mirror, the Maven 4 format, `settings.security`, and a Gradle-native credential override if one is wanted (it needs designing as a documented convention, not an undocumented fallback).
2. **`mirrorOf` matching**: closed — full Maven grammar, effective ids, blocked mirrors. Remaining: the name-matching convention needs a real design discussion before anything user-visible ships, plus diagnostics for patterns that match nothing.
3. **Repository identity**: see the reporting details above. A mirror-aware descriptor should probably carry both URLs and the auth state actually in use.
4. **Surface**: environment-variable/DSL opt-in, interaction with `dependencyResolutionManagement` repositories and repository content filtering, Ivy repositories.

Both issues found in the original review of this branch are now fixed: a mirror `<url>` with no scheme reports the unusable protocol instead of NPEing inside credential resolution, and the null-username NPE is moot with `NTLMCredentials` going away.

### Test coverage notes

`MavenSettingsMirrorIntegrationTest` (26 tests) passes under the `embedded`, `configCache` and `isolatedProjects` executers. Several properties were verified by mutation rather than assumed — each fix below fails its test when the fix is reverted:

- Replacing the declared configuration cache input with an unrelated file kills exactly the invalidation tests and leaves the rest green.
- A multi-project test checks that editing `settings.xml` re-mirrors *every* project, not only whichever one configured first and populated the cache.
- Rotating a master password behind a `<relocation>` invalidates the cache; stop walking the chain and that test fails.
- Tests that distinguish one mirror from another do it at the HTTP layer — only the mirror in force is served — so reaching for the wrong one fails the build rather than merely logging.

The documented sample is checked from both sides, so neither check can pass vacuously:

- The snippet's own check runs it as a reader would find it, with no Maven settings, and pins the resulting `FAILED` dependency report in `dependencies.out`. `dependencies` reports unresolved entries without failing, so without that file the check passed while resolving nothing.
- `MavenSettingsMirrorSampleIntegrationTest` takes the same sample through `@UsesSample`, installs `m2_home/settings.xml` as the Maven settings of an isolated home, and asserts the dependency resolves with no `FAILED`, for both DSLs. Removing the enabling property makes both fail. It needs the network, since the sample mirrors to Maven Central, and is guarded by the `Online` precondition.

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


platforms/documentation/docs/src/docs/userguide/reference/core-plugins/publishing_maven.adoc
```diff
@@ -178,6 +178,92 @@ This change also marks the deprecation of the OSSRH (OSS Repository Hosting) ser
 Going forward, publishing to Maven Central **requires a dedicated plugin**.
 You can learn more by watching link:https://youtu.be/nd2ULXyBaV8[this video].
 
+[[publishing_maven:maven_settings_mirrors]]
+=== Reusing Maven mirror settings
```
big-guy:
```suggestion
=== Reusing mirror settings from Maven
```

platforms/documentation/docs/src/docs/userguide/reference/core-plugins/publishing_maven.adoc
```diff
@@ -178,6 +178,92 @@ This change also marks the deprecation of the OSSRH (OSS Repository Hosting) ser
...
+repository manager by declaring a `<mirror>` in `~/.m2/settings.xml`. Gradle can read those same
+mirrors instead of requiring the configuration to be duplicated in your build. It is meant for
+exactly that case — a team with a working Maven mirror setup that wants Gradle builds to honour it
+— and not as a general-purpose way to redirect repositories. If you are not already invested in
+Maven's `settings.xml`, prefer configuring repositories directly in your build, as described in
+link:https://blog.gradle.org/maven-central-mirror[Good Neighbors: How to Reduce Maven Central
+Traffic from Gradle Builds].
+
+Enable it with the `org.gradle.mirror.maven.settings` property, in `gradle.properties` or on the
+command line as a system property:
```
big-guy:
```suggestion
Organizations that already build with Maven often use an internal
repository manager by declaring a `<mirror>` in `~/.m2/settings.xml`. Gradle can read those same mirrors instead of requiring the configuration to be duplicated in your build. This is meant for teams with a working Maven mirror setup to reuse those mirrors in Gradle. This is not a general-purpose way to redirect repositories. 

Gradle will only replace HTTP/HTTPS Maven repositories that match mirror declarations in Maven settings. Local file, S3, GCS and Ivy repositories are unaffected. Maven repositories can be used for build script dependencies as well as project dependencies, so the Gradle Plugin Portal is also affected by Maven mirrors. 

If you are not already using mirrors defined in 
Maven's `settings.xml`, prefer configuring repositories directly in your build or setting up a mirror init script, as described in
link:https://blog.gradle.org/maven-central-mirror[Good Neighbors: How to Reduce Maven Central Traffic from Gradle Builds].

Enable Maven mirror reuse with the `org.gradle.mirror.maven.settings` property, in `gradle.properties` or on the command line as a system property:
```

platforms/documentation/docs/src/docs/userguide/reference/core-plugins/publishing_maven.adoc
```diff
@@ -178,6 +178,92 @@ This change also marks the deprecation of the OSSRH (OSS Repository Hosting) ser
...
+link:https://maven.apache.org/settings.html[the Settings Reference] for the file as a whole.
+
+[[publishing_maven:maven_settings_mirrors_publishing]]
+==== Publishing is not affected
+
+Mirrors apply to *resolution only*. Maven itself never mirrors deployment, and neither does
+Gradle: publishing always uploads to the URL declared on the repository, using that repository's
+own credentials, even when a mirror matches it. A `mirrorOf` of `*` does not redirect your
+`publish` tasks.
+
```
big-guy:
```suggestion
Mirrors apply to *resolution only*. Publishing always uploads to the repository URL declared in the build.

```

platforms/documentation/docs/src/docs/userguide/reference/core-plugins/publishing_maven.adoc
```diff
@@ -178,6 +178,92 @@ This change also marks the deprecation of the OSSRH (OSS Repository Hosting) ser
...
+[[publishing_maven:maven_settings_mirrors_unsupported]]
+==== Unsupported Maven settings
+
```
big-guy:
```suggestion
==== Unsupported Maven mirror settings

```

platforms/documentation/docs/src/docs/userguide/reference/core-plugins/publishing_maven.adoc
```diff
@@ -178,6 +178,92 @@ This change also marks the deprecation of the OSSRH (OSS Repository Hosting) ser
...
+Only mirrors and the `<server>` entries that back them are read. In particular, the following are
+not supported:
```
big-guy:
```suggestion
Only mirrors and the `<server>` entries that back them are supported. All features may not be supported:
```

platforms/documentation/docs/src/docs/userguide/reference/core-plugins/publishing_maven.adoc
```diff
@@ -178,6 +178,92 @@ This change also marks the deprecation of the OSSRH (OSS Repository Hosting) ser
...
+  Gradle already refuses plain HTTP repositories unless the repository sets
+  `allowInsecureProtocol`. Mirrors you declare as `<blocked>` yourself are honoured and fail
+  resolution of the repositories they match.
+* *`<mirrorOfLayouts>`.* Gradle has a single repository layout, so the layout filter is ignored.
+* *Everything else in `settings.xml`* — profiles, proxies, `<pluginGroups>`, `<localRepository>`
+  for anything other than `mavenLocal()`, and interactive password prompting.
+
+A repository matched by a mirror keeps its Gradle name, but its resolved URL, and therefore its
+identity in build scans and the dependency cache, becomes the mirror's. Turning the property on or
+off changes that identity and re-downloads artifacts.
```
big-guy:
```suggestion
* *Multiple HTTP headers per server.* Only the first `<httpHeaders>` entry is supported by Gradle.
* *Relocating `settings-security.xml`.* Only the default location for `settings-security.xml` is checked.
* *`<mirrorOfLayouts>`.* Gradle has a single repository layout, so the layout filter is ignored.
* *The Maven 4 password encryption format.* Only the Maven 3 format produced by
  `mvn --encrypt-password` is understood.
* *Everything else in `settings.xml`* Profiles, proxies, and `<pluginGroups>` do not apply to Gradle. 

```

big-guy:
@bot-gradle test this

big-guy:
@bot-gradle test this

blindpirate:
❓ The PR description says "`getUrl()` and the repository descriptor still report the original URL", but `createDescriptor()` calls `validateUrl()`, which now applies the mirror — so the descriptor (and build scans) carry the *mirror* URL. The user guide added here documents the opposite of the description ("its identity in build scans and the dependency cache, becomes the mirror's") and matches the code.

Which is intended? This decides repository identity in build scans and the dependency cache, so it's worth settling before the approach is finalized. If the descriptor should keep the declared URL, mirror application needs to move from `validateUrl()` into `createResolver()`.

big-guy:
@bot-gradle test rfr

big-guy:
> Which is intended? This decides repository identity in build scans and the dependency cache, so it's worth settling before the approach is finalized.

Thanks for noticing this. The PR description was heavily outdated.

big-guy:
@bot-gradle test this

PR #38888, opened 2026-08-19
Protect against deadlock from null command/arguments

octylFractal:
We disallow null upfront for a clearer outcome there, and also harden setEndStateInfo so it no longer deadlocks on certain fatal errors

Found while investigating https://github.com/gradle/gradle-private/issues/5316

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-runtime/process-services-base/src/main/java/org/gradle/process/internal/DefaultExecHandle.java
+++ b/platforms/core-runtime/process-services-base/src/main/java/org/gradle/process/internal/DefaultExecHandle.java
@@ -216,16 +216,9 @@ public class DefaultExecHandle implements ExecHandle, ProcessSettings {
     }
 
     private void setEndStateInfo(ExecHandleState newState, int exitValue, Throwable failureCause) {
+        ExecHandleState currentState = getState();
         ExecResultImpl newResult = null;
         try {
-            ExecHandleState currentState;
-            lock.lock();
-            try {
-                currentState = this.state;
-            } finally {
-                lock.unlock();
-            }
-
             newResult = new ExecResultImpl(exitValue, execExceptionFor(failureCause, currentState), displayName);
 
             ShutdownHooks.removeShutdownHook(shutdownHookAction);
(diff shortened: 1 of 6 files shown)


platforms/core-runtime/process-services-base/src/main/java/org/gradle/process/internal/DefaultExecHandle.java
```diff
@@ -211,31 +216,38 @@ private boolean stateIn(ExecHandleState... states) {
...
+            newResult = new ExecResultImpl(exitValue, execExceptionFor(failureCause, currentState), displayName);
+
+            ShutdownHooks.removeShutdownHook(shutdownHookAction);
+            buildCancellationToken.removeCallback(shutdownHookAction);
+
+            if (!currentState.isTerminal() && newState != ExecHandleState.DETACHED) {
+                try {
+                    broadcast.getSource().executionFinished(this, newResult);
+                } catch (Exception e) {
+                    newResult = new ExecResultImpl(exitValue, execExceptionFor(e, currentState), displayName);
```
abstratt, line 231: 💭 

`execExceptionFor` (actually, `failureMessageFor`)  should be made fault tolerant or else `newResult` would potentially stay null

platforms/core-runtime/process-services-base/src/test/groovy/org/gradle/process/internal/DefaultExecHandleSpec.groovy
```diff
@@ -174,6 +174,20 @@ class DefaultExecHandleSpec extends ConcurrentSpec {
...
+        execHandle.start()
+
+        then:
+        def e = thrown(ProcessExecutionException)
+        e.message == "A problem occurred starting process 'awesome'"
+        execHandle.state == ExecHandleState.FAILED
+    }
+
     void "aborts process"() {
         def execHandle = handle().args(args(SlowApp.class)).build()
```
abstratt, line 213: 💭 

This test shows we may end up losing the exception if building a message itself can fail.

```
void "loses the failure when the error message cannot be built"() {
    given:
    // Make any command line count as "too long", so failureMessageFor() inspects the cause chain...
    System.setProperty("org.gradle.internal.cmdline.max.length", "1")

    // ...and fail the start with an exception whose getMessage() is null.
    // connectStreams() runs before started(), so the handle is still STARTING.
    def streamsHandler = Stub(FinishNotifyingStreamsHandler) {
        connectStreams(_, _, _) >> { throw new RuntimeException() }
    }
    def execHandle = handle().setDisplayName("awesome").streamsHandler(streamsHandler).build()

    when:
    execHandle.start()

    then:
    // BUG: no exception is thrown
    thrown(ProcessExecutionException)

    cleanup:
    System.clearProperty("org.gradle.internal.cmdline.max.length")
}
```

abstratt (approved):
Posted a couple of suggestions/questions

PR #37793, opened 2026-05-05
Remove IDE file generation tasks

jbartok:
Finishes:
* https://github.com/gradle/gradle/pull/37636

### Context
# Remove the file generation tasks of the IDE plugins

This is the Gradle 10 removal of everything deprecated by #37636.

#### What is removed

- The 8 file generation task classes (`GenerateIdeaModule`, `GenerateIdeaProject`, `GenerateIdeaWorkspace`, `GenerateEclipseProject`, `GenerateEclipseClasspath`, `GenerateEclipseJdt`, `GenerateEclipseWtpComponent`, `GenerateEclipseWtpFacet`) and all lifecycle/clean/`openIdea` task wiring — the `idea` and `eclipse` plugins are now model-only.
- The deprecated model members: `idea.module.iml`, `idea.project.ipr`, `idea.workspace`, the `pathFactory` properties, `eclipse.jdt.javaRuntimeName` and the `file { }` hooks on `EclipseJdt`.
- The `eclipse.wtp.facet` DSL and the file-merging hooks on `eclipse.wtp.component` (but **not** `eclipse.wtp` itself - see finding 1).
- Public API orphaned by the removal: `IdeaModel.targetVersion`, `IdeaModel.pathVariables`, `IdeaModule.pathVariables`, `IdeaProject.projectLibraries`, and the XML domain types `Module`, `Project`, `Workspace`, `ProjectLibrary` (idea) and `Jdt`, `Facet`, `WtpFacet`, `WtpComponent`, `WbDependentModule` (eclipse).
- `withXml` and the XML transformer accessors on `XmlFileContentMerger` (see finding 5); `beforeMerged`/`whenMerged` on the eclipse project and classpath models survive, as the Tooling API model builder applies them.

Test coverage for still-relevant behavior was converted to Tooling API cross-version tests (`r50`–`r83` specs) before the removal.

#### Findings while implementing

1. **The wholesale `eclipse.wtp` deprecation in #37636 was over-broad.** The `eclipse.wtp.component` configuration feeds the WTP classpath attributes (`org.eclipse.jst.component.dependency` / `nondependency`) surfaced via the Tooling API, which Buildship consumes. Only the facet DSL and the file-merging hooks match the deprecation's stated scope. The 9.x deprecations were re-scoped accordingly (see #38498).

[...]

PR #38889, opened 2026-08-19
Skip Scala configuration in IdeaPlugin with Isolated Projects, leave it only for project generator tasks

### LINKED ISSUE #27363: Support fetching Idea models for Scala projects with Isolated Projects
Examples of current violations are mostly related to the `IdeaScalaConfigurer`

(1) https://github.com/gradle/gradle/blob/6ca2a07a039037459da0f7c738729cf844d39d51/platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/idea/IdeaPlugin.java#L515

(2) https://github.com/gradle/gradle/blob/96280d6058595297a7042bd72b15614fb8b56542/platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/idea/internal/IdeaScalaConfigurer.java#L73

See an aspirational test in the `IsolatedProjectsToolingApiIdeaProjectIntegrationTest`

fru1tworld:
Fixes #27363

`IdeaScalaConfigurer` requires cross-project access that violates Isolated Projects. Since it only configures the deprecated `gradle idea` task and IntelliJ uses the Tooling API path, skip it instead of failing.


finalchild:
Looks like #37793 will remove the blocker altogether in Gradle 10. Do we need fix in Gradle 9? Maybe

ljacomet:

alllex:
At first glance, the short-circuiting of the logic under Isolated Projects (early return) doesn't seem like the right approach, because we'd be changing behavior between the modes.

If the PR targeting Gradle 10 linked above is actually removing the offending code in question, it might be better to wait, given that the major release is scheduled for just later this year.

Having said that, if we have the capacity until then, we'll investigate this further. At the same time, it's not our current focus area, so it might not happen soon.

ov7a:
@bot-gradle test this

ov7a:
I've updated the PR to preserve the behavior for generation tasks until [they are removed](https://github.com/gradle/gradle/pull/37793) in Gradle 10.

The original fix may result in the generated project files being incomplete/invalid.

Below is the status of the build for master, original (full skip) and my update:

invocation | master, no IP | master IP | original, no IP | original IP | update no IP | update IP
-- | -- | -- | -- | -- | -- | --
TAPI import (IntelliJ) | ✅ | ❌  | ✅ | ✅ | ✅ | ✅
help / build | ✅ | ❌  | ✅ | ✅ | ✅ | ✅
gradle idea | ✅ | ❌  | ✅ | ⚠️ no Scala SDK | ✅ | ❌
:sub:ideaModule | ✅ | ❌  | ✅ | ⚠️ no Scala SDK | ✅ | ❌
:ideaProject, single project | ✅ | ❌ | ✅ | ⚠️ no Scala SDK | ✅ | ❌
:ideaProject alone, multi-project | ✅ | ❌  | ✅ | ❌ | ✅ | ❌ 

where ✅ - build succeeds, ❌ - build fails with an error, ⚠️ - build succeeds, but with invalid/incorrect state

reinsch82:
@bot-gradle test this

PR #38890, opened 2026-08-19
Parse HTTP directory listings without jsoup

big-guy:
### Context

`resources-http` was the only place jsoup was used in the Gradle distribution, in a single call site: `ApacheDirectoryListingParser`, which reads the HTML directory listing an Ivy repository served over HTTP returns when resolving a dynamic version. Dropping it removes jsoup from the distribution entirely.

Only the `href` of each anchor is of interest, so a document tree is more machinery than the job needs. The new `HtmlAnchorHrefScanner` scans the markup for start tags instead, which is indifferent to how the surrounding elements nest and so needs no position on how to repair the ones that don't. It skips comments, doctypes, CDATA sections and the content of raw text elements such as `<script>`, accepts single-quoted and unquoted attribute values, and resolves the character references that can appear in a URL.

### How this was validated

Beyond the unit tests, the scanner was compared directly against jsoup:

- **The three repository listing fixtures in this repo** (Artifactory, Maven Central, Nexus): 22 anchors each, **reproduced exactly**.
- **30,000 fuzzed documents built from well-formed fragments**: 2,796 differ (9.3%), and every one falls into two classes, both artifacts of jsoup's tree building rather than of the markup:
  - 2,778 where jsoup **duplicates** an anchor. Leaving an `<a>` unclosed across a block element makes jsoup's active-formatting-elements reconstruction clone it, so the entry is listed twice. Deduplicated, the two results are identical.
  - 18 where jsoup **reorders** anchors foster-parented out of a `<table>`. The scanner keeps document order.

  Neither changes which entries a listing yields. `ResourceVersionLister` collects results into a `List` that already admits duplicates from multiple patterns, and version selection compares with a version comparator, so neither count nor order is semantically meaningful downstream.

One deliberate divergence, for malformed markup: where a raw text element such as `<script>` is **never closed**, jsoup treats the rest of the document as text, which would silently discard every remaining link. The scanner reads the remainder as markup instead — recovering links from already-broken markup seemed better than losing them. A memo of which tags were found unterminated keeps that recovery linear rather than quadratic on a document full of unclosed tags (covered by a test).

### On using a JDK parser instead

`javax.swing.text.html.parser.ParserDelegator` is the only HTML parser in the JDK, and it was measured too. It matches jsoup on the three real fixtures, but on the fuzz corpus it differs from jsoup on 27% of documents versus this scanner's 9.5%: 849 where a valueless `href` comes back as the sentinel string `#DEFAULT`, and 497 structural, which include **silently dropping** anchors that appear inside a `<table>` outside a row or cell (its HTML 3.2 content model). A version vanishing from resolution with no error is the worst available failure mode here. It also lives in `java.desktop`, which no Gradle production code currently depends on, and `resources-http` is on the client path. The `javax.xml` parsers are not an option — real listings contain `<hr>`, unquoted attributes and `&nbsp;`.

### Contributor Checklist
- [x] Provide integration tests (under `<subproject>/src/integTest`) to verify changes from a user perspective. — existing HTTP directory-listing integ tests cover this path end to end; no behaviour change to add tests for
- [ ] Update User Guide, DSL Reference, and Javadoc for public-facing changes. — n/a, internal
- [ ] Ensure that tests pass locally: `./gradlew <changed-subproject>:quickTest`. — ran `:resources-http:test` (331 tests), the HTTP directory-listing integ tests in `:dependency-management` (76 tests), `checkstyleMain`, `codenarcTest` and `projectHealth`; relying on CI for the full `quickTest`

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/resources-http/src/main/java/org/gradle/internal/resource/transport/http/ApacheDirectoryListingParser.java
+++ b/platforms/software/resources-http/src/main/java/org/gradle/internal/resource/transport/http/ApacheDirectoryListingParser.java
@@ -16,11 +16,9 @@
 
 package org.gradle.internal.resource.transport.http;
 
+import org.apache.commons.io.IOUtils;
 import org.gradle.api.resources.ResourceException;
 import org.gradle.internal.resource.UriTextResource;
-import org.jsoup.Jsoup;
-import org.jsoup.nodes.Document;
-import org.jsoup.select.Elements;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -33,11 +31,12 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
 
 public class ApacheDirectoryListingParser {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ApacheDirectoryListingParser.class);
+
+    private final HtmlAnchorHrefScanner anchorScanner = new HtmlAnchorHrefScanner();
 
     public List<String> parse(URI baseURI, InputStream content, String contentType) throws Exception {
         baseURI = addTrailingSlashes(baseURI);
@@ -45,13 +44,13 @@ public class ApacheDirectoryListingParser {
             throw new ResourceException(baseURI, String.format("Unsupported ContentType %s for directory listing '%s'", contentType, baseURI));
         }
         Charset contentEncoding = UriTextResource.extractCharacterEncoding(contentType, StandardCharsets.UTF_8);
-        Document document = Jsoup.parse(content, contentEncoding.name(), baseURI.toString());
-        Elements elements = document.select("a[href]");
-        List<String> hrefs = elements.stream()
-            .map(it -> it.attr("href"))
-            .collect(Collectors.toList());
+        List<String> hrefs = anchorScanner.scan(stripByteOrderMark(IOUtils.toString(content, contentEncoding)));
         List<URI> uris = resolveURIs(baseURI, hrefs);
         return filterNonDirectChilds(baseURI, uris);
+    }
+
+    private static String stripByteOrderMark(String html) {
+        return !html.isEmpty() && html.charAt(0) == '\uFEFF' ? html.substring(1) : html;
     }
 
     private URI addTrailingSlashes(URI uri) throws IOException, URISyntaxException {
--- a/platforms/software/resources-http/src/main/java/org/gradle/internal/resource/transport/http/ApacheDirectoryListingParser.java
+++ b/platforms/software/resources-http/src/main/java/org/gradle/internal/resource/transport/http/ApacheDirectoryListingParser.java
@@ -44,13 +44,9 @@ public class ApacheDirectoryListingParser {
             throw new ResourceException(baseURI, String.format("Unsupported ContentType %s for directory listing '%s'", contentType, baseURI));
         }
         Charset contentEncoding = UriTextResource.extractCharacterEncoding(contentType, StandardCharsets.UTF_8);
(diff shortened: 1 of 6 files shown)


platforms/software/resources-http/src/main/java/org/gradle/internal/resource/transport/http/HtmlAnchorHrefScanner.java
```diff
@@ -0,0 +1,306 @@
...
+                valueStart = position;
+                while (position < html.length() && !isUnquotedValueEnd(html.charAt(position))) {
+                    position++;
+                }
+                valueEnd = position;
+            }
+
+            if (anchorHref) {
+                href = resolveCharacterReferences(html.substring(valueStart, valueEnd));
+                hrefSeen = true;
```
jvandort: Is there a difference between hrefSeen == true and href != null?

platforms/software/resources-http/src/main/java/org/gradle/internal/resource/transport/http/ApacheDirectoryListingParser.java
```diff
@@ -33,27 +31,28 @@
...
-        List<String> hrefs = elements.stream()
-            .map(it -> it.attr("href"))
-            .collect(Collectors.toList());
+        List<String> hrefs = anchorScanner.scan(stripByteOrderMark(IOUtils.toString(content, contentEncoding)));
         List<URI> uris = resolveURIs(baseURI, hrefs);
         return filterNonDirectChilds(baseURI, uris);
     }
 
+    private static String stripByteOrderMark(String html) {
+        return !html.isEmpty() && html.charAt(0) == '\uFEFF' ? html.substring(1) : html;
```
jvandort: Why do we need this? Why should or should it not belong in the scanner? Can we add a comment here about where this byte comes from?

jvandort (approved):

platforms/software/resources-http/src/main/java/org/gradle/internal/resource/transport/http/HtmlAnchorHrefScanner.java
big-guy: No — you're right, they're exactly equivalent. `href` starts as `null` and is only ever assigned `""` or the resolved value, never `null`, so `hrefSeen` was just a second name for `href != null`.

I added it when I made a valueless `href` yield `""`, on the theory that the empty string was now a real value that needed distinguishing from "not seen" — but `null` was still doing that job perfectly well. Removed in b547c79.

platforms/software/resources-http/src/main/java/org/gradle/internal/resource/transport/http/ApacheDirectoryListingParser.java
big-guy: Good question, and the answer turned out to be that we don't need it — removed in b547c79.

Where the byte comes from: a server serving the listing as UTF-8 with a byte order mark, so the bytes start `EF BB BF`. Java's UTF-8 decoder doesn't strip that, unlike the UTF-16 one, so it decodes to a leading U+FEFF. I added the strip only because jsoup did the same thing internally and I was preserving parity with it.

On whether it belongs in the scanner: it belongs in neither. The scan looks for the next `<` wherever it happens to be, so a leading U+FEFF is skipped as text like any other character that isn't a tag — the strip was a no-op. Rather than move unnecessary code around I've deleted it and pinned the behaviour with a test on the parser, so the whole chain from raw bytes through the charset decode is covered rather than just the scanner.

platforms/software/resources-http/src/main/java/org/gradle/internal/resource/transport/http/HtmlAnchorHrefScanner.java
big-guy: blah, sorry, I didn't mean to let the agent reply

Issue #36808, opened 2026-02-19
IllegalStateException: Expected exactly one run for grouping node when Develocity Test Distribution reschedules tests after remote executor disconnections


rpalcolea:
### Current Behavior

Build fails with:

```
Execution failed for task ':test'.
> Expected exactly one run for grouping node :com.netflix.MyTest but found: 2
```

The exception is thrown from [`TestTreeModelResultsProvider.createEmptyClassResult()`](https://github.com/gradle/gradle/blob/878f0f5a4c4dacb799ced1d5a9dd6e07f9b23625/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/TestTreeModelResultsProvider.java#L184-L189), which asserts `perRootInfos.size() == 1`.

## What happened

During a build using Develocity Test Distribution, the Develocity server became unavailable (repeated 502/503 responses). This caused multiple remote test executor sessions to disconnect unexpectedly across three waves (22 total session disconnects). The test distribution plugin rescheduled affected tests each time, as expected.

One test class (`MyTest`) was started on 3 different remote agents due to repeated rescheduling. Its last test method was ultimately marked `SKIPPED`. After all rescheduling settled, the internal test results model contained 2 `PerRootInfo` entries for this grouping node, which violates the assertion in `createEmptyClassResult()`.

## Impact

The build fails at test report generation time, even though the actual test executions all passed (or were skipped). The test results are valid, but the report generator cannot handle the duplicate grouping node.

### Expected Behavior

When test distribution rescheduling produces multiple runs for the same test class, the test report generator should handle this gracefully — either by merging the runs or by selecting the most recent/complete run — rather than throwing an `IllegalStateException`.

### Self-contained Reproducer Project

hard to reproduce as it requires Test Distribution

### Gradle version

9.3.1

octylFractal:
Fixed by #36814

PR #38902, opened 2026-08-19
Merge all childless runs when a later run reports children

### LINKED ISSUE #38891: Test report generation: expected exactly one run for grouping node: two childless runs before a run with children survive finalizePath() merging
### Current Behavior

Test report generation fails with:

```
java.lang.IllegalStateException: Expected exactly one run for grouping node :com.example.ContextLoadsTest but found: 2
	at org.gradle.api.internal.tasks.testing.report.generic.TestTreeModelResultsProvider.createEmptyClassResult(TestTreeModelResultsProvider.java:185)
```

This is the same assertion as #36808, which was fixed by #36814 (milestone 9.4.0 RC2). I've
confirmed that fix is present in the versions below (`caa5b8c..v9.5.1` reports
`behind_by: 0`), so this is a distinct, surviving case rather than a regression.

`TestTreeModel.finalizePath()` merges the several runs of one path into a single node:

| arriving entry | behaviour |
| --- | --- |
| list is empty | add the entry |
| **leaf** (no children) | merge into an existing **non-leaf** if one exists, otherwise **append** |
| **non-leaf** | prefer an existing non-leaf, otherwise merge into the **first** entry |

Appending leaves is deliberate, the comment says so:

> Only merge into non-leaf nodes. Leaf nodes might be repeated by test retries, so we'll
> want to add them all to the model if possible.

That is right for a genuine leaf: a retried *test method* should show as two runs. The
problem is that `isLeaf()` cannot distinguish **"this node is a test method"** from
**"this node is a class that happened to report no children"**. A class run that reported
no test method is indistinguishable from a test method, so it gets retry treatment and is
appended.

A leaf therefore only merges when a non-leaf is *already present*, and two childless runs
arriving before any run with children both survive:

1. childless 1 → list empty → appended. **1 entry** (a leaf).
2. childless 2 → leaf, no non-leaf present → appended. **2 entries** (two leaves).
3. run with children → non-leaf, no non-leaf present → merges into the *first* entry only.
   **Still 2 entries.**

`createEmptyClassResult` then asserts exactly one and throws.

#36808 covered `children, then childless`: the childless leaf finds the non-leaf and folds
in. That case is fixed and stays fixed. The gap is two childless runs arriving *first*.

### Expected Behavior

Report generation should not fail. The runs should merge into a single grouping node, the
same way the `children, then childless` sequence already does after #36814: two childless
leaves are no less mergeable than one.

More generally, `createEmptyClassResult` throwing `IllegalStateException` means an internal
invariant is being enforced against a state that `finalizePath()` itself can produce. If
multiple surviving runs are ever legitimate, this should fold them rather than assert; if
they are not, `finalizePath()` should guarantee it.

### Context (optional)

This surfaces in CI on a large multi-project build, where a `test` task fails intermittently
during report generation while the tests themselves pass. Because it fires in report
generation rather than in a test, `ignoreFailures` and `--continue` don't help, and the
error names an internal grouping node, so it reads as a build infrastructure failure with no
actionable signal for the team whose build broke.

### Why this is hard to reproduce in the wild

Reaching this state needs a conjunction of conditions, which is likely why it has gone
unreported despite #36808 being fixed six months ago:

1. **Gradle 9.4+**, for the generic test report path this code does not exist in Gradle 8.
2. **A mechanism that starts a class on a worker that dies before any method reports.**
   In practice that means **Develocity Test Distribution** rescheduling after remote
   executor disconnections, the same trigger as #36808, where one class was started on
   three agents.
3. **The same class must lose its executor twice.** A single disconnect only produces the
   already-fixed `children, one childless` shape. Two consecutive childless runs require
   *correlated* executor loss, not an isolated flaky agent. In the #36808 incident this was
   Develocity 502/503s causing **22 executor disconnects in 3 waves**, a class rescheduled
   off one dying agent landing on another that was also dying.
4. **The class must then complete normally with children.** If no run ever reports a child,
   no grouping node is treated as a class and nothing throws (see `emitTwoChildless` in the
   reproducer).

### Retries alone cannot cause this

Worth stating explicitly, because retries are the intuitive suspect and the `finalizePath`
comment is written in terms of them. I tested this and it does not reproduce: under
`org.gradle.test-retry`, the **first** run always has children, that is what triggered the
retry so it registers as a non-leaf and every later childless run folds into it correctly.
Retries are a *multiplier* on the window (more runs per class, more chances for a disconnect
wave to catch the same class twice), but they cannot produce the leading childless run.

That combination of TD rescheduling under correlated backend failure, on 9.4+, on a build
large enough for it to hit some class  is why an organic reproducer is impractical. The
attached project reproduces the merging defect itself, deterministically, without any of it.

https://github.com/rpalcolea/gradle-childless-retry-repro

```bash
./gradlew verifyEmitTwoChildlessThenChildren   # FAILS: "but found: 2"
./gradlew verifyAll --continue                 # all scenarios, controls included
```

The reproducer emits the event sequence directly through the public
`TestEventReporterFactory` API, so it needs no Test Distribution, no retry plugin, and no
Gradle source checkout, and it is deterministic.

One caveat, stated up front: the HTML report generated by `TestEventReporterFactory` does
**not** go through `TestTreeModelResultsProvider` (it emits no `TEST-*.xml`), so each
scenario has a paired `verify*` task that drives the JUnit XML path explicitly via
`TestTreeModelResultsProvider.useResultsFrom(..)`. The reproducer therefore isolates the
merging defect deterministically rather than reproducing the full distributed scenario; the
production occurrence goes through a plain `Test` task with Test Distribution.

Scenario matrix: exactly one fails, and the controls bound the trigger:

| Task | Run sequence | Entries | Result |
| --- | --- | --- | --- |
| `verifyEmitOneChildless` | childless | 1 | passes |
| `verifyEmitFixedCase` | children, childless | 1 | passes  #36808, correctly fixed |
| `verifyEmitOneChildlessThenChildren` | childless, children | 1 | passes |
| `verifyEmitTwoChildless` | childless, childless | — | passes; no run ever reports a child, so no grouping node is treated as a class |
| **`verifyEmitTwoChildlessThenChildren`** | **childless, childless, children** | **2** | **FAILS** |

`emitFixedCase` and `emitOneChildlessThenChildren` are the same two runs in opposite orders
and both pass, which isolates the trigger to the *count* of leading childless runs rather
than to ordering alone. Two is the minimum.

The repo also keeps three plain `Test`-task scenarios (`@BeforeAll` failure, static
initializer failure, halted worker) that were run and **do not** reproduce, documented in
the README with the reason each one fails to reach the state.

### Gradle version

Reproduced on **9.4.0, 9.5.1, 9.6.0 and 9.7.0**

### Build scan URL (optional)

_No response_

- Gradle 9.5.1 (build time 2026-05-12, revision `fd78213f09782e62ca4957f9cfd3d90c6c3f1767`)
- JVM: Azul Zulu 25.0.2+10-LTS
- OS: Mac OS X 15.7.8 aarch64
- Kotlin 2.3.20, Groovy 4.0.29
- Configuration cache disabled in the reproducer (unrelated variable in report generation)
- Production occurrence: Develocity Test Distribution + `org.gradle.test-retry` 1.6.5

rpalcolea:
Fixes #38891

`TestTreeModel.finalizePath()` merges the several runs of a test group into a single
`PerRootInfo` entry. A run with no children (a leaf) only merges into an existing
non-leaf entry; if none exists yet, it is appended, correct for a genuinely retried
test method, which should show as separate runs.

That rule breaks down when the leaf-shaped runs are not test methods but childless runs
of a *container* (e.g. a class rescheduled onto a worker that was lost before any method
could report). If two or more such runs arrive before the run that finally reports
children, the incoming non-leaf only merges into the *first* existing entry, leaving the
rest stranded as separate `PerRootInfo` entries for the same grouping node.
`TestTreeModelResultsProvider.createEmptyClassResult` then asserts exactly one entry and
throws `IllegalStateException`.

This was a surviving gap in the fix for #36808 (PR #36814): that fix covered
`children, then childless`; this covers `childless, childless, ..., then children`.

## Fix

Fold every existing leaf entry into one before merging in the arriving non-leaf, instead
of merging only into the first. A node that arrives as a non-leaf was never a genuine
test method, so its earlier leaf-shaped entries are childless runs of the same
container, not retries, folding them together is safe and confined to that case.

## Testing

- Added two tests to `TestTreeModelTest` reproducing the failing sequence directly and
  through `TestTreeModelResultsProvider`, both failing before the fix and passing after.
- Verified end-to-end against a standalone reproducer
  ([rpalcolea/gradle-childless-retry-repro](https://github.com/rpalcolea/gradle-childless-retry-repro))
  that drives the event sequence through the public `TestEventReporterFactory` API.

### Context
Any build using Gradle 9.4+'s generic test report, where a test group (typically a test
class) can be started on more than one worker before any of those runs reports a test
method, hits this. In practice that means Develocity Test Distribution rescheduling a
class after a remote executor disconnects, the same trigger as #36808, but this time
the class loses its executor *twice* before it finally runs to completion with children.

Because the failure happens during report generation rather than during the tests
themselves, `--continue` and `ignoreFailures` don't help, and the thrown
`IllegalStateException` names an internal grouping node with no actionable signal, so it
reads as a build infrastructure failure rather than a flaky test. This is especially
disruptive on large, multi-project builds under heavy Test Distribution use, where
correlated executor loss (e.g. a wave of backend disconnects) is far more likely to hit
the same class twice.

#36808 / #36814 fixed the `children, then childless` ordering six months ago, but the
`childless, childless, ..., then children` ordering was never covered and still throws
today on every release since, including the latest (9.4.0 through 9.7.0 confirmed). A
deterministic, minimal reproducer isolating the defect (no Test Distribution or retry
plugin required) is at
[rpalcolea/gradle-childless-retry-repro](https://github.com/rpalcolea/gradle-childless-retry-repro).

*This PR, including the diagnosis, fix, tests, and this description, was developed with
AI assistance (Claude Code); the reproducer, fix approach, and verification were
reviewed and validated by me before submission.*

### Contributor Checklist

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/TestTreeModel.java
+++ b/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/generic/TestTreeModel.java
@@ -197,12 +197,25 @@ public class TestTreeModel {
                         existingRootInfos.add(rootInfo);
                     }
                 } else {
-                    // Merge into the one that is also not a leaf if possible, otherwise just merge into the first one.
-                    PerRootInfo.Builder toMerge = existingRootInfos.stream()
+                    // Merge into the one that is also not a leaf if possible.
+                    PerRootInfo.Builder existingNonLeaf = existingRootInfos.stream()
                         .filter(info -> !info.isLeaf())
                         .findFirst()
-                        .orElseGet(() -> existingRootInfos.get(0));
-                    toMerge.merge(rootInfo);
+                        .orElse(null);
+                    if (existingNonLeaf != null) {
+                        existingNonLeaf.merge(rootInfo);
+                    } else {
+                        // Every existing entry is a leaf, but this one has children, so none of them were ever
+                        // a genuine test method retry -- they're earlier runs of this same node that happened to
+                        // report no children (e.g. a class rescheduled onto a worker that was lost before any
+                        // test method could report). Fold them all into the first entry along with this one,
+                        // rather than stranding the rest, which would leave more than one entry for this node.
+                        PerRootInfo.Builder first = existingRootInfos.get(0);
+                        for (int i = existingRootInfos.size() - 1; i > 0; i--) {
+                            first.merge(existingRootInfos.remove(i));
+                        }
+                        first.merge(rootInfo);
+                    }
                 }
             } else {
                 existingRootInfos.add(rootInfo);
--- a/platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/report/generic/TestTreeModelTest.groovy
+++ b/platforms/software/testing-base/src/test/groovy/org/gradle/api/internal/tasks/testing/report/generic/TestTreeModelTest.groovy
@@ -32,10 +32,12 @@ import java.nio.file.Path
  * Tests for {@link TestTreeModel}, specifically the merging behavior in finalizePath().
  *
  * <p>When a test group (e.g. a test class) is emitted multiple times within the same root,
- * some executions may have no children (e.g. when all methods are skipped). These childless
- * executions appear as leaves and must be merged into any existing non-leaf entry for the same
- * node, rather than being added as separate entries. True leaf retries (where all entries are
- * leaves) should still be preserved as separate entries.</p>
+ * some executions may have no children (e.g. when all methods are skipped, or a worker is
+ * lost before any method reports). These childless executions appear as leaves and must be
+ * merged into a non-leaf entry for the same node, regardless of whether that non-leaf entry
+ * already exists or arrives later -- otherwise every childless execution before the first
+ * execution with children would survive as its own entry. True leaf retries (where all
+ * entries are leaves) should still be preserved as separate entries.</p>
  */
 class TestTreeModelTest extends Specification {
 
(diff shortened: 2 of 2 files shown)


ljacomet:

big-guy:
Thanks, we need to look at this and add an integ test.

PR #38904, opened 2026-08-20
Kill worker daemons still running abandoned work after a task timeout

blindpirate:
Fixes https://github.com/gradle/gradle-private/issues/5086

### Context

`Task.timeout` does not stop work submitted with `processIsolation()`. It fails the task, and then the build blocks for the full remaining duration of the abandoned work item.

When a task times out while a work item is running in a worker daemon:

1. The thread waiting on the work item is interrupted. `AbstractConditionalExecution.await()` responds by cancelling the execution, which interrupts the thread inside `WorkerDaemonClient.execute()`. The client is therefore released back to the **idle pool**, even though the worker process is still running the work item and has never reported a result.
2. Session teardown then asks that worker to stop gracefully. `RequestProtocol.stop()` and `run()` arrive on the same incoming queue in the worker (`WorkerAction`), so the stop request simply queues behind the in-flight work item. The build then blocks in `DefaultExecHandle.waitForFinish()`, which has **no timeout**, until the abandoned work item finishes on its own.

The only path that actually kills a worker, `WorkerDaemonClientsManager.killAllWorkers()`, is wired solely to `WorkerDaemonClientCancellationHandler` (build cancellation). Nothing on the timeout path.

Standalone demonstration on 9.7, no Gradle checkout involved — a task with `timeout = Duration.ofMillis(500)` submitting process isolation work items that block for 30s:

```
> Task :block FAILED
Requesting stop of task ':block' as it has exceeded its configured timeout of 500ms.
BUILD FAILED in 2s          <- what Gradle reports
WORK ACTION FINISHED ...    <- every work item still runs to completion
real 33.1s                  <- actual wall clock
```

Wall clock tracks the abandoned work 1:1: 30s work → ~34s build, 10s work → ~13s build, work never started → ~4s.

There is a second, independent consequence of (1): an abandoned client goes straight back into the idle pool, so the next task to reserve a worker can be handed one whose process is still executing the previous work item, and queues behind it. No timeout required to hit that.

### Why the test is flaky rather than always red

`TaskTimeoutIntegrationTest.timeout stops long running work items with process isolation` has been intermittently red for years and is currently the top failure in the Windows flaky-test quarantine builds ([116452314](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_FlakyQuarantine_Check_Quick_2/116452314), [116426068](https://builds.gradle.org/buildConfiguration/Gradle_Master_Check_FlakyQuarantine_Check_Quick_2/116426068)).

It hangs only when a worker daemon wins a race against the 500ms task timeout. When the daemon has not connected yet, teardown aborts it mid-startup and the test passes in ~3s. When it has reached the blocking section, the build waits out the 90s work item and the 60s `@IntegrationTestTimeout` kills it. In one quarantine build the five reruns took 60.2s, 60.3s, **3.2s**, 60.6s, 60.6s — the fast one being the run where no work item ever started.

### Changes

`WorkerDaemonClient` records whether a submitted work item was abandoned before the worker reported a result. Such a client:

- is no longer returned to the idle pool, since its worker is still busy with the old request;
- is killed rather than stopped gracefully when workers are stopped.

Well-behaved workers keep the existing graceful stop.

### Reproducer

The added integration test removes the race, so it fails deterministically rather than intermittently. A `:warmup` task forks a worker daemon and returns it to the idle pool, so `:block` reuses a warm daemon and its work item starts immediately; the task action then waits on a marker file so the timeout provably fires while the worker is inside its blocking section. The test asserts the marker exists, so a lost race fails loudly instead of passing silently.

Before the fix, 4/4 runs:

```
Build took 61912ms, i.e. it waited for the abandoned 60000ms work item
instead of stopping the worker daemon executing it.
```

### Verification

Local (macOS):

| | before | after |
|---|---|---|
| `build does not wait for abandoned process isolation work after a timeout` | 60.8s FAIL (4/4) | **5.65s pass** |
| `timeout stops long running work items with process isolation` | 60.0s FAIL | **1.58s pass** |
| `TaskTimeoutIntegrationTest` | 2 failed | **13/13 pass** |
| `:workers:embeddedIntegTest` + `:workers:test` | | **426 tests, 0 failures** |

Windows (`Rerun Flaky Test - Windows Amd64`, Java 17, 10 iterations per build), which is where this has been failing:

| build | commit | test | result |
|---|---|---|---|
| [116525063](https://builds.gradle.org/build/116525063) | repro **without** the fix | `build does not wait for abandoned process isolation work after a timeout` | **0/30 pass**, 67.4–69.4s each |
| [116505410](https://builds.gradle.org/build/116505410) | with the fix | same test | **10/10 pass**, ~12.7s each |
| [116524572](https://builds.gradle.org/build/116524572) | with the fix | `timeout stops long running work items with #isolationMode isolation` (the long-standing flaky test) | **30/30 pass** — `no` 10/10 (6.2–7.3s), `classLoader` 10/10 (2.4–2.7s), **`process` 10/10 (5.9–7.8s)** |

- [x] `./gradlew sanityCheck` — green on CI (`Sanity Check on Linux`, `Sanity Check (Quick Feedback - Linux Only)`)

### Note for reviewers

This is a deliberate behaviour change, not only a bug fix: a work item that was still running when its task timed out is now **terminated mid-flight** rather than left to run to completion. That seems right, since the build has already discarded its result and the task is failing regardless — but it is worth an explicit opinion.

Not included, deliberately: a bounded `waitForStop` that escalates to a kill on expiry. That is worth having as defence in depth for a worker wedged for some reason this change does not classify as abandoned, but as the primary fix it would be a timeout papering over the real gap, and it would have muddied the signal from the reproducer. Happy to add it here or separately.

### Contributor Checklist
- [x] Existing unit tests (`WorkerDaemonClientsManagerTest`) pass; no new unit test, the behaviour is only observable end to end.
- [ ] Update User Guide, DSL Reference, and Javadoc for public-facing changes. — n/a, internal types only
- [x] Ensure that tests pass sanity check: `./gradlew sanityCheck`. — green on CI
- [x] Ensure that tests pass locally: `./gradlew :workers:quickTest` equivalent, see table above.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonClient.java
+++ b/platforms/core-execution/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonClient.java
@@ -26,7 +26,6 @@ import org.gradle.process.internal.worker.MultiRequestClient;
 import org.gradle.process.internal.worker.WorkerProcess;
 
 import java.util.Optional;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 class WorkerDaemonClient implements Stoppable, Describable {
     public static final String DISABLE_EXPIRATION_PROPERTY_KEY = "org.gradle.workers.internal.disable-daemons-expiration";
@@ -36,8 +35,8 @@ class WorkerDaemonClient implements Stoppable, Describable {
     private final LogLevel logLevel;
     private final ActionExecutionSpecFactory actionExecutionSpecFactory;
     private int uses;
-    private final AtomicBoolean executing = new AtomicBoolean();
-    private final AtomicBoolean abandoned = new AtomicBoolean();
+    private volatile boolean executing;
+    private volatile boolean abandoned;
     private boolean cannotBeExpired = Boolean.getBoolean(DISABLE_EXPIRATION_PROPERTY_KEY);
 
     public WorkerDaemonClient(DaemonForkOptions forkOptions, MultiRequestClient<TransportableActionExecutionSpec, DefaultWorkResult> workerClient, WorkerProcess workerProcess, LogLevel logLevel, ActionExecutionSpecFactory actionExecutionSpecFactory) {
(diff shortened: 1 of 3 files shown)


octylFractal (approved):
I cleaned up some of the logic and Javadoc, and fixed a bug. LGTM now.

blindpirate:

PR #38913, opened 2026-08-20
Stop capturing the release notes date formatters

### LINKED ISSUE #38911: `:docs:releaseNotes` fails to store configuration cache state: captured DateTimeFormatter holds a Locale
### Current Behavior

Building Gradle from source fails while the configuration cache stores its entry. The task is `:docs:releaseNotes`:

```text
* What went wrong:
Configuration cache state could not be cached: field `locale` of
`java.time.format.DateTimeFormatter` bean found in captured state from
method `gradlebuild.docs.GradleReleaseNotesPlugin.generateReleaseNotes`
of lambda of type `org.gradle.api.Transformer` returning `java.lang.String`
found in field `transformer` of `org.gradle.api.internal.provider.TransformBackedProvider`
bean found in field `provider` of `org.gradle.api.internal.provider.TransformBackedProvider`
bean found in field `providerOfEntries` of `org.gradle.api.internal.provider.MapCollectors$EntriesFromMapProvider`
bean found in field `buffer` of `org.gradle.api.internal.provider.AppendOnceList`
bean found in field `collectors` of `org.gradle.api.internal.provider.DefaultMapProperty$CollectingSupplier`
bean found in field `__replacementTokens__` of task `:docs:releaseNotes`
of type `gradlebuild.docs.DecorateReleaseNotes`: error writing value of type 'java.util.Locale'
> 'ObjectOutputStream.putFields' is not supported by the Gradle configuration cache.

BUILD FAILED
Configuration cache entry discarded due to serialization error.
```

The configuration cache is enabled by default in this repository, so nothing has to be turned on for this to happen.

`--no-configuration-cache` alone does not work around it, because Isolated Projects is enabled as well:

```text
* What went wrong:
Configuration Cache cannot be disabled when Isolated Projects is enabled.
```

Both have to be turned off together:

```console
-Dorg.gradle.unsafe.isolated-projects=false --no-configuration-cache
```

### When it reproduces

Only when the version is requested without a build timestamp in it. Plain invocations pass, which makes the report easy to dismiss:

| Invocation | Result |
| --- | --- |
| `./gradlew :docs:releaseNotes` | passes |
| `./gradlew :docs:releaseNotes -PfinalRelease=true` | fails |
| `./gradlew :docs:releaseNotes -PrcNumber=1` | fails |
| `./gradlew :docs:releaseNotes -PbuildVersionQualifier=x` | passes |

Reproduced on `v9.7.1` and on `master` (Gradle 9.8.0-milestone-1). A full build is not required, since the entry is rejected while it is stored.

### Expected Behavior

`:docs:releaseNotes` completes with the configuration cache left at its default setting.

### Context (optional)

I maintain the `gradle` package in ALT Linux Sisyphus, the rolling development branch the distribution builds its releases from: https://packages.altlinux.org/en/sisyphus/srpms/gradle

The package is built from the upstream sources rather than repackaged from a binary distribution, inside an rpm buildroot with no network access and a vendored dependency cache. I am updating it from 8.14.3 to 9.7.1, and the build passes `-PfinalRelease=true`, so it hits this every time.

On 8.14.3 the spec passed `--no-configuration-cache` and that was enough. It now needs both flags above, and I would rather not ship a distribution package that turns off two upstream defaults.

This is not specific to my package. Linux distributions ship final releases only, never snapshots, release candidates or milestones. A plain invocation of this build produces a timestamped snapshot, since `snapshot` stays `true` unless `finalRelease`, `rcNumber` or `milestoneNumber` is set, so `-PfinalRelease=true` is the only way to get a clean `9.7.1`. That flag is also one of the two that trigger the failure, so every attempt to package Gradle for a distribution runs into this, while the plain command a triager would try first looks healthy.

The problem is in the build logic of this repository, so a separate reproducer project is not possible. Steps to reproduce:

```console
git clone https://github.com/gradle/gradle.git
cd gradle
./gradlew :docs:releaseNotes -PfinalRelease=true
```

This fails in a few seconds. The same happens on the `v9.7.1` tag.

### Gradle version

`master` (wrapper resolves 9.8.0-milestone-1) and the `v9.7.1` source tree (wrapper resolves 9.7.0).

### Build scan URL (optional)

_No response_

```text
OS:      ALT Regular Sisyphus (VERSION_ID 20260316), x86_64
Kernel:  6.18.35-6.18-alt1
CPU:     AMD Ryzen 5 7535U, 12 threads
RAM:     14 GiB
JDKs:    java-11-openjdk 11.0.31.1, java-17-openjdk 17.0.19,
         java-21-openjdk 21.0.11 (distribution packages)
         Temurin 25.0.3+9 LTS, provisioned by toolchain auto-provisioning
```

Also reproduced on macOS 15 (arm64) with Oracle JDK 25.0.2.

IvanKhanas:
Fixes #38911

### Context

Building Gradle from source fails at `:docs:releaseNotes` whenever the version carries no build timestamp. `-PfinalRelease`, `-PrcNumber` and `-PmilestoneNumber` all produce such a version, so a release build cannot store its configuration cache entry.

This matters well beyond the few people who run the build. Distributions ship final releases only, never snapshots, release candidates or milestones, and here the version stays a snapshot unless one of those three properties is set. `-PfinalRelease=true` is therefore not optional for a packager, which puts every distribution build on the failing path. The plain command a reviewer would try first passes, which is what kept this quiet.

A packager is a chokepoint rather than an audience. Everyone who installs Gradle from a distribution repository gets whatever came out of that build, and so does everyone who consumes those machines downstream. Working around the failure means turning the configuration cache and Isolated Projects off, so the binary those users receive is produced in a configuration you do not test.

I hit it while updating the `gradle` package in ALT Linux Sisyphus from 8.14.3 to 9.7.1: https://packages.altlinux.org/en/sisyphus/srpms/gradle

The package builds from the upstream sources in an rpm buildroot with no network access. Until now the spec passed `--no-configuration-cache`, which no longer works on its own because Isolated Projects is enabled as well, so it needs `-Dorg.gradle.unsafe.isolated-projects=false --no-configuration-cache`.

The same failure reproduces on `master` and on `v9.7.1`, and a full build is not needed for it:

```console
./gradlew :docs:releaseNotes -PfinalRelease=true
```

### Contributor Checklist

Notes on the checklist, since `build-logic` does not work quite like a product subproject:

**Integration tests.** `build-logic`, `build-logic-commons` and `build-logic-settings` have no `src/integTest` directory. The test went to `src/test` next to `FindBrokenInternalLinksTest`, which drives a real build through TestKit in the same way. It is parameterised over the four version shapes and fails on three of them without the production change.

**Documentation.** Nothing here is public facing, so there was nothing to update.

**Sanity check.** Run, passes. It still does not cover this change: none of its eleven tasks reaches `build-logic`.

**quickTest.** `build-logic` has no `quickTest` task. The equivalent for an included build is `gradle.includedBuilds.map { it.task(":check") }`, which belongs to `checkBuildLogic` rather than to `sanityCheck`. I ran it as `./gradlew :build-logic:check` and it passes.

### DIFF
--- a/build-logic/documentation/src/main/groovy/gradlebuild/docs/GradleReleaseNotesPlugin.java
+++ b/build-logic/documentation/src/main/groovy/gradlebuild/docs/GradleReleaseNotesPlugin.java
@@ -42,6 +42,9 @@ import java.time.format.DateTimeFormatter;
  * TODO: Maybe eventually convert this asciidoc too, so everything uses the same markup language.
  */
 public class GradleReleaseNotesPlugin implements Plugin<Project> {
+    private static final DateTimeFormatter BUILD_TIMESTAMP_FORMAT = DateTimeFormatter.ofPattern("yyyyMMddHHmmssZ");
+    private static final DateTimeFormatter RELEASE_DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+
     @Override
     public void apply(Project project) {
         ProjectLayout layout = project.getLayout();
@@ -77,9 +80,8 @@ public class GradleReleaseNotesPlugin implements Plugin<Project> {
 
             MapProperty<String, String> replacementTokens = task.getReplacementTokens();
             Provider<String> buildTimestamp = moduleIdentity.getBuildTimestamp();
-            DateTimeFormatter inputFormatter = DateTimeFormatter.ofPattern("yyyyMMddHHmmssZ");
-            DateTimeFormatter outputFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
-            Provider<String> dateTime = buildTimestamp.map(timestamp -> ZonedDateTime.parse(timestamp, inputFormatter).format(outputFormatter));
+            Provider<String> dateTime = buildTimestamp.map(timestamp ->
+                ZonedDateTime.parse(timestamp, BUILD_TIMESTAMP_FORMAT).format(RELEASE_DATE_FORMAT));
 
             replacementTokens.put("releaseDate", dateTime);
             replacementTokens.put("version", moduleIdentity.getVersion().map(GradleVersion::getVersion));
--- a/build-logic/documentation/src/test/groovy/gradlebuild/docs/GradleReleaseNotesPluginTest.groovy
+++ b/build-logic/documentation/src/test/groovy/gradlebuild/docs/GradleReleaseNotesPluginTest.groovy
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2026 Gradle and contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package gradlebuild.docs
+
+import org.gradle.testkit.runner.GradleRunner
+import spock.lang.Specification
+import spock.lang.TempDir
+
+import java.time.LocalDate
+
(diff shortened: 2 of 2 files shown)


ov7a:

lkasso (approved):
Thank you for your contribution. I will go ahead and merge it.

IvanKhanas:
> Thank you for your contribution. I will go ahead and merge it.

Thanks for the review 🤩

PR #38917, opened 2026-08-21
Prevent signing from realizing component variants

### LINKED ISSUE #38909: `signing.sign(publication)` resolves version-mapped classpaths at configuration time in 9.8.0-M1, breaking later `dependencies {}` blocks
# Current Behavior

Since Gradle 9.8.0-M1, calling `signing.sign(mavenPublication)` on a publication that uses `versionMapping { fromResolutionResult() }` eagerly resolves the mapped configurations (e.g. `compileClasspath`) **while `sign()` is being called**, i.e. during project configuration.

Resolving the project's own `compileClasspath` builds the local component and marks the project's consumable configurations (`apiElements`, `runtimeElements`, …) as observed. Any dependency declared afterwards in the same project fails with:

```
Cannot mutate the dependencies of configuration ':api' after the configuration's child configuration ':apiElements' was published as a variant. After a configuration has been observed, it should not be modified.
```

This is particularly painful when signing is configured by a convention plugin applied in `plugins {}`: the plugin runs before the script body, so every `dependencies {}` block in the build script fails. This broke publishing in the Gradle build itself (`:kotlin-dsl-plugins:publishPlugins` and the nightly promotion build) as soon as the wrapper was bumped to 9.8.0-M1. Locally it goes unnoticed because signing is only enabled when the signing key env vars are present (CI).

Even when no later mutation happens, resolving classpaths at configuration time for every signed publication is a performance regression and is incompatible with the lazy-configuration model.

# Expected Behavior

`signing.sign(publication)` should not resolve any configuration. It worked lazily in 9.7.0 (same reproducer passes).

# Context

Root cause, as far as I can tell from stack traces (full trace below):

1. `Sign.sign(Publication)` → `publicationInternal.allPublishableArtifacts(artifact -> isNoSignatureArtifact(artifact))` → `artifact.getFile()` — this has always been eager.
2. For the POM artifact, `SingleOutputTaskMavenArtifact.getFile()` does `task.getOutputs().getFiles().getSingleFile()` on `generatePomFileFor<Pub>Publication`, which visits *all* task properties via `TaskPropertyUtils.visitProperties`.
3. New in 9.8.0-M1: commit `23d63d19cdf` "Declare input annotations for MavenPom" made `GenerateMavenPom#getPom()` / `MavenPomInternal#getDependencies()` (and other members) `@Nested`. The property walker now unpacks the nested `Provider<MavenPomDependencies>`.
4. That provider is `DefaultMavenPublication` → `DefaultDependencyCoordinateResolverFactory`, which is built from `configuration.getIncoming().getResolutionResult().getRootComponent().zip(getRootVariant(), …)` for the version-mapped configurations → resolution happens right there.

# Steps to Reproduce

Self-contained, single-project Kotlin DSL build:

`settings.gradle.kts`
```kotlin
rootProject.name = "signing-repro"
```

`build.gradle.kts`
```kotlin
plugins {
    `java-library`
    `maven-publish`
    signing
}

group = "org.example"
version = "1.0"

publishing {
    publications {
        create<MavenPublication>("maven") {
            from(components["java"])
            versionMapping {
                usage("java-api") { fromResolutionResult() }
            }
        }
    }
}

signing {
    useInMemoryPgpKeys("not-a-real-key", "not-a-real-passphrase")
    sign(publishing.publications["maven"])
}

// Fails on 9.8.0-M1: compileClasspath was already resolved by signing.sign() above
dependencies {
    api("com.google.guava:guava:33.0.0-jre")
}
```

Run `gradle help`.

- Gradle 9.7.0: `BUILD SUCCESSFUL`
- Gradle 9.8.0-M1:
  ```
  * Where:
  Build file '…/build.gradle.kts' line: 28

  * What went wrong:
  Cannot mutate the dependencies of configuration ':api' after the configuration's child configuration ':apiElements' was published as a variant. After a configuration has been observed, it should not be modified.
  ```
- Gradle 9.8.0-M1 with the `versionMapping {}` block removed: `BUILD SUCCESSFUL` (so the resolution-backed POM dependencies provider is the trigger).

Stack trace of the configuration-time resolution (captured with an init script adding `configurations.configureEach { incoming.beforeResolve { Thread.dumpStack() } }`, trimmed):

```
CONFIG-TIME RESOLVE ::compileClasspath
	at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.resolveGraphIfRequired(DefaultConfiguration.java:654)
	at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration$ResolverResultsResolutionResultProvider.getValue(DefaultConfiguration.java:636)
	at org.gradle.api.internal.artifacts.result.DefaultResolutionResult.lambda$getRootComponent$0(DefaultResolutionResult.java:67)
	... (MergeProvider / FlatMapProvider / TransformBackedProvider chain from DefaultDependencyCoordinateResolverFactory)
	at org.gradle.api.internal.provider.DefaultProperty.calculateValueFrom(DefaultProperty.java:158)
	at org.gradle.api.internal.provider.AbstractMinimalProvider.getOrNull(AbstractMinimalProvider.java:105)
	at org.gradle.internal.properties.annotations.AbstractTypeMetadataWalker$InstanceTypeMetadataWalker.lambda$walkNestedProvider$0(AbstractTypeMetadataWalker.java:139)
	at org.gradle.internal.properties.annotations.AbstractTypeMetadataWalker$InstanceTypeMetadataWalker.walkNestedProvider(AbstractTypeMetadataWalker.java:138)
	... (walking @Nested GenerateMavenPom#getPom() -> @Nested MavenPomInternal#getDependencies())
	at org.gradle.api.internal.tasks.TaskPropertyUtils.visitAnnotatedProperties(TaskPropertyUtils.java:58)
	at org.gradle.api.internal.tasks.TaskPropertyUtils.visitProperties(TaskPropertyUtils.java:35)
	at org.gradle.api.internal.tasks.DefaultTaskOutputs.getFileProperties(DefaultTaskOutputs.java:157)
	at org.gradle.api.internal.tasks.DefaultTaskOutputs$TaskOutputUnionFileCollection.visitChildren(DefaultTaskOutputs.java:249)
	at org.gradle.api.internal.file.AbstractFileCollection.getSingleFile(AbstractFileCollection.java:151)
	at org.gradle.api.publish.maven.internal.artifact.SingleOutputTaskMavenArtifact.getFile(SingleOutputTaskMavenArtifact.java:42)
	at org.gradle.plugins.signing.Sign.isNoSignatureArtifact(Sign.java:221)
	at org.gradle.plugins.signing.Sign.lambda$sign$2(Sign.java:212)
	at org.gradle.api.publish.maven.internal.publication.DefaultMavenPublication.allPublishableArtifacts(DefaultMavenPublication.java:358)
	at org.gradle.plugins.signing.Sign.sign(Sign.java:211)
	at org.gradle.plugins.signing.SigningExtension.lambda$createSignTaskFor$0(SigningExtension.java:503)
	at org.gradle.plugins.signing.SigningExtension.createSignTaskFor(SigningExtension.java:501)
	at org.gradle.plugins.signing.SigningExtension.sign(SigningExtension.java:466)
	at Build_gradle$2.execute(build.gradle.kts:23)
	at Build_gradle.<init>(build.gradle.kts:21)
```

Workaround: defer the call, e.g. `afterEvaluate { signing.sign(publication) }`. This avoids the mutation error but still resolves the classpaths at configuration time.

# Gradle version

Regression in 9.8.0-M1 (`gradle-9.8.0-milestone-1`). Works in 9.7.0. Introduced by `23d63d19cdf` ("Declare input annotations for MavenPom").

# Build scan URL

Gradle build itself, failing in `:build-cache` (first published module to be configured), both on 9.8.0-M1:

- https://ge.gradle.org/s/kvstw4vm5h4du — `clean :kotlin-dsl-plugins:publishPlugins` (Util_PublishKotlinDslPlugin)
- https://ge.gradle.org/s/bg4epk2shqvds — `clean promotionBuild` (Gradle_Master_Promotion_Nightly, master @ `697afbe54bf`)

jvandort:
#38424 introduced input annotations to GenerateMavenPom. This caused calls to generateMavenPom.getOutputs().getFiles() to realize the properties of GenerateMavenPom, including the @Nested pom and its nested dependencies provider. Realizing that provider parses the dependencies of the publication, which finalizes the variants of the published component and resolves any configurations providing version-mapped dependency versions.

Signing eagerly realizes publication artifact files, which for Maven POMs, GMM files, and Ivy descriptor files, were provided by getOutputs(). As a result, after #38424, signing a publication caused failures during configuration time when any mutations to the set of variants or set of dependencies were made after sign(Publication).

We resolve this by directly wiring the known POM, GMM, and Ivy descriptor output files to the publications' metadata artifacts rather than relying on getOutputs(). Note that explicit calls to generateMavenPom.getOutputs() will continue to cause this same behavior change, but workarounds exist similar to ours: directly wire the task's output file properties without eagerly realizing its output properties. The benefits to Maven publication and configuration cache compatibility resulting from #38424 are likely worth this trade-off.

Note that #38424 does not cause a regression with Ivy, as GenerateIvyDescriptor does not yet declare input annotations for its descriptor, but we apply the same fix as we do for Maven for consistency and for forward compatibility with a similar change to Ivy when it is implemented.

Fixes #38909

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things

### DIFF
--- a/platforms/software/maven/src/main/java/org/gradle/api/publish/maven/internal/artifact/SingleOutputTaskMavenArtifact.java
+++ b/platforms/software/maven/src/main/java/org/gradle/api/publish/maven/internal/artifact/SingleOutputTaskMavenArtifact.java
@@ -17,7 +17,9 @@
 package org.gradle.api.publish.maven.internal.artifact;
 
 import org.gradle.api.Task;
+import org.gradle.api.file.RegularFile;
 import org.gradle.api.internal.tasks.TaskDependencyInternal;
+import org.gradle.api.provider.Provider;
 import org.gradle.api.tasks.TaskProvider;
 import org.gradle.api.internal.tasks.TaskDependencyFactory;
 
@@ -25,13 +27,15 @@ import java.io.File;
 
 public class SingleOutputTaskMavenArtifact extends AbstractMavenArtifact {
     private final TaskProvider<? extends Task> generator;
+    private final Provider<RegularFile> file;
     private final String extension;
     private final String classifier;
     private final TaskDependencyInternal buildDependencies;
 
-    public SingleOutputTaskMavenArtifact(TaskProvider<? extends Task> generator, String extension, String classifier, TaskDependencyFactory taskDependencyFactory) {
+    public SingleOutputTaskMavenArtifact(TaskProvider<? extends Task> generator, Provider<RegularFile> file, String extension, String classifier, TaskDependencyFactory taskDependencyFactory) {
         super(taskDependencyFactory);
         this.generator = generator;
+        this.file = file;
         this.extension = extension;
         this.classifier = classifier;
         this.buildDependencies = taskDependencyFactory.visitingDependencies(context -> context.add(getGenerator()));
@@ -39,7 +43,7 @@ public class SingleOutputTaskMavenArtifact extends AbstractMavenArtifact {
 
     @Override
     public File getFile() {
-        return getGenerator().getOutputs().getFiles().getSingleFile();
+        return file.get().getAsFile();
     }
 
     private Task getGenerator() {
--- a/platforms/software/maven/src/main/java/org/gradle/api/publish/maven/internal/artifact/SingleOutputTaskMavenArtifact.java
+++ b/platforms/software/maven/src/main/java/org/gradle/api/publish/maven/internal/artifact/SingleOutputTaskMavenArtifact.java
@@ -16,38 +16,38 @@
 
 package org.gradle.api.publish.maven.internal.artifact;
 
-import org.gradle.api.Task;
 import org.gradle.api.file.RegularFile;
+import org.gradle.api.internal.tasks.TaskDependencyFactory;
 import org.gradle.api.internal.tasks.TaskDependencyInternal;
 import org.gradle.api.provider.Provider;
-import org.gradle.api.tasks.TaskProvider;
-import org.gradle.api.internal.tasks.TaskDependencyFactory;
 
 import java.io.File;
 
 public class SingleOutputTaskMavenArtifact extends AbstractMavenArtifact {
-    private final TaskProvider<? extends Task> generator;
     private final Provider<RegularFile> file;
+    private final Provider<Boolean> enabled;
     private final String extension;
     private final String classifier;
     private final TaskDependencyInternal buildDependencies;
 
-    public SingleOutputTaskMavenArtifact(TaskProvider<? extends Task> generator, Provider<RegularFile> file, String extension, String classifier, TaskDependencyFactory taskDependencyFactory) {
+    public SingleOutputTaskMavenArtifact(
+        Provider<RegularFile> file,
+        Provider<Boolean> enabled,
+        String extension,
+        String classifier,
+        TaskDependencyFactory taskDependencyFactory
+    ) {
         super(taskDependencyFactory);
-        this.generator = generator;
         this.file = file;
+        this.enabled = enabled;
         this.extension = extension;
         this.classifier = classifier;
-        this.buildDependencies = taskDependencyFactory.visitingDependencies(context -> context.add(getGenerator()));
+        this.buildDependencies = taskDependencyFactory.visitingDependencies(context -> context.add(file));
     }
 
     @Override
     public File getFile() {
         return file.get().getAsFile();
-    }
-
-    private Task getGenerator() {
-        return generator.get();
     }
 
     @Override
@@ -66,7 +66,7 @@ public class SingleOutputTaskMavenArtifact extends AbstractMavenArtifact {
     }
 
     public boolean isEnabled() {
-        return getGenerator().getEnabled();
+        return enabled.get();
     }
 
     @Override
--- a/platforms/software/ivy/src/main/java/org/gradle/api/publish/ivy/internal/artifact/SingleOutputTaskIvyArtifact.java
+++ b/platforms/software/ivy/src/main/java/org/gradle/api/publish/ivy/internal/artifact/SingleOutputTaskIvyArtifact.java
@@ -16,39 +16,44 @@
 
 package org.gradle.api.publish.ivy.internal.artifact;
 
-import org.gradle.api.Task;
 import org.gradle.api.file.RegularFile;
-import org.gradle.api.internal.TaskInternal;
 import org.gradle.api.internal.tasks.TaskDependencyFactory;
 import org.gradle.api.internal.tasks.TaskDependencyInternal;
 import org.gradle.api.provider.Provider;
 import org.gradle.api.publish.ivy.internal.publisher.IvyPublicationCoordinates;
 import org.gradle.api.tasks.TaskDependency;
-import org.gradle.api.tasks.TaskProvider;
 import org.jspecify.annotations.Nullable;
 
 import java.io.File;
 
 public class SingleOutputTaskIvyArtifact extends AbstractIvyArtifact {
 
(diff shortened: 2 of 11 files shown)


jvandort:
@bot-gradle test rfr

jvandort:
@bot-gradle rfr

jvandort:
@bot-gradle test rfr

platforms/software/maven/src/main/java/org/gradle/api/publish/maven/internal/artifact/SingleOutputTaskMavenArtifact.java
```diff
@@ -62,7 +66,7 @@ protected TaskDependencyInternal getDefaultBuildDependencies() {
     }
 
     public boolean isEnabled() {
-        return getGenerator().getEnabled();
+        return enabled.get();
```
hfhbd, line 69: What's the point of the `enabled` provider? `getFile` will execute the providing task due to calling `get()`.
Why not using `file.getAsFile().getOrNull()`? If the task is not enabled, it should return null, shouldn't it?

 Or is it needed only until Provider Migration that migrates to `public Provider<RegularFile> getFile()` and `isEnabled()` can be removed?

octylFractal: I think you misunderstand how Providers work? A `Provider.get()` does not execute any tasks, the tasks are only executed if they are a dependency of another task or selected as an entry point. Typically inputs and outputs of a task are wired regardless of if they're enabled or not, so it's not really expected that a disabled task returns `null` from its properties.

octylFractal (approved):

platforms/software/maven/src/main/java/org/gradle/api/publish/maven/internal/artifact/SingleOutputTaskMavenArtifact.java
hfhbd: > Typically inputs and outputs of a task are wired regardless of if they're enabled or not, so it's not really expected that a disabled task returns null from its properties.

That's true, but getting the file provider `file.get().getAsFile()` in `getFile(): File` will execute the task, won't it?

For example the `GenerateMavenPom` task: `file` is the output file provider of the `GenerateMavenPom` task, so calling `get` will execute the task, unless the task is not enabled: https://github.com/gradle/gradle/pull/38917/changes/dc007096840419b3375e09aa7b402f334388ba32#diff-89219ce596f6130a5aa444f2cd62f16a588bfdb1ebd13afa72714f7d9f1986cbR199

So you need to call `enabled.get()` first to check if the task will produce the file specified in the file provider. Otherwise, after calling `getFile()`, you get a path to a file that does not exist (because the task did not run). And that additional check (and so the `isEnabled` provider) makes me curios if using lazy providers everywhere would solve this by only evaluating 1 provider.

octylFractal: No, that's what I'm saying. A `get()` on a `Provider` does not perform task execution at all. It's simply related to the value, which will still be there even if the file doesn't exist on disk. I'm reasonably confident that the `enabled` check is also evaluated before task execution in some cases (i.e. under CC), so I think doing an file existence check would not work.

PR #38918, opened 2026-08-21
Reuse dependency caches in ProjectBuilder

### LINKED ISSUE #38915: `ProjectBuilder` silently bypasses the dependency cache, causing full re-downloads per test run — now triggers Maven Central 429 blocks
## The problem

Projects created with `org.gradle.testfixtures.ProjectBuilder` cannot reuse the
on-disk dependency cache. `ProjectBuilderImpl` wires its services from
`TestGlobalScopeServices`, whose `createCacheFactory` returns
`TestInMemoryCacheFactory` — so every "persistent" cache, including the module
metadata store fronting `~/.gradle/caches/modules-2`, is an empty in-memory map
in each test JVM.

Consequently, any test that performs real dependency resolution through a
`ProjectBuilder` project re-downloads its entire dependency graph from remote
repositories on **every test run**, even when the machine's cache is fully
warm. `withGradleUserHomeDir(...)` does not change this: the user home,
resolver ids, and on-disk metadata all look correct while resolution still
reports a miss, which makes the behavior very hard to diagnose.

Until recently this cost only test time. Since Sonatype tightened Maven
Central's consumption limits (May 2026,
https://central.sonatype.org/faq/429-error/), it gets whole machines and CI
egress IPs blocked with HTTP 429/403: our plugin test suite (17 specs
resolving a real graph per run) got a developer machine blocked for ~3 hours,
unrelated builds included. Per the FAQ, tools that bypass dependency caches
are exactly the traffic pattern the enforcement targets — and `ProjectBuilder`
bypasses it invisibly.

We found this behavior documented nowhere: not in the `ProjectBuilder`
Javadoc, not in the plugin-testing guide, and in no existing issue.

## Current behavior

Full re-download of the resolved graph on every test run; no warning; all
observable cache state looks correct.

## Steps to reproduce

```kotlin
// In a JUnit test, run twice:
val project = ProjectBuilder.builder().build()
project.repositories.mavenCentral()
val config = project.configurations.create("probe")
project.dependencies.add("probe", "com.google.guava:guava:33.0.0-jre")
config.resolve()
```

Observe (e.g., with a network monitor) that the second run re-downloads
everything despite `~/.gradle/caches/modules-2` containing the artifacts.
Real-world diagnosis and a working mitigation are in
https://github.com/SpineEventEngine/core-jvm-compiler/pull/111 : a
build-prepared local stub repository plus
`gradle.startParameter.isOffline = true` on the stub project.

### Expected Behavior

Either of:

1. `ProjectBuilder` projects reuse the real dependency cache (read-only would
   suffice), or an opt-in exists to that effect; **or, at minimum:**
2. The `ProjectBuilder` Javadoc and the *Testing Gradle plugins* guide state
   plainly that dependency resolution in `ProjectBuilder` tests always
   bypasses the persistent caches and hits the network on every run — so
   users can make an informed choice before Central blocks their machines.

### Context (optional)

We have several subprojects that introduce Gradle plugins, and we test them using `ProjectBuilder` API. With the new policies at Maven Central we face the major re-work of the test of our Gradlew plugins.

https://github.com/SpineEventEngine/core-jvm-compiler/pull/111

### Gradle version

9.7.1

### Build scan URL (optional)

_No response_

- Gradle 9.7.1 (behavior present in earlier versions as well; the
  `TestInMemoryCacheFactory` wiring is long-standing)
- Reproduced on macOS and in GitHub Actions runners (Linux, Windows)

itsCodeTide:
Fixes #38915

## Problem

ProjectBuilder used TestInMemoryCacheFactory for persistent caches, preventing dependency metadata and artifacts from being reused between test runs.

## Changes

- Use DefaultCacheFactory for ProjectBuilder services.
- Document persistent cache behavior for withGradleUserHomeDir().
- Add an integration test that resolves a dependency, creates a second ProjectBuilder with the same Gradle User Home, enables offline mode, and verifies the dependency is reused without network access.

## Verification

- ProjectBuilder integration test passed in the existing checkout.
- Checkstyle passed.
- CodeNarc passed.
- git diff --check passed.


platforms/extensibility/unit-test-fixtures/src/main/java/org/gradle/testfixtures/ProjectBuilder.java
```diff
@@ -81,8 +81,8 @@ public ProjectBuilder withProjectDir(@Nullable File dir) {
     }
 
     /**
-     * Specifies the Gradle user home for the builder. If not set, an empty directory under the project directory
-     * will be used.
+     * Specifies the Gradle user home for the builder. Persistent Gradle caches, including the dependency cache,
+     * are stored in this directory. If not set, an empty directory under the project directory will be used.
```
hfhbd: IMHO there should also be a method to easily get the default gradle user home directory:
`withGradleUserHomeDir(getDefaultGradleUserHomeDir())`. Otherwise, you need to pass the variable from your build logic to the tests.

itsCodeTide: Thanks for the feedback. I added ProjectBuilder.getDefaultGradleUserHomeDir(), including @Incubating and the correct `@since` 9.8.0 metadata.

Callers can now share the default dependency cache with:

ProjectBuilder.builder()
    .withGradleUserHomeDir(ProjectBuilder.getDefaultGradleUserHomeDir())
    .build()

Explicit Gradle user homes use persistent caches, while the default temporary ProjectBuilder user home retains isolated in-memory caches for test isolation.

Added coverage for the new API and cache behavior. ProjectBuilder tests and binary compatibility checks pass. The signed commit has been pushed to this PR branch.

itsCodeTide:
I reproduced the reported plugins-model-native failures locally. The deepest error is a duplicate generated model class definition, not a bootstrap-classpath configuration error.

The same test task fails on untouched upstream/master with the same classloader/generated-model failures, so this is not introduced by PR #38918. The PR does not modify JVM arguments, boot classpaths, or native test configuration.

I have not added an unrelated workaround. Please rerun the CI jobs or classify this as a base-branch/environment failure.

itsCodeTide:
The latest logs show two unrelated baseline failures:

- unit-test-fixtures:projectHealth reports an existing dependency-analysis recommendation in build.gradle.kts. That file is unchanged by this PR.
- plugins-model-native:test reports duplicate generated model classes and leftover test files. I reproduced the same failures on untouched upstream/master.

The Develocity authentication message is informational and is not the build failure.

The ProjectBuilder cache changes remain validated by the unit-test-fixtures checks, integration tests, compile checks, binary compatibility, DCO, and documentation checks. Please classify or rerun the baseline-failing jobs, or let me know if a separate issue should be opened for them.

ov7a:

The fate of this PR will be decided by the team after discussion of [the issue](https://github.com/gradle/gradle/issues/38915).

reinsch82:
Thanks for working on this — making dependency resolution from ProjectBuilder tests fast by reusing caches addresses a real gap. However, the current approach has a few structural problems we need to resolve before this can move forward.

**1. `withGradleUserHomeDir(...)` must not implicitly opt into persistent caches.**
The change keys persistence on `gradleUserHomeDir != null`. That method has been public API since Gradle 3.0 and is widely used with throwaway temp dirs purely for isolation. All those callers would silently switch from in-memory to real file-locked caches: held `.lock` files break temp-dir cleanup on Windows, parallel test workers contend on exclusive locks, and resolution semantics change. Persistence should be an explicit opt-in, e.g. `withPersistentCaches()`, while `withGradleUserHomeDir` keeps meaning "location" only.

**2. Never resolve to the real `~/.gradle` implicitly.**
`getDefaultGradleUserHomeDir()` reads the test JVM's `gradle.user.home` system property / `GRADLE_USER_HOME` env var and falls back to the user's real home. Test workers don't inherit the invoking build's user home (`-g` travels as serialized internal config, never as a system property), so in practice this resolves to the real `~/.gradle` — a directory the invoking build may never have populated. Since ProjectBuilder runs Gradle embedded in the test JVM, the test process itself then holds the on-demand exclusive lock on `modules-2`; a test JVM paused in a debugger makes concurrent real builds on the machine fail with a 60s lock timeout.

I'd follow the TestKit precedent: persistent mode defaults to a dedicated per-user directory (analogous to TestKit's `.gradle-test-kit-<username>` under the temp dir), an explicitly provided home is honored, and the real home is only ever used when the user deliberately points there (and the javadoc warns about lock contention). Sharing the invoking build's cache stays possible by explicitly forwarding `gradle.user.home` to the test JVM — a deliberate act, not a fallback.

**3. Persistent caches need a lifecycle story.**
ProjectBuilder exposes only `build()`. The persistent registry introduced here is never closed, and user-home services are never released — the PR's own integration test needs the internal `org.gradle.internal.reuse.user.home.services` property plus internal `ProjectBuilderImpl.stop()` to clean up, levers users of the public API don't have. Consequences: held locks, non-daemon cache-worker threads and open caches leak for the JVM lifetime, cache cleanup never runs (a reused cache grows unboundedly), and pending async writes are never flushed on shutdown, forcing dirty-cache recovery in the next real build.

To be fair, this lifecycle gap predates this PR — ProjectBuilder has never released its services. But with in-memory caches the leak was process-private (heap, threads) and harmless at JVM exit, whereas real file-locked caches make the same gap machine-visible: held OS locks, open handles, a bound contention socket, and unflushed on-disk state. This change is what makes solving it a requirement.

Since the value of persistent caches is staying warm across tests, cleanup belongs at two levels:
- per test: an `AutoCloseable` project handle (or public equivalent of `ProjectBuilderImpl.stop`) releasing build-tree services;
- per JVM: a public static release method (for `@AfterAll`/`cleanupSpec`) that closes user-home registries and the cache factory, plus a shutdown hook registered when persistent mode first activates so abandoned JVMs still shut down cleanly.

Smaller items:
- `@since 9.8.0` will fail `checkBinaryCompatibility` — version.txt is 9.9.0.
- `TestGlobalScopeServices.createCacheFactory` should call `super.createCacheFactory(...)` for the persistent branch instead of hand-constructing `DefaultCacheFactory`, so it can't drift from production behavior.
- The new integration test also passes with in-memory caches (metadata bytes reach disk via the plain file store either way); it should assert on something only the real cache factory creates, e.g. the `.lock` file under `caches/modules-2`, or resolve in a fresh JVM.

Summary: explicit opt-in for persistence, a dedicated default home that is never the global one, and a public release/close mechanism. Happy to discuss the design here.

itsCodeTide:
Addressed the review feedback:

- Persistent caches are now opt-in through `withPersistentCaches()`.
- `withGradleUserHomeDir()` only selects the user-home location and no longer changes cache behavior.
- Persistent mode defaults to an isolated per-user temporary directory rather than the real ~/.gradle.
- Added public project and global-service cleanup APIs, with shutdown-hook cleanup.
- Updated API metadata to `@since` 9.9.0.
- Persistent cache creation now delegates to the production implementation.
- Integration coverage verifies the real cache lock and offline dependency reuse.

Validation: `:unit-test-fixtures:check` passes on top of current `upstream/master`.

reinsch82:
@bot-gradle test this

reinsch82:
Thanks for working on this.

After some internal discussion, we realized that the lifecycle story is more complex than I initially assumed. We'll probably need to go through some design iterations to get this right.
I'll follow up with more.

reinsch82:
see https://github.com/gradle/gradle/issues/38915#issuecomment-5662264906

PR #38923, opened 2026-08-21
Update signing configuration

blindpirate:
Backport of #38874 to `xperimental`.

After the move to a signing subkey, `useInMemoryPgpKeys` needs the key ID as its first argument — without it the signing plugin falls back to the primary key. The promotion build configuration also has to pass it as `env.PGP_SIGNING_KEY_ID`.

This branch needs it because `Gradle_Xperimental` has versioned settings enabled and syncs its TeamCity configuration from `xperimental`, so `Gradle_Xperimental_Promotion` currently has no `env.PGP_SIGNING_KEY_ID` — while the promotion build itself runs `gradle-promote` `master`, which already expects one.

Clean cherry-pick of the original commit (`4bdf0da`); no adaptation needed.

Docs: https://docs.gradle.org/current/userguide/signing_plugin.html#sec:subkeys
Part of gradle/gradle-private#5321


blindpirate:
Closing: we don't need the key-ID backport on `xperimental`.

Note for the record — `Gradle_Xperimental_Promotion` therefore keeps no `env.PGP_SIGNING_KEY_ID`, while its promotion builds run gradle-promote `master`, which passes the key ID to `useInMemoryPgpKeys`. If that line is ever promoted it will sign with the primary key rather than the subkey. Tracked in gradle/gradle-private#5321.

PR #38925, opened 2026-08-21
Reduce distribution size

ov7a:
Minify BouncyCastle and the embedded Kotlin compiler jars to reduce bin distribution size by 16.5 Mb from 144.6 Mb to 128.1 Mb (-11.4 %).

Note: the initial attempt was done with R8, but it quickly revealed its limitations in our scenarios, so ProGuard is used.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


ov7a:
@bot-gradle test RfR

ov7a:
@bot-gradle test RfR

build-logic-commons/basics/src/main/kotlin/gradlebuild/basics/transforms/Minify.kt
```diff
@@ -79,53 +96,92 @@ abstract class Minify : TransformAction<Minify.Parameters> {
...
+                        out.putNextEntry(ZipEntry(name))
+                        Files.copy(path, out)
+                        out.closeEntry()
+                    }
                 }
             }
-        } catch (exception: Exception) {
-            throw ClassAnalysisException("Could not write shaded Jar $jarFile", exception)
         }
+        partial.renameTo(library)
```
blindpirate: Shall we capture the return value of `renameTo` in case it fails?

blindpirate (approved):
LGTM with a tiny question

build-logic-commons/basics/src/main/kotlin/gradlebuild/basics/transforms/Minify.kt
ov7a: Thanks, fixed

PR #38931, opened 2026-08-22
Introduce ProviderFactory.absent(), present(T) and presentIfNotNull(@Nullable T)

### LINKED ISSUE #12623: Please add to ProviderFactory: <T> Provider<T> provider​(T value)
As we are introducing APIs built upon the `Provider`/`Property` types we sometimes find ourselves creating `Provider` instances around a static value.

Right now, this requires doing `providerFactory.provider(Callable{ value})`. It would be nice to not have to have to instantiate a `Callable`. An internal implementation of `Provider` with a static value would be a good addition.

### LINKED ISSUE #32724: Make Providers.of and Providers.ofNamed public available
### Expected Behavior

```kotlin
// myExtension.kt
class MyExtension {
  fun oldUsage(resolvableConfiguration: Configuration, withParam: Int) = newLazyFunction(Providers.ofNamed(resolvableConfiguration), Providers.of(withParam))

  fun newLazyUsage(resolvableConfiguration: NamedDomainObjectProvider<Configuration>, withParam: Provider<Int>) {
    // my logic
  }
}

// consumer.gradle.kts
val someOldApiFromAnotherPlugin = configurations.getByName("runtimeClasspath")
val someStaticValue = 42

val newRequiredApi = configurations.resolvable("foo")

myExtension {
  oldUsage(someOldApiFromAnotherPlugin, someStaticValue)

  newLazyUsage(newRequiredApi, provider(someStaticValue))
}
```

```kotlin
// myExtension.kt
class MyExtension {
  fun oldUsage(resolvableConfiguration: Configuration, withParam: Int) = realLogic(project.provider { resolvableConfiguration }, resolvableConfiguration.name, project.provider { withParam })

  fun newLazyUsage(resolvableConfiguration: NamedDomainObjectProvider<Configuration>, withParam: Provider<Int>) = realLogic(resolvableConfiguration, resolvableConfiguration.name, withParam)

  private fun realLogic(resolvableConfiguration: Provider<Configuration>, name: String, withParam: Provider<Int>) { 
    // my logic
  }
}

// consumer.gradle.kts
val someOldApiFromAnotherPlugin = configurations.getByName("runtimeClasspath")
val someStaticValue = 42

val newRequiredApi = configurations.resolvable("foo")

myExtension {
  oldUsage(someOldApiFromAnotherPlugin, someStaticValue)

  newLazyUsage(newRequiredApi, project.provider { someStaticValue })
}
```

### Context

I want to migrate the extension of my plugin to support Providers. I did update my logic to support providers only, but that changes requires the users to either use `project.providers` on the call side or writing a wrapper function that needs to wrap the raw values into Providers to support both cases until all users and their logic/applied plugins uses Providers only.

Also, the `provider` callback is lazy (and not stored in CC?), while the static function `Providers.of` could be stored.

bamboo:
Fixes #12623
Fixes #32724

# Wrap a known value in a `Provider` without paying for laziness

## What you get

Three new `ProviderFactory` methods (incubating, `@since 9.9.0`) for the case where you already
*have* the value and just need it in the Provider world:

```java
Provider<String> name    = providers.present("mylib");             // always present
Provider<String> missing = providers.absent();                     // always absent
Provider<String> maybe   = providers.presentIfNotNull(readEnv());  // present iff non-null
```

**Before:** the only general-purpose factory was `providers.provider(() -> value)` — a `Callable`
wrapped around a constant. It reads as lazy but isn't, and every query (`get()`, `isPresent()`,
`orElse(...)`) re-runs the callable through the provider-evaluation machinery, including its
re-entrancy guard. "No value" had no direct spelling at all: `provider(() -> null)` recomputes its
absence on every presence check.

**Now:** `present(value)` returns a fixed-value provider whose queries return the value at constant
cost; `absent()` returns the shared no-value provider; `presentIfNotNull(value)` picks between the
two for you — the natural bridge from nullable Java/Kotlin values into lazy APIs.

The names are the two states a provider can be in. The `Provider` javadoc now says so explicitly —
"a provider that has a value is said to be *present*, and a provider that has no value is said to
be *absent*" — right where it first explains that a value may be unavailable, and lists
`present(T)`/`absent()` among the common ways to create a provider. So a reader meeting
`providers.absent()` in `ProviderFactory` already knows the word from `Provider.isPresent()`.

`present(value)` also returns the new **`PresentProvider<T>`** — a `Provider<T>` subtype that encodes
"always present" in the type: `get()` is documented to never fail, `getOrNull()` loses its
`@Nullable` (Kotlin sees a non-null `orNull`), and both `orElse(...)` overloads return the provider
itself. Plugin authors can use it in their own signatures to tell consumers a value is always set,
with the compiler holding the promise.

Nothing to set up or migrate: the API is additive and incubating, existing code is untouched.

## How it behaves

- `present(null)` fails eagerly (`IllegalArgumentException`), mirroring `provider(null)`;
  `presentIfNotNull` is the null-tolerant spelling.
- `absent()`'s provider supports the full `Provider` contract (`orElse`, `map`, …) — `map` on it is
  a no-op, `orElse` yields the fallback.
- Transform results deliberately stay plain `Provider`: `map`/`filter`/`flatMap`/`zip` may produce
  absence (a transformer can return null), so `PresentProvider` refines only what is provably
  present — `get`, `getOrNull`, `getOrElse`, `isPresent`, `orElse`.
- Configuration Cache treats the returned providers as the fixed values they are — they serialize
  as plain values, and a `PresentProvider`-typed reference survives a cache round trip.

## For reviewers

Small additive commits (~240 insertions): (1) the three `ProviderFactory` methods, implemented in
`DefaultProviderFactory` over the existing internal `Providers` factories; (2) `PresentProvider`,
implemented by the internal fixed-value provider, whose `orElse` overloads already returned `this`;
(3) two rename commits recording how the names settled (`none/some/nullable` → `maybe` →
`absent/present/presentIfNotNull`); (4) a documentation-only commit that defines "present" and
"absent" in the `Provider` javadoc and phrases the new methods in those terms — kept separate so the
wording can be iterated on its own. Binary-compatibility, architecture tests, and the provider unit
suites are green; no accepted-API-change entries needed (the methods are unreleased). Deferred:
adopting `PresentProvider` elsewhere in the public API (e.g. `Provider.orElse`) — return-type
changes on existing methods are breaking, so out of scope here.

### Reviewing cheatsheet

- ❌ ❓**must** be fixed
- 🤔 💅 **should** be fixed
- 💭 **may** be fixed
- 🎉 celebrate happy things


bamboo:
@bot-gradle test

subprojects/core-api/src/main/java/org/gradle/api/provider/PresentProvider.java
```diff
@@ -0,0 +1,93 @@
...
+ * {@link #getOrNull()} never returns {@code null} and {@link #isPresent()} always returns {@code true}.
+ * Only the presence of the value is guaranteed; the value itself is not necessarily constant.
+ *
+ * <p><b>Note:</b> This interface is not intended for implementation by build script or plugin authors.
+ * A provider that always has the given value can be created via {@link ProviderFactory#some(Object)}.
+ *
+ * @param <T> Type of value represented by the provider
+ * @since 9.8.0
+ */
+@Incubating
```
asodja, line 38: 💅 We could add:
```
@HasInternalProtocol
@NonExtensible
```

the same way as on Provider. In JavaDoc is already mentioned it should not be extended

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/AbstractProviderWithValue.java
```diff
@@ -16,12 +16,13 @@
...
 package org.gradle.api.internal.provider;
 
+import org.gradle.api.provider.PresentProvider;
 import org.gradle.api.provider.Provider;
 
 /**
  * A {@link org.gradle.api.provider.Provider} that always has a value defined. The value may not necessarily be final.
  */
-public abstract class AbstractProviderWithValue<T> extends AbstractMinimalProvider<T> {
+public abstract class AbstractProviderWithValue<T> extends AbstractMinimalProvider<T> implements PresentProvider<T> {
```
asodja: ❓ 🤖 `PresentProvider` promises that querying "never fails", but two of the four subclasses of this class can't honour that:
  `DelegatingProviderWithValue.calculateOwnValue` throws `IllegalStateException(nonPresentMessage)` when the delegate is absent (used by `DefaultMutableAttributeContainer:97`), and `BuildableBackedProvider.calculateOwnValue` throws `NullPointerException` when the factory returns  null.

Can we double check this or weaken the JavaDoc or move interface just to FixedValueProvider?

subprojects/core-api/src/main/java/org/gradle/api/provider/PresentProvider.java
```diff
@@ -0,0 +1,93 @@
...
+public interface PresentProvider<T> extends Provider<T> {
```
asodja, line 39: ❌ 🤖 From AI review:

The PR description says "a `PresentProvider`-typed reference survives a cache round trip". Two decode paths in
  `FixedValueReplacingProviderCodec.decodeValue` (`ProviderCodecs.kt:139-151`) return non-`PresentProvider` implementations:

  - byte `2` → `ExecutionTimeValue.ofNullable(read())`, with the existing comment *"nullable because serialization may replace value with null, e.g.
  when using provider of Task"*. `TaskReferenceCodec.decode()` returns `null` for `PROHIBITED` (`TaskReferenceCodec.kt:59-63`), so
  `providers.some(otherTask)` decodes to `missing()` → `Providers.notDefined()` → `NoValueProvider extends AbstractMinimalProvider`.
  - byte `3` → `ExecutionTimeValue.ofNullable(v).withSideEffect(...)`, whose `toProvider()` yields `WithSideEffectProvider extends
  AbstractMinimalProvider` (`ValueSupplier.java:844`).

  Restoring either into a `PresentProvider`-typed field fails on the reflective set instead of surfacing the existing CC problem message. Path 2
  can't arise from `some()` itself; path 1 can. 

At minimum this needs a CC integration test - there's no integration-level test in the PR at all.

Can we validate this?

subprojects/core-api/src/main/java/org/gradle/api/provider/ProviderFactory.java
```diff
@@ -77,6 +78,51 @@ public interface ProviderFactory extends GradleService, ProjectService, Settings
...
+     *
+     * @param value The value of the provider. May be null.
+     * @param <T> the type of the value of the provider
+     * @return The provider. Never returns null.
+     * @see #none()
+     * @see #some(Object)
+     * @since 9.8.0
+     */
+    @Incubating
+    <T> Provider<T> nullable(@Nullable T value);
```
asodja: 🤔 Would `ofNullable` be more common name for such methods, then non-nullable could be `ofValue()` and would look more symetric. Although `nullable` may be more consistent with other methods that exists here, like `environmentVariable`, `systemProperty` etc

I guess some/none comes from functional vocabulary that is not that common in Java/Kotlin

asodja:
Nice addition to ProviderFactory 🎉

subprojects/core-api/src/main/java/org/gradle/api/provider/ProviderFactory.java
bamboo: > Although nullable may be more consistent with other methods that exists here

That's what I thought too.

subprojects/core-api/src/main/java/org/gradle/api/provider/PresentProvider.java
bamboo: `@NonExtensible` shouldn't be unnecessary but it's harmless, will do.

platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/AbstractProviderWithValue.java
bamboo: It seems `DelegatingProviderWithValue` violates the `AbstractProviderWithValue` invariant then. But yes, in any case, restricting it to `FixedValueProvider` is better.

subprojects/core-api/src/main/java/org/gradle/api/provider/PresentProvider.java
bamboo: I added integration tests.

alllex:
The intent behind the new API is clear, and I support adding the new methods as they serve as basic yet important building blocks in the Provider-algebra. However, the naming scheme `none/some/nullable` for the methods is unconventional for the Java world and has to be justified or at least compared to the alternative options.

Firstly, there is no prior API in Gradle that would have used a similar naming approach, making it somewhat of a novelty. Moreover, neither JDK nor Kotlin uses this vocabulary for similar methods. JDK has established `empty/of/ofNullable` for `Optional` and `Stream`, which we should likely treat as the default choice and argue on why it's not applicable. Groovy and Kotlin do use `none/some` in their vocabulary, but it's for _collection predicates_. I can infer some reasoning myself, but I would be curious about an explicit answer either way.

The `empty` doesn't seem to work for Providers, because we instead always had a concept of _absence_. What's "worse", we've also used `empty` as an operation on container-like Providers that clears the container to make it empty, yet the Provider itself has value and therefore is not absent.

The `of`, as I've noted in the past discussions, was an unfortunate choice of the method name for the `ProviderFactory.of(ValueSource.class, Action)`. Though, if we only have the two-arg version of it, we might still use the `.of(valueInstance)` as the alternative for `.some(valueInstance)`.

Next, the `none/some/nullable` doesn't align with the vocabulary of absent/present providers. The presence has a direct representation in the API -- `Provider.isPresent`. There is no explicit API for the _absence_. However, the user-facing documentation as well as the user-facing errors have consistently used the wording "[provider] _has no value_". Again, if we are departing from this naming, we should justify it.

The JDK's approach to naming is nice, because it creates consistency between the predicate and the construction: `Optional.isEmpty()` vs `Optional.empty()`. Hence, an option to consider is `ProviderFactory.present()`, which makes it directly aligned with the naming and, importantly, with the returned `PresentProvider` type.

As a point of reference, Guava's Optional has `absent/of/fromNullable`. Also, it seems none of the known libraries go for just `nullable` and instead it's `ofNullable` or `fromNullable`.

Another thing to consider is the kinds of methods that already live on the `ProviderFactory`. Most of the methods are within the domain of the source of or type of the underlying information (env vars, creds, props, etc). Only few methods are truly generic: `provider` and `zip`. Only `provider` is about generic construction. Therefore, I'd say that it doesn't necessarily establish enough of a precedent for the naming convention of the provider-constructing methods, and we shouldn't necessarily align the new API to it, and instead call it an exception in the naming, after we've introduced the new methods. This is to the point that it might feel like the new methods must be noun-like, describing the result, but I argue that it's not the case.

Last but not least, we are making these methods as instance methods on a provider factory as opposed to making them static on `Provider`. There are many good reasons for this approach, such as discoverability and making people standardize on the use of the `ProviderFactory`. But the callers will be treating these methods as static construction methods anyway. So the strong force in the naming scheme should come from how we'd name the methods in the static context. There, again, `none/some/nullable` doesn't feel right, at least to me.

I don't know what the best option is. However, without deeper discussion, it seems we can add even more variety to the salad of Provider-related concepts, and I'd rather we consolidate instead, even if we have to sacrifice brevity for consistency.

bamboo:
> But the callers will be treating these methods as static construction methods anyway.

I don't see why. They will be coming from a particular context in the configuration model tree so they could be carrying context (even if it's just provenance).

> The intent behind the new API is clear, and I support adding the new methods as they serve as basic yet important building blocks in the Provider-algebra. However, the naming scheme none/some/nullable for the methods is unconventional

Do the names convey the intended meaning? I say they do. And it doesn't seem you have a concrete suggestion for all three. Can you make a concrete proposal of your preferred names for the three methods in a single place so we can assess how they work together?

subprojects/core-api/src/main/java/org/gradle/api/provider/ProviderFactory.java
bamboo: I renamed `nullable => maybe` for consistency with the `none` / `some` terminology.

bamboo:
@bot-gradle test

jvandort:
I agree with Alex in that these names just don't feel very "Java-y". Your names do the job just fine, but vibes are real and I think going with a familiar naming scheme would be best.

Based on gut feel I would have expected empty,ofNullable,of.

bamboo:
> these names just don't feel very "Java-y"

Have you ever considered that could actually be a good thing? ;)

> empty,ofNullable,of.

`of` is already taken here and it's out of reach. `ofNullable` is out of character in an interface method and would hint at a nonexistent relationship with `of`. `empty` is already taken for collections to mean _present with no elements_.

In any case, the team will have a call this week to settle on the right triple and we will be considering other candidates.

subprojects/core-api/src/main/java/org/gradle/api/provider/ProviderFactory.java
bamboo: The methods now carry the new names the team chose.

bamboo:
@bot-gradle test this

asodja (approved):
LGTM

Issue #1843, opened 2017-04-18
Terminal automatically scrolls down on never ending tasks with the new console


felipefzdz:
Regarding the console, the new timer on `98% EXECUTING [1m 51s]` is pretty cool. However, when executing tasks that conceptually never end (like our build scan service), there is some UX caveat. Let's say that there was some interesting log message during the startup of my service. If I want to read it, I'll scroll up on the CLI. However, as that clock is continuously updated, the terminal (at least Terminator on Linux Mint), automatically scrolls down to that changing output, i.e. I need to stop the service/kill the build to being able to read the logs. Not sure if Gradle is aware of 'infinite' tasks, so then it can disable that timer for that use case.

eriwen:
@felipefzdz You mean using `--continuous` build?

felipefzdz:
@eriwen that might be also, but I was referring to whatever task that triggers something designed to be executed without a fixed end, like an application server. In the particular example that I faced, I was running a Ratpack service through a [JavaExec](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.JavaExec.html) task.

eriwen:
Thanks for the clarification. For now, we recommend using `--console plain` until we come up with something better. Ideally, we'd have logs available outside the build output, even more ideally in a build scan.

felipefzdz:
Cool, that's a good workaround. Thanks!

eriwen:
Let's handle this with the solution proposed in https://github.com/gradle/gradle/issues/2252 and close this as a duplicate

Issue #1842, opened 2017-04-18
Expose public API for declaring a SoftwareComponent for publishing


bmuschko:
Plugin developers tend to use internal types for declaring software components that should be published. Any change in the internal APIs breaks external plugins.

- https://github.com/novoda/bintray-release/issues/112
- https://github.com/johnrengelman/shadow/issues/287

### Expected Behavior

- Expose a public API for declaring a `SoftwareComponent` as publishable artifact. Make it easy to declare dependencies to be included in the POM similar to the internal definitions of components.java` and `components.web`.
- The functionality will be available for the Maven and Ivy publish plugins.
- Expand the user guide for both plugins.
- Write a getting started guide on how to publish custom artifacts that avoids the use of internal APIs in the interim.

### Current Behavior

The method [`MavenPublication.from(SoftwareComponent)`](https://docs.gradle.org/3.4.1/dsl/org.gradle.api.publish.maven.MavenPublication.html#org.gradle.api.publish.maven.MavenPublication:from(org.gradle.api.component.SoftwareComponent)) only allows the use of predefined software components e.g. `components.java` and `components.web`. A user cannot define own types of `SoftwareComponent` without using the internal API as the type is internally casted to [`SoftwareComponentInternal`](https://github.com/gradle/gradle/blob/master/subprojects/maven/src/main/java/org/gradle/api/publish/maven/internal/publication/DefaultMavenPublication.java#L119). As a result external plugins use `SoftwareComponentInternal` and `UsageContext` which have moved to a different package in 3.4 breaking any plugin depending on it.

The public API for assigning artifacts with [`MavenPublication:artifact(Object)`](https://docs.gradle.org/3.4.1/dsl/org.gradle.api.publish.maven.MavenPublication.html#org.gradle.api.publish.maven.MavenPublication:artifact(java.lang.Object)) and modifying the POM with the help of `withXml` can lead to code that's hard to maintain.

### Context

Users want to define their own publishable artifacts e.g. an AAR file that needs to include runtime dependencies as well. AAR files are not a predefined type.

adammurdoch:
We should move the public API towards variant aware dependency resolution, so that the API allows the user to declare variants of their software, that can be referenced either locally via a project dependency or used directly as an input or added to a publication for publishing.

wilkinsona:
[...]

PR #1841, opened 2017-04-17
Minor performance improvements to the task execution plan

ghale:

PR #1840, opened 2017-04-17
Remove deprecated method in IdeaModuleDependency

w25r:
### Context
Fixes https://github.com/gradle/gradle/issues/1724

### Contributor Checklist
- [x] Update User Guide, DSL Reference, and Javadoc for public-facing changes
- [x] Ensure that tests pass locally: `./gradlew quickCheck <impacted-subproject>:check`

### Gradle Core Team Checklist
- [ ] Verify design and implementation 
- [ ] Verify test coverage and [CI build status](https://builds.gradle.org/project.html?projectId=Gradle_Branches_Checkpoints&tab=projectOverview&branch_Gradle_Branches_Checkpoints=wl-tooling-api-dep-removal-take2)
- [ ] Verify documentation including proper use of `@since` and `@Incubating` annotations for all public APIs

PR #38932, opened 2026-08-24
Defer signing wiring until projects are evaluated

blindpirate:
Nightly promotion of `master` has been red since 2026-08-20 (#1840, #1841, #1842, #1843), so there
have been no new master nightlies since the 2026-08-19 build. `Publish Kotlin DSL Plugin` is red for
the same reason. Both fail during **configuration**:

```
* Where:
Build file '.../platforms/core-execution/build-cache/build.gradle.kts' line: 10

* What went wrong:
Cannot mutate the dependencies of configuration ':build-cache:api' after the configuration's child
configuration ':build-cache:apiElements' was published as a variant. After a configuration has been
observed, it should not be modified.
```

## Why

As described in #38917: #38424 made `GenerateMavenPom`'s `pom` a `@Nested` input, so realizing the
task's output files also realizes the pom's nested dependencies provider, which finalizes the
variants of the published component. Signing eagerly realizes a publication's artifact files, so
`sign(Publication)` now freezes the project's dependency configurations at configuration time.

Our publishing plugins call `signing.sign(publication)` while the plugin is being applied — that is,
from the `plugins { }` block of every published module, *before* that module's `dependencies { }`
block runs. Hence the failure above, and hence it only affects builds where signing is active
(`PGP_SIGNING_KEY` set), which is why ordinary master CI is green and only promotion/publishing jobs
are broken.

#38917 fixed the build tool on 2026-08-21, but that does not unblock promotion on its own: the
promotion build bootstraps with the wrapper's **released** distribution, currently
`9.8.0-milestone-1`, which predates the fix. And no newer 9.8 nightly exists to bump to, because
nightlies are produced by the very job that is broken — the latest published nightly is still
`9.8.0-20260819011813+0000`. So the build logic has to stop tripping over the released wrapper.

## What this does

Defers wiring `signing.sign(...)` until the project has been evaluated, in the three places that set
it up, and only when signing is actually enabled — so nothing changes for non-signing builds.

`afterEvaluate` is already used in build logic for the same "must run after the applying script"
reason (`gradlebuild.minify`, `gradlebuild.no-module-annotation`, `JvmCompilation`).

The alternative would be reverting the wrapper to 9.7.1, but the wrapper bot bumps to the latest
released version, so it would go straight back to `9.8.0-milestone-1` and break again.

## Verification

Ran the promotion command locally on this branch with the `9.8.0-milestone-1` wrapper
(`clean promotionBuild --dry-run -Dorg.gradle.unsafe.isolated-projects=false --no-configuration-cache`,
with dummy artifactory credentials so the `taskGraph.whenReady` credential check passes):

| | result |
|---|---|
| `PGP_SIGNING_KEY*` set, before this change | fails at `platforms/core-execution/build-cache/build.gradle.kts` line 10, as on CI |
| `PGP_SIGNING_KEY*` set, after this change | `BUILD SUCCESSFUL`, full task graph built, **29 `sign*Publication` tasks** present |
| no `PGP_SIGNING_KEY*`, after this change | `BUILD SUCCESSFUL`, 0 sign tasks and 32 `publish*PublicationTo*` tasks — unchanged from before |

So signing is still wired into the publishing graph, and non-signing builds are untouched.

Also confirmed on a 10-line standalone reproducer that this is a wrapper-version behaviour change:
`java-library` + `maven-publish` + `signing`, a publication with
`versionMapping { usage("java-api") { fromResolutionResult() } }`, `sign()` wired before
`dependencies { api(...) }` — fails on 9.8.0-milestone-1, passes on 9.7.0.

## Follow-up

This workaround can be removed once the wrapper points at a distribution that contains #38917.

Internal tracking: gradle/gradle-private#5323

### DIFF
--- a/build-logic-commons/publishing/src/main/kotlin/gradlebuild.kotlin-dsl-plugin-bundle.gradle.kts
+++ b/build-logic-commons/publishing/src/main/kotlin/gradlebuild.kotlin-dsl-plugin-bundle.gradle.kts
@@ -177,8 +177,14 @@ tasks.withType<Sign>().configureEach { isEnabled = signArtifacts }
 
 signing {
     useInMemoryPgpKeys(pgpSigningKeyId.orNull, pgpSigningKey.orNull, pgpSigningPassPhrase.orNull)
-    publishing.publications.configureEach {
-        if (signArtifacts) {
+}
+
+// See the comment on the same workaround in gradlebuild.publish-public-libraries: signing a
+// publication eagerly finalizes the published component, so it must not be wired up before the
+// applying build script has declared its dependencies.
+if (signArtifacts) {
+    afterEvaluate {
+        publishing.publications.configureEach {
             signing.sign(this)
         }
     }
--- a/build-logic-commons/publishing/src/main/kotlin/gradlebuild.publish-public-libraries.gradle.kts
+++ b/build-logic-commons/publishing/src/main/kotlin/gradlebuild.publish-public-libraries.gradle.kts
@@ -56,8 +56,22 @@ signing {
         pgpSigningKey.orNull,
         pgpSigningPassPhrase.orNull
     )
-    publishing.publications.configureEach {
-        if (signArtifacts) {
+}
+
+// Signing eagerly realizes the artifact files of a publication, which finalizes the variants and
+// the dependencies of the published component. Wiring signing up while this project is still being
+// configured therefore freezes its dependency configurations, and the `dependencies { }` block of
+// the build script applying this plugin then fails with "Cannot mutate the dependencies of
+// configuration ':x:api' after the configuration's child configuration ':x:apiElements' was
+// published as a variant". Deferring the wiring until the project is evaluated avoids that.
+//
+// The build tool side of this was fixed by https://github.com/gradle/gradle/pull/38917, but the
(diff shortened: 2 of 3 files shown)


blindpirate:
@bot-gradle test tf

blindpirate:
@bot-gradle test qf

blindpirate:
End-to-end verification on the real promotion pipeline: **Publish Branch Snapshot (from Quick Feedback)** [#551 / 116578599](https://builds.gradle.org/buildConfiguration/Gradle_Master_Promotion_PublishBranchSnapshotFromQuickFeedback/116578599) — green in 12m29s, promoting this branch's commit `d5688b3239134ddd13008060ac14e50c15899786`, published as `9.8.0-branch-boFixPromotionEagerSigning-20260824020331+0000`.

That run exercises exactly what nightly promotion does — same inner command (`clean promotionBuild` with the released `9.8.0-milestone-1` wrapper) and the real signing subkey:

- 0 occurrences of `Cannot mutate the dependencies of configuration ...` (nightly #1840–#1843 fail here)
- **29 `sign*Publication` tasks executed** — same count as in the task graph before the change, so signing is genuinely still wired up, not silently skipped
- `.asc` signatures uploaded for `gradle-9.8.0-{all,bin,src,docs}.zip` and `gradle-wrapper.jar`
- all three promotion steps exited 0

PR #38935, opened 2026-08-24
Fix JUnit platform launcher version for custom test suites

### LINKED ISSUE #38920: Custom `JvmTestSuite` loses the JUnit Platform launcher version with `mavenLocal()` first
### Current Behavior

The custom `integrationTest` suite uses Gradle's default JUnit Jupiter toolchain. Gradle adds a versionless `org.junit.platform:junit-platform-launcher` runtime dependency and expects the JUnit BOM constraint to supply its version.

### Expected Behavior

The default framework dependencies of a custom `JvmTestSuite` should resolve consistently when `mavenLocal()` appears before Maven Central. Repository order should not turn Gradle's default JUnit Platform launcher dependency into a dependency with an empty version.

## Prerequisite

Maven local must contain `org.junit.jupiter:junit-jupiter:5.12.2`. Maven stores its POM and JAR, but not Gradle's `.module` metadata. To populate Maven local when needed:

```shell
mvn dependency:get -Dartifact=org.junit.jupiter:junit-jupiter:5.12.2
```

## Reproduce the failure

Use Gradle 9.7.0:

```shell
gradle --version
gradle --refresh-dependencies resolveIntegrationTestRuntime \
  --configuration-cache \
  -Dorg.gradle.unsafe.isolated-projects=true
```

The build fails while Gradle stores the configuration cache. Its output contains:

```text
Could not find org.junit.platform:junit-platform-launcher:.
```

## Reverse repository order

Run the same task with Maven Central first:

```shell
gradle --refresh-dependencies -PmavenLocalFirst=false resolveIntegrationTestRuntime \
  --configuration-cache \
  -Dorg.gradle.unsafe.isolated-projects=true
```

This resolves `junit-platform-launcher:1.12.2` through the JUnit BOM constraint exposed by Maven Central's Gradle module metadata.

https://github.com/snazy/gradle-jvm-test-suite-metadata-repro

### Gradle version

9.7.0

### Build scan URL (optional)

_No response_

_No response_

itsCodeTide:
Fixes #38920

## Problem

Custom `JvmTestSuite`s using JUnit Jupiter add
`junit-platform-launcher` without a version. When `mavenLocal()` is first and
Gradle module metadata is unavailable, the JUnit BOM constraint is not found,
causing dependency resolution to fail.

## Solution

Add the matching JUnit BOM as a regular platform dependency whenever JUnit
Jupiter is configured. This makes the JUnit Platform launcher version
available independently of Gradle module metadata.

## Tests

- `:plugins-jvm-test-suite:compileJava`
- `:testing-jvm:compileIntegTestGroovy`
- Filtered `embeddedIntegTest`
- `:plugins-jvm-test-suite:checkstyleMain`

All passed offline.


itsCodeTide:
The regression test verifies that a custom JUnit Jupiter suite resolves
junit-platform-launcher:1.12.2 through the JUnit Jupiter BOM.

ov7a:

big-guy:
@itsCodeTide This fixes this one particular occurrence of this problem, but there's a larger problem. 

This shows up when using mavenLocal, but it can occur for other reasons, like a repository mirror that does not serve .module files. 

Like I said in #38920, putting mavenLocal() last (or not using it at all) is the easiest fix for this.

We won't be able to include these changes in 9.8.0.

itsCodeTide:
Thanks for the clarification. I understand that the root cause is broader than JUnit and can occur with any repository that does not provide Gradle module metadata. I’ve removed the JUnit-specific workaround and kept the general missing-version diagnostic with guidance to use complete metadata or place mavenLocal() after other repositories. I also understand this will target a release after 9.8.0.

PR #38937, opened 2026-08-24
Update Gradle wrapper to 9.8.0-milestone-2 and drop the signing deferral

blindpirate:
Two changes that belong together:

1. **Wrapper `9.8.0-milestone-1` → `9.8.0-milestone-2`.** M2 was promoted by [Release - Milestone #85](https://builds.gradle.org/buildConfiguration/Gradle_Master_Promotion_Milestone/116591954) and contains #38917.
2. **Revert of #38932** — the temporary signing deferral in `gradlebuild.publish-public-libraries`, `gradlebuild.kotlin-dsl-plugin-bundle` and `packaging/public-api`.

## Why now

#38932 worked around the fact that the released 9.8.0-milestone-1 wrapper predates #38917: signing eagerly realized a publication's artifact files, which finalized the published component's variants, so wiring `signing.sign(publication)` at plugin-apply time froze each module's dependency configurations and every published module failed to configure. That broke nightly promotion (#1840–#1843) and `Publish Kotlin DSL Plugin`.

There was no distribution to bump to at the time — nightlies are produced by the very job that was broken — so the build logic had to stop tripping over the released wrapper. With M2 out, the underlying fix is in the wrapper and the workaround is obsolete.

It should not be kept: Gradle's own best practice, `best_practices_general.adoc#avoid_after_evaluate` (added in 9.6.0), says *"Do not use `project.afterEvaluate {}` to configure tasks"* and *"afterEvaluate should be a last resort, not a first choice."* Wiring Sign tasks is exactly that, so it was only ever a stopgap.

## Verification

Ran the promotion command locally with signing enabled and the workaround removed, on the M2 wrapper
(`clean promotionBuild --dry-run -Dorg.gradle.unsafe.isolated-projects=false --no-configuration-cache`,
dummy artifactory credentials so the `taskGraph.whenReady` credential check passes):

`BUILD SUCCESSFUL` in 15m5s on `gradle-9.8.0-milestone-2`, with:

- **0** occurrences of `Cannot mutate the dependencies of configuration ...`
- **29 `sign*Publication` tasks** in the task graph — the same count as with the workaround in place, so
  signing is still wired up without it

So M2 does carry #38917, and the deferral is no longer needed. Nothing else in the repo referenced
`9.8.0-milestone-1`.

For contrast, the same command on this branch with the M1 wrapper is what produced the original
`Cannot mutate the dependencies of configuration ':build-cache:api' ...` failure.

Refs: #38932, #38917, #38424, gradle/gradle-private#5323

### DIFF
--- a/build-logic-commons/publishing/src/main/kotlin/gradlebuild.kotlin-dsl-plugin-bundle.gradle.kts
+++ b/build-logic-commons/publishing/src/main/kotlin/gradlebuild.kotlin-dsl-plugin-bundle.gradle.kts
@@ -177,14 +177,8 @@ tasks.withType<Sign>().configureEach { isEnabled = signArtifacts }
 
 signing {
     useInMemoryPgpKeys(pgpSigningKeyId.orNull, pgpSigningKey.orNull, pgpSigningPassPhrase.orNull)
-}
-
-// See the comment on the same workaround in gradlebuild.publish-public-libraries: signing a
-// publication eagerly finalizes the published component, so it must not be wired up before the
-// applying build script has declared its dependencies.
-if (signArtifacts) {
-    afterEvaluate {
-        publishing.publications.configureEach {
+    publishing.publications.configureEach {
+        if (signArtifacts) {
             signing.sign(this)
         }
     }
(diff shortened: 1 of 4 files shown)


blindpirate:
@bot-gradle test qf

ov7a (approved):

blindpirate:
Local verification finished: `clean promotionBuild --dry-run` with signing enabled, on the `9.8.0-milestone-2` wrapper and with the deferral reverted — `BUILD SUCCESSFUL` in 15m5s, **0** `Cannot mutate` occurrences, **29 `sign*Publication` tasks** in the graph (same count as with the workaround in place). M2 carries #38917, so the revert is safe. PR body updated.

